PR #5, opened 2015-04-26
Fixes to make ./configure --cover work.

daxtens:
This fixes building with ./configure --cover on x86_64 (and probably other architectures).


Dead2:
These seem like obvious fixes, thanks :)

PR #6, opened 2015-04-26
Run tests on Travis CI with .travis.yml

daxtens:
I'm in the process of building a test suite and it would be good if it ran automatically on Travis. This gets the existing set of tests to run. We get builds on both GCC and Clang, which is a nice win.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,5 @@
+language: c
+compiler:
+  - gcc
+  - clang
+script: ./configure && make && make check


Dead2:
A test-suite to catch regressions is very much needed, and although I myself have never used Travis CI before, I am very eager to see a good testsuite implemented for zlib. Hope to see followup patches to this.
Will pull.

daxtens:
Follow-up patches are indeed coming over the next few days :)

PR #9, opened 2015-04-29
Add security related tests

daxtens:
This provides test cases to cover 4 out of the 5 CVEs for zlib[0].

It also cleans up the build system, mostly around testing.

[0] http://www.cvedetails.com/vulnerability-list/vendor_id-72/product_id-1820/GNU-Zlib.html


daxtens:
There's more to come in terms of a test corpus, but I think this is an excellent starting point.

daxtens:
And now we cover all 5 known CVEs.

daxtens:
On second thoughts, a whole set of bugs have been pointed out to me, so maybe hold off on this for a couple of days until I fix it.

Dead2:
I'd like to get the test framework merged so development on more tests can continue.
What is stopping us from merging these now?

daxtens:
Cmake parity! I'm working on one final patch to make that work, I'm just
blocked on cmake not properly invoking testCVEinputs.sh out of tree. I'm a
bit squished for time to do that, but it's on the way. You can merge them
without out if you want, but if you could just wait until I check I've
pushed the latest version that'd be great. I'll try to ensure that's pushed
by the end of my day.

daxtens:
Branch is pushed. Cmake doesn't test stuff yet: I have a WIP patch but it's broken for cmake out of tree so I haven't pushed it yet.

daxtens:
woops, no wait, that breaks travis.

daxtens:
What have you done to gzfile? https://travis-ci.org/daxtens/zlib-ng/jobs/62321077#L217

Dead2:
No need to test gzfile api if it is disabled (WITH_GZFILEOP)

PR #14, opened 2015-05-03
Fix building out of source tree.

### LINKED ISSUE #1: Error when trying to configure and build out of tree
Calling the configure script from another directory in order to build out of tree will create an error :
cannot open zlib.h: No such file

mtl1979:
Fixes Dead2/zlib-ng#1
- Adds **SRCDIR** variable to _configure_, and _Makefile.in_ files in root directory and subdirectories.
- _configure_ duplicates directory tree from source directory inside build directory
- _configure_ reads from source directory and writes to build directory preserving the relative path.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,8 +37,10 @@ CMakeCache.txt
 CMakeFiles
 Testing
 *.cmake
+zconf.h.included
 
 configure.log
+a.out
 
 /arch/arm/Makefile
 /arch/generic/Makefile
--- a/Makefile.in
+++ b/Makefile.in
@@ -40,7 +40,9 @@ TAR=tar
 SHELL=/bin/sh
 EXE=
 
-ARCHDIR=
+SRCDIR=.
+
+ARCHDIR=arch/generic
 ARCH_STATIC_OBJS=
 ARCH_SHARED_OBJS=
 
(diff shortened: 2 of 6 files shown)


Dead2:
Looks pretty good.

But I am wondering, wouldnt it be more clear if SRCDIR always pointed to the source rootdir?
That would mean changes in the subdir makefiles, but they would then reference actual directorynames instead of ../../ stuff.

Like the patch is now, the ../../ stuff can reference outside of the source tree, such as /usr/ instead of /usr/src/zlib/. This can be fixed in other ways too, but the above would be a win for readability and consistency at the same time.

mtl1979:
If SRCDIR would point to root of the source tree, then it would be inconsistent with paths in the build tree, as paths in the build tree have to be relative because the absolute path of build root is never actually stored in a variable.
I was considering using two variables, one for current directory inside the source tree and one for root directory of the source tree, but that wouldn't work as it would really only work with three variables, the third being root of the build directory tree. That's kind of too many variables to keep track with.
The third option is kinda something in between, store just absolute path of root directory of both source tree and build tree in a variable and then use those to construct current location in both source and build directory trees.

Dead2:
Hmm, that is true.
What about removing the old "-I../../" from the submakefiles?
That should not be needed anymore, or am I missing something? I am not currently able to experiment with the code, otherwise I would just test it myself.

mtl1979:
-I../.. is used to find zconf.h as far as I remember
The new version in the latest commit was shortest and cleanest alternative I could make...

PR #15, opened 2015-05-04
Compile with -Wall -Werror

daxtens:
-Wall -Werror should help us to keep code quality up.

This patch set makes it compile and then enforces it in ./configure.


Dead2:
Most of these are minor fixes I was looking into myself, thanks for saving me the trouble.

Regarding the inline, I agree it can be removed. GCC does not want to inline it even with alwaysinline, so I have been considering either removing it or changing the code to not need it.

PR #17, opened 2015-05-05
Make 'make' make more magnificently.

daxtens:
The make fixes for building out of tree were a bit quirky.
- Docs were built as part of distclean.
- Out of tree builds touched zconf.h in SRCDIR. Fix that. It's still a little quirky, but at least it means that building out of tree doesn't dirty the source tree.


daxtens:
So interestingly this is breaking my Travis build, but it works locally both in and out of tree. _sigh_

PR #19, opened 2015-05-07
Fix signed/unsigned mismatch in match.c

mtl1979:


Dead2:
I'd like to keep this one on hold for a little while, since I am doing some research in the same code areas and this does not fix any real bugs.

mtl1979:
I don't mind... sign mismatches only avoid overflow in rare cases when a lot of data is accessed. I just separated this from another patch I was working on.

mtl1979:
**NOTE:** Commit a73defc28c29abcfefdebcf995100f20ba528f01 touches same file...

PR #20, opened 2015-05-07
Makefile cleanup continues...

mtl1979:
- Remove remaining unnecessary dependencies from Makefile.in's distclean target
- Make sure zconf.h exists and is up-to-date in win32/Makefile.msc


Dead2:
This looks fine to me really, but since you and Daxtens are both doing work in the same areas, I would love it if you both either comment on potential problems or leave a +1 comment if it seems fine. That way I know it has at least been seen by someone else than me and the committer.

Feel free to comment on any unrelated pull requests too, even if it is just to let me know that you do not see anything obviously wrong with it.

If I don't see any replies within a fair amount of time, I will go with my gut feeling and hope for the best.

I have seen so many pull requests in other projects just wasting away for months/years with little to no comments, so I'd like to avoid such a situation and rather keep an open dialogue not only with the contributors but between them. This is a community project after all, so lets try to get into a good habit that future contributors can follow. :)

I realize this is getting very much off topic, but; Great work guys! You are solving real problems with the build system and cleaning it up good. I am currently swamped with work, so I cannot dedicate much time to zlib-ng at the moment but I try to stay on top of contributions. I am doing research, profiling and other tests on some bigger changes to the code when I can, unfortunately several good looking possibilities have came up as net-losses.

mtl1979:
It's a lot about overlapping work... That's partly because I have also other projects I'm working on and Daxtens has his own cleanup with the test system that depends on what I'm working on... I try to comment on Daxtens commits, but sometimes it might take day or three. There is still open task for who is going to keep the Borland and Cygwin makefiles (win32/Makefile.bor and win32/Makefile.gcc) in sync as Borland C compiler is really old and Cygwin is a alternative platform which possibly we should drop supporting too...

I don't want to sound rude or ungrateful when it comes to people who do great work on this project, but sometimes it takes more than one, two or three developers to get things done cleanly without breaking support for any platform that we really should keep supporting.

Dead2:
Dropping Borland is fine with me, but I'd rather like to keep cygwin support.

mtl1979:
As long as it's not me who has to keep the Cygwin makefile in sync with the others... I really don't think blindly changing it without possibility to actually check that it works is what we want to do...

daxtens:
I'll try and keep an eye on things and +1 them as necessary.

I think cygwin should work exactly like standard _nix... but I haven't used it in a *loooong_ time. I might see if I can get a company Windows image and test on that.

daxtens:
By the way, if you use an at sign before my username, I'll get a notification --- like @daxtens, @Dead2 or @mtl1979.

Dead2:
@daxtens Ah, so that's how its done, thanks. I was actually looking for that :)

mtl1979:
Only ***\* thing with Cygwin is that one always needs to run dos2unix and unix2dos on any accessed file, because git assumes all Windows installations can handle gracefully both LF and CR+LF, but Cygwin chokes on CR+LF. MinGW's Git also assumes all text files are converted to CR+LF before committing.

PR #21, opened 2015-05-08
Unbreak things: import the right zconf.h

daxtens:
Previously I imported <zconf.h>. This worked on my box because I had a
system-wide zconf.h. It didn't work on Travis. This should fix that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,4 +2,8 @@ language: c
 compiler:
   - gcc
   - clang
-script: ./configure && make && make check
+env:
+  - BUILDDIR=. SRCDIR=.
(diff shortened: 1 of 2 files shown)


daxtens:
I've also added more tests to Travis because an early version of this broke out-of-tree builds and I'd like to make sure I don't break that in the future.

Dead2:
@mtl1979 Would this conflict with your work? I'd like to pull this for the travis fixes.

mtl1979:
@Dead2 I'm pretty sure it doesn't touch same lines that I'm working with... Nothing seems wrong either.

PR #31, opened 2015-05-16
Hacknslash

Dead2:
Ongoing conversion to stdint types.

Please review and comment, will be pulled tomorrow if no objections.

Compiles fine and makes no difference to speed or binary output on x86_64 w/linux gcc.


Dead2:
Replaced

PR #52, opened 2015-10-29
Fix for MSVC compilation when X86_ defines are enabled

matbech:
This adds several patches to compile it with MSVC when the intel X86\* defines are enabled.
I have successfully compiled it with:
X86_PCLMULQDQ_CRC
X86_SSE2_FILL_WINDOW
X86_CPUID
X86_SSE4_2_CRC_HASH
UNALIGNED_OK
UNROLL_LESS


Dead2:
This looks good to me, but I want a win32 user to test it before I merge it.
@mtl1979 do you have time to do a little test?

Also @matbech I'd love to see some numbers for before and after performance for your system if possible. I know what these patches enable, but everyone loves numbers :grinning:

matbech:
Is there a standardized performance test app available?

mtl1979:
I can't seem to checkout the branch... git fetch can't find it...

matbech:
Performance results for:
test data:
Silesia corpus
http://sun.aei.polsl.pl/~sdeor/index.php?page=silesia

Codename        Ivy Bridge
Specification       Intel(R) Core(TM) i7-3770 

file: silesia/mozilla

# 32-bit

## zlib-ng

Compressing data: raw data 51220480, compressed 19089128, factor 0.37, compression level (default = -1) -1, buffer size 32768, time 2625ms

## zlib-ng patch

Compressing data: raw data 51220480, compressed 19077704, factor 0.37, compression level (default = -1) -1, buffer size 32768, time 1594ms

## zlib-ori asm

Compressing data: raw data 51220480, compressed 19089124, factor 0.37, compression level (default = -1) -1, buffer size 32768, time 2047ms

# 64-bit

## zlib-ng

Compressing data: raw data 51220480, compressed 19089128, factor 0.37, compression level (default = -1) -1, buffer size 32768, time 1969ms

## zlib-ng patch

Compressing data: raw data 51220480, compressed 19077704, factor 0.37, compression level (default = -1) -1, buffer size 32768, time 1406ms

## zlib-ori asm

Compressing data: raw data 51220480, compressed 19089124, factor 0.37, compression level (default = -1) -1, buffer size 32768, time 1984ms

file: silesia\webster

# 32-bit

## zlib-ng

Compressing data: raw data 41458703, compressed 12213947, factor 0.29, compression level (default = -1) -1, buffer size 32768, time 1937ms

## zlib-ng patch

Compressing data: raw data 41458703, compressed 12208641, factor 0.29, compression level (default = -1) -1, buffer size 32768, time 1500ms

## zlib-ori asm

Compressing data: raw data 41458703, compressed 12213947, factor 0.29, compression level (default = -1) -1, buffer size 32768, time 1578ms

# 64-bit

## zlib-ng

Compressing data: raw data 41458703, compressed 12213947, factor 0.29, compression level (default = -1) -1, buffer size 32768, time 1454ms

## zlib-ng patch

Compressing data: raw data 41458703, compressed 12208641, factor 0.29, compression level (default = -1) -1, buffer size 32768, time 1313ms

## zlib-ori asm

Compressing data: raw data 41458703, compressed 12213947, factor 0.29, compression level (default = -1) -1, buffer size 32768, time 1485ms

matbech:
For anybody that wants to run their own tests:
https://github.com/matbech/zlib-perf
includes the binaries

arch/x86/x86.c
```diff
@@ -14,6 +14,22 @@ ZLIB_INTERNAL int x86_cpu_has_sse2;
 ZLIB_INTERNAL int x86_cpu_has_sse42;
 ZLIB_INTERNAL int x86_cpu_has_pclmulqdq;
 
+#ifdef _MSC_VER
+#include <intrin.h>
+#define CPU_PROCINFO_AND_FEATUREBITS 1
+
+void ZLIB_INTERNAL x86_check_features(void) {
+	unsigned int registers[4];
+	__cpuid(registers, CPU_PROCINFO_AND_FEATUREBITS);
```
mtl1979: Using define here is unnecessary as it is only used once.

arch/x86/x86.c
```diff
@@ -14,6 +14,22 @@ ZLIB_INTERNAL int x86_cpu_has_sse2;
...
+#ifdef _MSC_VER
+#include <intrin.h>
+#define CPU_PROCINFO_AND_FEATUREBITS 1
+
+void ZLIB_INTERNAL x86_check_features(void) {
+	unsigned int registers[4];
+	__cpuid(registers, CPU_PROCINFO_AND_FEATUREBITS);
+
+	unsigned int ecx = registers[2];
+	unsigned int edx = registers[3];
```
mtl1979: These two lines are unnecessary as value is not modified below.

arch/x86/x86.c
```diff
@@ -14,6 +14,22 @@ ZLIB_INTERNAL int x86_cpu_has_sse2;
...
+
+	x86_cpu_has_sse2 = edx & 0x4000000;
```
mtl1979:
```
x86_cpu_has_sse2 = registers[3] & 0x4000000;
```

arch/x86/x86.c
```diff
@@ -14,6 +14,22 @@ ZLIB_INTERNAL int x86_cpu_has_sse2;
...
+	x86_cpu_has_sse42 = ecx & 0x100000;
```
mtl1979:
```
x86_cpu_has_sse42 = registers[2] & 0x100000;
```

arch/x86/x86.c
```diff
@@ -14,6 +14,22 @@ ZLIB_INTERNAL int x86_cpu_has_sse2;
...
+	x86_cpu_has_pclmulqdq = ecx & 0x2;
```
mtl1979:
```
x86_cpu_has_pclmulqdq = registers[2] & 0x2;
```

mtl1979:
## Simple large file speed test with level 9

Original size: 17694914560

(matbech)
Compress time: 37m40.977s
Compressed size: 8400015342
Uncompress time: 7m36.393s

(dead2)
Compress time: 46m15.234s
Compressed size: 8400117371
Uncompress time: 8m25.673s

arch/x86/x86.c
matbech: This was done by intention to make it clear what registers are used.

arch/x86/x86.c
matbech: changed: https://github.com/matbech/zlib-ng/commit/d716e3b0b3c22917164cb113e33a30c0a341f15c

arch/x86/x86.c
Dead2: Knowing that the registers are ecx/edx etc is actually very useful for readability, so I'd prefer it stays like this or gets documented clearly. I would think the compiler will be able to optimize out the extra variable assignments here in any case.

mtl1979: The problem here is that either the code is cleaned up or there is long span of duplicated code that makes it really hard to follow what exactly is compiler-dependent code and what is not... I'm all for documenting things but in the end this is nothing more than work-in-progress as it is missing the part where all the optimizations actually get enabled at build time.

Dead2: For future expansion of the cpuid capability when we need to check for more settings, it would be nice if this is as obvious as possible.

But I am thinking that it might be better to only duplicate the cpuid instruction and population of the variables, while the actual checking of features should not be duplicated.

In other words, the following part should be common code, while the code to populate eax,ebx,ecx and edx variables should be split out to a new function (x86_get_cpuid for example) selected by defines. This way both parts can be easily maintained, readability is increased and addition of checks for more cpuid features can be implemented with minimal chances of breaking the compiler-dependant parts of the code.

```
     x86_cpu_has_sse2 = edx & 0x4000000;
     x86_cpu_has_sse42 = ecx & 0x100000;
     x86_cpu_has_pclmulqdq = ecx & 0x2;
```

mtl1979: Completely agree...

Dead2:
This looks good to me. Could you rebase it and preferably clean up the commits?

Also now that you cleaned up the old asm code in cpuid, I agree that the preprocessor test would be nicer inside the function instead of enclosing the whole function.

matbech:
I'm retracting this patch because I do not see it as a good fit for this repository.

PR #55, opened 2015-11-09
Fix uninitialized variable

jibsen:
I was trying to get the squash unit tests to run on MSVC, and got a runtime error about this uninitialized variable in the debug build.


Dead2:
Thanks :)

PR #65, opened 2016-04-24
Consistently use Z_NULL to initialize the opaque object.

nmlgc:
There's also an actual benefit to this. When using the unmodified zconf.h.in, compress.c and uncompr.c end up without a definition for NULL in both GCC and Clang, when compiling for the x86_64-unknown-linux-gnu target. Using Z_NULL here allows zlib-ng (without the gzip parts) to be compiled under this setup by just symlinking zconf.h.in to zconf.h, without having to run ./configure or CMake before, which is especially useful for projects that statically link zlib-ng or use Git submodules.

### DIFF
--- a/compress.c
+++ b/compress.c
@@ -31,7 +31,7 @@ int ZEXPORT compress2(unsigned char *dest, size_t *destLen, const unsigned char 
 
     stream.zalloc = (alloc_func)0;
     stream.zfree = (free_func)0;
-    stream.opaque = NULL;
+    stream.opaque = Z_NULL;
 
     err = deflateInit(&stream, level);
     if (err != Z_OK)
--- a/deflate.c
+++ b/deflate.c
@@ -187,7 +187,7 @@ int ZEXPORT deflateInit2_(z_stream *strm, int level, int method, int windowBits,
     strm->msg = Z_NULL;
     if (strm->zalloc == (alloc_func)0) {
         strm->zalloc = zcalloc;
-        strm->opaque = NULL;
+        strm->opaque = Z_NULL;
     }
     if (strm->zfree == (free_func)0)
         strm->zfree = zcfree;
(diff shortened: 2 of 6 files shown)


Dead2:
Looks good to me.
@mtl1979 Would this cause any problems with windows/mingw?

mtl1979:
@Dead2 Instead of re-introducing Z_NULL, we should fix that NULL is always properly defined as running configure or cmake is already optional on various targets that don't use bash. Only point of running configure or cmake is to detect on native compile that specific instruction sets are usable.

As far as I know, the root cause is that &lt;stddef.h&gt; is only included when building for Windows. This is the standard header defining NULL. Using Z_NULL is in my opinion a bad idea, because it masquerades error when using Z_NULL instead of literal 0 for non-pointer variables.

Dead2:
@nmlgc could you please have a look at this commit and give your thoughts?
[Replace Z_NULL with NULL. Fix incorrect uses of NULL/Z_NULL.](https://github.com/mtl1979/zlib-ng/commit/7f30ca41b119f7da61fd9158fd9fd7b95d29e621#diff-ce3e5ac484a4ca73e6533421950dbe15R121)

nmlgc:
I only went with Z_NULL because zlib.h [explicitly designates it as the constant to be used in this case](https://github.com/Dead2/zlib-ng/blob/63523d388d4ce0b38f0ee8c059240ed1a3be1f76/zlib.h#L205), but if you want to get rid of it and include stddef.h by default, that works too.

mtl1979:
We have been cleaning non-standard types from zlib-ng and Z_NULL is one of the last ones... It was most likely used to support non-standard C or C++ compilers that don't support "((void *)0)", which is standard definition for NULL. Like I mentioned above, one big reason to stick to standard types is so we get compiler warnings or errors when we assume compiler doesn't implicitly cast to compatible type when left and right side of assignment or operator have mismatched types.

Dead2:
I think we should keep a Z_NULL define in zlib.h though, for compatibility. I think it gets used externally.

mtl1979:
@Dead2 Yeah... But we should define it as NULL instead of 0. See f6c1e8ff9704b34a71e5874c97655e008ea7d8b1

Dead2:
A NULL cleanup patch has been merged to develop

Issue #73, opened 2016-07-11
Incorrect format string conversion specifiers


nemequ:
Running cppcheck on squash, several issues came up in zlib-ng.  One is the same issue as #51, the others are about incorrect conversion specifiers… AFAICT they're all pretty trivial to fix.

```
[zlib-ng/deflate.c:1035]: (warning) %d in format string (no. 1) requires 'int' but the argument type is 'unsigned int'.
[zlib-ng/inflate.c:1463]: (error) Shifting a negative value is undefined behaviour
[zlib-ng/trees.c:477]: (warning) %d in format string (no. 3) requires 'int' but the argument type is 'unsigned int'.
[zlib-ng/trees.c:752]: (warning) %ld in format string (no. 1) requires 'long' but the argument type is 'unsigned long'.
[zlib-ng/trees.c:752]: (warning) %ld in format string (no. 2) requires 'long' but the argument type is 'unsigned long'.
[zlib-ng/trees.c:775]: (warning) %ld in format string (no. 1) requires 'long' but the argument type is 'unsigned long'.
[zlib-ng/trees.c:778]: (warning) %ld in format string (no. 1) requires 'long' but the argument type is 'unsigned long'.
[zlib-ng/trees.c:781]: (warning) %ld in format string (no. 1) requires 'long' but the argument type is 'unsigned long'.
[zlib-ng/trees.c:838]: (warning) %ld in format string (no. 1) requires 'long' but the argument type is 'unsigned long'.
[zlib-ng/trees.c:838]: (warning) %ld in format string (no. 2) requires 'long' but the argument type is 'unsigned long'.
[zlib-ng/trees.c:841]: (warning) %ld in format string (no. 1) requires 'long' but the argument type is 'unsigned long'.
[zlib-ng/trees.c:841]: (warning) %ld in format string (no. 2) requires 'long' but the argument type is 'unsigned long'.
[zlib-ng/trees.c:983]: (warning) %d in format string (no. 1) requires 'int' but the argument type is 'unsigned int'.
[zlib-ng/trees.c:993]: (warning) %d in format string (no. 1) requires 'int' but the argument type is 'unsigned int'.
[zlib-ng/trees.c:282]: (warning) %u in format string (no. 1) requires 'unsigned int' but the argument type is 'signed int'.
[zlib-ng/trees.c:287]: (warning) %u in format string (no. 1) requires 'unsigned int' but the argument type is 'signed int'.
```

Dead2:
[...]

PR #74, opened 2016-07-17
Using correct format specifiers

Majlen:
Fixes bug #73. 

Changing code from unsigned to int in compress_block should not cause any issues, since the value never exceeds 285 in that function and is never less than 0.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -1032,7 +1032,7 @@ void check_match(deflate_state *s, IPos start, IPos match, int length) {
         z_error("invalid match");
     }
     if (z_verbose > 1) {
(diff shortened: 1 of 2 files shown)


Dead2:
Most of this looks good, and I would pull it if it did not also contain the change of datatype.
If you could split that into a separate pull request, that would be great. That way we can pull the format fixes but let the datatype change receive some additional testing and scrutiny.

One reason for this is that using unsigned can let the compiler do some additional optimizations since it can assume that the value will never be negative, so even though we know it will never actually be negative, the compiler might emit extra code to make sure.

Majlen:
Sure, I'll send the next pull request in a minute.

PR #92, opened 2017-02-27
call memset for read after write dependences at distance 1

sebpop:
On a benchmark using zlib to decompress a PNG image this change shows a 20%
speedup.  It makes sense to special case distance = 1 of read after write
dependences because it is possible to replace the loop kernel with a memset
which is usually implemented in assembly in the libc, and because of the
frequency at which distance = 1 appears during the PNG decompression:

Distance	Frequency
1		1009001
6		64500
9		29000
3		25500
144		14500
12		10000
15		3500
7		2000
24		1000
21		1000
18		1000
87		500
22		500
192		500

### DIFF
--- a/inffast.c
+++ b/inffast.c
@@ -244,16 +244,21 @@ void ZLIB_INTERNAL inflate_fast(z_stream *strm, unsigned long start) {
                     }
                 } else {
                     from = out - dist;          /* copy direct from output */
-                    do {                        /* minimum length is three */
-                        *out++ = *from++;
-                        *out++ = *from++;
-                        *out++ = *from++;
-                        len -= 3;
-                    } while (len > 2);
-                    if (len) {
-                        *out++ = *from++;
-                        if (len > 1)
+                    if (dist == 1) {
+                        memset (out, *from, len);
+                        out += len;
+                    } else {
+                        do {                        /* minimum length is three */
                             *out++ = *from++;
+                            *out++ = *from++;
+                            *out++ = *from++;
+                            len -= 3;
+                        } while (len > 2);
+                        if (len) {
+                            *out++ = *from++;
+                            if (len > 1)
+                                *out++ = *from++;
+                        }
                     }
                 }
             } else if ((op & 64) == 0) {          /* 2nd level distance code */


ghost:
FWIW, I've done a similar, overlapping thing [here](https://codereview.chromium.org/2722063002).  I didn't want to call out to `memset()` and `memcpy()` because they tend to have a lot of preamble determining how best to deal with very short copies and deciding if they can exploit the alignment of the data, while `deflate_fast()` doesn't need those checks because it's guaranteed plenty of wiggle room and rarely does a usefully aligned copy.

sebpop:
Hi Simon,
I like your patch.  Let me benchmark it and look at the perf profile.

Also, thank you for the pointers to the Chrome zlib bugs and patches.
I would vote for chromium to update its zlib to zlib-ng:
https://bugs.chromium.org/p/chromium/issues/detail?id=687631
for the reasons explained in
https://docs.google.com/document/d/1XMa2z5LmYn3TLjDVRz-C2VcHBoeWwzCmPQyPrnJsGW8/edit#

Sebastian

ghost:
It's not necessarily optimal because the ideal register width likely varies by CPU type, and there's essentially no pipelining as you would find in an aggressive `memcpy()` (and implementing `memset()` by unrolling then copying is clearly suboptimal) but since the copies are so short pipelining may not be a factor.  I was just trying to capture many cases at once with a restricted patch.

Try fiddling with the type in inffast.h to see what fits your CPU best.  Though I've only passed casual tests with 8-byte and 16-byte types.  Anything else _might_ go down in flames.

Dead2:
@simhos01 That patch looks very interesting indeed, and I hope you could make a pull request for zlib-ng as well. That way we can more easily discuss and review the changes.

I am very happy to finally see some patches that improve inflate performance, as we have been much too focused on deflate so far. :)

sebpop:
On Fri, Mar 3, 2017 at 4:05 AM, Hans Kristian Rosbach <
notifications@github.com> wrote:

> I am very happy to finally see some patches that improve inflate
> performance, as we have been much too focused on deflate so far. :)
>
I have also seen the recently submitted patch to vectorize adler32 for NEON:
https://codereview.chromium.org/2676493007/diff/100001/third_party/zlib/neon_adler32.c
This patch boosts the performance of inflate by about 7% on Exynos-M1.
Currently adler32 accounts for ~25% of the instructions executed by a PNG
decompress and inf_fast accounts for another ~25%.

Optimizing the cost of deflate is good for saving power on the web servers,
and that has direct ROI for the companies running the web.
On the other side, optimizing inflate mostly benefits the client side
(chromium, android, etc.) and the electricity costs for the devices running
zlib inflate are split among the billion users.

sebpop:
The performance with Simon's patch looks much better than with my memset patch: it brings the performance up by ~37% on my png image decompression.  On an AArch64 the total number of instructions executed on my benchmark goes from 3.5Bn to 2.5Bn; memory reads 831M -> 560M; writes 671M -> 398M.

Simon's patch touches more than the hotest loop: it addresses the problem in other cases as well.
When replacing Simon's chunkunroll/chunkcopy with memset only in the hotest loop, I see a 5% degradation in the performance, that could be explained by the added overhead of function call.

As recommended by Simon, I have also experimented with different types in inffast.h.
All types lower than 128 bits, as in Simon's patch, perform worse:
(the base 0% is with Simon's "typedef uint8x16_t inffast_chunk_t;")

On Juno r0:
```
-25% uint8_t
-11% uint16_t
 -8% uint32_t
 -4% uint64_t
  0% __uint128_t
```
On Exynos-M1:
```
-30% uint8_t
-15% uint16_t
 -7% uint32_t
 -3% uint64_t
  0% __uint128_t
```
Hans, let's revert the patch that I submitted, and let's take Simon's patch instead.

sebpop:
There is an extra 10% speedup on top of Simon's patch by adapting the memset idea to Simon's idea of helping the compiler with generating a good memcpy.
Here is the patch on top of Simon's changes:
```diff --git a/antutu/png/src/zlib/inffast.c b/antutu/png/src/zlib/inffast.c
index 0ed2a83..dd1a2de 100644
--- a/antutu/png/src/zlib/inffast.c
+++ b/antutu/png/src/zlib/inffast.c
@@ -48,6 +48,24 @@ static inline void storechunk(unsigned char* d, inffast_chunk_t c) {
     __builtin_memcpy(d, &c, sizeof(c));
 }
 
+static inline unsigned char *setchunk(unsigned char *d, unsigned char c, unsigned len) {
+  unsigned sz = sizeof(inffast_chunk_t);
+
+  while (len > sz) {
+    __builtin_memset(d, c, sz);
+    d += sz;
+    len -= sz;
+  }
+
+  // Remainder loop.
+  while (len > 0) {
+    *d++ = c;
+    len--;
+  }
+
+  return d;
+}
+
 /*
    Behave like memcpy, but assume that it's OK to overwrite at least
    INFFAST_CHUNKSIZE bytes of output even if the length is shorter than this,
@@ -391,6 +409,9 @@ unsigned start;         /* inflate()'s starting value for strm->avail_out */
                 }
                 else {
 #ifdef INFFAST_CHUNKSIZE
+                  if (dist == 1) {
+                    out = setchunk(out, *(out - dist), len);
+                  } else {
                     /* Whole reference is in range of current output.  No
                        range checks are necessary because we start with room
                        for at least 258 bytes of output, so unroll and roundoff
@@ -399,6 +420,7 @@ unsigned start;         /* inflate()'s starting value for strm->avail_out */
                      */
                     out = chunkunroll(out, &dist, &len);
                     out = chunkcopy(out, out - dist, len);
+                  }
 #else
                     from = out - dist;          /* copy direct from output */
                     do {                        /* minimum length is three */
```

Dead2:
The problem with @simhos01 patch is that it is currently BSD licensed, not zlib, and I do not want to introduce code with differing licenses in the library. Either the code needs to be re-licensed, or someone would need to write the code again based on the same ideas.
I am hoping @simhos01 will get permission to contribute directly to zlib-ng, as this is not the only patch currently being blocked by that problem from the same company.

Good work both of you though :)

sebpop:
Thanks Hans for your explanations.

I am also thinking that both the chunk memset and chunk memcpy
should be predicated by the total length of the data:
for a large enough length, we should just call memset or memcpy.
I will do some experiments to determine the magic length.

I will submit a patch based on the idea of helping the compiler
to generate nice memset and memcpy for small chunks, and a call to
the libc for larger chunks.

I am also working on implementing this same transform in the
Loop Idiom Recognition with runtime checks in LLVM.
This will help generating good code in any libraries with the
same patterns, like copy constructors in libcxx.

PR #93, opened 2017-03-16
inflate: improve performance of memory copy operations

sebpop:
When memory copy operations happen byte by byte, the processors are unable to
fuse the loads and stores together because of aliasing issues.  This patch
clusters some of the memory copy operations in chunks of 16 and 8 bytes.

For byte memset, the compiler knows how to prepare the chunk to be stored.
When the memset pattern is larger than a byte, this patch builds the pattern for
chunk memset using the same technique as in Simon Hosie's patch
https://codereview.chromium.org/2722063002

This patch improves by 50% the performance of zlib decompression of a 50K PNG on
AArch64 and x86_64 when compiled with gcc-7 or llvm-5.

### DIFF
--- a/inffast.c
+++ b/inffast.c
@@ -7,6 +7,7 @@
 #include "inftrees.h"
 #include "inflate.h"
 #include "inffast.h"
+#include "memcopy.h"
 
 /* Return the low n bits of the bit accumulator (n < 16) */
 #define BITS(n) \
@@ -195,10 +196,9 @@ void ZLIB_INTERNAL inflate_fast(z_stream *strm, unsigned long start) {
        input data or output space */
     do {
         if (bits < 15) {
-            hold += (uint32_t)(*in++) << bits;
-            bits += 8;
-            hold += (uint32_t)(*in++) << bits;
-            bits += 8;
+            hold += load_short(in, bits);
+            in += 2;
+            bits += 16;
         }
         here = lcode[hold & lmask];
       dolen:
@@ -222,10 +222,9 @@ void ZLIB_INTERNAL inflate_fast(z_stream *strm, unsigned long start) {
             }
             Tracevv((stderr, "inflate:         length %u\n", len));
             if (bits < 15) {
-                hold += (uint32_t)(*in++) << bits;
-                bits += 8;
-                hold += (uint32_t)(*in++) << bits;
-                bits += 8;
+                hold += load_short(in, bits);
+                in += 2;
+                bits += 16;
             }
             here = dcode[hold & dmask];
           dodist:
@@ -303,10 +302,18 @@ void ZLIB_INTERNAL inflate_fast(z_stream *strm, unsigned long start) {
                     if (op < len) {             /* still need some from output */
                         len -= op;
                         out = chunkcopysafe(out, from, op, safe);
-                        out = chunkunroll(out, &dist, &len);
-                        out = chunkcopysafe(out, out - dist, len, safe);
+                        if (dist == 1) {
+                            out = byte_memset(out, len);
+                        } else {
+                            out = chunkunroll(out, &dist, &len);
+                            out = chunkcopysafe(out, out - dist, len, safe);
+                        }
                     } else {
-                        out = chunkcopysafe(out, from, len, safe);
+                        if (from - out == 1) {
+                            out = byte_memset(out, len);
+                        } else {
+                            out = chunkcopysafe(out, from, len, safe);
+                        }
                     }
 #else
                     from = window;
@@ -347,46 +354,30 @@ void ZLIB_INTERNAL inflate_fast(z_stream *strm, unsigned long start) {
                             from = out - dist;  /* rest from output */
                         }
                     }
-                    while (len > 2) {
-                        *out++ = *from++;
-                        *out++ = *from++;
-                        *out++ = *from++;
-                        len -= 3;
-                    }
-                    if (len) {
-                        *out++ = *from++;
-                        if (len > 1)
-                            *out++ = *from++;
-                    }
+
+                    out = chunk_copy(out, from, (int) (out - from), len);
 #endif
                 } else {
 #ifdef INFFAST_CHUNKSIZE
-                    /* Whole reference is in range of current output.  No
-                       range checks are necessary because we start with room
-                       for at least 258 bytes of output, so unroll and roundoff
-                       operations can write beyond `out+len` so long as they
-                       stay within 258 bytes of `out`.
-                     */
-                    out = chunkunroll(out, &dist, &len);
-                    out = chunkcopy(out, out - dist, len);
-#else
-                    from = out - dist;          /* copy direct from output */
-                    if (dist == 1) {
-                        memset (out, *from, len);
-                        out += len;
+                    if (dist == 1 && len >= sizeof(uint64_t)) {
+                        out = byte_memset(out, len);
                     } else {
-                        do {                        /* minimum length is three */
-                            *out++ = *from++;
-                            *out++ = *from++;
-                            *out++ = *from++;
-                            len -= 3;
-                        } while (len > 2);
-                        if (len) {
-                            *out++ = *from++;
-                            if (len > 1)
-                                *out++ = *from++;
-                        }
+                        /* Whole reference is in range of current output.  No
+                           range checks are necessary because we start with room
+                           for at least 258 bytes of output, so unroll and roundoff
+                           operations can write beyond `out+len` so long as they
+                           stay within 258 bytes of `out`.
+                         */
+                        out = chunkunroll(out, &dist, &len);
+                        out = chunkcopy(out, out - dist, len);
                     }
+#else
+                    if (len < sizeof(uint64_t))
+                      out = set_bytes(out, out - dist, dist, len);
+                    else if (dist == 1)
+                      out = byte_memset(out, len);
+                    else
+                      out = chunk_memset(out, out - dist, dist, len);
 #endif
                 }
             } else if ((op & 64) == 0) {          /* 2nd level distance code */
--- a/inflate.c
+++ b/inflate.c
@@ -84,6 +84,7 @@
 #include "inftrees.h"
 #include "inflate.h"
 #include "inffast.h"
+#include "memcopy.h"
 
 #ifdef MAKEFIXED
 #  ifndef BUILDFIXED
@@ -1149,17 +1150,27 @@ int ZEXPORT inflate(z_stream *strm, int flush) {
                 }
                 if (copy > state->length)
                     copy = state->length;
+                if (copy > left)
+                    copy = left;
+                left -= copy;
+                state->length -= copy;
+                if (copy >= sizeof(uint64_t))
+                    put = chunk_memcpy(put, from, copy);
+                else
+                    put = copy_bytes(put, from, copy);
             } else {                             /* copy from output */
-                from = put - state->offset;
+                unsigned offset = state->offset;
+                from = put - offset;
                 copy = state->length;
+                if (copy > left)
+                    copy = left;
+                left -= copy;
+                state->length -= copy;
+                if (copy >= sizeof(uint64_t))
+                    put = chunk_memset(put, from, offset, copy);
+                else
+                    put = set_bytes(put, from, offset, copy);
             }
-            if (copy > left)
-                copy = left;
-            left -= copy;
-            state->length -= copy;
-            do {
-                *put++ = *from++;
-            } while (--copy);
             if (state->length == 0)
                 state->mode = LEN;
             break;
--- a/memcopy.h
+++ b/memcopy.h
@@ -1,0 +1,394 @@
+/* memcopy.h -- inline functions to copy small data chunks.
+ * For conditions of distribution and use, see copyright notice in zlib.h
+ */
+#ifndef MEMCOPY_H_
+#define MEMCOPY_H_
+
+#if (defined(__GNUC__) || defined(__clang__))
+#define MEMCPY __builtin_memcpy
+#define MEMSET __builtin_memset
+#else
+#define MEMCPY memcpy
+#define MEMSET memset
+#endif
+
+/* Load a short from IN and place the bytes at offset BITS in the result. */
+static inline uint32_t load_short(const unsigned char *in, unsigned bits) {
+    union {
+        uint16_t a;
+        uint8_t b[2];
+    } chunk;
+    MEMCPY(&chunk, in, sizeof(uint16_t));
+
+#if BYTE_ORDER == LITTLE_ENDIAN
+    uint32_t res = chunk.a;
+    return res << bits;
+#else
+    uint32_t c0 = chunk.b[0];
+    uint32_t c1 = chunk.b[1];
+    c0 <<= bits;
+    c1 <<= bits + 8;
+    return c0 + c1;
+#endif
+}
+
+static inline unsigned char *copy_1_bytes(unsigned char *out, unsigned char *from) {
+    *out++ = *from;
+    return out;
+}
+
+static inline unsigned char *copy_2_bytes(unsigned char *out, unsigned char *from) {
+    uint16_t chunk;
+    unsigned sz = sizeof(chunk);
+    MEMCPY(&chunk, from, sz);
+    MEMCPY(out, &chunk, sz);
+    return out + sz;
+}
+
+static inline unsigned char *copy_3_bytes(unsigned char *out, unsigned char *from) {
+    out = copy_1_bytes(out, from);
+    return copy_2_bytes(out, from + 1);
+}
+
+static inline unsigned char *copy_4_bytes(unsigned char *out, unsigned char *from) {
+    uint32_t chunk;
+    unsigned sz = sizeof(chunk);
+    MEMCPY(&chunk, from, sz);
+    MEMCPY(out, &chunk, sz);
+    return out + sz;
+}
+
+static inline unsigned char *copy_5_bytes(unsigned char *out, unsigned char *from) {
+    out = copy_1_bytes(out, from);
+    return copy_4_bytes(out, from + 1);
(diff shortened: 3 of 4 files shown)


Dead2:
This looks really good. I have not done any testing with it yet, but I read through the code-changes, and noticed nothing that seems objectionable.

Two things though;
- Can you add a copyright header to memcopy.h? I am assuming this is code that you wrote and that you can zlib license it.
- A small nitpick really, if you could change memcopy.h to use a line length of 120 chars, that would be great. Most people are not limited to 80 chars wide terminals any more, so we have changed zlib-ng to use 120 chars, since there are quite a few places where that makes the code easier to read.

sebpop:
Hans, thank you for your fast review.
Would the following text be enough to be added to the header file?

/* memcopy.h -- inline functions to copy small data chunks.
 */

I am preparing an updated patch with your recommendations.

Dead2:
Yeah, that header would be good :)

memcopy.h
```diff
@@ -0,0 +1,181 @@
...
+    unsigned sz = sizeof(chunk);
+    unsigned char *from = out - dist;
+
+    /* assert(len >= sz); */
+
+    /* Double up the size of the memset pattern until reaching the largest pattern of size less than SZ. */
+    while (dist < len && dist < sz) {
+        __builtin_memcpy(&chunk, from, sz);
+        __builtin_memcpy(out, &chunk, sz);
+
```
sebpop: This is the pattern that appears in Simon's patch, and I think I know a better way to improve its performance:
we should only read DIST bytes and only write SZ bytes.
The memset pattern should be computed in registers with | and <<.
This will improve performance as we will reduce the number of memory accesses.

Let's not commit this patch yet.

ghost: For NEON the pattern is a trivial (eg.,) `vreinterpretq_u8_u16(vld1q_dup_u16((void *)from));` and family, but that only covers powers of two.  For non-powers of two I could swear I posted the algorithm in a comment somewhere [here](https://github.com/Dead2/zlib-ng/pull/92) but I can't find it now.

Dead2:
It was brought to my attention that this patchset probably does not remain compatible with big-endian machines. We should probably fix that, either by just using the original code for those, or adapting the new code to work on big-endian as well where possible.
@sebpop If you can, we'd love to see you in #zlib-ng on the freenode irc network.

ghost:
If you have Ubuntu, you might try testing big-endian [this way](https://gist.github.com/simhos01/5eda2d649cabba8cf5d50eca78cd386b).

sebpop:
Thanks Hans for pointing out the issue with big-endian.
Thanks Simon for the instructions on how to test on qemu big-endian code.
I have access to several machines that are big-endian in the gcc farm:
https://gcc.gnu.org/wiki/CompileFarm
Let me investigate the problem and I will post an amended patch.

inffast.c
```diff
@@ -107,10 +108,11 @@ void ZLIB_INTERNAL inflate_fast(z_stream *strm, unsigned long start) {
        input data or output space */
     do {
         if (bits < 15) {
-            hold += (uint32_t)(*in++) << bits;
-            bits += 8;
-            hold += (uint32_t)(*in++) << bits;
-            bits += 8;
+            unsigned short chunk;
+            __builtin_memcpy(&chunk, in, sizeof(unsigned short));
+            hold += (chunk << bits);
```
sebpop: This hunk with the other similar one below is the only change that is unsafe for big-endian machines.
make check passes on powerpc-linux with these two hunks removed.

sebpop:
The new patch passes on big-endian powerpc-linux.

inffast.c
ghost: It might be worth putting a generic `uint16_t` getter in a header file so it can be implemented differently depending on endianness and unaligned load support, but I couldn't demonstrate any real performance gain over handful of files with this technique anyway.  Though it was the first change I tried, and it's buried under a lot of other performance problems, probably including branch misprediction.

sebpop:
I think I addressed all the requested changes in the new patch.

I also did a comparison with Simon's "data privatization" by adding padding to the end of the window in order to remove some of the branches and checks that avoid runing over the allocated buffer.  This had a small effect on the performance, about 5% on PNG decompression.  Also those changes do not make possible the change to inflate.c (that works on a different buffer than the window in inffast.c) and the change to inflate.c accounts for about 20% speedup in a PNG decompression.  Overall, I think that adding the end of window padding complexifies the assumptions under which the memcpy operations are legal without much performance benefit.

sebpop:
The last version of the patch passes on x86_64-linux and aarch64-linux with the extended set of tests that Mika suggested in https://github.com/sebpop/zlib-ng/commit/c37fd4195ae8846f1c35c31950edc9168d6b9173#commitcomment-21405774

I have also measured the performance of the change when used in a png and gzip decompression:

The number of executed instructions reported by valgrind --tool=cachegrind
on the decompression of a 50K PNG file on aarch64-linux:
- before the patch:
I   refs:      3,783,757,451
D   refs:      1,574,572,882  (869,116,630 rd   + 705,456,252 wr)

- with the patch:
I   refs:      2,569,373,984
D   refs:        908,492,140  (524,196,258 rd   + 384,295,882 wr)

The compression of a 260MB directory containing the code of llvm into a tar.gz
of 35MB and decompressing that with minigzip -d
on i7-4790K x86_64-linux, it takes 0.533s before the patch and 0.493s with the patch,
on Juno-r0 aarch64-linux A57, it takes 2.796s before the patch and 2.467s with the patch,
on Juno-r0 aarch64-linux A53, it takes 4.055s before the patch and 3.604s with the patch.

RJVB:
On Tuesday March 21 2017 11:51:19 sebpop wrote:

> The compression of a 260MB directory containing the code of llvm into a tar.gz
> of 35MB 

Is that with or without the tar step? Creating the .tar is probably a not insignificant bottleneck.

sebpop:
On Tue, Mar 21, 2017 at 2:10 PM, René Bertin <notifications@github.com>
wrote:
>
> Is that with or without the tar step? Creating the .tar is probably a not
> insignificant bottleneck.
>
>
The numbers are only for the inflate part: "minigzip -d test.tar.gz", no
tar, no compression.

Dead2:
I had to make a difficult choice here, since your and simon's patches are so similar.
But it seems to me that simon's patch is a good middle-ground that your changes can mostly build upon or enhance. I hope you will be willing to re-base this on top of the current hacknslash6 branch.

memcopy.h
```diff
@@ -0,0 +1,108 @@
...
+            *out++ = *from++;
+        return out;
+    }
+
+    return chunk_memcpy(out, from, len);
+}
+
+/* Byte by byte semantics: copy LEN bytes from FROM and write them to OUT. Return OUT + LEN. */
+static inline unsigned char *chunk_copy(unsigned char *out, unsigned char *from, unsigned len) {
+    int dist = (int)(from - out);
```
sh1boot: I recommend keeping this logic outside of this function.  The reason is that otherwise you can end up calculating differences between pointers to different buffers, and that's not safe.

memcopy.h
```diff
@@ -0,0 +1,128 @@
...
```
ghost: [from the right account this time] I recommend keeping this logic outside of this function.  The reason is that otherwise you can end up calculating differences between pointers to different buffers, and that's not safe.

sebpop: Right, thanks for catching this: I will send an updated patch.

sebpop:
I have tested the performance and correctness of the new patch on x86_64-linux and ppc64-linux.  I will test aarch64-linux tomorrow when I will have access to the phones and Juno boards.

inflate.c
```diff
@@ -1157,9 +1158,13 @@ int ZEXPORT inflate(z_stream *strm, int flush) {
                 copy = left;
             left -= copy;
             state->length -= copy;
-            do {
-                *put++ = *from++;
-            } while (--copy);
+            if (copy >= sizeof(long)) {
+                put = chunk_copy(put, from, copy);
```
ghost: I think the easy way to fix the comment in memcopy.h is to duplicate this code, and put different versions in the if and else blocks around line 1153 above.  Since the first case can never come from the output buffer, it can never be a `memset()` operation, so it's simpler.

mtl1979:
Ok for me when all the comments have been addressed... Still leaves some code that can be optimized further, like using do-while loops for less than 4 iterations. Like I mentioned earlier, those loops should be unrolled. It makes the code a little longer, but branch prediction in most current cpus can handle unrolled short loops better.

ghost:
I had imagined that the chunk copies I'd put in `inffast.c` would be hoisted out to `memcopy.h` by this patch, and that the preprocessor would choose between implementations for different architectures as appropriate.  My patch was conservative about touching any code that I wasn't going to test (ie., implementations for alignment-constrained or other cores I can't benchmark on) which left a lot of preprocessor mess inside `inffast.c`.  But I think that if we're bold (are we bold?) then it would be reasonable to throw away the preprocessor stuff and always make calls to equivalent inline functions.

The original zlib was clearly written with no faith in the compiler's ability to inline and optimise, but that was a long time ago and I think we can afford to lean on the compiler more.

The original bytewise copy loops are probably still the appropriate fallback implementation for architectures where we don't know if un-aligned load/store are available.  However, if they're abstracted into inline functions then the non-overlapping version can take `restrict` pointers so the compiler might find the best optimisation for the target architecture.

That's what I'd imagined.  What do others think?

sebpop:
> using do-while loops for less than 4 iterations
Could you please be a bit more specific about what you mean by this and in which context you want to apply this transform?

If there are still small loops doing byte copies around it is because I was not able to get a better performance by rewriting those loops in another way.  Of course I have not tried all possible ways, so if you could point in the code which parts you would like me to rewrite in a different way, I can run some perf experiments to see whether that is worth changing the code.

sebpop:
Simon, I like the idea of moving the code from inffast.c to memcopy.h: that would make the code in inffast.c more readable, and would move the complexity of selecting the best way to generate the memory operations to memcopy.h.

After reading again the code in inffast.c, I am thinking that this should be done as a separate cleanup patch.

mtl1979:
long is 8 bytes on Unix derivates and 4 bytes on Windows when compiled on a 64-bit operating system... When UNALIGNED_OK is defined, we can assume 7 byte copy can be implemented as unrolled 4+2+1 byte copy and 3 byte copy can be implemented as 2+1 byte copy using uint32_t, uint16_t and uint8_t (or unsigned char), which are fixed size types defined in stdint.h... When UNALIGNED_OK is undefined, we can still save some decrements when we use same tactic as above, but use just `4*uint8_t+2*uint8_t+1*uint8_t or 2*uint8_t+1*uint8_t` and decrement len after each batch instead of after each byte. Instead of 7 conditional jumps we get 3 conditional jumps, and in the latter case we save 1 conditional jump.

ghost:
Keep in mind that when enough data following the intended output is writable and enough data following the input is readable, then the most efficient 7-byte copy is probably an 8-byte copy.  I'm not so sure about all smaller copies being slower than an 8-byte copy but if it takes a lot of tests to recognise them then they probably will be slower in the end.

Using [infgen](https://github.com/madler/infgen) and the `gzip < infgen.c` example it looks like < 8-byte copies are about 54% of matches, and < 16-byte copies are 85% of matches.  `gzip < infgen` (the x86_64 executable) has 71% of matches being less than 8-byte copies.  Worth paying careful attention to, I think.

mtl1979:
8-byte copy is always faster than multiple smaller copies on 64-bit processor, but we can't guarantee that code outside the function call can handle garbage at the end of the output... I recently fixed one buffer overflow issue in zlib-ng which had existed since day 0. Nevertheless 1-byte copy is the most inefficient version and anything else than that will be faster except switch-case block that tries to unroll each alternative of the loop separately ;) Even stock zlib used 3-byte copy in place of 1-byte copy in some places.

zlib-ng tries to avoid 8-byte copies when possible so it can be compiled for 32-bit processors.

ghost:
Well figuring that out is what we're here for.  Initially I wanted `chunk_copy` to _alway_ assume that it can widen at least the first iteration to a large copy because that's a core assumption in `inflate_fast()` (except when it has to wiggle around the edges of the rolling window, then assumptions break down).  I did a sub-optimal back-up implementation because the broken-down assumptions should have been rare; but now we're looking at more cases in `inflate.c` where the copy may also be shorter than the available output space (`copy < left` cases).

mtl1979:
I was actually thinking about reducing the branches in inlined intermediate code to as minimum as possible... if we make too many special cases and try to optimize them separately, we end up having equal amount or more branches than non-optimized version. My first test actually replaced all asserts with calls to less strict but still optimized version.

inflate.c
```diff
@@ -1149,17 +1150,33 @@ int ZEXPORT inflate(z_stream *strm, int flush) {
...
-                from = put - state->offset;
+                unsigned offset = state->offset;
+                from = put - offset;
                 copy = state->length;
+                if (copy > left)
+                    copy = left;
+                left -= copy;
+                state->length -= copy;
+                if (copy >= sizeof(long)) {
+                    put = chunk_memset(put, from, offset, copy);
```
ghost: This might be a problem.  `chunk_memset()` assumes that there's room to write at least two times the chunks size to the output to implement the unroll.  How much it needs is variable, but it should always be less than `2*sizeof(...)`.

That said, if it doesn't need to unroll (if the copies don't overlap) then we could test `left >= sizeof(...)` instead of `copy >= sizeof(...)`; which allows it to paint with a fatter brush.

ghost: Oh, except this happens right after `left` is modified.  I mean to test the _old_ value of `left`.

inflate.c
```diff
@@ -1149,17 +1150,33 @@ int ZEXPORT inflate(z_stream *strm, int flush) {
                 }
                 if (copy > state->length)
                     copy = state->length;
+                if (copy > left)
+                    copy = left;
+                left -= copy;
+                state->length -= copy;
+                if (copy >= sizeof(long)) {
```
ghost: I think this one would definitely benefit from `if (left >= sizeof(...))`... except that `left -= copy` needs to be moved to after the test (in fact since it can go after the copy, it can even go outside of the if/else block).

ghost:
One of the ways I saw to reduce the tests in `inflate_fast()` was to repeat all the copy loops multiple times down every if/else path (basically to eliminate the use of phi instructions in the compiler)... but the code would look terrible.

For example, in the window reference case, there's a block of extremely rarely executed code that does a memset if the reference is beyond the reach of the window.  This sets up a condition where subsequent copies from within the window _might_ be right at the limit where a chunky copy isn't safe; but they only need to do a safety check _if_ that special case got executed.  The fix is to duplicate all the subsequent code, _including_ implied safety checks, within the original special-case block and then to jump to the end of the loop.  This way the code that follows the special case _never_ has to think about awkward conditions that special case might have set up, so redundant code can be omitted.  This process then has to be repeated over and over for all the other tests that _must_ be performed to elide code that only need to be performed if other tests came out true.

I think this is what global branch prediction tries to fix, but that can only help with part of the problem.

But anyway... using the tests that one cannot avoid, like whether we're copying from the window or from the output buffer, we get some free insights like knowing that a copy from the window is never a memset-like operation.  I suspect there's a bit of a logic puzzle to finding the right minimum set of tests for maximum exploitable insight.

mtl1979:
Some developers like to use inline functions in private headers to hide all the ugly code. One good thing in well written inline functions is that compiler can guess when parts of the function can be discarded without making the code behave undefined. That's why builtin functions like memcpy() are inefficient unless compiler can deduce either the length parameter is constant or that the maximum possible value of length parameter is constant and defined above the call.

sebpop:
> uncompression time for large binary file (about 2 GB) dropped from 3.1 to 2.8 seconds

I think this is a compelling example.  Can we commit my patch as it is, and then push any other changes needed to get the performance up on gzip decompression?

sebpop:
Mika, could you please check the performance of your benchmark with the loops unrolled?

sebpop:
I think it makes sense to unroll the loops as I am seeing the following overall profile change:
```
$ diff -u before after-unroll-by8
[...]
            Ir          Dr          Dw          Bc 
-2,505,282,223 538,576,448 403,497,162 341,413,310  PROGRAM TOTALS
+2,363,644,332 512,538,462 378,115,696 307,745,872  PROGRAM TOTALS
```
There are fewer overall instructions, fewer loads, fewer stores, and fewer branches.

Dead2:
I am willing to pull these changes, but the conflicts in inffast.c must be fixed first.

The other conflicts are very trivial, just remove func_table before insert_string, or rebase on top of hacknslash7.

sebpop:
I can confirm the observation from Mika that hoisting the check for "dist > 7" outside the switch(len) improves performance.  I am updating the patch.

Dead2:
The functable commit is not supposed to go into hacknslash7 yet, please remove it.

memcopy.h
```diff
@@ -29,9 +29,98 @@ static inline uint32_t load_short(const unsigned char *in, unsigned bits) {
...
+    return out + sz;
+}
+
+/* Copy byte by byte LEN bytes from FROM into OUT. Return OUT + LEN. */
+static inline unsigned char *memcopy_byte_by_byte(unsigned char *out, unsigned char *from, unsigned len) {
+    switch (len) {
+    default:
+        Assert(len < 8, "memcopy_byte_by_byte should be called with less than 8 bytes");
+    case 7:
+        *out++ = *from++;
```
mtl1979: This might cause misalignment on architectures where unaligned read/writes are not allowed. It depends on how wise the compiler is but only in best case scenario memcopy reverts to byte-by-byte loop. Currently there is one supported arch (out of 4) that has strict alignment rules.

sebpop:
Thanks Hans, Mika, and Simon for your reviews.

Dead2:
Thank you @sebpop for staying with it :)

PR #95, opened 2017-03-18
Inflate using wider loads and stores and a minimum of branches.

ghost:
In inflate_fast() the output pointer always has plenty of room to write.  This
means that so long as the target is capable, wide un-aligned loads and stores
can be used to transfer several bytes at once, and most conditionals to avoid
output overflow can be deferred.

When the reference distance is too short simply unroll the data a little to
increase the distance.

### DIFF
--- a/inflate.c
+++ b/inflate.c
@@ -373,11 +373,10 @@ static int updatewindow(z_stream *strm, const unsigned char *end, uint32_t copy)
     if (state->window == NULL) {
 #ifdef INFFAST_CHUNKSIZE
         unsigned wsize = 1U << state->wbits;
-        state->window = (unsigned char FAR *)
-                        ZALLOC(strm, wsize + INFFAST_CHUNKSIZE,
-                               sizeof(unsigned char));
-        if (state->window == Z_NULL) return 1;
-        zmemzero(state->window + wsize, INFFAST_CHUNKSIZE);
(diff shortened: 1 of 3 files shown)


inflate.c
```diff
@@ -371,9 +371,18 @@ static int updatewindow(z_stream *strm, const unsigned char *end, uint32_t copy)
 
     /* if it hasn't been done already, allocate space for the window */
     if (state->window == NULL) {
+#ifdef INFFAST_CHUNKSIZE
+        unsigned wsize = 1U << state->wbits;
+        state->window = (unsigned char FAR *)
+                        ZALLOC(strm, wsize + INFFAST_CHUNKSIZE,
+                               sizeof(unsigned char));
```
sebpop: This is an interesting change that guarantees chunk memset and chunk memcpy operations to not run over the end of the allocated memory.  This has the benefit of removing some performance costly branches from the hot paths of the memory operations.

ghost: I also tried using `mmap()` to fold the window buffer around on itself to remove even more of these tests, but to show any real benefit it's necessary to tweak the behaviour so that inflate is performed directly into the window and then copied out when it's done.  Obviously this only suits callers that pass very small output buffers, but that's what libpng (a very performance-hostile caller) does.  Some platforms require that the window's padded to 64kB to do this, though.

PR #99, opened 2017-03-30
Added a new longest_match method

JunHe77:
This method uses "farthest hash" to reduce check loops. The idea is based on algorithm described at:
http://www.gildor.org/en/projects/zlib
 Benchmarked it with lzbench against Silesia's corpus. The performance gain can be at ~150% max. For most cases in level6-9 it can get 40%~60% compression speed up.

### DIFF
--- a/configure
+++ b/configure
@@ -111,6 +111,7 @@ sse4flag="-msse4"
 pclmulflag="-mpclmul"
 without_optimizations=0
 without_new_strategies=0
+use_distant_search=0
 gcc=0
 warn=0
 debug=0
@@ -145,6 +146,7 @@ case "$1" in
       echo '  configure [--zlib-compat] [--prefix=PREFIX]  [--eprefix=EXPREFIX]' | tee -a configure.log
       echo '    [--static] [--32] [--64] [--libdir=LIBDIR] [--sharedlibdir=LIBDIR]' | tee -a configure.log
       echo '    [--includedir=INCLUDEDIR] [--archs="-arch i386 -arch x86_64"]' | tee -a configure.log
+      echo '    [--without-new-strategies] [--use_distant_search]' | tee -a configure.log
         exit 0 ;;
     -p*=* | --prefix=*) prefix=`echo $1 | sed 's/.*=//'`; shift ;;
     -e*=* | --eprefix=*) exec_prefix=`echo $1 | sed 's/.*=//'`; shift ;;
@@ -168,6 +170,7 @@ case "$1" in
     --localstatedir=*) echo "ignored option: --localstatedir" | tee -a configure.log; shift ;;
     -noopt | --without-optimizations) without_optimizations=1; shift;;
     -oldstrat | --without-new-strategies) without_new_strategies=1; shift;;
+    -distant | --use_distant_search) use_distant_search=1; shift;;
     -w* | --warn) warn=1; shift ;;
     -d* | --debug) debug=1; shift ;;
     *)
@@ -741,6 +744,12 @@ fi
 if test $without_new_strategies -eq 0; then
     CFLAGS="${CFLAGS} -DMEDIUM_STRATEGY"
     SFLAGS="${SFLAGS} -DMEDIUM_STRATEGY"
+fi
+
+# Enable distant_search for deflate_slow
+if test $use_distant_search -eq 1; then
+    CFLAGS="${CFLAGS} -DUSE_DISTANT_SEARCH"
+    SFLAGS="${SFLAGS} -DUSE_DISTANT_SEARCH"
 fi
 
 ARCHDIR='arch/generic'
--- a/match.c
+++ b/match.c
@@ -16,7 +16,11 @@
       non-gcc compilers since the __builtin_ctzl() function might not be optimized. */
 #  if defined(__GNUC__) && defined(HAVE_BUILTIN_CTZL) && ((__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__) \
         || defined(__LITTLE_ENDIAN__))
+#ifndef USE_DISTANT_SEARCH
 #    define std3_longest_match
+#else
+#    define std4_longest_match
+#endif
 #  elif(defined(_MSC_VER) && defined(_WIN32))
 #    define std3_longest_match
 #  else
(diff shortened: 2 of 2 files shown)


gildor2:
Hi guys.

Sorry for entering discussion.
The algorithm implemented in this PR provides performance benefit only when used for "lazy evaluation". I've described my algorithm on http://www.gildor.org/en/projects/zlib to be more "cleaver", I think. It boost performance for regular longest_match search as well. I'm switching to a different offset every time when match grows.
Why doing "own implementation" while there's already fully functional, optimized and tested code?
https://github.com/gildor2/fast_zlib

Thanks,
Konstantin (gildor)

JunHe77:
Hi, Konstantin,

Really glad to have your comments here! And thanks for your idea and algorithm!
Yes, this implementation is different with yours, but also provides boost for both regular search(level1-3) and "lazy mode"(level4-9).
The boost for level1-3 comes from get_match_len which compares multiple bytes at one time(4bytes on 32bit and 8bytes on 64bit).
The boost for level4-9 comes from combination of get_match_len and your idea.
The reason I create my "own implementation" is based on the test result with lzbench(https://github.com/inikep/lzbench) against Silesia's corpus(http://sun.aei.polsl.pl/~sdeor/index.php?page=silesia). It is observed that for most Silesia's cases:
1. Compression speed: For level1-4, code-g(https://github.com/gildor2/fast_zlib) provides poorer performance than original zlib. The drop is about 2~3.x% for different case. For level5, code-g provides almost the same performance as original zlib. For level6-9, code-g provides big boost than original zlib, especially in level8-9.
2. For compress ratio, code-g provides similiar ratio as original zlib for 1 case(nci), while for other 11 cases it  produces larger files than original zlib. The difference is about 0.5%~3%.
3. For decompress speed, the stream produced by code-g are generally decompressed slower than stream created by original zlib.

So I created another implementation, tuned and tested with these corpus. The implementation in this PR provides better compression speed, almost the same compress ratio and slightly better decompress speed compare to original zlib in all levels. If you're instrested with the test, I can uploads the result, maybe provide a lzbench fork, so you can run the test.

Again, thank you for sharing us your idea and comments!

gildor2:
Hi,

1. I didn't test compression for levels below 5. My code shouldn't provide any benefit for it because zlib works in very different way for "fast compression" - this affects not just lazy matches, but also hash table manipulations. My code provides benefit only when it allows to iterate over smaller number of hash chains. But if deflate iterates over 5 chains, compression speed could be slightly slower because of the algorithm's complexity.

2. Worse compression. It's just a question of probability. My code iterates over different hash chains, this is heuristic. And perhaps sometimes it goes through hash chain which will give worse compression ratio, if this chain will be stopped before it finishes (low chain_length value). I used nearly 1Gb of real data for testing, and in average compression ratio is nearly the same as with original zlib. I didn't test smaller data blocks, like - 10Mb of data etc, just because of difficulty of catching difference, and doing accurate profiling. Plus, using larger data for testing could guarantee statistically better performance, not optimizing code just for some particular test suite.

3. Decompression speed. It heavily depends on number of LZ chains. When data contains lots of huffman codes for separate literals, decompression will be slower. Plus, some fortune, again. I believe compression algorithm should be optimized for faster compression speed, with compression ratio to be not worse than original code. But you're bothering about tiny performance drop with decompression.

I've tested decompressors as well, and I can say that its performance varies a lot from run to run. However zlib-ng (Win32/Visual Studio build) gives the worst compression and decompression performance, compared to original zlib, asm-optimized zlib (by Gilles Vollant), and my code.

Your code boosts only lazy matches, and doesn't boost regular match search. I think if you'll use the whole function from my git, you'll get better performance boost. Perhaps it's worth using different longest_match function for levels 1-5.

gildor2:
You've confused me with your test results, so I decided to test my code with "silesia" data set by myself. I used the whole tar file, without trying to pack every file separately. Test results are attached (screenshot of console window):
test-results

There are 5 results for levels 1,3,5,7,9. NewAsm - my assembly code, NewC - my C code, DLL - there are two: zlib-ng (marked with zlib-ng_32.dll) and original asm optimization (zlibwapi32.dll). And the last line is original C code. Note: there's no zlib-ng for level=1 just because application crashed somewhere inside zlib-ng (on memory deallocation).

You may see, that for levels < 5 my code always produces exactly the same result as original zlib. So, decompression time **should match**, but it differs - I mentioned that in my previous post. Compression performance is comparable with original zlib. I mentioned that because you told me that my code generates data which are slower to unpack.

Zlib-ng is significantly faster with compression on level 3. Its decompression speed is always slower (sorry, but this is always true in any of my tests).

My code starts to bring benefit in compression from level 5 - comparing to "original zlib" - just as it should (again, mentioned in previous post, that at levels < 5 my code works just like original one).

My code is a little bit slower than zlib-ng at level 5, but it provides a little bit better compression, so it is hard to compare. On higher levels, compression benefit grows.

JunHe77:
Hi, Konstantin,
I've forked your fast_zlib repo (https://github.com/JunHe77/fast_zlib) and lzbench (https://github.com/JunHe77/lzbench) which I mentioned before.
Both your code in "Sources/match.h" and code in this PR are added to these two forks, so we can bench them with same base.
The binaries are built and tested on Ubuntu 16.04 (64bit version), GCC 5.4.0 running on Intel  E5-2650v3.
Test with Silesia:
fastzlib_compare
Test with enwiki dump:
fastzlib_compare_enwiki

Test with Silesia using lzbench: (lzbench:original zlib, lzbench-p: zlib with code in this PR, lzbench-g: zlib with code in fast_zlib)
lzbench_compare

These results should clearly describe the points I mentioned in previous post. 
As I don't have visualStudio build env, you may try my fast_zlib fork and see the difference between NewC and PR.

gildor2:
Hi (sorry, but I don't know your name)

I've tested your code with my test application. Adopted it for Visual C++ (inline -> __inline, __builtin_ctzl -> _BitScanForward). It still runs slower than my code. Much slower on higher compression levels. With my own test data, and with silesia. I even tested it with gcc (started Linux under VMWare) - still slower.

**Important.** I see some mess in your output. For some reason, "my" code produced results on levels 1-4 which differs from original zlib (and even worse that original zlib, in terms of compression ratio). It **must** be the same. And in my tests it is the same. I mean: on my PC resulting file size for lower levels is the same for my code, your code, and original zlib, and it equals to file sizes (for "siletsia") for "PR" and "Orig" tests. If everything is correct on your size, there's probably a bug in my code and it **must be** fixed. Compressed data should not depend on the machine.

Several questions. What did you put into "zlib" source directory? Which gcc version do you use? What CPU do you have (Intel/AMD/etc)?

Thanks,
Konstantin

gildor2:
May be there are other people who interested in helping with this, who reads this thread? Could anyone else test the code?

JunHe77:
The build/test env is mentioned in previous comments:
"The binaries are built and tested on Ubuntu 16.04 (64bit version), GCC 5.4.0 running on Intel E5-2650v3."
As for zlib sources in fast_zlib, it is HEAD@https://github.com/madler/zlib

gildor2:
Sorry, I missed that specs.

gildor2:
I've found and fixed the problem, I think. The problem was with gcc/64: it treats "long" type as 64-bit, unlike Visual C++ which works with it like with 32-bit. I'm using 32-bit Linux on my machine, so the problem didn't appear to me. I've exactly reproduced your compression results when I replaced "ulg" in match.h with 64-bit integer type.

The fix is here: https://github.com/gildor2/fast_zlib/commit/3c0c4626a71a827a8e2d7dc98a365daff26078e9

JunHe77:
Cool!
I'll merge the fix and make some tests.

JunHe77:
Hi, Konstantin,
I updated https://github.com/JunHe77/fast_zlib with your fix and tested against silesia corpus on both 32bit Ubuntu and 64bit Ubuntu on Intel E5-2650v3. Here are the results:
32bit-ubuntu:
image

64bit-ubuntu:
image

JunHe77:
One more thing, for Orig in both tests, they are built with "UNALIGNED_OK" defined.
(Added:) The PR implementation in this test is a combination of 64bit-optimized _get_match_len_ and  Konstantin's distant search loop code.

gildor2:
Nice catch with UNALIGNED_OK. I was sure zlib defines it automatically, and was surprised that it is defined only for MSDOS :)

Regarding your tests. I didn't understand why your code still faster until I looked at your "fast_zlib" repository ([match_pr.h](https://github.com/JunHe77/fast_zlib/blob/master/Sources/match_pr.h)). I see that unlike before, _now_ you're using my longest_match code (1:1), with changed comments, reordered variable declarations and several renames of variables. Just compare code: [yours](https://github.com/JunHe77/fast_zlib/blob/master/Sources/match_pr.h#L34) vs [mine](https://github.com/gildor2/fast_zlib/blob/master/Sources/match.h#L23) Perhaps you forgot to mention that in new tests you used different algorithm. The difference with original code now is only with use of optimized version of match comparison (get_match_len), and different handling of "limit". 

What's yours there now? :) You're comparing _my code_ with _my code_, the performance is nearly identical. Optimization won't give significant result - sometimes your changes slightly faster (only in 64 bits), sometimes - slightly slower. No difference to be considered. However, if you'll do some optimizations to my algorithm, it is worth doing PR to my repository, but not posting this work here as if you (re-)created algorithm by yourself with just using algorithm's description from my site.

Sorry for criticism, I tried to be as much polite as possible.

JunHe77:
Hi, Konstantin,
Yes, the latest benchmark of "PR" is result of  64bit-optimized _get_match_len_ and your distant search implementation combination. I did this to verify if there is a faster implementation as you have fixed the "long" type  bug on 64bit arch.
I'm very sorry for not point it out in the comment, and any misunderstanding that I re-create the algorithm. It's all your credit.

As for the original PR implementation benchmark, pls check following:
=====> Benching level-1 ....
    PR:1   Data: 202.1 Mb   Time: 3.1   s   Size: 77259877 bytes   Speed: 65.49 Mb/s   Ratio: 2.74   Unpack: 217.48 Mb/s
  NewC:1   Data: 202.1 Mb   Time: 3.6   s   Size: 77259877 bytes   Speed: 55.59 Mb/s   Ratio: 2.74   Unpack: 203.47 Mb/s
  Orig:1   Data: 202.1 Mb   Time: 3.5   s   Size: 77259877 bytes   Speed: 57.72 Mb/s   Ratio: 2.74   Unpack: 210.31 Mb/s
=====> Benching level-2 ....
    PR:2   Data: 202.1 Mb   Time: 3.5   s   Size: 75000662 bytes   Speed: 57.95 Mb/s   Ratio: 2.83   Unpack: 234.05 Mb/s
  NewC:2   Data: 202.1 Mb   Time: 3.8   s   Size: 75000662 bytes   Speed: 53.40 Mb/s   Ratio: 2.83   Unpack: 226.42 Mb/s
  Orig:2   Data: 202.1 Mb   Time: 4.0   s   Size: 75000662 bytes   Speed: 50.76 Mb/s   Ratio: 2.83   Unpack: 204.41 Mb/s
=====> Benching level-3 ....
    PR:3   Data: 202.1 Mb   Time: 4.5   s   Size: 72968297 bytes   Speed: 44.71 Mb/s   Ratio: 2.90   Unpack: 229.26 Mb/s
  NewC:3   Data: 202.1 Mb   Time: 4.5   s   Size: 72968297 bytes   Speed: 44.95 Mb/s   Ratio: 2.90   Unpack: 236.27 Mb/s
  Orig:3   Data: 202.1 Mb   Time: 4.6   s   Size: 72968297 bytes   Speed: 43.47 Mb/s   Ratio: 2.90   Unpack: 231.89 Mb/s
=====> Benching level-4 ....
    PR:4   Data: 202.1 Mb   Time: 5.0   s   Size: 71002183 bytes   Speed: 40.61 Mb/s   Ratio: 2.98   Unpack: 221.20 Mb/s
  NewC:4   Data: 202.1 Mb   Time: 4.8   s   Size: 70963970 bytes   Speed: 42.35 Mb/s   Ratio: 2.99   Unpack: 224.35 Mb/s
  Orig:4   Data: 202.1 Mb   Time: 5.3   s   Size: 71002183 bytes   Speed: 38.15 Mb/s   Ratio: 2.98   Unpack: 227.89 Mb/s
=====> Benching level-5 ....
    PR:5   Data: 202.1 Mb   Time: 6.9   s   Size: 69006376 bytes   Speed: 29.41 Mb/s   Ratio: 3.07   Unpack: 209.41 Mb/s
  NewC:5   Data: 202.1 Mb   Time: 7.0   s   Size: 68858828 bytes   Speed: 28.92 Mb/s   Ratio: 3.08   Unpack: 209.52 Mb/s
  Orig:5   Data: 202.1 Mb   Time: 6.7   s   Size: 69162276 bytes   Speed: 30.36 Mb/s   Ratio: 3.06   Unpack: 221.58 Mb/s
=====> Benching level-6 ....
    PR:6   Data: 202.1 Mb   Time: 8.4   s   Size: 68164765 bytes   Speed: 24.13 Mb/s   Ratio: 3.11   Unpack: 235.38 Mb/s
  NewC:6   Data: 202.1 Mb   Time: 7.9   s   Size: 67961099 bytes   Speed: 25.53 Mb/s   Ratio: 3.12   Unpack: 231.45 Mb/s
  Orig:6   Data: 202.1 Mb   Time: 9.4   s   Size: 68227945 bytes   Speed: 21.47 Mb/s   Ratio: 3.11   Unpack: 234.33 Mb/s
=====> Benching level-7 ....
    PR:7   Data: 202.1 Mb   Time: 9.3   s   Size: 67887653 bytes   Speed: 21.78 Mb/s   Ratio: 3.12   Unpack: 231.58 Mb/s
  NewC:7   Data: 202.1 Mb   Time: 8.8   s   Size: 67786382 bytes   Speed: 22.90 Mb/s   Ratio: 3.13   Unpack: 235.81 Mb/s
  Orig:7   Data: 202.1 Mb   Time: 11.2  s   Size: 67939146 bytes   Speed: 17.99 Mb/s   Ratio: 3.12   Unpack: 226.65 Mb/s
=====> Benching level-8 ....
    PR:8   Data: 202.1 Mb   Time: 11.9  s   Size: 67681192 bytes   Speed: 16.95 Mb/s   Ratio: 3.13   Unpack: 239.35 Mb/s
  NewC:8   Data: 202.1 Mb   Time: 9.0   s   Size: 67639356 bytes   Speed: 22.36 Mb/s   Ratio: 3.13   Unpack: 242.43 Mb/s
  Orig:8   Data: 202.1 Mb   Time: 17.0  s   Size: 67694499 bytes   Speed: 11.92 Mb/s   Ratio: 3.13   Unpack: 242.98 Mb/s
=====> Benching level-9 ....
    PR:9   Data: 202.1 Mb   Time: 11.7  s   Size: 67642749 bytes   Speed: 17.29 Mb/s   Ratio: 3.13   Unpack: 235.37 Mb/s
  NewC:9   Data: 202.1 Mb   Time: 9.0   s   Size: 67634024 bytes   Speed: 22.42 Mb/s   Ratio: 3.13   Unpack: 241.67 Mb/s
  Orig:9   Data: 202.1 Mb   Time: 21.6  s   Size: 67643258 bytes   Speed:  9.36 Mb/s   Ratio: 3.13   Unpack: 245.75 Mb/s

So yours outperforms PR, especially in level8/9. Would you like to create a PR for your implementation?

gildor2:
Hi Jun,

I did an attempt to push my changes to the stock zlib - https://github.com/madler/zlib/issues/231
However Mark Adler is very slow in response (he didn't do anything on GitHub for months), so I'm still waiting. Regarding pull request for zlib-ng, I'm ok if someone else will integrate my code - I mentioned that in zlib's issue which I open, and explained why someone else should do that :)

Thanks,
Konstantin

PR #101, opened 2017-04-05
Merge PRs #93 and #95 (ARM and non-ARM code paths for inflate).

ghost:
Roughly what I have in mind.

### DIFF
--- a/memcopy.h
+++ b/memcopy.h
@@ -11,33 +11,53 @@
 #define MEMCPY memcpy
 #define MEMSET memset
 #endif
-
-/* Load a short from IN and place the bytes at offset BITS in the result. */
-static inline uint32_t load_short(const unsigned char *in, unsigned bits) {
-    union {
-        uint16_t a;
-        uint8_t b[2];
-    } chunk;
-    MEMCPY(&chunk, in, sizeof(uint16_t));
-
-#if BYTE_ORDER == LITTLE_ENDIAN
-    uint32_t res = chunk.a;
-    return res << bits;
+#if __STDC_VERSION__ >= 199901L
+#define RESTRICT restrict
 #else
-    uint32_t c0 = chunk.b[0];
-    uint32_t c1 = chunk.b[1];
-    c0 <<= bits;
-    c1 <<= bits + 8;
-    return c0 + c1;
+#define RESTRICT
 #endif
+
+#if (defined(__ARM_NEON__) || defined(__ARM_NEON))
+#  include <arm_neon.h>
+typedef uint8x16_t copychunk_t;
+#else
+#  include <stdint.h>
+typedef uint64_t copychunk_t;
+#endif
+
+#define COPYCHUNKSIZE sizeof(copychunk_t)
+#define COPYCHUNK_UNROLL 8
+
+/*
+   Ask the compiler to perform a wide, unaligned load with a machine
+   instruction appropriate for the copychunk_t type.
+ */
+static inline copychunk_t loadchunk(const unsigned char *s) {
+    copychunk_t c;
+    MEMCPY(&c, s, sizeof(c));
+    return c;
 }
 
-static inline unsigned char *copy_1_bytes(unsigned char *out, unsigned char *from) {
+/*
+   Ask the compiler to perform a wide, unaligned store with a machine
+   instruction appropriate for the copychunk_t type.
+ */
+static inline void storechunk(unsigned char *d, copychunk_t c) {
+    MEMCPY(d, &c, sizeof(c));
+}
+
+static inline void copychunk(unsigned char **d, const unsigned char **s) {
+    storechunk(*d, loadchunk(*s));
+    *d += COPYCHUNKSIZE;
+    *s += COPYCHUNKSIZE;
+}
+
+static inline unsigned char *copy_1_bytes(unsigned char * RESTRICT out, const unsigned char * RESTRICT from) {
     *out++ = *from;
     return out;
 }
 
-static inline unsigned char *copy_2_bytes(unsigned char *out, unsigned char *from) {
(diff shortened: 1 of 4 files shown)


memcopy.h
```diff
@@ -342,53 +291,108 @@ static inline unsigned char *byte_memset(unsigned char *out, unsigned len) {
...
+   output buffer is beyond the end of the current copy, and this can still be
+   exploited.
+ */
+static inline unsigned char *chunkcopy_safe(unsigned char *out, const unsigned char *from, unsigned len,
+                                           unsigned char *limit) {
+    Assert(out + len <= limit, "chunk copy exceeds safety limit");
+    if (limit - out < COPYCHUNKSIZE) {
+        /* Output buffer is too small to write a whole COPYCHUNKSIZE chunk, so
+           do things slowly and carefully. */
         return copy_bytes(out, from, len);
```
ghost: These don't join up.  copy_bytes() offers a safe implementation for the tail case, but it only goes up to 8, and copychunk_t could be (and is on ARM) larger than that.

memcopy.h
```diff
@@ -342,53 +291,108 @@ static inline unsigned char *byte_memset(unsigned char *out, unsigned len) {
...
+   iteration, which will hopefully make the branch prediction more reliable.
+ */
+static inline unsigned char *chunkcopy_relaxed(unsigned char *out, const unsigned char *from, unsigned len) {
+    --len;
+    storechunk(out, loadchunk(from));
+    out += (len % COPYCHUNKSIZE) + 1;
+    from += (len % COPYCHUNKSIZE) + 1;
+    len /= COPYCHUNKSIZE;
+#if defined(COPYCHUNK_UNROLL)
+    Assert(COPYCHUNK_UNROLL <= 8, "Unroll code needs more unrolling");
```
mtl1979: This should be "< 8".

memcopy.h
```diff
@@ -342,53 +291,108 @@ static inline unsigned char *byte_memset(unsigned char *out, unsigned len) {
...
+    len /= COPYCHUNK_UNROLL;
+    while (len > 0) {
+        if (COPYCHUNK_UNROLL >= 1) copychunk(&out, &from);
+        if (COPYCHUNK_UNROLL >= 2) copychunk(&out, &from);
+        if (COPYCHUNK_UNROLL >= 3) copychunk(&out, &from);
+        if (COPYCHUNK_UNROLL >= 4) copychunk(&out, &from);
+        if (COPYCHUNK_UNROLL >= 5) copychunk(&out, &from);
+        if (COPYCHUNK_UNROLL >= 6) copychunk(&out, &from);
+        if (COPYCHUNK_UNROLL >= 7) copychunk(&out, &from);
+        if (COPYCHUNK_UNROLL >= 8) copychunk(&out, &from);
```
mtl1979: This is bad... Too many if statements...

ghost: They're all const-const comparisons.  I started with the same thing in #if blocks, but it looked terrible.

memcopy.h
```diff
@@ -342,53 +291,108 @@ static inline unsigned char *byte_memset(unsigned char *out, unsigned len) {
...
+static inline unsigned char *chunkunroll_relaxed(unsigned char *out, unsigned *dist, unsigned *len) {
+    const unsigned char *from = out - *dist;
+    while (*dist < *len && *dist < COPYCHUNKSIZE) {
+        storechunk(out, loadchunk(from));
+        out += *dist;
+        *len -= *dist;
+        *dist += *dist;
+    }
+    return out;
 }
```
ghost: In a revised version of this patch (internal review pending) I've separated `chunkcopy_*()` and `chunkcopy_lapped_*()`, such that the former takes `restrict` pointers which must _never_ overtlap, and the latter takes a pointer and a distance which _implicitly_ overlap.  This forces all calls to be explicit about whether or not they need the unrolling or `memset()`-like behaviour.

In the `chunkunroll_lapped_safe()` case, this also means internalising the guard for unrolling too close to the end of the buffer, so there's no confusion (or less confusion) about how much working room it needs.

memcopy.h
mtl1979: I'm only saying the code will be most likely rejected as it can be written better without multiple if statements. Every if statement has speed penalty and if we're trying to make the code more efficient, that means we have to minimize if  and other branch statements that have unnecessary speed penalty.

memcopy.h
```diff
@@ -12,68 +12,83 @@
...
-    return c0 + c1;
+#  include <stdint.h>
+typedef uint64_t copychunk_t;
 #endif
+
+#define COPYCHUNKSIZE sizeof(copychunk_t)
+#define COPYCHUNK_UNROLL 8
+
+/*
+   Ask the compiler to perform a wide, unaligned load with an machine
```
mtl1979: Should be "with a machine", as M is not a vowel.

memcopy.h
```diff
@@ -12,68 +12,83 @@
...
+   instruction appropriate for the copychunk_t type.
+ */
+static inline copychunk_t loadchunk(const unsigned char *s) {
+    copychunk_t c;
+    MEMCPY(&c, s, sizeof(c));
+    return c;
+}
+
+/*
+   Ask the compiler to perform a wide, unaligned store with an machine
```
mtl1979: Same as above...

memcopy.h
ghost: Well the compiler will eliminate the tests and simply insert as many copychunk calls as match the condition.  I could use a loop, but I don't trust the compiler to unroll it.

mtl1979: I'm seeing same thing done two different ways in lines 280-293 and 295-304... Only difference is the "count" variable... Some people would call it unnecessary code duplication, some people would call it unnecessary variation.

ghost: If you want to combine those two you're talking about Duff's device, which most people seem to regard as too esoteric to consider.

PR #108, opened 2017-04-28
Lazily initialize functable members.

mtl1979:
* Split functableInit() function as separate functions for each functable member, so we don't need to initialize full functable in multiple places in the zlib-ng code, or to check for NULL on every invocation. 
* Optimized function for each functable member is detected on first invocation and the functable item is updated for subsequent invocations.
* Remove NULL check in adler32() and adler32_z() as it is no longer needed.


Dead2:
Thanks

PR #111, opened 2017-05-23
Various small fixes

RJVB:
Mostly to the CMake file but also to zlib.h to make zlib-ng a drop-in replacement for libz in ZLIB_COMPAT mode (cf. https://github.com/Dead2/zlib-ng/issues/94)

### DIFF
--- a/zlib.h
+++ b/zlib.h
@@ -109,6 +109,9 @@ typedef struct z_stream_s {
     int                   data_type;  /* best guess about the data type: binary or text
                                          for deflate, or the decoding state for inflate */
     uint32_t              adler;      /* Adler-32 or CRC-32 value of the uncompressed data */
+#if defined(ZLIB_COMPAT) && defined(X86_64)
+    uint32_t              padding;    /* pad out to the same size as the zlib struct */
+#endif
     unsigned long         reserved;   /* reserved for future use */
 } z_stream;
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,12 @@ check_include_file(stddef.h    HAVE_STDDEF_H)
 #
 # Options parsing
 #
-set(ARCH ${CMAKE_SYSTEM_PROCESSOR})
+if(CMAKE_OSX_ARCHITECTURES)
+    # if multiple architectures are requested (universal build), pick only the first
+    list(GET CMAKE_OSX_ARCHITECTURES 0 ARCH)
+else()
+    set(ARCH ${CMAKE_SYSTEM_PROCESSOR})
+endif()
 message(STATUS "Architecture: ${ARCH}")
 
 option (ZLIB_COMPAT "Compile with zlib compatible API" OFF)
@@ -98,8 +103,8 @@ elseif(MSVC)
         message(STATUS "Ignoring WITH_NATIVE_INSTRUCTIONS; not supported on this configuration")
     endif()
 else()
-    execute_process(COMMAND ${CMAKE_C_COMPILER} --version OUTPUT_VARIABLE COMPILER_VERSION)
-    if("${COMPILER_VERSION}" MATCHES "gcc" OR "${COMPILER_VERSION}" MATCHES "GCC" OR "${COMPILER_VERSION}" MATCHES "clang")
+    # catch all GNU C compilers as well as Clang and AppleClang
+    if(CMAKE_C_COMPILER_ID MATCHES "GNU" OR CMAKE_C_COMPILER_ID MATCHES "Clang")
         set(__GNUC__ ON)
     endif()
     if(WITH_NATIVE_INSTRUCTIONS)
@@ -313,19 +318,24 @@ else()
     else()
         set(CMAKE_REQUIRED_FLAGS "${PCLMULFLAG}")
     endif()
-    check_c_source_compile_or_run(
-        "#include <immintrin.h>
-        #include <wmmintrin.h>
-        int main(void)
-        {
-            __m128i a = _mm_setzero_si128();
-            __m128i b = _mm_setzero_si128();
-            __m128i c = _mm_clmulepi64_si128(a, b, 0x10);
-            (void)c;
-            return 0;
-        }"
-        HAVE_PCLMULQDQ_INTRIN
-    )
+    if(NOT (APPLE AND ${ARCH} MATCHES "i386"))
+        # the pclmul code currently crashes on Mac in 32bit mode. Avoid for now.
+        check_c_source_compile_or_run(
+            "#include <immintrin.h>
+            #include <wmmintrin.h>
+            int main(void)
+            {
+                __m128i a = _mm_setzero_si128();
+                __m128i b = _mm_setzero_si128();
+                __m128i c = _mm_clmulepi64_si128(a, b, 0x10);
+                (void)c;
+                return 0;
+            }"
+            HAVE_PCLMULQDQ_INTRIN
+        )
+    else()
+        set(HAVE_PCLMULQDQ_INTRIN NO)
+    endif()
 endif()
 
 #
@@ -411,7 +421,7 @@ if(WITH_OPTIM)
             add_definitions(-DX86_SSE4_2_CRC_HASH)
             set(ZLIB_ARCH_SRCS ${ZLIB_ARCH_SRCS} ${ARCHDIR}/insert_string_sse.c)
             add_feature_info(SSE4_CRC 1 "Support CRC hash generation using the SSE4.2 instruction set, using \"${SSE4FLAG}\"")
-            add_intrinsics_option(${SSE4FLAG})
+            add_intrinsics_option("${SSE4FLAG}")
             if(WITH_NEW_STRATEGIES)
                 add_definitions(-DX86_QUICK_STRATEGY)
                 set(ZLIB_ARCH_SRCS ${ZLIB_ARCH_SRCS} ${ARCHDIR}/deflate_quick.c)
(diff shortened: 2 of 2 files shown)


Dead2:
This fails compilation on Macs
https://travis-ci.org/Dead2/zlib-ng/jobs/235325802

It looks interesting though.

RJVB:
That's strange, in my case it builds. Looks like we'd to check first if `CMAKE_OSX_ARCHITECTURES` is defined and fall back to the current architecture if not. My bad, I had forgotten that my build scripts define the variable on the cmake commandline.

RJVB:
OK, so how do I get rid of that supposed conflict and turn it into the diff that it should be?

mtl1979:
@RJVB You need to rebase your changes on top of HEAD of Dead2:hacknslash7 after doing "git fetch --all". That way the conflict disappears as resolved.

RJVB:
On Thursday June 08 2017 10:07:05 Mika Lindqvist wrote:

> @RJVB You need to rebase your changes on top of HEAD of Dead2:hacknslash7 after doing "git fetch --all". That way the conflict disappears as resolved.

Grrr, had to redo this 3x ... once I messed up, then I thought I had forgotten to update the working copy of my own fork, and then in the end I decided to force-push because apparently the hacknslash7 branch is behind the develop branch I had used as the basis for my tweaks. >:(

mtl1979:
hacknslash7 is actually ahead of develop, it is testing branch.

zlib.h
```diff
@@ -109,6 +109,9 @@ typedef struct z_stream_s {
     int                   data_type;  /* best guess about the data type: binary or text
                                          for deflate, or the decoding state for inflate */
     uint32_t              adler;      /* Adler-32 or CRC-32 value of the uncompressed data */
+#if defined(ZLIB_COMPAT) && defined(X86_64)
+    uint32_t              padding;    /* pad out to the same size as the zlib struct */
+#endif
```
mtl1979, line 114: This is wrong, because member "adler" is not aligned correctly due to "data_type" being first 32-bit member after 64-bit "void" member. When "adler" was "unsigned long/uint64_t", it had extra 4-byte implicit padding added before it. In zlib-ng that implicit padding is unnecessary due to "int" and "uint32_t" both being 4 bytes.

RJVB:
The point here is not the alignment of the adler member but the total size of the structure in *ZLIB_COMPAT* mode. Without the change zlib-ng may be API compatible but it certainly isn't ABI compatible. Runtime errors are raised when you replace an existing stock shared libz with the one from zlib-ng.

So exactly what is wrong here and why is it important? Can you give an example where this change leads to errors? I can't recall having noticed a change of alignment of the adler variable, and wouldn't expect it to change (does it?).
AFAICT the z_stream_s structure is only used at runtime and is not represented "verbatim" in compressed files. There's nothing of interest after the added padding member. Even if alignment of the adler member does change and that leads to noticeable (not just measurable ;)) performance degradation that could be seen as a trade-off.

 I for one would consider it an important aspect of a ZLIB compatibility mode that users can replace shipped copies of the stock library and distributions can provide alternative packages. A "drop-in replacement" compatibility mode can be a powerful argument in mission-critical applications as it allows testing the library across a much larger set of applications than would be feasible if you had to rebuild all of them.
I'm willing to accept small performance as a trade-off for such compatibility (and I'm guessing ZLIB_COMPAT mode already comes with such trade-offs).

FWIW, I tested this change with GCC and Clang on Linux and Mac. I expect identical behaviour on MSWin PCs when using one of those compilers but cannot verify. Maybe one should not add the extra padding when building with MSVC?

mtl1979:
@RJVB The main problem is that if the "adler" member is updated inside zlib-ng, the pointer to it is different than when updated outside it in the program or library using zlib-ng.

RJVB:
Isn't that an extra reason to ensure it behaves as one would expect in ZLIB_COMPAT mode?

The alternative might be to use the exact same z_stream_s struct definition as stock zlib, in ZLIB_COMPAT mode. Would that be safer?

mtl1979:
@RJVB We used smaller variables because it avoided spilling registers when handling a lot of variables at the same time. The real solution requires knowing if the target machine is little- or big-endian and deciding what padding is required to align the variables correctly so that high 32-bits are ignored correctly inside zlib-ng. We should not rely on compiler implicitly padding the members to keep them aligned. We have to remember that zlib was written years before 64-bit machines or operating systems existed.

RJVB:
My impression with this particular struct is that padding is mostly of importance in that it must be the same as for the zlib version if you want to use zlib-ng with software that was built against the zlib headers. From what I've seen padding has little of any importance in the zlib-ng code itself, except possibly the effect it has on performance when it causes suboptimal alignment. And as I said, that's a trade-off.

As to ignoring the high 32bits in zlib-ng, I suppose one could use unions for that. Something like

```
#ifdef ZLIB_COMPAT && X86_64
	union {
		uLong zlib_adler;
		uint32_t ng_adler;
	} crc32;
	#define adler crc32.ng_adler
#endif

(or an even more hackish solution with a struct or a uint32_t[2] array inside the union)

It's been a while, but I think all the other members in the z_stream_s are of equivalent types, right?

mtl1979:
@RJVB The issue only happens if there is more than 2 uint32_t members and one of them was originally "unsigned long"... There is only one uint32_t member before "unsigned long" member earlier in the struct. Essentially you get different value when reading versus when writing because the pointer is 4 bytes off inside and outside zlib-ng. This is only important for fields like "adler" that are usable outside zlib-ng code and usually are compared with value from another function. The issue gets worse with big-endian machines because all uint32_t (that are unsigned long in zlib) variables need to be padded before the actual value to align the low 32-bits of both types. I haven't talked with @Dead2 yet to discuss which of the methods is the cleanest solution to fully address the issue.

RJVB:
Ah, if the adler variable can be read (written too?) outside of the zlib code there is indeed a problem with my patch.

Reading the above reply I really wonder why not simply use the same struct definition as zlib does in ZLIB_COMPAT mode. There isn't really any other option if you want to aim for ABI or "drop-in" compatibility.
A quick compare of the 2 structure definitions shows that indeed only "adler" is different (still not double-checked).

What kind of overhead are we talking about if you add a masking-out of the high (unwanted) 32 bits in every write to the "adler" variable where that's necessary?

R.

mtl1979:
@RJVB When the structs are correctly padded, you don't even need to mask out the high bits, because those should be all zeroes. If the structs are incorrectly padded, no masking out will fix it as it will write completely to wrong struct member.

RJVB:
On Wednesday July 05 2017 08:32:15 Mika Lindqvist wrote:
>@RJVB When the structs are correctly padded, you don't even need to mask out the high bits, because those should be all zeroes. If the structs are incorrectly padded, no masking out will fix it as it will write completely to wrong struct member.

Evidently... The ZLIB_COMPAT goal would be to get all public structures to have the same size and padding as the struct has in stock zlib. It would have helped immensely if zlib had used packed structure but that's just wishful thinking now.

"Incorrect padding" here means "not the same padding as when the code was compiled", right? Doesn't that mean that if you compile zlib-ng with a "stock" z_stream_s struct you get a library that will function correctly with binaries that were built against stock zlib? Provided you use a compatible compiler, of course?

I have the impression we're talking in circles here...

mtl1979:
@RJVB The problem with uint32_t/uInt vs. uint64_t/uLong (crc32 and adler32) and uint16_t vs. uint32_t (fill_window and window_size) was originally caused by changes needed to support assembler optimized calculations as some platforms (including PowerPC) are strongly typed, which means using incorrect size integer cause compiler error instead of implicit cast.

Dead2:
As far as I can tell, that struct is not to be used externally, BUT it can be temporarily stored externally in case the program needs to work on multiple streams or for some other reason want to pause/resume the compression.
The thing that causes compatibility problems is that some code (like libpng) checks the size of the struct in order to detect whether an obviously incompatible zlib is being used, that check is actually very broken, since it does not guarantee anything at all.
The struct can be identical but the zlib can still be incompatible, or the struct can have changed but the zlib would still have been fully compatible except for that size check.

So padding it to the right length seems like the appropriate solution to me.

mtl1979:
I still say we should add 4 byte padding (uint32_t) before and after "adler" member... That way it doesn't rely on compiler adding hidden 4 byte padding after the explicit "padding" to align the last member of the struct. This doesn't guarantee low 32 bits of "adler" member are correctly aligned on 64-bit big-endian platforms but currently we don't support any. However on 64-bit little-endian machines, the low 32 bits of "adler" member should be correctly aligned. When we add support for 64-bit big-endian platforms, we can reuse the first padding before "adler" member and swap positions of "adler" member and second padding to align "adler" member in regard of stock zlib's struct layout. It is essential in this case that we use two uint32_t paddings instead of one uint64_t padding as the latter forces hidden implicit padding of the explicit padding, which is just counterproductive.

RJVB:
And that would be compared to what version of the struct? My change adds a single uint32_t padding after adler. If I understood and remember the above exchange correctly, zlib-ng uses a 32bit adler member because it doesn't get any padding before it. If true, then adding 32bit of padding in front of (as well as after) it will increase the struct size, which is not what we want.

mtl1979:
@RJVB Adding 4 byte padding before "adler" doesn't add more padding than was in stock zlib, because stock zlib had 64-bit "adler" that had implicit 4-byte padding because there was only 1 uint32_t before uint64_t and uint64_t has twice the alignment compared to uint32_t. You can see this clearly in https://github.com/Dead2/zlib-ng/issues/94#issuecomment-303396792 where the error message says ```stream_size=112,expected=104``` as 112-104 bytes = 8 bytes = 64 bits = 1 x uint64_t = 2 x uint32_t.

Dead2:
Applied

PR #112, opened 2017-05-25
Fix: wrong register for BMI1 bit

matbech:
The BMI1 bit is in the ebx register and not in ecx. See reference: https://software.intel.com/sites/default/files/article/405250/how-to-detect-new-instruction-support-in-the-4th-generation-intel-core-processor-family.pdf


Dead2:
Thanks

PR #113, opened 2017-05-25
x86: use TZCNT instead of BSF

matbech:
Summary: TZCNT is faster on AMD than BSF. On processors that do not support TZCNT, the instruction byte encoding is executed as BSF.

### DIFF
--- a/arch/x86/ctzl.h
+++ b/arch/x86/ctzl.h
@@ -2,24 +2,32 @@
 #define X86_CTZL_H
 
 #include <intrin.h>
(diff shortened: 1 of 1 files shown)


Dead2:
I am not sure this works? Does this not replace a runtime check with a compiletime check?
I haven't tested it yet though.

matbech:
It works without the runtime check. It was documented in the patch: "- On processors that do not support TZCNT, the instruction byte encoding is executed as BSF. "

Dead2:
Thanks

PR #114, opened 2017-06-20
configure: For Windows builds, add the CROSS_PREFIX to $RC and $STRIP.

nmlgc:
zlib's original win32/Makefile.gcc did the same, but this was removed in 7d17132436431d5f62cf5089623073d72d07deb0. It is kind of essential for cross-compiling a Win32 build on Linux, since `windres` most certainly doesn't exist, and the regular `strip` may not be able to handle DLLs.

It should probably actually be something like

	RC="${RC-${CROSS_PREFIX}windres}"

and

	STRIP="${STRIP-${CROSS_PREFIX}strip}"

to be consistent with the assignments of $AR, $RANLIB and $NM, but this didn't work for some reason.


mtl1979:
@nmlgc I'm pretty sure the conditional version didn't work because the two variables are defined as empty strings at the top of the file instead of being undefined. I'm OK with the code that actually works.

PR #115, opened 2017-08-17
Merge Hacknslash7

Dead2:
Lets merge the ARM improvements and various other improvements and fixes accumulated during the last half year.

If anyone wants to test and sign off on this before it gets merged, that'd be great. Otherwise, I'll just merge it sometime in the next few days.

After this, no big code changes will be accepted into the 'develop' branch before we do a stable release. Some build-system changes will be needed though, those will be committed to a new branch before being merged into 'develop'.


mtl1979:
Obviously there is still some issues that need to be fixed before stable release, but I'm not sure if we should do the fixes incrementally to make sure we don't try to fix something that doesn't actually degrade performance, or cause zlib-ng to crash or to create corrupted output. I'm aware of that current code uses some shortcuts to improve performance, but lately people have reported that at least one of the shortcuts actually can cause corruption due to read of uninitialized memory.

Dead2:
The bug with uninitialized memory was reported 10 months ago, but it took us a long time to locate, understand and fix it. It could easily be reproduced using valgrind, but finding what caused it was the problem.

PR #134, opened 2017-11-08
Fix make distclean with non-standard make

orgads:
Like mingw32-make


mtl1979:
Looks OK to me...

Dead2:
Thanks

PR #137, opened 2017-11-12
[compat] Remove WITH_GZFILEOP from public headers, use unsigned long in compress-related functions.

### LINKED ISSUE #136: Problem linking against kmod and Qt-5.9.2
Good day

Im currently in process of cross-compiling system for Raspberry PI 3 with 
buildroot (https://buildroot.org/)  and when linking against zlib-ng (I build it with "--zlib-compat --neon --acle" ) I encountered few things and also have few questions:

1. kmod package complains about missin gzFile type. Commenting out WITH_GZFILEOP from zconf.h & zlib.h helped. I noticed that vanilla zlib.h and zconf.h does not have WITH_GZFILEOP at all, so why was it added ?

2. Qt-5.9.2 QByteArray uses compress/uncompress functions but they nag about pointer mismatch. It expected uLong for length arguments like in original zlib.h but I understand that size_t is the standard way to handle length argument. So who should patch this (Qt? vanilla zlib.h ? ) ?

3. The --zlib-compat configure switch is needed so that packages expecting Zlib API will build correctly right?

4. What is the lowest ARM arch/cpu that zlib-ng should work? Would it example work with Raspberry PI Zero W that has Arch/CPU of: ARMv6Z / ARM1176JZF-S ?
(After several tries, I have zlib-ng currently compiled for 32-bit mode of ARMv8-A / Cortex-A53 with following gcc flags: -march=armv8-a+crc --mfpu=crypto-neon-fp-armv8 --mtune=cortex-a53 )

Best Regards
-S-

mtl1979:
* Don't check for WITH_GZFILEOP in public headers as that is only defined when actually compiling zlib-ng
* Use unsigned long instead of size_t when passing size parameters to compress/compressBound/uncompress

Fixes #136.

### DIFF
--- a/compress.c
+++ b/compress.c
@@ -8,8 +8,10 @@
 #define ZLIB_INTERNAL
 #if defined(ZLIB_COMPAT)
 # include "zlib.h"
+# define z_size_t unsigned long
 #else
 # include "zlib-ng.h"
+# define z_size_t size_t
 #endif
 
 /* ===========================================================================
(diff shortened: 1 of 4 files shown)


Dead2:
Don't you think the z_size_t define should be moved to zutil.h or zconf.h?

mtl1979:
@Dead2 I thought of that but currently it is only needed in these two files, so moving it to header file is kinda overkill.

Dead2:
This needs to be rebased.

mtl1979:
Rebased.

PR #144, opened 2018-01-18
Runtime crc32_z optimized implementation detection

grooverdan:
The commit message contains the means of the changes in a rather large amount of detail.

In short, this provided a GLIBC  way of resolving to an optimized routine on first usage without changing the API. In this detection and resolution its possible to perform initialisation. Non-glibc, static libraries etc will fall back to a function pointer implementation.

The avoid the need for detection routines in the middle of functions (e.g. getauxval in #109)

As you may have noticed this is the rework of one of the patches in https://github.com/madler/zlib/pull/335.

Feedback welcome. I'll continue to re-pack the Power optimized crc32 for here too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -263,6 +263,57 @@ if(HAVE_BUILTIN_CTZL)
     add_definitions(-DHAVE_BUILTIN_CTZL)
 endif()
 
+#
+# check for attribute(ifunc) support in the compiler
+#
+check_c_source_compiles(
+    "
+    static unsigned int crc32_real_optimized(unsigned int crc, unsigned char *p,
+                              unsigned long len)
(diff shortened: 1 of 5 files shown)


mtl1979:
This solution pretty much duplicates the functable solution we already use and changes function signatures which we are trying to avoid. "functable" isn't gcc-specific and works on other compilers too.

grooverdan:
Apologies, I didn't see the functable as it wasn't part of the crc32 that I looked at but certainly could extend it there. I thought I only changed internal function signatures for consistency but I'll be careful. I did fall back to function pointers like functable however I'll try again, adding crc32 as a functable (with crc32_acle as a function) and see how that looks for a start.

PR #145, opened 2018-01-19
wrap crc32 in functable

grooverdan:
crc32_acle is defined with size_t as last argument in arch/*/crc32_acle.c so the
declaration was fixed.

Internal function crc32_generic changed to have size_t as its last argument to match the size_t from crc32_z where it is called from.


mtl1979:
This will break on non-GCC builds.

grooverdan:
ack. reverted crc32_generic.

functable.c
```diff
@@ -82,3 +85,10 @@ ZLIB_INTERNAL uint32_t adler32_stub(uint32_t adler, const unsigned char *buf, si
 
     return functable.adler32(adler, buf, len);
 }
+
+ZLIB_INTERNAL uint32_t crc32_stub(uint32_t crc, const unsigned char *buf, size_t len) {
+    // Initialize default
+    functable.crc32=crc32_z_ifunc();
```
mtl1979: Completely unnecessary... Should be inlined.

functable.c
```diff
@@ -32,13 +32,16 @@ extern uint32_t adler32_c(uint32_t adler, const unsigned char *buf, size_t len);
 extern uint32_t adler32_neon(uint32_t adler, const unsigned char *buf, size_t len);
 #endif
 
+extern uint32_t (*(crc32_z_ifunc(void)))(uint32_t, const unsigned char *, size_t);
```
mtl1979: Unnecessary.

crc32.c
```diff
@@ -213,8 +214,12 @@ const uint32_t * ZEXPORT get_crc_table(void) {
     return (const uint32_t *)crc_table;
 }
 
-uint32_t ZEXPORT crc32_z(uint32_t crc, const unsigned char *buf, size_t len) {
-    if (buf == NULL) return 0;
+
+ZLIB_INTERNAL
+uint32_t (*(crc32_z_ifunc(void)))(uint32_t, const unsigned char *, size_t)
+{
```
mtl1979: Unnecessary, should be done inside crc32_stub.

grooverdan:
I moved it inside crc32_stub in the latest commit. The extent of the changes was why I was avoiding that approach.

Due to the crc32_generic not using size_t there is a compile warning:

functable.c:123:24: warning: assignment from incompatible pointer type [-Wincompatible-pointer-types]
         functable.crc32=crc32_generic;
                        ^

If crc32_generic breaks on non-gcc compilers so will crc32_z, crc32_little and crc32_big.

mtl1979:
The problem was not crc32_generic but using gcc syntax without actually checking that the used compiler is gcc... gcc covers only one third of the current user base. 
"size_t" is non-standard extension and can be 32-bit even on modern system, and that can cause integer truncation warnings if not handled correctly. One of the reasons why we switched to "functable" approach is that it enforces that all variants use same function signature.

grooverdan:
As above, moved all internal functions to use z_off64_t so they all have the same function signature.

mtl1979:
I'm not completely sure we should use z_off64_t as replacement for size_t as it has same issues as size_t... In my opinion, the best solution would be to use uint64_t in internal versions of both adler32_* and crc32_*, but I'm scared it will grow the patch a lot bigger and prone to introducing more errors instead of reducing them. I think we have to wait what @Dead2 thinks as he ultimately decides which solution is the best compromise between correct operation and compatibility.

grooverdan:
ack. Looking forward to more specific guidance as to the way to go.

grooverdan:
so push forced the uint64_t as the internal length.

external apis are:
crc32(uint32_t crc, const unsigned char *buf, uint32_t len)
crc32_z(uint32_t crc, const unsigned char *buf, size_t len)

Both these end up calling functable.crc32(crc, buf, len) aka
uint32_t crc32_generic(uint32_t, const unsigned char *, uint64_t) (or optimized versions).

Based on previous assertions that uint32_t < uint64_t and size_t <= uint64_t we have met the compatibility and provide a stable internal API and an consistent external API.

As size_t implementation can be bigger that uint64_t in the C standard added a compile Assert.

I haven't touched adler32 yet as there's enough changes for now.

mtl1979:
@grooverdan It's cleaner if the changes for adler32 variants will be in separate pull request.

grooverdan:
thanks.

PR #146, opened 2018-01-19
Use CMake to generate cmakein file

donny-dont:
When building on a Windows host `sed` is not available. This patch uses CMake commands directly to create the zconf.h.cmakein file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -461,7 +461,7 @@ macro(generate_cmakein input output)
 
         if (_line STREQUAL "#define ZCONF_H")
             file(APPEND ${output} "#cmakedefine Z_HAVE_UNISTD_H\n")
-            file(APPEND ${output}"#cmakedefine Z_HAVE_STDARG_H\n")
+            file(APPEND ${output} "#cmakedefine Z_HAVE_STDARG_H\n")
         endif()
     endforeach()
 endmacro(generate_cmakein)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -454,11 +454,16 @@ message(STATUS "Architecture-specific source files: ${ZLIB_ARCH_SRCS}")
 #============================================================================
 
 macro(generate_cmakein input output)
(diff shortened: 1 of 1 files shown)


donny-dont:
@Dead2 @mtl1979 you guys should be able to setup https://www.appveyor.com/ if you want to do Windows builds after this patch lands.

mtl1979:
@donny-dont I actually run Windows 10, so I can do Windows build even locally.

donny-dont:
@mtl1979 I was referring to having a CI. I don't believe travis supports windows at all and I have used appveyor for doing windows builds. As this patch removes the requirement of having `sed` then a CI for windows builds would be possible.

CMakeLists.txt
```diff
@@ -454,11 +454,16 @@ message(STATUS "Architecture-specific source files: ${ZLIB_ARCH_SRCS}")
 #============================================================================
 
 macro(generate_cmakein input output)
-    execute_process(COMMAND sed "/#define ZCONF_H/ a\\\n#cmakedefine Z_HAVE_UNISTD_H\\\n#cmakedefine Z_HAVE_STDARG_H\n"
-                    INPUT_FILE ${input}
-                    OUTPUT_FILE ${output}
-)
-
+    file(REMOVE ${output})
+    file(STRINGS ${input} _lines)
```
donny-dont, line 458: `file(STRINGS)` is available in CMake 2.8.4

https://cmake.org/cmake/help/v2.8.4/cmake.html#command:file

CMakeLists.txt
```diff
@@ -454,11 +454,16 @@ message(STATUS "Architecture-specific source files: ${ZLIB_ARCH_SRCS}")
...
+    foreach(_line IN LISTS _lines)
```
donny-dont, line 459: `foreach(IN)` is available in CMake 2.8.4.

https://cmake.org/cmake/help/v2.8.4/cmake.html#command:foreach

donny-dont (commented):
Responding to @Dead2 's comments in IRC.

This setup would work in the minimum CMake required, 2.8.4. It produces identical output with the current `sed` command as it just appends the two `#cmakedefin` lines.

mtl1979:
@donny-dont : @Dead2 handles the updating of CI configurations, but Windows builds are my responsibility. Even if we add CI for Windows builds, it's still easier to diagnose possible issues locally.

PR #148, opened 2018-01-31
Rename library when compiled without --zlib-compat

Dead2:
Renaming the library to libz-ng when not compiled in zlib-compat mode lets zlib-ng co-exist with zlib in distros and in applications that might be linked with a library that is linked with zlib. Currently those will cause conflicts and result in unreliable operation.

This branch has been running in production servers for several months now, dynamically linked into nginx while the rest of the system uses the distro-supplied zlib. This required patching nginx to detect and use the correct library, function names and structs, but after those simple changes were made, no problems have been detected.

Renaming the library also lets us make some more changes to the operation of zlib-ng without having to think about retaining compatibility with cold-war-era systems.

I should note that the current function definitions are not necessarily the final ones for the renamed library. We might want to consider making some more changes to variable types before making a stable version that we will have to keep supporting.

This is now open to discussion, and we are willing to look at changes and ideas related to improving or extending this process. Do we for example remove the zng_adler32 / zng_adler32_z function duplication (and same for crc32)? I think we should, even though this makes is 0.1% more work to port an application to zlib-ng. Anything else?


mtl1979:
We already had discussion about adler32/crc32 elsewhere and I suggested changing the last parameter to uint64_t because size_t has issues under Windows. If the last parameter is uint64_t, we don't need separate version for size_t.

Dead2:
I agree with that, and I'd welcome a pull request for that ;)
But is there anything else? I haven't had time to have a good look at our exported functions with this in mind yet, but I hope to do so soon.

mtl1979:
@Dead2 I could do a pull request but like I mentioned it would overlap with at least one other pull request. There is too many things we need to clean up and getting all merged bit by bit would not make sense.

Dead2:
Of course, I didn't mean to imply we'd do that before merging the waiting PRs. But I would like to get a sense of where we are going after this :)

mtl1979:
@Dead2 Eh... We need a roadmap to know where we are going... We have so many things unfinished and others we want to postpone to later releases...

PR #149, opened 2018-02-06
Fix build problems about NEON

richael02:
NEON is enabled by default on armv8-a platforms, and so NEON related
objects should be included when the platform is armv8-a. Errors about
adler32_neon will occur when you run ./configure on armv8-a platforms
without --neon option, because zlib-ng uses --neon option to include
NEON related objects regardless of Arm architecture.
You will have similar issue when you build the project with cmake.

This patch fixes the problem by including NEON related objects when
the platform is armv8-a(including aarch64).


mtl1979:
It was intentional choice that NEON support can be disabled on all ARM platforms. If there is build errors when NEON support is disabled, that is something that will be fixed when people post appropriate build logs.

richael02:
@mtl1979  if running ./configure  without --neon on AArch64, the end part of the build logs are as follow:

make[1]: Leaving directory '/home/ubuntu/zym/compress/zlib-ng/arch/aarch64'
cp arch/aarch64/fill_window_arm.o fill_window_arm.o
ar rc libz.a adler32.o compress.o crc32.o deflate.o deflate_fast.o deflate_medium.o deflate_slow.o functable.o match.o infback.o inffast.o inflate.o inftrees.o trees.o uncompr.o zutil.o fill_window_arm.o
gcc -L. -o example example.o libz.a -lc
libz.a(functable.o): In function `adler32_stub':
functable.c:(.text+0xa8): undefined reference to `adler32_neon'
functable.c:(.text+0xb0): undefined reference to `adler32_neon'
functable.c:(.text+0xb8): undefined reference to `adler32_neon'
Makefile:163: recipe for target 'example' failed
make: *** [example] Error 1

mtl1979:
@richael02 So adler32_stub needs to be fixed... I will look into it...

richael02:
@mtl1979  My patch is meant to fix the above problem, so would you please have a review?

mtl1979:
@richael02 We need to be able to disable NEON support in all files for the test suites... So you need to separate adding new ARM variations and fixing the build errors, otherwise the whole patch will be rejected.

richael02:
@mtl1979  Thank you for your comment. My thought was the same to you originally. But other ARM architecture engineers suggested that I should make NEON default when the platform is armv8-a. So if NEON must be a choice that can be disabled, I will fix this soon.

mtl1979:
@richael02 This has been an issue in the past too... We are an independent project that can't follow guidelines from contributors, we have to follow our own guidelines and that means we sometimes have to reject patches even if they seem to be valid in eyes of an outsider. I'm only a maintainer of some of the architectures and platforms, @Dead2 is the lead developer and he can ultimately reject even my contributions if there is overlap in pull requests or other issues.

CMakeLists.txt
```diff
@@ -393,6 +393,7 @@ if(WITH_OPTIM)
             add_feature_info(ACLE_CRC 1 "Support CRC hash generation using the ACLE instruction set, using \"${ACLEFLAG}\"")
         endif()
         if(WITH_NEON)
+            set(NEONFLAGS "${NEONFLAGS} -DWITH_NEON")
```
mtl1979: All ARM-related external defines should start with "ARM_" ... Look at the isolated example I gave in the latest PR by me.

richael02: Thank you for advice. It's reasonable that the flag starts with ARM. But I don't think it's good to name it ARM_NEON_ADLER32. Since option "--neon" means using the NEON related optimization on ARM platforms, it should not be restricted to adler32_neon. Supposing we have many other optimizations that use NEON, and we have to add many flags like " add_definitions("-DARM_NEON_........") ". So what about  naming it "ARM_USE_NEON" , taking into consideration distinguishing it from the  gcc's parameter "__ARM_NEON__" ?

mtl1979: I just backported the define from our internal test version... There is currently only two defines to enable NEON-support in individual functions, in fill_window and adler32, so it's not an issue yet and it makes it really easy to add those to the test suite when it will be implemented.

richael02: OK ... If so, I change it to ARM_NEON_ADLER32

richael02:
@Dead2  @mtl1979 the merged patch https://github.com/Dead2/zlib-ng/pull/151/commits/d00a2d89bcb40ed80f85a27699caa8bacd563372  still have problems on ARM platforms, please review my new modification: armv7 and armv8 shouldn't be put in the same category!

configure
```diff
@@ -883,17 +883,28 @@ case "${ARCH}" in
...
 
                 ARCH_STATIC_OBJS="${ARCH_STATIC_OBJS} crc32_acle.o insert_string_acle.o"
                 ARCH_SHARED_OBJS="${ARCH_SHARED_OBJS} crc32_acle.lo insert_string_acle.lo"
 
                 if test $buildneon -eq 1; then
-                    CFLAGS="${CFLAGS} ${floatabi} -mfpu=neon -DARM_NEON_ADLER32"
-                    SFLAGS="${SFLAGS} ${floatabi} -mfpu=neon -DARM_NEON_ADLER32"
-
+                    CFLAGS="${CFLAGS} ${floatabi} -DARM_NEON_ADLER32"
+                    SFLAGS="${SFLAGS} ${floatabi} -DARM_NEON_ADLER32"
```
mtl1979, line 907: What's the reason for removing "-mfpu=neon"?

richael02: On armv7, it requires "-mfpu=neon" to enable NEON instructions . But for armv8 , there's no need to use "-mfpu= neon", NEON instructions is enabled by default. And if compiled with "-mfpu=neon", there will be compiling errors.

mtl1979: If it compiles correctly on @Dead2's Raspberry, then it's OK for me...

richael02: I test it  on AArch64 servers. Does Raspberry  use armv8 or armv7 ?

mtl1979: I think it has v8-A, but it only did boot 32-bit Linux.

richael02: What's the output of "uname -m" on Raspberry ? Is it AArch32?

PR #150, opened 2018-02-06
Fix the bug in crc32_acle

richael02:
On armv8-a platforms if --acle is enabled, zlib-ng will use crc32_acle
instead of the default crc32. However, in crc32_acle the __crc32b() is
used to calculate the crc result of two variables with types uint32_t
and uint64_t, which gives an error result.The correct function used
should be __crc32d().

### DIFF
--- a/arch/aarch64/crc32_acle.c
+++ b/arch/aarch64/crc32_acle.c
@@ -42,16 +42,16 @@ uint32_t crc32_acle(uint32_t crc, const unsigned char *buf, size_t len) {
 
 #ifndef UNROLL_LESS
     while (len >= 32) {
-        c = __crc32b(c, *buf8++);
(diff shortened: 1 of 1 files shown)


mtl1979:
Good catch... It's actually hard to find ARM hardware that can run 64-bit version of Linux so this wasn't tested yet. A lot of 64-bit ARM boards still run only 32-bit Linux so it's identified as AArch64 in operating system.

richael02:
@mtl1979  I found the problem when I use zlib-ng  as a binary replacement of zlib on our AArch64 platform. And I have tested it on AArch64.

mtl1979:
@richael02 Sometimes it takes extra pair of eyes to notice small errors ;) gcc doesn't always issue warnings when compiling especially if char or char* is involved.

PR #153, opened 2018-02-08
Add detection for 64bit long int to support padding

richael02:
Zlib-ng introduces a 32bit padding member in its data
struct to make zlib-ng compatible with zlib, without
which there will be runtime errors when making zlib-ng
a binary replacement of zlib. Now the padding is enabled
only when both ZLIB_COMPAT and X86_64 are defined, so it
needs support for other platforms where sizeof(long int)
is not equal to sizeof(int).

This patch fixes the problem by adding the detection for
64bit long int.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,6 +217,23 @@ check_include_file(unistd.h Z_HAVE_UNISTD_H)
 check_include_file(stdarg.h Z_HAVE_STDARG_H)
 
 #
+# Check for 64bit long int
+#
+check_c_source_runs(
+    "int main(void)
+    {
+        if(sizeof(long int) == 8)
+            return 0;
+        else
+            return 1;
+    }"
+    IS_64BIT_LONG
+)
+if(IS_64BIT_LONG)
+    add_definitions(-DIS_64BIT_LONG)
+endif()
+
+#
 # Check if we can hide zlib internal symbols that are linked between separate source files using hidden
 #
 check_c_source_compiles(
@@ -368,7 +385,7 @@ set(ZLIB_ARCH_SRCS)
 set(ARCHDIR "arch/generic")
 if("${ARCH}" MATCHES "x86_64" OR "${ARCH}" MATCHES "AMD64")
     set(ARCHDIR "arch/x86")
-    add_definitions(-DX86_64 -DX86_NOCHECK_SSE2 -DUNALIGNED_OK -DUNROLL_LESS)
+    add_definitions(-DX86_NOCHECK_SSE2 -DUNALIGNED_OK -DUNROLL_LESS)
     add_feature_info(SSE2 1 "Use the SSE2 instruction set, using \"${SSE2FLAG}\"")
 elseif("${ARCH}" MATCHES "arm")
     set(ARCHDIR "arch/arm")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -229,9 +229,6 @@ check_c_source_runs(
     }"
     IS_64BIT_LONG
 )
-if(IS_64BIT_LONG)
-    add_definitions(-DIS_64BIT_LONG)
-endif()
 
 #
 # Check if we can hide zlib internal symbols that are linked between separate source files using hidden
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -368,7 +385,7 @@ set(ZLIB_ARCH_SRCS)
 set(ARCHDIR "arch/generic")
 if("${ARCH}" MATCHES "x86_64" OR "${ARCH}" MATCHES "AMD64")
     set(ARCHDIR "arch/x86")
-    add_definitions(-DX86_64 -DX86_NOCHECK_SSE2 -DUNALIGNED_OK -DUNROLL_LESS)
+    add_definitions(-DX86_NOCHECK_SSE2 -DUNALIGNED_OK -DUNROLL_LESS)
```
mtl1979: This will break the assembler code.

configure
```diff
@@ -769,8 +790,8 @@ case "${ARCH}" in
 
         case "${ARCH}" in
             x86_64)
-                CFLAGS="${CFLAGS} -DX86_64 -DX86_NOCHECK_SSE2"
-                SFLAGS="${SFLAGS} -DX86_64 -DX86_NOCHECK_SSE2"
+                CFLAGS="${CFLAGS} -DX86_NOCHECK_SSE2"
+                SFLAGS="${SFLAGS} -DX86_NOCHECK_SSE2"
```
mtl1979: Bad idea... See above.

zlib.h
```diff
@@ -109,8 +109,9 @@ typedef struct z_stream_s {
     int                   data_type;  /* best guess about the data type: binary or text
                                          for deflate, or the decoding state for inflate */
     uint32_t              adler;      /* Adler-32 or CRC-32 value of the uncompressed data */
-#if defined(ZLIB_COMPAT) && defined(X86_64)
-    uint32_t              padding;    /* pad out to the same size as the zlib struct */
+#if defined(ZLIB_COMPAT) && defined(IS_64BIT_LONG)
+    uint32_t              padding;    /* pad out to the same size as the zlib struct: padding
+                                         is needed when the size of long int is 64 bits */
```
mtl1979: This code was incorrect originally and will be removed very soon. Assumes little-endian order which isn't true for all supported processors.

configure
richael02: the flag "X86_64" is only covered in these places:

CMakeLists.txt:383:    add_definitions(-DX86_64 -DX86_NOCHECK_SSE2 -DUNALIGNED_OK -DUNROLL_LESS)
configure:772:                CFLAGS="${CFLAGS} -DX86_64 -DX86_NOCHECK_SSE2"
configure:773:                SFLAGS="${SFLAGS} -DX86_64 -DX86_NOCHECK_SSE2"
zlib.h:112:#if defined(ZLIB_COMPAT) && defined(X86_64)

So from my view it's only defined for enabling padding

mtl1979 (changes requested):
Conflicts with changes in other pull requests. Breaks compiling on AMD64 platform.

richael02:
So this new commit only adds the support our ARM platforms .And I have verified it.

configure
mtl1979: The define has existed long time before the padding was added... Patch for one architecture shouldn't touch code for another architecture. That however doesn't include lines that apply for multiple architectures and only change is adding more tests.

zlib.h
```diff
@@ -109,8 +109,9 @@ typedef struct z_stream_s {
     int                   data_type;  /* best guess about the data type: binary or text
                                          for deflate, or the decoding state for inflate */
     uint32_t              adler;      /* Adler-32 or CRC-32 value of the uncompressed data */
-#if defined(ZLIB_COMPAT) && defined(X86_64)
-    uint32_t              padding;    /* pad out to the same size as the zlib struct */
+#if defined(ZLIB_COMPAT) && (defined(X86_64) || defined(ARM_64BIT_LONG))
+    uint32_t              padding;    /* pad out to the same size as the zlib struct: padding
+                                         is needed when the size of long int is 64 bits */
```
mtl1979, line 114: This code is still wrong and pending removal within a week... The problem is that compiler automatically adds extra padding after variable "padding" even though the extra padding should be between "data_type" and "adler" on little-endian machines because both are 32-bit variables in zlib-ng, but "adler32" is 64-bit in zlib so still aligned differently.

richael02: I don't know how will you deal with padding finally . What I do is just making zlib-ng work well on out platforms. So I will wait for your final modification and then look into it.

mtl1979: ARM is little-endian by default, so you should have padding before and after "adler32"... both should be uint32_t when long int is 64 bits.
We will use two versions for zlib.h and the one used for ZLIB_COMPAT will be reverted to use same types as original zlib.

Dead2:
zlib.h / zlib-ng.h changes needs new review

mtl1979:
Unnecessary with latest merges.

PR #154, opened 2018-02-13
Fix dependency problem about cmake options

richael02:
According to the content of CMakeLists.txt, if building with "-DZLIB_COMPAT=ON",
the value of WITH_GZFILEOP should be ON too. However, WITH_GZFILEOP is OFF
actually when you run "cmake .. -DZIB_COMPAT=ON", which will cause errors if you
use gzfile related functions.

This patch fixes the problem by adjusting the position of WITH_GZFILEOP
option.


mtl1979:
Looks good to me.

PR #158, opened 2018-03-01
Fix the bug about byte-order on Arm platforms

richael02:
Zlib-ng tries to locate endian.h from <machine/endian.h> if __arm__
or __aarch64__ is defined.But it fails to find it and error comes.

This patch fixes the problem by relocating endian.h when __arm__ or
__aarch64__ is defined.

Change-Id: Ib35dd0992d03767d7321ef5db14246f5524c3f77

### DIFF
--- a/gzendian.h
+++ b/gzendian.h
@@ -15,7 +15,7 @@
 # else
 #  error Unknown endianness!
 # endif
-#elif defined(__APPLE__) || defined(__arm__) || defined(__aarch64__)
+#elif defined(__APPLE__)
 # include <machine/endian.h>
 #elif defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__bsdi__) || defined(__DragonFly__)
 # include <sys/endian.h>


mtl1979:
Doesn't break anything here using either gcc 5.3 or gcc 6.2.

richael02:
@mtl1979 I tested it both on AArch64  server and  Raspberry Pi3，and there came the problem. What platforms do you run on?

mtl1979:
@richael02 gcc 4.x, gcc 5.x and gcc 6.x ... All Linaro development kits for either 32-bit or 64-bit ARM. Also have SDK for Windows/ARMv7, but need to get a signing key for that...

richael02:
@mtl1979 So you use cross-compiling ,not on native platforms? Maybe it's the cause of difference.

mtl1979:
@richael02 Only native we have tested on is @dead2's Raspberry Pi... And the original version didn't fail there either... It's possible there is some gcc versions where headers are different as noticed when using for example size_t.

richael02:
@mtl1979 I also tested it on Raspberry Pi3 with gcc6.3.0, and endian.h is under /usr/include

mtl1979:
I compared the files and there is two lines missing in the version here... That's why it doesn't work for you... You can use other operating systems than Linux on ARM and the embedded version has <machine/endian.h> but Linux has <endian.h>

```
#ifdef __MINGW32__
# include <sys/param.h>
#elif defined(WIN32) || defined(_WIN32)
# define LITTLE_ENDIAN 1234
# define BIG_ENDIAN 4321
# if defined(_M_IX86) || defined(_M_AMD64) || defined(_M_IA64) || defined (_M_ARM)
#  define BYTE_ORDER LITTLE_ENDIAN
# else
#  error Unknown endianness!
# endif
#elif defined(__linux__)
# include <endian.h>
#elif defined(__APPLE__) || defined(__arm__) || defined(__aarch64__)
# include <machine/endian.h>
#elif defined(__FreeBSD__) || defined(__NetBSD__) || defined(__OpenBSD__) || defined(__bsdi__) || defined(__DragonFly__)
# include <sys/endian.h>
#elif defined(__sun) || defined(sun)
# include <sys/byteorder.h>
# if !defined(LITTLE_ENDIAN)
#  define LITTLE_ENDIAN 4321
# endif
# if !defined(BIG_ENDIAN)
#  define BIG_ENDIAN 1234
# endif
# if !defined(BYTE_ORDER)
#  if defined(_BIG_ENDIAN)
#   define BYTE_ORDER BIG_ENDIAN
#  else
#   define BYTE_ORDER LITTLE_ENDIAN
#  endif
# endif
#else
# include <endian.h>
#endif
```

richael02:
But as I know， we have no other operating system on AArch64 now. And I tried some third-party non-linux operating system on 32bit Arm( Raspberry Pi), but I haven't made  zlib-ng work there.

mtl1979:
All I know is that the old code works correctly and the last two revisions don't... What I use is based on work by Linaro team and verified to work on @Dead2's Raspberry Pi...

```
arm-eabi-gcc -O3 -Wall -DHAVE_HIDDEN -DHAVE_INTERNAL -DHAVE_BUILTIN_CTZL -DMEDIUM_STRATEGY -DUNALIGNED_OK -DUNROLL_LESS -I/c/build/git/zlib-ng -c -o crc32.o /c/build/git/zlib-ng/crc32.c
In file included from C:/build/git/zlib-ng/crc32.c:14:0:
C:/build/git/zlib-ng/gzendian.h:38:21: fatal error: endian.h: No such file or directory
 # include <endian.h>
                     ^
compilation terminated.
make: *** [Makefile:151: crc32.o] Error 1
```

Dead2:
Obsoleted

PR #164, opened 2018-03-14
Fix the problem about rule to make target "zconf.h" on Arm platforms

richael02:
If building zlib-ng with --acle option on Arm platforms, the building
process will stop in the meantime with the message "No rule to make
target zconf.h needed by crc32_acle.o".

This patch fixes the problem by including zconf.h or zconf-ng.h
according to the fact that whether ZLIB_COMPAT is defined or not in
crc32_acle.c.

Change-Id: Ib050c5b0e65d86210c8babdff5dbe670729fc63a


richael02:
I just follow the other files : they don't depend on zconf.h and zconf-ng.h, but zconf${SUFFIX}.h is add to them in Makefile.in.  But in fact I think we should remove the unnecessary zconf${SUFFIX}.h if the file doesn't depend on it.

mtl1979:
zconf.h/zconf-ng.h is used so we don't need to add special checks for system headers in every file... Essential system headers are always included.... This means we can use fixed-size integer types the same way in all files.

PR #165, opened 2018-03-14
Fix the problem about the consistency of crc32_acle on Arm platforms

richael02:
Make crc32_acle optional even though the Arm architectures have CRC
extention. Use crc32_acle only when built with --acle or -DWITH_ACLE=ON.

Change-Id: Ice51801d1bcdc57431c836db457fb31ec5ed0704


CMakeLists.txt
```diff
@@ -395,6 +395,7 @@ if(WITH_OPTIM)
             set(ZLIB_ARCH_SRCS ${ZLIB_ARCH_SRCS} ${ARCHDIR}/crc32_acle.c ${ARCHDIR}/insert_string_acle.c)
             set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${ACLEFLAG}")
             add_definitions("-DARM_ACLE_CRC_HASH")
+            add_definitions("-DARM_ACLE_CRC32")
```
mtl1979, line 398: Using two defines for same purpose is an overkill.

richael02: But crc32_acle.c  and insert_string_acle.c are for different purpose. Or if you think they should be enabled by the same define, we should merge ARM_ACLE_CRC_HASH and ARM_ACLE_CRC32 into one, like ARM_ACLE ?

richael02: and change it to : add_definitions("-DARM_ACLE")

mtl1979: If you want to keep ACLE enabled for insert_string_acle, make a new define for it instead... I'm just not sure at all it will compile if using more than one define for ACLE support...

richael02: I think it's better to use one define for both insert_stream_acle  and crc32_acle, because they both use the functions from arm_acle.h.  And do you think whether "ARM_ACLE"  is a reasonable define for them?

mtl1979: I think we should just keep the current define... What comes to functable detection, it will be rewritten soon...

Dead2:
This should be changed differently.

PR #166, opened 2018-03-14
Fix the bug when using arm-eabi as the tool chain

richael02:
Arm-eabi's target ARCH is ARMv7 bare-metal, not armv8-a.

Change-Id: Id2cd32515be09b40d3fd0866f8f65a2b6f775a4e

### DIFF
--- a/configure
+++ b/configure
@@ -268,12 +268,10 @@ if test "$gcc" -eq 1 && ($cc $CFLAGS -c $test.c) >> configure.log 2>&1; then
       if test "${uname}" = "eabi"; then
         uname=arm
       fi
-      if test $buildacle -eq 1; then
-        if test $native -eq 0; then
-          ARCH=armv8-a+crc
-        else
-          ARCH=native
-        fi
+      if test $native -eq 0; then
+        ARCH=armv7-a
+      else
+        ARCH=native
       fi ;;
     aarch64)
       if test "${uname}" = "elf"; then


configure
```diff
@@ -268,12 +268,10 @@ if test "$gcc" -eq 1 && ($cc $CFLAGS -c $test.c) >> configure.log 2>&1; then
...
-      if test $buildacle -eq 1; then
-        if test $native -eq 0; then
-          ARCH=armv8-a+crc
-        else
-          ARCH=native
-        fi
+      if test $native -eq 0; then
+        ARCH=armv7-a
+      else
+        ARCH=native
```
mtl1979, line 274: This block has nothing to do with arm-eabi... The previous block just sets uname to "arm" because eabi toolchain doesn't set it to recognizable value.

configure
```diff
@@ -268,12 +268,10 @@ if test "$gcc" -eq 1 && ($cc $CFLAGS -c $test.c) >> configure.log 2>&1; then
       if test "${uname}" = "eabi"; then
         uname=arm
       fi
```
mtl1979, line 270: This block sets arch to generic "arm" when arm-eabi toolchain is used.

configure
richael02: Although it sets uname to arm, but it goes to the wrong ARCH branch if built with --acle. It goes to  armv8-a+crc which is wrong!

mtl1979: We only allow enabling ACLE on armv8-a... NEON can be enabled on ARMv7 also...

Dead2:
This needs to be fixed differently.

richael02:
Hi,
    I  don’t know why you close this. I’m still working on it . And since the gcc tool chain  with suffix “eabi” is for ARMv7 which has no ACLE instructions support , so there’s no need to do “buildacle” judgement here. And shouldn’t make it armv8-a+crc here.

Best Regards,
Richael

mtl1979:
It was closed because proper fix needs complete overhaul of the configure script, separating pre-ARMv8 configuration flow from ACLE-supporting ARM flow...

richael02:
I’m wondering why every time I’m still working on the issue or bug I put forward, you just close my pull request or just denied my modification ,and commit a new one …  Would you please leave the issue to me? And if my pull request needs modification, you can give your suggestion and then we discuss?

mtl1979:
@richael02 Only @Dead2 can close issues and he has pretty strict rules about combining work by multiple developers... As I'm basically official maintainer of all the non-Intel ports (ARM, PPC) of zlib-ng, every commit touches my code, so I have the final say before Dead2. In past we have had to reject a lot of patches from people working for various subdivisions of ARM due to low quality or copyright issues.

We are pretty close to first public release, so we're running on tight schedule and that means fastest patch that actually works will get pushed to primary branch and other patches need to be rebased on top of the current branch tip.

PR #167, opened 2018-03-14
[ARM/AArch64] Add run-time detection of ACLE and NEON instructions under Linux.

mtl1979:
* Use getauxval() to check support for ACLE CRC32 instructions
* Allow disabling CRC32 instruction check


richael02:
I tested it on our AArch64 server, 32bit environment, as well as using the cross-compile toolchains on   https://releases.linaro.org/components/toolchain/binaries/5.4-2017.05/. And until now there are two  problems.
1. no modification in arch/aarch64/insert_string_acle.c ? Error comes when compiling on AArch64
 2.when using arm-eabi-gcc there's a  strange error: missing binary operator before the  token "(" WINAPI_PARTITION_PHONE_APP)"
in " #elif defined(_M_ARM) && defined(WINAPI_FAMILY_PARTITION) && WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_PHONE_APP)"

mtl1979:
@richael02 

1. Fixed
2. The strange error comes if gcc preprocessor parses the line in backwards order, which is something that it shouldn't do... To fix it, that needs to be split as two tests...

richael02:
Retested it and  there are no other problems with my test environment now.

landfillbaby:
you could use the [system register intrinsics](https://developer.arm.com/documentation/101028/0012/11--System-register-access) to do this independent of OS. you'd need ID_AA64ISAR0_EL1 for aarch64 and ID_ISAR5_EL1 for aarch32

mtl1979:
@landfillbaby Using low-level intrinsics is possible, but it will make the code unreadable to most people and prone to errors we can't test before merging in.

Dead2:
The commit message and the merge message state that this commit does completely different things. The commit message says it just moves code around, while the merge message says it introduces new features. This should be rectified. Perhaps split into a move-code commit and a feature commit?

mtl1979:
@Dead2 Seems like GitHub totally fucked up the pull request... I did rebase the commit and restore the original commit message...

Dead2:
@richael02 Are you able to do a performance test before/after this patch? I'd like to make sure that we actually improve performance with this, and that can't be tested in an emulator.

mtl1979:
I don't see the reason for performance test as whole purpose of the patch is to make sure code doesn't crash if run on processor without CRC32 instructions. Not crashing is always the preferred result over crashing.

Dead2:
@mtl1979 Is that so? Neither the commit message or pull request mention any crash fixes. That changes things.

mtl1979:
@Dead2 That's because this was part of bigger patch set I made a long time ago... To not crash requires also not using crc32 instrinsics for hashing, which is not part of this patch. Disabling crc32 intrinsics for hashing will slow down performance on processors that actually support those intrinsics, but that is the price of compatibility as we can't runtime switch macro contents.

Dead2:
Reverted due to build errors.
undefined reference to `arm_has_neon'

Needs a fix and better testing/review

PR #169, opened 2018-03-23
Separate feature checks for x86 and x86_64

mtl1979:
* Don't check for SSE2 on anything else than i686
* Don't check for PCLMULQDQ on anything else than i686 or x86_64
* Don't check for tzcnt support if maximum value of Basic CPUID information is less than 7
* Check for SSE4.2 CRC intrinsics

This avoids checking ARCH twice for correct defines.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -318,6 +318,17 @@ else()
         }"
         HAVE_SSE42_INTRIN
     )
+    check_c_source_compile_or_run(
+        "int main(void)
+        {
+            unsigned crc = 0;
+            char c = 'c';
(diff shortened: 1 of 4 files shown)


Dead2:
This is missing cmake support for X86_SSE4_2_CRC_INTRIN, would be nice to get that in at the same time. Otherwise, this looks great.

mtl1979:
@Dead2 I can add X86_SSE4_2_CRC_INTRIN check in CMakeLists.txt ;) Always value your feedback...

Dead2:
Thanks

PR #171, opened 2018-04-02
Add the check for -mfpu=neon

richael02:
For 64bit armv8-a there's no need to use "-mfpu=neon" to enable NEON.
But for 32bit system "-mfpu=neon" is required.

This patch adds the detection for -mfpu=neon flag.

Change-Id: Iec4f9e29f78806ae0e7021187038bb0b99eec30b

### DIFF
--- a/configure
+++ b/configure
@@ -1004,13 +1004,17 @@ case "${ARCH}" in
                 fi
 
                 if test $buildneon -eq 1; then
+                    CFLAGS="${CFLAGS} ${floatabi}"
+                    SFLAGS="${SFLAGS} ${floatabi}"
+
                     if test $MFPU_NEON_AVAILABLE -eq 1;then
-                        CFLAGS="${CFLAGS} ${floatabi} -mfpu=neon -DARM_NEON_ADLER32"
-                        SFLAGS="${SFLAGS} ${floatabi} -mfpu=neon -DARM_NEON_ADLER32"
-                    else
-                        CFLAGS="${CFLAGS} ${floatabi} -DARM_NEON_ADLER32"
-                        SFLAGS="${SFLAGS} ${floatabi} -DARM_NEON_ADLER32"
+                        CFLAGS="${CFLAGS} -mfpu=neon"
+                        SFLAGS="${SFLAGS} -mfpu=neon"
                     fi
+
+                    CFLAGS="${CFLAGS} -DARM_NEON_ADLER32"
+                    SFLAGS="${SFLAGS} -DARM_NEON_ADLER32"
+
                     ARCH_STATIC_OBJS="${ARCH_STATIC_OBJS} adler32_neon.o"
                     ARCH_SHARED_OBJS="${ARCH_SHARED_OBJS} adler32_neon.lo"
                 fi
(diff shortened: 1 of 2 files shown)


Dead2:
@mtl1979 Any comments?

mtl1979:
@Dead2 Obviously we should always test if some flag is actually supported on the current toolchain. Assuming is real bad habit.

richael02:
I said that for 64bit armv8-a there's no need to use "-mfpu=neon" to enable NEON.
But for 32bit armv8-a system and  armv7 system, "-mfpu=neon" is required. I both verified on our 64bit armv8-a  server and the 32bit  environment.  I add the check because I think it's a simple way to distinguish whether it's a 64bit armv8-a or a 32bit one. Of course I have thought that there's other way to distinguish them. This's just a choice.

Dead2:
This needs to be rebased due to a conflict in configure.

configure
```diff
@@ -992,9 +1004,13 @@ case "${ARCH}" in
...
-                    CFLAGS="${CFLAGS} ${floatabi} -DARM_NEON_ADLER32"
-                    SFLAGS="${SFLAGS} ${floatabi} -DARM_NEON_ADLER32"
-
+                    if test $MFPU_NEON_AVAILABLE -eq 1;then
+                        CFLAGS="${CFLAGS} ${floatabi} -mfpu=neon -DARM_NEON_ADLER32"
+                        SFLAGS="${SFLAGS} ${floatabi} -mfpu=neon -DARM_NEON_ADLER32"
+                    else
+                        CFLAGS="${CFLAGS} ${floatabi} -DARM_NEON_ADLER32"
+                        SFLAGS="${SFLAGS} ${floatabi} -DARM_NEON_ADLER32"
+                    fi
```
mtl1979: Instead of duplicating the add of `${floatabi}` in else block, you should move adding it above the IF block, and adding `-DARM_NEON_ADLER32` below the IF block, so they are done no matter what value $MFPU_NEON_AVAILABLE has... This is consistent how it is done on other architectures. It makes the code a little longer, but a lot more readable.

richael02: Would it be better to put ${floatabi} and -DARM_NEON_ADLER32 together below IF block?
"
                    if test $MFPU_NEON_AVAILABLE -eq 1;then
                        CFLAGS="${CFLAGS} -mfpu=neon"
                        SFLAGS="${SFLAGS} -mfpu=neon"
                    fi
                    CFLAGS="${CFLAGS} ${floatabi} -DARM_NEON_ADLER32"
                    SFLAGS="${SFLAGS} ${floatabi} -DARM_NEON_ADLER32"
                    ARCH_STATIC_OBJS="${ARCH_STATIC_OBJS} adler32_neon.o"
                    ARCH_SHARED_OBJS="${ARCH_SHARED_OBJS} adler32_neon.lo"
"

mtl1979: I usually don't like changing order of compiler parameters, because the result can be unpredictable. That's why we previously used empty variables instead when something was optional.

Dead2:
Just a thought here, do we need to reset CMAKE_REQUIRED_FLAGS after running the check or does it auto-reset? If it does not, then I suspect we can at some point get problems where CheckCXXSourceCompiles fails on some platforms because an unrelated flag was set for a previous check and was assumed to be empty in the later check.

I checked the documentation, but it says nothing about this, so I think it does not reset. If so, I think we should make sure that each check re-sets the variables back to empty after the check has completed. Are there any cmake experts around?

Other than that, this PR looks good to me.

mtl1979:
@Dead2 With most cmake variables, you need to explicitly append or prepend to them if you want to preserve the old contents.

richael02:
@Dead2  Since we use “  **set (CMAKE_REQUIRED_FLAGS    ”new-flag”)**  “ here， the  ”new-flag“ will replace the old  CMAKE_REQUIRED_FLAGS. It's different from this way: "  **set (CMAKE_REQUIRED_FLAGS    " ${CMAKE_REQUIRED_FLAGS}   new-flag")**   "   . And if  we want to clear the last CMAKE_REQUIRED_FLAGS，” **set (CMAKE_REQUIRED_FLAGS)**  " can do it.

Dead2:
@richael02 That's what I expected then. So if there are two compile-checks after each other, and the first one sets a flag, then the next one also gets that flag, changing the way that test works. So I think we'd better start resetting these flags after each test, before we get bitten by a bug.

richael02:
@Dead2  Maybe there's some misunderstanding. I mean  if  there are two compile-checks like this:
1. set(CMAKE_REQUIRED_FLAGS  "flag-A")
.....
2. set(CMAKE_REQUIRED_FLAGS "flag-B")
.....
the second one will not get the first flag-A. And there's no need to reset it. This is what we use in CMakeLists.txt. 
But if we write like this:
1. set(CMAKE_REQUIRED_FLAGS  "${CMAKE_REQUIRED_FLAGS}   flag-A")
.....
2. set(CMAKE_REQUIRED_FLAGS  "${CMAKE_REQUIRED_FLAGS}  flag-B")
.....
the second one will get the first flag-A and others.

Dead2:
@richael02 Yes, but what if the second test does not set the flag at all, and expects it to be empty?

For example first there is a test checking for compiler flag, and the second test is to check for a supported intrinsic. If the compiler flag does not get reset after the first test, the second test will run with both the compiler flag and the intrinsic. So if the compiler flag fails, the intrinsic also will fail, even though it should have passed.

richael02:
@Dead2 Hum I forgot to consider this situation.... If so we should reset CMAKE_REQUESTED_FLAGS to empty after each related check.

Dead2:
@richael02 Yes, I think that is the best approach to avoid future problems.

richael02:
@Dead2 Should I open another pull request to do this, for it's not related to "neon" issue?

Dead2:
You could add a reset after the check in this PR, but resets for the other checks should be in a separate PR, yes please 👍

richael02:
Ok, no problem.

Dead2:
Thanks 👍

PR #172, opened 2018-04-19
Add ARM cross build and validation

JunHe77:
Update configure and test scripts to cross compile
and validate arm build-outs on x86 by using qemu

Change-Id: I183d003ebafcf686de26fd1705704ded4b344580
Signed-off-by: Jun He <jun.he@arm.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,4 +47,79 @@ matrix:
       compiler: clang
       env: BUILDDIR=../build TOOL="cmake ../zlib-ng"
 
+    # Cross compiling for arm variants
+    - os: linux
+      compiler: aarch64-linux-gnu-gcc
+      addons:
+        apt:
+          packages:
+            - qemu
+            - gcc-aarch64-linux-gnu
+            - libc-dev-arm64-cross
+      # For all aarch64 implementations NEON is mandatory, while crypto/crc are not.
+      # So for aarch64 NEON should be automatically enabled, and "acle" should be explicitly defined
+      env: CHOST=aarch64-linux-gnu BUILDDIR=. TOOL="./configure --warn --zlib-compat"
+    - os: linux
+      compiler: aarch64-linux-gnu-gcc
+      addons:
+        apt:
+          packages:
+            - qemu
+            - gcc-aarch64-linux-gnu
+            - libc-dev-arm64-cross
+      # AArch64 toolchain in Trusty is 4.8.2, which doesn't support acle intrinsics.
+      # Until TravisCI provides Xenial dist (which comes with gcc-aarch64-5.3.1) 'acle' option cannot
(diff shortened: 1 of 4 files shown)


mtl1979:
This is bad. Does not actually test qemu supports running non-native binaries.

JunHe77:
Hi, @mtl1979 , may I know why qemu cannot be used for testing non-native binaries? I can try to solve them if possible. 😃 

As for this patch, enabling cross compile will help build test coverage for people without actual hw.
And as the built-out binaries are all working in user mode, qemu-* should be enough to validate if these binaries are basicaly working.
FYR, an sample output can be found at: https://travis-ci.org/JunHe77/zlib-ng/builds/368500583

Pls let me know how I can improve this patch, cause I think it is really matters to verify zlib-ng for non x86 platforms in CI. Thanks.

mtl1979:
Obviously it should test that the QEMU_RUN executable actually exists as it can fail if someone tries to configure outside Travis.

JunHe77:
If someone doesn't have QEMU installed, the whole build process still goes well. It only fails when the user runs "make test" or "make check", output as following:
`/bin/sh: 2: /bin/sh: 2: qemu-aarch64: not foundqemu-aarch64: not found`
This error message is pretty straightforward to know what is missing and test is stopped without creating any misleading/wrong result. I'll say this behavior is acceptable. 😁 

If you prefer to give user more/direct information on this failure, is following way OK?
`builder@06d6c1a2f66a:~/zlib-ng$ make -C test`
`make: Entering directory '/home/builder/zlib-ng/test'`
`Makefile:24: *** "You need QEMU to run tests on non-native platform".  Stop.`
`make: Leaving directory '/home/builder/zlib-ng/test`

mtl1979:
I think human readable error message is better...

JunHe77:
Patch is updated. If no qemu-* installed, user will see following error message:
`Makefile:24: *** "You need QEMU to run tests on non-native platform". Stop.`

mtl1979:
Just a minor style issue remaining in the patch... in "configure" you removed two empty lines outside the patched section... Generally those kind of changes should not be bundled with code changes. It's easier for git to track changes, when the patch is as short as possible.

JunHe77:
Thanks for pointing it out. I've updated it. Pls review.

JunHe77:
Ping, @mtl1979 . Is there any other issues need to be fixed? Thanks.

mtl1979:
@JunHe77 I'm waiting for @Dead2 to give his opinion... I don't have commit rights...

Dead2:
Looks great, thanks

JunHe77:
Thank you, @mtl1979, for patient review and help. Thank you, @Dead2 for merge this. 😄

PR #174, opened 2018-05-02
Travis: add tests for native linux ppc64le

grooverdan:
Also fixes to not assume i[36]86 if no other architecture.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,21 @@ matrix:
       compiler: clang
       env: BUILDDIR=../build TOOL="cmake ../zlib-ng"
 
+    # compiling for linux-ppc64le variants
+    - os: linux-ppc64le
+      compiler: gcc
+      env: BUILDDIR=. TOOL="cmake ."
+    - os: linux-ppc64le
+      compiler: gcc
+      env: BUILDDIR=../build TOOL="cmake ../zlib-ng"
+
+    - os: linux-ppc64le
+      compiler: clang
+      env: BUILDDIR=. TOOL="./configure --warn --zlib-compat"
+    - os: linux-ppc64le
+      compiler: clang
+      env: BUILDDIR=../build TOOL="cmake ../zlib-ng"
+
     # Cross compiling for arm variants
     - os: linux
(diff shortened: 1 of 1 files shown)


mtl1979:
Too early for this... Should wait until other PPC related patches are merged in... Otherwise very good idea...

grooverdan:
ok. I've been meaning to get back to the PPC patches. I'll hurry up.

Dead2:
This needs to be rebased due to conflicts in CMakeLists.txt

grooverdan:
rebased. Should succeed with a ARCH=generic now. I will get back to ppc64 optimizations. found a big endian problem that's taking a bit longer than expected.

mtl1979:
I got a lot of issues when I tried PPC big-endian due to some optimizations for x86 that conflicted with how PPC works... I had to basically make own version of some functions just for PPC, so it doesn't try to use the x86 versions as those are unusable on big-endian architectures.

grooverdan:
Have you got a list of functions/patches/fixes available for testing? I'll clean them up if you don't have time.

mtl1979:
@grooverdan Basically the functions I was talking about are in commit https://github.com/mtl1979/zlib-ng/commit/0c4dc5a170b397a8bdf789dad4f74fb8840aea0f

Whole test branch I used can be downloaded from https://github.com/mtl1979/zlib-ng/tree/ppc-test

The gcc version I used for testing didn't support PPC64 fully so I couldn't compile the VSX optimized code.

Dead2:
Thanks, even without ppc-specific code, this helps a little to avoid potential regressions.

grooverdan:
thanks for merging.

@mtl1979 thanks for the code. I'll test in on the gcc-4.8.5 from SUSE/RHEL7 which I think is the oldest supported around.

mtl1979:
@grooverdan The PPC64 toolkit I had did have broken headers, I haven't checked if later versions have fixed them. Due to PPC being strict about integer types when using vectors, it would have required casting back and forth between unsigned and signed types to get it work correctly and that made the code look really ugly.

PR #175, opened 2018-05-15
CMake: don't assuming i[36]86 if others don't match

grooverdan:
there are more architectures than just x86_64, arm, aarch64 and i[36]86 so lets actually do a pattern match before just assuming its i[36]86]. There's a arch/generic build dir so that will be suitable for everything else however forcing SSE2 is unlikely to work.


mtl1979:
The pattern is wrong... if we want to ignore the fact that i386 doesn't even support SSE, we should include everything from i386 through i686, not just the two processor classes. There is a lot of Linux versions that use i586 instead of i686 and we should not forget i486, that did also exist.

grooverdan:
quite right updated. Notice runtime checks for sse2 are occuring so didn't bother with compile time checks.

Dead2:
Thanks

grooverdan:
thanks for the merge

PR #176, opened 2018-05-31
Reset CMAKE_REQUIRED_FLAGS

richael02:
Reset CMAKE_REQUIRED_FLAGS after each check to avoid the following
checks being influenced by the previous results.

Change-Id: I2e34f6127ef1c617f4eea363a2cb80bc49b3bcab

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -303,6 +303,8 @@ else()
         }"
         HAVE_SSE2_INTRIN
     )
+    set(CMAKE_REQUIRED_FLAGS)
+
     if(WITH_NATIVE_INSTRUCTIONS)
         set(CMAKE_REQUIRED_FLAGS "${NATIVEFLAG}")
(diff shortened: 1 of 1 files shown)


mtl1979:
Resetting CMAKE_REQUIRED_FLAG before it is unconditionally set is overkill... It should be only reset if we can't guarantee the contents are unconditionally overwritten.

Dead2:
@mtl1979 I don't see any negative sides to doing this. It helps us avoid future oversights causing strange regressions and keeps the Cmake code easy to read. Resetting the variable does not cost much cpu-time and is only a single line of code.

mtl1979:
@Dead2 It doesn't remove the possibility that someone makes a patch that adds code to wrong side of the reset command... It doesn't actually make the code easier to read, it makes it more confusing for anyone who actually can use cmake properly.

PR #177, opened 2018-06-12
replaced include_directories() with target_include_directories()

SleepySpaceBear:
using target_include_directories() with the zlib libraries prevents people from having to manually include those directories when linking to those libraries


Dead2:
Thanks

PR #178, opened 2018-06-21
Make functable structure thread-local variable

mtl1979:
* Using thread-local variable for structure is faster than making other threads wait for first thread to update the structure members one by one
* C doesn't have portable implementation of locks and mutexes like C++
* Use __declspec(thread) for Visual C++ and __thread for other compilers


Dead2:
I'd like this a lot more if the "Move private defines" commit actually only moved code and did not introduce new arch-specific code. Those should be split into two commits. And this PR only seems to reference the other commit, so maybe that first commit is not actually intended to be in this PR at all?

Also, do we need thread-local variables if we are not thread-safe in the first place?

mtl1979:
This PR depends on the changes of the previous commit, so GitHub decided to show it also here... When that commit is merged in, I can rebase this branch/PR to look a lot more cleaner.

This is just first part of making zlib-ng thread-safe again... 
Eliminating all code that makes it not thread-safe.

Dead2:
Cherry-picked

PR #188, opened 2018-08-16
Enable clang scan build static analysis

gdevenyi:
Enable the scan-build function for clang-based builds.

The ``--status-bugs`` option marks the build as "failed" if scan-build detects any bugs.

scan-build currently finds one bug:
```
In file included from zlib-ng/inflate.c:88:
zlib-ng/memcopy.h:298:5: warning: Value stored to 'from' is never read
    from += rem;
    ^       ~~~
1 warning generated.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: c
 cache: ccache
-dist: trusty
+dist: xenial
 
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,61 +6,64 @@ matrix:
   include:
     - os: linux
       compiler: gcc
-      env: BUILDDIR=. TOOL="./configure --warn"
+      env: MAKE="make" BUILDDIR=. TOOL="./configure --warn"
     - os: linux
       compiler: gcc
-      env: BUILDDIR=. TOOL="cmake . -DZLIB_COMPAT=OFF -DWITH_GZFILEOP=ON -DWITH_NEW_STRATEGIES=YES -DWITH_OPTIM=ON"
+      env: MAKE="make" BUILDDIR=. TOOL="cmake . -DZLIB_COMPAT=OFF -DWITH_GZFILEOP=ON -DWITH_NEW_STRATEGIES=YES -DWITH_OPTIM=ON"
     - os: linux
       compiler: gcc
-      env: BUILDDIR=../build TOOL="../zlib-ng/configure --warn --zlib-compat"
+      env: MAKE="make" BUILDDIR=../build TOOL="../zlib-ng/configure --warn --zlib-compat"
     - os: linux
       compiler: gcc
-      env: BUILDDIR=. TOOL="./configure --warn --zlib-compat --without-optimizations --without-new-strategies"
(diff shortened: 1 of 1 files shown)


mtl1979:
That one line can be safely removed to make Travis happy...

gdevenyi:
Happy to make a seperate PR for that. I'm not well-versed enough in this code to assume it was safe to remove :)

gdevenyi:
Rebased to run against recent code. Still has some bugs.

gdevenyi:
This is looking okay, except that OSX changed its name for scan-build, I think?

And cmake is failing on some compile tests on the ppc64le arch.

Reading the discussion on oss-fuzz, should I refactor so there's only one static analysis run?

Dead2:
Perhaps MAKE can be set globally and only overridden for the few instances where it is needed?
Also moving MAKE to the end of the line, to match the logical order of commands, would make it easier to read the travis file.

We should make sure we don't run too many duplicate tests on travis, there is a certain amount of cpu time we can use, so we should try to minimize that if possible.

The PR would of course need to pass travis tests too, either by fixing bugs in the code or disabling tests that cannot be fixed.

I am honestly also thinking this could be done more cleanly and more usefully inside cmake instead of limiting it to only travis.  cmake -DWITH_SCANBUILD or something would be a useful command to run before doing a commit instead of waiting for the (often delayed) travis build.
Do you feel like you could refactor this to happen inside cmake instead?

Dead2:
Looks like travis blocked my account due to suspected abuse (I know nothing more than that), and then it just silently does not run the tests. I finally finished my buildkite setup though, and got it installed on one server, so we can at least do some testing until travis fixes my account.

But this certainly shows us that we cannot completely trust travis, and should if possible make it easy to implement testing on other CIs too, so count that as a vote for scan-build in cmake :)

Also, with buildkite, because it only runs on our own servers, we are free to use a lot more cputime :)
For now only an x86_64 linux server, but I have three more servers to add when I get the time, it will give us native testing on arm, aarch64 and a sparc-something.

gdevenyi:
I don't know much about cmake other than using it, but maybe this is my excuse to figure it out :)

Re: travis, what about CircleCI?

Dead2:
Looks like CircleCI fills the exact same role as BuildKite does, so I guess we will stay with BuildKite for now.

Cmake is not hard, although it is not always logical and the documentation often leaves something to be desired (like relevant examples), but it is relatively easy to get the hang of it. And it is a good skill to have.

gdevenyi:
When will that be up and running so I can iterate against those builds rather than Travis?

Dead2:
If you get it running in CMake, then there is no need for iterating against Travis or similar. You can just run "cmake -Dscanbuild=ON && make" for example to test it locally, and once it works it can be run the same way in Buildkite.
Buildkite is already being run with basic tests against recent PRs and Commits.

gdevenyi:
Okay, so based on all my reading, its not recommended to do hard-bake scan-build into the cmake config.

Instead they simply recommend "scan-build cmake" and "scan-build make". If you can expose the BuildKite config somehow I can work on that, otherwise, it looks like you've dropped Travis, although I got it working since :)

https://travis-ci.org/gdevenyi/zlib-ng/jobs/466647961

PR #189, opened 2018-08-16
Include option to build minizip

nmoinvaz:
This pull requests adds a _WITH_MINIZIP_ option to be able to build the minizip project against zlib-ng as an external project in cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -700,21 +700,23 @@ endif()
 # minizip
 #============================================================================
 
-option(WITH_MINIZIP "Build minizip" ON)
+option(WITH_MINIZIP "Build minizip" OFF)
+add_feature_info(WITH_MINIZIP WITH_MINIZIP "Build minizip project")
 if (WITH_MINIZIP)
-    get_target_property(ZLIB_LIBRARY zlibstatic LOCATION_${CMAKE_BUILD_TYPE})
     ExternalProject_Add(minizip
         GIT_REPOSITORY    https://github.com/nmoinvaz/minizip.git
         PREFIX            ${CMAKE_BINARY_DIR}/minizip.prefix
         SOURCE_DIR        ${CMAKE_BINARY_DIR}/minizip
         BINARY_DIR        ${CMAKE_BINARY_DIR}/minizip
         INSTALL_DIR       ${CMAKE_INSTALL_PREFIX}/../minizip
-        INSTALL_COMMAND   ${CMAKE_COMMAND} --build . --target install --config ${CMAKE_BUILD_TYPE}
         CMAKE_ARGS
             -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
-            -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
-            -DZLIB_LIBRARY=${ZLIB_LIBRARY}
+            -DZLIB_LIBRARY=$<TARGET_FILE:zlibstatic>
             -DZLIB_INCLUDE_DIR=${CMAKE_SOURCE_DIR}
             -DBUILD_TEST=ON
     )
 endif()
+
+#============================================================================
+
+FEATURE_SUMMARY(WHAT ALL INCLUDE_QUIET_PACKAGES)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -708,7 +708,7 @@ if (WITH_MINIZIP)
         PREFIX            ${CMAKE_BINARY_DIR}/minizip.prefix
         SOURCE_DIR        ${CMAKE_BINARY_DIR}/minizip
         BINARY_DIR        ${CMAKE_BINARY_DIR}/minizip
-        INSTALL_DIR       ${INSTALL_BIN_DIR}
+        INSTALL_DIR       ${CMAKE_INSTALL_PREFIX}
         CMAKE_ARGS
             -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
             -DZLIB_LIBRARY=$<TARGET_FILE:zlibstatic>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -708,7 +708,7 @@ if (WITH_MINIZIP)
         PREFIX            ${CMAKE_BINARY_DIR}/minizip.prefix
         SOURCE_DIR        ${CMAKE_BINARY_DIR}/minizip
         BINARY_DIR        ${CMAKE_BINARY_DIR}/minizip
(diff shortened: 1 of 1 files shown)


donny-dont:
Should it be versioned to a tag or is always master ok?

nmoinvaz:
I think master would be okay. The calls to the zlib-ng library will almost never change.

Dead2:
I get the following errors when I try this
``` CMake Error at /usr/share/cmake/Modules/ExternalProject.cmake:1654 (file):
   file problem creating directory: /usr/local/../minizip
   /usr/share/cmake/Modules/ExternalProject.cmake:3052 (_ep_set_directories)
   CMakeLists.txt:706 (ExternalProject_Add)

 CMake Error at /usr/share/cmake/Modules/ExternalProject.cmake:1656 (message):
   dir '/usr/local/../minizip' does not exist after file(MAKE_DIRECTORY)
   /usr/share/cmake/Modules/ExternalProject.cmake:3052 (_ep_set_directories)
   CMakeLists.txt:706 (ExternalProject_Add)
```

The problem is likely with
```INSTALL_DIR       ${CMAKE_INSTALL_PREFIX}/../minizip```
And at least on my machine, this resolves to /usr/minizip.

Now, I think minizip's INSTALL_DIR should probably be set to the same as INSTALL_BIN_DIR.
This is from my CMakeCache.txt
``` INSTALL_BIN_DIR:PATH=/usr/local/bin ```

**Comment by nmoinvaz** [comment]
I'm not sure what the value of INSTALL_DIR should be:
 
When I tried `${INSTALL_BIN_DIR}/minizip`, it tries to install into `C:/Program Files (x86)/zlib/bin/minizip`.
When I tried `${CMAKE_INSTALL_PREFIX}/minizip` it tries to install into `C:/Program Files (x86)/zlib/minizip`.

I think what we want is it to install into `C:/Program Files (x86)/minizip`.

**Comment by Dead2** [comment]
Hmm, but if it is installed as part of zlib-ng, it might make sense to install it to the same place as zlib-ng. That might also be required for it to find the zlib-ng library.
I am not sure how to best solve this for windows, as there is no one correct way to do this AFAIK.
But we might have to handle windows separately if installing into the same folder as zlib-ng is not acceptable.

**Comment by nmoinvaz** [comment]
I think as an ExternalProject, that might be the way to go - having minizip inside of zlib-ng directory.

**Comment by nmoinvaz** [comment]
I have updated the install directory. Travis failure is due to failure of Travis to run instance.

**Comment by Dead2** [comment]
This is still a problem on *nix platforms. Installing it into a subdirectory of */bin/ would make it not present in PATH. It should be installed into the same directory as zlib-ng. So directly to ${INSTALL_BIN_DIR}.
I am unsure what BINARY_DIR specifies, and whether that also needs to be changed.

**Comment by nmoinvaz** [comment]
Using ${INSTALL_BIN_DIR} it created `zlib-ng/bin/bin`, `zlib-ng/bin/include`, `zlib-ng/bin/lib`, etc. I think where it should have been installed to is the same directory as zlib-ng which is what ${CMAKE_INSTALL_PREFIX} does. I removed the `minizip` from the INSTALL_DIR and that should fix the PATH issue you mentioned. BINARY_DIR is the build directory.
https://cmake.org/cmake/help/latest/module/ExternalProject.html

**Comment by Dead2** [comment]
I re-tested this now, and noticed a few things.
Minizip does not support zlib-ng without zlib-compat enabled, so either we need to check for that and error out in cmake if both are enabled (see similar cmake test in PR #206), or minizip needs to gain support for zlib-ng function prefixes etc.

When compiling zlib-ng with minizip enabled, minizip will automatically try to install itself during the compilation phase. It seems really weird that cmake defaults to this, but I think we need to workaround this (it will fail the compile when run as non-root user for example). I think perhaps the best solution is if minizip sets install dir to the zlib-ng build dir, and then pretend the minizip files are part of zlib-ng for the install step. I found some info around this:
https://stackoverflow.com/questions/37838786/how-to-not-make-install-step-when-building-external-project-with-cmake/37855224

I also noticed that minizip compilation will break if libbsd-devel is not installed, even though its cmake is checking for its presence. Perhaps something more needs to be disabled if libbsd is not present? Or is libbsd a hard requirement? If so, then perhaps you should fail during cmake instead of during linking.
```bash
Scanning dependencies of target minigzip
[ 81%] Building C object CMakeFiles/minigzip.dir/test/minigzip.c.o
[ 82%] Linking C executable minigzip
[ 82%] Built target minigzip
[ 84%] Performing update step for 'minizip'
Current branch master is up to date.
Already up to date!
[ 85%] Performing configure step for 'minizip'
-- Using CMake version 3.11.0
-- Found ZLIB: /home/hansr/github/zlib/zlib-ng/libz.a (found version "1.2.11") 
-- Using ZLIB 
-- Could NOT find BZip2 (missing: BZIP2_LIBRARIES BZIP2_INCLUDE_DIR) 
-- Checking for module 'libbsd'
--   Package 'libbsd', required by 'virtual:world', not found
-- Configuring done
-- Generating done
-- Build files have been written to: /home/hansr/github/zlib/zlib-ng/minizip
[ 87%] Performing build step for 'minizip'
[ 94%] Built target libminizip
[ 96%] Linking C executable minizip
libminizip.a(mz_crypt_brg.c.o): In function `mz_crypt_rand':
mz_crypt_brg.c:(.text+0xa): undefined reference to `arc4random_buf'
make[5]: *** [CMakeFiles/minizip.dir/build.make:101: minizip] Error 1
make[4]: *** [CMakeFiles/Makefile2:68: CMakeFiles/minizip.dir/all] Error 2
make[3]: *** [Makefile:130: all] Error 2
make[2]: *** [CMakeFiles/minizip.dir/build.make:111: minizip.prefix/src/minizip-stamp/minizip-build] Error 2
make[1]: *** [CMakeFiles/Makefile2:216: CMakeFiles/minizip.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

nmoinvaz:
I'm closing this pull request. Instead of including minizip in zlib cmake, I've included zlib in minizip cmake as ExternalProject (see my dev branch). If you use these options in minizip cmake, it will compile with zlib-ng:

`-DZLIB_REPOSITORY=https://github.com/Dead2/zlib-ng -DZLIB_TAG=develop`

nmoinvaz:
I'm closing this pull request. Instead of including minizip in zlib cmake, I've included zlib in minizip cmake as ExternalProject. If you use these options in minizip cmake, it will compile with zlib-ng:

`-DZLIB_REPOSITORY=https://github.com/Dead2/zlib-ng -DZLIB_TAG=develop`

PR #193, opened 2018-08-20
Fix: unknown type name 'va_list'

mikeeremin:
Fix: unknown type name 'va_list' on linux build (gcc 7.3.1 on Centos 7)

### DIFF
--- a/zlib.h
+++ b/zlib.h
@@ -31,6 +31,7 @@
 */
 
 #include <stdint.h>
+#include <stdarg.h>
 #include "zconf.h"
 
 #ifdef __cplusplus


amosbird:
I just ran into this too. I wonder why this happened?

mtl1979:
It's not uncommon that compiler vendors change the standard headers... Sometimes they separate two headers from each other to make sure there will be no conflicts due to including two or more headers in unexpected order.

As long as new header files are not introduced or old ones removed, it will be easy to fix this kind of issues.

amosbird:
@mtl1979 Well it only happens to Centos. Ubuntu build works fine with the same version of gcc

mtl1979:
@amosbird As far as I know, Centos is based on RedHat and Ubuntu is based on Debian... As such they have diverged a lot during history of Linux and have different people maintaining the packages.

amosbird:
>  It's not uncommon that compiler vendors change the standard headers

I thought you were talking about gcc or clang. Do you suggest that somehow the centos distro maintainers split the headers by themselves?

Dead2:
cherry-picked this to develop, thanks!

Issue #184, opened 2018-08-09
MemorySanitizer: use-of-uninitialized-value in arch/x86/crc_folding.c:423


sebpop:
```
export CC=clang-6.0
export CFLAGS="-O3 -g -fsanitize=memory"
export LDFLAGS="$CFLAGS"
./configure                                                                                                                                                                                                                                                    
make clean                                                                                                                                                                                                                                                     
make test
```
produces the following warnings:
```
==19318==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x4c950a  (/work/spop/zlib-ng/minigzip+0x4c950a)
    #1 0x4c1775  (/work/spop/zlib-ng/minigzip+0x4c1775)
    #2 0x4a041b  (/work/spop/zlib-ng/minigzip+0x4a041b)
    #3 0x49706a  (/work/spop/zlib-ng/minigzip+0x49706a)
    #4 0x4976ca  (/work/spop/zlib-ng/minigzip+0x4976ca)
    #5 0x498ce3  (/work/spop/zlib-ng/minigzip+0x498ce3)
    #6 0x7f3b62fd782f  (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
    #7 0x41a9c8  (/work/spop/zlib-ng/minigzip+0x41a9c8)

SUMMARY: MemorySanitizer: use-of-uninitialized-value (/work/spop/zlib-ng/minigzip+0x4c950a) 
Exiting
zlib version 1.9.9 = 0x1990, compile flags = 0xa9
==19329==WARNING: MemorySanitizer: use-of-uninitialized-value
==19322==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x7f9fc6832aba  (/work/spop/zlib-ng/test/../libz-ng.so.1+0x36aba)
    #1 0x7f9fc6832c03  (/work/spop/zlib-ng/test/../libz-ng.so.1+0x36c03)
    #2 0x7f9fc6807845  (/work/spop/zlib-ng/test/../libz-ng.so.1+0xb845)
    #3 0x4972da  (/work/spop/zlib-ng/minigzipsh+0x4972da)
    #4 0x49793a  (/work/spop/zlib-ng/minigzipsh+0x49793a)
    #5 0x498f53  (/work/spop/zlib-ng/minigzipsh+0x498f53)
    #6 0x7f9fc645282f  (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
[...]

PR #203, opened 2018-08-26
fix bug #184: clear out buf to avoid msan use-of-uninitialized-value

sebpop:
Do not use bzero as suggested by Mika Lindqvist:
> You shouldn't use bzero() in new code as some compilers, like Visual C++,
> don't have it... New code should just use memset().


Dead2:
Tests ok, and looks good to me.
Slight performance reduction in minigzip.

Issue #207, opened 2018-09-19
undefined behavior in deflate_slow.c:112


sebpop:
```
deflate_slow.c:112:21: runtime error: unsigned integer overflow: 45871 - 45872 cannot be represented in type 'unsigned int'
```

sebpop:
#202 under review fixes some of the errors that I reported.  I removed those entries from the description.

sebpop:
```
deflate_slow.c:112:21: runtime error: unsigned integer overflow: 45871 - 45872 cannot be represented in type 'unsigned int'
```
occurs because
```
zutil.h:#define MIN_MATCH  3

// s->lookahead is equal 2 when executing the following statement:
unsigned int max_insert = s->strstart + s->lookahead - MIN_MATCH;
[...]
if (unlikely(insert_cnt > max_insert - s->strstart))
```
so `max_insert` is `45871` and `s->strstart` is `45872`.

mtl1979:
If the result of either side can be negative, then either the both sides must be evaluated as signed types or possibly the test needs to be reorganized so that substraction on one side instead becomes addition on the other side.

So ```insert_cnt > max_insert - s->strstart``` equals to ```s->strstart + insert_cnt > max_insert```.

sebpop:
The problem with that is that the next statement under the condition is again the same expression:
```
  if (unlikely(insert_cnt > max_insert - s->strstart))
      insert_cnt = max_insert - s->strstart;

  functable.insert_string(s, s->strstart + 1, insert_cnt);
```
if we replace the condition with `insert_cnt + s->strstart > max_insert` we still have to replace the computation of insert_cnt.

I think the problem is that s->lookahead is too small.  At the beginning of the loop we have the following condition:
```
/* Make sure that we always have enough lookahead, except
* at the end of the input file. We need MAX_MATCH bytes
* for the next match, plus MIN_MATCH bytes to insert the
* string following the next match.
*/
if (s->lookahead < MIN_LOOKAHEAD) {
  functable.fill_window(s);
  if (s->lookahead < MIN_LOOKAHEAD && flush == Z_NO_FLUSH) {
    return need_more;
  }
  if (s->lookahead == 0)
    break; /* flush the current block */
}

deflate.h:#define MIN_LOOKAHEAD (MAX_MATCH+MIN_MATCH+1)
zutil.h:#define MIN_MATCH  3
zutil.h:#define MAX_MATCH  258
```
[...]

PR #209, opened 2018-09-24
fix bug #207: avoid undefined integer overflow

sebpop:
zlib-ng used to fail when compiled with UBSan with this error:
deflate_slow.c:112:21: runtime error: unsigned integer overflow: 45871 - 45872 cannot be represented in type 'unsigned int'

The bug occurs in the code under `#ifndef NOT_TWEAK_COMPILER`.
When `lookahead < MIN_MATCH` we now go back to the original
step by step computation under `#ifdef NOT_TWEAK_COMPILER`.

### DIFF
--- a/deflate_slow.c
+++ b/deflate_slow.c
@@ -108,11 +108,13 @@ ZLIB_INTERNAL block_state deflate_slow(deflate_state *s, int flush) {
 #else
             {
                 unsigned int mov_fwd = s->prev_length - 2;
-                unsigned int insert_cnt = mov_fwd;
-                if (unlikely(insert_cnt > max_insert - s->strstart))
-                    insert_cnt = max_insert - s->strstart;
+                if (max_insert > s->strstart) {
+                    unsigned int insert_cnt = mov_fwd;
+                    if (unlikely(insert_cnt > max_insert - s->strstart))
+                        insert_cnt = max_insert - s->strstart;
 
-                functable.insert_string(s, s->strstart + 1, insert_cnt);
+                    functable.insert_string(s, s->strstart + 1, insert_cnt);
+                }
                 s->prev_length = 0;
                 s->match_available = 0;
                 s->match_length = MIN_MATCH-1;


deflate_slow.c
```diff
@@ -106,7 +106,17 @@ ZLIB_INTERNAL block_state deflate_slow(deflate_state *s, int flush) {
...
             s->strstart++;
 #else
-            {
+            if (s->lookahead < MIN_MATCH) {
+                s->prev_length -= 2;
+                do {
+                    if (++s->strstart <= max_insert) {
+                        functable.insert_string(s, s->strstart, 1);
+                    }
+                } while (--s->prev_length != 0);
```
mtl1979: Calling insert_string inside do-while loop is very inefficient, so should be avoided. I'm not sure this is the best way to fix the issue. 
Shouldn't we calculate minimum of ```s->prev_length``` and ```max_insert - s->strstart```?
The problem in the original code was that it used ```unsigned int``` in the substraction when it should have used ```int```.

sebpop: You are correct.  See updated patch 0352a24.

deflate_slow.c
```diff
@@ -108,11 +108,13 @@ ZLIB_INTERNAL block_state deflate_slow(deflate_state *s, int flush) {
 #else
             {
                 unsigned int mov_fwd = s->prev_length - 2;
-                unsigned int insert_cnt = mov_fwd;
-                if (unlikely(insert_cnt > max_insert - s->strstart))
-                    insert_cnt = max_insert - s->strstart;
+                if (max_insert > s->strstart) {
+                    unsigned int insert_cnt = mov_fwd;
+                    if (unlikely(insert_cnt > max_insert - s->strstart))
+                        insert_cnt = max_insert - s->strstart;
```
sebpop, line 114: Note that the above 3 lines are the computation of the `min(mov_fwd, max_insert - s->strstart)`.
This is the same code as what existed before the change.

The only change is the condition under which `insert_string` is called: `if (max_insert > s->strstart)`.

Issue #210, opened 2018-09-25
undefined behavior in trees.c:482


sebpop:
```
$ ./configure --build-fuzzers
$ make check
```
fails with:
```
trees.c:482:53: runtime error: unsigned integer overflow: 6 - 7 cannot be represented in type 'unsigned int'
trees.c:482:68: runtime error: unsigned integer overflow: 4294967295 * 3 cannot be represented in type 'unsigned int'
```

mtl1979:
The second error seems result of wrap around... Number 4294967295 is FFFF FFFF in hexadecimal... In decimal that would be -1.
I'm not sure what would be the correct fix for the first error. Fixing the first error would likely also fix the second error though.

sebpop:
Right.
I have not seen many differences in trees.c between zlib.org and zlib-ng. There are a few type differences: signed to unsigned, although those differences do not impact this computation.  I will try to see if this problem exist in zlib.org as well.

sebpop:
I see the same UB in zlib.org: https://github.com/sebpop/zlib/commit/fc7e134be55c6e83517c1e4d8c975c012e127a9f
```
trees.c:556:42: runtime error: unsigned integer overflow: 6 - 7 cannot be represented in type 'unsigned long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior trees.c:556:42 in 
trees.c:556:57: runtime error: unsigned integer overflow: 18446744073709551615 * 2 cannot be represented in type 'unsigned long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior trees.c:556:57 in 
trees.c:556:28: runtime error: unsigned integer overflow: 167178 + 18446744073709551614 cannot be represented in type 'unsigned long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior trees.c:556:28 in 
```

mtl1979:
So it needs to be fixed in stock zlib first, then we can apply the patch... That file is beyond my skills...

mtl1979:
Essentially the code...
```
s->opt_len += (unsigned long)((bits - tree[m].Len) * tree[m].Freq);
```
...should be equal to...
```
s->opt_len += (unsigned long)(bits * tree[m].Freq);
s->opt_len -= (unsigned long)(tree[m].Len * tree[m].Freq);
```
...if my math deduction skills have not rusted too much during the 21 years since I left school.

Separating it as two lines avoids wrap around and overflow as side effect.

[...]

PR #211, opened 2018-09-26
fix bug #210: split statement in two to avoid subtract overflow

sebpop:


sebpop:
oss-fuzz enables checks for unsigned overflows even if it is defined behavior.

> -fsanitize=unsigned-integer-overflow checks for unsigned overflow, which is defined behavior but often causes bugs.
I'd like to try enabling -fsanitize=unsigned-integer-overflow in ubsan builds.
By default, it will not cause the targets to crash on UIO, it will only print a message once,
but will provide extra coverage signal that could be useful.

PR #221, opened 2018-11-06
Fix a bug that can crash deflate on some input when using Z_FIXED.

sebpop:
This bug was reported by Danilo Ramos of Eideticom, Inc. It has
lain in wait 13 years before being found! The bug was introduced
in zlib 1.2.2.2, with the addition of the Z_FIXED option. That
option forces the use of fixed Huffman codes. For rare inputs with
a large number of distant matches, the pending buffer into which
the compressed data is written can overwrite the distance symbol
table which it overlays. That results in corrupted output due to
invalid distances, and can result in out-of-bound accesses,
crashing the application.

The fix here combines the distance buffer and literal/length
buffers into a single symbol buffer. Now three bytes of pending
buffer space are opened up for each literal or length/distance
pair consumed, instead of the previous two bytes. This assures
that the pending buffer cannot overwrite the symbol table, since
the maximum fixed code compressed length/distance is 31 bits, and
since there are four bytes of pending space for every three bytes
of symbol space.

### DIFF
--- a/deflate.h
+++ b/deflate.h
@@ -228,7 +228,7 @@ typedef struct internal_state {
     /* Depth of each subtree used as tie breaker for trees of equal frequency
      */
 
-    unsigned char *l_buf;       /* buffer for literals or lengths */
+    unsigned char *sym_buf;       /* buffer for distances and literals/lengths */
 
     unsigned int  lit_bufsize;
     /* Size of match buffer for literals/lengths.  There are 4 reasons for
@@ -250,13 +250,8 @@ typedef struct internal_state {
      *   - I can't count above 4
      */
 
-    unsigned int last_lit;        /* running index in l_buf */
-
-    uint16_t *d_buf;
-    /* Buffer for distances. To simplify the code, d_buf and l_buf have
-     * the same number of elements. To use different lengths, an extra flag
-     * array would be necessary.
-     */
+    unsigned int sym_next;      /* running index in sym_buf */
+    unsigned int sym_end;       /* symbol table full when sym_next reaches this */
 
     unsigned long opt_len;        /* bit length of current block with optimal trees */
     unsigned long static_len;     /* bit length of current block with static trees */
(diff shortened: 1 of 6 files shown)


Dead2:
Were there any conflicts that needed fixing? Is there anything we should make sure we review/test well?

I have never used Z_FIXED myself, and I am not sure whether we currently have good tests for that.

sebpop:
There were several conflicts that I fixed.  The conflicts were due to standard types used in zlib-ng.  The files deflate_{slow,medium,fast}.c were added in zlib-ng and I impacted the change to these files as well.

We test Z_FIXED in a fuzzer by selecting the strategy based on the first byte of the input data:
https://github.com/Dead2/zlib-ng/blob/develop/test/fuzz/example_dict_fuzzer.c#L52
```
int strategy = data[0] % 5;       /* [0..4]
      #define Z_FILTERED            1
      #define Z_HUFFMAN_ONLY        2
      #define Z_RLE                 3
      #define Z_FIXED               4
      #define Z_DEFAULT_STRATEGY    0 */
``` 
Please let me know whether we need more tests than that.

Dead2:
I ran some testing of this now, and found no corruption problems at any compression level.
It does seem to slow down compression by around 1.22% average on x86_64 across all compression levels except levels 0 and 1, but I do not see any way to avoid that.

Does deflate_quick need a similar fix I wonder?

mtl1979:
I think it slows down because in the original code it writes whole contents of the variable in one instruction, but in the new code it splits it in half and shifts... This might need to be reverted/revised...

deflate.h
```diff
@@ -373,20 +368,22 @@ void ZLIB_INTERNAL bi_windup(deflate_state *s);
 
 # define _tr_tally_lit(s, c, flush) \
   { unsigned char cc = (c); \
-    s->d_buf[s->last_lit] = 0; \
-    s->l_buf[s->last_lit++] = cc; \
+    s->sym_buf[s->sym_next++] = 0; \
+    s->sym_buf[s->sym_next++] = 0; \
```
mtl1979, line 372: This is bad...

deflate.h
```diff
@@ -373,20 +368,22 @@ void ZLIB_INTERNAL bi_windup(deflate_state *s);
...
-    flush = (s->last_lit == s->lit_bufsize-1); \
+    flush = (s->sym_next == s->sym_end); \
   }
 # define _tr_tally_dist(s, distance, length, flush) \
   { unsigned char len = (unsigned char)(length); \
     uint16_t dist = (uint16_t)(distance); \
-    s->d_buf[s->last_lit] = dist; \
-    s->l_buf[s->last_lit++] = len; \
+    s->sym_buf[s->sym_next++] = dist; \
+    s->sym_buf[s->sym_next++] = dist >> 8; \
```
mtl1979, line 381: This is bad...

trees.c
```diff
@@ -929,8 +930,9 @@ void ZLIB_INTERNAL _tr_flush_block(deflate_state *s, char *buf, unsigned long st
 int ZLIB_INTERNAL _tr_tally(deflate_state *s, unsigned dist, unsigned lc) {
     /* dist: distance of matched string */
     /* lc: match length-MIN_MATCH or unmatched char (if dist==0) */
-    s->d_buf[s->last_lit] = (uint16_t)dist;
-    s->l_buf[s->last_lit++] = (unsigned char)lc;
+    s->sym_buf[s->sym_next++] = dist;
+    s->sym_buf[s->sym_next++] = dist >> 8;
```
mtl1979, line 934: This is bad...

trees.c
```diff
@@ -978,14 +958,15 @@ static void compress_block(deflate_state *s, const ct_data *ltree, const ct_data
...
     int code;           /* the code to send */
     int extra;          /* number of extra bits to send */
 
-    if (s->last_lit != 0) {
+    if (s->sym_next != 0) {
         do {
-            dist = s->d_buf[lx];
-            lc = s->l_buf[lx++];
+            dist = s->sym_buf[sx++] & 0xff;
+            dist += (unsigned)(s->sym_buf[sx++] & 0xff) << 8;
```
mtl1979, line 968: This is bad...

deflate.h
Dead2: What is bad?
I see you posted a bunch of those comments, but no info about what is bad about it and why it is bad, and what could be done to improve on it.

mtl1979: @Dead2 These lines cause unnecessary slowdown compared to the previous version of the code. I would get rid of overlays completely and use separate arrays for all buffers... This means each buffer can be as small as possible and we avoid all shifts and ors...

Dead2: So there is nothing wrong with this patch, except that it could be followed up with an optimization patch.

mtl1979: @Dead2 Wrong.

PR #223, opened 2018-11-15
Fast crc32

sebpop:


Dead2:
Is it correct that this only adds new API functions and improves code only used externally? I cannot see any references to this code from our end, unless I am missing something. I also see no impact on minigzip deflate/inflate benchmarks, so that seems to support that notion.

I currently have no way of testing this myself, but it looks good to me.

sebpop:
Correct. For example libpng relies on computing crc32.

Issue #117, opened 2017-08-26
Speed up inflate_fast by up to 1.5x on x86_64


nigeltao:
FYI I have an upstream PR to speed up inflate_fast by up to 1.5x on x86_64. Happy to discuss how you'd like to merge it into zlib-ng, if you're interested.

https://github.com/madler/zlib/pull/292

See also https://github.com/madler/zlib/pull/256 by Simon Hosie.

Dead2:
Your changes look interesting, and we'd love to see a PR from you.
If your changes are arch-generic, then they can go into the global files, if they are arch-dependent then they need to go into the arch subdirs. It should be pretty easy to figure out.

I have not looked into @simhos01 PR that you reference, but I suspect they have already been merged in our codebase. There was some collaboration in this repo where Simon's patches were suggested and improved upon, so I think the result of that is probably what is in that PR.

Your changes will not be included into the next stable release, but we are ready for the one after that. For now, just make the PR against the develop branch, I can change it to another branch when we make one ready for the next release.

nigeltao:
FYI, I will make a PR, but I'm currently waiting on discussions with other Chromium zlib experts on how to best structure the patch (in terms of e.g. file layout, variable names, macro usage) to maximize the likelihood of being accepted upstream to https://github.com/madler/zlib. I've asked madler himself (on https://github.com/madler/zlib/pull/292), but I don't think he's very responsive (which is fine, I'm sure zlib is not the only thing in his life).

mtl1979:
@nigeltao Just remember that what's accepted in madler/zlib isn't same as what's accepted here as zlib-ng has deprecated some integer type aliases due to fact that we are targeting more modern compilers.

sebpop:
All the good bits from nigeltao's patch have been integrated in zlib-ng.
Can we close this issue?

nmoinvaz:
I think this issue can be closed.

PR #224, opened 2018-11-16
bug #117: speed up inflate_fast

sebpop:
Based on a patch by Nigel Tao:
https://github.com/madler/zlib/pull/292/commits/e0ff1f330cc03ee04843f857869b4036593ab39d

This patch makes unzipping of files up to 1.2x faster on x86_64.  The other part
(1.3x speedup) of the patch by Nigel Tao is unsafe as discussed in the review of
that pull request.  zlib-ng already has a different way to optimize the memcpy
for that missing part.

The original patch was enabled only on little-endian machines. This patch adapts
the loading of 64 bits at a time to big endian machines.

### DIFF
--- a/memcopy.h
+++ b/memcopy.h
@@ -4,23 +4,15 @@
 #ifndef MEMCOPY_H_
 #define MEMCOPY_H_
 
-/* Load a short from IN and place the bytes at offset BITS in the result. */
-static inline uint32_t load_short(const unsigned char *in, unsigned bits) {
-    union {
-        uint16_t a;
-        uint8_t b[2];
-    } chunk;
-    MEMCPY(&chunk, in, sizeof(uint16_t));
+/* Load 64 bits from IN and place the bytes at offset BITS in the result. */
+static inline uint64_t load_64_bits(const unsigned char *in, unsigned bits) {
+  uint64_t chunk;
+  MEMCPY(&chunk, in, sizeof(chunk));
 
 #if BYTE_ORDER == LITTLE_ENDIAN
-    uint32_t res = chunk.a;
-    return res << bits;
+  return chunk << bits;
 #else
-    uint32_t c0 = chunk.b[0];
-    uint32_t c1 = chunk.b[1];
-    c0 <<= bits;
-    c1 <<= bits + 8;
-    return c0 + c1;
+  return ZSWAP64(chunk) << bits;
 #endif
 }
 
--- a/infback.c
+++ b/infback.c
@@ -453,7 +453,8 @@ int ZEXPORT PREFIX(inflateBack)(PREFIX3(stream) *strm, in_func in, void *in_desc
 
         case LEN:
             /* use inflate_fast() if we have enough input and output */
-            if (have >= 6 && left >= 258) {
+            if (have >= INFLATE_FAST_MIN_HAVE &&
+                left >= INFLATE_FAST_MIN_LEFT) {
                 RESTORE();
                 if (state->whave < state->wsize)
                     state->whave = state->wsize - left;
--- a/inffast.c
+++ b/inffast.c
@@ -113,8 +113,8 @@ static inline unsigned char* chunkunroll(unsigned char *out, unsigned *dist, uns
    Entry assumptions:
 
         state->mode == LEN
-        strm->avail_in >= 6
-        strm->avail_out >= 258
+        strm->avail_in >= INFLATE_FAST_MIN_HAVE
+        strm->avail_out >= INFLATE_FAST_MIN_LEFT
         start >= strm->avail_out
         state->bits < 8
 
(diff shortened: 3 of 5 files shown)


sebpop:
I will run some oss-fuzz checks on one of the PPC64 big-endian machines of the gcc farm https://cfarm.tetaneutral.net and I will report whether that testing exposed bugs.

sebpop:
I ran `make test` with msan and all other sanitizers on PPC64 without any extra fails: I locally reverted 8cc303d to be able to run the tests.

To run oss-fuzz one needs to be able to use docker containers and I don't have permissions to install or to run docker on the compiler farm.  I will install a separate libfuzzer and run the fuzzers manually on PPC64.

Dead2:
@sebpop I'd like to know what compiler output you get when C99 compilation fails, so we can decide whether it is something we can fix to adhere better to the C99 standard or whether relaxing the standard is the only choice.

memcopy.h
```diff
@@ -12,23 +12,15 @@
...
-    uint32_t res = chunk.a;
-    return res << bits;
+    return chunk << bits;
 #else
-    uint32_t c0 = chunk.b[0];
-    uint32_t c1 = chunk.b[1];
-    c0 <<= bits;
-    c1 <<= bits + 8;
-    return c0 + c1;
+    return __builtin_bswap64(chunk) << bits;
```
Dead2: Is this actually portable? I suspect we might need to provide a __builtin_bswap64 function as a fallback where the compiler does not provide one.

Dead2:
Benchmark runs: 7, tested levels: 0-7, testfile 100M

develop at 796ad10efb749d264fe262b624842a921f05e45c with -O3:
 ```
Level  Comp   Comptime min/avg/max  Decomptime min/avg/max
 0     100.02%       0.01/0.01/0.02          0.08/0.09/0.11 
 1      47.08%       0.49/0.50/0.51          0.37/0.39/0.40 
 2      36.02%       1.10/1.12/1.13          0.39/0.39/0.40 
 3      34.77%       1.32/1.34/1.37          0.38/0.38/0.38 
 4      33.41%       1.50/1.53/1.56          0.37/0.37/0.38 
 5      33.07%       1.85/1.87/1.90          0.36/0.37/0.38 
 6      32.83%       2.54/2.57/2.61          0.36/0.37/0.38 
 avg    45.31%                 1.28                   0.34
 tot                          62.60                  16.58
```

PR224 with -O3:
 ```
Level  Comp   Comptime min/avg/max  Decomptime min/avg/max
 0     100.02%       0.01/0.01/0.02          0.09/0.09/0.10 
 1      47.08%       0.49/0.50/0.51          0.37/0.37/0.38 
 2      36.02%       1.09/1.11/1.13          0.38/0.38/0.39 
 3      34.77%       1.32/1.34/1.38          0.35/0.36/0.38 
 4      33.41%       1.49/1.52/1.54          0.36/0.36/0.37 
 5      33.07%       1.85/1.88/1.93          0.35/0.36/0.37 
 6      32.83%       2.55/2.58/2.65          0.35/0.35/0.36 
 avg    45.31%                 1.28                   0.33
 tot                          62.48                  16.02
```

So I see about a 5.4% speedup on my x86_64 machine, not quite the 1.2x speedup but a nice speedup nevertheless. This benchmark measures the total execution time of minigzip, so that might have caused some inefficiencies.
At -O2, I only see a 2.7% speedup.

memcopy.h
sebpop: I will prepare a patch to detect at configure/cmake time compilers without __builtin_bswap64 and provide an implementation for those.

sebpop:
> @sebpop I'd like to know what compiler output you get when C99 compilation fails, so we can decide whether it is something we can fix to adhere better to the C99 standard or whether relaxing the standard is the only choice.

The compiler is happy to compile the code without warnings, however it silently miscompiles minigzip, see the `make test` output on ppc64-linux here: https://github.com/zlib-ng/zlib-ng/commit/8cc303d05634a2169079a858b902ade256431fd5#commitcomment-31333964

sebpop:
Thanks @Dead2 for the perf analysis.
I added the perf analysis notes to the commit message.

memcopy.h
sebpop: Fixed in 13cd6d9.

memcopy.h
```diff
@@ -18,9 +18,19 @@ static inline uint64_t load_64_bits(const unsigned char *in, unsigned bits) {
   MEMCPY(&chunk, in, sizeof(chunk));
 
 #if BYTE_ORDER == LITTLE_ENDIAN
-    return chunk << bits;
+  return chunk << bits;
+#elif defined(HAVE_BUILTIN_BSWAP64)
+  return __builtin_bswap64(chunk) << bits;
 #else
-    return __builtin_bswap64(chunk) << bits;
+  chunk = ((chunk & 0xFF00000000000000u) >> 56u) |
```
Dead2: For clarity and easier re-use of code, perhaps we should make this a macro instead? I don't have any personal experience with coding for big endian systems, so is it likely that we will need this elsewhere at some point?
If it is not likely to be needed elsewhere, I'd still prefer to have this as a macro that checks for __builtin_bswap64 and provides it if it is not provided by the compiler. Hmm, that would also avoid the whole need for checks in cmake/configure I think. Thoughts?

mtl1979: When I did the patch for big-endian PPC systems, I had all byteswap macros grouped in one header file. This made it pretty easy to cleanup some places where data was stored byte by byte instead of byte swapping and then using single store for each variable or struct member.

sebpop: I found out that zutil.h already had a `ZSWAP32` macro.  I amended the previous patch by removing the configure/cmake detection and providing the `ZSWAP64` macro defined under the same logic as `ZSWAP32`.

sebpop: I separately pushed the ZSWAP64 change in #232 as we will also need a ZSWAP16.

Dead2:
@sebpop This needs a little rebase, since the ZSWAP-PR was merged :)

PR #227, opened 2018-12-06
Faster compression time and smaller compressed files

sebpop:


mtl1979:
We actually got better compression when checking for hash map collisions with some files, so I would rather see these as two separate pull requests so we can benchmark them individually.

sebpop:
Separately pushed that change at #228

Dead2:
Current develop @ edb9fb1e3acc4ff82845518fd50cf51185c47ad7:
```
 Level  Comp   Comptime min/avg/max  Decomptime min/avg/max
 0     100.02%       0.01/0.01/0.02          0.08/0.09/0.10 
 1      47.08%       0.49/0.51/0.52          0.37/0.38/0.39 
 2      36.02%       1.14/1.16/1.20          0.39/0.39/0.40 
 3      34.77%       1.36/1.38/1.40          0.38/0.38/0.39 
 4      33.41%       1.51/1.56/1.60          0.36/0.37/0.38 
 5      33.07%       1.87/1.90/1.92          0.36/0.37/0.38 
 6      32.83%       2.56/2.59/2.62          0.36/0.37/0.38 
 avg    45.31%                 1.30                   0.34
 tot                          63.69                  16.54
```

This PR:
```
Level  Comp   Comptime min/avg/max  Decomptime min/avg/max
 0     100.02%       0.07/0.07/0.08          0.12/0.13/0.13 
 1      47.08%       1.54/1.56/1.59          1.09/1.10/1.11 
 2      36.92%       3.12/3.14/3.18          1.19/1.20/1.21 
 3      34.95%       4.29/4.33/4.39          1.10/1.13/1.16 
 4      34.27%       4.79/4.82/4.87          1.10/1.10/1.11 
 5      33.81%       5.56/5.61/5.75          1.07/1.09/1.12 
 6      33.41%       7.24/7.31/7.41          1.06/1.07/1.09 
 avg    45.78%                 3.83                   0.97
 tot                         187.89                  47.74
```

As you can see, this PR is a lot slower than the current code. And it also compresses worse with my testdata. I even re-ran the benchmark to make sure it was not my mistake.
There is also a whole lot going on in this PR, and I am at a loss as to why most of these changes would be wanted in the first place.

sebpop:
You are right, search algorithms may be better/worse on different data sets: no free lunch.
As this is a pretty intrusive change, I'm abandoning it.

PR #229, opened 2018-12-07
remove `unaligned store` UBsan warnings

sebpop:
This patch addresses several warnings from `make test` when
zlib-ng was configured -with-fuzzers -with-sanitizers:
```
zlib-ng/trees.c:798:5: runtime error: store to misaligned address 0x63100125c801 for type 'uint16_t' (aka 'unsigned short'), which requires 2 byte alignment
0x63100125c801: note: pointer points here
 00 80 76  01 8b 08 00 00 00 00 00  00 03 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00
              ^
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior zlib-ng/trees.c:798:5 in
zlib-ng/trees.c:799:5: runtime error: store to misaligned address 0x63100125c803 for type 'uint16_t' (aka 'unsigned short'), which requires 2 byte alignment
0x63100125c803: note: pointer points here
 76  01 f5 08 00 00 00 00 00  00 03 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00
              ^
```
Instead of using `*(uint16_t*) foo = bar` to write a uint16_t, call
__builtin_memcpy which will be safe in case of memory page boundaries.

Without the patch:
```
 Performance counter stats for './minigzip -9 llvm.tar':

      13384.872467      task-clock (msec)         #    1.000 CPUs utilized
                16      context-switches          #    0.001 K/sec
                 0      cpu-migrations            #    0.000 K/sec
               130      page-faults               #    0.010 K/sec
    58,761,512,991      cycles                    #    4.390 GHz
   <not supported>      stalled-cycles-frontend
   <not supported>      stalled-cycles-backend
    77,600,648,174      instructions              #    1.32  insns per cycle
    17,486,109,042      branches                  # 1306.408 M/sec
       197,490,626      branch-misses             #    1.13% of all branches

      13.385829390 seconds time elapsed

45408 -rw-rw-r-- 1 spop spop 46493896 Dec  7 12:09 llvm.tar.gz
```
With remove-unaligned-stores patch:
```
 Performance counter stats for './minigzip -9 llvm.tar':

      13466.556633      task-clock (msec)         #    1.000 CPUs utilized
                34      context-switches          #    0.003 K/sec
                 0      cpu-migrations            #    0.000 K/sec
               127      page-faults               #    0.009 K/sec
    58,904,472,339      cycles                    #    4.374 GHz
   <not supported>      stalled-cycles-frontend
   <not supported>      stalled-cycles-backend
    77,488,290,366      instructions              #    1.32  insns per cycle
    17,479,687,094      branches                  # 1298.007 M/sec
       197,853,485      branch-misses             #    1.13% of all branches

      13.468135221 seconds time elapsed

45408 -rw-rw-r-- 1 spop spop 46493896 Dec  7 12:08 llvm.tar.gz
```

### DIFF
--- a/deflate.h
+++ b/deflate.h
@@ -13,6 +13,7 @@
 /* @(#) $Id$ */
 
 #include "zutil.h"
+#include "gzendian.h"
 
 /* define NO_GZIP when compiling if you want to disable gzip header and
    trailer creation by deflate().  NO_GZIP would be used to avoid linking in
@@ -302,26 +303,13 @@ typedef enum {
  * Output a short LSB first on the stream.
  * IN assertion: there is enough room in pendingBuf.
  */
-#ifdef UNALIGNED_OK
-/* Compared to the else-clause's implementation, there are few advantages:
- *  - s->pending is loaded only once (else-clause's implementation needs to
- *    load s->pending twice due to the alias between s->pending and
- *    s->pending_buf[].
- *  - no instructions for extracting bytes from short.
- *  - needs less registers
- *  - stores to adjacent bytes are merged into a single store, albeit at the
- *    cost of penalty of potentially unaligned access. 
- */
-#define put_short(s, w) { \
-    *(uint16_t*)(&s->pending_buf[s->pending]) = (w) ; \
-    s->pending += 2; \
-}
-#else
-#define put_short(s, w) { \
-    put_byte(s, (unsigned char)((w) & 0xff)); \
-    put_byte(s, (unsigned char)((uint16_t)(w) >> 8)); \
-}
+static inline void put_short(deflate_state *s, uint16_t w) {
+#if BYTE_ORDER == BIG_ENDIAN
+  w = ZSWAP16(w);
 #endif
+  MEMCPY(&(s->pending_buf[s->pending]), &w, sizeof(uint16_t));
+  s->pending += 2;
+}
 
 #define MIN_LOOKAHEAD (MAX_MATCH+MIN_MATCH+1)
 /* Minimum amount of lookahead, except at the end of the input file.


Dead2:
Since this code was ifdef'ed with UNALIGNED_OK, what does this change in that regard?

I might be wrong, but it looks to me like we exploit the fact that the cpu handles unaligned stores fine, but the sanitizer complains about it because it breaks C standards and thus would break on platforms where we do not set UNALIGNED_OK, is that about right?

Also, would this new memcpy be able to replace the one in the fallback part of the code as well? And if so, is it fast enough that we can drop the whole UNALIGNED_OK code-path? I see that it is slightly slower than the unaligned store, but that could be worth it if we can rid the code of the special-handling.

mtl1979:
@Dead2 I think it assumes all systems where unaligned stores are allowed are little-endian, and the ones that don't allow unaligned stores are big-endian, so if the second part also uses MEMCPY(), the input variable needs to be byteswapped if the system is big-endian.

Dead2:
Hmm, sounds like we need a big-endian put_short too then, with a separate check or at least a comment to clarify its use.
But for little-endian, does it make sense to keep two separate ways to do this at all?

mtl1979:
@Dead2 I don't think one can really optimize more than memcpy already is, for basic types it is inlined and as efficient as simple C-style assignment. Temporary variables usually get optimized away, so there will be no penalty using them for readability unless optimizations are disabled.

Dead2:
This benchmarks to around 0.4-0.5% slower than current develop branch does with the unaligned store.
Since this cleans up the code a lot, I am fine with that performance regression.

I'd like to see a few yay/nay for this asap. @mtl1979 @sebpop and any others following this discussion.

mtl1979:
@Dead2 I'm OK with little performance regression as we can still improve the code where two put_byte calls are used instead of one put_short call.

Dead2:
@sebpop A little rebase would be nice here too, some of the commits are already in develop.

PR #230, opened 2018-12-07
fix bug #210: split statement in two to avoid substract overflow

sebpop:
make check used to fail with:
trees.c:482:53: runtime error: unsigned integer overflow: 6 - 7 cannot be represented in type 'unsigned int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior trees.c:482:53 in

Patch from Mika Lindqvist.

### DIFF
--- a/trees.c
+++ b/trees.c
@@ -479,7 +479,8 @@ static void gen_bitlen(deflate_state *s, tree_desc *desc) {
                 continue;
             if (tree[m].Len != bits) {
                 Trace((stderr, "code %d bits %d->%u\n", m, tree[m].Len, bits));
-                s->opt_len += (unsigned long)((bits - tree[m].Len) * tree[m].Freq);
+                s->opt_len += (unsigned long)(bits * tree[m].Freq);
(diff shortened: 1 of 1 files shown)


Dead2:
Looks good to me

Dead2:
I did a benchmark of this to make sure it does not cause a big regression. Seems like this actually makes the code easier to optimize for GCC 8.2.1, because my results on x86_64 shows an overall speedup of 0.4 to 0.8%.
I did not believe the results at first, so I re-compiled both versions and ran the benchmarks again, same results.
(level 0-7 tested 7 times each, using averages to compare, turbo and idlestates disabled, source file in ram, destination file /dev/null, minigzip given realtime priority).

Dead2:
I did not notice until now that this actually increases compression ratio with my 100M testfile.
Levels 0 and 1 see no change.
Levels 2-6 on average go from 34.02% to 33.95% (% of original file size).
Level 6 goes from 32.83% to 32.77%.

Not a huge change, but it looks like this bug has been making a mess.

PR #232, opened 2018-12-11
define ZSWAP64 and ZSWAP16

sebpop:
Add definition of ZSWAP64 and ZSWAP16 in the same place as ZSWAP32.


Dead2:
Really nice :)
Looks good to me.

PR #238, opened 2018-12-12
Correct the initialization requirements for deflateInit2().

sebpop:


Dead2:
Should probably be copied to zlib-ng.h as well

PR #243, opened 2018-12-12
Avoid the use of ptrdiff_t.

sebpop:
This isn't the right type anyway to assure that it contains a
pointer. That type would be intptr_t or uintptr_t. However the C99
standard says that those types are optional, so their use would not
be portable. This commit simply uses size_t or whatever configure
decided to use for size_t. That would be the same length as
ptrdiff_t, and so will work just as well. The code checks to see if
the length of the type used is the same as the length of a void
pointer, so there is already protection against the use of the
wrong type. The use of size_t (or ptrdiff_t) will almost always
work, as all modern architectures have an array size that is the
same as the pointer size. Only old segmented architectures would
have to fall back to the slower CRC-32 calculation, where the
amount of memory that can be accessed is larger than the maximum
array size.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -330,6 +330,29 @@ if(HAVE_BUILTIN_CTZL)
     add_definitions(-DHAVE_BUILTIN_CTZL)
 endif()
 
+#
+# check for ptrdiff_t support
+#
+check_c_source_compiles(
+    "#include <stddef.h>
+     int main() { ptrdiff_t *a; return 0; }"
+    HAVE_PTRDIFF_T
+)
+if(NOT HAVE_PTRDIFF_T)
+  set(NEED_PTRDIFF_T 1)
+
+  check_type_size("void *" SIZEOF_DATA_PTR)
+  message(STATUS "sizeof(void *) is ${SIZEOF_DATA_PTR} bytes")
+
+  if(${SIZEOF_DATA_PTR} MATCHES "4")
+    set(PTRDIFF_TYPE "uint32_t")
+  elseif(${SIZEOF_DATA_PTR} MATCHES "8")
+    set(PTRDIFF_TYPE "uint64_t")
+  else()
+    message(FATAL_ERROR "sizeof(void *) is neither 32 nor 64 bit")
+  endif()
+endif()
+
 # Macro to check if source compiles when cross-compiling
 # or runs when compiling natively
 macro(check_c_source_compile_or_run source flag)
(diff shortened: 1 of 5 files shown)


mtl1979:
size_t won't work on non-GNU compilers like Visual C++ or gcc in C99 mode. size_t can be just 32 bits using those compilers even though pointer is 64 bits.

Dead2:
I am thinking we might rather define our own intptr_t/uintptr_t/ptrdiff_t if those are not already defined by the compiler.
That way we use the modern names for clarity, and have a fallback for compilers/platforms where it is needed.
Of course that requires us to have some detection macros that figures our what type is the correct size and uses that.

sebpop:
> define our own intptr_t/uintptr_t/ptrdiff_t if those are not already defined by the compiler.

This is not easy. One would need to know at configure time the size of a `void *` pointer and use that to decide whether to `typedef uint64_t ptrdiff_t;` or `typedef uint32_t ptrdiff_t;`.  The compiler already knows the `sizeof(void *)` as it needs to output the layout of static structures.  For instance autoconf does something like this:
```
struct s {
  int check_PTRDIFF:
      PTRDIFF_MIN == TYPE_MINIMUM (ptrdiff_t)
      && PTRDIFF_MAX == TYPE_MAXIMUM (ptrdiff_t)
      ? 1 : -1;
```
as we can see it generated in `binutils-gdb/gdb/gnulib/configure`

mtl1979:
It's pretty easy to detect size of pointer during configure time... We already do it in some places in the configure script to decide which optimizations to enable.

sebpop:
We can do a static check in C as the one from autoconf as shown above in binutils-gdb:
```
#define COMPILE_TIME_ASSERT(pred)            \
struct s { int check_PTRDIFF: (pred) ? 1 : -1; };
```
a struct field declared to have a length of 1 will pass compile, and will error when the field size is -1.

The other way to do a static check is as explained in https://stackoverflow.com/a/9284679/9252818
```
#define COMPILE_TIME_ASSERT(pred)            \  
    switch(0){case 0:case pred:;}
```
If pred evaluates to 0, like a false boolean expression does in C, the compiler will throw an error.

We can use either of those from configure trying to compile:
```
void fun () {
  COMPILE_TIME_ASSERT(sizeof(void *) == sizeof(uint64_t));
}
```
I will push a patch that uses a static assert to check for pointer size.

Dead2:
Sorry, needs a rebase

Dead2:
What about cmake?

Dead2:
It is looking really good now, I just have one tiny change to request. That cmake prints the test status, like we do for other header files we check for. Ex:
-- Looking for stddef.h
-- Looking for stddef.h - found

We should probably keep doing that for consistency, and I like the idea of providing some information where workarounds get enabled.

sebpop:
I am seeing the following lines in the output of `cmake ..`:
```
-- Performing Test HAVE_PTRDIFF_T
-- Performing Test HAVE_PTRDIFF_T - Success
```
Is that enough?

Dead2:
Yes, that is fine. I had a thinko and thought we were looking for a header file and not a data type, so I didn't even notice that one.

PR #246, opened 2018-12-17
cleanup x86/crc_pclmulqdq

sebpop:
configure --with-msan; make test fails on x86_64 with:

Uninitialized bytes in __interceptor_fwrite at offset 14 inside [0x72e000000000, 49152)
==706==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x498eba in zng_gzclose /home/spop/s/zlib-ng/test/minigzip.c:240:13
    #1 0x499982 in gz_compress /home/spop/s/zlib-ng/test/minigzip.c:313:9
    #2 0x49c352 in main /home/spop/s/zlib-ng/test/minigzip.c:514:13
    #3 0x7f8b954ae82f in __libc_start_main /build/glibc-Cl5G7W/glibc-2.23/csu/../csu/libc-start.c:291
    #4 0x41a9c8 in _start (/work/spop/zlib-ng/minigzip+0x41a9c8)

SUMMARY: MemorySanitizer: use-of-uninitialized-value /home/spop/s/zlib-ng/test/minigzip.c:240:13 in zng_gzclose

I think this is a bug in the memory sanitizer that may be confused by the use of
intrinsics in crc_fold_init. By moving the initialization of strm->adler out of
the crc_reset function, the memory sanitizer stops warning.


mtl1979:
If I remember correctly, there was discussion a while ago about fixing a bug in our usage of CRC_SAVE() macro. Essentially all code after the macro is ignored (as unreachable code) by normal compilers. The person who originally wrote the code thought no-one cares if compiler ignores the code as it saves him from writing two different variations of same macro,

Dead2:
Having that much code duplication in deflate seems like a bad solution to the problem, that code should rather go into crc_reset, at least I see that as the logical place for it to live.
@mtl1979's theory would be easily tested by moving the intialization to the top of crc_fold_init.

The do starting in CRC_LOAD and ending in CRC_SAVE seems unnecessarily confusing though, It would be better to have that outside the macros IMO. Also, is CRC_LOAD really needed in crc_fold_init, except to provide the do{ part?

sebpop:
Moving the intialization `s->strm->adler = 0;` to the top of crc_fold_init does not fix the msan problem.

sebpop:
I cleaned up that code by removing the CRC_{LOAD,SAVE} macros, and I moved the crc_reset functions to deflate.c to make them inline-able. MSan then stops warning about the uninitialized variable.

sebpop:
Let's commit this patch as a cleanup patch, and not one that fixes msan warnings.

There may be a configure/make problem as I see several files under arch/x86 or arch/aarch64 that do not get compiled with -fsanitize=memory, which confuse the compiler as it does not see some of the memory markers in place.

The following command passes without fails even without this current patch:
```
$ CC=clang CFLAGS="-g -fsanitize=memory" LDFLAGS="-g -fsanitize=memory" ./configure --with-fuzzers --with-msan ; make clean && make -j; make test
```

sebpop:
#279 fixes the MSan warning.  I will resubmit this patch as a cleanup.

Dead2:
Ok, it looks like this PR should be split into two commits at least.

1. I think the macro cleanup would be better if it just didn't have the do{ }while stuff in the macro. Having the actual intrinsics inside the macros does not seem like a problem to me, in fact it makes the code easier to read. Unless you have a reason for using the intrinsics directly?

2. The crc_pclmulqdq.c removal is going in the wrong direction IMO. We'd like to keep the size of deflate.c down and keep code separated in files that hint at the contents.

How about we roll crc_finalize() and copy_with_crc() into crc32.c, where they probably should have been all along? There is really no need for that part of the code to be split out into a separate file for pclmulqdq. And then we could put crc_reset() into crc_reset_p.h, that gets included and inlined in deflate.c. A comment in that header file should also state why crc_reset() is being treated special, since it is not for performance reasons.

3. copy_with_crc() could also get the inline keyword if that is likely to have any impact on performance, but perhaps in a separate commit.


Does this sound like a good plan? I am open for discussion on these points.

sebpop:
I am closing this pull request and follow up with patches as needed.
To address point 1. of the cleanup, please see #291.

For point 2. "roll crc_finalize() and copy_with_crc() into crc32.c", please see #292.

For point 3, https://github.com/sebpop/zlib-ng/commit/6a82743dd4c474995f5c9659686da746a840d1d4 shows no significant performance difference on x86_64 when inlining `copy_with_crc()`.

PR #247, opened 2018-12-17
fix several memory sanitizer errors on aarch64

sebpop:
With this patch we have a clean run of make test on aarch64 when zlib-ng is
configured with --with-msan --with-fuzzers.


Dead2:
Looks good to me

PR #248, opened 2018-12-17
Don't compute check value for raw inflate if asked to validate.

sebpop:


Dead2:
That merge is wrong, it adds an if instead of replacing it.

Dead2:
Looks good to me

PR #249, opened 2018-12-18
Implement scan-build test inside travis-CI

gdevenyi:
Also upgrades travis to Xenial to get access to LLVM natively in travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: c
 cache: ccache
-dist: trusty
+dist: xenial
 
 matrix:
   include:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ language: c
 cache: ccache
 dist: xenial
 
+env:
+  global:
+    - MAKE="make"
+
 matrix:
   include:
     - os: linux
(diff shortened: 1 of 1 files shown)


Dead2:
I see you are still putting the MAKE="make" at the front instead of last in each line. I think it is much more logical to have it at the end, after configure, since that is after all the order that it gets executed in.

gdevenyi:
Makes sense, will refactor

gdevenyi:
Refactored to put the make command override at the end of line.

gdevenyi:
Any other comments here? Is this a reasonable change?

mtl1979:
I'm not sure it's the best way to define MAKE="make" separately in each target, it should be set as global default and overridden only when necessary...

gdevenyi:
Okay. I can refactor again.

gdevenyi:
Refactor complete.

Dead2:
Looks good to me.
The current travis build shows no bug reports, so what happens if there is a bug?

gdevenyi:
Travis will throw an error on build, and you'll get lines like this:
https://github.com/zlib-ng/zlib-ng/pull/188#issue-208722186

Dead2:
Tanks :)

PR #250, opened 2018-12-18
Return an error if the gzputs string length can't fit in an int.

sebpop:
Tested with `configure --with-gzfileops`.


gzwrite.c
```diff
@@ -335,8 +334,12 @@ int ZEXPORT PREFIX(gzputs)(gzFile file, const char *s) {
 
     /* write string */
     len = strlen(s);
-    ret = (int)gz_write(state, s, len);
-    return ret == 0 && len != 0 ? -1 : ret;
+    if ((int)len < 0 || (unsigned)len != len) {
```
mtl1979, line 337: Easier than cast is to just compare ```len > INT_MAX``` as that macro should be defined in all modern compilers.

PR #251, opened 2018-12-18
Small speedup to inflate [psumbera].

sebpop:
Seeing a few percent speedup by using a pointer instead of an
assigned structure. This seems to help the compiler to optimize
better.


Dead2:
Seeing an average of 1.8% speedup on inflate w/gcc 8.2 on x86_64

PR #252, opened 2018-12-18
Avoid some conversion warnings in gzread.c and gzwrite.c.

sebpop:


gzread.c
```diff
@@ -285,7 +285,7 @@ static size_t gz_read(gz_state *state, void *buf, size_t len) {
     got = 0;
     do {
         /* set n to the maximum amount of len that fits in an unsigned int */
-        n = -1;
+        n = (unsigned)-1;
```
mtl1979, line 288: Shouldn't this be ```n = UINT_MAX;```.

Dead2: Perhaps, but that should probably be a separate commit, since it that deviates from what madler did.

PR #254, opened 2018-12-18
Avoid a conversion error in gzseek when off_t type too small.

sebpop:
This is a problem in the odd case that the second argument of
LSEEK is a larger type than off_t. Apparently MinGW defines off_t
to be 32 bits, but _lseeki64 has a 64-bit second argument.

Also undo a previous commit to permit MinGW to use _lseeki64.


Dead2:
No need to add the linebreak

PR #255, opened 2018-12-18
Fix CLEAR_HASH macro to be usable as a single statement.

sebpop:
As it is used in deflateParams().


Dead2:
Do we want that extra linebreak there?
Zlib-ng is using 120 wide lines as a guideline.

PR #257, opened 2018-12-18
Limit hash table inserts after switch from stored deflate.

sebpop:
This limits hash table inserts to the available data in the window
and to the sliding window size in deflate_stored(). The hash table
inserts are deferred until deflateParams() switches to a non-zero
compression level.


Dead2:
I see no regressions with minigzip use from this, so this looks good to me.

PR #258, opened 2018-12-18
Permit a deflateParams() parameter change as soon as possible.

sebpop:
This commit allows a parameter change even if the input data has
not all been compressed and copied to the application output
buffer, so long as all of the input data has been compressed to
the internal pending output buffer. This also allows an immediate
deflateParams change so long as there have been no deflate calls
since initialization or reset.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -509,7 +509,7 @@ int ZEXPORT PREFIX(deflateResetKeep)(PREFIX3(stream) *strm) {
     else
 #endif
         strm->adler = functable.adler32(0L, NULL, 0);
-    s->last_flush = Z_NO_FLUSH;
+    s->last_flush = -2;
 
     _tr_init(s);
 
@@ -585,12 +585,13 @@ int ZEXPORT PREFIX(deflateParams)(PREFIX3(stream) *strm, int level, int strategy
     }
     func = configuration_table[s->level].func;
 
-    if ((strategy != s->strategy || func != configuration_table[level].func) && s->high_water) {
+    if ((strategy != s->strategy || func != configuration_table[level].func) &&
+        s->last_flush != -2) {
         /* Flush the last buffer: */
         int err = PREFIX(deflate)(strm, Z_BLOCK);
         if (err == Z_STREAM_ERROR)
             return err;
-        if (strm->avail_out == 0)
+        if (strm->avail_in || (s->strstart - s->block_start) + s->lookahead)
             return Z_BUF_ERROR;
     }
     if (s->level != level) {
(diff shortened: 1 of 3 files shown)


Dead2:
This commit uses tabs instead of spaces.

deflate.c
```diff
@@ -585,12 +585,13 @@ int ZEXPORT PREFIX(deflateParams)(PREFIX3(stream) *strm, int level, int strategy
...
 
-    if ((strategy != s->strategy || func != configuration_table[level].func) && s->high_water) {
+    if ((strategy != s->strategy || func != configuration_table[level].func) &&
+        s->last_flush != -2) {
         /* Flush the last buffer: */
         int err = PREFIX(deflate)(strm, Z_BLOCK);
         if (err == Z_STREAM_ERROR)
             return err;
-        if (strm->avail_out == 0)
+        if (strm->avail_in || (s->strstart - s->block_start) + s->lookahead)
```
mtl1979, line 594: This statement does not make sense at all... something is missing or in wrong place.

sebpop: The change comes from this patch: https://github.com/madler/zlib/commit/f9694097dd69354b03cb8af959094c7f260db0a1

mtl1979: I've never liked Mark Adler's way of coding and even with 35 years experience in coding, it's almost impossible for me to decipher that line.

sebpop: My understanding of that code is that changing parameters of deflate is not allowed when not all the data has been compressed with the old parameters: i.e., strm->avail_in non zero. 

The call to deflate() with Z_BLOCK a few lines above ensures that deflate has completed the emission of a block. If deflate_state does not reflect that property `(s->strstart - s->block_start) + s->lookahead != 0` it is also an error to change parameters in the middle of emitting a block.

I think this is also what the documentation says in zlib[-ng].h.

mtl1979: Implicit cast of integer expression to bool is always a bad idea for readability.

sebpop: I agree that it is hard to read.

When I did the merges I tried to minimize the changes needed to apply each patch: there are a few coding styles that zlib-ng has adopted and I tried to follow these.  If you are saying that zlib-ng code should have `if (expr != 0)` instead of just `if (expr)`, let's document that somewhere and then I will amend the patch.

My preference is to have minimal changes in coding style with respect to zlib.org in order to minimize the effort of merges in the future.  If zlib-ng decides to go for a different style, I will follow the guidelines.

mtl1979: @sebpop Maybe @Dead2 wants as project leader to contribute his opinion on the matter too... So I'm putting the guideline on vote now before deciding if we should merge the patch as is or not...

Dead2: I think we should keep the common code as similar as possible, for ease of merging changes from upstream.
But I agree that in our own code, we should go for the more verbose but easier to read coding-style.

PR #261, opened 2018-12-19
Chromium zlib benchmark

sebpop:


Dead2:
I am a little unsure whether we want this inside the zlib-ng repo or not, and I am leaning towards not.
What about we make a new repo under the zlib-ng organization? As zlib-ng/chromium_bench or something? If so, I can give you commit rights there directly and avoid cluttering the main repo.
I'd like to prevent a name like zlib-ng_bench or similar, since it would make it difficult to separate different benches from eachother, so some kind of unique name would be nice.

I am working on cleaning up my own python script that does performance testing using minigzip so it can be made public in its own repo.

Also, I am thinking about making a repo that can contain patches for 3rd-party programs to allow them to detect and use zlib-ng using it's own API instead of the zlib-compat one. I currently have one such hackish patch for nginx that is not suitable for upstreaming (since it replaces zlib support instead of adding zlib-ng separately).

mtl1979:
We should always avoid code that uses trademark (registered or not) of other project/company in
executable names or visible text, except when it is merely mention in copyright text.

sebpop:
I am abandoning this patch as the library repo is not the right place for the benchmark.

Issue #263, opened 2018-12-20
[x86_64] deflate of zlib-ng is 20 to 30% slower than zlib.org on jpeg and pdf when compiled with clang-6.0


sebpop:
Running the chromium benchmark on snappy data sets (as setup in #261) shows that zlib-ng deflate speed is 20 to 30% slower than zlib.org.  The data below comes from an x86_64-linux i7-4790K at 4.4GHz, both libraries compiled with clang at `-O3`:
```
$ clang --version
clang version 6.0.0-1ubuntu2~16.04.1 (tags/RELEASE_600/final)
```

```
zlib.org : fireworks.jpeg  GZIP -2: [b 1M] bytes 123093 -> 122931 99.9% comp  56.7 ( 56.7) MB/s uncomp 658.1 (658.3) MB/s
zlib-ng  : fireworks.jpeg  GZIP -2: [b 1M] bytes 123093 -> 122935 99.9% comp  42.0 ( 42.6) MB/s uncomp 753.6 (753.8) MB/s
compression rate : zlib.org 99.9%, zlib-ng 99.9%: diff 0% 
deflate speed    : zlib.org 56.7 MB/s, zlib-ng 42.0 MB/s: diff -14.7 MB/s (-25.9%)
inflate speed    : zlib.org 658.1 MB/s, zlib-ng 753.6 MB/s: diff 95.5 MB/s (14.5%)

zlib.org : fireworks.jpeg  GZIP -3: [b 1M] bytes 123093 -> 122931 99.9% comp  55.8 ( 56.2) MB/s uncomp 657.7 (657.9) MB/s
zlib-ng  : fireworks.jpeg  GZIP -3: [b 1M] bytes 123093 -> 122930 99.9% comp  42.7 ( 42.8) MB/s uncomp 753.4 (753.5) MB/s
compression rate : zlib.org 99.9%, zlib-ng 99.9%: diff 0% 
deflate speed    : zlib.org 55.8 MB/s, zlib-ng 42.7 MB/s: diff -13.1 MB/s (-23.4%)
inflate speed    : zlib.org 657.7 MB/s, zlib-ng 753.4 MB/s: diff 95.7 MB/s (14.5%)

zlib.org : fireworks.jpeg  GZIP -4: [b 1M] bytes 123093 -> 122930 99.9% comp  54.3 ( 54.4) MB/s uncomp 662.8 (663.0) MB/s
[...]

PR #264, opened 2018-12-20
fix bug #263: hoist invariant loads

sebpop:
This patch speeds up longest_match for x86_64 to match the performance of zlib.org.

Performance results on x86_64 in commit message of:
https://github.com/zlib-ng/zlib-ng/commit/774e028f4d34b7fceb3e8b045ebc44262f31069a

### DIFF
--- a/match.c
+++ b/match.c
@@ -347,21 +347,23 @@ ZLIB_INTERNAL unsigned longest_match(deflate_state *const s, IPos cur_match) {
  */
 
 ZLIB_INTERNAL unsigned longest_match(deflate_state *const s, IPos cur_match) {
+    unsigned int strstart = s->strstart;
     unsigned chain_length = s->max_chain_length;/* max hash chain length */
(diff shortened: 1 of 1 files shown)


Dead2:
I really think this commit should mention why and what compiler(s) need this.

The code itself looks good to me, have not tested yet.

mtl1979:
I see s->window_size used in multiple places, it might speed up the code even more if it is stored as local variable, so it doesn't need indirect load either... That should allow accessing end pointer of the sliding window using offset load directly.

sebpop:
>  s->window_size used in multiple places

There are several such fields that are loaded again and again.  I'm not sure it is beneficial to systematically hoist all those expressions out of the loop: that may create a lot of spills and reloads as we run out of registers. From that point of view the patch is minimal, i.e., it only hoists two field accesses, which is enough to get the performance up on x86_64.

Dead2:
This shows an average of 0.53% speedup on compression in my tests using gcc8.2 on x86_64

PR #267, opened 2019-01-02
on x86_64 call libc's memset

sebpop:
On x86_64-linux by just calling libc's memset, we get faster inflate over all
the chromium benchmarks with up to 10% speedup.
https://gist.github.com/sebpop/326c5120b6fa6cffc0e5e2017e907383

On aarch64-linux calling libc's memset ends up being just regressions:
https://gist.github.com/sebpop/0260bf9f76b568f5b4763b5d5562790b

### DIFF
--- a/memcopy.h
+++ b/memcopy.h
@@ -266,6 +266,11 @@ static inline unsigned char *chunk_memcpy(unsigned char *out, unsigned char *fro
     return out;
 }
 
+#undef CALL_MEMSET
+#if defined(__x86_64) || defined(__i386_)
+# define CALL_MEMSET 1
+#endif
+
 /* Memset LEN bytes in OUT with the value at OUT - 1. Return OUT + LEN. */
 static inline unsigned char *byte_memset(unsigned char *out, unsigned len) {
     unsigned sz = sizeof(uint64_t);
@@ -273,6 +278,11 @@ static inline unsigned char *byte_memset(unsigned char *out, unsigned len) {
 
     unsigned char *from = out - 1;
     unsigned char c = *from;
+
+#if defined(CALL_MEMSET)
+    MEMSET(out, c, len);
+    return out + len;
+#endif
 
     /* First, deal with the case when LEN is not a multiple of SZ. */
     MEMSET(out, c, sz);


memcopy.h
```diff
@@ -266,6 +266,12 @@ static inline unsigned char *chunk_memcpy(unsigned char *out, unsigned char *fro
     return out;
 }
 
+#if defined(__arm__) || defined(__aarch64__) || defined(_M_ARM)
```
sebpop: I know that calling memset is better for x86_64, so it is better to change the ifdef like this:
```
#undef CALL_MEMSET
#if defined(__x86_64) || defined(__i386_)
# define CALL_MEMSET 1
#endif
```

Dead2:
Has this been tested on multiple compilers?
GCC will change memset() calls to __builtin_memset if length is known and it thinks it will be faster.
No idea what LLVM and other compilers would do.

Also, older glibc (on EL6 for example) might not perform as well, I think they improved memset/memcpy etc a lot sometime after that, but I can't remember what version that was.

Just want to make sure we don't assume too much.

sebpop:
LLVM does the same as gcc: it recognizes memset calls or memset loops and if the number of bytes to set is below a threshold defined by the target back-end it will generate custom inline code.

Both gcc and clang show good speedups on x86_64:
gcc https://gist.github.com/sebpop/0785dcf1c70c70b89b040c358184ae1b
clang https://gist.github.com/sebpop/326c5120b6fa6cffc0e5e2017e907383

Dead2:
Hm, I am not quite convinced. Those tests show a lot of instances of being slower with the patch too.
I guess there is no summary for how many tests were faster/slower and whether the total time for all tests was slower/faster?

Dead2:
In my tests using gcc 8.2 on x86_64, the results are as follows:
compression average 0.71% speedup
decompression average 4% slower
I re-ran the benchmarks to make sure the decompression results were correct, and they are indeed consistent.

Compression at level 0 (store) seems to be around 25% faster, but the inaccuracy at that level is high because it is so fast on the 40MiB testfile I used.

Could it be that the part causing most of the speedup here is from storing literals? If so, perhaps that part of the code can be optimized directly instead.

In its current form, I do not want to merge this, the 4% slowdown on decompression is not worth the 0.7% speedup on compression. But I would like to see this merged in another form that fixes the unintentional slowdown.

Dead2:
Hmm, since this patch only touches byte_memset, and that is only used by inffast, it should not speed up deflate at all. But it might have to do with deflate-relevant code segments getting moved around and being placed in more cache-optimal locations and/or cache aliasing.

I ran the same tests using clang (llvm 6.0.1), also with only the default -O2.
Compression: 0.14% faster  (averaged over 70 effective testruns, this might be noise)
Decompression: 1.31% slower

So, for me, this patch does not do what it claims since I see a big slowdown instead..
This is x86_64 using an Haswell cpu, with gcc8.2, zlib-ng compiled with -O2 (like most distros would do), and glibc 2.27.
What are the details of the systems that show a speedup with this?
It might still be optimal in some cases, but we'd need to narrow down when that is.

GCC with -O3 sees a 0.7% speedup with this patch, so it might be that this patch requires some compiler optimizations that are not enabled by default in -O2.
We could the detect the -O3 in CFLAGS in configure/cmake and set a define that enables this code bypass. Or we could narrow down what flag is in -O3 that helps us here, and possibly auto-enable that flag by default..

sebpop:
The system I benchmarked this patch on is an i7-4790K CPU @ 4.00GHz
gcc (Ubuntu 5.4.0-6ubuntu1\~16.04.10) 5.4.0 20160609
clang version 6.0.0-1ubuntu2\~16.04.1 (tags/RELEASE_600/final)
-O3, glibc 2.23

I am abandoning this patch as its performance impact is not consistent across systems.

PR #268, opened 2019-01-02
Remove unused flag

sebpop:


arch/x86/deflate_quick.c
```diff
@@ -92,11 +92,11 @@ static inline long compare258(const unsigned char *const src0, const unsigned ch
         "cmp        $256 + 16, %[ax]\n\t"
         "jb         1b\n\t"
 
-#ifdef X86
+# if !defined(__x86_64)
```
sebpop, line 95: Without this change I see a fail when configuring for 32-bit x86:
```
$ ./configure --32; make -j
[...]
arch/x86/deflate_quick.c: Assembler messages:
arch/x86/deflate_quick.c:90: Error: `movzwq' is only supported in 64-bit mode
```

mtl1979: For me it looks there is underlying problem in the configure script as it should override ARCH when "--32" is specified on command line.

sebpop: The error seems to be in the way we detect GCC_ARCH in configure:
```
GCC_ARCH=`$CC $CFLAGS -dumpmachine | sed 's/-.*//g'`
```
Here is the output of gcc and clang:
```
$ gcc  -m32 -std=c99 -dumpmachine 
x86_64-linux-gnu
$ clang  -m32 -std=c99 -dumpmachine 
i386-pc-linux-gnu
```
It looks like it is a known problem: https://gcc.gnu.org/ml/gcc-help/2018-05/msg00053.html
I think a good solution is to check whether the compiler is gcc in which case we should do:
```
$ gcc -print-multiarch
x86_64-linux-gnu
$ gcc -print-multiarch -m32
i386-linux-gnu
```
I will prepare a separate patch that fixes this problem.

sebpop:
The revert of the "remove -DX86" patch was only intended for my testing.  I will push again without it.

arch/x86/deflate_quick.c
mtl1979: @sebpop The patch should also handle cases where the compiler doesn't support ```-print-multiarch```.

configure
```diff
@@ -860,7 +865,7 @@ fi
 fi
 
 # Check for PCLMULQDQ intrinsics
-if test "${ARCH}" = "i686" || test "${ARCH}" = "x86_64"; then
+if test "${ARCH}" = "i386" -o "${ARCH}" = "i686" -o "${ARCH}" = "x86_64"; then
```
Dead2: Hm, seeing this code makes me wonder whether we could clean it up by using something like this instead.
  case $ARCH in i386 | i486 | i586 | i686 | x86_64)
It is kind of abusing the case system, but I think it would be A LOT easier to read than all the clunky if tests.
What do you think?

arch/x86/deflate_quick.c
```diff
@@ -92,11 +92,11 @@ static inline long compare258(const unsigned char *const src0, const unsigned ch
...
         "movzwl     -16(%[src0], %[ax]), %[dx]\n\t"
-#else
+# else
         "movzwq     -16(%[src0], %[ax]), %[dx]\n\t"
-#endif
+# endif
```
Dead2, line 99: I see no reason to change the whitespace here.
The whitespace makes it easier to see that this if is nested inside another if.

Dead2:
I like the "This patch fixes a problem with detecting the target that gcc compile" commit, although that commit does not fit with the current PR description.

The removal of the flags seems fine to me. The original thinking was we might begin using them for more stuff, but instead pretty much everything has introduced their own defines. And as long as we can depend on compiler-provided defines, that is better.

arch/x86/deflate_quick.c
sebpop: > The whitespace makes it easier to see that this if is nested inside another if.

Yes, this is the case: `#if !defined(__x86_64) #else #endif` is nested within the `#else` clause of `#ifdef _MSC_VER`.

configure
sebpop: I will amend the patch with the suggested pattern.

arch/x86/deflate_quick.c
Dead2: I must have been confused, I could swear I thought you were removing the spaces. All good then.

Dead2:
This patchset looks really good now, I haven't tested it yet.

Dead2:
Tested locally, and I see nothing wrong now. Thanks!

PR #269, opened 2019-01-04
Rename {in,de}flate_copyright -> zlibng_{in,de}flate_copyright.

yasushi-saito:
If we compile zlib-ng in non-backward-compatible mode, and link both
zlib and zlibng into an application, we get duplicate symbol errors
on these names.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -55,7 +55,7 @@
 #include "match.h"
 #include "functable.h"
 
-const char deflate_copyright[] = " deflate 1.2.11.f Copyright 1995-2016 Jean-loup Gailly and Mark Adler ";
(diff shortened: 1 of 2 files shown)


mtl1979:
You should use PREFIX() macro from zbuild.h so it matches how it is done elsewhere in the code.

yasushi-saito:
I found this really doesn't work yet. The context is that we have a cgo wrapper for zlibng (https://github.com/yasushi-saito/zlibng), which works great as long as (1) you don't use both zlibng and the system-standard libz, or (2) enable zlib-compatibility functions in zlibng. But in our case, we statically link zlibng w/o zlib-compatibility functions, and there are other codepaths that force linking libz. That causes libz to call into zlibng (or may be reverse) and crashes the application.

For now, I worked this around by adding "zng_" prefix to all the non-static symbols defined in zlibng in the cgo code. Doing this always shouldn't cause any harm.

PR #270, opened 2019-01-10
move longest_match to match.h and make it static inline

sebpop:
x86_64-linux compiled with clang and gcc at -O3 shows that inlining
longest_match() brings good speedup overall with speedups of up to 8%.
clang https://gist.github.com/sebpop/17d9c340cfaf16c1854017efefce6fc3
gcc https://gist.github.com/sebpop/451160356fc4e73541a1b8eef7dceb97

performance is also better on aarch64-linux with speedups up to 3.5%.
clang https://gist.github.com/sebpop/aec3e068a61c1daddf9c72ab4bc33281
gcc https://gist.github.com/sebpop/f0d4e6bfd58d95b313c955b5d747a74f


Dead2:
We already use the postfix '_p.h' for header files containing code to be inlined, so this should be named match_p.h

I'd also prefer if this change was done with 'git mv' instead of 'git rm', so changes to the contents actually appear as such instead of looking like a whole new file.

Dead2:
Seems like git mv was done, so that might need to be done in a separate commit in order for github to be able to show only the changed contents properly.

Dead2:
In my tests, this shows an average of 2.9% speedup on compression using gcc 8.2 on x86_64 (using the default cflags of -O2).

PR #271, opened 2019-01-16
Unroll less by default

Dead2:
Replace the UNROLL_LESS define with UNROLL_MORE, making UNROLL_LESS the default.

Performance benchmarks have so far not shown that any platform benefits from UNROLL_MORE,
although this might be beneficial on older compilers/cpus or for compiling without optimizations.

The extra UNROLL_MORE code should be considered for removal since it is never enabled by us
and will likely only serve to confuse and contribute to bitrot.


sebpop:
Looks good to me.

PR #272, opened 2019-01-16
Enable deflate_medium by default.

Dead2:
Let deflate_medium be enabled by default when compiling without using our configure or cmake scripts.
Also cuts down on the defines we send to the compiler by default.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -75,7 +75,7 @@ static void slide_hash            (deflate_state *s);
 static block_state deflate_stored (deflate_state *s, int flush);
 ZLIB_INTERNAL block_state deflate_fast         (deflate_state *s, int flush);
 ZLIB_INTERNAL block_state deflate_quick        (deflate_state *s, int flush);
-#ifdef MEDIUM_STRATEGY
+#ifndef NO_MEDIUM_STRATEGY
 ZLIB_INTERNAL block_state deflate_medium       (deflate_state *s, int flush);
 #endif
 ZLIB_INTERNAL block_state deflate_slow         (deflate_state *s, int flush);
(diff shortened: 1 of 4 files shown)


deflate.c
```diff
@@ -75,7 +75,7 @@ static void slide_hash            (deflate_state *s);
 static block_state deflate_stored (deflate_state *s, int flush);
 ZLIB_INTERNAL block_state deflate_fast         (deflate_state *s, int flush);
 ZLIB_INTERNAL block_state deflate_quick        (deflate_state *s, int flush);
-#ifdef MEDIUM_STRATEGY
+#ifndef NO_MEDIUM_STRATEGY
```
sebpop, line 78: Can we avoid double negations? 
"not defined no medium" means "defined medium" which is easier to read and non ambiguous.
Can we avoid all the changes below to the `deflate*.c` files by changing the way we set the `-DMEDIUM_STRATEGY` flag in configure and cmake?

Dead2: The problem is that since we depend on MEDIUM_STRATEGY being set, any users that do not use our cmake or configure scripts will never see those default features at all unless they set them manually. (Yes, I am looking at you @nemequ Squash).

If you look at squashcompression.com, and compare our performance/compression with that of stock zlib, it will be near identical, because no defines will be set and that results in a build that mostly uses stock zlib code. Probably nobody compiling zlib-ng standalone will ever build it in that specific configuration with everything disabled.

There probably are other projects that also include zlib similarly by just including the source/header files into their own project and build it without thinking about any optional defines.

I agree that double negation is bad, but I don't really see any way around that here, unless we make a proxy define that takes the NO_MEDIUM_STRATEGY define and turns it into the old MEDIUM_STRATEGY.
Another alternative is calling it DISABLE_MEDIUM_STRATEGY for example, that would make it a little more clear, but much longer.

I'd like to get rid of other safe-by-default defines that we do too, such as UNALIGNED_OK and possibly some of the x86 defines that enable runtime detection of instruction sets.

Dead2: Fyi, I have been making a little progress on cleaning up Squash's zlib-ng integration, but it is still a long way from good. It should for example disable compat mode and use our prefixed functions instead.
See https://github.com/Dead2/squash
I have been unable to make it actually compile zlib-ng with the correct defines even though CMakeCache.txt contains the correct options, not sure what is going on there. Hardcoding the defines would work, but it would be better to let our cmake script be able to analyze the system and set the correct ones.

Not sure I can spend any more time on that for now, but feel free to contribute, I want to make a PR to Squash with the current changes at the very least.

sebpop: I understand.  Let's keep the current form.
The patch looks good to me.

PR #273, opened 2019-01-16
ENH: Remove superflous setting for cmake 2.6.0+

hjmjohnson:
As of CMake 2.6.0 the ELSE() and ENDIF() constructs can be empty. The
same is true for closing constructs on ENDMACRO(), ENDFUNCTION(), and
ENDFOREACH().

If you require 2.4.x compatibility, CMake 2.4.3 or greater
recognizes the CMAKE_ALLOW_LOOSE_LOOP_CONSTRUCTS option (which is
superfluous in 2.6.0)


sebpop:
Looks good to me.  Thanks for the patch!

PR #274, opened 2019-01-16
Update min cmake version

hjmjohnson:


sebpop:
The changes look good to me.

CMakeLists.txt
```diff
@@ -1,10 +1,41 @@
...
+if(CMAKE_VERSION VERSION_LESS 3.12)
+  cmake_policy(VERSION ${CMAKE_VERSION})
+else()
+  cmake_policy(VERSION 3.5.1...3.13.2)
+endif()
+
 set(CMAKE_MACOSX_RPATH 1)
 
-project(zlib C)
+# If not specidied on the command line, enable C99 as the default
```
donny-dont: Typo "specified"

donny-dont (commented):
Looks fine just saw a typo

Dead2:
Hmm, I am a bit torn on this one. I like the changes, but..

Doing this would mean we no longer support the cmake version currently shipped in the latest version of RHEL, Centos and Scientific Linux without requiring pulling in the EPEL repo to obtain cmake3. (EPEL contains cmake 3.6.1 btw)
It likely also means we loose cmake support on platforms I have no access to test on, HPUX, Solaris, etc..

On the other hand, we also have configure, and that is being maintained at least as good as our cmake config.

Feel free to weigh in with comments pro/con and +1/-1.

mtl1979:
Combining bug fixes and enhancements in same pull request or patch set in my opinion is very bad idea. Features that require cmake version more recent than 2.8.4 should be conditional or postponed until all distributions we have decided to support will have the version in main repository.

hjmjohnson:
@Dead2 & @mtl1979:  Regarding the update of cmake version.

The line of code at  https://github.com/zlib-ng/zlib-ng/blame/develop/CMakeLists.txt#L709  is invalid for  versions of cmake < 3.0.     My guess is that no one has attempted to use cmake < 3.0 in the past 7 months.

```
johnsonhj@hj$ git rebase upstream/develop    # Current develop branch does not build with cmake < 3.0
johnsonhj@hj$ module load cmake/2.8.9 
johnsonhj@hj$ mkdir ../zlib-ng-2.8.9
johnsonhj@hj$ cd ../zlib-ng-2.8.9/
johnsonhj@hj$ cmake ../zlib-ng
<clip content>
CMake Error at CMakeLists.txt:709 (target_include_directories):
  Unknown CMake command "target_include_directories".
```
 
When choosing 3.5.1, that is the version of cmake delivered in the previous version of Ubuntu 16.04 (Xenial).  All other platforms will need to install a custom version of cmake anyway for the past 7 months.  Moving back to 3.0.0 would not be that hard, but there are no distributions that I know of that delivered less than 3.5.1 by default.

@mtl1979 In principle I agree.  I thought that all of these were "enhancements".  There was not a particular "bug" that I was fixing.   Unfortunately, these changes are not easily decoupled as independent PR's on the develop branch.  I'd be happy to submit them one at a time (waiting for the previous one to be approved before the next one was submitted), but I think the overall story of intent is more difficult to follow.

hjmjohnson:
@mtl1979 See #285 that only increments cmake to a version that allows building.

mtl1979:
@hjmjohnson I have tried to be really careful about what patches to accept to CMakeLists.txt but in the end it is @Dead2 who makes the final decision... Originally we set oldest supported tools to ones released 2013 or later, but it might be wise to adjust the year sometime in the future as 2013 is now 6 years in the past. Cmake 2.8.4 was selected as minimum because it was at that time the most recent version that all distributions we have access to did have as package in main repository.

hjmjohnson:
@mtl1979 Thank you for the explanation.  I can make any range of values work, but if we change the stated minimum back to a value prior to cmake version 3.0, then we will also need to modify changes made 7 months ago as well.

mtl1979:
@hjmjohnson The minimum required cmake version is something we have to revisit with @Dead2 when he is online. I don't like scolding him for making bad decisions, but if one of his decisions introduce a bug, then he himself must revert the decision. Nobody of us is perfect, but if we screw up, we must fix things as soon as possible, not wait months like has happened before...

hjmjohnson:
I will wait for @Dead2  to provide feedback, and then make appropriate modifications to support the version that is requested.

Dead2:
Our automated Buildkite testing is currently running cmake 2.8.12 for its tests, and the tests do not fail with the current scripts, but they do fail with the ones in this PR of course. Not sure why this does not fail currently, but it is possible that RHEL has backported the most important features like "target_include_directories" to avoid all their customers getting problems.

3.6.1 is a version I know I can set up automated testing of, with some work. 3.5.1 would be a lot more work for me to set up any automated testing for.
I glanced through the 3.6 changelog, but didn't see anything really interesting, either version would be fine with me.

Is it perhaps possible to set up travis to run one of its tests using an older ubuntu version, that runs either 3.6.1 or 3.5.1? If so, having one or two of the travis tests be a validation of the cmake version requirements fitting the actual code, would be great. Also, is there any way to get cmake to lint or complain about errors in the code even if it is not in a codepath it is currently running?  "-Wdev -Werror=dev" perhaps?

Older platforms can run our configure script. However, that does require Bash, so it could potentially be a pain on some platforms. The primary reasons for using Bash instead of Ash is that Bash is what we use for testing, and I know of no easy way to automate testing to verify ash-compatibility through for example Travis.

I am leaning towards increasing the cmake minimum version requirement from 2.8.4 to 3.5.1 or 3.6.1.
**Lets put it to a vote: All for Cmake version upgrade give thumbs up, all against give thumbs down.**

mtl1979:
I have 3.5.1 on Linux and 3.10.2 on MinGW...

hjmjohnson:
@Dead2 I am shocked that Buildkite is working with cmake 2.8.12.  I've downloaded cameo 2.8.12 and verified that the build does not work.    I choose 3.5.1 because it is the default version for ubuntu 16.04 (trusty) environment.

hjmjohnson:
👍   <-- for increasing cmake version to at least 3.1 (anything newer than that will reduce maintenance burden for core developers at the cost of a higher requirement).  I choose 3.5.1 as a balance between the competing requirements.

Dead2:
@sebpop @mtl1979 No votes for/against?

Dead2:
Well, with a whopping count of 1 vote, this gets approved ;)
Thanks @hjmjohnson

hjmjohnson:
@Dead2 Thanks.  I'll keep my eye on the issues to see if problems show up.

mtl1979:
I didn't see any reason to vote myself, as I would have voted both ways depending on the new minimum version... I only commented on which version should be set as minimum if approved.

PR #275, opened 2019-01-16
Fix missing header warning

hjmjohnson:


Dead2:
I like the 'Fix missing header warning' commit, if the PR did not contain all the other commits, I would have merged this now.

hjmjohnson:
@Dead2 Thank you for the feedback.  I isolated this fix that is independent of the other changes.

Dead2:
Looks good to me.

Issue #225, opened 2018-11-27
make test fails when configured with --debug


sebpop:
```
./configure --debug; make; make test
[...]
make -C test
make[1]: Entering directory '/work/spop/zlib-ng/test'
hello world
zlib version 1.9.9 = 0x1990, compile flags = 0x1a9
uncompress(): hello, hello!
inflate(): hello, hello!
bad compressed size
                *** zlib test FAILED ***
Makefile:49: recipe for target 'teststatic' failed
make[1]: *** [teststatic] Error 1
make[1]: Leaving directory '/work/spop/zlib-ng/test'
Makefile:119: recipe for target 'test' failed
make: *** [test] Error 2
```
Fails with both gcc 5.4.0 and clang 6.0.0

sebpop:
It passes if I comment out the asserts:
```
#   define Assert(cond, msg) // {if(!(cond)) z_error(msg);}
```

sebpop:
With this patch it passes:
```
--- a/trees.c
+++ b/trees.c
@@ -909,7 +909,7 @@ void ZLIB_INTERNAL _tr_flush_block(deflate_state *s, char *buf, unsigned long st
         s->compressed_len += 3 + s->opt_len;
 #endif
     }
-    Assert(s->compressed_len == s->bits_sent, "bad compressed size");
+    // Assert(s->compressed_len == s->bits_sent, "bad compressed size");
     /* The above check is made mod 2^32, for files larger than 512 MB
      * and unsigned long implemented on 32 bits.
      */
```

sebpop:
It fails with:
```
s->compressed_len = 3786
s->bits_sent = 3802
```

mtl1979:
This issue has existed since very early days of zlib-ng. Back then I tried to trace where the discrepancy is caused, but my skills were not adequate enough.

sebpop:
`example` passes if I change `Z_BEST_SPEED` to `Z_BEST_COMPRESSION`, and it also fails at `Z_NO_COMPRESSION`.

The problem is in intel's patches in `deflate_quick` which is only called at compression level 1.

sebpop:
Fixed.

PR #282, opened 2019-01-17
fix bug #225: account bits sent

sebpop:


sebpop:
This fix follows the same accounting as used in trees.c:
```
 * Send one empty static block to give enough lookahead for inflate.
 * This takes 10 bits, of which 7 may remain in the bit buffer.
 */
void ZLIB_INTERNAL _tr_align(deflate_state *s) {
    send_bits(s, STATIC_TREES << 1, 3);
    send_code(s, END_BLOCK, static_ltree);
#ifdef ZLIB_DEBUG
    s->compressed_len += 10L; /* 3 for block type, 7 for EOB */
#endif
    bi_flush(s);
}
```

Dead2:
Looks good to me, but will need a rebase.

Issue #208, opened 2018-09-19
store to misaligned address in `make check` with sanitizers


sebpop:
```
trees.c:1096:9: runtime error: store to misaligned address 0x63100012d3e3 for type 'uint16_t' (aka 'unsigned short'), which requires 2 byte alignment
0x63100012d3e3: note: pointer points here
 2f  48 08 5d 3f 00 be be be  be be be be be be be be  be be be be be be be be  be be be be be be be
              ^ 

trees.c:798:5: runtime error: store to misaligned address 0x63100003c801 for type 'uint16_t' (aka 'unsigned short'), which requires 2 byte alignment
0x63100003c801: note: pointer points here
 00 80 21  00 c1 31 01 00 00 00 c2  a0 f5 4f 6d 0b 2f a0 00  00 00 00 00 00 00 00 00  00 00 00 00 00
              ^ 

trees.c:799:5: runtime error: store to misaligned address 0x63100003c803 for type 'uint16_t' (aka 'unsigned short'), which requires 2 byte alignment
0x63100003c803: note: pointer points here
 21  00 00 00 01 00 00 00 c2  a0 f5 4f 6d 0b 2f a0 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00
              ^ 
```

sebpop:
As of today I do not see the errors in trees.c anymore.  However with #279 I see the following errors:
```
$ CC=clang CFLAGS="-g" LDFLAGS="-g" ./configure --with-fuzzers --with-sanitizers ; make clean && make -j ; make test
[...]
zlib-ng/arch/x86/deflate_quick.c:154:5: runtime error: store to misaligned address 0x631000014801 for type 'unsigned int', which requires 4 byte alignment
0x631000014801: note: pointer points here
 00 80 02  d3 07 00 00 be be be be  be be be be be be be be  be be be be be be be be  be be be be be
              ^ 
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior zlib-ng/arch/x86/deflate_quick.c:154:5 in 
[...]
zlib-ng/trees.c:574:19: runtime error: unsigned integer overflow: 0 - 1 cannot be represented in type 'unsigned long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior zlib-ng/trees.c:574:19 in 
[...]
zlib-ng/trees.c:576:27: runtime error: unsigned integer overflow: 0 - 8 cannot be represented in type 'unsigned long'
[...]

PR #283, opened 2019-01-17
fix bug #208: let the compiler generate code for unaligned stores

sebpop:
to avoid this error:
zlib-ng/arch/x86/deflate_quick.c:154:5: runtime error: store to misaligned address 0x631000014801 for type 'unsigned int', which requires 4 byte alignment
0x631000014801: note: pointer points here
00 80 02  d3 07 00 00 be be be be  be be be be be be be be  be be be be be be be be  be be be be be
               ^
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior zlib-ng/arch/x86/deflate_quick.c:154:5 in

### DIFF
--- a/arch/x86/deflate_quick.c
+++ b/arch/x86/deflate_quick.c
@@ -23,6 +23,7 @@
 #  include <nmmintrin.h>
 #endif
 #include "deflate.h"
+#include "memcopy.h"
 
 #ifdef ZLIB_DEBUG
 #include <ctype.h>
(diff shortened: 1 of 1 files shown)


Dead2:
Code looks good to me, but I need to test it later.

Dead2:
This resulted in a 1-2% speedup on deflate_quick in my tests, seems like the compiler is better able to optimize the code now.
All is fine in my tests.

PR #285, opened 2019-01-17
BUG:  CMake 2.8.4 does not support required features

hjmjohnson:
CMake Error at CMakeLists.txt:682 (target_include_directories):
  Unknown CMake command "target_include_directories".

target_include_directories was introduced in cmake 3

C_STANDARD 99 support was added in cmake 3.1

Ubuntu 16.04 (Xenial) was distributed with cmake 3.5.1 by default.

CMake versions > 3.3 allow simplified implementations of modern
cmake compilation support.

===
For newer versions of cmake (upto a maximum validated version [3.13.2 in this case]),
use newer cmake policies available.  The newer policies often provide better
diagnostics for subtle build related issues.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,10 @@
-cmake_minimum_required(VERSION 2.8.4)
+cmake_minimum_required(VERSION 3.5.1)
+if(CMAKE_VERSION VERSION_LESS 3.12)
+  cmake_policy(VERSION ${CMAKE_VERSION})
+else()
+  cmake_policy(VERSION 3.5.1...3.13.2)
+endif()
+
 set(CMAKE_MACOSX_RPATH 1)
 
 project(zlib C)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -20,13 +26,7 @@ include(CheckCSourceRuns)
...
 
 # Enable C99
-if (CMAKE_VERSION VERSION_LESS "3.1")
-    if (CMAKE_C_COMPILER_ID STREQUAL "GNU")
-        set (CMAKE_C_FLAGS "-std=c99 ${CMAKE_C_FLAGS}")
-    endif ()
-else ()
-    set (CMAKE_C_STANDARD 99)
-endif ()
+set (CMAKE_C_STANDARD 99)
```
donny-dont: Just a nit here which more applies to the file's formatting in general. Whether there's a space after a `set` is not consistent.

donny-dont (commented):
LGTM

CMakeLists.txt
Dead2: True, the same goes for if/else/endif.
All these are mainly (I see a few exceptions) used without a space before the parentheses, so sticking to the same style would be best.

Dead2:
Is this still needed?

hjmjohnson:
Mo.

PR #287, opened 2019-01-21
crc32: aarch64 accelerated crc32 calculate based on linux kernel

RuiSunSilhouette:
For arm64 plantform ,the hardware crc32 instructions will improve crc32 performance. base on the linux kernel crc32 acceletated calculation, we add it to arch/aarch64 and test on Cortex-A72 which get improvements for about 40%

Hope these patches would merged into zlib-ng mainline for others want to test on arm64.

Thanks & Best Regards.

cc: Ard Biesheuvel <ard.biesheuvel@linaro.org>

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -103,10 +103,14 @@ check: test
 
 .SECONDARY:
 
-$(ARCHDIR)/%.o: $(SRCDIR)/$(ARCHDIR)/%.c
+$(ARCHDIR)/%.o: $(SRCDIR)/$(ARCHDIR)/%.c 
+	$(MAKE) -C $(ARCHDIR) $(notdir $@)
+$(ARCHDIR)/%.o: $(SRCDIR)/$(ARCHDIR)/%.S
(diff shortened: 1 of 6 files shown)


Dead2:
This code is licensed under GPLv2, and we can therefore not merge this into zlib-ng.
Zlib-ng uses the Zlib license.
Would it be possible to get this code re-licensed?

Dead2:
Actually, we would like not to use asm code, since that makes it very hard to maintain.
It would need to be translated into compiler intrinsics.

mtl1979:
I don't see much difference comparing to the intrinsic code we currently have except the reverse instructions for big-endian processors.

RuiSunSilhouette:
Will you accept inline assembler?

mtl1979:
We prefer intrinsics because that way we catch differences between compilers more easily. Using intrinsics also allows compiler to better integrate the code to surrounding lines and optimize as whole.

Issue #289, opened 2019-01-22
GCC8 warning: argument to ‘sizeof’ in ‘strncpy’ same as source


Dead2:
```
/home/hansr/github/zlib/zlib-ng/test/example.c: In function ‘test_dict_inflate’:
/home/hansr/github/zlib/zlib-ng/test/example.c:465:48: warning: argument to ‘sizeof’ in ‘strncpy’ call is the same expression as the source; did you mean to use the size of the destination? [-Wsizeof-pointer-memaccess]
     strncpy((char*)uncompr, garbage_str, sizeof(garbage_str));
```
I am a bit confused about this warning.
Would it not be better to check the size of the source before copying, so you don't copy garbage due to copying more than the source buffer size is? Although, writing more than the dst buffer size would also be bad.

The most secure would probably be if strncpy took in two size variables and only respected the smallest of them. But as it is, strncpy only takes one, and passing it the smallest one would be the logical solution to me.

So, do we change anything or do we disable the warning?

mtl1979:
I think the warning is valid. Length of the second parameter is implied as it needs to be zero-terminated.

Dead2:
Ohh, of course, I forgot that it was a string copy so it would stop on null. I was thinking of memcpy behavior. Yep that makes perfect sense then, and is a bug we need to fix.

sebpop:
git blame was pointing to code that I modified in 06bd4c8b105fc97cf4998d2ac22acbbc41df951a
The fix reverts the use of strncpy back to strcpy.
I also tried with the original: to only copy one "garbage" instead of 3 and that was still failing with "unititialized memory" when configured with --with-msan.

Dead2:
Confirmed fixed in develop.

PR #290, opened 2019-01-22
fix bug #289: use strcpy instead of strncpy

sebpop:
to avoid a GCC 8 warning:

test/example.c:465:48: warning: argument to ‘sizeof’ in ‘strncpy’ call is the same expression as the source; did you mean to use the size of the destination? [-Wsizeof-pointer-memaccess]
     strncpy((char*)uncompr, garbage_str, sizeof(garbage_str));


Dead2:
Thanks!

nmoinvaz:
Maybe it should be: `strncpy((char*)uncompr, garbage_str, uncomprLen);`

PR #291, opened 2019-01-23
cleanup: remove partial do { } while constructs from macros

sebpop:
also remove dead code that was there only to close the open braces of the
do-while construct.

### DIFF
--- a/arch/x86/crc_folding.c
+++ b/arch/x86/crc_folding.c
@@ -25,32 +25,12 @@
 
 #include "crc_folding.h"
 
-
-#define CRC_LOAD(s) \
-    do { \
-        __m128i xmm_crc0 = _mm_loadu_si128((__m128i *)s->crc0 + 0);\
-        __m128i xmm_crc1 = _mm_loadu_si128((__m128i *)s->crc0 + 1);\
-        __m128i xmm_crc2 = _mm_loadu_si128((__m128i *)s->crc0 + 2);\
-        __m128i xmm_crc3 = _mm_loadu_si128((__m128i *)s->crc0 + 3);\
-        __m128i xmm_crc_part = _mm_loadu_si128((__m128i *)s->crc0 + 4);
-
-#define CRC_SAVE(s) \
-        _mm_storeu_si128((__m128i *)s->crc0 + 0, xmm_crc0);\
-        _mm_storeu_si128((__m128i *)s->crc0 + 1, xmm_crc1);\
-        _mm_storeu_si128((__m128i *)s->crc0 + 2, xmm_crc2);\
-        _mm_storeu_si128((__m128i *)s->crc0 + 3, xmm_crc3);\
-        _mm_storeu_si128((__m128i *)s->crc0 + 4, xmm_crc_part);\
-    } while (0);
-
 ZLIB_INTERNAL void crc_fold_init(deflate_state *const s) {
-    CRC_LOAD(s)
-
-    xmm_crc0 = _mm_cvtsi32_si128(0x9db42487);
-    xmm_crc1 = _mm_setzero_si128();
-    xmm_crc2 = _mm_setzero_si128();
-    xmm_crc3 = _mm_setzero_si128();
-
-    CRC_SAVE(s)
+    /* CRC_SAVE */
+    _mm_storeu_si128((__m128i *)s->crc0 + 0, _mm_cvtsi32_si128(0x9db42487));
+    _mm_storeu_si128((__m128i *)s->crc0 + 1, _mm_setzero_si128());
+    _mm_storeu_si128((__m128i *)s->crc0 + 2, _mm_setzero_si128());
+    _mm_storeu_si128((__m128i *)s->crc0 + 3, _mm_setzero_si128());
 
     s->strm->adler = 0;
 }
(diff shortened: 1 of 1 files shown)


turol:
It's considered good practice to keep your macro contents inside a no-op `do` .. `while` loop. Consider this:

```
if (somecondition)
    CRC_LOAD(s);
```

If there's a loop this will do what you expect. Without a loop it will unconditionally execute all the statements except the first one.

mtl1979:
@turol no-op loop wouldn't work as it declares new variables. The original problem was that end of the loop was after "return" statement and as such dead code.

arch/x86/crc_folding.c
```diff
@@ -39,8 +38,7 @@
         _mm_storeu_si128((__m128i *)s->crc0 + 1, xmm_crc1);\
         _mm_storeu_si128((__m128i *)s->crc0 + 2, xmm_crc2);\
         _mm_storeu_si128((__m128i *)s->crc0 + 3, xmm_crc3);\
-        _mm_storeu_si128((__m128i *)s->crc0 + 4, xmm_crc_part);\
-    } while (0);
+        _mm_storeu_si128((__m128i *)s->crc0 + 4, xmm_crc_part);
 
 ZLIB_INTERNAL void crc_fold_init(deflate_state *const s) {
     CRC_LOAD(s)
```
Dead2: CRC_LOAD is effectively dead code here now and could be removed. Variables below would need __m128i prefix though.

Dead2:
I would be fine with having a complete "do { } while 0" in each macro, and that is a more safe code-style even though it is not currently needed here.

Both alternatives are much, much better than the current code that has partial do-while in each macro.

Dead2:
Uhm, why did you remove the macros completely, the previous iteration was looking very close to ready..

sebpop:
Every use of those macros had some problem: either loading unnecessary data, defining variables that would not be used (and the compiler complaining about), and I was left with only one place where the macros were used.  So why using them if they produce slower code, or compiler warnings?

I find the amended version much better than with the macros.

arch/x86/crc_folding.c
sebpop: This is one of the reasons why the macros stand in the way.

arch/x86/crc_folding.c
```diff
@@ -391,8 +378,10 @@ uint32_t ZLIB_INTERNAL crc_fold_512to32(deflate_state *const s) {
 
     uint32_t crc;
     __m128i x_tmp0, x_tmp1, x_tmp2, crc_fold;
-
-    CRC_LOAD(s)
+    __m128i xmm_crc0 = _mm_loadu_si128((__m128i *)s->crc0 + 0);
+    __m128i xmm_crc1 = _mm_loadu_si128((__m128i *)s->crc0 + 1);
+    __m128i xmm_crc2 = _mm_loadu_si128((__m128i *)s->crc0 + 2);
+    __m128i xmm_crc3 = _mm_loadu_si128((__m128i *)s->crc0 + 3);
```
sebpop, line 392: This one does not need xmm_crc_part and the compiler would complain that I define a variable never used after.

arch/x86/crc_folding.c
```diff
@@ -251,7 +230,11 @@ ZLIB_INTERNAL void crc_fold_copy(deflate_state *const s, unsigned char *dst, con
     unsigned long algn_diff;
     __m128i xmm_t0, xmm_t1, xmm_t2, xmm_t3;
 
-    CRC_LOAD(s)
+    __m128i xmm_crc0 = _mm_loadu_si128((__m128i *)s->crc0 + 0);
+    __m128i xmm_crc1 = _mm_loadu_si128((__m128i *)s->crc0 + 1);
+    __m128i xmm_crc2 = _mm_loadu_si128((__m128i *)s->crc0 + 2);
+    __m128i xmm_crc3 = _mm_loadu_si128((__m128i *)s->crc0 + 3);
+    __m128i xmm_crc_part;
```
sebpop, line 239: This one would be a redundant load with the macro: xmm_crc_part gets initialized on all paths below before it get used.

arch/x86/crc_folding.c
```diff
@@ -25,32 +25,11 @@
...
-    xmm_crc0 = _mm_cvtsi32_si128(0x9db42487);
-    xmm_crc1 = _mm_setzero_si128();
-    xmm_crc2 = _mm_setzero_si128();
-    xmm_crc3 = _mm_setzero_si128();
-
-    CRC_SAVE(s)
+    _mm_storeu_si128((__m128i *)s->crc0 + 0, _mm_cvtsi32_si128(0x9db42487));
+    _mm_storeu_si128((__m128i *)s->crc0 + 1, _mm_setzero_si128());
+    _mm_storeu_si128((__m128i *)s->crc0 + 2, _mm_setzero_si128());
+    _mm_storeu_si128((__m128i *)s->crc0 + 3, _mm_setzero_si128());
```
sebpop, line 33: As you said, there is no need to load in any of the xmm_crc* variables, and there is no use of xmm_crc_part.

Dead2:
All good points, and I have to agree.
It would be nice to have a one-line comment above each of those intrinsic sections to add a bit of clarity to the pretty dense code though. Not that the rest of that code is much better, but lets improve things along the way if we can. :)

sebpop:
> It would be nice to have a one-line comment above each of those intrinsic sections to add a bit of clarity to the pretty dense code

What should that comment line be?
As I am not changing the behavior of the current algorithm,
I have no idea what I should write for those loads and stores to improve the readability of the code.

I think we better leave the code as is, or commit the patch without adding comments.

Dead2:
I was thinking something simple like
// LOAD
and
// SAVE
for example. Thus providing the same level of context info as the macro names currently does.

PR #292, opened 2019-01-23
cleanup: move code from arch/x86/crc_pclmulqdq.c to crc32.c

sebpop:


Dead2:
Looks good, but updates to win32/ makefiles are missing, would be nice to keep those in sync.
Maybe we should have some automated win32 testing too, since I have no idea whether those things even compile currently.

Dead2:
I my quick testing, this surprised me by showing a 1-10% speedup. Obviously the compiler really likes this.
I'll try to do a more controlled test later.

nmoinvaz:
@Dead2 - Why do you have specific win32 makefiles? Isn't cmake able to build the DLLs?

sebpop:
Thanks @nmoinvaz for the suggestion: #296

sebpop:
Updated dependences in win32/ makefile.

PR #293, opened 2019-01-23
ARM: remove configure flags --neon and --acle

sebpop:
This patch series enables NEON vectorization on ARM processors by default without having to explicitly configure with --neon and --acle.  This makes configure for ARM processors symmetric to the way configure works for x86 processors.  The patches also make the configure flag --without-optimization work for ARM the same way the configure flag works for x86.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,7 +79,6 @@ matrix:
             - gcc-aarch64-linux-gnu
             - libc-dev-arm64-cross
       # For all aarch64 implementations NEON is mandatory, while crypto/crc are not.
-      # So for aarch64 NEON should be automatically enabled, and "acle" should be explicitly defined
       env: CHOST=aarch64-linux-gnu BUILDDIR=. TOOL="./configure --warn --zlib-compat"
     - os: linux
       compiler: aarch64-linux-gnu-gcc
@@ -89,12 +88,6 @@ matrix:
             - qemu
             - gcc-aarch64-linux-gnu
             - libc-dev-arm64-cross
-      # AArch64 toolchain in Trusty is 4.8.2, which doesn't support acle intrinsics.
-      # Until TravisCI provides Xenial dist (which comes with gcc-aarch64-5.3.1) 'acle' option cannot
-      # be enabled in subset. But the command
-      # "CHOST=aarch64-linux-gnu ./configure --warn --zlib-compat --acle"
-      # should work if manually test it in a dist with gcc-aarch64 5.1 or higher
-      # env: CHOST=aarch64-linux-gnu BUILDDIR=. TOOL="./configure --warn --zlib-compat --acle"
       env: CHOST=aarch64-linux-gnu BUILDDIR=. TOOL="./configure --warn --zlib-compat"
     # Hard-float subsets
     - os: linux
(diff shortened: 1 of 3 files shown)


Dead2:
Travis config file needs to get it's --neon flag removed.
Also needs to be rebased.

I looked over the code, and did not see anything objectionable. But you arm-guys would be in a better position to say whether this works properly on the many many arm platforms :)
The simplification looks really good though!

mtl1979:
This is bad... It breaks cross-compilers as those target for armv2 or armv7 by default.

sebpop:
> This is bad... It breaks cross-compilers as those target for armv2 or armv7 by default.

@mtl1979 could you please be more specific about which configurations are broken by the patch?

.travis.yml
```diff
@@ -114,7 +107,7 @@ matrix:
             - qemu
             - gcc-arm-linux-gnueabihf
             - libc-dev-armhf-cross
-      env: CHOST=arm-linux-gnueabihf BUILDDIR=. TOOL="./configure --warn --zlib-compat --neon"
+      env: CHOST=arm-linux-gnueabihf BUILDDIR=. TOOL="./configure --warn --zlib-compat"
```
sebpop: > Travis config file needs to get it's --neon flag removed.

@Dead2 is this change enough to make the Travis CI tester pass, or is there another change that I need?

sebpop: I see that the tester just passed.  I don't know why the tester hasn't picked up this change yesterday.

sebpop:
I changed the patch such that we have configure flags to disable neon and acle.
The default is changed to the more useful configuration with neon and with acle.

mtl1979:
Option to disable is much better than removing the whole option... Now this just needs to be tested on various ARM compilers ́including Linaro toolsets. If we get any regressions, it's easier to just disable the specific support on per case basis and/or issue a warning to user.

sebpop:
Tested both plain `configure` and `configure --without-neon --without-acle`  with the following toolchains:
```
https://releases.linaro.org/components/toolchain/binaries/latest-7/arm-linux-gnueabihf/
https://releases.linaro.org/components/toolchain/binaries/latest-7/armv8l-linux-gnueabihf/
https://releases.linaro.org/components/toolchain/binaries/latest-7/aarch64-linux-gnu/
```
zlib-ng fails to configure with the bare-metal toolchains:
```
https://releases.linaro.org/components/toolchain/binaries/latest-7/arm-eabi/
https://releases.linaro.org/components/toolchain/binaries/latest-7/aarch64-elf/
```
as it detects ARCH=x86_64 instead.

sebpop:
Auto-detection of cross-compilation fails for bare-metal, i.e., when only setting `CC=aarch64-elf-gcc ./configure`.
When also specifying `CHOST=` the following configure command works:
```
CC=.../linaro-7.4.1-2019.02-x86_64_aarch64-elf/usr/bin/aarch64-elf-gcc CHOST=aarch64 ./configure
```
zlib-ng passes configure+make test with `aarch64-elf`.

I still see some undefined symbols when linking with the `arm-eabi` toolchain:
```
libz-ng.so.1.9.9: undefined reference to `__tls_get_addr'
deflate.c:(.text+0x1ec): undefined reference to `__aeabi_read_tp'
```

mtl1979:
gcc flag "-mtp=cp15" should fix the missing "__aeabi_read_tp" symbol if the target processor and compiler supports it. For the incorrectly detected architectures, we might want separate pull request and patch.

sebpop:
The `arm-eabi` toolchain also passes the configure+make test with:
```
CC=.../linaro-7.4.1-2019.02-x86_64_arm-eabi/usr/bin/arm-eabi-gcc CHOST=arm CFLAGS="-mtp=cp15" ./configure --static
```

sebpop:
I pushed a separate patch for the auto-detection of cross-compilers: #305

PR #295, opened 2019-01-24
cleanup: remove unused file

sebpop:


mtl1979:
This is bad...

sebpop:
@mtl1979 could you please be more specific about what you mean?
Is the commit bad?
Or is it bad because we have a file that is unused?

mtl1979:
For both reasons... Obviously the file was meant to be used to quickly find out what intrinsic functions are available on specific compiler versions but rationale for including it was never documented.

Dead2:
No, it finds what intrinsics are referenced from source files, making it easier to figure out what SSE/AVX etc versions are required by different source files.
In its current form, it likely has little value, but in a more developed form, it could possibly be helpful for code reviews (in any kind of code, not just zlib-ng).
I do not object to its removal.

PR #296, opened 2019-01-24
cleanup: remove win32/

sebpop:


nmoinvaz:
Nice work. I think the windows resource files (.rc) can be generated in cmake as well.
https://stackoverflow.com/questions/6693100/how-to-generate-windows-dll-versioning-information-with-cmake

nmoinvaz:
On second inspection it looks like it might not be that straight forward. The only other suggestion I can make is rename `win32-zlib*.rc` to just `zlib*.rc`. I think the root directory would look cleaner that way.

mtl1979:
This is bad... We have configure/make and cmake for *nix systems and nmake/cmake for Windows, because cmake is not installed by default. It can be installed as optional for Visual C++ 2015 and 2017 though.

nmoinvaz:
I don't understand why you would put effort into maintaining two different makefiles. It is not that difficult to say for Windows users that you must download cmake. Also, I think most Windows developers are not going to bother with the old nmake makefiles - they'll use cmake to generate the Visual Studio solution files.

mtl1979:
I have both nmake and cmake installed, but I use nmake because it's less buggy. I have upgraded cmake to latest available version whenever I need to start working on a project that doesn't have files for nmake just to see if some or all of the bugs have been fixed. It has been already pointed out in other pull requests that cmake evolves too quickly and the files for it need to be constantly updated to make sure everything works as smoothly as possible. nmake files only need updating if we add new files or change preprocessor macros.

sebpop:
I don't have a preference on keeping or removing the win32 makefiles.
Either way, I will follow up with patches to either fix the existing makefiles or remove them.
However I don't have a system where to test the changes.

mtl1979:
That's why it has been so far my responsibility to check that files in that directory still work. Not having a system to test submitted patches is not valid reason to remove files, it's actually prerequisite to have system to test the effect of the patch.

Dead2:
I'd like to keep these files, I know for a fact that the dependencies are out of date, but I have no idea whether it still compiles however.
I would like to get one of those CI tools to verify that they keep working. That would require someone to set up a config file, test it, and then make a PR with it.

nmoinvaz:
If cmake format changes that much then why use it in the first place for the other platforms. I think that if it is good enough for the other platforms it should be good enough for Windows too. Having two different makefiles requires more work to support and maintain. I mean there are probably %0.1 of Windows developers who are going to actually be using nmake. There isn't just one nmake makefile, there are two, one for each architecture. With cmake you get all the added benefits of everything being handled for you - it figures out your architecture and the source files you'll need among other things. Also, with cmake you can probably generate nmake makefiles. So in the slim chance somebody will need nmake they can just use cmake to generate them. That's just my 2 cents.

@Dead2, Makefile.msc fails to compile:

```
deflate.obj : error LNK2001: unresolved external symbol _deflate_medium
zlib-ng1.dll : fatal error LNK1120: 1 unresolved externals
```
I don't have an ARM machine in order to test Makefile.arm.

mtl1979:
We didn't remove cmake support after we forked from zlib even though it was seriously broken. Instead we gradually fixed it and merged MinGW/Cygwin support to it. We intentionally didn't merge Visual C++ support to it because Visual Studio 2017 wasn't finished yet so no compiler suite on native Windows did have cmake support. At that time 50% of the development team did run zlib-ng on Windows, so it was a lot more than 0.1% of possible users.

donny-dont:
The ninja generator and visual studio generator work fine through CMake on Windows. We use zlib-ng for the WinCairo port of WebKit and build it through [vcpkg](https://github.com/Microsoft/vcpkg).

https://github.com/WebKitForWindows/WinCairoRequirements/tree/master/ports/zlib

There are a couple minimal patches in there to make it conform to what vcpkg wants but overall it works. I don't see any reason to bother with nmake.

mtl1979:
@donny-dont The whole point is that we shouldn't require external programs for the build. vcpkg requires minimum of VS2015 with latest patches, but our minimum is still VS2013 without any patches. Using nmake Makefiles is better than storing Visual Studio project files on the repository as each version uses slightly different format and/or contents due to added features.

donny-dont:
@mtl1979 I'm just noting the fact that if you're building on Windows CMake works perfectly in both a Ninja and Visual Studio configuration. I don't personally see any need to use NMake as someone who primarily develops on Windows.

sebpop:
@Dead2 said:
> I'd like to keep these files,

I am closing this pull request and I will try to update the win32/Makefiles.  I will need some help testing that patch on win32.

@nmoinvaz thanks for the windows error report, please try with #299.

PR #297, opened 2019-01-25
ARM: check cpu feature once at init time

sebpop:
This makes the checks for arm cpu features as inexpensive as on the x86 side
by calling the runtime feature detection once in deflate/inflate init and then
storing the result in a global variable.


mtl1979:
Good idea but fails the build tests...

PR #298, opened 2019-01-25
zlib adler_simd.c

sebpop:
Add SSSE3 implementation of the adler32 checksum, suitable for
both large workloads, and small workloads commonly seen during
PNG image decoding. Add a NEON implementation.

Speed is comparable to the serial adler32 computation but near
64 bytes of input data, the SIMD code paths begin to be faster
than the serial path: 3x faster at 256 bytes of input data, to
~8x faster for 1M of input data (~4x on ARMv8 NEON).

For the PNG 140 image corpus, PNG decoding speed is ~8% faster
on average on the desktop machines tested, and ~2% on an ARMv8
Pixel C Android (N) tablet, https://crbug.com/762564#c41

Update x86.{c,h} to runtime detect SSSE3 support and use it to
enable the adler32_simd code path and update inflate.c to call
x86_check_features(). Update the name mangler file names.h for
the new symbols added, add FIXME about simd.patch.

Ignore data alignment in the SSSE3 case since unaligned access
is no longer penalized on current generation Intel CPU. Use it
in the NEON case however to avoid the extra costs of unaligned
memory access on ARMv8/v7.

NEON credits: the v_s1/s2 vector component accumulate code was
provided by Adenilson Cavalcanti. The uint16 column vector sum
code is from libdeflate with corrections to process NMAX input
bytes which improves performance by 3% for large buffers.

Update BUILD.gn to put the code in its own source set, and add
it conditionally to the zlib library build rule. On ARM, build
the SIMD with max-speed config to produce the smallest code.

No change in behavior, covered by many existing tests.

Bug: http://crbug.com/762564
Reviewed-on: https://chromium-review.googlesource.com/660019
Commit-Queue: Chris Blume <cblume@chromium.org>
Reviewed-by: Adenilson Cavalcanti <cavalcantii@chromium.org>
Reviewed-by: Chris Blume <cblume@chromium.org>


sebpop:
This is a preliminary port of the chromium patch for adler32 vectorization on x86 with SSSE3.
I would appreciate comments on the cmake, configure+Make system.

The patch brings in a threshold (<= 64 bytes) that chromium developers use to continue using the C version of adler32 and thus avoid an unprofitable use of the SIMD version of adler32 on both x86 and arm cpus.

I still need to review the performance of our NEON code for adler32 with respect to what chromium uses.

arch/x86/adler32_ssse3.c
```diff
@@ -0,0 +1,192 @@
+/* adler32_simd.c
+ *
+ * Copyright 2017 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the Chromium source repository LICENSE file.
+ *
```
Dead2, line 6: This states it is licensed under a BSD-license. It would need to be re-licensed to zlib if we are to use this.

sebpop: I sent an email to the author of the patch Noel Gordon to ask whether he would consider re-licensing the patch under the zlib license.
@Dead2 I added your linkedin email address in CC to that conversation.

adler32.c
```diff
@@ -62,11 +61,24 @@ static uint32_t adler32_combine_(uint32_t adler1, uint32_t adler2, z_off64_t len
...
+
 /* ========================================================================= */
-uint32_t adler32_c(uint32_t adler, const unsigned char *buf, size_t len) {
+uint32_t ZLIB_INTERNAL adler32_c(uint32_t adler, const unsigned char *buf, size_t len) {
     uint32_t sum2;
     unsigned n;
 
+#if defined(X86_SSSE3_ADLER32)
+    if (buf && len >= 64 && x86_cpu_has_ssse3)
+        return adler32_ssse3(adler, buf, len);
```
sebpop, line 76: Here is the change similar to the one under review in #303.

arch/x86/adler32_ssse3.c
nmoinvaz: I emailed @Adenilson and he said to reach out via the Chromium bug tracker for re-licensing.

nmoinvaz: https://bugs.chromium.org/p/chromium/issues/detail?id=1009797

adler32.c
mtl1979: The test for x86_cpu_has_ssse3 should be first so it does not even try other tests if that fails. It makes no sense for me to do two unnecessary tests if third fails.

arch/x86/adler32_ssse3.c
mtl1979: Mixing licenses is in my opinion always a bad thing due to viral nature of copyright law. If ever our code is merged with more permissive license, essentially we are giving permission for others to use the whole source using that license. On the other hand, if we merge code with more restrictive license, it might require us to remove some existing code or forbid other people linking against zlib-ng due to their code having incompatible license.

This is different if code is linked statically or dynamically as the general rule is that one binary must have only one license that covers all source files. Essentially when code is dual-licensed, the author is creating new license that combines the two licenses as such that more permissive license overrides conflicting requirements.

nmoinvaz: The bug report has been closed. It looks like chromium developers don't see the necessity or want to relicense that bit of source code under zlib license.

mtl1979: Then that code is clearly no-go...

sebpop:
Closing because of incompatible license.

PR #299, opened 2019-01-26
win32: add missing makefile dependence

sebpop:


Dead2:
I am unable to test this myself

nmoinvaz:
Both Makefile.msc and Makefile.arm compile without a problem. 

_Edit:_ I previously had problems compiling Makefile.arm because I wasn't doing distclean between compilations.

PR #301, opened 2019-01-28
Remove functable

sebpop:
This series of patches intends to reduce the number of indirect jumps (which are more costly than conditional jumps) by removing the function pointers in functable.  This narrows the performance gap that I see when comparing against chromium/zlib.


mtl1979:
This is broken and bad idea... This essentially makes zlib-ng test for correct optimization every single time the specific function is called. With functable, even though it uses indirect jump, the check is done only once per process.

sebpop:
I am closing this pull request as I agree with @mtl1979 that the indirect branches with only one run time target are cheaper than a load from a global variable followed by a direct call.

Looking at the performance runs on x86_64 I see several regressions, the case that I investigated below seems to be related to an increase in register spilling.

base: gcc -O3
```
alice29.txt : ZLIB -2: [b 1M] bytes 148481 ->  60245 40.6% comp  89.6 ( 89.7) MB/s uncomp 347.6 (347.7) MB/s
```
compression speed degrades by 2% with the patches to remove indirect jumps:
```
alice29.txt : ZLIB -2: [b 1M] bytes 148481 ->  60245 40.6% comp  86.2 ( 86.2) MB/s uncomp 351.5 (351.7) MB/s
```
`valgrind --tool=cachegrind --branch-sim=yes ./benchmark; cg_annotate --auto=yes cachegrind.out` reports the number of executed instructions (Ir), data reads (Dr), writes (Dw), conditional branches (Bc), and indirect branches (Bi):

base:
```
           Ir  I1mr ILmr          Dr       D1mr  DLmr          Dw      D1mw   DLmw          Bc        Bcm         Bi       Bim  file:function
  850,647,600    35   19 299,944,080    522,291    11 151,206,480   605,804  4,552  79,333,200  8,321,468 20,810,160         5  deflate_fast.c:deflate_fast
```
patch to remove indirect jumps:
```
           Ir I1mr ILmr          Dr       D1mr  DLmr          Dw      D1mw   DLmw          Bc        Bcm         Bi       Bim  file:function
  735,682,320   75   18 273,087,720    537,770    11 135,657,360   605,445  4,552  79,333,200  8,344,879          0         0  deflate_fast.c:deflate_fast
  210,021,840   11    3  54,303,480        360     0  36,255,240       359      0  20,782,800        776          0         0  deflate_p.h:deflate_fast
```
We see that the 20M indirect branches were transformed into conditional branches. The total number of instructions goes up from 850M to 735+210=945M by about 100M instructions. Looking at  `insert_string` in `deflate_p.h`:
```

         Ir I1mr ILmr         Dr D1mr DLmr         Dw D1mw DLmw         Bc Bcm Bi Bim 
          .    .    .          .    .    .          .    .    .          .   .  .   .  #ifdef X86_SSE4_2_CRC_HASH
 62,354,520    1    1 20,783,880  360    0          0    0    0 20,785,320 780  0   0      if (x86_cpu_has_sse42)
147,637,440  377    4 33,526,800    0    0 36,231,840    0    0          0   0  0   0          return insert_string_sse(s, str, count);
```
we see that there are 33M more reads and 36M more writes. The code of `deflate` where this function gets inlined has a lot of spills and this change may aggravate the problem.

PR #303, opened 2019-02-01
 only call NEON adler32 for more than 16 bytes

sebpop:


mtl1979:
This is broken...

Dead2:
Needs rebase :)

Dead2:
I currently have no benchmarks that use adler32, could you post some chromium-bench before/after benches please?
I see no regressions with minigzip at least.

mtl1979:
This will cause unnecessary revert when we add PowerPC and other architectures that support vectors... We need special case in the stub initialization for arm/aarch64 path...

Dead2:
I am wondering whether special-casing len < 16 only for arm is necessary?
Did you test whether this could also be beneficial for other arches?
I guess neon might be more expensive to setup, but it seems like a thing that should be explored.

sebpop:
> I am wondering whether special-casing len < 16 only for arm is necessary?
> Did you test whether this could also be beneficial for other arches?
> I guess neon might be more expensive to setup, but it seems like a thing that should be explored.

There is a similar change with `len >= 64` in the patch that vectorizes adler32 on x86 with ssse3: https://github.com/zlib-ng/zlib-ng/pull/298/commits/938cc7e25d5c349d4e828bddf978407bf966db7a

Dead2:
I am currently against removing adler32 from functable.

However, could this not also be solved by having the adler32-neon call the adler32-c function if the length is smaller than 64? That should be a pretty small and simple change, whether it is still a performance win is the question however.

sebpop:
The patch follows the comments from @Dead2 by adding to the neon implementation of `adler32` the scalar computation of adler32 for short strings.  The impact is about 0.3% better performance on A73 overall the chromium benchmark running on the snappy data set.

I am not sure where to move the "duplicated code" that fails in CodeFactor: maybe we can move that to a new file `adler32.h`?

mtl1979:
> I am not sure where to move the "duplicated code" that fails in CodeFactor: maybe we can move that to a new file `adler32.h`?

Should be adler32_p.h as it would be private header...

arch/arm/adler32_neon.c
```diff
@@ -79,16 +79,59 @@ static void NEON_handle_tail(uint32_t *pair, const unsigned char *buf, size_t le
...
+            sum2 += adler;
+        }
+        if (adler >= BASE)
+            adler -= BASE;
+        MOD28(sum2);            /* only added so many BASE's */
+        return adler | (sum2 << 16);
+    }
+
     /* The largest prime smaller than 65536. */
     const uint32_t M_BASE = 65521;
```
mtl1979: ```BASE``` and ```M_BASE``` are for same purpose, and as such duplicates.

Dead2:
@sebpop Could you run the perf benchmark on x86_64 too?

sebpop:
I just finished two runs of the chromium benchmark on x86-64: there is no performance difference with the patch.

Dead2:
Thanks for running the benchmark, no unintended regressions is important :)

PR #304, opened 2019-02-01
enable INFFAST_CHUNKSIZE on x86

sebpop:
speeds up inflate on x86 by up to 85% on an i7-4790K CPU @ 4.00GHz.


mtl1979:
This is broken...

Dead2:
Looks pretty good, however this is a quite big change and splitting it up a bit would be nice.
Please split the -DX86_SSE2_FILL_WINDOW change into a separate commit before or after the rest (it can still be part of the same PR).

I have not done any benchmarking of this yet, (have been sick and very busy lately) but it is on my prio list for next week.

Also, I would very much like to talk to you in #zlib-ng on irc.freenode.net sometime @sebpop

sebpop:
Ok, I will split this patch up.

I have another improvement for aarch64 for the case the memset is of size 3 and 6: https://github.com/sebpop/zlib-ng/commit/24e4324b4a645e64f00473575c3af5e71915ce03
that I will be adding to the memset patch.
One of my colleagues remarked that for image decompression the case 3 is one of the most frequent
and that the patch speeds up considerably inflate.  In my testing of the snappy data, there were almost no case 3 and 6: that just shows that the chromium benchmark does not contain image decompression.
They may want to add a raw image to the benchmark such that it gets compressed as a png and then decompressed as png compressed data.

sebpop:
I split up the patch from general cleanup changes in #310.  Closing this PR.

PR #309, opened 2019-03-01
 rename X86_SSE4_2_CRC_HASH to X86_SSE4_2

sebpop:


CMakeLists.txt
```diff
@@ -562,17 +553,14 @@ if(WITH_OPTIM)
...
-            if(HAVE_SSE42CRC_INTRIN)
-                add_definitions(-DX86_SSE4_2_CRC_INTRIN)
-            endif()
+            add_feature_info(SSE42 1 "Use the SSE4.2 instruction set, using \"${SSE42FLAG}\"")
+            add_intrinsics_option("${SSE42FLAG}")
             if(WITH_NEW_STRATEGIES)
                 add_definitions(-DX86_QUICK_STRATEGY)
                 set(ZLIB_ARCH_SRCS ${ZLIB_ARCH_SRCS} ${ARCHDIR}/deflate_quick.c)
-                add_feature_info(SSE4DEFLATE 1 "Support SSE4.2-accelerated quick decompression")
+                add_feature_info(SSE42_DEFLATE 1 "Support SSE4.2-accelerated quick decompression")
```
Dead2, line 563: I am thinking this should rather be renamed DEFLATE_QUICK for example, since we don't currently have SSE4.2 features in any other deflate strategy, but we might in the future. Deflate-quick is a feature, not an optimization in the existing deflate strategy.

Also, this says decompression, but deflate is compression.

Dead2:
This PR needs more description, it does not only rename a define, it combines two checks and defines into one. That potentially changes how it functions.

Also, if we do this, I'd like to see the define default to enabled and only be needed to disable the SSE4.2 support. That makes sure people using zlib-ng as a subproject without running our cmake or configure scripts get the optimized version by default. If they get compile failures, they should be able to figure out how to either run them or set the necessary defines manually.

sebpop:
I am inclined to abandon this change: it is not really needed for the inffast_chunk patch and it only changes define names to avoid mentioning `crc_hash` as the sse42 instructions may be used in other components of zlib that have nothing to do with hashing.

I will fix the typo in the cmake file.  #325

PR #310, opened 2019-03-01
enable INFFAST_CHUNKSIZE on x86

sebpop:
This is a split up series of patches from https://github.com/zlib-ng/zlib-ng/pull/304


Dead2:
This is a big PR, and renaming chunk_copy to chunk_copy_1 caused me some confusion, until I saw it was renamed back again in the next commit. Would you be willing to clean that up please?

Other than that, things are looking good I think.
This speeds up inflate by ~9.7% on my x86_64 machine.

memcopy.h
```diff
@@ -22,6 +22,12 @@ typedef uint8x16_t inffast_chunk_t;
 #  define INFFAST_CHUNKSIZE sizeof(inffast_chunk_t)
 #endif
 
+# if defined(X86_SSE2)
```
nmoinvaz: May need to fix whitespacing here. # if contains a space between the two. And also at endif.

Dead2: It does look like correct indentation for a nested if is used from this line and down, but the two if blocks just above here break the pattern and make it hard to understand, they should also preferably be indented so we can easily see where this block starts (at MEMCOPY_H in line 4).

Dead2: Honestly, I personally prefer to indent the # sign as well, instead of leaving it at the beginning of the line. That makes the code easier to read at a glance. I use syntax highlighting, so I still see the preprocessor macros are there.

Dead2:
needs a rebase

PR #311, opened 2019-03-01
fix all ASan errors on aarch64

sebpop:
when running `./configure --with-sanitizers --with-fuzzers && make && make test`
there were several errors of the form:

arch/aarch64/insert_string_acle.c:35:13: runtime error: load of misaligned address 0xffff90614801 for type 'unsigned int'

This patch fixes all those errors by using the MEMCPY builtin.

### DIFF
--- a/match_p.h
+++ b/match_p.h
@@ -192,7 +192,7 @@ static inline unsigned longest_match(deflate_state *const s, IPos cur_match) {
         chain_length >>= 2;
 
     /*
-     * Do not looks for matches beyond the end of the input. This is
+     * Do not look for matches beyond the end of the input. This is
      * necessary to make deflate deterministic
      */
     nice_match = (unsigned int)s->nice_match > s->lookahead ? s->lookahead : s->nice_match;
@@ -205,8 +205,8 @@ static inline unsigned longest_match(deflate_state *const s, IPos cur_match) {
 
     scan = s->window + s->strstart;
     strend = s->window + s->strstart + MAX_MATCH - 1;
-    scan_start = *(uint16_t *)scan;
-    scan_end = *(uint16_t *)(scan + best_len-1);
+    memcpy(&scan_start, scan, sizeof(scan_start));
+    memcpy(&scan_end, scan + best_len - 1, sizeof(scan_end));
 
     Assert((unsigned long)s->strstart <= s->window_size - MIN_LOOKAHEAD, "need lookahead");
     do {
@@ -226,9 +226,13 @@ static inline unsigned longest_match(deflate_state *const s, IPos cur_match) {
          * is limited to the lookahead, so the output of deflate is not
          * affected by the uninitialized values.
          */
-        if (likely((*(uint16_t *)(match + best_len - 1) != scan_end)))
+        uint16_t val;
+        memcpy(&val, match + best_len - 1, sizeof(val));
+        if (likely(val != scan_end))
             continue;
-        if (*(uint16_t *)match != scan_start)
+
+        memcpy(&val, match, sizeof(val));
+        if (val != scan_start)
             continue;
 
         /* It is not necessary to compare scan[2] and match[2] since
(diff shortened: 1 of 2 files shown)


Dead2:
I am currently unable to benchmark this for any performance regressions (my arm machines need some work), could you show some benchmarks before and after please?

match_p.h
```diff
@@ -246,11 +250,37 @@ static inline unsigned longest_match(deflate_state *const s, IPos cur_match) {
...
+            scan += sizeof(sval);
+
+            MEMCPY(&mval, match, sizeof(mval));
+            MEMCPY(&sval, scan, sizeof(sval));
+            if (mval != sval)
+              break;
+            match += sizeof(mval);
+            scan += sizeof(sval);
+
+        } while (scan < strend);
```
mtl1979, line 282: This introduces unnecessary code duplication... It should compare the memory locations directly using largest possible data type or use only one pair of MEMCPY per iteration...

arch/aarch64/crc32_acle.c
```diff
@@ -15,66 +15,80 @@
...
-    if ((len > 2) && ((ptrdiff_t)buf & 2)) {
-        buf2 = (const uint16_t *) buf;
-        c = __crc32h(c, *buf2++);
-        len -= 2;
-        buf4 = (const uint32_t *) buf2;
-    } else {
-        buf4 = (const uint32_t *) buf;
+    if ((len > sizeof(uint16_t)) && ((ptrdiff_t)buf & 2)) {
+        uint16_t val;
+        MEMCPY(&val, buf, sizeof(val));
```
mtl1979: This seems unnecessary as buf is guaranteed to be aligned...

arch/arm/crc32_acle.c
```diff
@@ -15,53 +15,77 @@
...
-    if ((len > 2) && ((ptrdiff_t)buf & 2)) {
-        buf2 = (const uint16_t *) buf;
-        c = __crc32h(c, *buf2++);
-        len -= 2;
-        buf4 = (const uint32_t *) buf2;
-    } else {
-        buf4 = (const uint32_t *) buf;
+    if ((len > sizeof(uint16_t)) && ((ptrdiff_t)buf & 2)) {
+        uint16_t val;
+        MEMCPY(&val, buf, sizeof(val));
```
mtl1979: Like above, this seems unnecessary as buf is guaranteed to be aligned.

arch/aarch64/crc32_acle.c
```diff
@@ -15,66 +15,80 @@
...
     }
 
-    if ((len > 2) && ((ptrdiff_t)buf & 2)) {
-        buf2 = (const uint16_t *) buf;
-        c = __crc32h(c, *buf2++);
-        len -= 2;
-        buf4 = (const uint32_t *) buf2;
-    } else {
-        buf4 = (const uint32_t *) buf;
+    if ((len > sizeof(uint16_t)) && ((ptrdiff_t)buf & 2)) {
```
mtl1979: This is wrong... ```buf & 2``` should be ```buf & sizeof(uint16_t)```.

arch/aarch64/crc32_acle.c
```diff
@@ -15,66 +15,80 @@
...
 uint32_t crc32_acle(uint32_t crc, const unsigned char *buf, uint64_t len) {
-    register uint32_t c;
-    register const uint16_t *buf2;
-    register const uint32_t *buf4;
-    register const uint64_t *buf8;
+    uint32_t c = ~crc;
 
-    c = ~crc;
     if (len && ((ptrdiff_t)buf & 1)) {
         c = __crc32b(c, *buf++);
```
mtl1979: This aligns ```buf``` to 2-byte boundary.

arch/aarch64/crc32_acle.c
```diff
@@ -15,66 +15,80 @@
...
         len--;
     }
 
-    if ((len > 2) && ((ptrdiff_t)buf & 2)) {
-        buf2 = (const uint16_t *) buf;
-        c = __crc32h(c, *buf2++);
```
mtl1979: This aligns ```buf2``` and ```buf4``` below to 4-byte boundary.

arch/aarch64/crc32_acle.c
```diff
@@ -15,66 +15,80 @@
...
+    if ((len > sizeof(uint16_t)) && ((ptrdiff_t)buf & 2)) {
+        uint16_t val;
+        MEMCPY(&val, buf, sizeof(val));
+        c = __crc32h(c, val);
+        buf += sizeof(val);
+        len -= sizeof(val);
     }
 
-    if ((len > 4) && ((ptrdiff_t)buf & 4)) {
-        c = __crc32w(c, *buf4++);
```
mtl1979: This aligns ```buf4``` to 8-byte boundary.

arch/aarch64/crc32_acle.c
```diff
@@ -15,66 +15,80 @@
...
-        buf4 = (const uint32_t *) buf;
+    if ((len > sizeof(uint16_t)) && ((ptrdiff_t)buf & 2)) {
+        uint16_t val;
+        MEMCPY(&val, buf, sizeof(val));
+        c = __crc32h(c, val);
+        buf += sizeof(val);
+        len -= sizeof(val);
     }
 
-    if ((len > 4) && ((ptrdiff_t)buf & 4)) {
```
mtl1979: I think this should have ```buf4 & 4``` at the end instead... See line 37 of original code.

sebpop: Both `buf & 4` and `buf4 & 4` are equivalent:
on line 37 of original code `buf4 = buf`,
on line 35 of original code `buf4 = buf2` and on lines 32 and 33 we have `buf2 = buf + sizeof(uint16_t)`.
The bit we are looking at with `buf & 4` or with `buf4 & 4` is the same.

sebpop: I agree that it is less confusing to read `buf4 & 4`.  I will prepare a separate patch.

arch/aarch64/crc32_acle.c
sebpop: This and the other comments below are now fixed in the new patch.
The new patch minimizes the number of changes: it keeps all the auxiliary pointers buf2, buf4, and buf8.

match_p.h
sebpop: I will fix this in a follow-up patch that will change the performance.
We will need to special case big/little endian when iterating through the bytes of the loaded `uint64_t`.

arch/aarch64/crc32_acle.c
sebpop: Fixed in #322.

match_p.h
mtl1979: Using oprimized memcmp() you wouldn't need to byte swap anything... Basically compiler will automatically use temporary variable correctly,

Dead2:
Please rebase

PR #312, opened 2019-03-02
remove redundant configure time flag

sebpop:
only rely on compiler defined __ARM_FEATURE_CRC32


mtl1979:
Obviously this flag is not redundant as we need to have flag to disable features even if compiler supports it. This is because compiler default might be higher than the target processor, especially when doing validity tests.

sebpop:
You are right, we need the flag to disable the feature.  Closing PR.

PR #313, opened 2019-03-02
replace use of ptrdiff_t with unsigned long

sebpop:


mtl1979:
This is bad... "unsigned long" is not guaranteed to be same size as pointer.

sebpop:
We don't need to have the exact same size as a pointer: even a shorter type would work as we are only interested in the lower bits for the alignment test. Also see the discussion about casting a pointer to `unsigned long` for the alignment test in https://stackoverflow.com/questions/1898153/how-to-determine-if-memory-is-aligned

mtl1979:
@sebpop The thing is that the code would be invalid as you're only allowed to cast a pointer to integer of same size, anything else would cause compiler trap. When you cast pointer to anything smaller, you are actually required to cast it to widest integer size first and then zero the highest N bits. This is because there is no assembler instruction to truncate pointers (as in clearing highest N bits without making copy). Even if variable uses less than maximum number of bits, the generated assembler code will use all available bits, but explicitly zero unused bits.

When you truncate normal integers, you are actually making a copy of the lower bits of the value and then zero- or sign-extending the register.

sebpop:
Thanks @mtl1979 for the review.
I am closing the PR as the benefit of the cleanup does not compensate for the cost of introducing bugs.

PR #316, opened 2019-03-04
factor out duplicated code reported by codefactor

sebpop:


mtl1979:
Instead of moving architecture specific code out of "arch" tree, we should merge the "arm" and "aarch64" directories and combine the differences using preprocessor tests...

sebpop:
I was also thinking about merging `arch/{arm,aarch64}` into one dir `arch/arm` instead of moving the files at the top level.
I will send a patch.

mtl1979:
Originally we kept the two as own directories because there was different flags in different compiler versions for the features and testing all of them would have made the lines too long and confusing but maintaining both directories doesn't really make sense anymore...

sebpop:
Closing this PR, refactored code in #317.

PR #317, opened 2019-03-04
 factor out code in arch/{arm,aarch64}

sebpop:


arch/arm/crc32_acle.c
```diff
@@ -5,17 +5,36 @@
  *
 */
 
-#ifdef __ARM_FEATURE_CRC32
-#include <arm_acle.h>
-#ifdef ZLIB_COMPAT
+#if __ARM_FEATURE_CRC32
```
Dead2: Was the change from ifdef to if intentional? Or was it supposed to be if defined()?

Dead2:
This refactor commit is mixing several independent tasks, making it hard to review or bisect.
Could you split it into preferably three commits?
1. Cleanups that do not change functionality, such as the #ifdef to #if changes. Also curious of why? Also the cmakelists.txt changes that change x86 stuff should go here.
2. Refactor crc32 code
3. Unify aarch64 and arm codebases into one

arch/arm/crc32_acle.c
sebpop: This change was intentional to make it look as the other uses of that compiler defined macro.
It can be used both ways, either as `#if __ARM_FEATURE_CRC32` or `#if defined(__ARM_FEATURE_CRC32)` as both gcc and clang have it defined as `1` when compiling with `-march=armv8-a+crc`:
```
$ arm-linux-gnueabihf-gcc -march=armv8-a+crc -x c /dev/null -dM -E | grep __ARM_FEATURE_CRC32
#define __ARM_FEATURE_CRC32 1
```
and it is not defined when `-march` does not include crc:
```
$ arm-linux-gnueabihf-gcc -march=armv8-a -x c /dev/null -dM -E | grep __ARM_FEATURE_CRC32 | wc -l
0
```

Dead2: I guess that is fine then. My only worry is that it implies that we care about the value and not just whether it has been defined. For readability, I think just checking whether it has been defined would be more readable and easier to understand, but if you think this is better, I can be swayed.

Dead2:
Looks good, but needs a rebase.
There are a few conflicts between the various PRs now, so please tell me if you want them pulled in a special order. Currently I am thinking that this one might be the first arm-related PR to pull, but if that makes it hard to rebase something else, please tell me.

PR #319, opened 2019-03-05
unify uses of __ARM_FEATURE_CRC32

sebpop:


functable.c
```diff
@@ -111,7 +111,7 @@ ZLIB_INTERNAL uint32_t crc32_stub(uint32_t crc, const unsigned char *buf, uint64
     if (sizeof(void *) == sizeof(ptrdiff_t)) {
 #if BYTE_ORDER == LITTLE_ENDIAN
       functable.crc32=crc32_little;
-#  if __ARM_FEATURE_CRC32 && defined(ARM_ACLE_CRC_HASH)
+#  if defined(__ARM_FEATURE_CRC32) && defined(ARM_ACLE_CRC_HASH)
```
Dead2, line 114: This one I feel is warranted, checking whether it is defined is simpler (for readability) than checking the value of the define is (where a reviewer/coder would need to understand whether the value actually matters).

The change of "ifdef" to "if defined()" is more questionable, it will compress better (the source code) but I honestly don't see any other benefit of this. Unless we wanted to change all "ifdef" blocks in the code that is.
Unless there is a compelling reason, I'd prefer to keep ifdef for the cases where we only check a single define. It is shorter and I feel it is more readable.

sebpop: Agreed.
Fixed.

PR #321, opened 2019-03-05
use memcpy instead of MEMCPY

sebpop:
MEMCPY stands for __builtin_memcpy when that is supported by the compiler.
This patch systematically changes calls to memcpy to MEMCPY when the size of
the copied memory is known at compile time. This allows the compilers to
better optimize the code.

### DIFF
--- a/arch/x86/deflate_quick.c
+++ b/arch/x86/deflate_quick.c
@@ -152,7 +152,7 @@ static inline void quick_send_bits(deflate_state *const s,
     s->bi_valid = width - (bytes_out * 8);
 
     /* Taking advantage of the fact that LSB comes first, write to output buffer */
-    MEMCPY(s->pending_buf + s->pending, &out, sizeof(out));
+    memcpy(s->pending_buf + s->pending, &out, sizeof(out));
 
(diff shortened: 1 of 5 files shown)


Dead2:
The change of the hardcoded builtin_memcpy is something I fully agree on, we should not hardcode the builtin.

The concept of changing from stock memcpy in general however..?
IIRC, GCC will auto-replace memcpy calls with its builtin if:
- size is known
- size is small enough to profit from it
- optimization is enabled
- compiler builtins are enabled

So I don't see why we hardcode the builtin? Does LLVM not do the same?
I made an experiment where i benchmarked this patch and one where I override MEMCPY and MEMSET in zutil.h to be the stock ones always, and even with several rounds of testing each I am unable to measure any performance change at all. If there is one, it is below 0.1% (that is approx what my benchmarks can vary from run to run).
Even when only looking at the fastest run out of the 490 of each, stock memcpy/memcmp is the winner (by an extremely minimal margin).

sebpop:
You are right.
I will send an amended patch that replaces MEMCPY with memcpy and that also fixes the __builtin_memcpy.

PR #322, opened 2019-03-05
 use sizeof(type) instead of magic numbers

sebpop:


gdevenyi:
Accidentally includes #311 and #317

arch/arm/crc32_acle.c
```diff
@@ -5,39 +5,88 @@
...
         c = __crc32b(c, *buf++);
         len--;
     }
 
-    if ((len > 2) && ((ptrdiff_t)buf & 2)) {
-        buf2 = (const uint16_t *) buf;
-        c = __crc32h(c, *buf2++);
+    if ((len > sz2) && ((ptrdiff_t)buf & sz2)) {
+        uint16_t val;
+        MEMCPY(&val, buf, sizeof(val));
```
mtl1979: I still don't see any valid reason to use MEMCPY() as buf is guaranteed to be always aligned.

nmoinvaz: Why are you using sz2 and sz4 variables? Is there some performance improvement over just using sizeof() in all those places?

sebpop: I fixed both issues.  Thanks.

Dead2:
The sizeof commit looks good. Will pull when the dependencies are in.

Dead2:
needs a rebase

Dead2:
Please rebase

PR #323, opened 2019-03-06
Change force-sse2 behavior.

Dead2:
Default sse2 to be available on x86_64 arch.
Allow x86 arch to optionally force sse2 availability.
It still depends on sse2 intrinsics being detected and optimizations enabled.

PR #324, opened 2019-03-06
Update x86 and x86_64 arch checks

Dead2:
Update x86 and x86_64 arch checks to use the
recommended define names, resulting in improved compiler support.
Based on the overviews from several sites, such as:
http://nadeausoftware.com/articles/2012/02/c_c_tip_how_detect_processor_type_using_compiler_predefined_macros


Dead2:
This depends on #323, since that carries a similar change to one of the places I did not cover here.

sebpop:
Looks good to me.

PR #327, opened 2019-03-26
Fix building with gcc 8.2.1 and -Wall -Wextra -pedantic -Werror

iii-i:
* ptrdiff_t check always failed because of unused parameter
* sizeof(void *) check always failed because of double semicolon
* Sign issue in nice_match assignment
* dist parameter of set_bytes may be unused
* Parameters of main may be unused in test/example.c
* snprintf requires a _POSIX_C_SOURCE #define in test/minigzip.c,
  because a _POSIX_SOURCE #define is present


sebpop:
Looks good to me.

PR #328, opened 2019-03-26
Fix endianness detection in memcopy.h

iii-i:
When memcopy.h is included into inffast.c, endianness-related
preperocessor defines are not set, which leads to
BYTE_ORDER == LITTLE_ENDIAN condition being always true. This breaks
decompression at least on s390x.


sebpop:
Looks good to me.  Thanks!

PR #329, opened 2019-03-26
Increase verbosity required to warn about bit length overflow.

iii-i:
This is a cherry-pick of https://github.com/madler/zlib/commit/21c66cd5ac8876f48a19882074933c6275eaa022.
I found it useful to have when debugging zlib-ng.

---

When debugging the Huffman coding would warn about resulting codes
greater than 15 bits in length. This is handled properly, and is
not uncommon. This increases the verbosity of the warning by one,
so that it is not displayed by default.

### DIFF
--- a/trees.c
+++ b/trees.c
@@ -449,7 +449,7 @@ static void gen_bitlen(deflate_state *s, tree_desc *desc) {
     if (overflow == 0)
         return;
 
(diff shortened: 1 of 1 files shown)


sebpop:
Looks good to me.

PR #330, opened 2019-03-26
Add support for IBM Z hardware-accelerated deflate

iii-i:
Corresponding zlib pull request: https://github.com/madler/zlib/pull/410

---

Future versions of IBM Z mainframes will provide DFLTCC instruction,
which implements deflate algorithm in hardware with estimated
compression and decompression performance orders of magnitude faster
than the current zlib-ng and ratio comparable with that of level 1.

This patch adds DFLTCC support to zlib-ng. In order to enable it, the
following build commands should be used:

    $ ./configure --dfltcc
    $ make

When built like this, zlib-ng would compress in hardware on level 1, and in
software on all other levels. Decompression will always happen in
hardware. In order to enable DFLTCC compression for levels 1-6 (i.e. to
make it used by default) one could either add -DDFLTCC_LEVEL_MASK=0x7e
at compile time, or set the environment variable DFLTCC_LEVEL_MASK to
0x7e at run time.

Two DFLTCC compression calls produce the same results only when they
both are made on machines of the same generation, and when the
respective buffers have the same offset relative to the start of the
page. Therefore care should be taken when using hardware compression
when reproducible results are desired. One such use case - reproducible
software builds - is handled explicitly: when SOURCE_DATE_EPOCH
environment variable is set, the hardware compression is disabled.

DFLTCC does not support every single zlib-ng feature, in particular:

    * inflate(Z_BLOCK) and inflate(Z_TREES)
    * inflateMark()
    * inflatePrime()
    * deflateParams() after the first deflate() call

When used, these functions will either switch to software, or, in case
this is not possible, gracefully fail.

This patch tries to add DFLTCC support in a least intrusive way.
All SystemZ-specific code was placed into a separate file, but
unfortunately there is still a noticeable amount of changes in the
main zlib-ng code. Below is the summary of those changes.

DFLTCC takes as arguments a parameter block, an input buffer, an output
buffer and a window. Since DFLTCC requires parameter block to be
doubleword-aligned, and it's reasonable to allocate it alongside
deflate and inflate states, ZALLOC_STATE, ZFREE_STATE and ZCOPY_STATE
macros were introduced in order to encapsulate the allocation details.
The same is true for window, for which ZALLOC_WINDOW and
TRY_FREE_WINDOW macros were introduced.

While for inflate software and hardware window formats match, this is
not the case for deflate. Therefore, deflateSetDictionary and
deflateGetDictionary need special handling, which is triggered using the
new DEFLATE_SET_DICTIONARY_HOOK and DEFLATE_GET_DICTIONARY_HOOK macros.

deflateResetKeep() and inflateResetKeep() now update the DFLTCC
parameter block, which is allocated alongside zlib-ng state, using
the new DEFLATE_RESET_KEEP_HOOK and INFLATE_RESET_KEEP_HOOK macros.

In order to make unsupported deflateParams(), inflatePrime() and
inflateMark() calls to fail gracefully, the new DEFLATE_PARAMS_HOOK,
INFLATE_PRIME_HOOK and INFLATE_MARK_HOOK macros were introduced.

The algorithm implemented in hardware has different compression ratio
than the one implemented in software. In order for deflateBound() to
return the correct results for the hardware implementation, the new
DEFLATE_BOUND_ADJUST_COMPLEN and DEFLATE_NEED_CONSERVATIVE_BOUND macros
were introduced.

Actual compression and decompression are handled by the new DEFLATE_HOOK
and INFLATE_TYPEDO_HOOK macros. Since inflation with DFLTCC manages the
window on its own, calling updatewindow() is suppressed using the new
INFLATE_NEED_UPDATEWINDOW() macro.

In addition to compression, DFLTCC computes CRC-32 and Adler-32
checksums, therefore, whenever it's used, software checksumming needs to
be suppressed using the new DEFLATE_NEED_CHECKSUM and
INFLATE_NEED_CHECKSUM macros.

DFLTCC will refuse to write an End-of-block Symbol if there is no input
data, thus in some cases it is necessary to do this manually. In order
to achieve this, bi_reverse and flush_pending were promoted from static
to ZLIB_INTERNAL and exposed via deflate.h.

Since the first call to dfltcc_inflate already needs the window, and it
might be not allocated yet, inflate_ensure_window was factored out of
updatewindow and made ZLIB_INTERNAL.


Dead2:
This looks pretty good already, but I have some comments.

- The commit info detailing how this works could perhaps be added as a documentation file for future reference.
- CMake support appears to be missing
- I would prefer it if new code used stdint.h data types where possible without breaking compatibility or impacting performance, especially instead of for example int that does not have a cross-platform set size.
- zlib-ng uses 120 char line lengths, please refrain from breaking lines up excessively, at the very least in the files that already use 120 char lengths. Yes, I guess we should have a code-style document somewhere.
- DFLTCC define is a bit vague, we try to prefix defines with arch information where that is relevant. S390_DFLTCC perhaps?
- I'd like to see a short comment in-code for non-arch-specific code about why things like bufsize is special-cased when DFLTCC is defined (in minigzip and gzguts for example).
- zlib-ng reading environment variables is not good, as this opens up programs using zlib-ng as a library to another vectors of security holes that they do not have any control over.

Env variables either needs be removed or at the very least put behind a compile-time option that is disabled by default and has a big fat warning for anyone wanting to enable it. In any case, I would prefer that the env-var related code was a separate commit, to make both code review and bisect easier.

I have not looked too close at what the code itself does yet, and I will probably be unable to test this. But from what I have seen so far, it looks neat, although there are a number of hooks into the generic code. We'll have to have a closer look at these before merging them.

I'd like to see benchmarks (both time and size) for at least a few of the levels, before this PR and after this PR with DFLTCC enabled/disabled.

iii-i:
@Dead2 thanks for the feedback! I will update the PR once I have resolved the issues you found.

Unfortunately, IBM does not allow me to disclose the exact throughput and compression ratio related numbers at this point, except for what is already stated in the commit message - "with estimated compression and decompression performance orders of magnitude faster than the current zlib-ng and ratio comparable with that of level 1". I hope IBM will lift this restriction in the future, and I will be able to post the results of e.g. https://quixdb.github.io/squash-benchmark/.

deflate.c
```diff
@@ -64,9 +64,29 @@ const char deflate_copyright[] = " deflate 1.2.11.f Copyright 1995-2016 Jean-lou
...
+#define DEFLATE_RESET_KEEP_HOOK(strm) do {} while (0)
+#define DEFLATE_PARAMS_HOOK(strm, level, strategy) do {} while (0)
+#define DEFLATE_BOUND_ADJUST_COMPLEN(strm, complen, sourceLen) do {} while (0)
+#define DEFLATE_NEED_CONSERVATIVE_BOUND(strm) 0
+#define DEFLATE_HOOK(strm, flush, bstate) 0
+#define DEFLATE_NEED_CHECKSUM(strm) 1
+
+#ifdef DFLTCC
+#  include "arch/s390/dfltcc_deflate.h"
+#endif
```
mtl1979, line 97: Defining macros first and then redefining them in dfltcc_deflate.h seems overkill to me... Maybe should just define dummy ones in ```#else``` block.

inflate.c
```diff
@@ -88,6 +88,23 @@
...
+#define ZFREE_WINDOW ZFREE
+#define INFLATE_RESET_KEEP_HOOK(strm) do {} while (0)
+#define INFLATE_PRIME_HOOK(strm, bits, value) do {} while (0)
+#define INFLATE_TYPEDO_HOOK(strm, flush) do {} while (0)
+#define INFLATE_NEED_CHECKSUM(strm) 1
+#define INFLATE_NEED_UPDATEWINDOW(strm) 1
+#define INFLATE_MARK_HOOK(strm) do {} while (0)
+
+#ifdef DFLTCC
+#  include "arch/s390/dfltcc.h"
```
mtl1979: This also seems overkill to me...

inflate.c
```diff
@@ -359,36 +378,19 @@ void makefixed(void) {
...
-
-   Providing output buffers larger than 32K to inflate() should provide a speed
-   advantage, since only the last 32K of output is copied to the sliding window
-   upon return from inflate(), and since all distances after the first 32K of
-   output will fall in the output data, making match copies simpler and faster.
-   The advantage may be dependent on the size of the processor's data caches.
- */
-static int updatewindow(PREFIX3(stream) *strm, const unsigned char *end, uint32_t copy) {
+int ZLIB_INTERNAL inflate_ensure_window(state)
     struct inflate_state *state;
```
mtl1979: Old style parameter type declaration here...

iii-i:
The new patch contains the following improvements:

* Added arch/s390/README.md
* Added CMake support (activated with -DWITH_DFLTCC=1)
* Removed excessive line breaking
* Renamed DFLTCC macro to S390_DFLTCC
* Added comments regarding buffer sizes
* Simplified hook macros
* Removed environment variables

Could you point to a few particular places where it would make sense to replace int with e.g. int32_t? I reviewed the code, and I think that int is used mostly as a bool replacement or as a type for values, for which zlib-ng already uses int (e.g. flush).

Even though I removed it in the new version, what do you think about re-introducing SOURCE_DATE_EPOCH environment variable support? Unfortunately, DFLTCC can produce different (but, of course, valid) compression results for the same input, and this can be bad for build reproducibility (e.g. if zlib-ng is used to compress debug info). So it would make sense to switch to software in such cases, and SOURCE_DATE_EPOCH is, in my opinion, a pretty good way to detect whether user wants a reproducible build or not.

Also, I'm not sure what to do about CodeFactor critical warning - it doesn't seem to like code duplication in minigzip, but it existed before this commit.

mtl1979:
I wouldn't worry about CodeFactor warnings in code that you didn't modify yourself...

iii-i:
Gentle ping.

configure
```diff
@@ -146,6 +147,7 @@ case "$1" in
       echo '    [--without-new-strategies]  Compiles without using new additional deflate strategies' | tee -a configure.log
       echo '    [--without-acle]            Compiles without ARM C Language Extensions' | tee -a configure.log
       echo '    [--without-neon]            Compiles without ARM Neon SIMD instruction set' | tee -a configure.log
+      echo '    [--dfltcc]                  Compiles with the DEFLATE COMPRESSION CALL instruction on IBM Z' | tee -a configure.log
```
Dead2: Should be --with-dfltcc

CMakeLists.txt
```diff
@@ -108,6 +108,10 @@ if("${ARCH}" MATCHES "arm" OR "${ARCH}" MATCHES "aarch64")
     option(WITH_NEON "Build with NEON intrinsics" ON)
 endif()
 
+if("${ARCH}" MATCHES "s390x")
+    option(WITH_DFLTCC "Build with DEFLATE CONVERSION CALL" OFF)
```
Dead2: This should mention that it is IBM Z/s390x, like it does in the configure help text

arch/s390/dfltcc.c
```diff
@@ -0,0 +1,778 @@
...
+#define CVT_CRC32 0
+#define CVT_ADLER32 1
+#define HTT_FIXED 0
+#define HTT_DYNAMIC 1
+
+struct dfltcc_param_v0 {
+    uint16_t pbvn;                     /* Parameter-Block-Version Number */
+    uint8_t mvn;                       /* Model-Version Number */
+    uint8_t ribm;                      /* Reserved for IBM use */
+    unsigned reserved32 : 31;
```
Dead2: Can these unsigned not be uint32_t?

arch/s390/dfltcc.c
```diff
@@ -0,0 +1,778 @@
...
+    }
+}
+
+/*
+   Extension of inflate_state and deflate_state. Must be doubleword-aligned.
+*/
+struct dfltcc_state {
+    struct dfltcc_param_v0 param;      /* Parameter block. */
+    struct dfltcc_qaf_param af;        /* Available functions. */
+    uLong level_mask;                  /* Levels on which to use DFLTCC */
```
Dead2: uint64_t?

arch/s390/dfltcc.c
```diff
@@ -0,0 +1,778 @@
...
+    register int r0 __asm__("r0") = fn;
+    register void *r1 __asm__("r1") = param;
+    register unsigned char *r2 __asm__("r2") = t2;
+    register size_t r3 __asm__("r3") = t3;
+    register const unsigned char *r4 __asm__("r4") = t4;
+    register size_t r5 __asm__("r5") = t5;
+    int cc;
+
+    __asm__ volatile(
+#ifdef S390_DFLTCC_USDT
```
Dead2: What is this USDT? I don't think it gets explained anywhere.

Dead2: If it is for debugging, we already have defines for enabling verbose debugging where the code will spit out a lot of information. This might be a case where that could be used instead?

inflate.c
```diff
@@ -359,6 +378,33 @@ void makefixed(void) {
 }
 #endif /* MAKEFIXED */
 
+int ZLIB_INTERNAL inflate_ensure_window(struct inflate_state *state)
```
Dead2, line 389: Splitting this out to a separate function should be a separate preparation commit (same PR). Makes it easier to bisect and benchmark.

trees.c
```diff
@@ -1044,7 +1043,7 @@ static int detect_data_type(deflate_state *s) {
  * method would use a table)
  * IN assertion: 1 <= len <= 15
  */
-static unsigned bi_reverse(unsigned code, int len) {
+ZLIB_INTERNAL unsigned bi_reverse(unsigned code, int len) {
```
Dead2, line 1046: This too in preparation-commit

deflate.h
```diff
@@ -336,6 +336,8 @@ void ZLIB_INTERNAL _tr_flush_bits(deflate_state *s);
 void ZLIB_INTERNAL _tr_align(deflate_state *s);
 void ZLIB_INTERNAL _tr_stored_block(deflate_state *s, char *buf, unsigned long stored_len, int last);
 void ZLIB_INTERNAL bi_windup(deflate_state *s);
+unsigned ZLIB_INTERNAL bi_reverse(unsigned code, int len);
+void ZLIB_INTERNAL flush_pending(PREFIX3(streamp) strm);
```
Dead2, line 340: This too in preparation-commit

deflate.c
```diff
@@ -83,7 +103,6 @@ static block_state deflate_rle   (deflate_state *s, int flush);
 static block_state deflate_huff  (deflate_state *s, int flush);
 static void lm_init              (deflate_state *s);
 static void putShortMSB          (deflate_state *s, uint16_t b);
-ZLIB_INTERNAL void flush_pending (PREFIX3(stream) *strm);
```
Dead2, line 86: This too in preparation-commit

Dead2:
Things are looking better, however I am not a big fan of the introduction of the hooks in deflate.c and inflate.c, such as `DEFLATE_RESET_KEEP_HOOK(strm);`. The way they are implemented now, they will be confusing, since they are basically noop on any other platform than s390x but it looks like it does something on all platforms.
I am not sure what I'd prefer though, either wrap each of them in an #ifdef, rename them with `S390X_` prefix for example, or add a comment to each of them. Either way would make the code more readable.

I think I'd also prefer there to be a way to compile DFLTCC only for inflate or only for deflate separately. I can at least imagine cases where the inflate would be preferred enabled but the deflate one disabled for compatibility reasons.


Regarding the SOURCE_DATE_EPOCH feature, I think the best way would be if we could add this to the zlib-ng API instead. It is a good parameter to have, and programs supporting the zlib-ng API could then hint that it wants us to make reproducible output (dwz for example). It would need to be a separate PR, and we might want to add more parameters later, so we'd need to have space for that while still remaining backwards-compatible when adding them. Would a bitmap do perhaps, or should we take short strings and parse them? If so, do we want `REPRODUCIBLE=1` style parameters in strings? One parameter per call? I think these things would be better discussed in its own PR though, but you now have the chance to suggest a new API function :)
We could possibly use this to allow an application to enable/disable usage of deflate_medium, deflate_quick, deflate_dfltcc, inflate_dfltcc, etc as well, and possibly change their level mappings. Just throwing out some possibilities here.

We should probably also have a function call in the API that can tell the application what features are compiled in and available, and what is enabled/disabled and what compression levels are using what compression methods.

Please give me some feedback on whether you want to make a PR for one or both of these. Otherwise me, @sebpop or @mtl1979 (or some other lurker) might pick it up.

Dead2:
I understand (even if I do not like) the lack of benchmark results, however I'd still like to see regression benchmarks for s390x:
- Before this commit
- With this commit but disabled completely
- With this commit and enabled, but on hardware not supporting the instruction (if possible)
I guess these should all be OK, since neither of them are with the DFLTCC instructions in actual use. And I guess that all of these can be run on older s390x cpus, and they should at least not be run in an emulator :)

From the looks of things, the preparatory commit I have requested you make will be the one most likely to negatively affect performance, and possibly on any platform, so I will be doing my own benchmarks on x86_64 at some point before I accept this PR.

arch/s390/dfltcc.c
iii-i: Ok, I'll change the types here as follows: level_mask:uint16_t (enough for levels 1-9), block_size:uint32_t (this has to match avail_in), block_threshold:size_t (this has to match total_in), dht_threshold:uint32_t (this also has to match avail_in).

arch/s390/dfltcc.c
iii-i: Yes, this is for debugging (https://myaut.github.io/dtrace-stap-book/app/proc.html). One has to have sys/sdt.h present to make use of it. Since the inserted probe points cost almost nothing, maybe I should simply replace S390_DFLTCC_USDT with something like HAVE_SYS_SDT_H?

iii-i:
Regarding hooks, I would start with documenting them. The first version of this patch had `#ifdef`s all over the place and people found it annoying, that's why I've decided to hide the new code behind the hooks.

One can produce the inflate-only version of zlib-ng already by building with `-DS390_DFLTCC -DDFLTCC_LEVEL_MASK=0`, which I could document. Or would it be better to be more explicit and split `-DS390_DFLTCC` into `-DS390_DFLTCC_DEFLATE` and `-DS390_DFLTCC_INFLATE`?

I like the idea with parsing string parameters, since it would allow zlib-ng to support a variety of different options. I could send a couple patches with the general infrastructure and the implementation of the `REPRODUCIBLE` option. I think I'll pass on the query API, at least until I'm done with this and parameters PRs.

Would it be sufficient to send the relative performance numbers, i.e. by how much percent the throughput and ratio have regressed or improved with this patch?

Dead2:
Comments for the hooks would help, or renaming them since they are s390x-only.

I think the separate defines for inflate/deflate would be better, along with the related options in configure and cmake of course.

Relative performance numbers would be ok. What I am looking for is evidence that things does not break or become slower, and that we don't merge an optimization that looks good on paper but ends up not actually helping any in the real-world.

iii-i:
The new patches contain the following improvements:
* Split introduction of `inflate_ensure_window` and hooks into separate patches
* Added comments for hooks
* Split inflate/deflate support
* Changed types of some variables to `u?int\d+_t`

I still have a TODO to re-run the [squash](https://quixdb.github.io/squash-benchmark/) benchmark. Or is there some other benchmark that you would prefer?

arch/s390/dfltcc.c
Dead2: It can stay like it is, I was just wondering whether it would print debugging messages similar to assert for example.

Dead2:
The problem with squash is that it by default compiles zlib-ng with a static makefile without any arch files and without any defines set since it is unable to run our configure/cmake scripts to set things up properly.
We have recently been improving our defaults to have safe things enabled by default, so cmake will instead disable them when needed, but it still leaves out things that cannot be safely on by default and some things have not been changed yet either.
I have a fork of Squash that has a couple improvements that takes us about half-way towards being able to use zlib-ng's cmake configuration properly. It needs some more love though, since it does not seem to run our cmake scripts properly, it just seems to use the default settings or something.

@sebpop uses chrome's png benchmark, or some such
I use minigzip for my benchmarking, spawning it through perf to capture accurate cpu-time (averaged over multiple runs on a cpu core that is otherwise idle, with cpu turbo disabled and sleep states beyond C1e also disabled).
Jsnell has also made a zlib bench that is pretty good.


Regarding the hook comments, I meant inline comments where they get used, since currently you would have to search to find the definition and see that it is doing nothing except when on s390x. So a comment like `// Hook for s390x's DFLTCC` for example would make it very easy to see that this hook is not in use in most cases. The comments where they get defined are also very nice though.

I just did a quick scroll through, and the rest of the changes look good so far, Good work!

iii-i:
I've added the comments to call sites as well (I did not comment memory management, NEED_CHECKSUM and NEED_UPDATEWINDOW, since those are used in multiple places and look self-explanatory).

So far I've managed to re-use my plain zlib benchmarking setup, where I build squash-benchmark as usual with all the downsides that you mentioned, but then LD_PRELOAD my zlib version during the run. I will normalize the figures and post them.

I'll also try minigzip - what types of files do you compress with it for your benchmarking?

iii-i:
The new patch contains the following improvements:

- The logic has been split between `dfltcc_common.c`, `dfltcc_deflate.c` and `dfltcc_inflate.c` in order to avoid building code for features that are off
- Added feature detection for `sys/sdt.h`

I have posted the normalized squash-benchmark figures observed on a development system running on a GA hardware here:

https://github.com/iii-i/zlib-ng/wiki/Performance-with-dfltcc-patch-applied-and-dfltcc-support-not-built
https://github.com/iii-i/zlib-ng/wiki/Performance-with-dfltcc-patch-applied-and-dfltcc-support-built

This should at least demonstrate that there are no noticeable performance regressions, since `compress_cpu` and `decompress_cpu` are pretty close to 1. There are outliers, but I think they occur because I could not find a dedicated system to run the benchmark.

minigzip figures:

```
$ time -p zlib-ng-plain/minigzip <rnd >/dev/null
real 3.76
user 3.73
sys 0.03
$ time -p zlib-ng-dfltcc-on/minigzip <rnd >/dev/null
real 3.78
user 3.75
sys 0.02
```

I've expected to be allowed to publish the normalized figures for the hardware deflate by now, but unfortunately this discussion is still ongoing. Sorry about that!

iii-i:
Here are the figures for the hardware deflate from the internal development system:

https://github.com/iii-i/zlib-ng/wiki/Performance-with-dfltcc-patch-applied-and-dfltcc-support-built-on-dfltcc-enabled-machine

It is enabled only on level 1, therefore all other levels have figures similar to those from the performance regression test tables. On level 1, one can observe up to x100 speed-ups in the `compress_cpu` column, depending on the data set.

xnox:
My preference in Ubuntu would be to use `secure_getenv()` instead of `getenv()` if possible, where available.

iii-i:
@Dead2 @mtl1979 Ping.

arch/s390/dfltcc_deflate.c
```diff
@@ -0,0 +1,370 @@
...
+        return 0;
+
+    return 1;
+}
+
+static inline void dfltcc_gdht(PREFIX3(streamp) strm)
+{
+    deflate_state *state = (deflate_state *)strm->state;
+    struct dfltcc_param_v0 *param = &GET_DFLTCC_STATE(state)->param;
+    size_t avail_in = avail_in = strm->avail_in;
```
mtl1979: This doesn't look right, assigning contents of variable to itself.

arch/s390/dfltcc_deflate.c
```diff
@@ -0,0 +1,370 @@
...
+    need_empty_block = flush == Z_FINISH && param->bcf && !param->bhf;
+
+    /* Translate stream to parameter block */
+    param->cvt = state->wrap == 2 ? CVT_CRC32 : CVT_ADLER32;
+    if (!no_flush) {
+        /* We need to close a block */
+        if (param->bcc) {
+            /* We have already asked the hardware to do this */
+        } else if (strm->avail_in == 0)
+            /* There is no input data, so the hardware will not nohor our
```
mtl1979: "nohor" is not a word. Typo?

arch/s390/dfltcc_deflate.c
mephi42: Most likely a leftover from before this function was factored out. I wonder why didn't the compiler spot this?!

arch/s390/dfltcc_deflate.c
mephi42: Fixed

arch/s390/dfltcc_deflate.c
iii-i: Most likely a leftover from before this function was factored out. I wonder why didn't the compiler spot this?!

mephi42:
* Fixed a bug when `Z_SYNC_FLUSH` usage led to incomplete EOBS write.

@xnox Thanks for the advice! Currently the usage of environment variables has been completely dropped, but if it comes to that again in some form, I will use `secure_getenv`.

iii-i:
* Fixed a bug when `Z_SYNC_FLUSH` usage led to incomplete EOBS write.

@xnox Thanks for the advice! Currently the usage of environment variables has been completely dropped, but if it comes to that again in some form, I will use `secure_getenv`.

arch/s390/dfltcc_deflate.c
```diff
@@ -0,0 +1,365 @@
...
+    return cc;
+}
+
+static inline void send_eobs(PREFIX3(streamp) strm, const struct dfltcc_param_v0 *param)
+{
+    deflate_state *state = (deflate_state *)strm->state;
+
+    send_bits(state, bi_reverse(param->eobs >> (15 - param->eobl), param->eobl), param->eobl);
+    flush_pending(strm);
+    if (state->pending != 0) {
```
iii-i, line 81: Fixed another bug, which caused a partial EOBS write.

iii-i:
I believe I have fixed all the known bugs in the latest patch - is there anything I can do to have this PR merged?

mtl1979:
@iii-i I'm all for supporting more platforms and instruction sets, but this patch touches so many parts of the code that accepting it with limited information and no chance to verify the code ourselves makes accepting the pull request a tough decision.

iii-i:
@mtl1979 I understand your concerns - I tried for quite some time to make the patch least intrusive, and still I'm not particularly happy with the end result. So if you have any suggestions on how this can be improved, I'm totally open to them.

I'd like to still point out that we are now at least at the point when common code changes are isolated in smaller separate commits (https://github.com/zlib-ng/zlib-ng/commit/93856b, https://github.com/zlib-ng/zlib-ng/commit/b4c48d and https://github.com/zlib-ng/zlib-ng/commit/c098fa for this iteration), which are hopefully easy to review and regression test on other platforms.

If you need some additional information, please let me know. I cannot promise I can provide it due to all the restrictions that are in place for now, but at least with the relative performance figures it worked out in the end.

Unfortunately, giving zlib-ng team access to the test system is not possible at all at the moment due to aforementioned restrictions. We could try to arrange this when such systems become generally available though.

We are currently integrating patches for the future hardware into open source software (just a small example - https://github.com/bminor/binutils-gdb/commit/ba354106f02e25c9adc3dc1364d6a974216ec94e) and distributions, so that when it comes out, its users can benefit from the new features right away. So for us it's important not only to integrate such patches at all, but also to do that preferably at an earlier point in time.

mtl1979:
@iii-i Currently we don't have enough test suites to automatically validate optimized functions against results from non-optimized "default" implementations. When those are added to the repository, it will be a lot easier to merge in code for new architectures. 

My second concern is that if we by default only use optimized deflate with just level 1, it will be basically marginal case as most applications would use levels 6 to 9 due to better compression rate when using non-optimized deflate. In my opinion, levels below 6 still can be tuned better on zlib-ng thanks to recent bug fixes.

iii-i:
@mtl1979 I could run the regression tests / benchmarks of your choice manually if that would help. I have access to a bunch of x86_64 and aarch64 machines.

We thought about enabling DFLTCC on level 6 in the upstream code, but came to a conclusion that it would be better if it’s rather distros who do this. There are two things that can go wrong:
- Compression ratio is lower than what people are used to on level 6.
- The compression results are not bit-reproducible, and some software might care about this. E.g. [debuginfo compression](https://github.com/bminor/binutils-gdb/blob/d87343/gas/compress-debug.c#L37), which uses `Z_DEFAULT_COMPRESSION`, will result in a non-reproducible build. It would be unfortunate if a user just builds the code from zlib-ng repository and runs into such a counter-intuitive behavior. However, distros are in better position to evaluate pros and cons of such a default.

Dead2:
> I like the idea with parsing string parameters, since it would allow zlib-ng to support a variety of different options. I could send a couple patches with the general infrastructure and the implementation of the `REPRODUCIBLE` option.

I am still awaiting this PR from you ;)

But I am willing to merge this PR for now. The overall impact on the arch-generic code is pretty low, and the hooks are commented to make them easier to spot/understand. I do hope you will help maintaining this code though :)

iii-i:
@Dead2 Thanks! We are interested in keeping this code in upstream zlib-ng long-term, so I will definitely help maintaining it. Feel free to summon me to relevant issues and PRs.

I have created a pull request for the parameters infrastructure: https://github.com/zlib-ng/zlib-ng/pull/340. After some thinking, I have decided against using strings - but let's discuss it there.

PR #332, opened 2019-04-01
fix configure output when compiler is clang

sebpop:
The patch cleans up the expected errors that clutter the output of configure:

$ CC=clang ./configure
Checking for compiler... clang-6.0
clang: error: unknown argument: '-print-multiarch'
clang: error: no input files

With this patch the last two lines disappear.

### DIFF
--- a/configure
+++ b/configure
@@ -265,7 +265,7 @@ if test "$gcc" -eq 1 && ($cc $CFLAGS -c $test.c) >> configure.log 2>&1; then
   CFLAGS="${CFLAGS} -std=c99"
 
   # Re-check ARCH if the compiler is a cross-compiler.
-  if $CC -print-multiarch && test -n "$($CC -print-multiarch)" 1> /dev/null 2>&1; then
+  if $CC -print-multiarch 1> /dev/null 2>&1 && test -n "$($CC -print-multiarch)" 1> /dev/null 2>&1; then
       CC_ARCH=`$CC $CFLAGS -print-multiarch | sed 's/-.*//g'`
   else
       CC_ARCH=`$CC $CFLAGS -dumpmachine | sed 's/-.*//g'`


Dead2:
Should we perhaps instead check for GCC first? Hiding the error might not be the best approach here.
It looks to me like we don't support cross-compiling with anything other than GCC, is that correct? If so, should we support others?

mtl1979:
@Dead2 Cross-compiling using clang works, but it needs one configuration file per target, so works differently than with gcc.

mtl1979:
Supporting cross-compiling using clang is tricky, but not harder than cross-compiling through cmake.

configure
```diff
@@ -265,7 +265,7 @@ if test "$gcc" -eq 1 && ($cc $CFLAGS -c $test.c) >> configure.log 2>&1; then
   CFLAGS="${CFLAGS} -std=c99"
 
   # Re-check ARCH if the compiler is a cross-compiler.
-  if $CC -print-multiarch && test -n "$($CC -print-multiarch)" 1> /dev/null 2>&1; then
+  if $CC -print-multiarch 1> /dev/null 2>&1 && test -n "$($CC -print-multiarch)" 1> /dev/null 2>&1; then
```
sebpop, line 268: The redirection to /dev/null does not hide the error, the return code of executing `$CC -print-multiarch` is used to find out whether the compiler `$CC` supports that flag.  clang for example does not support it.

sebpop:
> It looks to me like we don't support cross-compiling with anything other than GCC, is that correct?

This is not correct: I am able to use both gcc and clang as cross compilers from x86 to aarch64 and 32-bit arm.
Setting up those cross-compilers may be a bit hard, so I would recommend using linaro's cross compilers.

Dead2:
Good :)

PR #333, opened 2019-04-03
fix configure --help text to match parsed options

sebpop:


mtl1979:
The start of line is getting very long compared to the others, maybe we should split the options as two separate lines so the help text can still start at same position as other lines.

sebpop:
Thanks @mtl1979 for the suggestion: it helped improve the text wording as well.

mtl1979:
@sebpop That's one of the reasons I suggested splitting the line... The original text didn't sound very informational...

PR #335, opened 2019-04-04
Fix infcover.c build

iii-i:
infcover.c includes inflate.h, which makes use of PREFIX3 macro. This
macro is defined in zbuild.h, which is not included.


sebpop:
Looks good to me.

mtl1979:
Looks good to me... ```zbuild.h``` is supposed to always be the very first header file to be included in each source file of zlib-ng itself, so that's why it is not included in other header files by default, especially ones that might be included by external projects.

Issue #337, opened 2019-05-19
DYNAMIC_CRC_TABLE build error


nmoinvaz:
When building with #define `DYNAMIC_CRC_TABLE`, functable.c references static variable _crc_table_empty_ and static function _make_crc_table_, which are found in crc32.c.

```
1>functable.c(107): error C2065: 'crc_table_empty': undeclared identifier
1>functable.c(108): warning C4013: 'make_crc_table' undefined; assuming extern returning int
```

mtl1979:
I think we need "extern" declaration for those two symbols at the top of functable.c ... Makes no sense to create private header just so they can be used in two files.

nmoinvaz:
Looks like this was merged. Thanks.

PR #339, opened 2019-05-20
Fix build when DYNAMIC_CRC_TABLE is defined.

mtl1979:
* Fixes issue #337.


functable.c
```diff
@@ -10,6 +10,11 @@
 
 #include "gzendian.h"
 
+#ifdef DYNAMIC_CRC_TABLE
```
Dead2: I think this should move down to the crc32-area (although that area seems to be missing a comment to make it more visible) for example after the definiton of crc32_generic, around line 39.

PR #340, opened 2019-05-23
Add public zng_deflate{Set,Get}Params() API functions

iii-i:
These functions allow zlib-ng callers to modify and query the
compression parameters in a future-proof way. When the caller requests a
parameter, which is not supported by the current zlib-ng version, this
situation is detected and reported to the caller. The caller may modify
or query multiple parameters at once. Currently only "level" and
"strategy" parameters are supported. It is planned to add a
"reproducible" parameter, which would affect whether IBM Z DEFLATE
CONVERSION CALL is used.

Passing enum and void * buffer was chosen over passing strings, because
of simplicity for the caller. If strings were used, C callers would have
to call snprintf() and strtoul() for setting and getting integer-valued
parameters respectively, which is quite tedious.

Bulk updates were chosen over updating individual parameters separately,
because it might make sense to apply some parameters atomically, e.g.
level and strategy.

The new functions are defined only for zlib-ng, but not compat zlib.

### DIFF
--- a/zlib-ng.h
+++ b/zlib-ng.h
@@ -1793,6 +1793,13 @@ ZEXTERN int ZEXPORT zng_gzgetc_(gzFile file);  /* backward compatibility */
 typedef enum {
     Z_DEFLATE_LEVEL = 0,         /* compression level, represented as an int */
     Z_DEFLATE_STRATEGY = 1,      /* compression strategy, represented as an int */
+    Z_DEFLATE_REPRODUCIBLE = 2,
+    /*
+         Whether reproducible compression results are required. Represented as an int, where 0 means that it is allowed
+       to trade reproducibility for e.g. improved performance or compression ratio, and non-0 means that
+       reproducibility is strictly required. Reproducibility is guaranteed only when using an identical zlib-ng build.
+       Default is 0.
+    */
 } zng_deflate_param;
 
 typedef struct {
--- a/zlib-ng.h
+++ b/zlib-ng.h
@@ -1790,6 +1790,46 @@ ZEXTERN int ZEXPORT zng_gzgetc_(gzFile file);  /* backward compatibility */
 #endif /* WITH_GZFILEOP */
 
 
+typedef enum {
+    Z_DEFLATE_LEVEL = 0,         /* compression level, represented as an int */
+    Z_DEFLATE_STRATEGY = 1,      /* compression strategy, represented as an int */
+} zng_deflate_param;
+
+typedef struct {
+    zng_deflate_param param;  /* parameter ID */
+    void *buf;                /* parameter value */
+    size_t size;              /* parameter value size */
+    int status;               /* result of the last set/get call */
+} zng_deflate_param_value;
+
+ZEXTERN int ZEXPORT zng_deflateSetParams(zng_stream *strm, zng_deflate_param_value *params, size_t count);
+/*
+     Sets the values of the given zlib-ng deflate stream parameters. All the buffers are copied internally, so the
+   caller still owns them after this function returns. Returns Z_OK if success.
+
+     If the size of at least one of the buffers is too small to hold the entire value of the corresponding parameter,
+   or if the same parameter is specified multiple times, Z_BUF_ERROR is returned. The caller may inspect status fields
+   in order to determine which of the parameters caused this error. No other changes are performed.
+
+     If the stream state is inconsistent or if at least one of the values cannot be updated, Z_STREAM_ERROR is
+   returned. The caller may inspect status fields in order to determine which of the parameters caused this error.
+   Parameters, whose status field is equal to Z_OK, have been applied successfully. If all status fields are not equal
+   to Z_STREAM_ERROR, then the error was caused by a stream state inconsistency.
+
+     If there are no other errors, but at least one parameter is not supported by the current zlib-ng version,
+   Z_VERSION_ERROR is returned. The caller may inspect status fields in order to determine which of the parameters
+   caused this error.
+*/
+
+ZEXTERN int ZEXPORT zng_deflateGetParams(zng_stream *strm, zng_deflate_param_value *params, size_t count);
+/*
+     Copies the values of the given zlib-ng deflate stream parameters into the user-provided buffers. Returns Z_OK if
+   success, Z_VERSION_ERROR if at least one parameter is not supported by the current zlib-ng version, Z_STREAM_ERROR
+   if the stream state is inconsistent, and Z_BUF_ERROR if the size of at least one buffer is too small to hold the
+   entire value of the corresponding parameter.
+*/
+
+
 /* provide 64-bit offset functions if _LARGEFILE64_SOURCE defined, and/or
  * change the regular functions to 64 bits if _FILE_OFFSET_BITS is 64 (if
  * both are true, the application gets the *64 functions, and the regular
--- a/arch/s390/README.md
+++ b/arch/s390/README.md
@@ -20,7 +20,10 @@ Two DFLTCC compression calls produce the same results only when they
 both are made on machines of the same generation, and when the
 respective buffers have the same offset relative to the start of the
 page. Therefore care should be taken when using hardware compression
-when reproducible results are desired.
+when reproducible results are desired. In particular, zlib-ng-specific
+zng_deflateSetParams call allows setting Z_DEFLATE_REPRODUCIBLE
+parameter, which would disable DFLTCC if reproducible results are
+required.
 
 DFLTCC does not support every single zlib-ng feature, in particular:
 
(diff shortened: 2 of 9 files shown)


zlib-ng.h
```diff
@@ -1835,6 +1835,72 @@ ZEXTERN int              ZEXPORT zng_deflateResetKeep (zng_stream *);
 ZEXTERN int ZEXPORTVA zng_gzvprintf(gzFile file, const char *format, va_list va);
 #endif
 
+typedef enum {
+    Z_DEFLATE_LEVEL = 0,
+    /*
+         Compression level. Represented as an int.
+    */
```
mtl1979: No need for comment to span three lines. It should fit on same line as enum item.

zlib-ng.h
```diff
@@ -1835,6 +1835,72 @@ ZEXTERN int              ZEXPORT zng_deflateResetKeep (zng_stream *);
 ZEXTERN int ZEXPORTVA zng_gzvprintf(gzFile file, const char *format, va_list va);
 #endif
 
+typedef enum {
```
Dead2: All this should probably move up to line ~1792, above the #ifdef Z_LARGE64 parts.

Also, the comments are not following the established style for the file, they involve much more white-space than we are used to. The ones in the enum should be single-line comments for example.

Dead2:
I'd like to see example buffers both for set and get. I tried to figure it out from the code + comments, but I was not able to figure it out in the short time I have available for that right now.

Things look pretty solid though.

zlib-ng.h
iii-i: I'll change that.

iii-i:
The CI appears to have failed on a POWER machine. I have access to one and I could reproduce the segfault. Debugging...

iii-i:
So far I only managed to figure out that if I take `develop` branch and add a second `test_large_deflate` call to `example.c` as follows

```
--- a/test/example.c
+++ b/test/example.c
@@ -548,6 +548,7 @@ int main(int argc, char *argv[])

     test_large_deflate(compr, comprLen, uncompr, uncomprLen);
     test_large_inflate(compr, comprLen, uncompr, uncomprLen);
+    test_large_deflate(compr, comprLen, uncompr, uncomprLen);

     test_flush(compr, &comprLen);
     test_sync(compr, comprLen, uncompr, uncomprLen);
```

the same problem would occur. If I do it this way instead, the problem does not occur:

```
--- a/test/example.c
+++ b/test/example.c
@@ -546,6 +546,7 @@ int main(int argc, char *argv[])
     test_deflate(compr, comprLen);
     test_inflate(compr, comprLen, uncompr, uncomprLen);

+    test_large_deflate(compr, comprLen, uncompr, uncomprLen);
     test_large_deflate(compr, comprLen, uncompr, uncomprLen);
     test_large_inflate(compr, comprLen, uncompr, uncomprLen);
```

which makes me think this is related to contents of modified `uncompr` buffer. Valgrind and ASAN do not see anything, so this is unlikely to be a buffer overrun. Backtrace:

```
Program received signal SIGSEGV, Segmentation fault.
longest_match (cur_match=40073, s=0x100538b0) at /home/iii/zlib-ng/match_p.h:88
88	    scan_end1 = scan[best_len-1];
(gdb) bt
#0  longest_match (cur_match=40073, s=0x100538b0) at /home/iii/zlib-ng/match_p.h:88
#1  deflate_slow (s=0x100538b0, flush=<optimized out>) at /home/iii/zlib-ng/deflate_slow.c:65
#2  0x00000000100041b0 in zng_deflate (strm=0x3fffffffeed0, flush=<optimized out>) at /home/iii/zlib-ng/deflate.c:1026
#3  0x0000000010001418 in test_large_deflate (compr=compr@entry=0x10040010 "x\001\354̱\t", comprLen=40000, uncompr=uncompr@entry=0x10049c60 "", uncomprLen=uncomprLen@entry=40000)
    at /home/iii/zlib-ng/test/example.c:289
#4  0x0000000010000b14 in main (argc=<optimized out>, argv=<optimized out>) at /home/iii/zlib-ng/test/example.c:552
```

`best_len` is `0`, which appears to make `scan[best_len-1]` invalid, but I don't know yet whether this is normal and why this is happening.

mtl1979:
@iii-i My wild guess would be that it doesn't have a match yet... Other targets don't cause exception because the contents are not actually accessed, otherwise it would trap on the line that actually tries to dereference the out-of-bound array pointer.

If ```s->prev_length``` is ```0```, we could set ```best_len``` to ```1```, so it would not try to access array index ```-1```...

Line 69: ```best_len = s->prev_length ? s->prev_length : 1;```

mtl1979 (changes requested):

iii-i:
@mtl1979 Interesting point regarding other architectures! I added

```
--- a/match_p.h
+++ b/match_p.h
@@ -49,6 +49,7 @@
  *
  */
 static inline unsigned longest_match(deflate_state *const s, IPos cur_match) {
+    if (s->prev_length == 0) abort();
     const unsigned wmask = s->w_mask;
     const Pos *prev = s->prev;
```

and it did not fire on x86_64, but fired on ppc64 and s390x (without DFLTCC). I've added a watchpoint, and it caught the place where `0` is stored to `prev_length`:

```
Old value = 2
New value = 0
0x0000000010007774 in deflate_slow (s=0x100538b0, flush=<optimized out>) at /home/iii/zlib-ng/deflate_slow.c:57
57	        s->prev_length = s->match_length, s->prev_match = s->match_start;
```

however

```
(gdb) p s->match_length
$3 = 2
(gdb) p s->prev_length
$4 = 0
```

so I've no idea where is `0` coming from?! I'll try to make sense of the assembly.

mtl1979:
@iii-i My guess is that the ```p s->match_length``` returns ```2```, because the next line (line 58) sets it to 2 (MIN_MATCH=3, 3-1=2).

iii-i:
@mtl1979 thanks, you are right! This is because of compiler reordering. With `-O0`:

```
Old value = 2
New value = 0
0x000000001000b2b4 in deflate_slow (s=0x100538b0, flush=0) at /home/iii/zlib-ng/deflate_slow.c:57
57	        s->prev_length = s->match_length, s->prev_match = s->match_start;
(gdb) p s->match_length
$3 = 0
(gdb) p s->prev_length
$4 = 0
```

mtl1979:
@iii-i Now we just need to test there is no regressions if we set ```best_len``` to ```1``` instead of ```0```...

iii-i:
@mtl1979 Smoke test (`while true; do for l in $(seq 1 9); do head -c10000000 /dev/urandom | hexdump -C | ./minigzip -$l | ./minigzip -d | pv >/dev/null; done; done`) does not show any problems, and I can run something else that tries different kinds of data, e.g. squash, in addition, but I don't understand the following:

Wouldn't `longest_match` start returning `1` instead of `0` if doesn't find anything? Or is the fact that it's called in the first place somehow ensures that some match will be found? Or do returning `0` and `1` mean the same to callers, since both of these values are less than `MIN_MATCH`?

Also, now that `scan_end1` and `scan_end` are essentially fake, what would happen if, by accident, `match[best_len] == scan_end && match[best_len-1] == scan_end1`?

mtl1979:
@iii-i As far as I know it never checks if return value of longest_match() is exactly 0. In case where scan_end and scan_end1 would both be 0, it could confuse the code but as we fool it to think current match is single byte, it should never try to match against string of bytes at same buffer index.

iii-i:
@mtl1979 I let it run overnight, and, unfortunately, it has hit the `match[2]?` assert.

mtl1979:
@iii-i Then the only choice left is early exit if there is no match...

iii-i:
@mtl1979 I got back to working on this today again, and realized that bogus `scan_end` and `scan_end1` cannot lead to the `match[2]?` assertion error, because what's violated here is "if the first two bytes and the hash match, then the third byte matches as well" assumption, which has nothing to do with `scan_end` and `scan_end1`. So I built `develop` zlib-ng, and the assertion error occurred there as well. I will create a separate issue for this.

mtl1979:
@iii-i In the worst case we just need to get rid of that assert and compare all MIN_MATCH bytes... There is possible border case where the hash can be invalid at end of valid input buffer due to hashing always 2 bytes ahead of given memory location.

iii-i:
@mtl1979 In this case `scan` and `match` seem to be in the middle of input (`s->lookahead == 15490`). I've created https://github.com/zlib-ng/zlib-ng/issues/361 for this.

Dead2:
> I'd like to see example buffers both for set and get. I tried to figure it out from the code + comments, but I was not able to figure it out in the short time I have available for that right now.

Still waiting for this.

iii-i:
@Dead2 Might [this](https://github.com/zlib-ng/zlib-ng/pull/340/files#diff-7e7c6b06b673b61d8ee0da60976ba21cR254) be what you are looking for?

Dead2:
I see it now :)

Please rebase this, so we can see whether travis succeeds this time.

Dead2:
Thanks

Issue #338, opened 2019-05-19
CMake suggestions


nmoinvaz:
Had some suggestions while testing integration:

* Turn off `ZLIB_ENABLE_TESTS` by default. I don't think the tests should be turned on by default, since most cases they will only be built when somebody is testing the functionality. With minizip, I have my tests turned off by default. Maybe there is a good reason for it.
* Allow compilation without any of the Gzip functionality. 
  * gzguts.h might need to be moved to ZLIB_GZFILE_SRCS so it doesn't get compiled in with `WITH_GZFILEOP` turned OFF.
  * gzendian.h, why is endian detection part of gz* when it is used in numerous files. Does it make sense to rename this to zendian.h?
* Remove add_library for _zlibstatic_ and don't specify SHARED or STATIC in add_library for _zlib_. Instead, the library type for _zlib_ can be specified by using [-DBUILD_SHARED_LIBS=ON](https://cmake.org/cmake/help/v3.0/variable/BUILD_SHARED_LIBS.html). Maybe this was done to compile both SHARED and STATIC at the same time?

mtl1979:
1. For stable branch, testing should be disabled by default.
2. It should already be possible to disable gzip-related functionality. If that works incorrectly, people need to post error logs as separate issues 
3. Sometimes it is useful to be able to build both static and shared libraries at the same time. Some features require static linkage due to being private symbols. Test suites are one good example.

nmoinvaz:
Other suggestions I have:

* Prefix all the options with something so they don't conflict with other cmake projects. So like `WITH_GZFILEOP` would become `ZNG_WITH_GZFILEOP`.
* On windows, everytime I set up zlib-ng it complains about zconf.h not existing and some other file. I assume this is because the ./configure script doesn't work on windows. It would be great if cmake just detected if it were Windows and copied the zconf.h.in to zconf.h if it were Windows.

mtl1979:
@nmoinvaz extra prefixes is just overkill... We should never depend on other projects.

There is a lot of things that needs to be fixed with cmake but it has been quite low priority as cmake is not something that is native to Windows development environment. Visual Studio 2017 was first to introduce it and it really has issues that make it quite useless and repelling for general use. Personally I use SDK for Windows 7 or 8.1 depending on if I want to make 32-bit or 64-bit binaries.

nmoinvaz:
[...]

PR #341, opened 2019-05-24
Cleanup old legacy makefiles and scripts

nmoinvaz:
I found out that the issue on Windows (see #338) was due to missing files from the repository. In that even though it was using cmake, it still required the use of "zconf.in" and from there the "zconf.in.cmakein" was built. With these changes I have removed all the unix makefiles and scripts and now there only exists "cmakein" files. 

In addition I've cleaned up the support text files and moved some of their contents into README.md and make it so that the most important information for developers is available right there to start.

I also moved the CVE tests to use the cmake test apparatus. I had to use a python script (cross-platform) in order to evaluate a different exit code other than 1 which was necessary for the old tests.


mtl1979:
```
-- Renaming
--     /home/travis/build/zlib-ng/zlib-ng/zconf-ng.h.cmakein
-- to 'zconf-ng.h.cmakeincluded' because this file is included with zlib
-- but CMake generates it automatically in the build directory.
CMake Error: File /home/travis/build/zlib-ng/build/zconf-ng.h.cmakein does not exist.
CMake Error at CMakeLists.txt:653 (configure_file):
  configure_file Problem configuring file
```

nmoinvaz:
I have to still take a look at the CI and try and see what I can do to fix all the environments.

mtl1979:
Now Travis is failing because cross-compiler detection is not implemented correctly in CMakeLists.txt ...

nmoinvaz:
Will reopen when I got it working.

mtl1979:
The problem is that ```MAKE``` is special variable that many commands depend on, so it can't have any parameters specified... you need to use another variable to pass any required build parameters.

PR #343, opened 2019-05-27
Added cmake support for cross-compiling on ARM

nmoinvaz:
In this pull request:

- Added cmake support for cross-compiling on ARM
- Fixed bug detecting float abi in cmake, this was due to the use of `${CC}` which should have been `$ENV{CC}`. I have instead decided to use the CMAKE_C_COMPILER variable.
- Added cross-platform testing of CVEs in cmake. Here I created a python script that is similar to _testCVEinputs.sh_ but works across all platforms.
- Simplified the generation of fuzzer exes in cmake using foreach loop.
- Updated travis CI to use cmake only.
- Added Windows os to the matrix list in travis CI.
- Cleanup cmake a bit, some formatting and moving similar things together for readability.

There are other improvements that I have in mind but this is the first step toward cmake only builds.


Dead2:
This PR does too many things. Some things might be able to go into the same PR, but in that case they should be separate commits in the PR.

Removing configure from travis for example does not fit into this PR.
Formatting, comment-changes, moving code around and such that do not change functionality, should be a separate commit at the very least.

nmoinvaz:
This PR has been split up into #350, #351, #352, #353, #354, #355, #356.

PR #344, opened 2019-05-27
Delete old unix makefiles and scripts

nmoinvaz:
This pull request works in-conjunction with issue #343 which moves the repository over to cmake only build system. I have removed some unix makefiles and scripts and other ancient artifacts that are no longer relevant to the cmake build system.

Note: This pull request will fail any review checks that rely on these old systems.


Dead2:
As I already told you on IRC, the configure script is not going away.
At the very least we would then have to have cmake 2.8-support in place again. That was removed last year with the reasoning that we still had 'configure' that for example the long-term-stable enterprise distros can use, and thus we could safely make use of more of the modern cmake features from cmake 3+.

nmoinvaz:
I thought you said it was not going away until cmake support matched configure support, as with my PRs it now does.

PR #345, opened 2019-05-27
Remove old zlib artifacts

nmoinvaz:
I have created this as a separate pull request. I don't think anybody is interested in these files anymore. If they want them it is in the madler/zlib repository or in this repository history. Maybe there is some value in holding on to them for nostalgia, but I think it is very little. If they do convey something of importance, it might be better to move those certain informations to the Wiki page.

Additionally, some of the instructions in these readme are out-dated.


Dead2:
We might need to go through these files and copy out the information that is still relevant.

mtl1979:
I think we need to sort out the documentation and clean it up before first release... It might be the last thing before the release, but we can't postpone it too much...

nmoinvaz:
@Dead2, I tried to do that as much as possible with #346, but it would be good to have somebody else go over it as well.

Dead2:
The changelog and the readme we can remove. But I feel that the FAQ still contains things we might want to keep in some form. If you can change this PR to only remove the two former files, then at least we can get those two out of the way.

nmoinvaz:
@Dead2, would it be sufficient if I move the FAQ to the Wiki?

nmoinvaz:
I have found a change log in the inflate.c source code that I've removed as well. Looks like CodeFactor is failing due to the duplication that is fixed in #368.

nmoinvaz:
Also, I propose we get rid of treebuild.xml too.

Dead2:
The wiki is not currently a good place for a FAQ to live unfortunately. I think we should still have an (updated) FAQ file distributed with the source.
The rest looks good.

We also need to update the man page and pdf (or delete the pdf perhaps?).

mtl1979:
The problem with man files and pdf files is that man files are Unix-specific, but pdf files can only be created/updated on operating systems and distributions that include the conversion tools, but can be viewed on Windows too.

nmoinvaz:
I added the FAQ back and removed treebuild.xml and the zlib.pdf. Looks like this new CodeInspector and CodeFactor are failing due to errors that are not mine.

nmoinvaz:
Rebasing looks like it fixed the CI issues.

mtl1979:
I don't honestly know what tool uses treebuild.xml files and how they are generated or updated... It seems to me to be just one extra build step that possibly no-one of our users will/can use, especially if the file is constantly out-of-sync.

Dead2:
I am thinking the pdf might as well be replaced by contents in the readme and faq, for the non-linux platforms. I don't think there is anything we want to include that would require it to be in a .pdf file (images, math-formulas, etc).

So a PR updating the FAQ and possibly the readme if the manpage has anything important we want to have there too would be nice.

PR #347, opened 2019-05-28
Rename .h.in to .h.cmakein

nmoinvaz:
This pull request is intended to be rebased and merged after:

- #344 Delete old unix makefiles and scripts
- #343 Added cmake support for cross-compiling on ARM

This change will rename .h.in files that were intended for unix makefiles, to be .h.cmakein which are intended for cmake. 

Currently, the CMakeList.txt will create h.cmakein versions of the .h.in files in the CMakeList.txt code. Then it will process those configs. Since in #344, those old makefiles were removed, this step to convert the unix makefile header templates to cmake header templates is no longer necessary.

Note: It is expected that certain review checks will fail which rely on makefile header template .h.in files.


nmoinvaz:
Closed since configure is not going to be removed.

PR #352, opened 2019-05-29
Fixed cmake formatting

nmoinvaz:
- Unify most of the option() and add_feature_info() statements to the top.
- Updated comments to all begin with proper casing.
- Convert tabs to spaces.

PR #354, opened 2019-05-30
Added cmake support for testing against CVEs

nmoinvaz:
This adds cmake support for running and testing minigzip against the CVE gz archives. It adds a python script that will translate the error code from minigzip into what is acceptable for the test.


test/CVE.py
```diff
@@ -0,0 +1,22 @@
...
+    stdout, stderr = proc.communicate()
+    print stdout
+    #print stderr
+    # We expect 1 error code is OK and we expect 134 or 
+    # similar exit code for a vulnerable failure
+    if proc.returncode == 1 or proc.returncode == 0:
+        print('zlib is not vulnerable to {0} ({1})'.format(sys.argv[1], proc.returncode))
+        exit(0)
+    else:
+        print('zlib VULNERABLE to {0} ({1})'.format(sys.argv[1], proc.returncode))
```
mtl1979: "zlib is VULNERABLE to"

CMakeLists.txt
```diff
@@ -113,6 +114,7 @@ if("${ARCH}" MATCHES "s390x")
     option(WITH_DFLTCC_DEFLATE "Use DEFLATE COMPRESSION CALL instruction for compression on IBM Z" OFF)
     option(WITH_DFLTCC_INFLATE "Use DEFLATE COMPRESSION CALL instruction for decompression on IBM Z" OFF)
 endif()
+add_feature_info(WITH_CVES WITH_CVES "Build test/CVEs")
```
mtl1979: This doesn't look like best place for this line... Should be around line 217 where the other add_feature_info() lines are...

mtl1979:
Python is not installed by default on many systems, so this might fail without descriptive error message.

CMakeLists.txt
nmoinvaz: I put it here because it should merge in nicely after #352. Maybe there is a better way?

mtl1979: The original message on the pull request doesn't mention this pull request depends on #352.

nmoinvaz: Ok I've updated it.

test/CVE.py
nmoinvaz: Actually this is the way that the testCVEinputs.sh has it.

nmoinvaz: The line above it though, is different and does have "is". I have removed it so that both strings are the same as testCVEinput.sh.

mtl1979: As a teacher, I usually prefer proper English, but I don't always have time to go through all pull requests... Consistency is always good as long as it doesn't result in build or run-time errors :)

nmoinvaz:
Sadly cmake test doesn't have a way to say certain exit codes are success or failures. I tried to go with a scripting language that has cross-platform support. It is installed on all of the travis build machines at least, but I have also added a friendly error message when python doesn't exist to the cmake.

mtl1979:
@nmoinvaz Travis build images are in no way standard as they are based on very old Ubuntu release with most of the development related packages added with multiple versions. Windows for example doesn't have Python, Perl or cmake installed by default. Fun thing is that even essential C/C++ development files are not installed by default when installing recent versions of Visual Studio as they are not required for modern applications that only use C#.

nmoinvaz:
I only brought that up to say that it is covered on the test harnesses. I realize that it is not going to be installed on every machine, just like cmake is installed on every machine, or unix bash scripts aren't available on every machine.

mtl1979:
@nmoinvaz I know... Sometimes descriptive error message is better than trying to make the script in multiple scripting languages just to cover all possible operating systems. When we do proper documentation, we can put all the requirements for every step from configuring to building and finally to running the application.

test/CVE.py
nmoinvaz: Haha, no wonder I feel like I'm being graded on an exam! 😂 Yeah, I think here I will just keep it consistent, and if we want to later I can do another pull request for both files that corrects the grammer.

nmoinvaz:
@mtl1979 I am wondering if I should have the `WITH_CVES` switch, since configure script does not have it and seems to automatically build with those tests.

mtl1979:
@nmoinvaz You could change it that it defaults to running the tests if Python is available and default to not running when Python is not installed.

nmoinvaz:
Good idea, I have made that change.

Dead2:
Looks good, have not tested yet.

CMakeLists.txt
```diff
@@ -849,6 +849,26 @@ if (ZLIB_ENABLE_TESTS)
         add_test(example_dict_fuzzer example_dict_fuzzer${CMAKE_EXECUTABLE_SUFFIX} ${ALL_SRC_FILES})
         add_test(minigzip_fuzzer minigzip_fuzzer${CMAKE_EXECUTABLE_SUFFIX} ${ALL_SRC_FILES})
     endif()
+
+    # Use python script to translate exit code 0 or 1 as success
+    set(CVES CVE-2002-0059 CVE-2004-0797 CVE-2005-1849 CVE-2005-2096)
+    find_program(PYTHON_EXECUTABLE python)
```
iii-i: Wouldn't it be better to use python3?

mtl1979: @iii-i We should not depend on specific version of python... It should work on any version...

Dead2: Python3 should be supported, but should not be a requirement.
Also, most distros symlink `python` to `python2` or `python3` depending on their choice of primary python version.

test/CVE.py
```diff
@@ -0,0 +1,22 @@
+import sys
+import subprocess
+
+def main():
+    cmd = sys.argv[2:]
+    proc = subprocess.Popen(cmd,
+                        stdout=subprocess.PIPE,
+                        stderr=subprocess.STDOUT)
+    stdout, stderr = proc.communicate()
+    print stdout
```
iii-i: Shouldn't this be print(stdout) for py3 compatibility?

CMakeLists.txt
mtl1979: @Dead2 My point exactly... it should support both python2 and python3, but through the symlink... Not having a symlink is not valid excuse to default to version that might not exist in all systems we are intending to support.

test/CVE.py
nmoinvaz: Fixed thanks.

nmoinvaz:
I have modified the pull request to use a redirect cmake script similar to #362 instead of using python scripts. It adds the ability to that cmake script to specify `SUCCESS_EXIT` codes.

Dead2:
Thanks :)

PR #360, opened 2019-06-04
Added prefixes to zlib functions for dual zlib/zlib-ng linking.

nmoinvaz:
This pull requests adds PREFIX() to certain functions to allow for dual linking both zlib & zlib-ng.

### DIFF
--- a/arch/arm/armfeature.c
+++ b/arch/arm/armfeature.c
@@ -1,3 +1,4 @@
+#include "zbuild.h"
 #include "zutil.h"
 
 #if defined(__linux__)
--- a/arch/x86/x86.c
+++ b/arch/x86/x86.c
@@ -8,6 +8,7 @@
  * For conditions of distribution and use, see copyright notice in zlib.h
  */
 
+#include "zbuild.h"
 #include "zutil.h"
 
 #ifdef _MSC_VER
--- a/deflate.c
+++ b/deflate.c
@@ -55,7 +55,7 @@
 #include "match_p.h"
 #include "functable.h"
 
-const char deflate_copyright[] = " deflate 1.2.11.f Copyright 1995-2016 Jean-loup Gailly and Mark Adler ";
+const char PREFIX(deflate_copyright)[] = " deflate 1.2.11.f Copyright 1995-2016 Jean-loup Gailly and Mark Adler ";
 /*
   If you use the zlib library in a product, an acknowledgment is welcome
   in the documentation of your product. If for some reason you cannot
@@ -279,11 +279,11 @@ int ZEXPORT PREFIX(deflateInit2_)(PREFIX3(stream) *strm, int level, int method, 
 
     strm->msg = NULL;
     if (strm->zalloc == NULL) {
-        strm->zalloc = zcalloc;
+        strm->zalloc = PREFIX5(calloc);
         strm->opaque = NULL;
     }
     if (strm->zfree == NULL)
-        strm->zfree = zcfree;
+        strm->zfree = PREFIX5(cfree);
 
     if (level == Z_DEFAULT_COMPRESSION)
         level = 6;
@@ -547,7 +547,7 @@ int ZEXPORT PREFIX(deflateResetKeep)(PREFIX3(stream) *strm) {
         strm->adler = functable.adler32(0L, NULL, 0);
     s->last_flush = -2;
 
-    _tr_init(s);
+    PREFIX4(tr_init)(s);
 
     DEFLATE_RESET_KEEP_HOOK(strm);  /* hook for IBM Z DFLTCC */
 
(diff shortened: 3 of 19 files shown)


nmoinvaz:
One thing I am not so sure of is, it uses PREFIX() in zutil.h, which means everything will need to also include zbuild.h. Perhaps we should add include zbuild.h to the top of zutil.h? Or just nuke zbuild.h altogether and include its contents in zutil.h. My opinion is to do the later.

mtl1979:
zbuild.h is supposed to be build-time private header... All external and user code is supposed to hard-code those prefixes instead of relying on checking if compatibility mode is defined. No new code is supposed to use compatibility mode, it's just to get old code working when stock zlib is not needed.

nmoinvaz:
Hmm. Not sure what to do because PREFIX() is still needed whenever including zutil.h. What do you think? Leave as is? Another option would be to move zbuild.h into zutil.h but do this

```
#if !defined(ZNG_USERDEFINED_BUILD)
# if defined(ZLIB_COMPAT)
#  define PREFIX(x)  x
#  define PREFIX2(x) ZLIB_ ## x
#  define PREFIX3(x) z_ ## x
#  define PREFIX4(x) _ ## x
#  define PREFIX5(x) z ## x
#  define zVersion   zlibVersion
#  define z_size_t   unsigned long
# else
#  define PREFIX(x)  zng_ ## x
#  define PREFIX2(x) ZLIBNG_ ## x
#  define PREFIX3(x) zng_ ## x
#  define PREFIX4(x) zng_ ## x
#  define PREFIX5(x) zng_ ## x
#  define zVersion   zlibng_version
#  define z_size_t   size_t
# endif
#endif
```

So, if anybody wanted to they could define their own PREFIX() macros by setting #define ZNG_USERDEFINED_BUILD and PREFIX() before including zutil.h. This is just an idea.

mtl1979:
If zutil.h is externally visible, we need to make separate version for it like we have done for other similar headers... so zutil.h and zutil-ng.h ...

nmoinvaz:
I see what you are saying, zutil.h is not externally visible so it shouldn't matter. Ok, no need for any changes then I think. I think that makes sense why gzendian.h code is not in zutil.h also.

mtl1979:
The idea of being able to disable or override those prefixes has been in my mind too... I just haven't had time to finish implementing the feature properly... Basically it requires having third copy of those externally visible header file pairs, because some data types are incompatible and as such we can't just reuse the header files for compatibility mode.

nmoinvaz:
I could do another PR that would address that issue. What about something similar to this:

- Move zbuild.h contents into zutil.h
- In zutil.h include zbuild.h if ZNG_BUILD_PARAMS was set:

```
#if defined(ZNG_BUILD_PARAMS)
# include "zbuild.h"
#else
# if defined(ZLIB_COMPAT)
#  define PREFIX(x)  x
#  define PREFIX2(x) ZLIB_ ## x
#  define PREFIX3(x) z_ ## x
#  define PREFIX4(x) _ ## x
#  define PREFIX5(x) z ## x
#  define zVersion   zlibVersion
#  define z_size_t   unsigned long
# else
#  define PREFIX(x)  zng_ ## x
#  define PREFIX2(x) ZLIBNG_ ## x
#  define PREFIX3(x) zng_ ## x
#  define PREFIX4(x) zng_ ## x
#  define PREFIX5(x) zng_ ## x
#  define zVersion   zlibng_version
#  define z_size_t   size_t
# endif
#endif
```
Just throwing around ideas.

mtl1979:
Moving things to zutil.h is bad as the macros should be defined in user code before including any zlib-ng headers...

nmoinvaz:
But if doing that, it won't affect any source code internally that includes zutil.h correct?

mtl1979:
For it to work correctly, zlib-ng would have to be statically linked to the binary... Otherwise it would cause linker errors. This also means that the build system for zlib-ng needs to support overriding the prefix.

Dead2:
# Regarding custom prefixes
I think stock zlib supports compiling with a separate prefix, IIRC.
If we want to support custom prefixes, I guess that can be done relatively easily. But it only makes sense to do so in zlib-ng mode I think, and that means having only 2 different prefix-parameters, since the rest of them are identical.

I just reviewed PREFIX2, and I am thinking that might possibly be superfluous. I might be wrong but my immediate thought is that we should always use ZLIBNG_VERSION for checking shared-lib compatibility. If that holds true, then we can remove PREFIX2 and simplify our own set of prefixes to only a single variant, thus also making it easier to support custom prefixes.
Stock zlib requires you to edit zlib.h manually to add prefixes IIRC, and that is probably fine for us too. But if we had a way to generate it, that would be more elegant of course, and would also allow us to remove the two slightly different copies we have today.

Dead2:
Now, regarding this PR. I think we might be slightly off track with this one.
Most of the functions and other symbols in this PR are internal ones, that we do not export as part of our API. This means that we can freely rename them, and there is no need to use the complicated prefix method.
So, why not just rename them and add a static 'zng_' prefix?

nmoinvaz:
I have created a new pull request that should do exactly that.

PR #362, opened 2019-06-14
deflate_medium: fix handling of overlapping matches

### LINKED ISSUE #361: ppc64le: match[2] assertion error
When running the following script on ppc64le:

```
while true; do
    for l in $(seq 1 9); do
        head -c10000000 /dev/urandom | hexdump -C | ./minigzip -$l | ./minigzip -d | pv >/dev/null
    done
done
```

after a few minutes I get a `match[2]?` assertion error.

I patched `z_error` in order to get a core dump, and here is the backtrace:

```
(gdb) bt
#0  0x00007fffa823faf0 in raise () from /lib64/libc.so.6
#1  0x00007fffa8241e6c in abort () from /lib64/libc.so.6
#2  0x00000000100184d8 in z_error (m=0x1001c6d0 "match[2]?") at zutil.c:109
#3  0x000000001000a674 in longest_match (s=0x3eb3c0b0, cur_match=24931) at match_p.h:123
#4  0x000000001000b948 in deflate_medium (s=0x3eb3c0b0, flush=0) at deflate_medium.c:287
#5  0x0000000010006098 in zng_deflate (strm=0x3eb30028, flush=0) at deflate.c:1026
#6  0x0000000010001738 in zng_gzwrite (gz=0x3eb30010, buf=0x7fffd3f2ab58, len=16384) at test/minigzip.c:186
#7  0x0000000010001ddc in gz_compress (in=0x7fffa83e16c0 <_IO_2_1_stdin_>, out=0x3eb30010) at test/minigzip.c:312
#8  0x00000000100027ec in main (argc=0, argv=0x7fffd3f2f068) at test/minigzip.c:516
```

Some values:
```
(gdb) frame 3
(gdb) x/4bx scan-2
0x3eb49b7f:	0x2e	0x2e	0x33	0x2e
(gdb) x/4bx match-2
0x3eb43963:	0x2e	0x2e	0x73	0x61
(gdb) p/x s->ins_h
0x3df3
```

iii-i:
Fixes #361


mtl1979:
Looks good to me...

nmoinvaz:
It is missing cmake variant of the test.

mtl1979:
@nmoinvaz cmake doesn't support using redirection of standard input or standard output with add_test ... As usually we can't use wrapper shell script to circumvent as that would not be portable.

nmoinvaz:
It is possible to use something similar to the python script I created in #354. Maybe it might be worth it to create a generic redirection script.

mtl1979:
@nmoinvaz Generic redirection script might be a good idea if there will be enough tests that use it.

iii-i:
I have added the CMake version of the test based on https://stackoverflow.com/questions/3065220/ctest-with-multiple-commands

This should work on *nix and Windows, but I did not test the latter.

nmoinvaz:
Good idea to use cmake! I think I might make my pull request do something similar using cmake instead of python. 

- For indentation in `execute_process` and `add_test`, it should only be a single tab (4 spaces) indentation on the lines after the first one (at least this is how I've tried to unify the formatting in the CMakeLists.txt).
- run_and_redirect.cmake should go in the cmake directory

In order for the add_test to work on the Travis ARM CI tests you have to do something like:

```
add_test(NAME GH-361
    COMMAND ${CMAKE_COMMAND}
    "-DCOMMAND=${CMAKE_CROSSCOMPILING_EMULATOR};$<TARGET_FILE:minigzip>;-4"
    -DINPUT=${CMAKE_CURRENT_SOURCE_DIR}/test/GH-361/test.txt
    -P ${CMAKE_CURRENT_SOURCE_DIR}/run_and_redirect.cmake)
```
Double quotes around -DCOMMAND are weird but required. `CMAKE_CROSSCOMPILING_EMULATOR` specifies that it should use qemu-arm where necessary. It is defined in the toolchain file in cmake directory.

I have not tested on Windows.

iii-i:
Thanks for the advice! Hopefully this works everywhere now.

nmoinvaz:
- I'm not sure why the change in run-and-redirect.cmake, I think the `if(../win32)` should be how you had it before: `if(WIN32)`

- Because CMAKE_CROSSCOMPILING_EMULATOR is not set on Windows, when it converts the list to a string, it has an empty first item. Instead, we can set our values to a variable and then convert to a string and it seems to work on Windows:

```
    set(GH361_COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:minigzip> -4)
    add_test(NAME GH-361
        COMMAND ${CMAKE_COMMAND}
        "-DCOMMAND=${GH361_COMMAND}"
        -DINPUT=${CMAKE_CURRENT_SOURCE_DIR}/test/GH-361/test.txt
        -P cmake/run-and-redirect.cmake)
```

nmoinvaz:
But also we have to specify full path for -P argument `${CMAKE_CURRENT_SOURCE_DIR}/cmake/run-and-redirect.cmake` for out of source directory builds.

iii-i:
My IDE tried to be too smart with `WIN32` and I haven't noticed that. For now I've simply copied your `run-and-redirect.cmake` version. I've also added the intermediate variable for the test command.

Edit: thanks again, now it fully works!

Dead2:
I think we should have some per-test description (with references to issue/PR) so it is easier to keep track of the regression-tests. I am thinking a one-liner description and url to the issue for example.
(I did notice the GH-361 reference, but I fear that is a bit vague)

I think this should go into test/README

For reference, think about the php self-tests, they have thousands of tests and you get a one-liner description for each failed test. That way it is also easier to figure out what could have made it fail (dependencies etc), and it can tell you something about whether it is a bug you can safely ignore if you never use that part of the API for example.

mtl1979:
As far as I can remember stock zlib wasn't supposed to support overlapping matches, but it was still partially implemented. When the code was reworked, the support was removed. Fixing this issue however is better choice than letting the assertion to fail on certain systems.

iii-i:
I've added the description to `test/INDEX`, since this file was already there for that purpose.

iii-i:
The build has timed out because a couple jobs did not start. I have pushed a no-op change in order to retrigger it.

Dead2:
Ok, it seems like this increases compression runtime of deflate_medium slightly, but it also improves compression in some cases.

Before:
```
 Runs: 7
 Levels: 0-9
 Skipworst: True

 0     100.015%    0.005/0.007/0.010       0.031/0.034/0.038 
 1      49.603%    0.184/0.186/0.190       0.123/0.130/0.135 
 2      38.062%    0.444/0.451/0.455       0.125/0.129/0.134 
 3      37.021%    0.546/0.553/0.560       0.127/0.128/0.131 
 4      35.878%    0.607/0.618/0.630       0.123/0.127/0.129 
 5      35.290%    0.740/0.747/0.791       0.120/0.123/0.127 
 6      35.193%    1.045/1.059/1.102       0.121/0.123/0.129 
 7      34.832%    2.077/2.099/2.163       0.123/0.124/0.126 
 8      34.775%    3.794/3.852/4.062       0.119/0.123/0.128 
 9      34.766%    4.791/4.809/5.138       0.125/0.126/0.129 

 avg1   43.543%                1.438                   0.117
 avg2   37.269%                1.597                   0.126  (lvl 0 excluded)
 tot                          86.284                   7.008
```

After:
```
 0     100.015%    0.005/0.006/0.009       0.030/0.033/0.041 
 1      49.603%    0.185/0.190/0.195       0.128/0.129/0.130 
 2      38.062%    0.447/0.450/0.457       0.128/0.131/0.135 
 3      37.021%    0.546/0.552/0.566       0.123/0.129/0.135 
 4      35.832%    0.606/0.621/0.647       0.118/0.124/0.127 
 5      35.231%    0.735/0.753/0.769       0.121/0.124/0.129 
 6      35.125%    1.052/1.064/1.134       0.121/0.123/0.128 
 7      34.832%    2.095/2.116/2.215       0.118/0.123/0.129 
 8      34.775%    3.799/3.863/4.244       0.120/0.123/0.129 
 9      34.766%    4.808/4.846/4.997       0.123/0.125/0.130 

 avg1   43.526%                1.446                   0.117
 avg2   37.250%                1.606                   0.126  (lvl 0 excluded)
 tot                          86.777                   6.991
```

Dead2:
Thanks :)

Issue #358, opened 2019-06-04
ZLIB-NG Benchmark


powturbo:
zlib-ng is now integrated into [TurboBench - Compression Benchmark](https://github.com/powturbo/TurboBench).
Due to linker conflicts (in tree.c) it is not possible to 
have both zlib-ng (not zlib_compatible) and original zlib  compiled in the same executable.
The current zlib-ng (SSE4.2) in also benchmarked in the [Static/Dynamic web content compression benchmark](https://sites.google.com/site/powturbo/home/web-compression). Better compression and decompression speed than zlib. Decompression is even faster than brotli.

Here is another benchmark with large files
Text:enwik8 (100MB):
<pre><b>      C Size  ratio%     C MB/s     D MB/s   Name  (bold = pareto)</b> MB=1.000.0000
    36475786    36.5   <b>   21.08</b>   <b>  410.17</b>   <b>zlib_ng 9       </b>
    36475796    36.5      17.30     281.10   zlib 9          
    36548925    36.5   <b>   21.13</b>     280.70   <b>zlib 6          </b>
    36879096    36.9   <b>   28.39</b>     277.60   <b>zlib 5          </b>
    37022957    37.0   <b>   36.87</b>   <b>  440.90</b>   <b>zlib_ng 6       </b>
    42298778    42.3   <b>   63.79</b>     273.13   <b>zlib 1          </b>
    56030799    56.0   <b>  187.14</b>     351.88   <b>zlib_ng 1       </b>
   100000004   100.0   <b>14160.29</b>   <b>14351.32</b>   <b>memcpy          </b>
</pre>

Text 58% +binary 42%: silesia.tar
<pre><b>      C Size  ratio%     C MB/s     D MB/s   Name  </b>
    67646927    31.9   <b>   14.15</b>   <b>  464.59</b>   <b>zlib_ng 9       </b>
    67647204    31.9      11.46     349.50   zlib 9          
    68225985    32.2   <b>   26.60</b>     345.96   <b>zlib 6          </b>
    68504069    32.3   <b>   48.32</b>   <b>  479.54</b>   <b>zlib_ng 6       </b>
    69159650    32.6      37.20     339.46   zlib 5          
    77256401    36.5   <b>   74.80</b>     323.19   <b>zlib 1          </b>
   100862036    47.6   <b>  222.74</b>     433.71   <b>zlib_ng 1       </b>
[...]

PR #363, opened 2019-06-21
Add zng_ prefix to internal functions to avoid linking conflicts with zlib.

nmoinvaz:
This pull requests adds zng_ prefix to certain internal functions to allow for dual linking both zlib & zlib-ng. #358, #360

### DIFF
--- a/deflate_medium.c
+++ b/deflate_medium.c
@@ -22,15 +22,21 @@ struct match {
 
 #define MAX_DIST2  ((1 << MAX_WBITS) - MIN_LOOKAHEAD)
 
+static int tr_tally_dist(deflate_state *s, int distance, int length) {
+    return zng_tr_tally(s, distance, length);
+}
+
+static int tr_tally_lit(deflate_state *s, int c) {
+    return zng_tr_tally(s, 0, c);
+}
+
 static int emit_match(deflate_state *s, struct match match) {
     int flush = 0;
-    int mflush = 0;
 
     /* matches that are not long enough we need to emit as literals */
     if (match.match_length < MIN_MATCH) {
         while (match.match_length) {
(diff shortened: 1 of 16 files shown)


deflate_medium.c
```diff
@@ -22,21 +22,15 @@ struct match {
...
-
 static int emit_match(deflate_state *s, struct match match) {
     int flush = 0;
+    int mflush = 0;
 
     /* matches that are not long enough we need to emit as literals */
     if (match.match_length < MIN_MATCH) {
         while (match.match_length) {
-            flush += tr_tally_lit(s, s->window[match.strstart]);
+            zng_tr_tally_lit(s, s->window[match.strstart], mflush);
```
Dead2: These changes should not be part of this commit and likely not this PR.
Commits like the sweeping zng_ prefix changes should have zero effective change of code flow and performance.
The changes in this file might change both, and should get more scrutiny and not be hidden within a big commit.

nmoinvaz: I removed them since renaming them zng_tr_tally_dist and zng_tr_tally_lit would conflict with functions that already carried the same name. I will add them back.

nmoinvaz: I don't quite understand the purpose, of them; expect with the functions, you have additional function overhead for each call.

nmoinvaz:
I have made the PR so it is only renaming of functions like you suggested.

deflate_medium.c
nmoinvaz: I can add a new PR for the removal of those extra functions after this PR, if it is necessary.

nmoinvaz:
Looks like Travis CI stalled on one instance.

Dead2:
Looks good to me, and it passes my tests.

Dead2:
Thanks

PR #366, opened 2019-07-08
IBM Z DFLTCC: fix 31-bit build

iii-i:
* Add machine mode hint for STFLE.
* Adjust offset calculations.


Dead2:
I have no way of verifying this, but since it is your code, I'll merge it. :)

PR #367, opened 2019-07-17
Fixed use of uninitialized value found by memory sanitizer

nmoinvaz:
This issue has shown itself after duplicate gz functions was removed from the test code. Now all the test code is using the same gzip functions and one code path.

See conversation in #348. Credit to @sebpop for fixing it.


Dead2:
Thanks

PR #368, opened 2019-07-18
Dedup inflate macros, fixedtables() and inffixed.h contents

Dead2:
This removes a lot of lines of code from inflate.c, infback.c and inffast.c, moving those into inflate_p.h and making them easier to maintain and keep in sync.

Total number of lines reduced by 99.
Performance seems to be identical on x86_64.
Library size is reduced by 4120bytes or ~2.9%.

A couple of side notes:
- I am uncertain of whether inflate.c, infback.c and inffast.c are actually in sync currently. We don't actually ever test inflateBack(), do we? [I did not check]
- Do we actually want to support running zlib-ng with runtime-generated code tables? It is supposed to save around 2KiB of storage, so that should not be a problem since we are already much bigger than zlib ever was. Removing it would remove some of the maintenance burden, especially considering we don't even test this codepath. I do want to keep the functionality for generating the table files though.
This would also allow us to access the code tables directly instead of doing a state struct lookup on every access, so that could potentially end up a little faster as well.

### DIFF
--- a/infback.c
+++ b/infback.c
@@ -17,9 +17,6 @@
 #include "inffast.h"
 #include "inflate_p.h"
 
-/* function prototypes */
-static void fixedtables(struct inflate_state *state);
-
 /*
    strm provides memory allocation functions in zalloc and zfree, or
    NULL to use the library memory allocation functions.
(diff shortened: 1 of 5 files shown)


mtl1979:
Run-time generated tables already have issues because not all code paths support run-time generated tables. This was evident when we tried to change minimum match from 3 to 4 and SSE-optimized functions stopped working. If we want to continue supporting run-time generation, we have to fix the existing issues.

PR #369, opened 2019-07-20
Rename gzendian.h to zendian.h

nmoinvaz:
In this PR I renamed gzendian.h to zendian.h as I had mentioned in #338. The reason for this is that perhaps when gzendian.h was created it was only used in the gzip source code, but now it is being used outside that source code.


mtl1979:
+1

Issue #336, opened 2019-04-29
Architecture-specific .c files should use target attribute


Myriachan:
When using runtime CPU feature detection, as zlib-ng does, it is best to use `__attribute__((__target__)))` to enable special CPU intrinsics instead of command-line options such as `-msse4`.  The command-line options can cause subtle problems with the One-Definition Rule and lead to crashes on CPUs that do not support those features.

Compiler options such as `-msse4` (GCC/Clang) give compilers carte blanche to use SSE4.2 instructions within functions wherever they please, even if those functions do not use intrinsics.  This becomes a problem with inline functions (likewise template functions in C++).

Consider the following Godbolt example:

https://gcc.godbolt.org/z/G2oNqX

Because `-mssse3` is specified, GCC's vectorizer decides to implement the swaps using the SSSE3 instruction `pshufb`, even though it isn't explicitly using `_mm_shuffle_epi8`.

If `SwapTest` were in a header file that is also #included by a source file not using `-mssse3`, this sometimes is a fatal problem.  The linker would be then free to choose between two different instantiations of `SwapTest`.  The linker might choose the SSSE3 version to be linked in, used by the non-SSSE3 source files, and the program would crash on non-SSSE3 machines.

This can happen with functions declared in system headers, which are not under the control of zlib-ng.

The solution here is to not use the command-line options, and instead use `__attribute__((__target__))` on each function allowed to use processor extensions.

Visual Studio does not understand this attribute.  However, Visual Studio also allows using intrinsics without enabling those CPU features on the command line, so just use nothing in Visual Studio.  Hence something like this:

```
#ifdef _MSC_VER
#define TARGET_ATTRIBUTE_SSE4
#else
#define TARGET_ATTRIBUTE_SSE4 __attribute__((__target__("sse4")))
#endif
...
TARGET_ATTRIBUTE_SSE4
static inline long compare258(const unsigned char *const src0, const unsigned char *const src1) {
```

As a final note, for ARM, Clang's target syntax requires a + sign as the first character, e.g. `"+crypto"` not `"crypto"`, while GCC doesn't.  This requires Clang to be special-cased on ARM.

Dead2:
I agree. This has been way way back on my TODO list actually. Feel free to submit a PR.
Best would probably be to do a compile test during configure/cmake and then enable the feature if setting target attributes is supported by the compiler.
So rather than default it to on, default it to off. Remember there are various compilers out there in regular usage that may or may not support such parameters.

[...]

PR #370, opened 2019-07-20
Added target attribute and compiler detection for intrinsics

nmoinvaz:
This is a first stab at issue #336 to add attribute target to the functions that are using processor specific instructions. I have added HAVE_ATTRIBUTE_TARGET_[FEATURE] and others that set ZNG_TARGET_[FEATURE]. I do the compiler check in cmake. Maybe somebody else can do the configure script. 

I did find some SSSE3 instruction in the code (`_mm_shuffle_epi8`), but I don't see any cmake or configure detection for SSSE3.

I could really find an attribute target for NEON adler instructions.


nmoinvaz:
Some consistency formatting things I didn't do but would have liked to do:

1. Rename HAVE_HIDDEN to HAVE_ATTRIBUTE_HIDDEN
2. Rename HAVE_INTERNAL to HAVE_ATTRIBUTE_INTERNAL
3. Rename X86_SSE4_2_CRC_INTRIN to X86_SSE42_CRC_INTRIN and SSE4_CRC to SSE42_CRC
4. Rename ZLIB_INTERNAL to ZNG_INTERNAL

mtl1979:
Obviously all build errors need to be resolved, before anyone can approve this change...

nmoinvaz:
I can't seem to figure out why the 4 instances on Travis are failing due to ssse3 attribute. Maybe somebody else can take a look.

mtl1979:
@nmoinvaz I found old gcc regression report about ssse3 not working in function target attributes. Has been broken since gcc 5.0.

nmoinvaz:
I found the link to the report. https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57202 
The strange thing is the clang instance is failing too.

mtl1979:
https://gcc.gnu.org/ml/gcc-patches/2013-04/msg00740.html

nmoinvaz:
It looks like it requires `-DWITH_NATIVE_INSTRUCTIONS=ON` to compile correctly - which sets `-march=native` to enable the intrinsics for the native instruction set. Not sure what I should do about that. What should the default behavior be? Always default to `-msse3` flag when detecting `__GNU__` or always turn on `WITH_NATIVE_INSTRUCTIONS` when detecting `__GNU__`. Or maybe default `WITH_NATIVE_INSTRUCTIONS` to `ON`?

mtl1979:
Obviously we need to check if the gcc is broken and avoid using the function target attribute.

nmoinvaz:
Looks like I got it. It wasn't doing the attribute detection properly because the instructions were being optimized away by the compiler I guess. Now it can do a mixed mode, where it does target attributes if it detects it will work correctly, otherwise it will use compiler flags.

On compilers that have problems with target attributes, `-DWITH_NATIVE_INSTRUCTIONS=ON` should get everything working with target attributes. 

Side note `WITH_NATIVE_INSTRUCTIONS` is good, but I see the need for also having something like 'WITH_OPTIM_INSTRUCTIONS` which would use `-march=core2` which are minimum compiler features for the optimizations. Setting `-march=core2` (64-bit) or `-march=prescott` (32-bit) also fixes the issue and gets everything working with target attributes. Having `WITH_NATIVE_INSTRUCTIONS` might use other instructions that wouldn't work on the minimum processor.

nmoinvaz:
Actually looks the issue I am having is due to the use of multiple `__attribute__((__target__("")))` on the functions. 

The compiler doesn't like the use of `ZNG_TARGET_SSE2` with `ZNG_TARGET_SSSE3` and `ZNG_TARGET_PCLMULQDQ` attributes on a single functions. The issue only occurs on two functions in _crc_folding.c_. I can get it to work if the targets are all defined in one attribute `__attribute__((__target__("sse2,ssse3,pclmul")))`, even though GCC allows multiple function attributes.

Not sure how to handle that cleanly?

```
#if defined(HAVE_ATTRIBUTE_SSE2) && defined(HAVE_ATTRIBUTE_SSSE3) && defined(HAVE_ATTRIBUTE_PCLMULQDQ)
#  define ZNG_TARGET_SSE2_SSSE3_PCLMULQDQ __attribute__((__target__("sse2,ssse3,pclmul")))
#endif
```

nmoinvaz:
Travis CI is successful. Not sure what is wrong with buildkite since I don’t have access to it.

Dead2:
Remember that the sse/avx attributes are supersets of eachother.
SSE2 implies SSE, SSE3 implies SSE2 and SSE, etc. So you only need to set the highest one.

mtl1979:
@Dead2 That is mostly true, but in some cases specifying fallback target means alternative version of intrinsic can be used. For example a lot of AVX intrinsics have SSE equivalent that use less registers with same effect.

Dead2:
@mtl1979 Yes, but isn't this to enable us to compile code that requires a certain minimum instruction set, like we have always done? Such as the CRC32 instruction or PCLMULQDQ etc. So we should specify their minimum required instruction set to match the functable tests.
We don't have any functionality for instruction set fallbacks today, so that should not be part of this PR.
GCC also supports making multiple versions of a function, compiled for different instruction sets, and such optimizations would need in-depth research to justify the added complexity by improving speed significantly.

march serves a similar but different purpose, allowing the compiler to use more optimal instructions for tasks that can be solved with more simple instructions, and the distro march setting must be respected here (allowing the user to set a minimal march is also OK, but must not by default override a march set for example by rpm macros.)

PS: I am writing this from a mountain cabin with severely limited internet connectivity, so I have not reviewed any changes, only read comments notifications by email.

mtl1979:
@Dead2 As far as I understand the function target attribute works in conjunction with "march" specified on command line... When "march" is set to lower than highest specified in target attribute, it chooses the instruction set that intersect both choices. 
If there is only one target specified, it will fail with error during compilation. Main purpose of function "target" attributes is to allow compiling inlined code that will use run-time check to choose the right function or group of functions.
It works little similar with "mtune" on command line, but doesn't actually generate code, it only verifies already generated code that was wrapped in "target" pragma block.
If you ask gcc to generate ifunc-style wrapper for multiple targets using single function with attribute "target_clones", it will of course generate new code if the function is used at least once.

nmoinvaz:
@Dead2, let me know which way you would like me to go with this PR after you've had a chance to look at the code. Enjoy the cabin. :-)

nmoinvaz:
I have noticed something interesting. The code in `crc_folding.c` uses `sse2,ssse3,sse4.1,pclmulqdq` while the code in `insert_string_sse.c` and `deflate_quick.c` uses `sse4.2`. So I need to account for the possibility sse4.1 versus sse4.2. What I think is also interesting, is that there is no path that uses sse4.2 crc functions from `crc32.c`.

nmoinvaz:
I have found a separate issue in `CMakeLists.txt` where if using MSVC it does not use `crc_folding.c` because the checks for intrinsics are not being done - and for that matter it isn't using `deflate_quick_sse.c` or `insert_string_sse.c`.

nmoinvaz:
I see the reason why no `sse4.2` path in crc functions from `crc32.c` - it is because intel crc32 intrinics are CRC32-C not CRC32-B. I suppose this is okay for `deflate_quick` and `quick_insert_string` where it might just be comparing two hashes.

functable.c
```diff
@@ -12,14 +12,14 @@
 
 /* insert_string */
 #ifdef X86_SSE4_2_CRC_HASH
-extern Pos insert_string_sse(deflate_state *const s, const Pos str, unsigned int count);
+ZNG_TARGET_SSE42 extern Pos insert_string_sse(deflate_state *const s, const Pos str, unsigned int count);
```
Dead2: IIRC, the target attribute is not needed for the extern definitions, as it does not change the function definition itself.

arch/x86/deflate_quick.c
```diff
@@ -32,7 +32,7 @@
 extern void fill_window_sse(deflate_state *s);
 extern void flush_pending(PREFIX3(stream) *strm);
 
-static inline long compare258(const unsigned char *const src0, const unsigned char *const src1) {
+ZNG_TARGET_SSE2 ZNG_TARGET_SSE42 static inline long compare258(const unsigned char *const src0, const unsigned char *const src1) {
```
Dead2: This is wrong. Since SSE2 is a subset of SSE4.2, you only need to specify SSE4.2.
(That is also why you don't need to do "-msse2 -msse42")

zutil.h
```diff
@@ -20,6 +20,48 @@
...
+#if defined(HAVE_ATTRIBUTE_PCLMULQDQ)
+#  define ZNG_TARGET_PCLMULQDQ __attribute__((__target__("pclmul")))
+#  if defined(HAVE_ATTRIBUTE_SSE2)
+#    if defined(HAVE_ATTRIBUTE_SSE42)
+#       define ZNG_TARGET_SSE2_SSE42_PCLMULQDQ \
+          __attribute__((__target__("sse2,sse4.2,pclmul")))
+#    endif
+#    if defined(HAVE_ATTRIBUTE_SSSE3)
+#      define ZNG_TARGET_SSE2_SSSE3_PCLMULQDQ \
+         __attribute__((__target__("sse2,ssse3,pclmul")))
```
Dead2: sse2 can be removed from this and the above attribute definitions since it is a subset of both ssse3 and sse4.2.
pclmul is IIRC not a superset of either, so it needs to stay.
However, investigating whether pclmul is only supported in cpus with for example sse4.2 for example would mean that we would get more efficient code by also requiring sse4.2 instead of ssse3, by allowing the compiler the use of more recent instruction sets in code that only gets run on cpus that supports it anyway.

Dead2:
To avoid cluttering the function definitions more than needed, I suggest having the target attribute on the line before the function definition.

Also, I am wondering whether it might be easier and more easily understood and maintained to use target pragmas instead. https://gcc.gnu.org/onlinedocs/gcc-4.7.2/gcc/Function-Specific-Option-Pragmas.html#Function-Specific-Option-Pragmas
Those are what I usually use in other projects.

The nice thing about the pragmas is that they don't need any checks in the configure/cmake scripts, and no need for a big chunk of define macros to make the target attribute defines needed either.
The pragmas will also be used by compilers that try to be gcc compliant, and will just be ignored by the rest.
The pragmas can also work on a whole file (making them drop-in the same as the current -msseX parameters), and having the pragma in the file header (after includes) would likely make it pretty easily understood.
The only downside I see to the pragmas is that they do not limit themselves to one function, requiring a pragma reset after each function if you want that, but since we already have whole-file settings, that is not really a big downside for us.

Is there any reason not to use the gcc pragmas to make life easier for ourselves?
I understand that this does not help with MSVC and such, but that is not solved here anyways.

nmoinvaz:
@Dead2 I don't think clang supports GCC target param? But I think clang supports attribute target.

arch/x86/deflate_quick.c
nmoinvaz: Ok. This should be fixed now.

functable.c
nmoinvaz: Good catch. Thanks.

nmoinvaz:
The code inspector application keeps failing. I used to have code-inspector on my minizip repo but too many false positives. Instead I use LGTM. https://lgtm.com/

zutil.h
nmoinvaz: Ok I have made that change.

Dead2:
CodeInspector and a few of the others are just something I am testing out currently, I am aware that there is problems with some of them. I'll add LGTM, thanks for the suggestion.
I'll remove some of them again later, once I know what is worth keeping.

nmoinvaz:
Ok I have made the latest changes. Feel free to review now.

Dead2:
I just checked, and PCLMULQDQ first appeared in a Westmere cpu, and it does not appear in any cpu that does not support SSE4.2. Therefore there is likely no reason to disallow the compiler from utilizing instruction sets up to that if it deems it to be beneficial to performance, since those functions cannot be utilized on SSE4.2-less cpus anyway.

Also, I notice that you have changed what functions gets the flags, I'd prefer that to be a different PR so it can be more easily reviewed and checked for specific performance regressions.

For example currently all functions in deflate_quick gets the full sse4.2 instruction set, and that is fine because we never use deflate_quick on older cpus anyways.
With your current changes, only the functions utilizing sse4.2 intrinsics gets those instruction sets enabled, limiting the compiler to only use older instruction sets for the other functions. This also limits what functions the compiler can inline, since it cannot inline a sse4.2 or ssse3 function into a sse2 function for example.

So, as an initial step, please set the same instruction sets as before for all functions in a file that previously had a file-wide instruction set specified. Changing the instruction sets used per function should be a separate discussion and PR. There might well be benefits to be had there though.

This is also not compiling on my local machine currently, cmake elects to use -msse3 etc instead of the attributes, and the compiler does not recognize the attribute defines and complains about syntax error. Will investigate further later.

Dead2:
I am not sure how this is happening, but the cmake script is somehow only finding support for pclmulqdq as a target attribute, sseX is not detected correctly. This is using GCC9.1 on Fedora 30.

`[ 12%] Building C object CMakeFiles/zlib.dir/functable.c.o
/usr/bin/cc -DHAVE_ATTRIBUTE_PCLMUL -DHAVE_BUILTIN_CTZL -DHAVE_HIDDEN -DHAVE_INTERNAL -DUNALIGNED_OK -DX86_CPUID -DX86_PCLMULQDQ_CRC -DX86_QUICK_STRATEGY -DX86_SSE2 -DX86_SSE4_2_CRC_HASH -DX86_SSE4_2_CRC_INTRIN -DZLIB_DLL -D_LARGEFILE64_SOURCE=1 -D__USE_LARGEFILE64 -I/home/hansr/github/zlib/zlib-ng  -msse4 -O2 -DNDEBUG -fPIC   -std=c99 -o CMakeFiles/zlib.dir/functable.c.o   -c /home/hansr/github/zlib/zlib-ng/functable.c`

Also, the defines in zutil.h needs to have a fallback define that defines the name but without content (like the ZLIB_INTERNAL for example). (This would also have hidden the current failure though)

Also, a suggestion after having attempted to read through the cmake code to find out how this fails; How about we first check for target attribute support, and then check for either -msse3 OR msse3 as a target attribute? Currently we first check for -msse3 and if that is supported we try to upgrade to a target attribute. I think this could make the code simpler and also could help keep the instruction-set-support detection grouped instead of split up like now.

The compile test could potentially even be done using a single code example instead of two separate ones, by using a define to enable the target attribute, like we do in the real compile. I think that should make slightly shorter code total, but would require the test source code to reside in a string that gets shared by the two tests. That could allow us to group all the test source code in a separate area of the script, or even into a separate cmake file. (I think that would be acceptable for this PR, but a future PR might explore restructuring the arch-dependent checks into separate cmake files, as the cmake script is now getting very big and can be hard to grok due to the high level of inter-dependencies going on).

Dead2:
Btw, related to earlier comments, clang supports its own target attribute pragmas
https://clang.llvm.org/docs/AttributeReference.html#id267
See bottom answer for an example:
https://stackoverflow.com/questions/46165752/does-clang-have-something-like-pragma-gcc-target

nmoinvaz:
> So, as an initial step, please set the same instruction sets as before for all functions in a file that previously had a file-wide instruction set specified.

I'm not sure what you are referring to. If a function used a particular instruction set, I set it to have that particular attribute target. Was there a particular function that I did not set correctly? This is the way I had it before too I believe.

> I am not sure how this is happening, but the cmake script is somehow only finding support for pclmulqdq as a target attribute, sseX is not detected correctly. 

In order to debug the issue with cmake you can run `rm CMakeCache.txt & cmake . --debug-output` and see CMakeFiles\CMakeError.log to see the error.

I think maybe your idea for the pragma has some merit, but then the code would have many pragmas for the different compilers. 

Perhaps it would be better for somebody else to take over this who has more experience with target attributes.

Dead2:
@nmoinvaz Thanks for the cmake tip, I didn't know about the --debug-output parameter. I had wiped the cache and the CMakeFiles folder and looked in the log files to try to find the error, but it looked to me like there was no error, that it somehow just skipped the tests.

I will try to make a pragma-based PR soon-ish, then we can see whether that works out fine or not. I might need some help with the cmake file though, we will see. I feel like it has grown to a size where it needs to use include files. (like many other projects use with separate SRCDIR/cmake/*.cmake files where each file handles separate features (For example ssl, jemalloc, dtrace, ctest. I don't know what the best layout would be for us, but I am thinking one for each arch, one for ctest, one for fuzzing, one (or more) for compiler tests should be a good start.)

PR #371, opened 2019-07-22
Fixed optimizations not being used when compiler is msvc.

nmoinvaz:
This issue I mentioned in #370. Optimization code such as `crc_folding.c`, `deflate_quick_sse.c`, `fill_window_sse.c`, and `insert_string_sse.c` were not being compiled when the compiler was MSVC  because the checks for the instrincs were not being done and the `HAVE_[TARGET]_INTRIN` variables weren't being set. I could have simply set `HAVE_[TARGET]_INTRIN` variables to ON manually in the case of MSVC, but it is better this way to have one path for all the compilers (that it runs and checks some code for determination). I have just added MSVC code where necessary in the checks.


Dead2:
Section comments in the CMake file would be nice

nmoinvaz:
I thought about cleaning up the comments in all of the cmake, but I think maybe that is best saved for another PR.

nmoinvaz:
I have added comments for this PR changes only.

CMakeLists.txt
```diff
@@ -400,80 +400,88 @@ if(MSVC)
...
+    set(CMAKE_REQUIRED_FLAGS "${SSE4FLAG}")
+endif()
+check_c_source_compile_or_run(
+    "int main(void)
+    {
+        unsigned val = 0, h = 0;
+    #if defined(_MSC_VER)
+        { __asm mov edx, h __asm mov eax, val __asm crc32 eax, edx __asm mov val, eax }
+    #else
+        __asm__ __volatile__ ( \"crc32 %1,%0\" : \"+r\" (h) : \"r\" (val) );
```
Dead2, line 436: These are not compiler intrinsics, this is inline assembler code. Comment above should be changed, if this is actually what we want to check..?  I have not checked whether we use these inline assembler parts anywhere, but we generally try to avoid doing so because it makes the source harder to maintain and debug.
Also, the comment says we are testing for SSE4, but the asm references crc32 (part of SSE4.2), something the next check says it is testing.

So what does HAVE_SSE42_INTRIN really do (besides not checking for intrinsics at all, something HAVE_SSE42CRC_INTRIN does). I feel like the naming and comments need to get updated to match the actual code here.

I understand that this was also confusing before this commit, but this seems like a good time to fix it up while it is now being reworked.

HAVE_SSE42_INLINE_ASM would perhaps be a better name for this check.

Dead2: HAVE_SSE42CRC_INLINE_ASM would possibly be an even more relevant name, unless we use inline asm for something other than crc32 based on this test.

mtl1979: As target, 32-bit Windows has inline assembler, 64-bit Windows does not have... 64-bit Windows uses MASM or intrinsics.

nmoinvaz: I think the previous naming of HAVE_SSE42_INTRIN was a bit misleading and that HAVE_SSE42CRC_INLINE_ASM is more clear. I have made that change. I have also made a change to use inline asm for msvc in `insert_string_sse.c`.

nmoinvaz:
I have made some more of the suggested changes. Travis is failing due to network issues, some might need to be restarted.

Dead2:
Sorry, I got distracted and managed to completely botch the merge and had to do a force-push to cleanly remove it from develop again.
Could you make a new PR with this please?

PR #374, opened 2019-08-06
Clean up inflate.c/infback.c differences

Dead2:
Remove syntax, whitespace and comment differences between inflate.c and infback.c, making
actual differences much easier to spot, easing maintenance.
This also makes the Codefactor test find fewer but much bigger duplicate code sections, thus improving its accuracy and usefulness.

This also made it visible that we do not have the IBM Z deflate conversion call support in infback.c.
@iii-i Could you make a pull request (based on this commit) that adds the missing support, preferably also keeping the two files as identical as possible (do a side-by-side diff to see what I mean)?

### DIFF
--- a/infback.c
+++ b/infback.c
@@ -39,7 +39,7 @@ int ZEXPORT PREFIX(inflateBackInit_)(PREFIX3(stream) *strm, int windowBits, unsi
     }
     if (strm->zfree == NULL)
         strm->zfree = zng_cfree;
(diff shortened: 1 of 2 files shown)


nmoinvaz:
Looks ok.

iii-i:
@Dead2 I'll look into this. Do you know some users of this interface? I could not find any, so I've decided not to touch it in the first place.

Dead2:
@iii-i Unfortunately I do not, no.
We should probably extend our "testsuite" to add coverage for this though.

PR #375, opened 2019-08-06
Remove BUILDFIXED and MAKEFIXED.

Dead2:
- Remove BUILDFIXED support.
- Split out MAKEFIXED into a separate 'makefixed' util that is easy to use if we want to regenerate/verify inffixed.h

This builds on top of PR/374 and includes that commit, so please review only the second commit here.

### DIFF
--- a/zutil.c
+++ b/zutil.c
@@ -77,9 +77,6 @@ unsigned long ZEXPORT PREFIX(zlibCompileFlags)(void)
 #ifdef ZLIB_WINAPI
     flags += 1 << 10;
 #endif
-#ifdef BUILDFIXED
-    flags += 1 << 12;
-#endif
 #ifdef DYNAMIC_CRC_TABLE
     flags += 1 << 13;
 #endif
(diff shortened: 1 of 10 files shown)


Dead2:
Fixed indentation error in makefixed.c

zutil.c
```diff
@@ -77,9 +77,6 @@ unsigned long ZEXPORT PREFIX(zlibCompileFlags)(void)
 #ifdef ZLIB_WINAPI
     flags += 1 << 10;
 #endif
-#ifdef BUILDFIXED
-    flags += 1 << 12;
-#endif
```
nmoinvaz, line 82: Does it make sense to put in a comment that the 1 << 12 flag was previously used by _BUILDFIXED_ or does it not matter?

Dead2: We have already removed several flags from that area already,
and they are all documented in zlib-ng.h for the zng_zlibCompileFlags function, so I don't think we need to duplicate it here.

PR #378, opened 2019-08-09
Rename #defines for consistency

nmoinvaz:
These are changes I had mentioned in #370. I think it is better this way for the sake of consistency and clearity.

* Rename HAVE_HIDDEN to HAVE_ATTRIBUTE_HIDDEN.
* Rename HAVE_INTERNAL to HAVE_ATTRIBUTE_INTERNAL.
* Rename X86_SSE4_2_CRC_INTRIN to X86_SSE42_CRC_INTRIN.
* Rename X86_SSE4_2_CRC_HASH to X86_SSE42_CRC_HASH.
* Rename SSE4_CRC to SSE42_CRC in CMakeList.
* Rename SSE4DEFLATE to SSE42_DEFLATE in CMakeList.


Dead2:
This needs a rebase.

Also, I'd like to suggest using HAVE_ATTR_ instead of HAVE_ATTRIBUTE_ since it will save 5 chars per attribute per file in the compile log, making each line at least a little shorter. I don't think abbreviating it as ATTR would be confusing for anyone, do you agree?

Also, perhaps renaming SSE42_DEFLATE to DEFLATE_QUICK_SSE42 would be a good idea? That makes it clearer what it actually does, while also leaving naming-room for DEFLATE_QUICK_NEON or similar future implementations. (Not that this name is actually used by anything currently)

mtl1979:
I'm not completely sure grouping preprocessor defines by feature name instead of instruction set is the best idea, until at least we have more than 2 variants for same feature using different instruction sets.

What comes to ATTR vs. ATTRIBUTE, the main concern is do we ever get very large compile logs where we should start saving every possible byte...

Dead2:
@mtl1979 What preprocessor define are you talking about? SSE42_DEFLATE is not a preprocessor define in case that is the one you meant.

mtl1979:
@Dead2 Technically speaking it's unquoted string literal, as in it is concatenated inside add_feature_info in same manner as preprocessor macros can be converted to string containing the macro name using ## operator in C/C++.

Dead2:
And you don't agree that the add_feature_info printout should show DEFLATE_QUICK_SSE42 instead of SSE42_DEFLATE?  I find the latter name to be misleading, since it implies that it affects more than just the quick codepath. What would you suggest instead?

mtl1979:
@Dead2 If we change one feature name, it opens avalanche effect regarding inconsistency between that and the other feature names, which means we would have to rename all of the similar feature names. The new name is also a lot longer than any of the current feature names we use.

Dead2:
@mtl1979 I would rather we have good names in place before a stable release.
Regarding feature names and defines, they should be descriptive names of the related feature, not overly ambiguous, and preferably they should follow a common naming scheme for the type of feature in question (ex: not HAVE_HIDDEN + INTERNAL_ATTR). 

In this case SSE42_DEFLATE is not a good name, and should be replaced with something more specific.
If you have suggestions for names, please list them.

mtl1979:
Obviously there can't be perfect names, but we can make names less ambiguous... if we ignore length of the name, consistent name would be SSE42_DEFLATE_QUICK. The name will always be a compromise, no matter how many similar names we already have.

nmoinvaz:
Another option would be, `HAVE_VISIBILITY_HIDDEN` and `HAVE_VISIBILITY_INTERNAL`. Then for the attribute target we could just do `HAVE_TARGET_SSE2` and `HAVE_TARGET_SSE42`..

Dead2:
+1 HAVE_VISIBILITY_HIDDEN
+1 HAVE_TARGET_SSE42
I am also thinking about USE_DEFLATE_QUICK and USE_DEFLATE_MEDIUM

HAVE_* for compiler features we can use
USE_* for enabling our own features

What do you think, is that any good?

Dead2:
@nmoinvaz Lets go with the naming-scheme I suggested above then.
But would you perhaps want this PR to be delayed to come after the first or both of the other cmake PRs you currently have open? Your choice since it might require some rebasing, although I don't think it touches a lot of the same areas so might not be so bad.

nmoinvaz:
@Dead2, you said you were going to do a PR that did target attributes, so I was going to wait until that was done then update them. I'm not sure if you wanted me to change anything to `USE_DEFLATE_QUICK` or `USE_DEFLATE_MEDIUM`. Anyways i have updated `HAVE_ATTRIBUTE` to `HAVE_VISIBILITY`. There is nothing with `HAVE_TARGET` in this PR.

PR #379, opened 2019-08-12
Turn off enabling tests by default in cmake

nmoinvaz:
This change will turn off `ZLIB_ENABLE_TESTS` by default in cmake (also see #338). When incorporating zlib-ng's CMakeLists.txt it is sometimes not advantageous to build the tests by default. This would also likely need to be done to get a stable release out.

I have updated the travis.yml to set `-DZLIB_ENABLE_TESTS=ON` for cmake instances. There is no ./configure script equivalent.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,20 +13,20 @@ matrix:
     - os: windows
       compiler: clang
       env:
-        - GENERATOR="cmake . "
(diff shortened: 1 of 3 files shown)


nmoinvaz:
@Dead2 are you interested in this PR? If not I can close it. If so then I can rebase.

PR #380, opened 2019-08-15
Only build either shared or static library, but not both

nmoinvaz:
CMake has a flag called `BUILD_SHARED_LIBS` that is used to control whether or not a shared library or static library is built. The benefit of this change is that consumers of zlib-ng's CMakeList.txt can use this flag to control which version of the library they want built. Which is what you might do if using zlib-ng's CMakeList.txt from another project using ExternalProject CMake function.  (See #338)


Dead2:
What specifically is the pros and cons here?
It doesn't seem to change cmake complexity much at least.

mtl1979:
If any distribution wants to use cmake to make zlib-ng binary packages, then this will make building packages more complex. This is however one of those reasons why we would want to keep configure.

nmoinvaz:
You have to think that people will want to use/include our CMakeLists.txt or reference them from their own projects. Having them build both shared and static versions of the library in that use case doesn't make much sense and only adds to overall build time for them. Using `BUILD_SHARED_LIBS` is the preferred way to switch building static versus shared when using CMake. If you look at the changes I made in the CMakeList.txt I think it actually makes the CMake code more clear and simplified.

The con is probably that  we need to add specific Travis test to make sure that CMake builds properly with `BUILD_SHARED_LIBS=ON`. However, all current Travis tests would not build both static and shared which could reduce overall Travis build times. The tests were only ever testing against the shared build. At least by splitting it out this way by using the `BUILD_SHARED_LIBS` it is now possible to switch tests between shared and static.

@mtl1979 - I am not sure I understand the additional complexity that would be required. Wouldn't it just be a matter of using the `-DBUILD_SHARED_LIBS=ON` flag?

mtl1979:
@nmoinvaz As far as I know, the build system works by running the build script once and then grouping the output files from single source package to multiple binary packages. If cmake can only build either static or shared libraries, it means the build step needs to run twice in same output directory and that is forbidden on cmake.

nmoinvaz:
I have made a change so that if `BUILD_SHARED_LIBS` is not set it will output both SHARED and STATIC. If `BUILD_SHARED_LIBS` explicitly set to ON, then only SHARED library will be built. If `BUILD_SHARED_LIBS` explicitly set to OFF, then only STATIC library will be built.

CMakeLists.txt
```diff
@@ -747,64 +747,81 @@ if(MINGW OR MSYS)
...
+    add_library(zlibstatic STATIC ${ZLIB_SRCS} ${ZLIB_ARCH_SRCS}
+        ${ZLIB_DLL_SRCS} ${ZLIB_PUBLIC_HDRS} ${ZLIB_PRIVATE_HDRS})
 
-if(WITH_GZFILEOP)
-    target_sources(zlib PRIVATE ${ZLIB_GZFILE_SRCS})
-    target_sources(zlibstatic PRIVATE ${ZLIB_GZFILE_SRCS})
+    set(ZLIB_INSTALL_LIBRARIES zlib zlibstatic)
+else()
+    add_library(zlib ${ZLIB_SRCS} ${ZLIB_ARCH_SRCS}
+        ${ZLIB_DLL_SRCS} ${ZLIB_PUBLIC_HDRS} ${ZLIB_PRIVATE_HDRS})
```
Dead2: Would it perhaps be smart to combine "${ZLIB_SRCS} ${ZLIB_ARCH_SRCS} ${ZLIB_DLL_SRCS} ${ZLIB_PUBLIC_HDRS} ${ZLIB_PRIVATE_HDRS}" into a single variable ${ZLIB_ALL_LIB_SRCS} or similar?
That would avoid repeating that whole 2-line string 3 times in 8 lines of code. It has no functional difference but would make the code more readable.

CMakeLists.txt
```diff
@@ -747,64 +747,81 @@ if(MINGW OR MSYS)
...
+        LIBRARY DESTINATION "${INSTALL_LIB_DIR}")
 endif()
-if(NOT SKIP_INSTALL_HEADERS AND NOT SKIP_INSTALL_ALL )
-    install(FILES zlib${SUFFIX}.h DESTINATION "${INSTALL_INC_DIR}" RENAME zlib${SUFFIX}.h)
-    install(FILES ${CMAKE_CURRENT_BINARY_DIR}/zconf${SUFFIX}.h DESTINATION "${INSTALL_INC_DIR}" RENAME zconf${SUFFIX}.h)
+if(NOT SKIP_INSTALL_HEADERS AND NOT SKIP_INSTALL_ALL)
+    install(FILES zlib${SUFFIX}.h
+        DESTINATION "${INSTALL_INC_DIR}" RENAME zlib${SUFFIX}.h)
+    install(FILES ${CMAKE_CURRENT_BINARY_DIR}/zconf${SUFFIX}.h
+        DESTINATION "${INSTALL_INC_DIR}" RENAME zconf${SUFFIX}.h)
```
Dead2, line 819: Is there any special reason why you split the lines here and elsewhere? We use 120char line-lengths.

Some of the lines are otherwise unchanged, making reviewers spend time comparing them to spot what changed, only to discover it was only split. For that reason, I'd prefer if style-changes are not mixed with feature patches, so they can be reviewed separately :)

CMakeLists.txt
```diff
@@ -833,8 +852,11 @@ if (ZLIB_ENABLE_TESTS)
     add_executable(switchlevels test/switchlevels.c)
     configure_test_executable(switchlevels)
 
-    add_executable(makefixed tools/makefixed.c inftrees.c)
```
Dead2, line 836: The intention here was having this tool always compile as static, since it does not need anything other than two functions from inftrees.c.
I'd also like to add it as a ctest where it runs and compares its output to the fixed table-header we distribute with zlib-ng, I think this would give us good unit test coverage to the two functions in inftrees.c. Anyone feel free to provide a PR adding that, I have not had time to dig into the required ctest stuff yet, and likely will not for a while.

CMakeLists.txt
nmoinvaz: Done

CMakeLists.txt
nmoinvaz: I know you mention formatting changes. But maybe there is some leway that can be granted every so often. :-) The second line does exceed 120 characters. The first line is formatted so it is the same. The install() line above I had to change. I do try to keep formatting to a minimum.

CMakeLists.txt
nmoinvaz: I see. I have changed makefixed so it doesn't link against zlib.

nmoinvaz: I think that might be a good idea for a PR. Perhaps there is some place we can list all the TODOs.

nmoinvaz:
Ok ready.

CMakeLists.txt
Dead2: I have been reviewing alternatives to Trello (wanted to have a kanban board integrated in github), but I didn't find any that met my requirements. 
With Trello, since you need to be a team member to make changes, and any team member pretty much gets admin rights unless you pay for a business account (anyone can delete the whole board for example).

But as far as I can tell, taiga seems to be a better alternative, so I have set up a board that I am experimenting with.
https://tree.taiga.io/project/dead2-zlib-ng-first-stable-release/kanban
I need to figure out how the permissions system works, but I can probably begin inviting people I have the email address of soon(ish), if I dont find something better that is.

nmoinvaz: That looks good. There is also Jira. https://www.atlassian.com/software/views/open-source-license-request

nmoinvaz: I think we need to also add an item for investigating GitHub Actions when it comes out or even Azure Pipelines. Travis CI is awfully slow and Windows support is limited.

Dead2: I use Jira daily at work, and I think it'd be overkill for this project, at least at this point.
If we were to ditch the GitHub issues list, then Jira and others would look a lot better.

What would be great is a KanBan board integrated in the GitHub interface, but I have not found anything like that in the marketplace unfortunately. If it was integrated, then linking to issues and PRs would be easier, and it would be easier to find it.

Btw, buildkite is able to take travis-like configs, I just haven't been able to dedicate the time convert it. It runs on our own hardware, and can run on windows. I have plans to set it up on a couple of Raspberry Pi that I have (a 1 and a 3), so we can run native tests and also do automated benchmarks to spot regressions (requires some scripting).

If only I could spend 100% of my time on this project ;)

PR #381, opened 2019-08-15
Merge aarch64 and arm cmake sections

nmoinvaz:
This PR just cleans up the cmake a bit and merges the sections for ARCH detection for aarch64 and arm.


Dead2:
Looks good to me, but I'd like someone that has worked on the ARM stuff to review as well. @sebpop @mtl1979 ?

mtl1979:
This change does not take account anymore that Visual C++ does not support aarch64.

sebpop:
The change looks good to me.
I tested the patch on aarch64-linux.

nmoinvaz:
@mtl1979, can you tell me how the change I made doesn't work? I'm not seeing it.

mtl1979:
If compiler is MSVC, arch can never be aarch64, so it should never short-circuit neon detection as default and highest possible target is armv7. Only armv6 had neon as optional feature in some chips.

nmoinvaz:
MSVC does support ARM64 (at least newer versions of MSVC) maybe that is different than aarch64. I haven't had a chance to test it. But when you say short-circuit neon detection. I don't see where the changes I made short-circuit neon detection..

mtl1979:
I haven't installed VS2019 yet to check if it supports aarch64 but at least VS2015 does not support it and VS2017 run-time library is broken so I did uninstall it the same day I tried it. The point is we shouldn't make changes that are not tested.

VS2017 version 15.4 is first to include compiler that targets 64-bit ARM, but I don't know if it is usable for zlib-ng.

nmoinvaz:
I agree we shouldn't make changes that are not tested. I think in the current `develop` branch you will find that MSVC ARM support doesn't work already with CMakeLists.txt. But this PR is not about adding MSVC ARM support. The more I work on this project the less I want to continue doing so.

mtl1979:
I haven't always agreed in all decisions that @dead2 have made but in the end this is his project. Nevertheless when I see something that needs more work, I do comment. Visual Studio support has existed since start of zlib-ng and I see no reason to remove it even though current plans of Microsoft lean towards Linux subsystem for command-line applications.

A lot of things have been broken in CMakeLists.txt but eventually we plan to fix everything... That however does not mean introducing changes that are unlikely to do what originally intended. The sources expect specific defines to exist while compiling for specific compiler and processor pairs. gcc has used two different defines for neon in 32-bit ARM and third for 64-bit ARM.

nmoinvaz:
See #386 instead

PR #383, opened 2019-08-16
use relative header path

cielavenir:
so that header can be included without -I flag, which is potentially useful for embedding whole source.


Dead2:
Thanks, I was considering doing this myself actually. :+1: 
Looks good to me

PR #384, opened 2019-08-16
prefix gz_error and gz_intmax

cielavenir:
to avoid conflict against libz.a.

### DIFF
--- a/gzguts.h
+++ b/gzguts.h
@@ -151,7 +151,7 @@ typedef struct {
 typedef gz_state *gz_statep;
 
 /* shared functions */
-void ZLIB_INTERNAL gz_error(gz_state *, int, const char *);
+void ZLIB_INTERNAL zng_gz_error(gz_state *, int, const char *);
 
 /* GT_OFF(x), where x is an unsigned value, is true if x > maximum z_off64_t
    value -- needed when comparing unsigned to z_off64_t, which is signed
@@ -159,8 +159,8 @@ void ZLIB_INTERNAL gz_error(gz_state *, int, const char *);
 #ifdef INT_MAX
 #  define GT_OFF(x) (sizeof(int) == sizeof(z_off64_t) && (x) > INT_MAX)
 #else
-unsigned ZLIB_INTERNAL gz_intmax(void);
-#  define GT_OFF(x) (sizeof(int) == sizeof(z_off64_t) && (x) > gz_intmax())
+unsigned ZLIB_INTERNAL zng_gz_intmax(void);
+#  define GT_OFF(x) (sizeof(int) == sizeof(z_off64_t) && (x) > zng_gz_intmax())
 #endif
 
 #endif /* GZGUTS_H_ */
--- a/gzlib.c
+++ b/gzlib.c
@@ -29,7 +29,7 @@ static void gz_reset(gz_state *state) {
         state->how = LOOK;          /* look for gzip header */
     }
     state->seek = 0;                /* no seek request pending */
-    gz_error(state, Z_OK, NULL);    /* clear error */
+    zng_gz_error(state, Z_OK, NULL);    /* clear error */
     state->x.pos = 0;               /* no uncompressed data yet */
     state->strm.avail_in = 0;       /* no input data yet */
 }
@@ -319,7 +319,7 @@ z_off64_t ZEXPORT PREFIX(gzseek64)(gzFile file, z_off64_t offset, int whence) {
         state->eof = 0;
         state->past = 0;
         state->seek = 0;
-        gz_error(state, Z_OK, NULL);
+        zng_gz_error(state, Z_OK, NULL);
         state->strm.avail_in = 0;
         state->x.pos += offset;
         return state->x.pos;
@@ -461,7 +461,7 @@ void ZEXPORT PREFIX(gzclearerr)(gzFile file) {
         state->eof = 0;
         state->past = 0;
     }
-    gz_error(state, Z_OK, NULL);
+    zng_gz_error(state, Z_OK, NULL);
 }
 
 /* Create an error message in allocated memory and set state->err and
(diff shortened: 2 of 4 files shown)


mtl1979:
Prefixing internal symbols should not be necessary unless ZLIB_INTERNAL is no-op...

cielavenir:
then what is the reason of https://github.com/zlib-ng/zlib-ng/pull/363 ...

mtl1979:
gz_intmax should not be externally visible as it is defined only in internal header. What I'm saying is that ZLIB_INTERNAL might not be the correct attribute macro for it if it causes symbol conflict.

mtl1979:
I investigated this a bit and it seems to be limitation in "ar". People suggest using "ld -r" if using gcc or compatible.

Dead2:
Annoying that we need it, but I agree that we should fix this problem.
However, what about we rename the functions, but add a define using the old name to translate to the prefixed name? That way we don't change the code calling the function, but the symbol gets changed.
`#define gz_error zng_gz_error`
I guess we could have done that for some of the other internal functions as well.

mtl1979:
I've seen the rename/define trick used with a lot of MSVC code. We obviously can't use workaround that only works with gcc or compatible.

mtl1979:
I think easiest and cleanest solution would be to make the two functions inlined and move the definition to header... That way the symbols don't get generated in the object file and propagated to the static library by gnu ar.

dankegel:
Is the goal to allow dual static linking?

mtl1979:
If we allow dynamic linking of both zlib and zlib-ng, personally I think we should allow also static linking as it would allow us to do tests that are not possible with dynamic linking.

dankegel:
Only allowing dual dynamic linking opens fewer cans of worms.
If zlib-ng compat becomes widespread, allowing dual static linking might mean dual static linking of zlib-ng compat with regular zlib-ng, even across versions, which suggests even internal identifiers would need prefixing.
Unless there is some magic in .a linking I'm unaware of, anyway.

Taking the idea to the max, we could also support unity builds of apps with both zlib and zlib-ng.  In that world  (where all source files are essentially concatenated),  the practice of depending on file scope being useful for preventing clashes fails.

mtl1979:
@dankegel There is well known trick to create static libraries without gnu ar... Essentially gnu ld can be used to combine multiple object files and at the same time rewrite the symbol table to hide internal symbols.

nmoinvaz:
There is a CI instance that checks dual linking

dankegel:
This bug lacks a reproducible recipe for demonstrating the problem.  One would help considerably.

mtl1979, are you proposing that we use that gnu ld trick to create static libraries routinely?  That would be... unusual.  What about other operating systems that don't use gnu ld?

The #define renaming trick should work everywhere except unity builds (where defines cannot change values for different files).

The easiest resolution would be to create a bug "Dual static linking fails", including a test case, and target it at a future release.  It doesn't seem to be a crucial feature offhand.  

zlib-ng/test doesn't yet have a test that reliably links to both static system and static zlib-ng and verifies that they both work, does it?

mtl1979:
@dankegel No. I'm just iterating all the options with their upsides and downsides... Then we can collectively choose the best fix.

dankegel:
The first step is probably for @cielavenir to explain how to reproduce the bug he was seeing.

nmoinvaz:
@dankegel there is a CI instance test that links against system zlib and a CMake option for it.

nmoinvaz:
The test name is `Ubuntu GCC Link Zlib` and cmake option `ZLIB_DUAL_LINK`. I created them sometime after this ticket was reported. I never really encountered this error.

dankegel:
@nmoinvaz, I know about that option, and its use in CI, but I don't think it causes tests to call both zlib and zlib-ng routines, so it does not verify that they both work.

@cielavenir, please file an issue describing a situation where this matters in the real world, with a recipe for how to reproduce the problem.

Dead2:
Closing until a real-world testcase can be provided.

nmoinvaz:
@dankegel the test did discover other issues at the time, but you are probably right, a better test would be a simple .c file that uses functions from both.

dankegel:
Yeah, I can imagine the existing test might find a few problems, but it's rather limited.

cielavenir:
@dankegel 

1. build zlib-ng cmake with `ZLIB_ENABLE_TESTS=ON ZLIB_DUAL_LINK=ON`
2. in CMakeFiles/minigzip.dir:

```
ar x /usr/lib/x86_64-linux-gnu/libz.a gzlib.o
gcc gzlib.o gzlib.c.o
```

then you will get:

```
gzlib.c.o: In function `gz_error':
gzlib.c:(.text+0x870): multiple definition of `gz_error'
gzlib.o:(.text+0xb80): first defined here
```

My project supports cross-compilation, so I took the strategy to compile both zlib and zlin-ng from source.

cielavenir:
The point is that this command does not cause multiple definition:

```
gcc /usr/lib/x86_64-linux-gnu/libz.a gzlib.c.o
```

In this way ZLIB_DUAL_LINK=ON could not reproduce the issue (even if ZLIB_LIBRARY_RELEASE is set to /usr/lib/x86_64-linux-gnu/libz.a, not so).

nmoinvaz:
@cielavenir is there a way we can reproduce it in the CI environment? Some CMake flag that turns on use of ar? Forgive me I’m not super knowledgeable on Linux.

dankegel:
@cielavenir, linking two .o files in isolation is not really a test.  Can you provide a real example, one that does not involve unpacking the .a file?

nmoinvaz:
I tried `-DZLIB_ENABLE_TESTS=ON -DZLIB_DUAL_LINK=ON` with manually compiled zlib in both Debug and Release modes and there was no error at least on Windows. I thought maybe the fact that you compiled zlib from sources made a difference but it looks like it doesn't.

cielavenir:
@nmoinvaz linking zlib-ng with manually compiled .a (or .lib) works but linking with .o (or .obj) does not work.

dankegel:
Then don't do that.

Dead2:
@cielavenir Correct me if I am understanding this wrong.
1. Extract gzlib.o from libz.a (stock zlib)
2. Link extracted gzlib.o with gzlib.c.o from zlib-ng
3. Errors occur

Now, what I wonder is what the motivation to do this is? Why do you not link with the .a file instead?
Knowing why it matters helps understand the issue at hand.

cielavenir:
I prepared https://github.com/cielavenir/zlib-ng/blob/issue384/compile.sh; just running it (after installing submodule) should trigger the error.

dankegel:
You wrote

```
gcc -DWITH_GZFILEOP -O2 *.c zlib/*.c
```

Is your motivation that you don't want to be bothered using zlib-ng's build system?

e.g. you just want a single source file you can
include in your project, rather than invoking zlib-ng's build system?

If so, what build system are you using in your project?

PR #385, opened 2019-08-16
Fixed GCC version for bswap intrinsics

cielavenir:
according to https://stackoverflow.com/a/15034284/2641271 , 4.3 is least version supporting __builtin_bswap32. Also https://gcc.gnu.org/onlinedocs/gcc-4.8.0/gcc/Other-Builtins.html (and 4.7.0 manual) indicates __builtin_bswap16 is supported by gcc 4.8. For portability, I currently select 4.8.

### DIFF
--- a/zutil.h
+++ b/zutil.h
@@ -175,7 +175,7 @@ void ZLIB_INTERNAL   zng_cfree(void *opaque, void *ptr);
 #  define ZSWAP64(q) _byteswap_uint64(q)
 
 #elif defined(__Clang__) || (defined(__GNUC__) && \
-        (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 3)))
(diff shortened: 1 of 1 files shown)


mtl1979:
We might want to move checking these to configure/CMakeLists.txt, so it selects the individual builtins always when available...

Dead2:
@mtl1979 I must confess that I prefer preprocessor checks before cmake checks. And the check could be split if we want to support older than GCC 4.8.
EL6 is based on GCC 4.4, EL7 is based on 4.8. Not sure whether it is worth targeting anything older than 4.8 currently though, as even that is getting pretty old.

mtl1979:
gcc 4.x, 5.x, 6.x and 7.x have quite a lot of dud releases that miscompile code. That pretty much warrants special handling as most distros have one of them as default. 

Splitting the test is possibility but I'm not totally confident that afterwards the code is still clean and as readable as before this change.

Dead2:
This is a minimal fix to a real problem, so I'll pull this as-is.
Thanks :)

PR #386, opened 2019-08-19
Changes to support compilation with MSVC ARM & ARM64

nmoinvaz:
These changes are made to support cross-compilation ARM and ARM64 for MSVC on Windows. Currently, it doesn't look like Travis CI supports that configuration, otherwise I would have added the tests.

Of note, is no longer the use of arch/arm/ctzl.h since MSVC supports built-in `_BitScanForward` on ARM and ARM64. Could not use `_arm_clz` or `_arm_rbit` in arch/arm/ctzl.h due to them not being defined in arm64intr.h and only in armintr.h.


match_p.h
```diff
@@ -33,10 +33,8 @@
 
 
 #if defined(_MSC_VER) && !defined(__clang__)
-# if defined(_M_IX86) || defined(_M_AMD64) || defined(_M_IA64)
+# if defined(_M_IX86) || defined(_M_AMD64) || defined(_M_IA64) ||  defined(_M_ARM) || defined(_M_ARM64)
 #  include "arch/x86/ctzl.h"
```
mtl1979: This is bad... We need to move this header to more generic location.

Dead2:
Looks good to me, but I'd like to let this wait for your other big cmake PR. :)

Dead2:
I will merge this tomorrow if there are no new comments by then.

mtl1979 (changes requested):

match_p.h
nmoinvaz: What do you mean?

mtl1979: We need to move the header out of arch/x86 and possibly rename it to make it clear it is used for multiple archs using Visual C++ compiler.

nmoinvaz: Hmm. Are you suggesting to put it in the arch/generic folder?

nmoinvaz: Or I could rename it to arch/generic/msvc_ctzl.h

mtl1979: We haven't really decided how we are going to use arch/generic as most of the generic *_c functions are still in the project root directory.

nmoinvaz: Yeah I don't really have a good idea here.

mtl1979: Maybe /msvc_ctzl.h until we start to populate arch/generic...

nmoinvaz: @Dead2 thoughts?

Dead2: I am thinking that this is not really arch-specific, and it does not really fit into generic either.
The thought behind generic was to have for example fill_window_c there, and that would be used if there was no fill_window_X implementation in the activated arch. So sort of a place for the fallback code that has arch-specific versions. It would also simplify keeping it and the arch-specific versions in sync properly, both through better visibility and by easier diffing.

In this case, I suggest moving it into /fallback-builtins.h as what it is is a fallback implementation of a compiler builtin. And we might end up getting more of those at some point. We might also need arch-specific builtins at some point, those should go into the arch folder.

nmoinvaz: Done

PR #387, opened 2019-08-21
[Travis] Add codecov.io and several new tests + misc

Dead2:
Travis config changes:
- Add codecov.io coverage support
- Remove repeated identical test
- Add a GCC-9 test (one benefit is it gives much better warnings/error messages)
- Add some variety to some of the tests
- Add section comments and whitespace, making the config file more readable and using the same style for arm and other arches

See the two commits separately for easier reviewing, if anyone objects I can easily split these into several separate PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,12 +13,14 @@ env:
 matrix:
   include:
     # Windows tests
+    #   clang/cmake
     - os: windows
       compiler: clang
       env:
         - GENERATOR="cmake . "
         - MAKER="cmake --build . --config Release"
         - TESTER="ctest --verbose -C Release"
+
     - os: windows
       compiler: clang
       env:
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Getting some warnings on Windows `unrecognized option '/-coverage'`

```
  cl : Command line warning D9002: ignoring unknown option '-g' [C:\Users\travis\build\zlib-ng\zlibstatic.vcxproj]
  cl : Command line warning D9002: ignoring unknown option '--coverage' [C:\Users\travis\build\zlib-ng\zlibstatic.vcxproj]

  LINK : warning LNK4044: unrecognized option '/-coverage'; ignored [C:\Users\travis\build\zlib-ng\switchlevels.vcxproj]
```

Dead2:
@nmoinvaz Ah, as you noticed, the travis config suggested that gcc/clang was being used on windows as well, and I assumed that was correct.

Is it possible to set env variables depending on what compiler is chosen in the travis matrix perhaps? Otherwise, perhaps we could detect compiler and add the coverage flags if gcc/clang. The last option would be to duplicate the string for every test in the matrix, but I'd like to go for a more elegant solution.

It would also be nice to add coverage support for msvc, since some code is different on windows.

I'll add these to the KanBan board.

PR #388, opened 2019-08-22
Added regression test for zng_deflateEnd -3

nmoinvaz:
This issue adds regression test for zng_deflateEnd -3 return code mention in #382.

PR #382, opened 2019-08-16
[deflate_quick] try to consume available data if Z_FINISH is used

cielavenir:
the following test program fails with `zng_deflateEnd -3` without this patch.

```c
#include <stdio.h>
#include <sys/stat.h>
#include "zlib-ng.h"
int main(){
	FILE *f=fopen("FAQ.zlib","rb");
	struct stat st;
	fstat(fileno(f),&st);
	char *buf=malloc(st.st_size);
	char *compbuf=malloc(st.st_size);
	fread(buf,1,st.st_size,f);
	fclose(f);

	zng_stream z;
	int status;
	z.zalloc = Z_NULL;
	z.zfree = Z_NULL;
	z.opaque = Z_NULL;
	const int level=1;

	if((status=zng_deflateInit2(
		&z, level , Z_DEFLATED, -MAX_WBITS, level, Z_DEFAULT_STRATEGY
	)) != Z_OK){
		return status;
	}

	z.next_in = buf;
	z.avail_in = st.st_size;
	z.next_out = compbuf;
	z.avail_out = st.st_size;

	status = zng_deflate(&z, Z_FINISH);
	int csize = st.st_size - z.avail_out;
	status = zng_deflateEnd(&z);
	if(status != Z_OK){
		printf("zng_deflateEnd %d\n",status);
		return 1;
	}
	fwrite(compbuf,1,csize,stdout);
}
```

PR #390, opened 2019-08-23
Emit end block results in -3 return code for deflate_quick. 

nmoinvaz:
Other deflate strategies don't emit anything in this conditional block. See #382, #388.


Dead2:
Hmm, does this result in correct behavior though?
Seems like this will need some in-depth review and testing.

Edit: The comments in #388 sheds more light on this, it is not as risky as I first thought.

Dead2:
And does this mean that #382 should be reverted?

Edit: This was already answered in a comment in #388, this fix is in addition to that, not instead of. :)

Dead2:
This fails in my tests during decompression of a 40MiB deflate_quick compressed file.
```
./minigzip -d -c /tmp/zlib-testfil.gz
/tmp/zlib-testfil.gz: invalid stored block lengths
```

PR #391, opened 2019-08-23
Changes based on CloudFlare's fork

Dead2:
The first commit (match_p.h) is simply moving a if check earlier, and is identical to what CF did.

The second commit is based on the idea of what CF did, but I did not just unroll the define, I changed the define to hopefully achieve the same improvement, while keeping the ability to enable debugging.
For the debugging variant, there is no longer a function that duplicates the define's content, but the necessary debugging code is inserted into the top of the define instead. This is not for performance, but for simplicity, so we don't need to maintain as much duplicated code.

I am uncertain about merging this though, as I was unable to see any performance difference in my local tests, however those were done on x86_64 with GCC 9.1. There might be performance differences on other architectures or compiler, or older GCC versions might not be able to optimize this as much.

I also removed the following lines from the define, as I don't see why they are needed. Did we remove the need at some point when changing types perhaps?
int len = length;
int val = (int)value;

So, I am looking for feedback here, does this change make any sense to you?
And does it make any difference in performance on your platforms?

### DIFF
--- a/trees.c
+++ b/trees.c
@@ -474,6 +474,10 @@ static void send_tree(deflate_state *s, ct_data *tree, int max_code) {
     if (nextlen == 0)
         max_count = 138, min_count = 3;
 
+    // Temp local variables
+    int filled = s->bi_valid;
+    uint16_t bit_buf = s->bi_buf;
+
     for (n = 0; n <= max_code; n++) {
         curlen = nextlen;
         nextlen = tree[n+1].Len;
@@ -481,25 +485,25 @@ static void send_tree(deflate_state *s, ct_data *tree, int max_code) {
             continue;
         } else if (count < min_count) {
             do {
-                send_code(s, curlen, s->bl_tree);
+                send_code(s, curlen, s->bl_tree, bit_buf, filled);
             } while (--count != 0);
 
         } else if (curlen != 0) {
             if (curlen != prevlen) {
-                send_code(s, curlen, s->bl_tree);
+                send_code(s, curlen, s->bl_tree, bit_buf, filled);
                 count--;
             }
             Assert(count >= 3 && count <= 6, " 3_6?");
-            send_code(s, REP_3_6, s->bl_tree);
-            send_bits(s, count-3, 2);
+            send_code(s, REP_3_6, s->bl_tree, bit_buf, filled);
+            send_bits(s, count-3, 2, bit_buf, filled);
 
         } else if (count <= 10) {
-            send_code(s, REPZ_3_10, s->bl_tree);
-            send_bits(s, count-3, 3);
+            send_code(s, REPZ_3_10, s->bl_tree, bit_buf, filled);
+            send_bits(s, count-3, 3, bit_buf, filled);
 
         } else {
-            send_code(s, REPZ_11_138, s->bl_tree);
-            send_bits(s, count-11, 7);
+            send_code(s, REPZ_11_138, s->bl_tree, bit_buf, filled);
+            send_bits(s, count-11, 7, bit_buf, filled);
         }
         count = 0;
         prevlen = curlen;
@@ -511,6 +515,10 @@ static void send_tree(deflate_state *s, ct_data *tree, int max_code) {
             max_count = 7, min_count = 4;
         }
     }
+
+    // Store back temp variables
+    s->bi_buf = bit_buf;
+    s->bi_valid = filled;
 }
 
 /* ===========================================================================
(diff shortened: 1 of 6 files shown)


mtl1979:
Only reason for using functions instead of macros is that stack traces in debug code are easier to read. This is not an issue in release code as there is no symbol table anyways.

Dead2:
I don't see send_bits as critical in stacktraces really. It was possibly awesome during early development of zlib, but I don't think it is as important to us now. With the trace code still in place, and much better debug utilities now, I think we can safely remove that.

I remembered why the macro used the two temporary variables btw, it is because that is the only way to guarantee correct behavior if the input parameters are not guaranteed to be a plain variable. Imagine if the input parameter was a function call, that would get called multiple times, or even a val++ could be bad. So a macro would deviate from the way a function call behaves.
We don't use the macro that way though, and if we remove the function, then there would be no such difference in behavior between debug and release build either.

trees.c
```diff
@@ -770,12 +778,12 @@ static void send_all_trees(deflate_state *s, int lcodes, int dcodes, int blcodes
     Assert(lcodes >= 257 && dcodes >= 1 && blcodes >= 4, "not enough codes");
     Assert(lcodes <= L_CODES && dcodes <= D_CODES && blcodes <= BL_CODES, "too many codes");
     Tracev((stderr, "\nbl counts: "));
-    send_bits(s, lcodes-257, 5); /* not +255 as stated in appnote.txt */
-    send_bits(s, dcodes-1,   5);
-    send_bits(s, blcodes-4,  4); /* not -3 as stated in appnote.txt */
+    send_bits(s, lcodes-257, 5, s->bi_buf, s->bi_valid); /* not +255 as stated in appnote.txt */
+    send_bits(s, dcodes-1,   5, s->bi_buf, s->bi_valid);
+    send_bits(s, blcodes-4,  4, s->bi_buf, s->bi_valid); /* not -3 as stated in appnote.txt */
```
mtl1979: This seems inefficient... Too many indirect reads and writes...

mtl1979 (changes requested):

trees.c
Dead2: True, but how often do we call send_all_trees()?
I guessed it would not matter for performance, that is why I did not use temp variables there.
If it does matter for performance, we can easily add that.

Dead2:
Still looking for benchmarks for older x86/x86_64 and other arches.
Want to make sure this does not end up as a regression.

mtl1979:
@dead2 It's actually really easy to count how many times each function is called. I don't really care when there is just one or two indirect memory accesses but that code has a loop.

Dead2:
Added suggested changes and rebased on latest develop.

trees.c
```diff
@@ -769,21 +777,31 @@ static void send_all_trees(deflate_state *s, int lcodes, int dcodes, int blcodes
...
+    // Store back temp variables
+    s->bi_buf = bit_buf;
+    s->bi_valid = filled;
+
     send_tree(s, (ct_data *)s->dyn_ltree, lcodes-1); /* literal tree */
     Tracev((stderr, "\nlit tree: sent %lu", s->bits_sent));
 
     send_tree(s, (ct_data *)s->dyn_dtree, dcodes-1); /* distance tree */
     Tracev((stderr, "\ndist tree: sent %lu", s->bits_sent));
+
```
mtl1979: Do we really need empty line here?

Dead2: Yes, in fact we should have at least four. Just kidding, I'll amend the commit.

trees.c
mtl1979: I counted 6 indirect reads or writes per each call if not using temporary variables. Indirect reads and writes are issue on processor and compiler combinations that assume struct members are volatile by definition. 

This means the address of the pointer (relative to first member of struct) is recalculated each time and that causes slowdown. With temporary variable, only the value of the variable is modified in CPU register but not stored back in memory containing the struct contents until compiler runs out of usable registers or the current scope ends.

trees.c
mtl1979: I was thinking why there is extra line added when nothing else is changed in surrounding lines... Things like this happen usually when one commit in patch set reverts earlier addition.

trees.c
Dead2: This has already been addressed

trees.c
Dead2: I was only joking, and it has been corrected..

mtl1979:
Looks OK to me now...

Dead2:
rebased

PR #393, opened 2019-08-23
Import fix for a very rare potential crash in crc_folding_sse

Dead2:
This comes at a slight performance penalty of around 0.2% for compression on a skylake with GCC-9


mtl1979:
This seems to introduce code duplication, although only one line is duplicated...

Issue #284, opened 2019-01-17
x86_64 deflate_quick fails on chromium-benchmark fireworks.jpeg


sebpop:
In the past I have seen some errors on x86_64 with the snappy chromium-benchmark at compression level 1.  Check that these errors are fixed.

sebpop:
I still see it failing at level 1 with
```
fireworks.jpeg                           : compress failed (0)
```

sebpop:
It passes on aarch64 and fails on x86_64 which points to some bug in intel's patches for `deflate_quick`.

sebpop:
When replacing `deflate_quick` with `deflate_fast` it passes.

mtl1979:
It would be good to know why deflate_quick fails... That pretty much requires enabling debug output and possibly adding more sanity checks.

sebpop:
The benchmark uses an estimate of the compress size to resize a buffer before decompressing.
In the case of fireworks.jpeg with deflate_quick we grow the output by 5.2%: 
```
test/bench/snappy/fireworks.jpeg         : GZIP -1: [b 1M] bytes 123093 -> 129548 105.2% comp  34.9 ( 34.9) MB/s uncomp 139.7 (139.7) MB/s
```
The benchmark passes if we set the estimated compression size to double of the input:
```
 z_size_t ZEXPORT PREFIX(compressBound)(z_size_t sourceLen) {
-    return sourceLen + (sourceLen >> 12) + (sourceLen >> 14) + (sourceLen >> 25) + 13;
+    return 2 * sourceLen;
```
When printing the original estimated size, we see far less than 5%:
```
 zng_compressBound (sourceLen=123093)
(gdb) finish
Value returned is $7 = 123143
```
The fix is to special case intel's deflate_quick to have a different function to estimate its output buffer size.

sebpop:
There is at least one more bug that we see with lcet10.txt and others that returns -3 which is different than the jpeg which was returning 0 instead of 1.

mtl1979:
Maybe we should adjust one of the shifts to make it return slightly larger value... That way we don't completely butcher the logic inside compressBound...

```
    return sourceLen + (sourceLen >> 4) + (sourceLen >> 12) + (sourceLen >> 14) + (sourceLen >> 25) + 13;
```

This should allow 6.25~6.28% increase.

bmrzycki:
[...]

PR #396, opened 2019-09-07
Fixed chunkmemset should be called on larger chunks assert

nmoinvaz:
When compressing fireworks.jpeg using deflate level 1 and mem level 1, during inflate the following assert would be hit:

```
chunkmemset should be called on larger chunks
```

This PR fixes that by providing a path for copying lengths that are less than _sizeof(uint64_t)_ when using INFFAST_CHUNKSIZE. Also see #284.

### DIFF
--- a/inffast.c
+++ b/inffast.c
@@ -307,7 +307,9 @@ void ZLIB_INTERNAL zng_inflate_fast(PREFIX3(stream) *strm, unsigned long start) 
                        operations can write beyond `out+len` so long as they
                        stay within 258 bytes of `out`.
                     */
-                    if (dist >= len || dist >= INFFAST_CHUNKSIZE)
+                    if (len < sizeof(uint64_t))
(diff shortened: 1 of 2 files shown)


nmoinvaz:
It would be good to get some reviewers on my usage of the parameters passed into these functions. I'm pretty sure they are right.

Dead2:
Yep, I'd like someone who knows the inffast chunkcopy code better than me to review this.

Dead2:
Did you happen to benchmark this to see whether it made any difference?

nmoinvaz:
I didn't do any benchmarks, but I saw the code I was using wasn't working in a certain instance so I decided to close the PR. The assert still exists in the code, because len < sizeof(uint64_t). In the first change lines I made, that would be okay because of the comment above (although an assert is still thrown).  I'm not sure if it is better to just remove the asserts or what.

PR #397, opened 2019-09-07
inflate returns Z_BUF_ERROR during Z_FINISH when data left to be written

nmoinvaz:
See madler/zlib#446 for description and test case.


Dead2:
I see there was a reply from upstream, is this still relevant?

nmoinvaz:
I want to reinvestigate it when I have some time.

nmoinvaz:
I revisited it. There may have been something wrong I was doing while testing things originally. `inflate` with `Z_FINISH` will return `Z_OK` even with 1 byte buffer, so long as it isn't outputting any data. In other words, if `in > 1 and out == 0`. So I think perhaps that is what I was seeing. Then only when it needed to write it would return `Z_BUF_ERROR`. This probably didn't occur that often in my tests because everything was handled by a previous call to `inflate` using `Z_NO_FLUSH`.

PR #398, opened 2019-09-07
Added cmake tests to verify output of makefixed

nmoinvaz:
This PR adds two cmake tests, one that runs makefixed to output its result, and the second one that verifies that the output of makefixed is equal to the output of inffixed.h.


Dead2:
Do we need make clean and git_ignore to know about inffixed._h perhaps?

nmoinvaz:
Done

PR #400, opened 2019-09-08
Modify cmake to use object libraries

nmoinvaz:
This PR uses two object libraries to compile the source code. These are then used for linking the dynamic and static libraries. The benefit is that source code compilation only happens once. Previously, the source code libraries were compiled twice - once for the static library and once for the dynamic library. The down side is it makes the cmake file a bit more complicated and adds two new projects to the solution. Looking for feedback on whether we should merge it in.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -757,19 +757,15 @@ target_include_directories(zlibobj PUBLIC
         ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
 
 if(NOT DEFINED BUILD_SHARED_LIBS)
-    add_library(zlib SHARED)
-    add_library(zlibstatic STATIC)
+    add_library(zlib SHARED $<TARGET_OBJECTS:zlibobj>)
+    add_library(zlibstatic STATIC $<TARGET_OBJECTS:zlibobj>)
 
     set(ZLIB_INSTALL_LIBRARIES zlib zlibstatic)
 else()
-    add_library(zlib)
+    add_library(zlib $<TARGET_OBJECTS:zlibobj>)
 
     set(ZLIB_INSTALL_LIBRARIES zlib)
 endif()
-
-foreach(ZLIB_INSTALL_LIBRARY ${ZLIB_INSTALL_LIBRARIES})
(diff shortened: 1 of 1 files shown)


mtl1979:
I would think position-independent code is slower because it requires indirect memory accesses… Is there benchmark done to make sure the speed difference is insignificant.

nmoinvaz:
I had errors on one of the configurations complaining that the code needed to be compiled with PIC.

nmoinvaz:
https://travis-ci.org/nmoinvaz/zlib-ng/jobs/582354915

```
/usr/bin/ld: CMakeFiles/zlibobj.dir/adler32.c.o: relocation R_X86_64_32 against `.data' can not be used when making a shared object; recompile with -fPIC
CMakeFiles/zlibobj.dir/adler32.c.o: error adding symbols: Bad value
```

Not quite sure what that is all about.

mtl1979:
The problem is that you can't use same object files for static and shared libraries because the symbol tables are incompatible.

Dead2:
PIC code is not slower due to indirect memory access, but it does incur a small cost during startup or loading of shared library, when glibc will have to go through the code while loading it and change all jump addresses to the actual ones (This is also where LD_PRELOAD gets handled).

This startup cost can be removed by using the prelink application on your system, but that has had problems from time to time. It is generally not needed any more, since the loading of applications/libraries is very fast on modern hardware and we are talking milliseconds on startup. But it is still popular, and can be useful if you have a bash-script running the same quick commands in a loop (sort, uniq, etc).

Security-wise, PIC (shared libs) and PIE (executables) are harder to locate in memory and thus harder to abuse to hack a system. Prelinking does reduce that security a bit, since all libs will keep the same random addresses on that system instead of getting new random addresses on load if it is not already loaded in memory.

My understanding of this:
Shared libs needs to be compiled with -fPIC.
Static libs needs to be compiled without -fPIC, then the application using it should be compiled with -fPIE.

nmoinvaz:
I don't believe there is a way to do PIC for one library and not the other without recompiling from the sources. I think perhaps I will close this issue then and we will leave it so that it compiles the sources twice since it is needed.

PR #401, opened 2019-09-11
Add more tests cases for deflate

nmoinvaz:
These tests cases are taken from the PR at madler/zlib#438. I had to make some modifications to the test cases for formatting, PREFIX, old zlib types, etc. 

- I think in a future PR I can probably address all the duplicate code issues for this example.c. 
- deflate.h include was necessary because test_deflate_copy checks the status of the state which is an internal structure.

### DIFF
--- a/test/example.c
+++ b/test/example.c
@@ -11,6 +11,7 @@
 #else
 #  include "zlib-ng.h"
 #endif
+#include "deflate.h"
 
 #include <stdio.h>
 
@@ -85,7 +86,7 @@ void test_gzio(const char *fname, unsigned char *uncompr, z_size_t uncomprLen)
     fprintf(stderr, "NO_GZCOMPRESS -- gz* functions cannot compress\n");
 #else
     int err;
-    int len = (int)strlen(hello)+1;
+    size_t len = strlen(hello)+1;
     gzFile file;
     z_off_t pos;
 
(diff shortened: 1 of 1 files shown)


Dead2:
Looks really good.

Regarding the duplicate code, I was thinking about the possibility of using macros for some of it, such as the four common lines to set up and init c_stream could be one. If they are generic enough, some of those could also become useful macros for real users of zlib/zlib-ng, something that is relevant since this is after all in example.c.
Once such macros have been made, we could also consider whether some of them (not the c_stream one) should become part of zlib-ng.h to make use of zlib-ng easier.

mtl1979:
I don't think adding macros to zlib-ng.h is a good idea. We should have separate file so it is easier to find them as zlib-ng.h is already quite long.

Dead2:
I agree, that is probably a good idea. IF we end up with any macros we want to make public at all.

mtl1979:
Macros are good for private code, but inline functions are better for public use. You can do the same with static inline functions when passing parameters as pointers or references. Excess macros can cause the compiler to run out of stack space (ICE) as most multi-line macros are written as scoped code blocks that essentially have own variable scope but still inherit variables from parent scope. With inline functions only necessary variables are copied over. This is the same reason why fill_window on ARM didn't use fill_window_c() for unoptimized code path.

Dead2:
You can't use functions to set up variables though, so that wouldn't work in all cases. I was thinking macros to simplify the stuff that needs to be done in their code, not ours. But if you want to suggest new convenience functions for our API, I am listening.

mtl1979:
If same group of variables are setup exactly the same way in multiple places, it makes sense to make them member of a struct and pass reference to the struct to the initialization function.

nmoinvaz:
I think those function calls will be mainly only be used in our test code. I think most people who are using zlib in a single function would just call deflate/inflate once with `Z_FINISH`. Even if they weren't they are probably doing something too specialized to use a macro anyways.

PR #403, opened 2019-09-15
Added missing private headers to cmake

nmoinvaz:
Having all the headers makes it easier to find text in source code when searching using an IDE that cmake supports.


nmoinvaz:
Merged in latest changes. I don't like to rebase because sometimes it will add my username to all commits it does even though it isn't my checkin.

I also added crc32_p.h.

CMakeLists.txt
```diff
@@ -655,15 +659,23 @@ set(ZLIB_PUBLIC_HDRS
...
     inflate.h
+    inflate_p.h
     inftrees.h
+    match_p.h
+    memcopy.h
     trees.h
+    zbuild.h
+    zendian.h
     zutil.h
 )
```
Dead2, line 681: Missing fallback_builtins.h and trees_p.h, possibly others, not sure.

nmoinvaz: fallback_builtin.h is already added via ZLIB_ARCH_HDRS only for MSVC.

nmoinvaz: Added missing trees_p.h

nmoinvaz:
Added trees_p.h and double checked to make sure I didn't miss anything. Not sure why codecov failing. Maybe needs to be retried.

Dead2:
Needs a rebase. Would probably be a good idea if you squashed the commits too.

nmoinvaz:
It was easier to just create a new PR, see #447.

Issue #394, opened 2019-08-28
[x86_64][msan] make test fails


sebpop:
On x86_64 `make test` started failing after https://github.com/zlib-ng/zlib-ng/pull/348 when configured with `./configure --with-msan`.  There was a first fix that removed several errors: https://github.com/zlib-ng/zlib-ng/pull/367 and there still seem to be some problem left:

```
Step #20: ./switchlevels 1 5 9 3 </src/zlib-ng/test/GH-364/test.bin >/dev/null
Step #20: ==1341==WARNING: MemorySanitizer: use-of-uninitialized-value
Step #20:     #0 0x4b50f8 in longest_match /src/zlib-ng/match_p.h:473:17
Step #20:     #1 0x4b6169 in deflate_slow /src/zlib-ng/deflate_slow.c:65:31
Step #20:     #2 0x49f478 in zng_deflate /src/zlib-ng/deflate.c:1036:18
Step #20:     #3 0x494da1 in compress_chunk /src/zlib-ng/test/switchlevels.c:67:11
Step #20:     #4 0x495d4b in main /src/zlib-ng/test/switchlevels.c:103:13
Step #20:     #5 0x7f88fb70082f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
Step #20:     #6 0x41b6c8 in _start (/src/zlib-ng/test/switchlevels+0x41b6c8)
Step #20: 
Step #20:   Uninitialized value was stored to memory at
Step #20:     #0 0x4ccd98 in crc_fold_copy /src/zlib-ng/arch/x86/crc_folding.c:352:5
Step #20:     #1 0x4a9ade in read_buf /src/zlib-ng/deflate.c:1203:9
Step #20:     #2 0x4c4749 in fill_window_sse /src/zlib-ng/arch/x86/fill_window_sse.c:106:13
Step #20:     #3 0x4b550a in deflate_slow /src/zlib-ng/deflate_slow.c:39:13
Step #20:     #4 0x49f478 in zng_deflate /src/zlib-ng/deflate.c:1036:18
Step #20:     #5 0x494da1 in compress_chunk /src/zlib-ng/test/switchlevels.c:67:11
Step #20:     #6 0x495d4b in main /src/zlib-ng/test/switchlevels.c:103:13
Step #20:     #7 0x7f88fb70082f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
Step #20: 
[...]

PR #404, opened 2019-09-15
Fixed use of uninitialized value warning in crc_folding.c 

nmoinvaz:
This fixes an issue I encountered with uninitialized variable in crc_folding.c that is also referenced in #394. It is possible for `xmm_crc_part` to not be initialized when `goto done` is called in `crc_fold_copy`.


Dead2:
This looks correct to me.

PR #406, opened 2019-09-17
Split out maketrees and makecrch into separate files and tools

Dead2:
This also adds the tests to compare the generated header files with the existing header files, slightly improving our test coverage.

This removes support for compiling with DYNAMIC_CRC_TABLE, meaning we always have the CRC32 tables compiled-in instead of generating them on-demand. This pretty much only matters for embedded users that need every byte they can save, and zlib-ng is already a poor choice for embedded.

Changes to code that was not otherwise behind #ifdef:
- gen_codes is now ZLIB_INTERNAL instead of static
- The tr_static_init function is no longer used by zlib-ng
- trees.h has gotten whitespace changes in the tables, improving readability
- crc32.h has UL appended to the values of all tables, not only some of them.


nmoinvaz:
I think at work we use `DYNAMIC_CRC_TABLE` and it results in 8K reduction of executable size last time I looked.

nmoinvaz:
This change makes things a lot cleaner I think.

mtl1979:
* UL is incorrect suffix for uint32_t constants, "unsigned long" type is not guaranteed to be 32-bits, it should use UINT32_C() macro instead so it works correctly on both 32-bit and 64-bit targets
* removing the whitespace in trees.h causes the patch to be too big for GitHub's diff viewer

nmoinvaz:
Cmake looks pretty straightforward. I wonder if we should have a crc32_gen.exe that we can use to test the crc32 generation against known values.

mtl1979:
@nmoinvaz I think it's very good to have tests to compare crc32 and adler32 values against known correct values. It makes double checking optimized code a lot easier as not all bugs are caught by just trying to compress and uncompress files using minigzip.

Dead2:
@mtl1979 Half of the file already uses UL for uint32_t constants.
The alternatives are:
- Rewrite the write table function to conditionally support UL on or off, to make an identical file as before.
- Remove UL from the entire file.
- Add UL to the entire file.

I chose the last of the options, and I think that is the safest way for this PR.
I agree we might want to change this to ((uint32_t)0xdeadbeef) style instead, but that is outside the scope of this PR, and is best done in a completely separate PR. Feel free to submit a PR based on this that changes that.

Dead2:
@nmoinvaz We should certainly add crc32 and adler32 tests.
Perhaps a separate tool that just takes two parameters: algorithm and filename, and prints out the checksum? Or should it be like example.c that does lots of smaller tests in one run?

mtl1979:
* Casting constants to uint32_t is also incorrect as hexadecimal constants are always unsigned.
* With test for crc32/adler32, we should have a function/executable that tests output of all possible 32-bit values and compares the output with another file. I don't think we need separate test for all possible 8-bit and 16-bit input values or test for verifying output after two consecutive 32-bit input values.

nmoinvaz:
@Dead2 - I think it might be sufficient to put it in the example.c. Here are some test vectors for both algorithms that might be useful:

https://github.com/froydnj/ironclad/blob/master/testing/test-vectors/adler32.testvec
https://github.com/froydnj/ironclad/blob/master/testing/test-vectors/crc32.testvec

Dead2:
@mtl1979 Unsigned int is usually 32-bit, but it can also be 16-bit, even in millions of systems designed and manufactured today. Using that would mean those systems would be unable to use zlib-ng.
Do you have a suggestion to solve this without breaking those?

If I understand the constant typing correct, the important part it using a type that is big enough to contain the value, and UL is AFAIK always big enough to do that in this case.

Dead2:
@nmoinvaz Ironclad is under a BSD-type license, so we probably cannot use those examples as-is.
But we can certainly take the same approach and make our own samples. Were you thinking a 2-dimention array with 3 columns (text, ader32sum, crc32sum), if so then that sounds like a great idea.
The same array of samples could also be used for compression tests, some of the current tests use strings a bit too short to actually hit a lot of the code. One of the tests could perhaps concat all the samples (repeated until it hits more than 64K long).

Regarding test output, I am thinking we might want to replace some of the verbose output with a dot per test if the test is successful, but output stuff if it failed. (print to a buffer instead of stdout, then discard and print dot if successful, print to stdout if failed). There is after all very little useful information unless something fails. Optionally, we could have a verbose flag, but I don't think that is important.

mtl1979:
@Dead2 I see only two options, either use UINT32_C() or no suffix at all... I don't know how many platforms we still support where "int" is 16-bits, but it's the same issue as with "long"... "long long" is non-standard extension that is as wide as "long" on some platforms.

Dead2:
I checked and see that zlib uses UL suffix for their tables, and we have at some point removed those.
The crc_comb tables (that zlib does not have) has UL in our code, so it looks like we forgot that we had removed UL from the other tables when we merged the new tables.

The C spec says the constants will automatically use the datatype of the smallest size able to contain the value, I am unsure what that means for 0x00000000 (does it count leading zeroes as part of the value?) but we do know it does work fine.

I will go with removing UL from crc_comb. That does not revert our change and fixes the inconsistency caused by one of our merges.

crc32_p.h
```diff
@@ -0,0 +1,19 @@
...
+static uint32_t gf2_matrix_times(const uint32_t *mat, uint32_t vec) {
+    uint32_t sum = 0;
+    while (vec) {
+        if (vec & 1)
+            sum ^= *mat;
+        vec >>= 1;
+        mat++;
+    }
+    return sum;
+}
```
mtl1979, line 16: Only inline functions should be defined in header files

Dead2:
only rebased

crc32_p.h
Dead2: I'd say static functions are also fine if there are legit reasons for it to be there, but I've made it inline now.

mtl1979: The problem with non-inline functions in header files is that gcc might inline it anyways but some compilers will make duplicate copies if more than one compilation unit includes the header.

Dead2: IIRC static in that context means the compiler should make separate copies per compilation unit, so therefore that is as expected. It is not generally good practice of course, but in certain cases it can be necessary to avoid making it a global function in a separate compilation unit/source file.
But inlining this one is fine with me. :)

mtl1979: Making simple functions explicitly inline lessen the difference between compiler behaviours. It's true that static functions are local to the compilation unit they are included in, but the issue here was not being local or not, the issue was that different compilers treat non-inline static functions in header files differently.

PR #408, opened 2019-09-19
Travis and cmake changes to improve test coverage.

Dead2:
travis.yml changes:
 - Select minimal vm image instead of generic ('c'), reducing size and boot time.
   - Windows needs (lowercase) 'c' image to not hang on booting
 - Update from Xenial to Bionic for linux tests.
   - ppc64 images don't have Bionic, set to Xenial
   - ppc64 is having misc failures with Xenial tools, so use GCC-9 and Clang-6
   - ppc64 cannot run MSAN
 - Add another test for more complete testing across architectures.
 - Enable msan and/or sanitizers for various tests.
 - Disable ASAN leak detection on aarch64, since it crashes under qemu
 - Reorder tests to be better grouped and running windows/macox tests lasts.
 - Enable verbose messages from sanitizers

CmakeLists.txt changes:
 - Enable warnings by default.
 - Add MAINTAINER setting to cmake
   - Enables extra warnings.
   - Enables fuzzers.
 - Add CI detection to cmake, currently auto-enables MAINTAINER mode.
 - Add detection of several more sanitizer features.
 - Test sanitizer features together, not just alone.
   This also helps with their inter-dependencies.

.gitignore changes:
 - Add switchlevels.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,7 @@
 /minigzip
 /minigzip64
 /minigzipsh
+/switchlevels
 /zlib.pc
 /zlib-ng.pc
 /CVE-2003-0107
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,14 @@
-language: c
-#cache: ccache  # cannot be used because it caches gcno(coverage) files
-dist: xenial
+# Notes:
+# - Fuzzers get automatically enabled when cmake detect CI
+# - ccache cannot be used because it caches gcno(coverage) files,
+#   and it is slow during travis network issues
+# - "language" selects what vm image is chosen
+#   - Linux, "C" selects the full generic image, we only need the minimal image
+#   - Windows builds require "c" image
+
+language: minimal
 bundler_args: --retry 5
+dist: bionic
 
 env:
   global:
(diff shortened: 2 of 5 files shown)


gdevenyi:
I was working on this, might make sense to integrate here:
```
if(MAINTAINER)
  if ("${CMAKE_C_COMPILER_ID}" STREQUAL "Clang")
     # using Clang
     list(APPEND CMAKE_C_FLAGS "-Weverything")
  elseif ("${CMAKE_C_COMPILER_ID}" STREQUAL "GNU")
    # using GCC
    list(APPEND CMAKE_C_FLAGS "-Wall -Wextra -Wpedantic")
  elseif ("${CMAKE_C_COMPILER_ID}" STREQUAL "Intel")
    # using Intel C++
     list(APPEND CMAKE_C_FLAGS "-w3")
  elseif ("${CMAKE_C_COMPILER_ID}" STREQUAL "MSVC")
     # using Visual Studio C++
		list(APPEND CMAKE_C_FLAGS "/Wall")
  endif()
endif()

```

Dead2:
I am seeing quite a few new warnings with this..
```
make clean; rm -rf CMakeFiles; rm CMakeCache.txt; CI="trqavis" cmake . && make
```
Mostly -Wstrict-overflow warnings. It looks like some (most?) of them are simply false positives, but I am uncertain about some of them. If we can fix them; great, if not, then I'll set -Wno-strict-overflow.

Dead2:
Enabled warnings also for non-gcc/clang compilers, enabled extra warnings for maintainer mode, and also unified the method of applying warning flags.
Disabled LeakSanitizer if CMAKE_CROSSCOMPILING_EMULATOR is set, since it crashes when run under qemu.

gdevenyi:
I think you haven't pushed those changes yet

Dead2:
@gdevenyi Search for WARNFLAGS, I merged it into the existing if-structure handling compiler flags.

gdevenyi:
Sorry, see it now.

What do you think of the MSVC and Clang all warning flags?

Dead2:
IIRC:
- MSVC /Wall is an alias of /W3, so W4 enables more warnings. W5 is likely too much though.
  Will need to verify that W4 is what we want, unfortunately travis windows builds have been broken for most of the day today.
- Clang -Weverything enables ALL warnings Clang has, and is intended for Clang developers, not for real-world use. There are plenty of warnings I have not enabled in GCC as well, because they warn about things that are false-positives or things that really are not problems but more like a kind of trivia.
  It might be fun to run once in a while, just to see whether you spot something that could be worth spending the time to improve, but it is probably not so good for setups where you want to turn warnings into errors. :)

Dead2:
Sorry, that was incorrect. /Wall is /W4.
ICC has an additional level W5, MSVC does not.

Dead2:
All Travis tests are finally successful.

Unfortunately, "coverage" went down, but that is really because of codecov-related problems, actual test coverage is up.
But because the codecov.io upload script is unable to use gcno files created anything other than the default gcc used by the VM image. So when we compile using clang or gcc-9 for example, we get zero coverage data from those builds.
That is actually the problem I was trying to fix when I stumbled into a bunch of other issues and started making this and the previous two big PRs improving our cmake and travis scripts.

So, if nobody objects to this PR, I will apply it soon, and meanwhile I will again try to work on the codecov problems and hopefully find a good solution.

nmoinvaz:
Patch looks good. CMakeLists.txt is improved in some areas.

PR #410, opened 2019-09-19
Fix leak in example.c test_deflate_copy()

Dead2:
In test_deflate_copy(), we free the original struct but forget to free the copy stuct.
Found by msan.


nmoinvaz:
Looks good.

Issue #416, opened 2019-09-19
Clang warnings in example.c and gzclose.c


Dead2:
```
Scanning dependencies of target example
[ 56%] Building C object CMakeFiles/example.dir/test/example.c.o
/home/travis/build/zlib-ng/zlib-ng/test/minigzip.c:295:2048: warning: array index 3 is past the end of the array (which contains 3 elements) [-Warray-bounds]
[...]

PR #422, opened 2019-09-20
Potential fix for clang array warnings.

nmoinvaz:
Not sure if this will fix the issue with the clang warnings but at least it is cleaner. Maybe the compiler will think smarter when using []. See #416.

### DIFF
--- a/test/minigzip.c
+++ b/test/minigzip.c
@@ -273,45 +273,43 @@ int main(int argc, char *argv[])
 {
     int copyout = 0;
     int uncompr = 0;
+    int i = 0;
     gzFile file;
     char *bname, outmode[20];
 
     snprintf(outmode, sizeof(outmode), "%s", "wb6 ");
 
-    prog = argv[0];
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Not sure why codecov fails.

Dead2:
Codecov will be unstable until I get my travis+cmake update merged.
Turns out cmake caches the gcno coverage files, and of course that messes up coverage completely since they are sometimes shared across builds/branches/commits. That might have been semi-fine (not really), except it is cached per source file, so it can be re-using coverage for deflate.c even though the cached version had good coverage because of extra tests and the new one has lousy coverage with no tests. And since we do 5 simultaneous builds, those cached files will be updated and changes overwritten without being "thread-safe".

So ccache needs to be disabled, actually it spends a lot of time downloading and uploading that cache anyways, so compiling our little source does not matter much.

Dead2:
Looks good to me.

PR #423, opened 2019-09-20
Fixed assert during inflate fast when len < sizeof(uint64_t).

nmoinvaz:
This PR is for an issue that I am having on Windows where an `assert` will be hit inside of `chunkmemset` in memcopy.h:

```
static inline unsigned char *chunkmemset(unsigned char *out, unsigned dist, unsigned len) {
    Assert(len >= sizeof(uint64_t), "chunkmemset should be called on larger chunks");
```

It happens due to the fact that it is passed a len that is less than sizeof(uint64_t). 

Here in the following code, the comment says it is okay because we expect to have enough room to make the copy:
```
/* Whole reference is in range of current output.  No
    range checks are necessary because we start with room
    for at least 258 bytes of output, so unroll and roundoff
    operations can write beyond `out+len` so long as they
    stay within 258 bytes of `out`.
*/

if (dist >= len || dist >= INFFAST_CHUNKSIZE)
    out = chunkcopy(out, out - dist, len);
else
    out = chunkmemset(out, dist, len);
```

However, the assert will still be hit even though we have enough space, we are still passing in a len that is less than sizeof(uint64_t).

What I have done is change the `chunkmemset` call to `chunkmemsetsafe`. 

```
if (dist >= len || dist >= INFFAST_CHUNKSIZE)
    out = chunkcopy(out, out - dist, len);
else
    out = chunkmemsetsafe(out, dist, len, len);
```

And in `chunkmemsetsafe` I changed the first if statement from this:

```
 if (left < (unsigned)(3 * INFFAST_CHUNKSIZE)) {
```

To this, where it checks to see if the len is less than sizeof(uint64_t) and if so does a slow copy to avoid the assert:

```
if (len < sizeof(uint64_t) || left < (unsigned)(3 * INFFAST_CHUNKSIZE)) {
```

### DIFF
--- a/inffast.c
+++ b/inffast.c
@@ -310,7 +310,7 @@ void ZLIB_INTERNAL zng_inflate_fast(PREFIX3(stream) *strm, unsigned long start) 
                     if (dist >= len || dist >= INFFAST_CHUNKSIZE)
                         out = chunkcopy(out, out - dist, len);
                     else
-                        out = chunkmemsetsafe(out, dist, len, len);
+                        out = chunkmemset(out, dist, len);
 #else
                     if (len < sizeof(uint64_t))
                       out = set_bytes(out, out - dist, dist, len);
--- a/inffast.c
+++ b/inffast.c
@@ -310,7 +310,7 @@ void ZLIB_INTERNAL zng_inflate_fast(PREFIX3(stream) *strm, unsigned long start) 
                     if (dist >= len || dist >= INFFAST_CHUNKSIZE)
                         out = chunkcopy(out, out - dist, len);
                     else
-                        out = chunkmemset(out, dist, len);
+                        out = chunkmemsetsafe(out, dist, len, len);
 #else
                     if (len < sizeof(uint64_t))
                       out = set_bytes(out, out - dist, dist, len);
--- a/memcopy.h
+++ b/memcopy.h
@@ -298,7 +298,7 @@ static inline unsigned char *chunkmemset(unsigned char *out, unsigned dist, unsi
 }
 
 static inline unsigned char* chunkmemsetsafe(unsigned char *out, unsigned dist, unsigned len, unsigned left) {
-    if (left < (unsigned)(3 * INFFAST_CHUNKSIZE)) {
+    if (len < sizeof(uint64_t) || left < (unsigned)(3 * INFFAST_CHUNKSIZE)) {
         while (len > 0) {
           *out = *(out - dist);
           out++;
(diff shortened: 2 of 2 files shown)


nmoinvaz:
Actually I think the thing to do might be to just remove the assert:
```
    Assert(len >= sizeof(uint64_t), "chunkmemset should be called on larger chunks");
```

The code below it is able to handle when len < sizeof(uint64_t):

```
    unsigned char *from = out - dist;
    inffast_chunk_t chunk;
    unsigned sz = sizeof(chunk);
    if (len < sz) {
        do {
            *out++ = *from++;
            --len;
        } while (len != 0);
        return out;
    }
```

Dead2:
The assert is probably there just for performance reasons.

nmoinvaz:
It gets hit every time when run in debug mode. Do you prefer the original patch or removing the assert?

Dead2:
Honestly, I am not sure what I prefer yet.
It depends on what kind of performance loss we are dealing with.

If I understand the code correctly, the assert is complaining because chunkmemset is slower than other solutions for short lengths, so we should be using something else, if possible.

So the assert is valuable in its own way, but perhaps it should have been some kind of debug message instead of an assert.

The solution you have currently is probably good too.

nmoinvaz:
Removing the assert will end up in the same code being called for shorter lengths as would the patch. chunkmemset:

```
    unsigned char *from = out - dist;
    inffast_chunk_t chunk;
    unsigned sz = sizeof(chunk);
    if (len < sz) {
        do {
            *out++ = *from++;
            --len;
        } while (len != 0);
        return out;
    }
```
Versus the patch:
```
    if (len < sizeof(uint64_t) || left < (unsigned)(3 * INFFAST_CHUNKSIZE)) {
        while (len > 0) {
          *out = *(out - dist);
          out++;
          --len;
        }
        return out;
    }
```

I think it is better to just remove the assert. In my patch in inffast.c, I changed `chunkmemset` to `chunkmemsetsafe`. But the comment above those calls says it doesn't need to be worried about being safe. That is probably why I don't like the patch as much. 

Another option would be to comment out the assert and leave a note stating this can be turned on when debugging performance related issues.

Dead2:
Well, did you check whether that location (the one that calls chunkmemset) always calls it with a short length? If it sometimes has a bigger length, then it would likely be better to just comment out the assert and add the comment.
For bigger lengths, calling chunkmemsetsafe would add quite a bit of extra code (several extra if checks and an extra function call), and also it is not supposed to be necessary to avoid overlaps where it is called from (At least I think that is why it is called safe).
I do not know the memcopy.h code as well as I'd like unfortunately.

nmoinvaz:
I checked and most of the time `chunkmemset` (from that location) is called with a length of 258. I have commented out the assert and noted it could be turned on for performance debugging. Like I said previously, the condition with `len < sizeof(uint64_t)` is handled properly by the function's `if (len < sz)` branch. I tried to use the comment style of the surrounding code.

PR #425, opened 2019-09-20
Split arch detection into separate cmake module

Dead2:
-  Add archdetect as separate cmake include file.
 - Simplify architecture checking by setting BASEARCH in addition to ARCH.
 - Clean up parts of CMakeLists.txt
 - Unify detection of ARM getauxval code availability.
    We don't want to compile arch-specific code when WITH_OPTIM is not set,
    and the current checks don't take that into account.
 -  Configure: Don't enable optimizations if without-optimizations has been set.
    Add ARM_GETAUXVAL define for ARM platforms if optimizations is enabled.

This also fixes a problem where gcc ppc64le builds sometimes attempt to build with -msse2 etc flags.


cmake/archdetect.c
```diff
@@ -0,0 +1,96 @@
...
+// PowerPC
+#elif defined(__powerpc__) || defined(_ppc__) || defined(__PPC__)
+    #if defined(__64BIT__) || defined(__powerpc64__) || defined(__ppc64__)
+        #error archfound ppc64
+    #else
+        #error archfound ppc
+    #endif
+
+// --------------- Less common architectures alphabetically below ---------------
+#elif defined() || defined()
```
mtl1979: This doesn't look right...

Dead2: heh, no :)

Dead2: Fixed

cmake/archdetect.cmake
```diff
@@ -0,0 +1,77 @@
...
+# Make sure we have ARCH set
+if(NOT ARCH OR ARCH STREQUAL "unknown")
+    set(ARCH ${CMAKE_SYSTEM_PROCESSOR})
+    message(STATUS "Arch not recognized, falling back to cmake arch: '${ARCH}'")
+else()
+    message(STATUS "Arch detected: '${ARCH}'")
+endif()
+
+
+# Base arch detecion
```
nmoinvaz: Spelling

cmake/archdetect.c
```diff
@@ -0,0 +1,95 @@
...
+
+// PARISC
+#elif defined(__hppa__)
+    #error archfound parisc
+
+// RS-6000
+#elif defined(__THW_RS6000)
+    #error archfound rs6000
+
+// return 'unrecognized' if we do nok know what architecture this is
```
nmoinvaz: Spelling

nmoinvaz: Spelling

cmake/archdetect.cmake
nmoinvaz: Spelling

cmake/archdetect.cmake
```diff
@@ -0,0 +1,77 @@
+# archdetect.c -- Detect compiler architecture and set ARCH and BASEARCH
```
nmoinvaz: Should be archdetect.cmake not archdetect.c

cmake/archdetect.c
Dead2: Fixed

cmake/archdetect.cmake
Dead2: Fixed

cmake/archdetect.cmake
Dead2: Fixed

nmoinvaz:
Interesting use of c to detect the architecture. 

- Double quotes such as those around `"${ARCH}"` are no longer necessary when doing `MATCHES`. Maybe we can clean that all up in another PR.
- Consider setting variables instead of setting BASEARCH to a string.

So instead of:
```
elseif("${ARCH}" MATCHES "alpha")
    set(BASEARCH "alpha")
```
You'd do this:
```
elseif("${ARCH}" MATCHES "alpha")
    set(BASEARCH_ALPHA_FOUND TRUE)
```

Then in the CMakeLists.txt you can simply do:

```
if(BASEARCH_ALPHA_FOUND)
```

Dead2:
@nmoinvaz Those are good points :)
Not having to do regex checking of strings all the time should also lead to a slightly faster cmake.

nmoinvaz:
Thanks. Yeah it is pretty standard nomenclature for cmake for scripts/includes/functions to have a `*_FOUND` variable that is set.

It might make some sense to put this into this PR (maybe somewhere at the top of the CMakeLists.txt):
```
message(STATUS "Using CMake version ${CMAKE_VERSION}")
```

nmoinvaz:
Overall this PR is much needed. Nice work.

Dead2:
I think I'll set both, I was thinking the BASEARCH variable could also be useful in conjunction with setting ARCHDIR and such.

Regarding the *_FOUND variable, just want to clarify, you want a ARCHDETECT_FOUND variable, right?

I think we can get the version printout in there :)

Thanks for the detailed feedback, that helps a lot.

nmoinvaz:
Good point about being used in ARCHDIR. Well, I was thinking one _FOUND variable for each architecture. So there would be BASEARCH_X86_FOUND, etc.. There could also be an additional ARCHDETECT_FOUND to determine if any architecture was even detected.

Dead2:
Oh, absolutely, one _FOUND for each arch, that I already added localyl.
I was wondering whether you wanted an additional one to detect that the script was found, I'll add one in any case.

Pushed the changes now.

CMakeLists.txt
```diff
@@ -429,90 +388,92 @@ if(MSVC)
...
-    #else
-        crc = __builtin_ia32_crc32qi(crc, c);
-    #endif
-        (void)crc;
-        return 0;
-    }"
-    HAVE_SSE42CRC_INTRIN
-)
-set(CMAKE_REQUIRED_FLAGS)
+if("${BASEARCH}" MATCHES "x86")
```
nmoinvaz: Should be BASEARCH_X86_FOUND?

CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ if(CMAKE_VERSION VERSION_LESS 3.12)
 else()
   cmake_policy(VERSION 3.5.1...3.13.2)
 endif()
+message(STATUS "Using CMake version ${CMAKE_VERSION}")
```
nmoinvaz, line 7: "CMake" capitalization inconsistent with toolchain message() below..

CMakeLists.txt
```diff
@@ -101,7 +88,7 @@ option(WITH_OPTIM "Build with optimisation" ON)
 option(WITH_NEW_STRATEGIES "Use new strategies" ON)
 option(WITH_NATIVE_INSTRUCTIONS
     "Instruct the compiler to use the full instruction set on this host (gcc/clang -march=native)" OFF)
-if("${ARCH}" MATCHES "arm" OR "${ARCH}" MATCHES "aarch64")
+if(BASEARCH_ARM_FOUND)
     option(WITH_ACLE "Build with ACLE CRC" ON)
     option(WITH_NEON "Build with NEON intrinsics" ON)
 elseif("${ARCH}" MATCHES "s390x")
```
nmoinvaz: Consider `BASEARCH_S390_FOUND AND "${ARCH}" MATCHES "s390x"`.. for clarity and a bit of speed.

CMakeLists.txt
```diff
@@ -142,29 +129,22 @@ if(${CMAKE_C_COMPILER} MATCHES "icc" OR ${CMAKE_C_COMPILER} MATCHES "icpc" OR ${
...
     # /Oi ?
-    if(NOT ${ARCH} MATCHES "AMD64")
-        set(SSE2FLAG "/arch:SSE2")
-    endif()
-    if("${ARCH}" MATCHES "arm")
+    if(BASEARCH_X86_FOUND)
+        if(NOT ${ARCH} MATCHES "x86_64")
+            set(SSE2FLAG "/arch:SSE2")
+        endif()
+    elseif("${ARCH}" MATCHES "arm")
```
nmoinvaz: Intended to be BASEARCH_ARM_FOUND? I'm not sure on this one.

nmoinvaz:
Things are looking good.

CMakeLists.txt
Dead2: mismerge, thanks

CMakeLists.txt
Dead2: What is correct though?

CMakeLists.txt
Dead2: S360, but yeah, probably a good idea for clarity

CMakeLists.txt
Dead2: Another mismerge, thanks for finding those :)

CMakeLists.txt
nmoinvaz: Hmm. Probably "CMake" since that is what they show on their website. So probably the toolchain message() needs to be fixed.

Dead2: Agreed.

CMakeLists.txt
nmoinvaz: This one still doesn't show yet.

Dead2: Doh, I tried adding it to the wrong commit, no wonder it was confused. It's in now.

Issue #426, opened 2019-09-20
Tests failing when crosscompiling


Dead2:
Fuzzer tests are not run through the crosscompiling emulator at all, and fails on arm and ppc.

All tests fail on ppc, do we need a toolchain file that defines the qemu emulator perhaps?

Example build showing failures:
https://travis-ci.org/zlib-ng/zlib-ng/builds/587616712

nmoinvaz:
Try using this instead on CMakeLists.txt @ ~883:

```
add_test(${FUZZER}_fuzzer $<TARGET_FILE:${FUZZER}_fuzzer> ${ALL_SRC_FILES})
```

Then cmake will have no problem finding the executables.

nmoinvaz:
These few should be updated as well:
```
add_test(NAME example COMMAND $<TARGET_FILE:example>)
add_test(NAME example64 COMMAND $<TARGET_FILE:example64>)
add_test(NAME CVE-2003-0107 COMMAND $<TARGET_FILE:CVE-2003-0107>)
```

Dead2:
But shouldn't it also be using ${CMAKE_CROSSCOMPILING_EMULATOR} somehow? Like the CVE ones do..

nmoinvaz:
Yes that's right. I'll take a look at it and see what works.

Dead2:
I added your suggested changes to the ongoing "travis" branch/PR for now.
If you add them to a PR yourself, I'll just rebase it on top of that.

nmoinvaz:
I think this can be closed too

PR #431, opened 2019-09-20
Fixed running cmake tests on arm

nmoinvaz:
See #426.


nmoinvaz:
Also notice the double #if define(define(..))

Dead2:
I just fixed the double defined myself :)
Could you remove (amend) that and push again please?

Dead2:
Oh, and do it against develop, then I'll rebase on top of that when this goes in first.

nmoinvaz:
See #432

PR #434, opened 2019-10-05
Add CI support for GitHub Actions

nmoinvaz:
This PR adds support for GitHub Actions as a replacement for Travis CI. 

Things this does, that Travis CI didn't do:

- Adds compilation and testing PPC, PPC64, SPARC64, and S390X via qemu.
- All build configurations complete in less than 10 minutes.
- Tests all ARM configurations for both cmake and configure.
- Codecov support for clang and gcc (set `CODECOV_TOKEN` secret) that use Ninja
- Out of source builds

I split the CI yaml into one for Configure and one for CMake to simplify them.

The GitHub CI can be viewed here:
https://github.com/nmoinvaz/zlib-ng/actions


nmoinvaz:
Done

nmoinvaz:
One thing I noticed in the Travis CI yaml it says at the top:
```
# Fuzzers get automatically enabled when cmake detect CI
```
But I don't see anywhere where the detection takes place in CMakeLists.txt.

Dead2:
Line 73:
```
if(DEFINED ENV{CI})
    message(STATUS "CI detected, enabling maintainer mode.")
    set(MAINTAINER ON)
endif()
if(MAINTAINER)
    set(WITH_FUZZERS ON)
endif()
```

nmoinvaz:
Thanks I think my branch is out of date.

nmoinvaz:
It looks like there was a bug in CMakeLists.txt, where the options parsing was overwriting the value of `MAINTAINER` and `WITH_FUZZER` options. It was checking for `CI` environment variable then setting those variables and then it was being overwritten by options parsing.

nmoinvaz:
Everything appears to be working properly. Only the CMake codecov run is failing during Run tests stage due to #437 and #438.

Dead2:
- CODECOV_TOKEN is not needed when using Travis with public repos, does codecov not support GitHub Actions similarily?
- Travis sets CI env variable, GitHub Actions does not?
- I'd like to see an empty line between configs in the matrixes, to improve readability, similar to the Travis config.
- We want codecov on every build if possible, because the different builds will hit different code, and we want to calculate coverage data from all of them.
- I'd prefer if we could replace the looong parameters with shorthands, like I introduced in Travis recently. For example for the much-repeated `-DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-arm.cmake -DCMAKE_C_COMPILER_TARGET=arm-linux-gnueabi` and similar. Those are hard to read (several of those are almost identical) and clutter the file by making the lines very long, pushing more interesting parameters out of view.

To be able to compare coverage data, I think this PR should remove .travis.yml, otherwise it becomes cumulative with the Travis data. And the idea is to replace Travis in any case, right?

Possibly not for this PR: Kcov can gather coverage data (without hit counters though) when running code that has debug info (gcov does not need debuginfo), this can be helpful if we cannot get the coverage data by normal means (msvc?). Kcov can also give coverage data for bash scripts like configure, I have not found any coverage tools with support for cmake though.

nmoinvaz:
- `CODECOV_TOKEN` is required. It might not be required for you because you have the Codecov app installed in GitHub. I tried removing the token environment variable and it would not upload the report to Codecov.
- No it does not set the CI environment variable, which is why I set it manually.
- I have added the whitespaces.
- I'll take a look at it
- I have not found a way to do that in GitHub Actions. Azure Pipelines YAML supports a `variable:` syntax but GitHub Actions does not. I don't think it is possible at this time. The cmake toolchain parmeters for ARM are duplicated only a few times. But on each of the platforms, the file it references is actually different. Even the packages that are needed on each platform is slightly different.
- I have removed travis.yml

nmoinvaz:
For `--fsanitizer=memory` we should also consider setting in _CMake_: 
```
-O1 -fno-omit-frame-pointer -fno-optimize-sibling-calls -fsanitize-memory-track-origins
```
Clang documentation says it is necessary for perfect stack trace:

> To get a reasonable performance add -O1 or higher. To get meaningful stack traces in error messages add -fno-omit-frame-pointer. To get perfect stack traces you may need to disable inlining (just use -O1) and tail call elimination (-fno-optimize-sibling-calls).

The flag ` -fsanitize-memory-track-origins` also tracks origins of use-of-uninitialized variables.

https://clang.llvm.org/docs/MemorySanitizer.html

nmoinvaz:
> Currently, codecov does not require a secret token to be explicitly set up and used in a public GitHub repository for several popular CI services (e.g., Travis CI). Unfortunately, Azure is not yet on that list. As a result, with Azure, pull request opened from a fork cannot access the token, and hence codecov cannot report back for the pull request.

https://developercommunity.visualstudio.com/idea/615016/integration-with-codecov-without-token-for-public.html

nmoinvaz:
@Dead2, I'm not sure what to use for the codecov flags: 
```
ubuntu_gcc_ppc64le_compat
```
Or:
```
ubuntu,gcc,ppc64le,compat
```

The documention isn't too clear on which way I should go: https://docs.codecov.io/docs/flags

nmoinvaz:
I did some tests and the two methods for codecov flags did not produce significant difference in results in the code base.

nmoinvaz:
I have added codecov to all CI tests (except MSVC) - however it does do Windows GCC. Here is the difference in code coverage graph (left side is my repo and right-side is zlib-ng/zlib-ng)

image

I had to move setting of the coverage flags into CMakeLists.txt to simplify the CI script. Because Codecov is running on all platforms it makes the script a bit more complex. I made some comments in the CI script with issues I encountered and why I did some things the way I did. Currently there is about 48 configurations between the two CI scripts.

Dead2:
I wish the toolchain files and cmake fixes had been split into a separate PR, but I'll pull this now as-is.
A follow-up PR might want to fix the aarch64 only being tested in zlib-compat mode though..

Dead2:
Seems like apt gets 404 errors on the ARM builds, I am not used to apt, but perhaps it needs to update the apt repo indexes before attempting the apt install? I am guessing the apt indexes present in the VM images are old and outdated and that is causing it to error out. (The best would have been to not include any indexes in the images, thus requiring an update instead, but again I am not used to apt.)

PR #436, opened 2019-10-06
Fixed msan detection when compiler doesn't support it

nmoinvaz:
Clang-4.0 returns "unrecognized command line option".
`CMAKE_REQUIRED_FLAGS` was not being reset after check.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -288,12 +288,14 @@ endif()
 
 if(WITH_MSAN)
     set(CMAKE_REQUIRED_FLAGS "-fsanitize=memory")
-    check_c_source_compiles("int main() { return 0; }" HAS_MSAN FAIL_REGEX "not supported")
-    if(${HAS_MSAN})
+    check_c_source_compiles("int main() { return 0; }"
+        HAS_MSAN FAIL_REGEX "not supported|unrecognized command")
+    if(HAS_MSAN)
         set(SANITIZERS_FLAGS "-fsanitize=memory")
         message(STATUS "Adding memory sanitizer flag: ${SANITIZERS_FLAGS}")
         set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${SANITIZERS_FLAGS}")
     endif()
+    set(CMAKE_REQUIRED_FLAGS)
 endif()
 
 if(WITH_SANITIZERS)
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Not sure what to do about this one:
```
set(WARNFLAGS_DISABLE "-Wno-implicit-fallthrough")
```

It is not supported in clang-4.0 either. Clang-4.0 is what we are using on Ubuntu 14.04.

mtl1979:
If something is not supported by compiler, it should not be used, we already have tests to check which command line options are available.

nmoinvaz:
The check doesn't work if it is not supported.

nmoinvaz:
I could remove `-Wno-implicit-fallthrough`, thoughts @Dead2?

Dead2:
It should have a compiler check, similar to the other compiler parameter checks, and only be added to the list if the check does not fail.

The point is we want to enable -Werror for CI-builds at some point, but before that we want to weed out irrelevant warnings, and that one is pretty useless. It can be fixed by sprinkling /* fall-through */ hints here and there, but I don't know whether that is even an universally accepted compiler hint. I do know that GCC supports a whole bunch of different formatting of that hint.

nmoinvaz:
I'm unable to reproduce the `-Wno-implicit-fallthrough` error using older versions of clang. I'm not sure where that was coming from. Maybe resetting CMakeCache.txt did the trick.

nmoinvaz:
It looks like it is not an error, but just a warning I found in CMakeOutput.log when using Clang-4.0:
```
cc1: warning: unrecognized command line option "-Wno-implicit-fallthrough" [enabled by default]
```
Here is more information, it looks like the compilers just show a warning and continue with complication.
https://stackoverflow.com/questions/49593240/using-wno-implicit-fallthrough-on-older-versions-of-gcc
I think we are okay to continue doing it how we do.

nmoinvaz:
I think in a few places in CMake we should be using `check_c_compiler_flag` instead of `check_c_source_compiles`. Such as the check for `-fsanitize` and other sanitizer flags. But that would be for another PR. This PR is specifically just to fix MSAN not resetting CMAKE_REQUIRED_FLAGS and the check not checking for the right verbage for clang.

Dead2:
Uh, did I break github? I did a rebase and merge, and now this PR seems to be in a tristate between open and closed. Is says it is closed, but it appears in the Open PRs list..
The PR seems to have been merged to develop though, so I guess it is mostly cosmetic.

nmoinvaz:
Yes very strange. It shows merged icon and it seems to be merged into `develop` but doesn't say in this thread. I'll close this PR from my end.

Issue #440, opened 2019-10-07
Conversion warnings in deflate on Windows


nmoinvaz:
```
2>zlib-ng\deflate.c(1628,13): warning C4244:  '=': conversion from 'uint16_t' to 'unsigned char', possible loss of data
2>zlib-ng\deflate_fast.c(61,13): warning C4244:  '=': conversion from 'uint16_t' to 'unsigned char', possible loss of data
2>zlib-ng\deflate_slow.c(89,13): warning C4244:  '=': conversion from 'uint16_t' to 'unsigned char', possible loss of data
```

PR #441, opened 2019-10-07
Fixed conversion warnings in zng_tr_tally_dist on Windows

nmoinvaz:
See #440. This uses the same type as `zng_tr_tally` in trees.c And also the same type `zng_tr_tally_dist` when `ZLIB_DEBUG` is defined in deflate.h.

### DIFF
--- a/deflate.h
+++ b/deflate.h
@@ -365,7 +365,7 @@ void ZLIB_INTERNAL flush_pending(PREFIX3(streamp) strm);
   }
 # define zng_tr_tally_dist(s, distance, length, flush) \
   { unsigned char len = (unsigned char)(length); \
-    uint16_t dist = (uint16_t)(distance); \
+    unsigned dist = (unsigned)(distance); \
     s->sym_buf[s->sym_next++] = dist; \
     s->sym_buf[s->sym_next++] = dist >> 8; \
     s->sym_buf[s->sym_next++] = len; \


mtl1979:
Signedness is same after the change. This is actually case of missing explicit cast in the following lines. When shifting, explicit cast is required by most compilers as shift operator is assuming 32-bit signed integers. Truncating also requires explicit cast.

nmoinvaz:
Sorry I meant "conversion" but typed "signedness". The PR does fix the warnings in #440 even though I botched the terminology.  I've updated the issue titles to reflect.

mtl1979:
There is a reason why we try to avoid using types like "unsigned", "int" and "long" in zlib-ng. Those types don't specify bit width of the variable and such might be slow on one architecture and fast on another. Like I already mentioned earlier, they can also mask other errors in the code due to being pre-standard types, especially "unsigned".

nmoinvaz:
I agree it is not ideal to not specify fixed width type, but I am just using the same type as `zng_tr_tally`. Maybe another PR at a later day can go and clean all the non-fixed width types up.

Dead2:
Isn't unsigned char the correct type here? Considering the definition of sym_buf:
```
 unsigned char *sym_buf;
```
Am I missing something?

nmoinvaz:
dist can't be unsigned char because the next line does dist >> 8. I suppose the first assignment automatically does the cast to unsigned char.

This PR just makes that bit of code the same as this bit of code (as far as casting/types goes):
https://github.com/zlib-ng/zlib-ng/blob/edbfb283977a19d1d6c80d78b9d761dd545c8071/trees.c#L736

PR #444, opened 2019-10-07
Compiler hints and prefetch

Dead2:
- Clean up LIKELY/UNLIKELY definitions, making them upper-case to improve visibility.
- Add LIKELY_NULL hint.
- Add PREFETCH_L1, PREFETCH_L2 and PREFETCH_RW for GCC, Clang, ICC and MSVC.

The new hints are not yet being used, this PR is just about making them available for future PRs. 
This PR should make no changes in the generated binaries.

### DIFF
--- a/zutil.h
+++ b/zutil.h
@@ -215,19 +215,27 @@ void ZLIB_INTERNAL   zng_cfree(void *opaque, void *ptr);
 
 /* Only enable likely/unlikely if the compiler is known to support it */
 #if (defined(__GNUC__) && (__GNUC__ >= 3)) || defined(__INTEL_COMPILER) || defined(__Clang__)
-#  ifndef likely
-#    define likely(x)      __builtin_expect(!!(x), 1)
-#  endif
-#  ifndef unlikely
-#    define unlikely(x)    __builtin_expect(!!(x), 0)
-#  endif
+#    define LIKELY_NULL(x)      __builtin_expect((x) != 0, 0)
+#    define LIKELY(x)           __builtin_expect(!!(x), 1)
+#    define UNLIKELY(x)         __builtin_expect(!!(x), 0)
+#    define PREFETCH_L1(addr)    __builtin_prefetch(addr, 0, 3)
+#    define PREFETCH_L2(addr)    __builtin_prefetch(addr, 0, 2)
+#    define PREFETCH_RW(addr)    __builtin_prefetch(addr, 1, 2)
+#elif defined(__WIN__)
+#    include <xmmintrin.h>
+#    define LIKELY_NULL(x) x
+#    define LIKELY(x)      x
+#    define UNLIKELY(x)    x
+#    define PREFETCH_L1(addr)    _mm_prefetch((char *) addr, _MM_HINT_T0)
+#    define PREFETCH_L2(addr)    _mm_prefetch((char *) addr, _MM_HINT_T1)
+#    define PREFETCH_RW(addr)    _mm_prefetch((char *) addr, _MM_HINT_T1)
 #else
-#  ifndef likely
-#    define likely(x)      x
-#  endif
-#  ifndef unlikely
-#    define unlikely(x)    x
-#  endif
+#    define LIKELY_NULL(x) x
+#    define LIKELY(x)      x
+#    define UNLIKELY(x)    x
+#    define PREFETCH_L1(addr)    addr
+#    define PREFETCH_L2(addr)    addr
+#    define PREFETCH_RW(addr)    addr
 #endif /* (un)likely */
 
 #if defined(_MSC_VER)
--- a/arch/arm/fill_window_arm.c
+++ b/arch/arm/fill_window_arm.c
@@ -112,7 +112,7 @@ void fill_window_arm(deflate_state *s) {
 
             s->ins_h = s->window[str];
 
-            if (unlikely(s->lookahead < MIN_MATCH))
+            if (UNLIKELY(s->lookahead < MIN_MATCH))
                 insert_cnt += s->lookahead - MIN_MATCH;
             slen = insert_cnt;
             if (str >= (MIN_MATCH - 2))
--- a/arch/arm/insert_string_acle.c
+++ b/arch/arm/insert_string_acle.c
@@ -21,7 +21,7 @@
 Pos insert_string_acle(deflate_state *const s, const Pos str, unsigned int count) {
     Pos p, lp, ret;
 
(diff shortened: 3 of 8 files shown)


zutil.h
```diff
@@ -215,19 +215,27 @@ void ZLIB_INTERNAL   zng_cfree(void *opaque, void *ptr);
...
-#  ifndef likely
-#    define likely(x)      __builtin_expect(!!(x), 1)
-#  endif
-#  ifndef unlikely
-#    define unlikely(x)    __builtin_expect(!!(x), 0)
-#  endif
+#    define LIKELY_NULL(x)      __builtin_expect((x) != 0, 0)
+#    define LIKELY(x)           __builtin_expect(!!(x), 1)
+#    define UNLIKELY(x)         __builtin_expect(!!(x), 0)
+#    define PREFECH_L1(addr)    __builtin_prefetch(addr, 0, 3)
```
mtl1979: Missing "T" in PREFETCH...

zutil.h
```diff
@@ -215,19 +215,27 @@ void ZLIB_INTERNAL   zng_cfree(void *opaque, void *ptr);
...
+#    define PREFECH_L2(addr)    __builtin_prefetch(addr, 0, 2)
+#    define PREFECH_RW(addr)    __builtin_prefetch(addr, 1, 3)
+#elif defined(__WIN__)
+#    include <xmmintrin.h>
+#    define LIKELY_NULL(x) x
+#    define LIKELY(x)      x
+#    define UNLIKELY(x)    x
+#    define PREFECH_L1(addr)    _mm_prefetch((char *) addr, _MM_HINT_T0)
+#    define PREFECH_L2(addr)    _mm_prefetch((char *) addr, _MM_HINT_T1)
+#    define PREFECH_RW(addr)    _mm_prefetch((char *) addr, _MM_HINT_T0)
```
mtl1979: This doesn't seem correct... Closest equivalent is _MM_HINT_ET1, because _MM_HINT_ET0 is unsupported. Difference between ET? and T? hints is the write intent.

mtl1979 (changes requested):
Fix spelling errors.

zutil.h
Dead2: Thanks, fixed.

zutil.h
Dead2: From what I found when I made this, the _ET* hints are not guaranteed to be defined by some versions of MSVC. I have been unable to find out what versions that is though, so I followed the method used by others and used _T0 for the RW-prefetch. It is not as good for multi-threading (cache invalidation across cores/cpus when writing takes some cycles), but that is not likely to be a big issue for us since we never multi-thread.

We might add _ET1 later when/if we figure out what MSVC versions can support it. It might also turn out to not be faster, since it only prefetches to L2 instead of L1. The sources I found were unclear on this aspect.

_ET1 could possibly be better for a PREFETCH_W(addr) or INVALIDATE(addr) where we don't want to read its contents after prefetching, but that is only a theory and would have to be benchmarked.

Basically, the normal prefetch does just that, it fetches a copy of the data. But the RW prefetch will make that cache copy exclusive and invalidate it in all other cores in the system. In general it is a shame that there isn't better native support for prefetch hinting and branch hinting in the x86 architecture.

mtl1979: Considering I have VS2015, VS2017.9 and VS2019.3 installed, I can easily check if one of them does not have ET hints defined. I've noticed some defines have been added or removed on updates, but the ones I have installed are the latest available.

Going through VS2015-2019 and Intel Intrinsics Guide, it seems write hints are not supported on any x86 processors except some Xeon Phi processors starting from Knights Corner core. I don't know why GCC allows them without triggering build-time error, but it seems best for me, to just not use them on any compiler we support. It might be that gcc is silently ignoring the hint which would make the macro completely useless.

Dead2: As long as GCC accepts it, it is fine, then it can decide whether to use it or not for the targeted cpu. GCC is also used for non-x86 cpus, and future x86 cpus, so it is still relevant.

These are only hints, and cpus are allowed to ignore those completely at their own discretion, so we cannot depend on them for correctness in any case.

mtl1979: I'm just worried adding them might backfire... I've seen prefetch used in really old code and then being commented out, because the effect was something else than expected. For me, the instructions and intrinsics sound more relic than future feature. Write hints only make sense when used with write-back policy, not write-through. As write-back policy can cause cpu stalls, I hardly can think it can cause speed improvement.

Dead2: I can assure you that prefetching is very viable even on skylake cpus, the automatic prefetchers are not foolproof even though they have improved vastly. The current prefetchers often have to learn (they even call it deep learning) how your software behaves, and after a few iterations on hot code it will most often be able to prefetch with good effect. Lack of available data in the L1 cache is after all one of the biggest causes of pipeline stalls, and results in under-utilization of the cores. And with the ever increasing amount of ticks the cores have to wait to fetch stuff from ram, this is one of the areas cpu designers focus on the most.

The problem is that not all code is hot enough to stay in cache long enough to be learned, or it keeps going in and out of that cache and has to be re-learned. And you also have code that is not repetitive enough to trigger the automated prefetching.

A lot of software does suffer from prefetching too late (after all it has to prefetched ~100-200 cycles in advance to be able to actually arrive in the cache before it is needed), prefetching after a branch (that might not be caught early enough by the prefetchers), or even prefetching the obvious stuff that the modern-ish cpus will all be able to handle just fine by itself (unless prefetching is disabled in the MSRs by the bios).

But what you are arguing about is premature, since this only adds the prefetch macros, it does not actually get used at all as of yet. Feel free to argue the use of these macros whenever a PR comes along that suggests their use. At that time it makes sense to do a thorough review and benchmarks on several cpus to see whether it is actually a win and does not have regressions elsewhere.

PS: As a reference, in some complex software we use at work, I recently removed a pipeline-stall in hot code by using prefetching, and it improved server scalability by A LOT (double digits). The speed of doing the operation in single-task mode did not improve more than about 0.5%, but when running hundreds of the same in parallel, it ensured the cores does not waste time by waiting for ram by stalling and effectively blocking the other tasks from doing useful work.

mtl1979: I'm not arguing prefetching does not benefit, I'm just arguing write hint is not supported in most of the 32-bit or 64-bit Intel/AMD processors, which means it is totally pointless to add macro for it even if gcc does not trigger error when using the intrinsic. We should only add the macro for architectures that actually have the relevant instruction and likely benefit from using it. For other architectures, it should be no-op, not random implementation-specific effect.

PR #448, opened 2019-10-10
Use check_c_compiler_flag for compiler flag check in CMake

nmoinvaz:


nmoinvaz:
It looks like there is a [bug](https://stackoverflow.com/questions/50024731/ld-unrecognized-option-push-state-no-as-needed) in the linker in GCC when using -fsanitize flags. So we will have to keep it the way it is, since it checks both the compiler and linker to ensure that the flag is supported.

Issue #456, opened 2019-10-15
Unresolved external symbols in maketrees


nmoinvaz:
```
maketrees.c.obj : error LNK2019: unresolved external symbol _z_error referenced in function _tr_static_init
trees.c.obj : error LNK2001: unresolved external symbol _z_error
trees.c.obj : error LNK2019: unresolved external symbol _send_bits referenced in function _zng_tr_flush_block
trees.c.obj : error LNK2001: unresolved external symbol _z_verbose
```
This happens in debug mode: 
```
cmake . --DBUILD_SHARED_LIBS=ON -DCMAKE_BUILD_TYPE=Debug
cmake --build . --config Debug
```
It appears trees.c requires some functions which are in zlib library and only work with static library. 

If I change CMakeLists.txt and add zutil.c:
```
add_executable(maketrees tools/maketrees.c trees.c zutil.c)
```
And then move `send_bits()` function from deflate.c to trees.c it works, but I'm not sure moving `send_bits()` is something we want to do.

mtl1979:
I don't see send_bits() used anywhere else than trees.c as deflate_quick.c has its own version, so moving it to trees.c isn't really a bad idea at all...

PR #460, opened 2019-10-20
Fixed unresolved external symbols in maketrees

nmoinvaz:
See #456. I have moved `send_bits()` from deflate.c to trees.c and added zutil.c to maketrees project.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -895,7 +895,7 @@ if (ZLIB_ENABLE_TESTS)
     add_executable(makefixed tools/makefixed.c inftrees.c)
     target_include_directories(makefixed PUBLIC ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
 
-    add_executable(maketrees tools/maketrees.c trees.c)
+    add_executable(maketrees tools/maketrees.c trees.c zutil.c)
     target_include_directories(maketrees PUBLIC ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
 
     add_executable(makecrct tools/makecrct.c)


mtl1979:
Looks good to me...

nmoinvaz:
Looks like Travis CI instance timed out.

Dead2:
I think this is no longer necessary..

nmoinvaz:
@Dead2 it is still necessary. In the issue #456 related to this PR I mentioned it happens in Debug mode. `-DCMAKE_BUILD_TYPE=Debug`. That is perhaps why you don't see it in the CI where it uses Release mode.

Dead2:
You don't think bcb7a0b588851257e5a531df4203c001c39d3472 solved this issue?
I don't have time to check for myself right now, but I don't see how it could fail any more.

Dead2:
Also, good point about debug mode for the CI, I have had that in the back of my head for a while, and we should certainly get it in shape and into CI, at least as a single CI build just to make sure the core code builds and runs fine with it enabled.

Possibly we could also consider enabling asserts on all (?) CI builds (not the full debug tracing etc though), since those _should_ only trigger on bugs/unexpected situations, and should hopefully only affect speed, so it could be nice to have those for correctness-checking in the CI.

nmoinvaz:
With that check-in send_bits() still remains in deflate.c. Yes it is still an issue I verified it again today on my last comment with latest develop branch.

nmoinvaz:
You can see it failing in this build log:
https://pipelines.actions.githubusercontent.com/tWPv9jfVvkKXJmMNWRDhKqxjGhmdO7GMOQGdOYs8D35Wvrmi6A/_apis/pipelines/1/runs/357/signedlogcontent/96?urlExpires=2019-10-23T01%3A32%3A33.8909003Z&urlSigningMethod=HMACV1&urlSignature=V9VLitz8C1GOwsi0Z%2FVI11p%2BExcbW0qtSaVHX0h85BU%3D

nmoinvaz:
You can see it failing [here](https://github.com/nmoinvaz/zlib-ng/runs/270867899) in the debug configuration.

Dead2:
@nmoinvaz No, send_bits() does not remain in deflate.c with that commit, the debug-function is removed completely and only the macro in deflate.h is used instead.

The failed build you linked does not mention send_bits, but rather z_error and z_verbose.
So I think the only change needed here is probably the one you made in CMakeLists.txt, no need to re-introduce the send_bits() function however. (Although that macro could possibly be moved, idk)

nmoinvaz:
It looks like you are right. We only need the zutil.c. Perhaps the changes made in https://github.com/zlib-ng/zlib-ng/commit/bcb7a0b588851257e5a531df4203c001c39d3472 made it so we don't need to move the function any longer.

Issue #452, opened 2019-10-13
AddressSanitizer: heap-buffer-overflow in crc_folding.c:333


nmoinvaz:
```
==1907==AddressSanitizer Init done
/Users/runner/runners/2.159.0/work/zlib-ng/zlib-ng/arch/x86/crc_folding.c:252:20: runtime error: unsigned integer overflow: 0 - 4464207872 cannot be represented in type 'unsigned long'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/runner/runners/2.159.0/work/zlib-ng/zlib-ng/arch/x86/crc_folding.c:252:20 in 
=================================================================
==1907==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x00010a1ccda0 at pc 0x000106752bbe bp 0x7ffee95d5b90 sp 0x7ffee95d5b88
READ of size 16 at 0x00010a1ccda0 thread T0
    #0 0x106752bbd in crc_fold_copy crc_folding.c:333
    #1 0x10663ecd6 in copy_with_crc crc32.c:212
    #2 0x10668d0fe in read_buf deflate.c:1202
    #3 0x10674a232 in fill_window_sse fill_window_sse.c:78
    #4 0x1066a2c10 in deflate_medium deflate_medium.c:217
    #5 0x106663326 in zng_deflate deflate.c:1035
    #6 0x10677c1bc in gz_comp gzwrite.c:115
    #7 0x1067734e1 in gz_write gzwrite.c:214
    #8 0x10677198f in zng_gzwrite gzwrite.c:245
    #9 0x10662cd25 in test_gzio_deflate_inflate example.c:286
    #10 0x106637f82 in main example.c:1495
    #11 0x7fff5c1783d4 in start (libdyld.dylib:x86_64+0x163d4)

0x00010a1ccda3 is located 0 bytes to the right of 419235-byte region [0x00010a166800,0x00010a1ccda3)
2019-10-11T22:36:48.2287670Z allocated by thread T0 here:
2019-10-11T22:36:48.2287830Z     #0 0x1068cf143 in wrap_malloc (libclang_rt.asan_osx_dynamic.dylib:x86_64+0x59143)
2019-10-11T22:36:48.2287990Z     #1 0x10662c93f in test_gzio_deflate_inflate example.c:245
2019-10-11T22:36:48.2288130Z     #2 0x106637f82 in main example.c:1495
[...]

PR #462, opened 2019-10-22
Fixed crc32 calculation during gzwrite when buf size is not 128 bit aligned

nmoinvaz:
See #452 for details. What is happening is that the buf allocation in the call to `gzwrite` is not 128-bit aligned. When copying the remainder of the unaligned bytes for the crc32 calculation, it is not taking into account the fact that the buffer might not be 128-bit aligned and is using `_mm_load_si128` function. Instead this PR uses `memcpy` to copy the remaining bytes into `xmm_crc_part` so it does not try and read past the buffer size.

### DIFF
--- a/arch/x86/crc_folding.c
+++ b/arch/x86/crc_folding.c
@@ -249,7 +249,7 @@ ZLIB_INTERNAL void crc_fold_copy(deflate_state *const s, unsigned char *dst, con
         goto partial;
     }
 
-    algn_diff = (0 - (uintptr_t)src) & 0xF;
+    algn_diff = ((uintptr_t)0 - (uintptr_t)src) & 0xF;
     if (algn_diff) {
         xmm_crc_part = _mm_loadu_si128((__m128i *)src);
         _mm_storeu_si128((__m128i *)dst, xmm_crc_part);
@@ -311,7 +311,7 @@ ZLIB_INTERNAL void crc_fold_copy(deflate_state *const s, unsigned char *dst, con
             goto done;
 
         dst += 48;
-        xmm_crc_part = _mm_load_si128((__m128i *)src + 3);
+        memcpy(&xmm_crc_part, (__m128i *)src + 3, len);
     } else if (len + 32 >= 0) {
         len += 32;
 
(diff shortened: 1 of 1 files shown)


arch/x86/crc_folding.c
```diff
@@ -16,7 +16,7 @@
  * For conditions of distribution and use, see copyright notice in zlib.h
  */
 
-#ifdef X86_PCLMULQDQ_CRC
+#if X86_PCLMULQDQ_CRC
```
mtl1979: This is bad... Assumes X86_PCLMULQDQ_CRC is always defined

nmoinvaz: I noticed that as soon as I made the PR. Thanks. It is resolved now.

Dead2:
I am confused, is this about the data not being aligned or about reading beyond the buffer?
And was using _mm_loadu_si128() or even better (especially on older SSE3-capable cpus) _mm_lddqu_si128() considered?
I'd like to see benchmarks of this one.

mtl1979:
@Dead2 The main problem is that one can't do partial read or write using vector registers... so if len is less than vector size, the result is unoptimal and possibly invalid code.

Dead2:
@mtl1979 But at line 243 we already check for len < 16 and that code uses memcpy, so we already know that len is equal or bigger than the vector size.

mtl1979:
@Dead2 When I did read all the text, it said len = 3...

Dead2:
@mtl1979 I don't see any places saying len = 3 in the code, the PR comments or the issue comments. Where is this from?

mtl1979:
@Dead2 #452 mentioned on first post...

Dead2:
Hmm, so if that is correct, then the input len to the function should be 19.
First we have 19-64 = -45
That should then hit the ` } else if (len + 48 >= 0) {` code chunk and that should result in len becoming 3.
If I understand correctly that means that block handles the first 16 bytes, and the remaining 3 bytes go to be handled by the `:partial` code. And I guess it is copying those remainder 3 bytes into xmm_crc_part that is triggering the warning.

Doing an aligned load is not a problem, since src is 16-byte aligned (after we handle misalignment at line 253). So that suggests that over-reading the user-supplied buffer is the problem.

So, yes, that is no good and probably needs to be solved by using the memcpy.
@nmoinvaz If you agree, then please change the commit message, since it is misleading.
After all, we cannot assume the supplied buffer is a multiple of 16-bytes long, no matter what alignment it starts at. So even a 16-byte aligned buffer with a non-multiple-of-16 len should trigger this warning.

That function should really have some comments to make it easier to follow what it is doing.
(I also think line 247 loadu should be an aligned load, since partial_buf is supposedly aligned.)

PS: Please include some more details in commit/pr messages and comments, so I don't have to re-do the entire investigation that led to the changes when reviewing them, as that can sometimes take a lot of time.

nmoinvaz:
I have updated the commit message.

PR #473, opened 2019-10-26
Added regression test for deflateEnd returning -3 when using deflate quick.

nmoinvaz:
See #382. This commit adds a new test application called `deflateparams` that will allow us to create fine tuned test cases for `deflate`. 

In CMakeLists.txt I have added the regression test for #382, which was initially done in #388. I have decided to move the functionality into its own test application because I believe we can get more code coverage this way testing various deflateInit param configurations.


test/deflateparams.c
```diff
@@ -0,0 +1,254 @@
...
+{
+    PREFIX3(stream) c_stream; /* compression stream */
+    unsigned char compr[4096];
+    int err;
+
+    c_stream.zalloc = zalloc;
+    c_stream.zfree = zfree;
+    c_stream.opaque = (void *)0;
+    c_stream.total_in = 0;
+    c_stream.total_out = 0;
```
mtl1979: This is something that would benefit from using a macro to eliminate/shorten very long span of repeated lines... maybe INIT_STREAM(...)... 

We will always use zalloc/zfree functions...

test/deflateparams.c
```diff
@@ -0,0 +1,254 @@
...
+    }
+    else if (large_buf) {
+        deflate_params_large(buf, len, level, window_bits, mem_level, strategy);
+    }
+    else {
+        deflate_params(buf, len, level, window_bits, mem_level, strategy);
+    }
+
+    free(buf);
+    fclose(f);
```
mtl1979: Maybe these lines should be broken as separate function with file name as first parameter.

nmoinvaz: I was going toward something more like what minigzip has - in that it can work with file in/out or std in/out. But right now for the deflateEnd -3 test all it needs is file in/out.

test/deflateparams.c
nmoinvaz: I think we talked about doing that at some point in another thread. But I'm not so sure about how useful it would be.

nmoinvaz:
Does anybody have any thoughts about the name of the test project or the name of the c source file? Right now I just have it as `deflateparams`. I plan on doing a similar test app for inflate() only. I was thinking maybe to just call it `deflate` and `inflate`. So the simple hello world test could just go:

```
echo hello world | deflate | inflate | findstr "hello world"
```

Or something similar.

test/deflateparams.c
mtl1979: This is an issue that rises only with C code, as C++ has constructors that initialize the members with default values... I would rather not disable some warnings in external services just because the code has a lot of duplication.

test/deflateparams.c
mtl1979: I've seen applications where "-" as filename means standard input... 

Long methods are another issue specific with C as all variables need to be declared at the top and compiler needs to know how many variables can stay in registers. For that it needs to count how far from top any variable is modified or read, as those that are only used close to top can be safely spilled to stack after they are no longer needed and the register can be reused for another variable. More there is calls to other functions, more pressure there is for storing variables in registers instead of stack (memory).

mtl1979:
@nmoinvaz Name of the test application should not be confusing or ambiguous as if there will be more deflate-related tests, none of them can have same name as another...

Dead2:
* test/data/defneg3.dat needs copyright info.
* Helptext for small/large should perhaps not be in the middle of the regular deflate parameters, perhaps move those to the end?
* Header of deflateparams.c should contain a copyright/author header, and please include a section describing what this tool is and does.
* Does this output deflate streams with headers, or just raw deflate streams, or gzip files? If with deflate headers, then I'd like to suggest 'minideflate' as the name.
* I'd like to see inflate capability in the same binary with the -d parameter

* If this becomes capable of reading/writing files, and if it is dealing in deflate files, then the .zz file extension is probably the correct one.

Other than that, this looks great.

nmoinvaz:
- test/data/defneg3.dat is something I generated.
- I'm not sure what you mean about help text. Can you add a comment on the line you are referring to?
- I can add copyright.
- It is a utility that can test deflate() under various parameters. It can do raw deflate or gzip. It depends on what value you pass in for the `-w` (windowBits) parameter. I can rename it minideflate if you like that better.
- There is a separate program called inflateParams in #475 that is used for inflate(). I think it is better separated into two programs. One for inflate and one deflate. The parameters are different for each program. If I rename the other to minideflate, I can name this one miniinflate (notice the double i's).
- What would I do with that extension? The app outputs to stdout. The dat file I generated is just the smallest random binary data that reproduces the problem defneg3 issue.

test/deflateparams.c
```diff
@@ -0,0 +1,261 @@
...
+        CHECK_ERR(err, "deflate");
+
+    err = PREFIX(deflateEnd)(&c_stream);
+    CHECK_ERR(err, "deflateEnd");
+
+    fwrite(compr, 1, c_stream.total_out, stdout);
+    free(compr);
+}
+
+/* ===========================================================================
```
Dead2: This is the help text. Although it would have been nice if it could print that instead of only being a comment in the source.

Dead2:
I was thinking this could be an opposite to minigzip, one that could output deflate streams instead (those are actually smaller than gzip streams due to having a smaller header).

But if you want to make a pure test application, then it should have a name prefixed by 'test' to avoid people using it to make deflate files.

defneg3.dat should have copyright info nevertheless, we have to tighten up our lack of copyright info here and there. This might need to be traceable in the future, so better to put in the extra effort now.

The .zz extension comment was conditioned by an if. If it cannot do files, then of course it cannot use file extensions.

nmoinvaz:
I think this is more of a test tool. It could be used for non-test purposes if somebody wanted to. The problem with prefixing it by 'test' is that it is not consistent with the other test applications. switchlevels.c is a test app but it is not prefixed by 'test'. deflateparams tool is not like the example project where it has a bunch of small tests. We could rename everything to prefix it with 'test'. But I imagine only the project name would be prefixed with 'test' and not the actual source filename. Would we then also prefix 'example' project with 'test_?'. I write question mark because we had some discussions about it being renamed previously. I think perhaps all that might be better as a separate PR to make everything consistent.

test/deflateparams.c
nmoinvaz: I think maybe you selected the wrong lines but I think I know what you are talking about.

nmoinvaz: I think maybe you selected the wrong lines but I think I know what you are talking about.

Dead2: Weird, I didn't select any lines, I put the comment on the first line of the in-code-comment. If you go see it in GitHubs code viewer, it'll be pretty obvious.
I guess GitHub expects all comments to come after the code.

mtl1979: I agree, this should be printed on console if amount of command-line parameters is wrong or user specifies "--help" explicitly...

nmoinvaz: I have noticed that switchlevels and minigzip are also missing help information. Maybe we can do all three apps at once in a separate PR.

nmoinvaz:
@Dead2 I have added the licensing information that you wanted and moved the test file to GH-382 directory.

nmoinvaz:
Rebased.

Dead2:
What is the status of this (except it obviously needs a rebase)?

nmoinvaz:
The `deflateparams` utility is still a good testing tool. The original test case (deflateEnd returning -3) for it is still valid. It can be used for fine tuned testing deflate() using the command line only.

However, because not all code paths are used in unit testing (namely `deflate_params_large` and `deflate_params`), it may decrease code coverage %.

nmoinvaz:
Rebase done.

test/deflateparams.c
nmoinvaz: Help text for minigzip and switchlevels is now in #623.

nmoinvaz: Added help text.

nmoinvaz:
I have reworked the _deflateparams_ to work more like _minigzip_. I went back to the original commit to verify that my unit test still produced -3.

nmoinvaz:
Ready for review.

Dead2:
Looks good. Might want to add `--help` support before merging though.

nmoinvaz:
I am thinking of renaming it to minideflate and adding inflate functionality. @Dead2 what do you think? inflate portion will respect windowBits setting. Just like minigzip does both deflate/inflate. This will allow specialized deflate/inflate..

Dead2:
I think that would be great.

PR #475, opened 2019-10-30
Added unit test for inflate returning Z_BUF_ERROR during Z_FINISH 

nmoinvaz:
This adds unit test for the issue reported in #397.


nmoinvaz:
Even though the underlying test case for this `inflateparams` tool is no longer valid, the tool itself could be useful. Not sure what we want to do here. Maybe close this issue, we can resurrect it if we have a use case that requires this type of testing.

nmoinvaz:
I have moved the test app in this PR to a gist in case we want to end up using it one day.
https://gist.github.com/nmoinvaz/e0e315c41d4d6140d3aeff5e8e221204

Issue #478, opened 2019-11-24
armv6/v7: overriding -march, -mfloat-abi and --mfpu with incompatible values


MilhouseVH:
https://github.com/zlib-ng/zlib-ng/pull/386/files#diff-af3b638bc2a3e6c650974192a53c7291R189-R204 incorrectly detects `softfp` when cross-compiling on Linux as it overrides `-mfloat-abi` and `-mfpu` unnecessarily.

`zlib-ng` also incorrectly sets `-march=armv8-a+crc` when building for arvm7 and armv6.

From `CMakeOutput.log` (building for armv7, Raspberry Pi2):
```
The target system is: Linux - 1 - arm
The host system is: Linux - 4.4.0-166-generic - x86_64
Compiling the C compiler identification source file "CMakeCCompilerId.c" succeeded.
Compiler: /home/ubuntu/projects/LibreELEC.tv/build.LibreELEC-RPi2.arm-9.80-devel-next/toolchain/bin/armv7ve-libreelec-linux-gnueabi-gcc
Build flags: -march=armv7ve;-mtune=cortex-a7;-mabi=aapcs-linux;-Wno-psabi;-Wa,-mno-warn-deprecated;-mfloat-abi=hard;-mfpu=neon-vfpv4;-Wall;-pipe;;-O2;-fomit-frame-pointer
Id flags:
```

The result of `-dumpmachine` from gcc-9.2.0 when cross-compiling for armv7 is:
```
armv7ve-libreelec-linux-gnueabi
```

This results in:
```
-- The following features have been enabled:

 * CMAKE_BUILD_TYPE, Build type: MinSizeRel (selected)
 * ZLIB_COMPAT, Provide a zlib-compatible API
 * WITH_OPTIM, Build with optimisation
 * WITH_NEW_STRATEGIES, Use new strategies
 * WITH_ACLE, Build with ACLE CRC
 * WITH_NEON, Build with NEON intrinsics
 * NEON_FILLWINDOW, Support NEON instructions in fill_window_arm, using "-mfloat-abi=softfp -mfpu=neon"
 * ACLE_CRC, Support CRC hash generation using the ACLE instruction set, using "-march=armv8-a+crc"

-- The following features have been disabled:

 * WITH_GZFILEOP, Compile with support for gzFile-related functions
 * WITH_SANITIZERS, Build with address sanitizer and all supported sanitizers other than memory sanitizer
 * WITH_MSAN, Build with memory sanitizer
[...]

PR #480, opened 2019-11-25
configure: relax arm detection

MilhouseVH:
As discussed, [here](https://github.com/zlib-ng/zlib-ng/issues/478#issuecomment-557960064), this check is too specific and misses many popular pre-armv8 variants.


Dead2:
Thanks :)

PR #484, opened 2019-11-25
Compatibility fix for crc32()

### LINKED ISSUE #483: mariadb: error: conflicting types for 'crc32'
Building latest `mariadb-10.4.10` on Linux with latest `zlib-ng` (5bd3c582908503135fd8d6326ca0e2a6a947b44f from develop branch) and `--zlib-compat` for x86_64 host (not cross-compiling) results in the following build failure:
```
[16/179] Building C object mysys/CMakeFiles/mysys.dir/my_compress.c.o
FAILED: mysys/CMakeFiles/mysys.dir/my_compress.c.o 
/home/ubuntu/projects/LibreELEC.tv/build.LibreELEC-Generic.x86_64-9.80-devel/toolchain/bin/host-gcc -DHAVE_CONFIG_H -D_FILE_OFFSET_BITS=64 -I../wsrep-lib/include -I../wsrep-lib/wsrep-API/v26 -Iinclude -I../include -I../mysys -march=native -O2 -Wall -pipe -I/home/ubuntu/projects/LibreELEC.tv/build.LibreELEC-Generic.x86_64-9.80-devel/toolchain/include -Wno-format-security -pie -fPIC -Wl,-z,relro,-z,now -fstack-protector --param=ssp-buffer-size=4 -O2 -g -DNDEBUG -D_FORTIFY_SOURCE=2 -DDBUG_OFF   -fPIC -MD -MT mysys/CMakeFiles/mysys.dir/my_compress.c.o -MF mysys/CMakeFiles/mysys.dir/my_compress.c.o.d -o mysys/CMakeFiles/mysys.dir/my_compress.c.o   -c ../mysys/my_compress.c
In file included from ../mysys/my_compress.c:24:0:
/home/ubuntu/projects/LibreELEC.tv/build.LibreELEC-Generic.x86_64-9.80-devel/toolchain/include/zlib.h:1683:26: error: conflicting types for 'crc32'
 ZEXTERN uint32_t ZEXPORT crc32(uint32_t crc, const unsigned char *buf, uint32_t len);
                          ^
In file included from ../mysys/my_compress.c:20:0:
../include/my_sys.h:932:14: note: previous declaration of 'crc32' was here
 extern ulong crc32(ulong crc, const uchar *buf, uint len);
              ^
```
`mariadb` build log: http://ix.io/22Qz

No such build issue with upstream `zlib-1.2.11`.

Using `gcc-9.2.0`.

I don't currently have a fix for this, but happy to test any proposals.

mtl1979:
* Use "unsigned long" externally instead of "uint32_t" for crc
* Use "unsigned int" externally instead of "uint32_t" for len

Fixes #483

### DIFF
--- a/crc32.c
+++ b/crc32.c
@@ -64,9 +64,15 @@ ZLIB_INTERNAL uint32_t crc32_generic(uint32_t crc, const unsigned char *buf, uin
     return crc ^ 0xffffffff;
 }
 
+#ifdef ZLIB_COMPAT
+unsigned long ZEXPORT PREFIX(crc32)(unsigned long crc, const unsigned char *buf, unsigned int len) {
+    return (unsigned long) PREFIX(crc32_z)((uint32_t) crc, buf, len);
+}
+#else
 uint32_t ZEXPORT PREFIX(crc32)(uint32_t crc, const unsigned char *buf, uint32_t len) {
     return PREFIX(crc32_z)(crc, buf, len);
 }
+#endif
 
 /*
    This BYFOUR code accesses the passed unsigned char * buffer with a 32-bit
--- a/zlib.h
+++ b/zlib.h
@@ -1680,7 +1680,7 @@ ZEXTERN uint32_t ZEXPORT adler32_combine(uint32_t adler1, uint32_t adler2, z_off
    negative, the result has no meaning or utility.
 */
 
-ZEXTERN uint32_t ZEXPORT crc32(uint32_t crc, const unsigned char *buf, uint32_t len);
+ZEXTERN unsigned long ZEXPORT crc32(unsigned long crc, const unsigned char *buf, unsigned int len);
 /*
      Update a running CRC-32 with the bytes buf[0..len-1] and return the
    updated CRC-32.  If buf is NULL, this function returns the required


MilhouseVH:
Many thanks, `mariadb` now builds successfully!

crc32.c
```diff
@@ -64,9 +64,15 @@ ZLIB_INTERNAL uint32_t crc32_generic(uint32_t crc, const unsigned char *buf, uin
     return crc ^ 0xffffffff;
 }
 
+#ifdef ZLIB_COMPAT
+unsigned long ZEXPORT crc32(unsigned long crc, const unsigned char *buf, unsigned int len) {
+    return (unsigned long) crc32_z((uint32_t) crc, buf, len);
+}
+#else
 uint32_t ZEXPORT PREFIX(crc32)(uint32_t crc, const unsigned char *buf, uint32_t len) {
```
Dead2, line 72: Since this branch is zlib-ng native, we should perhaps not use the PREFIX(). Either that or we should use it in both branches.
Come to think of it, we might want to use it after all to enable custom function prefixes, since some (few) users do require that to avoid clashes between a statically compiled-in zlib(-ng) and a dl()'ed zlib(-ng) from for example a 3rd party plugin. Lets keep using PREFIX() to ease the use in those cases and reduce the amount of unexpected compile/runtime failures for those users.

mtl1979: Updated

Dead2: You forgot the return call to crc32_z :)

mtl1979: Updated again ;)

Dead2: Doesn't the len parameter also need a cast to avoid warnings?

mtl1979: "unsigned int" is either 16-bit or 32-bit, "uint32_t" is always 32-bit type. Because it would convert to same or larger size type, it doesn't need cast.

Dead2: Doh, of course it is. I was somehow reading both parameters as unsigned long. I so prefer the stdint names. :)

Issue #471, opened 2019-10-25
Debug configuration asserts in CI


nmoinvaz:
There are two debug mode asserts that I have run into on CI:

**compress_fuzzer**
Using input [match2.zip](https://github.com/zlib-ng/zlib-ng/files/3770892/match2.zip). Original extension is pdb.

```
scan += 2, match+=2;
Assert(*scan == *match, "match[2]?");
```
There are multiple of the same line but the one that was hit was in match_p.h at line 460. The comment above this line suggests that scan and match don't have to be equal. 

**example_dict_fuzzer**
Using input [CMakeLists.txt](https://github.com/zlib-ng/zlib-ng/files/3770909/CMakeLists.txt).

```
Assert((uint16_t)dist < (uint16_t)MAX_DIST(s) &&
               (uint16_t)lc <= (uint16_t)(MAX_MATCH-MIN_MATCH) &&
               (uint16_t)d_code(dist) < (uint16_t)D_CODES,  "zng_tr_tally: bad match");
```
This is in trees.c at line 747.

I'm not sure if these asserts are necessary. Maybe they can be converted to `Tracev` statements instead of `Assert` statements.

nmoinvaz:
Or perhaps we want to turn off fuzzers in debug configuration in CI.

Dead2:
That depends on whether these bugs are caused by the fuzzing itself or whether the fuzzing triggered a real bug.
If it is caused by the fuzzing, then perhaps we should bypass them by using a define check ZFUZZING for example. But if these are real bugs, then we would be much better off fixing them.
So this will need some investigation I guess.

mtl1979:
First two bytes of scan and match are not equal when it runs out of input... That check is incorrect as it should check MIN_MATCH bytes, which is 3 bytes, not 2... In some places we already reordered the test to avoid the assert completely.

Dead2:
I don't have time to work on this real soon, do any of you want to suggest PRs for one or both of these?
Without actually looking at the code, I am thinking perhaps the first assert might benefit from another conditional so it does not trigger on end-of-input, if that check is in the assert then it also won't affect performance of release builds.

The second one kind of looks like it could potentially be a real bug, and perhaps checking whether stock zlib does something different there, or even whether it fails on the same fuzzer. (Can our fuzzers run on stock zlib btw? That would actually be a great feature if they could.)

nmoinvaz:
[...]

PR #490, opened 2019-12-12
Modified zng_tr_tally to check for proper bounds per strategy

nmoinvaz:
This should fix one of the debug asserts in #471. Since these functions only get called when `ZLIB_DEBUG` is defined, it shouldn't affect performance.

### DIFF
--- a/deflate.h
+++ b/deflate.h
@@ -334,7 +334,7 @@ void ZLIB_INTERNAL slide_hash_c(deflate_state *s);
 
         /* in trees.c */
 void ZLIB_INTERNAL zng_tr_init(deflate_state *s);
-int ZLIB_INTERNAL zng_tr_tally(deflate_state *s, unsigned dist, unsigned lc);
+int ZLIB_INTERNAL zng_tr_tally(deflate_state *s, unsigned dist, unsigned lc, unsigned max_dist);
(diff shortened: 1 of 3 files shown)


nmoinvaz:
Another (probably safer alternative) is to abandon use of `MAX_DIST2` and replace it with `MAX_DIST(s)`.   By using `MAX_DIST2` we prevent anybody from inflating who has compiled with `INFLATE_STRICT`.

mtl1979:
We should document defines that are incompatible and make compiling fail hard if conflicting ones are used...

nmoinvaz:
@Dead2 what do you want to do in this instance? I think it would be better to stick with `MAX_DIST` everywhere for maximum compatibility. Compression might suffer when window size is less than the max. Stock zlib doesn't have a `MAX_DIST2`. We could do something like `DEFLATE_FORCE_MAX_DISTANCE` which would use `MAX_DIST2` but if not set it would default to `MAX_DIST`. If we did something like that, we would still need this PR. Perhaps the best solution is just remove `MAX_DIST2` altogether.

Dead2:
I think we should make sure we make deflate streams that are standards compliant. So if inflate strict fails, then that signifies a bug, so MAX_DIST2 should likely be changed.

nmoinvaz:
Closing this in favor of #493.

PR #491, opened 2019-12-12
Fixed architecture detection when cross compiling.

nmoinvaz:


cmake/archdetect.cmake
```diff
@@ -43,15 +45,14 @@ else()
     message(STATUS "Arch detected: '${ARCH}'")
 endif()
 
-
```
mtl1979: Mixing code and white-space changes is a no-go...

nmoinvaz: I understand that from before. Personally I am not so strict about it.

mtl1979: I'm just reminding...

PR #492, opened 2019-12-12
Added infcover code coverage test executable to cmake test.

nmoinvaz:
I modified infcover.c to compile with zlib-ng.

It fails on this test:
``
try("c c0 81 0 0 0 0 0 90 ff 6b 4 0", "invalid distance too far back", 1);
``
With inflateBack returning `Z_BUF_ERROR`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -903,6 +903,9 @@ if (ZLIB_ENABLE_TESTS)
     add_executable(switchlevels test/switchlevels.c)
     configure_test_executable(switchlevels)
 
+    add_executable(infcover test/infcover.c inftrees.c)
+    configure_test_executable(infcover)
+
     add_executable(makefixed tools/makefixed.c inftrees.c)
     target_include_directories(makefixed PUBLIC ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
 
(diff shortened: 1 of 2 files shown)


nmoinvaz:
Looks like it works with `INFLATE_STRICT` enabled.

nmoinvaz:
Now I recompile zlib with `INFLATE_STRICT` and use it for infcover.

nmoinvaz:
There is a problem with compiling with `INFLATE_STRICT` because deflate_medium strategy uses `MAX_DIST2` which is not compatible.

Dead2:
Is MAX_DIST2 also the reason why this requires to be built with INFLATE_STRICT? If not, I really think we should try to figure out why it requires that, because I think that might signal another bug.

nmoinvaz:
Yes it is the reason. If we remove MAX_DIST2 then we don’t need INFLATE_STRICT to be defined.

mtl1979:
Removing MAX_DIST2 will have penalty in compression rate and that will invalidate any comparisons with other forks that have deflate_medium with MAX_DIST2 intact.

Dead2:
@mtl1979 I am open to alternatives that makes standards-compliant bitstreams. If deflate-medium does not do so currently, then it needs to be fixed.

mtl1979:
@Dead2 I'm just saying it's not a bug, it's a feature... Bugs by definition are undocumented, but this relaxation of distance restriction is well-documented...

Dead2:
If it does not decompress with inflate_strict and zlib itself does not make such streams, then I believe it is likely not standards compliant, but I could be wrong. Honestly I don't see the problem with the value of MAX_DIST2, but obviously it matters. Possibly it is just about setting the window size to the max in the output header since MAX_DIST2 forces the full window size instead of using the currently selected window size (so inflate will know this and allocate a big enough window, etc).

In any case, by default we should create standards compliant streams, if not then we are not really zlib compatible. Non-compliant features could be kept as a special runtime setting operated through the new feature interface implemented for the DFLTCC features, and possibly an off-by-default compile setting. If MAX_DIST2 can be fixed, then we could possibly use that for all the deflate strategies.

mtl1979:
Long time ago I actually suggested disabling deflate_medium completely because it had multiple issues....

nmoinvaz:
I will need to take a look at this again after #493 is merged.

nmoinvaz:
Rebased and fixed.

PR #493, opened 2020-01-02
Removed MAX_DIST2 which is incompatible with INFLATE_STRICT.

nmoinvaz:
This PR removes `MAX_DIST2` in favor of `MAX_DIST(s)` which makes the output produced by deflate medium compatible when zlib is compiled with `INFLATE_STRICT` (see #492). It also fixes the check for proper bounds in zng_tr_tally as mentioned in #490 (so debug CI finally succeeds).

Once this is merged I can clean up or close the other 2 PRs.

### DIFF
--- a/deflate_medium.c
+++ b/deflate_medium.c
@@ -20,8 +20,6 @@ struct match {
     unsigned int orgstart;
 };
 
-#define MAX_DIST2  ((1 << MAX_WBITS) - MIN_LOOKAHEAD)
-
 static int tr_tally_dist(deflate_state *s, int distance, int length) {
     return zng_tr_tally(s, distance, length);
 }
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Once this is merged I can clean up the other 2 PRs.

Dead2:
Looks good :)

Dead2:
Surprisingly this seems to improve compression speed by ~0.1-0.2%.
I had expected it to become worse due to the extra indirect lookup (I suspect MAX_DIST2's purpose was to get rid of that), so I re-ran the before and after tests several times and with this it is always faster than even the best before run.

Dead2:
Oh, compressed size was the same in all my tests.

nmoinvaz:
Thank you for merging in, now the CI tests can finally pass.

Issue #487, opened 2019-11-30
Zlib-ng's CI detection in cmake causes my own CI to fail


malytomas:
There in the cmake:
https://github.com/zlib-ng/zlib-ng/blob/3ea747396dcc3d63924eedbdf05bacf7866e1b29/CMakeLists.txt#L100
you detect CI and change build configuration based on it.
This causes my own CI for a project that would integrate zlib-ng to fail at link stage with missing __gcov_init and few more symbols.
Furthermore, it prevents CI from testing the cmake configuration without the maintainer and codecoverage switches.
I suggest to set the maintainer switch directly in travis.yml instead (and also have at least one test without it).

mtl1979:
Looks like @Dead2 needs to revert the commit adding that if() block and do it in more compatible way...

nmoinvaz:
I think this can be closed now.

PR #495, opened 2020-01-06
Enable CI options explicitly so forked repositories don't fail CI.

nmoinvaz:
See issue #487.


Dead2:
I don't think "maintenance" is the right word here. The point is to convey that this is an option meant to be used by the maintainers. Similar to autoconf's "--enable-maintainer-mode" option for example. But "maintenance" means something slightly different and I fear the meaning gets lost.

nmoinvaz:
Yeah I didn't really know what words to use instead so I just went with `WITH_CODE_COVERAGE` and `WITH_MAINTENANCE_WARNINGS`.

nmoinvaz:
I could change it to something like, `WITH_ALL_WARNINGS`.

Dead2:
WITH_MAINTAINER_WARNINGS would be ideal I think, since we don't enable _ALL_ the warnings, but we enable warnings that makes no sense for an user to care about.

nmoinvaz:
I have renamed the option to `WITH_MAINTAINER_WARNINGS` and added it and `WITH_CODE_COVERAGE` to README.md.

Dead2:
I think you should revert the namechange of WARNFLAGS_MAINTAINER to WARNFLAGS_MAINTENANCE too, to avoid confusion.

nmoinvaz:
Done.

PR #497, opened 2020-01-10
Fixed formatting.

nmoinvaz:


mtl1979:
Number of spaces still seem inconsistent in preprocessor lines.

nmoinvaz:
I used 2 spaces intent for preprocessor lines. It seemed to be the majority style.

test/fuzz/checksum_fuzzer.c
```diff
@@ -13,72 +13,73 @@
...
-  uint32_t adler0 = PREFIX(adler32)(0L, NULL, 0);
-  uint32_t adler1 = adler0;
-  uint32_t adler2 = adler0;
-  uint32_t combine1, combine2;
-  /* Checksum with a buffer of size equal to the first byte in the input. */
-  uint32_t buffSize = data[0];
-  uint32_t offset = 0;
-  uint32_t op[32];
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t dataLen)
+{
```
Dead2: We try to use { at the same line as the function definition, so not sure why you changed it this one place..

Dead2:
This is a big step in the right direction.
I'd prefer a different style of #if etc indentation though.

Since all modern compilers ignore leading whitespace, and syntax highlighting is getting pretty widespread, I'd like #if within functions to be indented like the rest of the code, and code within the #if to also get indented like it would with a normal if.

example pseudocode:
``` C
void hello(){
    #ifdef foo
        printf("bar");
        #ifdef bar
            printf("foo");
        #endif
    #endif
    printf("baz");
}
```

I think that is a lot easier to read than this:
``` C
void hello(){
#ifdef foo
    printf("bar");
#ifdef bar
    printf("foo");
#endif
#endif
    printf("baz");
}
```
#if blocks outside of functions gets a bit more iffy. Multiple #if blocks within each other should trigger indentation. But #if blocks containing functions should not indent the function within.

nmoinvaz:
I think that style of #ifdef is a big departure from malder/zlib #ifdef style.

mtl1979:
I still think indenting preprocessor statements should be separate from indenting the actual code... It confuses people when code that belongs to same compiler block scope level has inconsistent indentation... It then looks like those preprocessor #if blocks are actually normal if blocks which they obviously are not.

mtl1979:
Only exception to the preprocessor indentation rule should in my opinion be the header guard, which doesn't increase the indentation level for inner preprocessor blocks... so both the header guard and the next inner #if block will have no spaces. This is because the header guard is essentially same as "#pragma once" which is single statement even though it has scope of full file.

Dead2:
Since you seem to feel strongly about the preprocessor indentation, I'll accept that even though I personally think the code looks very messy when functions has a many #if blocks.

There is still the inline comment above though.

test/fuzz/checksum_fuzzer.c
nmoinvaz: Done. I have also made all the other functions the same with the brace on the same line as the function definition.

nmoinvaz:
Ok. I think my goal was to keep it similar to original zlib style but to try and make it more uniform.

Issue #451, opened 2019-10-12
quick_insert_string needs MSVC inline assembly


nmoinvaz:
The function `quick_insert_string` that is part of deflate_quick.c needs MSVC inline assembly. 

```
static inline Pos quick_insert_string(deflate_state *const s, const Pos str) {
    Pos ret;
    unsigned h = 0;

#ifdef _MSC_VER
    h = _mm_crc32_u32(h, *(unsigned *)(s->window + str));
#else
    __asm__ __volatile__ (
        "crc32l (%[window], %[str], 1), %0\n\t"
    : "+r" (h)
    : [window] "r" (s->window),
      [str] "r" ((uintptr_t)str)
    );
#endif
```

Notice how `insert_string_sse` has the MSVC inline assembly version:

```
#if defined(X86_SSE42_CRC_INTRIN)
#  ifdef _MSC_VER
        h = _mm_crc32_u32(h, val);
#  else
        h = __builtin_ia32_crc32si(h, val);
#  endif
#else
#  ifdef _MSC_VER
        __asm {
            mov edx, h
            mov eax, val
            crc32 eax, edx
            mov val, eax
        };
#  else
        __asm__ __volatile__ (
            "crc32 %1,%0\n\t"
            : "+r" (h)
            : "r" (val)
        );
#  endif
#endif
```
It is also missing `#if defined(X86_SSE42_CRC_INTRIN)` and `h = __builtin_ia32_crc32si(h, val);`. I don't know enough assembly to do it. I figured it might be an easier task for somebody else.

nmoinvaz:
The solution I had come up with was:

```
static inline Pos quick_insert_string(deflate_state *const s, const Pos str) {
    Pos ret;
    unsigned h = 0;
    unsigned int val;

    memcpy(&val, (unsigned *)(s->window + str), sizeof(val));

#if defined(X86_SSE42_CRC_INTRIN)
#  ifdef _MSC_VER
        h = _mm_crc32_u32(h, val));
#  else
        h = __builtin_ia32_crc32si(h, val);
#  endif
#else
#  ifdef _MSC_VER
        __asm {
            mov edx, h
            mov eax, val
            crc32 eax, edx
            mov val, eax
        };
#  else
        __asm__ __volatile__ (
            "crc32 %1,%0\n\t"
            : "+r" (h)
            : "r" (val)
        );
#  endif
#endif
```

[...]

PR #499, opened 2020-01-18
Update assembly in quick_insert_string and support unaligned reads

nmoinvaz:
This PR does two things:

- Adds unaligned conditionally compiled code for insert_string_sse and quick_insert_string.
- Unifies sse42 crc32 assembly between insert_string_sse and quick_insert_string. #451

### DIFF
--- a/arch/x86/deflate_quick.c
+++ b/arch/x86/deflate_quick.c
@@ -201,16 +201,35 @@ static void static_emit_end_block(deflate_state *const s, int last) {
 static inline Pos quick_insert_string(deflate_state *const s, const Pos str) {
     Pos ret;
     unsigned h = 0;
+    unsigned int val;
 
-#ifdef _MSC_VER
-    h = _mm_crc32_u32(h, *(unsigned *)(s->window + str));
+#if defined(UNALIGNED_OK)
+    val = *(unsigned int *)(s->window + str);
 #else
+    memcpy(&val, s->window + str, sizeof(val)); 
+#endif
+
+#if defined(X86_SSE42_CRC_INTRIN)
+#  ifdef _MSC_VER
+    h = _mm_crc32_u32(h, val);
+#  else
+    h = __builtin_ia32_crc32si(h, val);
+#  endif
+#else
+#  ifdef _MSC_VER
+    __asm {
+        mov edx, h
+        mov eax, val
+        crc32 eax, edx
+        mov val, eax
+    };
+#  else
     __asm__ __volatile__ (
-        "crc32l (%[window], %[str], 1), %0\n\t"
-    : "+r" (h)
-    : [window] "r" (s->window),
-      [str] "r" ((uintptr_t)str)
+        "crc32 %1,%0\n\t"
+        : "+r" (h)
+        : "r" (val)
     );
+#  endif
 #endif
 
     ret = s->head[h & s->hash_mask];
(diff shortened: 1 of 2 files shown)


arch/x86/deflate_quick.c
```diff
@@ -201,16 +201,35 @@ static void static_emit_end_block(deflate_state *const s, int last) {
...
     Pos ret;
     unsigned h = 0;
+    unsigned int val;
 
-#ifdef _MSC_VER
-    h = _mm_crc32_u32(h, *(unsigned *)(s->window + str));
+#if defined(UNALIGNED_OK)
+    val = *(unsigned int *)(s->window + str);
+#else
+    memcpy(&val, (unsigned *)(s->window + str), sizeof(val));
```
mtl1979: This is wrong... memcpy parameters should always use "unsigned char *" or "char *", never pointers to larger types. This is because compiler might implement memcpy as a macro.

nmoinvaz: Interesting ok.

nmoinvaz: Fixed.

arch/x86/deflate_quick.c
```diff
@@ -201,16 +201,35 @@ static void static_emit_end_block(deflate_state *const s, int last) {
...
-        "crc32l (%[window], %[str], 1), %0\n\t"
-    : "+r" (h)
-    : [window] "r" (s->window),
-      [str] "r" ((uintptr_t)str)
+        "crc32 %1,%0\n\t"
+        : "+r" (h)
+        : "r" (val)
     );
+#  endif
 #endif
```
mtl1979, line 233: Instead of duplicating the code, it should be moved to private header as inline function.

nmoinvaz: Any thoughts on the name of the function?

mtl1979: Obviously it should not conflict with existing intrinsic names... and the name should clearly show the size of the data parameter (uint32)...

nmoinvaz: It doesn't seem to be possible because of the statement:
```
        if (s->level >= TRIGGER_LEVEL)
            val &= 0xFFFFFF;
```
I wonder if that is even necessary.

mtl1979: That statement is no issue as the inline function would start right after it.

nmoinvaz: What do you think about changing insert_string to be a function in deflate.c and doing a  `functable.compute_hash` function instead? The majority of the insert_string functions are the same except for the assembly for the hash.

mtl1979:
Honestly thinking that was the next step I was considering...

nmoinvaz:
Do you think there would be much of a performance hit by calling compute_hash function for each loop in insert_string? Before it was all inline.

mtl1979:
Indirect jump will always have performance hit, that's why I suggested inlining as much as possible.

nmoinvaz:
I don't think inlining would make a difference in this instance because it would be a call to a function pointer `functable.compute_hash`, right?

mtl1979:
It's one or the other, not both...

Dead2:
I'd suggest making an inlined function in a header file named for example zng_crc32c. (For those who don't know already; CRC-32C is the common name for the CRC32 variant used by SSE4.2, the last C stands for Castagnoli, and it should not be confused with the CRC32 variant used to detect errors in gzip files).

Also, the two changes in this PR are solving separate problems and in separate files, so I'd appreciate if they were in separate commits as well.

nmoinvaz:
This PR is superseded by #505.

Issue #427, opened 2019-09-20
configure script does weird things on osx


Dead2:
See: https://travis-ci.org/zlib-ng/zlib-ng/jobs/587639851

It does not fail, but it seems sed is not happy with us.

```
$ export GENERATOR="./configure --warn"

$ $GENERATOR
Checking for compiler... gcc
Checking for shared library support... Building shared library libz-ng.1.9.9.dylib with gcc.
Checking for off64_t... No.
Checking for _off64_t... No.
Checking for fseeko... Yes.
Checking for strerror... Yes.
Checking for unistd.h... Yes.
Checking for ptrdiff_t... Yes.
Checking for ANSI C compliant compiler...  Yes.
Checking for attribute(visibility(hidden)) support... Yes.
Checking for attribute(visibility(internal)) support... Yes.
Checking for __builtin_ctzl ... Yes.
Checking for SSE4.2 CRC intrinsics ... Yes.
Checking for PCLMULQDQ intrinsics ... Yes.
Checking for sys/sdt.h ... Yes.
ARCH: x86_64
Using arch directory: arch/x86
sed: 1: "Makefile": invalid command code M
sed: 1: "Makefile": invalid command code M
sed: 1: "Makefile": invalid command code M
sed: 1: "Makefile": invalid command code M
sed: 1: "Makefile": invalid command code M
sed: 1: "Makefile": invalid command code M
sed: 1: "Makefile": invalid command code M
sed: 1: "Makefile": invalid command code M
sed: 1: "Makefile": invalid command code M
sed: 1: "Makefile": invalid command code M
sed: 1: "Makefile": invalid command code M
sed: 1: "Makefile": invalid command code M
sed: 1: "Makefile": invalid command code M
sed: 1: "Makefile": invalid command code M
sed: 1: "arch/x86/Makefile": command a expects \ followed by text
sed: 1: "arch/x86/Makefile": command a expects \ followed by text
sed: 1: "arch/x86/Makefile": command a expects \ followed by text
sed: 1: "arch/x86/Makefile": command a expects \ followed by text
sed: 1: "arch/x86/Makefile": command a expects \ followed by text
sed: 1: "arch/x86/Makefile": command a expects \ followed by text
sed: 1: "arch/x86/Makefile": command a expects \ followed by text
sed: 1: "arch/x86/Makefile": command a expects \ followed by text
sed: 1: "arch/x86/Makefile": command a expects \ followed by text
[...]

PR #500, opened 2020-01-19
Fixed sed in configure script on macOS

nmoinvaz:
Changes:

- Fixed configure script use of sed on macOS. (see #427)
- Added macOS configure build (which was used to test this PR against)


mtl1979:
I remember that the sed incompatibility was discussed earlier... I think we should try to be as compatible as possible without making the scripts unreadable.

Issue #326, opened 2019-03-08
What tools do you use for benchmarking?


bjne:
Today I made some benchmarks using jsnell/zlib-bench, using the following config:

```
my @versions = (
    { id => 'baseline', repository => 'https://github.com/madler/zlib.git', commit_or_branch => 'cacf7f1' },
    { id => 'cloudflare', repository => 'https://github.com/cloudflare/zlib.git', commit_or_branch => 'e55212b' },
    { id => 'intel', repository => 'https://github.com/jtkukunas/zlib.git', commit_or_branch => '4b0ef4e'},
    { id => 'zlib-ng', repository => 'https://github.com/zlib-ng/zlib-ng.git', commit_or_branch => '3ee7109', CONFIGURE_FLAGS => '--zlib-compat'},
);
```

None of the results are in favor of zlib-ng, even tho it afaik uses patches from both intel and cf?
Am I missing something here? Why does intel outperform everyone on decompression?
What does cloudflare do at level 9?

results:
```
                     baseline               cloudflare             intel                  zlib-ng

 dec exec  time [s]:  1.78 ± 0.00 (100.00%)  1.52 ± 0.00 ( 85.74%)  1.37 ± 0.00 ( 77.07%)  1.56 ± 0.00 ( 87.75%)
 dec html  time [s]:  1.02 ± 0.00 (100.00%)  0.87 ± 0.00 ( 85.42%)  0.68 ± 0.00 ( 67.01%)  0.90 ± 0.00 ( 88.32%)
 dec jpeg  time [s]:  0.24 ± 0.00 (100.00%)  0.13 ± 0.00 ( 52.57%)  0.14 ± 0.00 ( 60.33%)  0.24 ± 0.00 (100.84%)
 dec pngp  time [s]:  1.12 ± 0.00 (100.00%)  0.86 ± 0.00 ( 76.44%)  0.74 ± 0.00 ( 66.07%)  0.89 ± 0.00 ( 79.71%)

compress executable -5 (2 iterations)
 exec 0.33 time [s]:  0.47 ± 0.00 (100.00%)  0.25 ± 0.00 ( 53.22%)  0.26 ± 0.00 ( 55.07%)  0.27 ± 0.00 ( 58.73%)
 html 0.33 time [s]:  0.28 ± 0.00 (100.00%)  0.16 ± 0.00 ( 55.74%)  0.15 ± 0.00 ( 52.17%)  0.18 ± 0.00 ( 64.24%)
[...]

PR #502, opened 2020-01-22
Add initial AVX2 support.

mtl1979:
* Add detection for AVX2
* Use AVX2 optimized slide_hash when available

### DIFF
--- a/win32/Makefile.msc
+++ b/win32/Makefile.msc
@@ -23,7 +23,7 @@ AR = lib
 RC = rc
 CP = copy /y
 CFLAGS  = -nologo -MD -W3 -O2 -Oy- -Zi -Fd"zlib" $(LOC)
-WFLAGS  = -D_CRT_SECURE_NO_DEPRECATE -D_CRT_NONSTDC_NO_DEPRECATE -DX86_PCLMULQDQ_CRC -DX86_SSE2 -DX86_CPUID -DX86_SSE42_CRC_HASH -DUNALIGNED_OK -DX86_QUICK_STRATEGY
+WFLAGS  = -D_CRT_SECURE_NO_DEPRECATE -D_CRT_NONSTDC_NO_DEPRECATE -DX86_PCLMULQDQ_CRC -DX86_SSE2 -DX86_CPUID -DX86_SSE42_CRC_INTRIN -DX86_SSE42_CRC_HASH -DX86_AVX2 -DUNALIGNED_OK -DX86_QUICK_STRATEGY
 LDFLAGS = -nologo -debug -incremental:no -opt:ref -manifest
 ARFLAGS = -nologo
 RCFLAGS = /dWIN32 /r
@@ -36,7 +36,7 @@ SUFFIX =
 
 OBJS = adler32.obj compress.obj crc32.obj deflate.obj deflate_fast.obj deflate_quick.obj deflate_slow.obj \
        deflate_medium.obj \
-       functable.obj infback.obj inflate.obj inftrees.obj inffast.obj slide_sse.obj trees.obj uncompr.obj zutil.obj \
+       functable.obj infback.obj inflate.obj inftrees.obj inffast.obj slide_avx.obj slide_sse.obj trees.obj uncompr.obj zutil.obj \
        x86.obj fill_window_sse.obj insert_string_sse.obj crc_folding.obj
 !if "$(ZLIB_COMPAT)" != ""
 WITH_GZFILEOP = yes
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,6 +144,7 @@ if(${CMAKE_C_COMPILER} MATCHES "icc" OR ${CMAKE_C_COMPILER} MATCHES "icpc" OR ${
         set(WARNFLAGS_MAINTAINER "-W4 -Wcheck")
         set(WARNFLAGS_DISABLE "")
         if(BASEARCH_X86_FOUND)
+            set(AVX2FLAG "-mavx2")
             set(SSE2FLAG "-msse2")
             set(SSE4FLAG "-msse4.2")
         endif()
@@ -152,6 +153,7 @@ if(${CMAKE_C_COMPILER} MATCHES "icc" OR ${CMAKE_C_COMPILER} MATCHES "icpc" OR ${
         set(WARNFLAGS_MAINTAINER "/W4 /Wcheck")
         set(WARNFLAGS_DISABLE "")
         if(BASEARCH_X86_FOUND)
+            set(AVX2FLAG "/arch:AVX2")
             set(SSE2FLAG "/arch:SSE2")
             set(SSE4FLAG "/arch:SSE4.2")
         endif()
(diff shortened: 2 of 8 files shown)


win32/Makefile.msc
```diff
@@ -23,7 +23,7 @@ AR = lib
 RC = rc
 CP = copy /y
 CFLAGS  = -nologo -MD -W3 -O2 -Oy- -Zi -Fd"zlib" $(LOC)
-WFLAGS  = -D_CRT_SECURE_NO_DEPRECATE -D_CRT_NONSTDC_NO_DEPRECATE -DX86_PCLMULQDQ_CRC -DX86_SSE2 -DX86_CPUID -DX86_SSE42_CRC_HASH -DUNALIGNED_OK -DX86_QUICK_STRATEGY
+WFLAGS  = -D_CRT_SECURE_NO_DEPRECATE -D_CRT_NONSTDC_NO_DEPRECATE -DX86_PCLMULQDQ_CRC -DX86_SSE2 -DX86_CPUID -DX86_SSE42_CRC_INTRIN -DX86_SSE42_CRC_HASH -Dx86_AVX2 -DUNALIGNED_OK -DX86_QUICK_STRATEGY
```
nmoinvaz: I think this should be -DX86_AVX2 (uppercase X)

mtl1979: Fixed.

Dead2:
Since slide_avx.c is not your code, I'd appreciate it if that is added first as a separate commit with a short comment of where it came from.
I'll let the unrelated change in Makefile.msc slide ;)

mtl1979:
slide_avx.c is modified from slide_sse.c ... Only thing I changed was the intrinsic names and increments in the loops...

Dead2:
Ah, that explains it.

I think you should add rewrite the copyright text at the top then. Not sure how much you changed, but it sounds like it is no longer Intel's work. Alternatively you could just add a comment something like "Based on Intels slide_sse implementation, modified to use AVX2 instead".
Currently it reads like it is Intels work and that you modified it a little to fit zlib-ng datatypes and such.

mtl1979:
I tried to be as safe as possible regarding copyright laws... US laws are stricter than EU laws. There is certain percentage that needs to be different so that copyright text can be truncated and in my opinion there is not enough difference to warrant truncating the original copyright.

Dead2:
You can still add a comment like the one I suggested though.

mtl1979:
@Dead2 Better?

Dead2:
Perfect :)

mtl1979:
I can't run benchmarks as my Linux doesn't have AVX2, only Windows machine has...

Dead2:
I'll benchmark it soonTM :)

Dead2:
In my benchmarks this only improves performance on average by ~0.55% with cmake while performance seems unchanged with configure (Might be caused by configure building it static).

While it improves performance on average, I notice that it is not consistently better. I run each compression level 11 times, and some levels on average end up faster while some are slower.

So, do you want to have another look at things to see whether anything can be improved first?

mtl1979:
On my Windows machine, AVX2 code is marginally faster than SSE2 equivalent code when CPU load is 100%. I suspect the speed improvement is less if CPU load is lower than 100% due to most modern CPUs reducing CPU clock speed if CPU load is less than 100%.

nmoinvaz:
@Dead2 what are you using for benchmarking?

Dead2:
I use my own python tool that also disables power saving features such as c-states beyond C1 and forces a fixed clock speed to avoid turbo and/or throttling from influencing things. The core is also disabled as far as the linux scheduler is concerned, so no other tasks (or interrputs) interrupts the benchmark (with the exception of some RCU stuff I guess). And the files are located on a tmpfs (ramdisk), so repeatability is pretty good with mostly sub 0.1% variation between runs.

For timings I run minigzip through perf and parse its output with only the cputime (not time elapsed) counter enabled.

Mostly what seems to cause some variation in benchmark runs is cache (other things on the system might eat up a lot of L3 cache for a short time), possibly memory alignment, and syscalls. Therefore I filter out the X worst results for each level. Currently I run 11 tests per level and filter out the 5 worst.

To avoid fork() and such taking up a lot of the time, I currently use a 40MiB file for testing.
Today I have been running every PR with both cmake and configure, each taking ~5-6min including the perf-testing.

mtl1979:
I would use multiple test files of different size when testing AVX2 as it heavily depends on file size due to short time spending inside slide_*.

Dead2:
@mtl1979 Does that mean you are done and this is ready for merging?

mtl1979:
@Dead2 If I do something, it makes more sense to do separate pull request...

gdevenyi:
>  I currently use a 40MiB file for testing.

Please take a look at https://github.com/zlib-ng/zlib-ng/issues/326#issuecomment-578414906 for some corner-case file types which might be worth adding to a benchmarking suite.

Dead2:
@gdevenyi My 40MiB file is actually a tar file containing multiple file types (testing small files separately leads to inaccuracy due to fork() and syscalls having a much bigger impact), but adding some new files to one of my testfiles would be very much interesting.  :)

gdevenyi:
@Dead2 ah cheers, that makes some sense. In the linked case the important thing is the "padded with zeros" files, which we expect act pretty differently than regular files, and hence are a corner case in benchmarking.

Dead2:
For some reason, codecov only reports a 4% coverage on the codechanges from this patch.

Dead2:
Have a look here: https://codecov.io/gh/zlib-ng/zlib-ng/commit/53f9257a24362386e75504836eaeb2788360b1e8

It looks like cpuid gets run, but the check in `fill_window_sse` is always false so `slide_hash_sse2` gets used instead.

nmoinvaz:
@Dead2 are you saying we need a CI configuration with SSE2 turned off?

Dead2:
No, but that is a good point, we probably need to be able to disable AVX2 so SSE2 can be tested..

But currently it seems like AVX2 detection fails in the CI setup, it would be very weird if github actions didn't support AVX2 since that was introduced with Haswell years ago and I don't think there would be any reason to disable it.. Strange to say the least.

mtl1979:
cpuid fails on WSL too. /proc/cpuinfo was fixed in WSL some time ago.

nmoinvaz:
I am also seeing AVX2 disabled in GitHub Actions. 

When doing CPU ID check for 7, it returns all zeros. 

```
4: cpu 0 - ebx 756e6547, ecx 6c65746e, edx 49656e69 - max 13
4: cpu 1 - eax 000306a9, ebx 03020800, ecx ffba222b, edx 1f8bfbff - max 13
4: cpu 7 - eax 00000000, ebx 00000000, ecx 00000000, edx 00000000
```

I searched for other code that does AVX2 detection, and it seems it uses 0 for subfunction (2nd parameter).

image

In this instance I am able to get back values:

```
5: cpu 0 - ebx 756e6547, ecx 6c65746e, edx 49656e69 - max 13
5: cpu 1 - eax 000306a9, ebx 03020800, ecx ffba222b, edx 1f8bfbff - max 13
5: cpu 7 - eax 00000000, ebx 00000283, ecx 00000000, edx bc000400
```

mtl1979:
@nmoinvaz Fixed in #528

PR #498, opened 2020-01-10
Disable alignment sanitizer when architecture supports unaligned reads.

nmoinvaz:
See #496. This change also allows you to specify `-DUNALIGNED_OK=ON` manually through the command line if you wish to force it on.

PR #503, opened 2020-01-26
Support unaligned access in put_short

nmoinvaz:
This seems to have been changed in 7a4df36aad308cc50b60bf64d0a8acc036ed8b08. When put together with #498 it will no longer cause any UBsan warnings for platforms that support unaligned access.

PR #504, opened 2020-01-26
Improvements to pending buffer functions

nmoinvaz:
- Moved put_short_msb to deflate.h along with the rest of the put functions and renamed it to be consistent with other put function naming.
- Added put_uint32 and put_uint32_msb.
- Replaced instances of put_byte with put_short or put_uint32.

This sanitizer checks should pass once  #498 is merged.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -115,7 +115,6 @@ ZLIB_INTERNAL block_state deflate_slow         (deflate_state *s, int flush);
 static block_state deflate_rle   (deflate_state *s, int flush);
 static block_state deflate_huff  (deflate_state *s, int flush);
 static void lm_init              (deflate_state *s);
-static void putShortMSB          (deflate_state *s, uint16_t b);
 ZLIB_INTERNAL unsigned read_buf  (PREFIX3(stream) *strm, unsigned char *buf, unsigned size);
 
 extern void crc_reset(deflate_state *const s);
@@ -745,16 +744,6 @@ unsigned long ZEXPORT PREFIX(deflateBound)(PREFIX3(stream) *strm, unsigned long 
 }
 
 /* =========================================================================
- * Put a short in the pending buffer. The 16-bit value is put in MSB order.
- * IN assertion: the stream state is correct and there is enough room in
- * pending_buf.
- */
-static void putShortMSB(deflate_state *s, uint16_t b) {
-    put_byte(s, (unsigned char)(b >> 8));
-    put_byte(s, (unsigned char)(b & 0xff));
-}
-
-/* =========================================================================
  * Flush as much pending output as possible. All deflate() output, except for
  * some deflate_stored() output, goes through this function so some
  * applications may wish to modify it to avoid allocating a large
@@ -859,12 +848,11 @@ int ZEXPORT PREFIX(deflate)(PREFIX3(stream) *strm, int flush) {
         if (s->strstart != 0) header |= PRESET_DICT;
         header += 31 - (header % 31);
 
-        putShortMSB(s, header);
+        put_short_msb(s, header);
 
         /* Save the adler32 of the preset dictionary: */
         if (s->strstart != 0) {
-            putShortMSB(s, (uint16_t)(strm->adler >> 16));
-            putShortMSB(s, (uint16_t)(strm->adler));
+            put_uint32_msb(s, strm->adler);
         }
         strm->adler = functable.adler32(0L, NULL, 0);
         s->status = BUSY_STATE;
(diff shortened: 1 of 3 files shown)


deflate.c
```diff
@@ -859,12 +848,12 @@ int ZEXPORT PREFIX(deflate)(PREFIX3(stream) *strm, int flush) {
...
 
-        putShortMSB(s, header);
+        put_short_msb(s, header);
 
         /* Save the adler32 of the preset dictionary: */
         if (s->strstart != 0) {
-            putShortMSB(s, (uint16_t)(strm->adler >> 16));
-            putShortMSB(s, (uint16_t)(strm->adler));
+            put_short_msb(s, (uint16_t)(strm->adler >> 16));
+            put_short_msb(s, (uint16_t)(strm->adler));
```
mtl1979: This can be combined.

deflate.h
```diff
@@ -314,6 +316,45 @@ static inline void put_short(deflate_state *s, uint16_t w) {
...
+ * IN assertion: there is enough room in pendingBuf.
+ */
+static inline void put_word(deflate_state *s, uint32_t w) {
+#if BYTE_ORDER == BIG_ENDIAN
+    w = ZSWAP32(w);
+#endif
+#if defined(UNALIGNED_OK)
+    *(uint32_t *)(&s->pending_buf[s->pending]) = w;
+#else
+    memcpy(&(s->pending_buf[s->pending]), &w, sizeof(uint32_t));
```
mtl1979: Second parameter is missing explicit cast.

nmoinvaz:
I have modified this PR for better aligned access support and also merged in #503.

nmoinvaz:
@mtl1979 I have made the changes you suggested.

mtl1979:
Big-endian builds seem to fail on CI...

deflate.h
```diff
@@ -300,18 +300,66 @@ typedef enum {
...
+    put_byte(s, ((dw >> 24) & 0xff));
+    put_byte(s, ((dw >> 16) & 0xff));
+    put_byte(s, ((dw >> 8) & 0xff));
+    put_byte(s, (dw & 0xff));
+#else
+    put_byte(s, (dw & 0xff));
+    put_byte(s, ((dw >> 8) & 0xff));
+    put_byte(s, ((dw >> 16) & 0xff));
+    put_byte(s, ((dw >> 24) & 0xff));
+#endif
```
mtl1979, line 343: Shifting already does explicit byte swap, so the last two cases can be merged.

deflate.h
```diff
@@ -300,18 +300,66 @@ typedef enum {
...
+#if defined(UNALIGNED_OK)
+    *(uint16_t *)(&s->pending_buf[s->pending]) = w;
+    s->pending += 2;
+#elif BYTE_ORDER == BIG_ENDIAN
+    put_byte(s, ((w >> 8) & 0xff));
+    put_byte(s, (w & 0xff));
+#else
+    put_byte(s, (w & 0xff));
+    put_byte(s, ((w >> 8) & 0xff));
+#endif
```
mtl1979, line 318: Shifting already does explicit byte swap, so last two cases can be merged.

deflate.h
```diff
@@ -300,18 +300,66 @@ typedef enum {
...
+ * Output a int MSB first on the stream.
+ * IN assertion: there is enough room in pending_buf.
+ */
+static inline void put_uint32_msb(deflate_state *s, uint32_t dw) {
+#if BYTE_ORDER != BIG_ENDIAN
+    dw = ZSWAP32(dw);
 #endif
-  memcpy(&(s->pending_buf[s->pending]), &w, sizeof(uint16_t));
-  s->pending += 2;
+    put_uint32(s, dw);
```
mtl1979: There is case when this might fail as put_uint32() on big-endian machine does explicit byte swap due to shifting even though it is not intended... This means put_uint32() should only be called here on little-endian machines when UNALIGNED_OK is also defined, and when UNALIGNED_OK is unset explicitly use put_byte() with reversed byte order.

mtl1979 (changes requested):
ZSWAP32() should not be used if following function uses put_byte() internally. This happens when UNALIGNED_OK is not set. Sequence of multiple put_byte() calls does write in same order on little- and big-endian machines, so only thing that matters is the shift amount when choosing LSB or MSB order.

deflate.c
nmoinvaz: Fixed.

deflate.h
nmoinvaz: I think this is from an older revision. I have removed the `memcpy` in newer revision.

deflate.h
nmoinvaz: I learn something new every day. Thank you.

deflate.h
nmoinvaz: I have revised the put functions.

mtl1979:
Big-endian builds seem to be fixed now... After @Dead2 has had time to review, I guess this is ready for merging...

Dead2:
This looks really good.

One nitpick though comments for put_uint32 and put_uint32_msb says:
"Output a int LSB first on the stream."
Since int does not always mean 32bit int (even though it pretty much always is nowadays), I wonder if it should say "Output a 32bit int..", what do you think? As I said, a minor nitpick :)

I'll try to get my tests running again this week so I can test this and a few of the other PRs.

mtl1979:
@Dead2 "32-bit" would be more correct... But I agree that ambiguous terms should be corrected...

nmoinvaz:
I have fixed the comment.

Dead2:
Needs rebase

nmoinvaz:
Rebase done.

Dead2:
configure: ~0.6% faster
cmake: ~0.3% faster

PR #505, opened 2020-01-27
Change insert_string abstraction to update_hash

nmoinvaz:
- Standardize insert_string and abstract only the hashing functionality across architectures. 
- Added unaligned conditionally compiled code for insert_string and quick_insert_string. 
- Unify sse42 crc32 assembly between insert_string and quick_insert_string. #451

This change follows what was discussed in #499. This sanitizer checks should pass once  #498 is merged.


deflate_p.h
```diff
@@ -27,21 +29,39 @@ void flush_pending(PREFIX3(stream) *strm);
...
+#if defined(UNALIGNED_OK)
+        val = *(unsigned int *)(&s->window[str_idx]);
+#else
+        memcpy(&val, &s->window[str_idx], sizeof(val));
+#endif
+
+        if (s->level >= TRIGGER_LEVEL)
+            val &= 0xFFFFFF;
+
+        s->ins_h = functable.update_hash(s, s->ins_h, val);
```
mtl1979: It might be faster to store function pointer from ```functable.update_hash``` in local variable before the loop, as then compiler can optimize it as simple jump using register. Some compilers might assume global struct members are volatile by default and as such require indirect jump.

mtl1979 (changes requested):

deflate_p.h
nmoinvaz: Fixed

nmoinvaz:
I'm not quite sure why some of the architectures are failing.

nmoinvaz:
Seems to only happen on big-endian machines.

mtl1979:
All big-endian machines should have UNALIGNED_OK unset and avoid anything larger than char with memcpy. This is because direct assign might unintentionally do byte swap, and as optimized memcpy acts like macro as in it does not do implicit cast, so anything wider than 8 bits might trigger CPU exception which is not fatal inside virtual machine as lowest bit(s) are internally used as bitmask.

nmoinvaz:
It looks like I have got it working now. I wonder now if it is worth all the trouble.

mtl1979:
I'd like to think all the recent changes have made it easier to add AltiVec support and possibly support for other vector instruction sets with minimal duplication.

Dead2:
Rebase needed.

functable.h
```diff
@@ -8,12 +8,18 @@
...
-    void     (* fill_window)    (deflate_state *s);
-    Pos      (* insert_string)  (deflate_state *const s, const Pos str, unsigned int count);
-    uint32_t (* adler32)        (uint32_t adler, const unsigned char *buf, size_t len);
-    uint32_t (* crc32)          (uint32_t crc, const unsigned char *buf, uint64_t len);
-    void     (* slide_hash)     (deflate_state *s);
+    fill_window_cb  fill_window;
+    update_hash_cb  update_hash;
+    adler32_cb      adler32;
+    crc32_cb        crc32;
+    slide_hash_cb   slide_hash;
```
Dead2, line 22: This should be a separate PR I think, I don't see any added value from typedef'ing this, it just moves the complexity a few lines and adds a new reference to that complexity. Nobody should be touching this struct unless they also want to touch that complexity, so I don't see the point of this honestly. Feel free to open a new PR with this where you can state the benefit and we can discuss it separately. :)

Dead2:
I like the idea in this PR, good job :)

functable.h
nmoinvaz: The `update_hash_cb` typedef is used in `insert_string` so there is no indirect reference in the loop.

Dead2: Ah, there it is, I completely missed that. :)

nmoinvaz:
Rebased again. Debug is failing due to existing issue mentioned in #471.

Dead2:
Hmm, this looks like it has a slight performance regression of ~0.1%.

I am thinking it is because where we previously took an indirect call penalty for each insert_string but insert_string had UPDATE_HASH inlined, we now take the penalty of update_hash not being inlined multiple times in the for loop in insert_string.

The cleanup is real though. So is there anything that can be done better to avoid or mitigate that penalty?

nmoinvaz:
@Dead2 we can pass in the update_hash_cb into insert_string. What do you think of that?

mtl1979:
@Dead2 I suggested we take it completely out of functable and use inlined functions that are stored in arch-specific header files... even if we need to do run-time check on every call, it might still be faster than going through functable.

Dead2:
@nmoinvaz I don't think that would do anything unfortunately.

Update_hash gets called a lot in insert_string, that is why it made sense to make several insert_string variants with update_hash inlined. If we inlined the runtime checks for update_hash into insert_string we would be doing an if and (usually) a jump for each iteration of the for loop, and I am not convinced that is any better than doing an unconditional jump to a slightly more remote location.

What I suggest is replace keep using insert_string through functable, but unify the insert_string implementations by using a couple of macros for the common (most of it) code and just have them differ in what update_hash they inline. That should give us a cleanup of insert_string_* while preserving the performance.


```C
// Pseudocode:
void insert_string_sse2(foo, bar){
TOP_PART()
update_hash_sse2(baz)
BOTTOM_PART()
}
```

The macros themselves will be a bit more messy though, but this should make it really easy to maintain multiple insert_string variants for different architectures.
This is where modern C++ with its template functions would shine, too bad we can't use any of that :)

nmoinvaz:
I see what you mean. Do you think it is worth it for the 0.1%? It would make it harder to debug the code.

Dead2:
I am a bit worried that this might have a bigger perf impact on other cpus, skylake like I am currently using is pretty good at reducing this kind of penalty compared to older cpus and possible compared to other architectures. Would be nice to get that tested at least.

functable.h
nmoinvaz: Actually it might not be necessary, because indirect lookup doesn't occur. Since it is not a pointer, there only a struct field reference. _I think_.

Dead2:
There is one trick that can be used to make debugging easier and keep the templating, it is kind of dirty but more maintainable than the above.

There are two closely related variants:
1. Have a header file with the insert_string function template, the function name suffix and the update_hash function call suffix is passed in by define from the .c file that includes it. The .c file then needs to include the header file several times with each of the different suffixes defined. This makes the .c file a little messy, but the insert_string_template.h would be very clean.

2. Use sed or the C preprocessor to convert insert_string_template.in into insert_string_sse2.c etc before compiling. This is probably the most elegant way, but I am not sure how this could work on windows platforms..?

mtl1979:
@Dead2 variant 1 might be doable using multiple small .c files, basically one generic and one for each arch with optimised version.

nmoinvaz:
I agree option 1 might be doable. One of the reasons why I broke it out the way I did with _update_hash_ is because `quick_insert_string` uses the hash function too so we need to account for that somehow.

nmoinvaz:
Closing in favor of #522.

PR #506, opened 2020-01-28
Support aligned alloc/free functions for Windows

nmoinvaz:
I have also moved the malloc.h include after `MY_ZCALLOC` incase it is defined then the include is not necessary. This essentially allows you to compile with `-DUNALIGNED_OK=OFF` on Windows.

### DIFF
--- a/zutil.c
+++ b/zutil.c
@@ -10,9 +10,6 @@
 #ifdef WITH_GZFILEOP
 #  include "gzguts.h"
 #endif
-#ifndef UNALIGNED_OK
-#  include <malloc.h>
-#endif
 
 const char * const zng_errmsg[10] = {
     (const char *)"need dictionary",     /* Z_NEED_DICT       2  */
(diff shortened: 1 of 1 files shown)


Dead2:
Rebase needed

Dead2:
Looks good to me, but I'd like to see another review of this in case I missed something.

mtl1979:
My compiler doesn't define ```WINDOWS```, but does define ```_WIN32```...

nmoinvaz:
`WINDOWS` is defined in zconf.h

Dead2:
Hmm, WIN32 could potentially be set without WINDOWS being set, does that ever happen in real life?
```c
#if defined(_WINDOWS) && !defined(WINDOWS)
#  define WINDOWS
#endif
#if defined(_WIN32) || defined(__WIN32__)
#  ifndef WIN32
#    define WIN32
#  endif
#endif
```

nmoinvaz:
Yeah I don't know if that is necessary anymore.

nmoinvaz:
Rebase done and changed to `_WIN32`.

Dead2:
@mtl1979 Does this look good to you now?

mtl1979:
I'm not really fan of changing ```#ifndef``` to ```#if !defined``` when there is only one condition... It makes the patch longer than necessary and essentially mixes style changes with code changes.

nmoinvaz:
Fixed.

PR #507, opened 2020-01-28
Fix double specification of -O2 in ./configure

gdevenyi:
./configure and cmake don't generate the same kind of builds right now, update ./configure to closer match.

### DIFF
--- a/configure
+++ b/configure
@@ -332,7 +332,7 @@ if test "$gcc" -eq 1 && ($cc $CFLAGS -c $test.c) >> configure.log 2>&1; then
     LDFLAGS="${LDFLAGS} ${ARCHS}"
   fi
   CFLAGS="${CFLAGS} -Wall"
-  SFLAGS="-O2 ${CFLAGS} -fPIC"
+  SFLAGS="${CFLAGS} -fPIC"
   if test $native -eq 1; then
     CFLAGS="${CFLAGS} -march=native"
     SFLAGS="${SFLAGS} -march=native"


gdevenyi:
Ref: https://github.com/zlib-ng/zlib-ng/issues/326#issuecomment-578414906

gdevenyi:
One fuzzer test fails here but it looks like this is unrelated as the fuzzer runs against a cmake build.

Dead2:
Thanks for spotting and fixing this. Will pull once I have verified it myself :)

The failing test is safe to ignore, we luckily have a fix for that in another PR.

Dead2:
The default cmake cflags on my platform is as follows:
`CMAKE_C_FLAGS_RELEASE:STRING=-O2 -DNDEBUG`
So I am wondering whether -O2/-O3 is distro-specific or perhaps version specific. I am on Fedora 30 with cmake 3.14.5. So for me this would actually make the two differ instead of making them the same.

I do see ~1.5% speedup on compression with -O3 though, so it might be worth changing our default to that. Decompression does seem to become ~1.3% slower though (similar percentage, although in actual time it is nowhere close to the same).

So, I'd prefer if this PR only fixed the double -O2 for now, and that a new PR (or issue) is opened regarding changing to -O3 as default for both cmake and configure.

gdevenyi:
Oh wow, that's really weird.

I'm using ``cmake version 3.14.5`` from https://apt.kitware.com/ on Ubuntu 14.04.5 and gcc-9.2.1 and I definitely get "-O3"

cmake seems to be completely opaque in its documentation on what the defaults are: https://cmake.org/cmake/help/latest/variable/CMAKE_CONFIGURATION_TYPES.html#variable:CMAKE_CONFIGURATION_TYPES

I will update

gdevenyi:
This setting appears to be older than 4 years on cmake:
https://github.com/Kitware/CMake/blame/master/Modules/Compiler/GNU.cmake

"-O3" is the default for release, I'm not sure how you're getting otherwise...

gdevenyi:
Its fedora's fault: https://src.fedoraproject.org/rpms/cmake/blob/master/f/cmake.spec#_96

gdevenyi:
I currently have removed the O2 to O3 change. Please let me know if you want me to re-add.

Sounds like we may need some benchmarking across -O levels?

Dead2:
Yeah, I suspected as much.
If we start defaulting to -O3 in configure, we should probably set the same default for cmake to avoid there being a big difference in performance between the two.

Although I just tested and our all-defaults cmake build is ~2.3% slower than the all-defaults configure build on my system. I checked the logs for a couple of files, but found no significant difference in compiler parameters for those. The only difference I know of is that minigzip is compiled static by configure and shared by cmake.
Also, minigzipsh made by configure for some reason is unable to find the libz-ng.so.1 without LD_PRELOAD, making testing much more awkward.

Issue #496, opened 2020-01-07
longest_match questions


nmoinvaz:
I have been looking at the longest_match function. 

1. It doesn't appear that the `UNALIGNED_OK` is necessary. std2_longest_match and std3_longest_match don't do any unaligned access.
1. Why is longest_match broken into 3 separate functions? It seems like it could be a single function, with #ifdef inside for the separate comparison lengths (1 at a time, 2 at a time, or 4/8 at a time).

mtl1979:
A lot of the code has been rewritten to avoid unaligned accesses even though x86 processors can handle them, because some compilers are not designed to emit correct instructions when some optimizations are enabled. I haven't tried to isolate each of the functions as separate files so I could compare the function contents line by line, but I suspect there is small differences in other parts of the functions than just different compare chunk size.

```UNALIGNED_OK``` can be used to both check if processor can handle unaligned loads and stores and if processor is likely little- or big-endian, as most big-endian processors don't allow unaligned accesses.

nmoinvaz:
I was not able to find anywhere where there was unaligned access so maybe `UNALIGNED_OK` is not necessary. 

It does seem like the only difference is the compare chunk size. The difference used to be that one was for `UNALIGNED_OK` access while others were not. I think while trying to fix ASAN errors, everything might have been changed to be aligned.

https://github.com/zlib-ng/zlib-ng/commit/72214719ca8db882ab49d31872528b8b456362aa
https://github.com/zlib-ng/zlib-ng/commit/bee695a6415bbf37087be132a3c40e2e761658c0

nmoinvaz:
Perhaps if `UNALIGNED_OK` is defined in CMakeLists.txt, it should not add `-fsanitize=alignment`.

mtl1979:
Maybe there should be pull request to cleanup/merge the longest_match variants... Would help maintaining the code as only one version needs to be checked for improvements/issues.

nmoinvaz:
I have abstracted out the methods of comparing 258 bytes and have done a performance testing on them:

https://gist.github.com/nmoinvaz/b70fd103c5bed104dfbcf09d5d927cfe

PR #508, opened 2020-01-29
Support unaligned reads in longest_match variants.

nmoinvaz:
See #496. At some point `longest_match` variants will still need to be cleaned up, but this should add back unaligned reads to `longest_match` for the short term.

This commit adds new macros that can be used for other areas where we have unaligned read supported, however I have not touched those due to pending pull requests.

### DIFF
--- a/match_p.h
+++ b/match_p.h
@@ -11,7 +11,7 @@
 #include "zbuild.h"
 #include "deflate.h"
 
-#if (defined(UNALIGNED_OK) && MAX_MATCH == 258)
+#if (MAX_MATCH == 258)
 
    /* ARM 32-bit clang/gcc builds perform better, on average, with std2. Both gcc and clang and define __GNUC__. */
 #  if defined(__GNUC__) && defined(__arm__) && !defined(__aarch64__)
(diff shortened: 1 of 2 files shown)


mtl1979:
ZCOPY32 seems unused. ZCOPY32 and ZCOPYULONG are essentially duplicates on all platforms that run Windows or are 32-bit.

nmoinvaz:
I plan on using `ZCOPY32` in #504 if it is merged.

mtl1979:
The problem is that memcpy is not guaranteed to do unaligned-safe copy when both size of parameters and copy length is known at compile-time. So this is adding unnecessary complexity in the code. Most analysis tools can't detect unaligned access inside memcpy and other functions that are not part of current compile unit.

mtl1979:
When memcpy is followed by compare, those should be combined. When unaligned access is allowed, both pointers should be cast to type exactly the size of copy length. When unaligned access is not allowed, operands should be compared byte by byte and false returned on first unequal byte. Loop can be unrolled as length is always constant.

nmoinvaz:
I see what you mean. Perhaps the changes made in bee695a6415bbf37087be132a3c40e2e761658c0 and 72214719ca8db882ab49d31872528b8b456362aa should be reverted - since those versions of `longest_match` were never intended for unaligned architectures..

mtl1979:
@nmoinvaz There is quite a few commits that @Dead2 merged in that should not have merged in in my opinion, because they are basically workarounds for border cases that make normal code a lot slower and don't actually fix anything but instead mask the real issues.

Usually reinventing standard C library functions is frowned upon, but in this case those standard library functions are broken and sub-optimal at the best case. 

It is common knowledge that 8-bit reads and writes are slow, but memcmp does always access 8 bits only, so there is no possible gain in speed. What is saved is actually first part of the optimization where it checks the alignment and tries to skip as many equal chunks as possible, but in this case the length is smaller than the chunk size so the first part of optimized memcmp is redundant.

nmoinvaz:
I will make another PR that reverts those changes and adds a comment in those places in case people try to make the same change again.

mtl1979:
@nmoinvaz Adding a comment is a very good idea as not many contributors are as familiar with compiler bugs and limitations as I am... I spent 6 months helping Linaro team fix issues in their ARM version of gcc...

PR #512, opened 2020-02-02
Fixed compiler warning when using BITS macro.

nmoinvaz:
This is the same cast that occurs in `NEEDBITS` and `DROPBITS`. I have seen warnings related to this.


Dead2:
Looks good to me.

PR #513, opened 2020-02-02
Change deflate_state's bi_buf to 32-bit

nmoinvaz:
This PR requires #496 and #504 which are the first two commits. In the third commit I changed `deflate_state`'s `bi_buf` to 32-bit and then in the fourth commit I replaced `quick_send_bits` with `send_bits` which is 32-bit now. All the code can benefit from 32-bit bit buffer now and the code path for all the deflate methods is the same.

### DIFF
--- a/arch/x86/deflate_quick.c
+++ b/arch/x86/deflate_quick.c
@@ -135,6 +135,9 @@ static inline void static_emit_ptr(deflate_state *const s, const int lc, const u
 
     send_bits(s, code1, len1, bi_buf, bi_valid);
     send_bits(s, code2, len2, bi_buf, bi_valid);
+#ifdef ZLIB_DEBUG
+    s->compressed_len += len1 + len2;
+#endif
 
     s->bi_valid = bi_valid;
     s->bi_buf = bi_buf;
@@ -147,6 +150,9 @@ static inline void static_emit_lit(deflate_state *const s, const int lit) {
     uint32_t bi_buf = s->bi_buf;
 
     send_bits(s, static_ltree[lit].Code, static_ltree[lit].Len, bi_buf, bi_valid);
+#ifdef ZLIB_DEBUG
+    s->compressed_len += static_ltree[lit].Len;
+#endif
 
     s->bi_valid = bi_valid;
     s->bi_buf = bi_buf;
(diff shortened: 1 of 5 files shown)


Dead2:
Postponing review of this one until the two preceding ones are in.
Looking forward to it though :)

Dead2:
Needs rebase

nmoinvaz:
I thought it might. Rebase done.

nmoinvaz:
It appears to be failing on debug build. I will have to take a look another time.

Dead2:
This seems to slow down deflate_quick slightly, but the rest get a nice speedup of about 0.5%.
Inflate remains unchanged.

nmoinvaz:
That is what I expected.

Dead2:
deflate_quick is about 1-3% slower, because it is so quick the margins of error are also quite big.
Is there anything that can be done to mitigate this somewhat? Otherwise I won't sweat that since deflate_quick is already so much faster than anything else here and it is a very nice code cleanup.

nmoinvaz:
Perhaps the only areas where I could see would be possibly combining the two calls to `send_bits` in `static_emit_ptr` (I don't know if it is possible). Or to try reorg the if statements in `send_bits` so that the first (less likely) condition is second.

Dead2:
The if reorg looks to me like a possible improvement.

nmoinvaz:
@Dead2 I took another look. Should be better. Also fixed the debug build issue (which I think has been in deflate_quick for some time).

Dead2:
On top of all of yesterdays changes, this now improves performance by:
cmake: 0.25-0.30%
configure: 0.92-1.02% (very surprising, but probably because the compiler is able to do some extra optimizations when compiling static)

inflate seems to stay mostly the same, with configure possibly a little slower than before..

(with this, configure-deflate is 1.54% faster than two days ago)

BTW: Making cmake link minigzip statically brings it in line with configures minigzip.
```patch
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -906,7 +906,7 @@ option(ZLIB_ENABLE_TESTS "Build test binaries" ON)
 if (ZLIB_ENABLE_TESTS)
     enable_testing()
     macro(configure_test_executable target)
-        target_link_libraries(${target} zlib)
+        target_link_libraries(${target} zlibstatic)
         target_include_directories(${target} PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
         if(NOT WITH_GZFILEOP)
             target_compile_definitions(${target} PUBLIC -DWITH_GZFILEOP)
```
The above makes minigzip and all tests link statically with zlib-ng. Maybe it is worth making (only) minigzip static by default (but add an option to disable it), it's not like zlib-ng is huge, dunno. :)
(I do want to keep the tests using shared linking though, so we actually test the shared library too.

Issue #516, opened 2020-02-07
Uniformize cmake vs configure build configuration


gdevenyi:
Following up on #507 and #515, @Dead2 noted that configure vs cmake on his system has a ~2% difference in performance.

We also discovered that cmake on Fedora is patched to be different than upstream cmake, which -O2 instead of -O3 for the "Release" configuration.

In order to avoid these inconsistencies, after #515 the cmake and configure settings should be uniformized to what zlib-ng devs consider the best choice by default, rather than relying on the platform.

nmoinvaz:
I think this should be resolved now. CMake now uses -O2. @dankegel do know anything we are missing? I think they should be building the same with pkgcheck.sh correct?

dankegel:
I know of no further problems (though I haven't really looked).  pkgcheck.sh is now hands off with respect to cflags.  OK to close as fixed IMHO.

PR #521, opened 2020-02-08
Change minigzip to compile against zlibstatic.

nmoinvaz:
This PR is done so that cmake output binaries match configure output binaries. It will help fix the gap we are seeing in performance testing between the two build systems. This should help address #516.

For CI, we have to keep everything statically linked due to issues with `qemu` powerpc configurations. I have previously commented in cmake.yml. That and also possible issues with code coverage results is why `-DBUILD_SHARED_LIBS=OFF` is explicitly stated in CI. _I believe_ that shared libraries are still tested in CI through configure script. 

When not using cmake for CI (or when `BUILD_SHARED_LIBS` is not defined) then `minigzip` and `minigzip64` will link statically against `zlibstatic`, in this instance all of the tests still link dynamically.


Dead2 (approved):
Looks good to me.

dankegel:
Is this the bug ci works around with static linking?
https://bugs.launchpad.net/bugs/1701798

nmoinvaz:
It could be. I know in the yaml I had written:
> Shared libaries turned off for qemu ppc* and sparc..

I had experienced the same problem again not too long ago with powerpc and qemu.

PR #522, opened 2020-02-08
Standardize insert_string functionality across architectures.

nmoinvaz:
This PR is a continuation of the discussion from #505.

* Standardize insert_string functionality across architectures.
* Added unaligned conditionally compiled code for insert_string and quick_insert_string.
* Unify sse42 crc32 assembly between insert_string and quick_insert_string. #451
* Modified quick_insert_string to work across architectures.

### DIFF
--- a/arch/arm/insert_string_acle.c
+++ b/arch/arm/insert_string_acle.c
@@ -10,44 +10,11 @@
 #include "../../zbuild.h"
 #include "../../deflate.h"
 
-/* ===========================================================================
- * Insert string str in the dictionary and set match_head to the previous head
- * of the hash chain (the most recent string with same hash key). Return
- * the previous length of the hash chain.
- * IN  assertion: all calls to to INSERT_STRING are made with consecutive
- *    input characters and the first MIN_MATCH bytes of str are valid
- *    (except for the last MIN_MATCH-1 bytes of the input file).
- */
-Pos insert_string_acle(deflate_state *const s, const Pos str, unsigned int count) {
-    Pos p, lp, ret;
+#define UPDATE_HASH(s, h, val) \
+    h = __crc32w(0, val)
 
-    if (UNLIKELY(count == 0)) {
-        return s->prev[str & s->w_mask];
-    }
+#define INSERT_STRING       insert_string_acle
+#define QUICK_INSERT_STRING quick_insert_string_acle
 
-    ret = 0;
-    lp = str + count - 1; /* last position */
-
-    for (p = str; p <= lp; p++) {
-        uint32_t val, h, hm;
-        memcpy(&val, &s->window[p], sizeof(val));
-
-        if (s->level >= TRIGGER_LEVEL)
-            val &= 0xFFFFFF;
-
-        h = __crc32w(0, val);
-        hm = h & s->hash_mask;
-
-        Pos head = s->head[hm];
-        if (head != p) {
-            s->prev[p & s->w_mask] = head;
-            s->head[hm] = p;
-            if (p == lp)
-                ret = head;
-        } else if (p == lp) {
-            ret = p;
-        }
-    }
-    return ret;
-}
+#include "../../insert_string_tpl.h"
 #endif
--- a/arch/x86/deflate_quick.c
+++ b/arch/x86/deflate_quick.c
@@ -23,6 +23,7 @@
 #  include <nmmintrin.h>
 #endif
 #include "../../deflate.h"
+#include "../../functable.h"
 #include "../../memcopy.h"
 
 #ifdef ZLIB_DEBUG
@@ -31,7 +32,6 @@
 
 extern void fill_window_sse(deflate_state *s);
 extern void flush_pending(PREFIX3(stream) *strm);
-extern Pos  quick_insert_string_sse4(deflate_state *const s, const Pos str);
 
 static inline long compare258(const unsigned char *const src0, const unsigned char *const src1) {
 #ifdef _MSC_VER
@@ -219,7 +219,7 @@ ZLIB_INTERNAL block_state deflate_quick(deflate_state *s, int flush) {
         }
 
         if (s->lookahead >= MIN_MATCH) {
-            hash_head = quick_insert_string_sse4(s, s->strstart);
+            hash_head = functable.quick_insert_string(s, s->strstart);
             dist = s->strstart - hash_head;
 
             if (dist > 0 && (dist-1) < (wsize - 1)) {
--- a/arch/x86/deflate_quick.c
+++ b/arch/x86/deflate_quick.c
@@ -31,6 +31,7 @@
 
 extern void fill_window_sse(deflate_state *s);
 extern void flush_pending(PREFIX3(stream) *strm);
+extern Pos  quick_insert_string_sse4(deflate_state *const s, const Pos str);
 
 static inline long compare258(const unsigned char *const src0, const unsigned char *const src1) {
 #ifdef _MSC_VER
@@ -186,26 +187,6 @@ static void static_emit_end_block(deflate_state *const s, int last) {
     s->block_open = 0;
 }
 
-static inline Pos quick_insert_string(deflate_state *const s, const Pos str) {
-    Pos ret;
-    unsigned h = 0;
-
-#ifdef _MSC_VER
-    h = _mm_crc32_u32(h, *(unsigned *)(s->window + str));
-#else
-    __asm__ __volatile__ (
-        "crc32l (%[window], %[str], 1), %0\n\t"
-    : "+r" (h)
-    : [window] "r" (s->window),
-      [str] "r" ((uintptr_t)str)
-    );
-#endif
-
-    ret = s->head[h & s->hash_mask];
-    s->head[h & s->hash_mask] = str;
-    return ret;
-}
-
 ZLIB_INTERNAL block_state deflate_quick(deflate_state *s, int flush) {
     IPos hash_head;
     unsigned dist, match_len;
@@ -238,7 +219,7 @@ ZLIB_INTERNAL block_state deflate_quick(deflate_state *s, int flush) {
         }
 
         if (s->lookahead >= MIN_MATCH) {
-            hash_head = quick_insert_string(s, s->strstart);
+            hash_head = quick_insert_string_sse4(s, s->strstart);
             dist = s->strstart - hash_head;
 
             if (dist > 0 && (dist-1) < (wsize - 1)) {
--- a/win32/Makefile.a64
+++ b/win32/Makefile.a64
@@ -35,7 +35,8 @@ SUFFIX =
 
 OBJS = adler32.obj armfeature.obj compress.obj crc32.obj deflate.obj deflate_fast.obj deflate_slow.obj \
        deflate_medium.obj \
-       functable.obj infback.obj inflate.obj inftrees.obj inffast.obj trees.obj uncompr.obj zutil.obj fill_window_arm.obj
+       functable.obj infback.obj inflate.obj inftrees.obj inffast.obj insert_string.obj \
+       trees.obj uncompr.obj zutil.obj fill_window_arm.obj
 !if "$(WITH_GZFILEOP)" != ""
 WFLAGS = $(WFLAGS) -DWITH_GZFILEOP
 OBJS = $(OBJS) gzclose.obj gzlib.obj gzread.obj gzwrite.obj
--- a/win32/Makefile.arm
+++ b/win32/Makefile.arm
@@ -39,7 +39,8 @@ SUFFIX =
 
 OBJS = adler32.obj armfeature.obj compress.obj crc32.obj deflate.obj deflate_fast.obj deflate_slow.obj \
        deflate_medium.obj \
-       functable.obj infback.obj inflate.obj inftrees.obj inffast.obj trees.obj uncompr.obj zutil.obj fill_window_arm.obj
+       functable.obj infback.obj inflate.obj inftrees.obj inffast.obj insert_string.obj \
+       trees.obj uncompr.obj zutil.obj fill_window_arm.obj
 !if "$(WITH_GZFILEOP)" != ""
 WFLAGS = $(WFLAGS) -DWITH_GZFILEOP
 OBJS = $(OBJS) gzclose.obj gzlib.obj gzread.obj gzwrite.obj
--- a/win32/Makefile.msc
+++ b/win32/Makefile.msc
@@ -36,7 +36,8 @@ SUFFIX =
 
 OBJS = adler32.obj compress.obj crc32.obj deflate.obj deflate_fast.obj deflate_quick.obj deflate_slow.obj \
        deflate_medium.obj \
-       functable.obj infback.obj inflate.obj inftrees.obj inffast.obj slide_avx.obj slide_sse.obj trees.obj uncompr.obj zutil.obj \
+       functable.obj infback.obj inflate.obj inftrees.obj inffast.obj insert_string.obj \
+       slide_avx.obj slide_sse.obj trees.obj uncompr.obj zutil.obj \
        x86.obj fill_window_sse.obj insert_string_sse.obj crc_folding.obj
 !if "$(ZLIB_COMPAT)" != ""
 WITH_GZFILEOP = yes
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -770,6 +770,7 @@ set(ZLIB_PRIVATE_HDRS
     inflate.h
     inflate_p.h
     inftrees.h
+    insert_string_tpl.h
     match_p.h
     memcopy.h
     trees.h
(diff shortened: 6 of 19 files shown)


nmoinvaz:
I'm not sure why configure script fails, so if somebody can point me in the right direction. I think `insert_string.c` needs to be included, but it looks like it should do that automatically.

arch/x86/deflate_quick.c
```diff
@@ -234,7 +215,7 @@ ZLIB_INTERNAL block_state deflate_quick(deflate_state *s, int flush) {
         }
 
         if (s->lookahead >= MIN_MATCH) {
-            hash_head = quick_insert_string(s, s->strstart);
+            hash_head = functable.quick_insert_string(s, s->strstart);
```
Dead2, line 222: Since deflate_quick itself depends on SSE, I think we can safely use quick_insert_string_sse directly here, and avoid the functable indirection. At least until we get a better insert_string or deflate_quick gets ported to another arch.

nmoinvaz: Ok. I think the only thing preventing it from getting ported is the `compare258` function.

nmoinvaz:
Debug configuration may fail due to existing issue #471.

win32/Makefile.a64
```diff
@@ -35,7 +35,8 @@ SUFFIX =
 
 OBJS = adler32.obj armfeature.obj compress.obj crc32.obj deflate.obj deflate_fast.obj deflate_slow.obj \
        deflate_medium.obj \
-       functable.obj infback.obj inflate.obj inftrees.obj inffast.obj trees.obj uncompr.obj zutil.obj fill_window_arm.obj
+       functable.obj infback.obj inflate.obj inftrees.obj inffast.obj insert_string.obj \
+	   trees.obj uncompr.obj zutil.obj fill_window_arm.obj
```
mtl1979: Mismatch in indent...

win32/Makefile.arm
```diff
@@ -39,7 +39,8 @@ SUFFIX =
 
 OBJS = adler32.obj armfeature.obj compress.obj crc32.obj deflate.obj deflate_fast.obj deflate_slow.obj \
        deflate_medium.obj \
-       functable.obj infback.obj inflate.obj inftrees.obj inffast.obj trees.obj uncompr.obj zutil.obj fill_window_arm.obj
+       functable.obj infback.obj inflate.obj inftrees.obj inffast.obj insert_string.obj \
+	   trees.obj uncompr.obj zutil.obj fill_window_arm.obj
```
mtl1979: Mismatch in indent...

win32/Makefile.msc
```diff
@@ -36,7 +36,8 @@ SUFFIX =
 
 OBJS = adler32.obj compress.obj crc32.obj deflate.obj deflate_fast.obj deflate_quick.obj deflate_slow.obj \
        deflate_medium.obj \
-       functable.obj infback.obj inflate.obj inftrees.obj inffast.obj slide_avx.obj slide_sse.obj trees.obj uncompr.obj zutil.obj \
+       functable.obj infback.obj inflate.obj inftrees.obj inffast.obj insert_string.obj \
+	   slide_avx.obj slide_sse.obj trees.obj uncompr.obj zutil.obj \
```
mtl1979: Mismatch in indent...

insert_string.h
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+/* ===========================================================================
+ * Insert string str in the dictionary and set match_head to the previous head
+ * of the hash chain (the most recent string with same hash key). Return
+ * the previous length of the hash chain.
+ * IN  assertion: all calls to to INSERT_STRING are made with consecutive
+ *    input characters and the first MIN_MATCH bytes of str are valid
+ *    (except for the last MIN_MATCH-1 bytes of the input file).
+ */
```
mtl1979: This comment block is in wrong place, does not apply to QUICK_INSERT_STRING as it requires MIN_MATCH+1 bytes (4 bytes) to be valid, not MIN_MATCH bytes (3 bytes).

arch/arm/insert_string_acle.c
```diff
@@ -10,44 +10,11 @@
...
-                ret = head;
-        } else if (p == lp) {
-            ret = p;
-        }
-    }
-    return ret;
-}
-#endif
+#include "../../insert_string.h"
+#endif
```
mtl1979: Missing empty line at end of file.

mtl1979 (changes requested):
Some minor issues ;)

win32/Makefile.a64
nmoinvaz: Good catch thanks!

win32/Makefile.arm
nmoinvaz: Good catch thanks!

arch/arm/insert_string_acle.c
nmoinvaz: Good catch thanks!

win32/Makefile.msc
nmoinvaz: Good catch thanks!

insert_string.h
nmoinvaz: Ok. I have moved it above insert_string.

nmoinvaz:
Changes made.

nmoinvaz:
@Dead2 perhaps `fill_window` implementations can be cleaned up the same way? I noticed in Git log you have experience with `fill_window` code, what do you think?

mtl1979:
@nmoinvaz AFAIR ```fill_window``` code was separated due to compiler bug in Linaro ARM compiler... The bug has been fixed quite a long time ago already...

nmoinvaz:
@mtl1979 do you think it would be a good idea to combine the code? 

There is some slight differences I'm not sure about..

image

It looks like may be the ARM `fill_window` is better in this area.

mtl1979:
@nmoinvaz bottom part of the left side is pretty much same as the code on the right side...

Dead2:
This looks good to me, but I have not tested it yet.

insert_string.h
```diff
@@ -0,0 +1,81 @@
+#ifndef INSERT_STRING_H_
+#define INSERT_STRING_H_
+
+/* insert_string.h -- Private insert_string functions shared with more than
+ *                    one insert string implementation
+ *
+ * Copyright (C) 1995-2013 Jean-loup Gailly and Mark Adler
+ * For conditions of distribution and use, see copyright notice in zlib.h
```
Dead2: I am pretty sure that quick_insert_string was not made by the two above, does this copyright notice need additions?

Dead2:
A little hard to review, I would have preferred the changes to move quick_insert_string in one commit and another commit for converting users of insert_string to quick_insert_string.
Other than that and the copyright comment, this is looking quite good.
It does need a rebase now though.

insert_string.h
nmoinvaz: Fixed.

nmoinvaz:
Rebased.

nmoinvaz:
I have split it into two commits.

Dead2:
Hmm, this one is giving me very mixed results.
Levels 1: Same compression ratio, 9.5% slower.
Levels 2-4: Slightly worse compression, 5.5-10% faster
Levels 5-9: Noticable drop in compression ratio, 13-50% slower

Performance difference between the two commits is not large, but with only the first commit, levels 1-4 have identical compression output (compared to without this PR), levels 5+ have worse compression but not quite as bad as with both commits. First commit alone is slower than both together.

There is so much going on in this PR that it is difficult to see what causes these changes.
Would it be beneficial to split this PR into smaller pieces?
One possible change could be to implement functable first, without changing the functions that gets called. Or the other way around, change the code without putting it in functable initially.
Any change that can be done separately could be separated out into separate PRs possibly, so this PR can shrink its scope a bit.

nmoinvaz:
Have you tried without the commit that does quick_insert_string changes? Perhaps that is the culprit.

nmoinvaz:
Those are discouraging results. Worse compression maybe due to level_mask.

nmoinvaz:
How can level 1 be 9.5% slower and the only change was to use functable?

Dead2:
Well, that's what I mean about splitting up the changes. If the changes to let deflate_quick use functable without other changes to the code was an initial commit, then we could easily see that that is actually the reason for that, and it would be easier to test alternatives that might improve upon that specific slowdown.

I try to live by the linux kernels patch rules if possible, and would prefer if others tried to do so as well. See especially rule #3 https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html

Make deflate_quick use functable for quick_insert_string is one logical change. Changing quick_insert_string to be arch-independent is one logical change. Changing it to be compatible with insert_string would be one logical change. Etc.  This would let us do a bisect with benchmarking to pinpoint what made things so slow, and would also allow us to bisect to find the source of a bug or compression result changes.

Dead2:
Btw, what I do when I want to split up commits, I do a soft git reset back to the commit before the ones I want to split up. Then I do `git add -p` (can optionally take file-names too), that lets you choose what changes to add interactively by patch-blocks. The 's' command is quite useful to split up multiple changes in one block into separate blocks, and you can also edit a block in case it has multiple changes in the same lines.

It is a fair bit of work though, but that at least makes it easier to do without having to re-do the same changes piecemeal.
I find the fixup/rebase trick from our wiki is also useful for the cases where I forgot to add a change to a previous commit and want to correct that.

nmoinvaz:
I think I can split things up more with regard to `level_mask`. A lot of work goes into these pull requests and getting them to pass across all the CI instances (I suppose that is my own fault since I added a lot more of them). Splitting things up it usually an after thought and is a lot of work in itself. I will try to split things up a bit better in the future.

I am probably going to have to do my own performance testing on this one to track down where the issue is.

Dead2:
I am the same, I always start doing multiple related or unrelated changes at the same time and have to split them up like described above afterwards, that is the tedious part of it and can often be time consuming. It is hard to be creative if you have to limit yourself to only do a single change at a time and having to plan out multiple changes/steps ahead of time.

mtl1979:
Developers don't have linear sense of time... We go back and forth between different projects and parts of code... Some things are written in mere minutes, but some things take years to finish. I still have quite a lot of unfinished patches from time when I started working on zlib-ng...

We're lucky that GitHub allows force-pushing, as for example SourceForge doesn't.

nmoinvaz:
I have split it up again. I removed the level mask code I had in there before. And I separated out the change to functable for easier performance testing. I did some tests using @neurolabusc's [pigz-bench-python](https://github.com/neurolabusc/pigz-bench-python/) on the final result which can be viewed [here](https://gist.github.com/nmoinvaz/00c973fd13bae8693337afc9a27b95d0).

nmoinvaz:
I've conducted some more tests and noticed that level masking with `TRIGGER_LEVEL` check in `insert_string` and `quick_insert_string` significantly hurt performance at levels 6 and 9.  The Cloudflare branch does not do this type of check.

From d306c75d3bb36cba73aec9b3b3ca378e31d1799e:

> For both new hash functions, we hash 4 bytes, instead of 3, for certain
levels. This shortens the hash chains, and also improves the quality
of each hash entry.

It might be better to remove the trigger level check from all `UPDATE_HASH` functions and always hash 4 bytes instead of 3.

mtl1979:
There is several issues with TRIGGER_LEVEL and optimized versions of insert_string... It might be time to remove it completely.

Dead2:
This looks better, performance-wise.

Current HEAD:
```
 1      49.603%    0.196/0.202/0.204       0.132/0.136/0.140  20804915
 2      38.062%    0.454/0.465/0.474       0.132/0.137/0.141  15964205
 3      37.021%    0.558/0.570/0.582       0.130/0.134/0.136  15527698
 4      35.832%    0.632/0.640/0.652       0.126/0.133/0.136  15028898
 5      35.231%    0.761/0.776/0.787       0.126/0.131/0.133  14776760
 6      35.125%    1.084/1.104/1.132       0.121/0.128/0.132  14732613
 7      34.832%    2.118/2.153/2.183       0.125/0.131/0.135  14609649
 8      34.775%    3.844/3.890/3.937       0.119/0.128/0.133  14585592

 avg    37.560%                1.225                   0.132
 tot                         147.013                  15.872
```
PR522:
```
 1      49.603%    0.208/0.216/0.220       0.130/0.135/0.137  20804915
 2      38.062%    0.430/0.433/0.439       0.131/0.135/0.137  15964310
 3      37.021%    0.525/0.536/0.542       0.127/0.133/0.136  15527879
 4      35.832%    0.595/0.606/0.613       0.128/0.131/0.133  15029207
 5      35.720%    0.652/0.662/0.678       0.123/0.129/0.134  14982036
 6      35.651%    0.786/0.805/0.822       0.125/0.129/0.131  14953101
 7      35.031%    1.436/1.455/1.475       0.125/0.128/0.133  14692984
 8      34.996%    2.254/2.284/2.348       0.124/0.130/0.133  14678469

 avg    37.740%                0.875                   0.131
 tot                         104.954                  15.730
```

As you can see, PR522 is significantly faster, but its compression is worse from level 5 and up.
deflate_quick is slightly slower, but we can live with that.

Most of the decrease in compression comes from the removal of this in insert_string.h
```
    if (s->level >= TRIGGER_LEVEL)
        val &= 0xFFFFFF;
```
This is the result with the above code snippet added back to insert_string.h without the TRIGGER_LEVEL check:
```
 5      35.230%    0.739/0.752/0.767       0.128/0.131/0.134  14776532
 6      35.127%    1.056/1.072/1.098       0.121/0.129/0.132  14733422
 7      34.833%    2.091/2.108/2.144       0.124/0.130/0.133  14610180
 8      34.776%    3.836/3.874/3.974       0.127/0.131/0.135  14585981
```

I tried to find out where the rest of the decrease in compression comes from, but I have not been able to, there is so much going on in this PR that it is very difficult to actually understand all the changes. Preferably, changes that change the output should be in separate commits from all other changes if possible.
We should know where this change to compression comes from before I merge this, for all I know it is a hidden bug causing this.


For reference I'll add a benchmark of stock gzip on Fedora 30:
```
 1      38.815%    0.659/0.664/0.666       0.249/0.251/0.254  16280312
 2      38.010%    0.713/0.717/0.719       0.243/0.246/0.249  15942360
 3      37.125%    0.891/0.893/0.896       0.238/0.240/0.242  15571251
 4      36.048%    0.892/0.897/0.904       0.232/0.236/0.241  15119798
 5      35.265%    1.273/1.277/1.280       0.231/0.234/0.237  14791400
 6      35.053%    2.101/2.120/2.144       0.231/0.234/0.235  14702131
 7      34.961%    2.744/2.752/2.756       0.228/0.232/0.234  14663578
 8      34.896%    4.722/4.736/4.749       0.233/0.233/0.234  14636481

 avg    36.272%                1.757                   0.238
```

mtl1979:
That masking with higher compression levels is so it always matches 4 bytes, to speed things up. If first three bytes are the same, but fourth byte isn't, the hash becomes different.

As far as I understand, trying to find 3 byte matches with higher compression levels doesn't make much sense except when similar four byte match (same first three bytes) has very long distance which is supposed to be quite unlikely.

However we have to remember that masking only works correctly on little-endian machines. On big-endian machines it will mask out the wrong byte.

It might be useful to test if change to using quick_insert_string for length 1 had effect on compression ratio.

Dead2:
Yep, quick_insert_string is missing a check that insert_string has.
I have not investigated whether this check is needed or what it actually does.

I changed the quick code to this:
```
    UPDATE_HASH(s, h, val);
    hm = h & s->hash_mask;

    head = s->head[hm];
    if (head != str) {
        s->prev[str & s->w_mask] = head;
        s->head[hm] = str;
    }
    return head;
```
- Added if check
- Added blank line to match insert_string
- Renamed 'ret' to 'head' to match insert_string

With this, levels 1-4 have identical compressed size as HEAD.
It looks like levels 5-9 generally get a slightly smaller compressed size, but occasionally a slight increase. (Would be identical to HEAD if it wasn't for the 3 vs 4 byte hashing)
Performance is virtually unchanged from the extra branch.

@nmoinvaz If you agree with the above changes, could you add them to the PR?
Other than this issue and a docstring for quick_insert_string, this looks ready to merge.

nmoinvaz:
Those changes look good. I have updated the PR. I'm looking forward to these changes being merged.

Dead2:
Changes look good, will probably merge tomorrow unless I find a problem.

Dead2:
@nmoinvaz Hey, had a thought while getting ready for bed; insert_string.h should really be named insert_string_tpl.h or something like that, to make it obvious it is actually a template and not just a place to stuff defines etc. Best start the naming scheme now, before we get more templates. :)

nmoinvaz:
Yes I had thought about that too, I figured I would make that change after the fact. But I can update the PR now that I know how you want it named.

nmoinvaz:
I have updated the PR with the new naming convention.

Dead2:
Good work 👍

PR #523, opened 2020-02-08
Codecov Improvements

Dead2:
- Fix codecov for pull requests.
- Fail build if codecov fails
- Don't disable optimizations and force debuginfo when enabling code coverage
- Make sure GCC links in libgcov
- Attempt to re-run codecov several times if it fails to connect.

This has been tested and verified to work on all current CI builds.
But sometimes it seems that the build VM just isn't able to connect to codecov.io despite retrying 4 times (plus codecov internal retries) over ~4 minutes. I am unsure whether this is possibly caused by bad GHA routing or something like that.
Nevertheless, I have seen several instances of this re-trying successfully working after codecov initially fails. Hopefully this will lead to more stable coverage percentages, no longer causing readme-changes to show a change in code coverage for example.


nmoinvaz:
I have increased the number of tries that the `codecov-python` project uses. I also added a command line argument `--trys` to specify the number of attempts if we need to increase it again. https://github.com/nmoinvaz/codecov-python/commit/e28a87fed0572c48ef1a96ac3286e7ddd246ffea

I think it would be better to use `codecov-python` that we were using before - no extra bash scripts, no having to pass the arguments to the extra script, etc. I can even transfer the fork to zlib-ng organization if you would prefer.

mtl1979:
"trys" is not a real word. Plural of "try" is "tries" similar to all words ending in -y.

nmoinvaz:
I was just going to keep the spelling the same as the original author had it but I changed it to `tries`.

CMakeLists.txt
```diff
@@ -258,10 +258,12 @@ else()
 endif()
 if(WITH_CODE_COVERAGE)
     if(CMAKE_C_COMPILER_ID MATCHES "Clang")
-        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O0 -g -coverage")
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -coverage")
```
nmoinvaz, line 261: Are you sure you want to remove `-O0`? Perhaps it will be better if it is `-O1`. If it is `O3` then maybe certain lines will be optimized away and code coverage % will decrease.

Btw, I think the sanitizers need `-O1`.

Dead2: The problem with using -O0 with coverage is that our CI then never even tests compiler optimizations, and that could be bad in itself. It'd be better if some of the CI runs are -O0 or -O1 instead of all of them.

Dead2:
I am so disappointed with codecov.io now.
- Random commits never get their reports processed completely/finished by codecov.io. [1]
- Pretty much every commit/PR has codecov upload failures, despite retrying and retrying. This has been happening for a long long time now (and we had the same issues when we used Travis), and their status page claims their systems are all up and running flawlessly.
- Very little development seems to be happening from codecov.io's side.
- The sourcegraph +codecov (+github) integration seems to be failing for us. On other github projects with this integration I see coverage/hits info inline with the code, and coverage data in the header. On our repo that does not seem to work. (I have been wondering whether this is because cmake puts the object files in its own proprietary dirs instead of putting them next to the source files., and makes codecov unable to identify the object<-> source connection. But I dunno.)
- The PR report (see above) is often missing data, such as several of the "flags" not having any coverage data. Is codecov.io loosing data internally in addition to external uploads failing?

So, I am considering testing coveralls again, they seem to have improved since the last time I looked at them, and they even support Github Actions.

[1] Random example: https://codecov.io/gh/zlib-ng/zlib-ng/commit/98666ba14a3a9a1f927dbc56090cbc7816c19504

nmoinvaz:
I agree that there appears to be little development on codecov.io or that it is going very slowly. Their python uploader has been languishing and they have been telling everybody to use their bash uploader. I had to make modifications to it, for this project and for work where we use GitLab private runners and don't have access to GitHub Action's nice `shell: bash` feature. Python is a better cross-platform solution anyways. Also for a project our size, the coverage reports are sometimes stuck saying "In Progress.."

Dead2:
I attempted to bring this to the attention of codecov.io support, but they didn't want to look at the issues due to it being a free open-source account. 

Coveralls.io will be my next test, right after they fix the security issue I exposed with them earlier today..

Dead2:
I deleted all the project data on codecov.io, thinking that might fix things.
And it does look better so far, hoping it will proceed like that.

Dead2:
Any objections to pulling this as-is? It would give us some coverage data for PRs at least.

codecov.io does not work great even with this unfortunately, and resetting the project data on codecov.io seems to have done almost nothing (it was better for a couple builds, then it was right back to how it was again).

.github/workflows/cmake.yml
```diff
@@ -325,10 +325,15 @@ jobs:
         LSAN_OPTIONS: ${{ matrix.lsan-options || 'verbosity=1' }}
 
     - name: Upload coverage report
-      if: matrix.codecov
+      if: matrix.codecov && ( env.CODECOV_TOKEN_SECRET != '' || github.repository == 'zlib-ng/zlib-ng')
```
nmoinvaz: There is an empty space after (

nmoinvaz:
Why still use the bash script?

nmoinvaz:
I see what you are trying to do.

.github/workflows/cmake.yml
Dead2: Fixed by adding another space at the end, to keep to the style used in the file.

CMakeLists.txt
Dead2: Modified a couple runs to compile with debug symbols, and one with -O1

Dead2:
Any ideas why the cmake CI failed now?
See https://github.com/zlib-ng/zlib-ng/actions/runs/38790706

The error messages are helpful as always..
```
.github#L1
The template is not valid. Unexpected value ''
```
```
.github#L1
One or more errors occurred.
```

It is annoying that all CI checks passes when the github actions config file is invalid..

nmoinvaz:
@Dead2 you changed the name from `Ubuntu 18.04 GCC OSB` to `Ubuntu 18.04 GCC OSB -O1` but you also have to change it in the matrix list above.

.github/workflows/cmake.yml
```diff
@@ -325,10 +326,15 @@ jobs:
         LSAN_OPTIONS: ${{ matrix.lsan-options || 'verbosity=1' }}
 
     - name: Upload coverage report
-      if: matrix.codecov
+      if: matrix.codecov && ( env.CODECOV_TOKEN_SECRET != '' || github.repository == 'zlib-ng/zlib-ng' )
```
nmoinvaz, line 329: I think this should be `secrets.CODECOV_TOKEN != ''`

.github/workflows/cmake.yml
```diff
@@ -325,10 +326,15 @@ jobs:
...
+      shell: bash
       run: |
-        cd ${{ matrix.build-dir || '.' }}
-        python -m codecov -F "${{ matrix.codecov }}" --name "${{ matrix.name }}" --gcov-exec="${{ matrix.gcov-exec || 'gcov' }}"
+        bash tools/codecov-upload.sh
       env:
         # Codecov does not yet support GitHub Actions
-        CODECOV_TOKEN: "${{secrets.CODECOV_TOKEN}}"
\ No newline at end of file
+        CODECOV_TOKEN_SECRET: "${{secrets.CODECOV_TOKEN}}"
```
nmoinvaz, line 335: Is `CODECOV_TOKEN_SECRET` needed?

mtl1979: It is needed for pull requests as the scope for them is the source repository not zlib-ng/zlib-ng which is the destination repository.

nmoinvaz: The variable is not used in the script.

mtl1979: See lines 329 and 336... It uses the hard-coded token for zlib-ng/zlib-ng repository but allows overriding it for other repositories...

.github/workflows/cmake.yml
mtl1979: It doesn't have access to secrets.CODECOV_TOKEN ...

.github/workflows/cmake.yml
nmoinvaz: Couldn't an overriding repository just set `secrets.CODECOV_TOKEN` in their repository? That is what I do in mine. There is no reason for `CODECOV_TOKEN_SECRET` or `env.CODECOV_TOKEN_SECRET` when the GitHub runner already provides `secrets.CODECOV_TOKEN`. Maybe I am missing something.

mtl1979: Yes... But the action still wouldn't have access to the variable... It's basically security restriction...

nmoinvaz: You mean the script wouldn't have access to that variable? It has access to `CODECOV_TOKEN` already, so there is no need for a duplicate token `CODECOV_TOKEN_SECRET`. The action does have access to `secrets.CODECOV_TOKEN`. The `if` statement above does not work in my repository. I had to modify it. There is a mismatch between an environment variable and a secrets variable I think. There is no need for an environment variable in the if statement `env.CODECOV_TOKEN_SECRET` because everything is run by a GitHub Actions runner which has secrets variables.

mtl1979: It doesn't have access to the variable when it's a pull request... There is no problems when it's normal commit.

nmoinvaz: I think that is why it is checking if repo is zlib-ng in the `if` statement for PR and the || statement in `CODECOV_TOKEN`.

mtl1979: My knowledge of codecov is quite limited, so I don't know how to fix it so it works 100%... I just know the limitations...

Dead2: These things are difficult for me to test, because I have access. A PR from someone external could experiment with whether this is actually needed or not, but as-is should work for everyone. My testing led me to believe that this was needed, but it might possibly not be.

nmoinvaz: https://github.community/t5/GitHub-Actions/jobs-lt-job-id-gt-if-does-not-work-with-env-secrets/td-p/38549
https://github.community/t5/GitHub-Actions/If-expression-with-context-variable/td-p/34556

There doesn't however appear to be a way for me to even set special `env` vars, so using it is not an option. I can only set secrets in my forked repo. And `if` statement in yaml doesn't support secrets.

PR #525, opened 2020-02-09
Added X86 AVX2, SSE2, SSE4, and PCLMULQDQ cmake options and CI configs

nmoinvaz:
This PR came from discussion in #502 and adds the following cmake options:

- WITH_AVX2
- WITH_SSE2
- WITH_SSE4
- WITH_PCLMULQDQ

They are added so we can independently disable them and verify that CI still works. I have added the associated CI configurations. And have also added CI configurations for turning off ACLE and NEON on ARM.

In deflate_quick.c I now use `functable.fill_window` because `-DWITH_SSE2=OFF` was not working because it was hard coded to use `fill_window_sse`.


README.md
```diff
@@ -95,6 +95,10 @@ Build Options
...
 | WITH_NATIVE_INSTRUCTIONS |                          | Instruct the compiler to use the full instruction set on this host (gcc/clang -march=native) | OFF                              |
 |                          | --force-sse2             | Assume SSE2 instructions are always available                                                | DISABLED (x86), ENABLED (x86_64) |
+| WITH_AVX2                |                          | Build with AVX2 slide hash                                                                   | ON                               |
+| WITH_SSE2                |                          | Build with SSE2 slide hash                                                                   | ON                               |
+| WITH_SSE4                |                          | Build with SSE4 CRC32-C                                                                      | ON                               |
+| WITH_PCLMULQDQ           |                          | Build with PCLMULQDQ CRC                                                                     | ON                               |
```
Dead2: I am thinking these options should not really be tied up to slide-hash or crc specifically, just let them be generic options to disable the instruction sets completely along with any and all features that depend on them. I do see that WITH_ACLE already refers to crc specifically, but I'd rather we avoid that.
Otherwise, we should name the options for example WITH_AVX2_SLIDE_HASH etc, and I don't think we need to have that level of granularity in the cmake files.

We could have more "hidden" checks for defines like DISABLE_AVX2_SLIDE_HASH to specifically disable one feature for debugging/testing (by short-circuiting the functable checks for example), but that is too verbose for CMake IMO.

mtl1979: PCLMULQDQ and ACLE are both instruction sets that have limited use and such are exceptions, other are generic enough.

nmoinvaz: @Dead2, are you saying I should just change the readme?

Dead2: @nmoinvaz Readme and the cmake descriptions, yes.

nmoinvaz:
These changes are coming in handy for determining which architecture specific code is causing a particular problem.

.github/workflows/cmake.yml
```diff
@@ -87,6 +117,13 @@ jobs:
             packages: qemu gcc-arm-linux-gnueabi libc-dev-armel-cross
             codecov: ubuntu_gcc_armsf
 
+          - name: Ubuntu 18.04 GCC ARM SF No ACLE
+            os: ubuntu-18.04
+            compiler: arm-linux-gnueabi-gcc
+            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-arm.cmake -DCMAKE_C_COMPILER_TARGET=arm-linux-gnueabi -DWITH_SANITIZERS=ON -WITH_ACLE=OFF
```
mtl1979: Missing D in -WITH_ACLE=OFF

.github/workflows/cmake.yml
```diff
@@ -115,6 +152,13 @@ jobs:
             packages: qemu gcc-arm-linux-gnueabihf libc-dev-armel-cross
             codecov: ubuntu_gcc_armhf
 
+          - name: Ubuntu 18.04 GCC ARM HF No NEON
+            os: ubuntu-18.04
+            compiler: arm-linux-gnueabihf-gcc
+            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-arm.cmake -DCMAKE_C_COMPILER_TARGET=arm-linux-gnueabihf -DWITH_SANITIZERS=ON -WITH_NEON=OFF
```
mtl1979: Missing D in -WITH_NEON=OFF

CMakeLists.txt
```diff
@@ -460,109 +470,114 @@ if(MSVC)
...
 
-    # Check whether compiler supports PCLMULQDQ intrinics
-    if(WITH_NATIVE_INSTRUCTIONS)
-        set(CMAKE_REQUIRED_FLAGS "${NATIVEFLAG}")
-    else()
-        set(CMAKE_REQUIRED_FLAGS "${PCLMULFLAG}")
+        # FORCE_SSE2 option will only be shown if HAVE_SSE2_INTRIN is true
+        if("${ARCH}" MATCHES "i[3-6]86")
+            cmake_dependent_option(FORCE_SSE2 "Always assume CPU is SSE2 capable" OFF "HAVE_SSE2_INTRIN" OFF)
+        endif()
```
mtl1979: This doesn't actually disable SSE2 as the value of HAVE_SSE2_INTRIN is cached. It must be explicitly cleared.

CMakeLists.txt
```diff
@@ -460,109 +470,114 @@ if(MSVC)
...
+            #if defined(_MSC_VER)
+                crc = _mm_crc32_u32(crc, c);
+            #else
+                crc = __builtin_ia32_crc32qi(crc, c);
+            #endif
+                (void)crc;
                 return 0;
             }"
-            HAVE_PCLMULQDQ_INTRIN
+            HAVE_SSE42CRC_INTRIN
```
mtl1979: See above, all cached variables must be explicitly cleared.

CMakeLists.txt
```diff
@@ -460,109 +470,114 @@ if(MSVC)
...
+                    __m128i c = _mm_clmulepi64_si128(a, b, 0x10);
+                    (void)c;
+                    return 0;
+                }"
+                HAVE_PCLMULQDQ_INTRIN
+            )
+        else()
+            set(HAVE_PCLMULQDQ_INTRIN NO)
+        endif()
+        set(CMAKE_REQUIRED_FLAGS)
```
mtl1979: See above, cached variables not cleared.

CMakeLists.txt
```diff
@@ -460,109 +470,114 @@ if(MSVC)
...
+            int main(void) {
+                __m256i x = _mm256_set1_epi16(2);
+                const __m256i y = _mm256_set1_epi16(1);
+                x = _mm256_subs_epu16(x, y);
+                (void)x;
+                return 0;
+            }"
+            HAVE_AVX2_INTRIN
+        )
+        set(CMAKE_REQUIRED_FLAGS)
```
mtl1979: See above, cached variables must be explicitly cleared.

.github/workflows/cmake.yml
nmoinvaz: Fixed.

CMakeLists.txt
nmoinvaz: Fixed.

CMakeLists.txt
nmoinvaz: Fixed.

.github/workflows/cmake.yml
nmoinvaz: Fixed thanks.

.github/workflows/cmake.yml
nmoinvaz: Fixed.

nmoinvaz:
Changes made.

CMakeLists.txt
```diff
@@ -88,8 +88,13 @@ option(WITH_OPTIM "Build with optimisation" ON)
...
 option(WITH_NATIVE_INSTRUCTIONS
     "Instruct the compiler to use the full instruction set on this host (gcc/clang -march=native)" OFF)
-if(BASEARCH_ARM_FOUND)
-    option(WITH_ACLE "Build with ACLE CRC" ON)
+if(BASEARCH_X86_FOUND)
+    option(WITH_AVX2 "Build with AVX2" ON)
+    option(WITH_SSE2 "Build with SSE2" ON)
+    option(WITH_SSE4 "Build with SSE4" ON)
+    option(WITH_PCLMULQDQ "Build with PCLMULQDQ" ON)
+elseif(BASEARCH_ARM_FOUND)
```
mtl1979: I see no reason why BASEARCH_ARM_FOUND shouldn't be the first as A is first letter in Latin alphabet.

CMakeLists.txt
```diff
@@ -105,8 +110,13 @@ add_feature_info(WITH_SANITIZERS WITH_SANITIZERS "Build with address sanitizer a
...
 add_feature_info(WITH_FUZZERS WITH_FUZZERS "Build test/fuzz")
 add_feature_info(WITH_NEW_STRATEGIES WITH_NEW_STRATEGIES "Use new strategies")
-if(BASEARCH_ARM_FOUND)
-    add_feature_info(WITH_ACLE WITH_ACLE "Build with ACLE CRC")
+if(BASEARCH_X86_FOUND)
+    add_feature_info(WITH_AVX2 WITH_AVX2 "Build with AVX2")
+    add_feature_info(WITH_SSE2 WITH_SSE2 "Build with SSE2")
+    add_feature_info(WITH_SSE4 WITH_SSE4 "Build with SSE4")
+    add_feature_info(WITH_PCLMULQDQ WITH_PCLMULQDQ "Build with PCLMULQDQ")
+elseif(BASEARCH_ARM_FOUND)
```
mtl1979: I see no reason why BASEARCH_ARM_FOUND shouldn't be the first as A is first letter in Latin alphabet.
Changing order of existing tests clutters the patch unnecessarily.

CMakeLists.txt
nmoinvaz: Good catch. I have corrected it.

CMakeLists.txt
nmoinvaz: Good catch. I have corrected it.

nmoinvaz:
@mtl1979 I have reordered the variables like you suggested.

Dead2:
Looks good. Thanks for splitting this into two commits, reviewing CMakeLists.txt without that was hard (until I checked whether it was split into more than one commit).

nmoinvaz:
I updated the codecov names in cmake.yml.

nmoinvaz:
Rebased.

nmoinvaz:
I must be dreaming. Did it say this increases code coverage by 13%?

arch/x86/deflate_quick.c
```diff
@@ -224,7 +224,7 @@ ZLIB_INTERNAL block_state deflate_quick(deflate_state *s, int flush) {
         }
 
         if (s->lookahead < MIN_LOOKAHEAD) {
-            fill_window_sse(s);
+            functable.fill_window(s);
```
Dead2: These changes do not seem to belong here.

Dead2:
I am thinking these options should be hidden by default, they have no value for the average user. Uncertain whether they should be listed in the options table too, perhaps a separate table for advanced options meant for debugging and development?
See: https://cmake.org/cmake/help/v3.0/command/mark_as_advanced.html

arch/x86/deflate_quick.c
nmoinvaz: They belong there because without them it is not possible to do `-WITH_SSE2=OFF`

nmoinvaz:
I have moved the architecture specific options to advanced options and created a separate table in the README.md.

arch/x86/deflate_quick.c
nmoinvaz: Ok I have removed them.

nmoinvaz:
Rebased.

CMakeLists.txt
```diff
@@ -105,13 +98,30 @@ add_feature_info(WITH_SANITIZERS WITH_SANITIZERS "Build with address sanitizer a
...
+elseif(BASEARCH_X86_FOUND)
+    set(WITH_AVX2 "Build with AVX2" ON)
+    set(WITH_SSE2 "Build with SSE2" ON)
+    set(WITH_SSE4 "Build with SSE4" ON)
+    set(WITH_PCLMULQDQ "Build with PCLMULQDQ" ON)
+    mark_as_advanced(WITH_AVX2)
+    mark_as_advanced(WITH_SSE2)
+    mark_as_advanced(WITH_SSE4)
+    mark_as_advanced(WITH_PCLMULQDQ)
+endif()
```
Dead2: All the above need to be option() not set(), otherwise they never show up in ccmake, even in advanced mode.

mark_as_advanced can mark multiple options as advanced.
According to the documentation of mark_as_advanced, it should either be used before making the option, or with the FORCE parameter to change it after the fact. I have not been able to reproduce any ill effects AFAIK, but I know there are potential problems with cache there.

I suggest adding this after the group of options.
```
    option(..)
    mark_as_advanced(FORCE WITH_AVX2 WITH_SSE2 WITH_SSE4 WITH_PCLMULQDQ)
```

nmoinvaz: Ok let me know what you think now.

Dead2: Unless you know something I don't, you are missing the FORCE parameter. According to how I understood the docs, it is required when setting it after creating the options to avoid problems with it getting set as a new empty variable due to weirdness in the caching.
I think they might have fixed it in some of the most recent versions (saw a ticket about it from last year I think it was).

nmoinvaz: Sorry I must have missed it. I’ve never used the mark_as_advanced macro before.

nmoinvaz:
I have reordered the commits and squashed them from 3 to 2, to make it easier to review, and added FORCE to mark_as_advanced.

CMakeLists.txt
nmoinvaz: Fixed.

Dead2:
Thanks

PR #528, opened 2020-02-10
Add cpuidex.

mtl1979:
* On Linux ecx wasn't cleared, so AVX2 test failed
* Add cpuidex() so ecx can be set to correct value for AVX2 test, so it works on both Windows and Linux

### DIFF
--- a/arch/x86/x86.c
+++ b/arch/x86/x86.c
@@ -33,15 +33,21 @@ static void cpuid(int info, unsigned* eax, unsigned* ebx, unsigned* ecx, unsigne
     *ecx = registers[2];
     *edx = registers[3];
 #else
-    unsigned int _eax;
-    unsigned int _ebx;
-    unsigned int _ecx;
-    unsigned int _edx;
-    __cpuid(info, _eax, _ebx, _ecx, _edx);
-    *eax = _eax;
-    *ebx = _ebx;
-    *ecx = _ecx;
-    *edx = _edx;
+    __cpuid(info, *eax, *ebx, *ecx, *edx);
+#endif
+}
+
+static void cpuidex(int info, int subinfo, unsigned* eax, unsigned* ebx, unsigned* ecx, unsigned* edx) {
+#ifdef _MSC_VER
+    unsigned int registers[4];
+    __cpuidex(registers, info, subinfo);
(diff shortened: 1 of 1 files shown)


nmoinvaz:
I think `__get_cpuid_count` takes register parameters as pointers: `__get_cpuid_count(info, subinfo, &_eax, &_ebx, &_ecx, &_edx);` Maybe that is why ASAN is failing.

Also it doesn't appear that ecx is being set to zero - or maybe I am wrong.

mtl1979:
info = eax, subinfo = ecx

arch/x86/x86.c
```diff
@@ -45,6 +45,20 @@ static void cpuid(int info, unsigned* eax, unsigned* ebx, unsigned* ecx, unsigne
...
+#ifdef _MSC_VER
+    unsigned int registers[4];
+    __cpuidex(registers, info, subinfo);
+
+    *eax = registers[0];
+    *ebx = registers[1];
+    *ecx = registers[2];
+    *edx = registers[3];
+#else
+    __get_cpuid_count(info, subinfo, eax, ebx, ecx, edx);
```
nmoinvaz: Perhaps we should use `__get_cpuid` in cpuid() too. What do you think?

mtl1979: __get_cpuid doesn't clear ecx register either... Only thing it does extra is check the maximum leaf number.

nmoinvaz: I was suggesting we standardize and use `__get_cpuid` in cpuid() and `__get_cpuid_count` in cpuidex(). Otherwise we should stick to `__cpuid` in cpuid() and `__cpuid_count` in cpuidex().

mtl1979: Seems that __get_cpuid and __get_cpuid_count cause some tests to fail... I can't open the logs to see why they fail, but __cpuid and __cpuid_count don't cause test fails, so I see no reason why we can't use them as we already test for maximum leaf number.

nmoinvaz: Sounds good.

nmoinvaz:
Looks good.

Dead2:
Looks good to me :)

PR #529, opened 2020-02-12
Fixed wrong command used in infcover unit test

nmoinvaz:


nmoinvaz:
There appears to be a MSAN error in infcover.

Dead2:
I noticed an infcover failure yesterday too, but have not had time to look into it.
```
infcover: /home/hansr/github/zlib/zlib-ng/test/infcover.c:562: try: Assertion `ret == Z_DATA_ERROR' failed.
Aborted (core dumped)
```

nmoinvaz:
I have modified _infcover_ so that certain tests that require  `INFLATE_ALLOW_INVALID_DISTANCE_TOOFAR_ARRR` to be defined are only run when it is defined.

Dead2:
@nmoinvaz Isn't that inverted logic?
If TOOFAR is defined, shouldn't it NOT error on those, while it should error on them if it is not defined?
I might be missing something here though.

nmoinvaz:
I think the logic is ok. If you look at the code for `inflateUndermine` it returns `Z_OK` if `INFLATE_ALLOW_INVALID_DISTANCE_TOOFAR_ARRR` is defined otherwise it returns `Z_DATA_ERROR`. For the try() statement, it is only checking that condition, if we have the #define to support it.

Dead2:
Hmm, I don't see what I thought I saw earlier. Looks ok now :)

nmoinvaz:
I am not able to reproduce this MSAN error on Ubuntu 14 or 16 using Clang 4.0.1 or 9.0.1 -- and I've tried to get MSAN to trace origin on CI without success.

```
Origin A (origin_id 9f200001):
  raw origin id: -1625292799
  Uninitialized value was stored to memory at
    #0 0x4277e3  (/home/runner/work/zlib-ng/zlib-ng/infcover+0x4277e3)
    #1 0x55728a  (/home/runner/work/zlib-ng/zlib-ng/infcover+0x55728a)

  Uninitialized value was created by an allocation of 'win.i' in the stack frame of function 'main'
    #0 0x493fb0  (/home/runner/work/zlib-ng/zlib-ng/infcover+0x493fb0)

==6228==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x557c0a  (/home/runner/work/zlib-ng/zlib-ng/infcover+0x557c0a)
    #1 0x557153  (/home/runner/work/zlib-ng/zlib-ng/infcover+0x557153)
    #2 0x427473  (/home/runner/work/zlib-ng/zlib-ng/infcover+0x427473)
    #3 0x7f8e5e2b6040  (/lib/x86_64-linux-gnu/libc.so.6+0x43040)
    #4 0x7f8e5e2b6139  (/lib/x86_64-linux-gnu/libc.so.6+0x43139)
    #5 0x7f8e5e294b9d  (/lib/x86_64-linux-gnu/libc.so.6+0x21b9d)
    #6 0x41a9d9  (/home/runner/work/zlib-ng/zlib-ng/infcover+0x41a9d9)

  raw origin id: -1625292799
  Uninitialized value was stored to memory at
    #0 0x4277e3  (/home/runner/work/zlib-ng/zlib-ng/infcover+0x4277e3)
    #1 0x55728a  (/home/runner/work/zlib-ng/zlib-ng/infcover+0x55728a)

  Uninitialized value was created by an allocation of 'win.i' in the stack frame of function 'main'
    #0 0x493fb0  (/home/runner/work/zlib-ng/zlib-ng/infcover+0x493fb0)

SUMMARY: MemorySanitizer: use-of-uninitialized-value (/home/runner/work/zlib-ng/zlib-ng/infcover+0x557c0a) 
```

Dead2:
Try enabling debug (-g3 for example)? Maybe MSAN can give a bit more info then, unless that "fixes" the bug of course.

nmoinvaz:
I'm not able to get it to reproduce even with `-g3`. Very strange. I have even confirmed MSAN is working by adding a dummy MSAN error.

test/infcover.c
```diff
@@ -590,7 +595,9 @@ static void cover_inflate(void) {
     try("4 80 49 92 24 49 92 24 f b4 ff ff c3 84", "invalid distances set", 1);
     try("4 c0 81 8 0 0 0 0 20 7f eb b 0 0", "invalid literal/length code", 1);
     try("2 7e ff ff", "invalid distance code", 1);
+#ifdef INFLATE_ALLOW_INVALID_DISTANCE_TOOFAR_ARRR
     try("c c0 81 0 0 0 0 0 90 ff 6b 4 0", "invalid distance too far back", 1);
+#endif
```
Dead2, line 603: This is the part I was wondering about earlier, and I am again now. So it'll probably need a comment if it is correct.
This looks to me like we disable the check of invalid distance unless we allow those invalid distances.

Isn't this supposed to check that we do return the correct error when distance is too far back? In that case, we should probably be testing for this by default, unless we allow the invalid distances, then it makes little sense to check them.

Dead2:
Should perhaps infcover.c be compiled with asserts enabled (not the whole codebase), since it relies a lot on assert()? Not sure about this, but it just seems like infcover loses most of its checks when asserts are disabled.

test/infcover.c
nmoinvaz: We disable the check because that logical branch does not exist. When `INFLATE_ALLOW_INVALID_DISTANCE_TOOFAR_ARRR` is defined, it will allow invalid distances. The try() function simply adds code coverage to the logic that is wrapped in `#ifdef INFLATE_ALLOW_INVALID_DISTANCE_TOOFAR_ARRR`. So we don't try to attempt that logical branch if `INFLATE_ALLOW_INVALID_DISTANCE_TOOFAR_ARRR` is defined because it doesn't exist.

nmoinvaz:
Yes you are correct about `assert`. Only the debug CI instance would fail, the release CI instances would succeed.

nmoinvaz:
I have added a comment about it. The best way I could re-enable the `assert` was to just `#undef NDEBUG` before including assert.h. There are more complicated solutions that involve string replacement of CMAKE variables.

test/infcover.c
nmoinvaz: Also, if allowing invalid distances in zlib-ng is something we want to do let me know. We can't just do it for infcover only because codecov will not see a zlibstatic-with-invalid-dist as the same thing as zlibstatic. So I left it turned off as is. And only enable check if turned on.

Dead2: I thought these tests are testing invalid behavior? If so, then shouldn't the test be enabled if too far distances are not allowed?
I have never completely understood infcover though, too much code, too little comments.

nmoinvaz: Are you suggesting we just get rid of the #define and always allow invalid distances in the lib?

Dead2: Not at all.
I suspect we are talking past each other.

nmoinvaz: It does test invalid behavior. I don’t know a clean way to enable it. We can’t just enable it for this application because setting the #define in zlib static or zlib projects would apply to all apps.
If we have a separate library with only the #define then I found that code coverage doesn’t see the zlib as the same as the other lib without the #define.

Dead2: I mean, isn't it supposed to test the invalid behavior and isn't it supposed to fail (and be catched by infcover, returning success if it failed) when the define isn't set? Doesn't this test check that the error is returned if it is invalid? Aren't we testing error-handling?
If we only run the test when it is not in fact invalid, then it would be a pointless test..  Do you see what I mean?

nmoinvaz: I made some modifications so that code still gets run even when the `#define` is not set.

nmoinvaz: I made the wrong assumption that the code did not need to be called if `#define` was not set. However, it should still be called to run the failure code path.

nmoinvaz:
It seems the distrinction between `INFLATE_STRICT` and `INFLATE_ALLOW_INVALID_DISTANCE_TOOFAR_ARRR` is:

* [INFLATE_STRICT](http://pipeline.lbl.gov/code/3rd_party/licenses.win/zlib-1.2.3/ChangeLog) is used to check distances against zlib header which restricts the window size.

* [INFLATE_ALLOW_INVALID_DISTANCE_TOOFAR_ARRR](http://mail.openjdk.java.net/pipermail/core-libs-dev/2012-March/009562.html) is used to allow distances beyond the window size with which it fills with zeros.

`inflateBack` does not support `INFLATE_ALLOW_INVALID_DISTANCE_TOOFAR_ARRR` and instead returns `Z_BUF_ERROR` for `try("c c0 81 0 0 0 0 0 90 ff 6b 4 0", "invalid distance too far back", 1);`

test/infcover.c
Dead2: Looks good now :)

PR #531, opened 2020-02-12
Disable alignment sanitizer with -fno-sanitize when unaligned reads supported

nmoinvaz:
Because `-fsanitize=undefined` is a sets a *group* of sanitizer flags (including _alignment_ sanitizer) we need to explicitly disable alignment sanitizer with `-fno-sanitize=alignment`.


nmoinvaz:
Rebased

nmoinvaz:
Codecov reported on GH before all checks completed..

Issue #533, opened 2020-02-13
ZLIB_COMPAT library artifact naming


kreuzerkrieg:
I'm trying to create [vcpkg](https://github.com/microsoft/vcpkg) port for `zlib-ng` and it looks like `-DZLIB_COMPAT` does not affect the output library name.
Looks like it should
```
if (ZLIB_COMPAT)
    add_definitions(-DZLIB_COMPAT)
    set(WITH_GZFILEOP ON)
    set(LIBNAME1 libz)
    set(LIBNAME2 zlib)
    set(SUFFIX "")
else()
    set(LIBNAME1 libz-ng)
    set(LIBNAME2 zlib-ng)
    set(SUFFIX "-ng")
endif()
```

but these `LIBNAMExxx` are not used

nmoinvaz:
@kreuzerkrieg can you verify that #534 fixes your issue?

kreuzerkrieg:
> @kreuzerkrieg can you verify that #534 fixes your issue?

affirmative

kreuzerkrieg:
Any chance above PR will be merged soon?

Dead2:
It will be soon I hope, but at least one week out.
I am trying to let codecov.io have a look at a stable state for the corruption bugs (on their end) when analyzing our repo/coverage data. 
Ref: https://community.codecov.io/t/codecov-server-500-error-and-endless-processing/1012/5

kreuzerkrieg:
Thanks for the update

nmoinvaz:
Fix was merged into `develop`.

PR #534, opened 2020-02-14
Fixed library not being named correctly on Windows with ZLIB_COMPAT

nmoinvaz:
See #533.

This change gets rid of `LIBNAME1` and `LIBNAME2` and just sets the output name of the library directly based on platform using `set_target_properties`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,12 +115,8 @@ add_feature_info(WITH_CODE_COVERAGE WITH_CODE_COVERAGE "Enable code coverage rep
 if (ZLIB_COMPAT)
     add_definitions(-DZLIB_COMPAT)
     set(WITH_GZFILEOP ON)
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Rebased

nmoinvaz:
Codecov reported on GH before all checks completed..

Dead2:
codecov edits its comment if more data is made available or a new change is pushed :)
And it actually seems to work now :)

Dead2:
Wouldn't this change break naming on linux, where we want it named libz when in compat mode?

nmoinvaz:
It still contains ```set_target_properties(${ZLIB_INSTALL_LIBRARIES} PROPERTIES OUTPUT_NAME z${SUFFIX})```. I suppose `LIBNAME1` was never used anywhere.

PR #537, opened 2020-02-20
Added cmake unit tests to verify gzip compatibility

nmoinvaz:
This PR confirms that gzip utility on Linux and Mac can decompress minigzip compressed data. Currently we only check to see that minigzip can decompress minigzip compressed data.

Buildkite seems to fail. I don't have access to that CI.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1024,6 +1024,7 @@ if (ZLIB_ENABLE_TESTS)
         endforeach()
         string(REPLACE ";" "" GZ_ARGS "${ARGN}")
         string(REPLACE "-" "" GZ_ARGS "${GZ_ARGS}")
+        # Test minigzip can decompress minigzip compressed
         add_test(NAME ${target}-${name}-${GZ_ARGS}-compr
             COMMAND ${CMAKE_COMMAND}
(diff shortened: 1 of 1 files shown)


Dead2:
Should we perhaps have gzip compress some files and have minigzip decompress them also?

nmoinvaz:
Yes that would be useful but perhaps that can be a separate PR. Do you know why Buildkite fails?

nmoinvaz:
I have added tests to verify minigzip can decompress gzip compressed.

Dead2:
gzip does not support -k (--keep) until version 1.6, that is too new for some current linux distros, like the most recent release of RHEL 7.x.

It might be worth using the long version of parameters, so they are understandable without looking up man files, in this case I found 3 online gzip man-pages that didn't list -k before I thought it might be --keep and googled for "gzip man keep" instead and finally found confirmation.

nmoinvaz:
Is the problem with buildkite the fact that -k is missing from gzip?

Dead2:
Yes, buildkite is running Scientific Linux 7, based on RHEL7

nmoinvaz:
I have made the suggested changes and rebased.

Issue #536, opened 2020-02-19
pigz compiled with zlib-ng generates errors with level=1


neurolabusc:
@reddydp discovered that pigz compiled with zlib-ng generates errors with compression level 1. This is specific to zlib-ng, as versions of pigz compiled to default, Intel or CloudFlare zlib does not show this error. Since the zlib-ng and lib-Intel use similar strategy for level-1 compression, I wonder if this is related to a fixed [intel zlib issue](https://github.com/jtkukunas/zlib/issues/8).

We have replicated the issue on Linux as well as MacOS. To see the issue run the following commands:

```
 git clone git clone https://github.com/neurolabusc/pigz-bench
 cd pigz-bench
 ./1compile.sh 
 ./4verify.sh
```

The issue is only seen with pigzNG with the level 1 compression:

```
Segmentation fault: 11  $cmd
pigzNG	1	1	33
pigzNG: skipping: /Users/chris/src/pigz-bench/corpus/asl16.nii.gz: corrupted -- invalid deflate data (invalid stored block lengths)
```

nmoinvaz:
It happens due to `deflateParams` call which changes the level mid-stream. It starts out as level 6 and then switches to level 1. 

I believe the max window size for deflate level 1 is 8192 (the size of `quick_dist_codes`). This is set as `(1 << 13)` here: https://github.com/zlib-ng/zlib-ng/blob/bd7d2226dde22fbd71477c96654934b0464981cb/deflate.c#L308

When the call to `deflateParams` happens, the `w_size` is never updated. This can allow `dist` in `deflate_quick` to exceed the size of `quick_dist_codes` array.  See the check here: https://github.com/zlib-ng/zlib-ng/blob/bd7d2226dde22fbd71477c96654934b0464981cb/arch/x86/deflate_quick.c#L240

I think that perhaps since `w_size` is dependent upon compression level, we should be setting it during `deflateParams` call. Also perhaps `configuration_level` should have a setting for max window size for each algorithm and we should use that.

Removing the check for `s->last_flush != 2` in  `deflateParams`  fixes the problem. It seems the change was made in https://github.com/zlib-ng/zlib-ng/commit/11413c0618ff9209639775a22bb7dbe492ef4b65#diff-8940271ef2146523af486ca4408361da and appears to not be part of Intel zlib.

Perhaps when changing the level, if the window size changes, it should force block to flush.

mtl1979:
[...]

PR #540, opened 2020-02-25
Fixed segmentation fault in deflate_quick() when switching levels using deflateParam

nmoinvaz:

I have added a unit test case that reproduces the segmentation fault. Here is the initial check-in of just the unit test causing the segmentation fault https://github.com/nmoinvaz/zlib-ng/runs/467173544 and afterwards after the fix is applied https://github.com/nmoinvaz/zlib-ng/runs/467198265.

Background:

- deflateInit would be initialized with a window size greater than 8K
- deflateParams called to switch to level 1 which does not update w_size
- Fault would occur because deflate_quick was not checking w_size bounds on dist when accessing quick_dist_codes

### DIFF
--- a/arch/x86/deflate_quick.c
+++ b/arch/x86/deflate_quick.c
@@ -29,6 +29,8 @@
 #  include <ctype.h>
 #endif
 
+#define QUICK_WINDOW_SIZE 8192
+
 extern void fill_window_sse(deflate_state *s);
 extern void flush_pending(PREFIX3(stream) *strm);
 
@@ -122,7 +124,7 @@ static inline long compare258(const unsigned char *const src0, const unsigned ch
 }
 
 static const unsigned quick_len_codes[MAX_MATCH-MIN_MATCH+1];
-static const unsigned quick_dist_codes[8192];
+static const unsigned quick_dist_codes[QUICK_WINDOW_SIZE];
 
 static inline void static_emit_ptr(deflate_state *const s, const int lc, const unsigned dist) {
     unsigned code1 = quick_len_codes[lc] >> 8;
@@ -209,6 +211,10 @@ static inline Pos quick_insert_string(deflate_state *const s, const Pos str) {
 ZLIB_INTERNAL block_state deflate_quick(deflate_state *s, int flush) {
     IPos hash_head;
     unsigned dist, match_len;
+    unsigned int wsize = s->w_size;
+
+    if (wsize > QUICK_WINDOW_SIZE)
+        wsize = QUICK_WINDOW_SIZE;
 
     if (s->block_open == 0) {
         static_emit_tree(s, flush);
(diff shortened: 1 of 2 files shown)


nmoinvaz:
Wrong branch name.

arch/x86/deflate_quick.c
```diff
@@ -122,7 +124,7 @@ static inline long compare258(const unsigned char *const src0, const unsigned ch
 }
 
 static const unsigned quick_len_codes[MAX_MATCH-MIN_MATCH+1];
-static const unsigned quick_dist_codes[8192];
+static const unsigned quick_dist_codes[QUICK_WINDOW_SIZE];
```
mtl1979, line 127: Completely unnecessary change...

arch/x86/deflate_quick.c
```diff
@@ -343,7 +349,7 @@ static const unsigned quick_len_codes[MAX_MATCH-MIN_MATCH+1] = {
     0x001c230d, 0x001d230d, 0x001e230d, 0x0000a308,
 };
 
-static const unsigned quick_dist_codes[8192] = {
+static const unsigned quick_dist_codes[QUICK_WINDOW_SIZE] = {
```
mtl1979, line 352: Completely unnecessary change...

arch/x86/deflate_quick.c
```diff
@@ -237,7 +243,7 @@ ZLIB_INTERNAL block_state deflate_quick(deflate_state *s, int flush) {
             hash_head = quick_insert_string(s, s->strstart);
             dist = s->strstart - hash_head;
 
-            if (dist > 0 && (dist-1) < (s->w_size - 1)) {
+            if (dist > 0 && (dist-1) < wsize) {
```
mtl1979, line 246: Should be ```wsize - 1```...

arch/x86/deflate_quick.c
nmoinvaz: It seems proper that `quick_len_codes` depends on #define value so `quick_dist_codes` should as well. I had to create the #define for checking bounds, might as well use it where can.

arch/x86/deflate_quick.c
nmoinvaz: I think here it should be `wsize`. I don't see the need for `-1` .

mtl1979: It's to avoid overflow... Obviously there has to be at least MIN_MATCH bytes before the match, otherwise there would not be anything to match with. So dist can't be more than ```wsize - MIN_MATCH```.

nmoinvaz: I thought here that `dist` refers to the distance where the match is found in the window, not the distance + `MIN_MATCH`. If `MIN_MATCH` is 3 then why isn't it `w_size - 3`?

mtl1979: In the worst case the original string starts at byte index 0, and match can be last 3 bytes of window (8189, 8190, 8191). I don't know why it isn't ```w_size - 3```, that is why I said the original test is also incorrect...

nmoinvaz: Let's move the conversation to the other PR #541. Sorry for the confusion.

PR #541, opened 2020-02-25
Fixed segmentation fault in deflate_quick() when switching levels using deflateParam

nmoinvaz:

I have added a unit test case that reproduces the segmentation fault. Here is the initial check-in of just the unit test causing the segmentation fault https://github.com/nmoinvaz/zlib-ng/runs/467173544 and afterwards after the fix is applied https://github.com/nmoinvaz/zlib-ng/runs/467198265.

Background:

- deflateInit would be initialized with a window size greater than 8K
- deflateParams called to switch to level 1 which does not update w_size
- Fault would occur because deflate_quick was not checking w_size bounds on dist when accessing quick_dist_codes

### DIFF
--- a/arch/x86/deflate_quick.c
+++ b/arch/x86/deflate_quick.c
@@ -209,6 +209,10 @@ static inline Pos quick_insert_string(deflate_state *const s, const Pos str) {
 ZLIB_INTERNAL block_state deflate_quick(deflate_state *s, int flush) {
     IPos hash_head;
     unsigned dist, match_len;
+    unsigned int wsize = s->w_size;
+
+    if (wsize > 8192)
+        wsize = 8192;
 
     if (s->block_open == 0) {
         static_emit_tree(s, flush);
@@ -237,7 +241,7 @@ ZLIB_INTERNAL block_state deflate_quick(deflate_state *s, int flush) {
             hash_head = quick_insert_string(s, s->strstart);
             dist = s->strstart - hash_head;
 
-            if (dist > 0 && (dist-1) < (s->w_size - 1)) {
+            if (dist > 0 && (dist-1) < (wsize - 1)) {
                 match_len = compare258(s->window + s->strstart, s->window + hash_head);
 
                 if (match_len >= MIN_MATCH) {
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1121,6 +1121,13 @@ if (ZLIB_ENABLE_TESTS)
             "-DCOMMAND=${GH_364_COMMAND}"
             -DINPUT=${CMAKE_CURRENT_SOURCE_DIR}/test/GH-364/test.bin
             -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/run-and-redirect.cmake)
+
+    set(GH_536_COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:switchlevels> 6 9744 1 91207)
+    add_test(NAME GH-536-segfault
+            COMMAND ${CMAKE_COMMAND}
+            "-DCOMMAND=${GH_536_COMMAND}"
+            -DINPUT=${CMAKE_CURRENT_SOURCE_DIR}/test/data/lcet10.txt
+            -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/run-and-redirect.cmake)
 endif()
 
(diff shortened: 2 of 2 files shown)


arch/x86/deflate_quick.c
```diff
@@ -122,7 +124,7 @@ static inline long compare258(const unsigned char *const src0, const unsigned ch
 }
 
 static const unsigned quick_len_codes[MAX_MATCH-MIN_MATCH+1];
-static const unsigned quick_dist_codes[8192];
+static const unsigned quick_dist_codes[QUICK_WINDOW_SIZE];
```
mtl1979: Completely unnecessary change...

arch/x86/deflate_quick.c
```diff
@@ -237,7 +243,7 @@ ZLIB_INTERNAL block_state deflate_quick(deflate_state *s, int flush) {
             hash_head = quick_insert_string(s, s->strstart);
             dist = s->strstart - hash_head;
 
-            if (dist > 0 && (dist-1) < (s->w_size - 1)) {
+            if (dist > 0 && (dist-1) < wsize) {
```
mtl1979: Should be ```wsize - 1```...

arch/x86/deflate_quick.c
```diff
@@ -343,7 +349,7 @@ static const unsigned quick_len_codes[MAX_MATCH-MIN_MATCH+1] = {
     0x001c230d, 0x001d230d, 0x001e230d, 0x0000a308,
 };
 
-static const unsigned quick_dist_codes[8192] = {
+static const unsigned quick_dist_codes[QUICK_WINDOW_SIZE] = {
```
mtl1979: Completely unnecessary change...

arch/x86/deflate_quick.c
nmoinvaz: I think here it should be `wsize`. I don't see the need for `-1` .  It is already a less than check.

arch/x86/deflate_quick.c
nmoinvaz: It seems proper that `quick_len_codes` depends on #define value so `quick_dist_codes` should as well. I had to create the #define for checking bounds, might as well use it where can.

arch/x86/deflate_quick.c
mtl1979: Obviously both matches need to fit in the window at the same time, otherwise the hash could not be calculated. The test is incorrect, but removing ``` - 1``` makes it less correct... Obviously ```dist``` can't be more than ```wsize - MIN_MATCH```.

arch/x86/deflate_quick.c
mtl1979: Because the array size doesn't change ever, using macro for the array size makes it prone for overflow.

arch/x86/deflate_quick.c
nmoinvaz: But then why wasn't the original check `s->w_size - 3`?

nmoinvaz: > In the worst case the original string starts at byte index 0, and match can be last 3 bytes of window (8189, 8190, 8191). I don't know why it isn't w_size - 3, that is why I said the original test is also incorrect...

I think we should only care that `dist-1` is less than 8192  or `quick_dist_codes` array size, because in `static_emit_ptr`:
```
    unsigned code2 = quick_dist_codes[dist-1] >> 8;
    unsigned len2  = quick_dist_codes[dist-1] & 0xFF;
```

mtl1979: I think we should add a comment here why it uses ```dist - 1``` in the second test... That way people don't need to scan the whole source to see the reason...

nmoinvaz: I think I found the reason which is that `hash_mask` is window size -1.

nmoinvaz:
This PR is unnecessary if #549 is accepted. But the test case can at least be salvaged.

arch/x86/deflate_quick.c
nmoinvaz: Ok I changed it for less code lines being changed.

arch/x86/deflate_quick.c
nmoinvaz: I have kept the -1.

arch/x86/deflate_quick.c
nmoinvaz: Changed.

nmoinvaz:
Simplified and rebased.

PR #542, opened 2020-02-25
Allow deflate_quick() to continue processing input data after flushing pending buffer

nmoinvaz:

This merge request relies on #541 and should be rebased after it is merged in.

I have added a unit test which covers this change. Before the change in https://github.com/nmoinvaz/zlib-ng/runs/467259498 it errors out with `deflate() did not consume 55397 bytes of input`. And after the change there is no such error https://github.com/nmoinvaz/zlib-ng/runs/467278325.

Background:

Switch levels and `pigz` check to see that all of the input buffer was consumed. `deflate_quick` was not processing all available input after flushing the pending buffer but instead was exiting early.

The relevant change is the check for `avail_in == 0` in:

```
if (s->pending + 4 >= s->pending_buf_size) {
    flush_pending(s->strm);
    if (s->strm->avail_in == 0 && flush != Z_FINISH) {
        return need_more;
    }
}
```


nmoinvaz:
Simplified and rebased.

Dead2:
Looks good to me

nmoinvaz:
Rebased.

PR #546, opened 2020-02-27
AVX2 support in memcopy.h

mtl1979:
* forces -mavx2 for files including memcopy.h because of ABI incompatibility.


nmoinvaz:
I notice you did a new CMake option. It might overlap with #525 where I added `-DWITH_AVX2`

mtl1979:
It needs to be separate from -DWITH_AVX2 as it doesn't use run-time detection and at least the MacOS CI and buildkite don't have AVX2.

nmoinvaz:
Why don't we do the same thing with SSE2 in that code?

mtl1979:
I did try to edit as little as possible the SSE2 part of the code... Any SSE2-specific changes can be done in separate pull request... The code uses both SSE2 instructions and AVX2 instructions, so AVX2 test is inside SSE2 test when there is overlap. When the files are compiled with -mavx2, compiler will take care of zeroing the upper bits of vector registers when necessary.

Dead2:
I think this PR is possibly what made the buildkite server crash a couple weeks back.

We cannot set the -mavx2 for files containing code that have to operate on cpus without that instruction set).
See what `gcc -Q --help=target` vs `gcc -mavx2 -Q --help=target` results in on an x86_64 machine:
```
--- defaults.gcc	2020-03-13 13:51:00.498966135 +0100
+++ mavx2.gcc	2020-03-13 13:51:06.231007377 +0100
@@ -26,2 +26,2 @@
-  -mavx                       		[disabled]
-  -mavx2                      		[disabled]
+  -mavx                       		[enabled]
+  -mavx2                      		[enabled]
@@ -114 +114 @@
-  -mno-sse4                   		[enabled]
+  -mno-sse4                   		[disabled]
@@ -124 +124 @@
-  -mpopcnt                    		[disabled]
+  -mpopcnt                    		[enabled]
@@ -152,4 +152,4 @@
-  -msse3                      		[disabled]
-  -msse4                      		[disabled]
-  -msse4.1                    		[disabled]
-  -msse4.2                    		[disabled]
+  -msse3                      		[enabled]
+  -msse4                      		[enabled]
+  -msse4.1                    		[enabled]
+  -msse4.2                    		[enabled]
@@ -159 +159 @@
-  -mssse3                     		[disabled]
+  -mssse3                     		[enabled]
@@ -183 +183 @@
-  -mxsave                     		[disabled]
+  -mxsave                     		[enabled]
```
Enabling the AVX2 instruction set in GCC will also enable all the preceding instruction sets, like SSE2/3/4 etc (SSE2 is missing from the list because it is already enabled by x86_64).

All code compiled with non-generic instruction sets enabled have to be hidden behind runtime instruction-set checks.
This means we cannot inline code containing instruction-set-specific intrinsics into the general code. The trick @nmoinvaz is doing with template files in PR #522 is one way to get around this limitation when it comes to reduce the amount of code duplication.

PR #547, opened 2020-03-05
Clean up zng_tr_tally code

nmoinvaz:
Now that we have `deflate_p.h` we don't need to have a #define for inlining and separate function for debugging. We can just have a single static inline function in `deflate_p.h`. I split the whole function into two, one for tallying literal and distance, similar to how the #defines were previously split up.

### DIFF
--- a/deflate_p.h
+++ b/deflate_p.h
@@ -48,6 +48,40 @@ static inline Pos insert_string_c(deflate_state *const s, const Pos str, unsigne
 }
 
 /* ===========================================================================
+ * Save the match info and tally the frequency counts. Return true if
+ * the current block must be flushed.
+ */
+
+extern const unsigned char ZLIB_INTERNAL zng_length_code[];
+extern const unsigned char ZLIB_INTERNAL zng_dist_code[];
+
+static inline int zng_tr_tally_lit(deflate_state *s, unsigned char c) {
+    /* c is the unmatched char */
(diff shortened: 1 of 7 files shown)


deflate_p.h
```diff
@@ -47,6 +47,41 @@ static inline Pos insert_string_c(deflate_state *const s, const Pos str, unsigne
...
+ * Save the match info and tally the frequency counts. Return true if
+ * the current block must be flushed.
+ */
+
+extern const unsigned char ZLIB_INTERNAL zng_length_code[];
+extern const unsigned char ZLIB_INTERNAL zng_dist_code[];
+
+static inline int zng_tr_tally_lit(deflate_state *s, unsigned c) {
+    /* c is the unmatched char */
+    unsigned char cc = (c);
```
mtl1979: Because this is private internal function, we don't need the second parameter to be "unsigned c", we can use "unsigned char c" directly...

mtl1979 (changes requested):

deflate_p.h
nmoinvaz: Fixed. Thank you.

Dead2:
Looks a lot like what I was trying to do a month or two back, but approached from the other end.
I like it.
Performance is within measurement accuracy, except for deflate_medium that seems to have gotten very slightly faster, for obvious reasons.

Issue #539, opened 2020-02-21
deflate_quick questions


nmoinvaz:
These are some questions I have about deflate_quick() that perhaps somebody (@jtkukunas ?) can answer - at least it is good to have it documented.

1. Why is deflate quick window only 8K? It seems like it would be beneficial for it to be 32K and lead to better compression rates.

2. Sometimes `dist` is greater than `s->w_size`. Is there anything that can be done to prevent it from trying to read at greater distances?
   1. Perhaps this has to do with switching the deflate level mid-stream. If it goes from level 6 (32k window size) to level 1 (8k window size) then it will still be reading 32k distances because window size doesn't change when level is updated in `deflateParams`.
3. When switching deflate level, from level 1 (8k window size) to level 6 (32k window size), since the window size never updates (nor the window buffer alloc itself), then compression performance at level 6 suffers because window size is permanently at 8k.
   1. I'm not suggesting the size of the window be reallocated or even updated, but rather that deflate_quick() support 32k window size with more lens and dist codes.
4. What method/algorithm was used in constructing the `quick_len_codes` and `quick_len_dists`? 
   1. If we had the algorithm perhaps we could have a makequick.exe that generated these automatically and is part of CI, similar to makefixed, makecrct, or maketrees..

nmoinvaz:
With 32k window size quick deflate is ~9% slower and compresses ~7% better. 
```
sslkeylogfile.log  154,465,349
sslkeylogfile8k.log.gz 96,917,999
sslkeylogfile16k.log.gz 93,809,389
sslkeylogfile32k.log.gz 90,912,856
```
To generate the quick dist codes where `QUICK_MAX_WINDOW` is 8192:
```
/* Generate quick distance codes */
for (dist = 1; dist <= QUICK_MAX_WINDOW; dist++) {
    int32_t code = d_code(dist-1);
    int32_t code_rev = bi_reverse((unsigned)code, 5);
    int32_t extra = (dist-1) - base_dist[code];
    int32_t extra_bits = extra_dbits[code];

    quick_dist_codes[dist-1] = (extra << 13) | (code_rev << 8) | (5 + extra_bits);
}
```
Pros:
* Better compression when switching compression levels
   * Level 1 with 8K window size -> level 6 with 32k window size (window not resized)
* Supports all window sizes up to 32k (still supports 8k window size for speed)

Cons:
* Requires 128k byte table for all codes
* Slower if using max window size

nmoinvaz:
[...]

PR #549, opened 2020-03-08
Unify emitting of literals and match dist/lengths.

nmoinvaz:
This PR unifies the code in the various deflate algorithms used for emitting literal and match dist/length codes. 

It also changes the `deflate_quick` algorithm so that it does not emit the match dist/length codes using pregenerated tables. This should reduce the overall compiled code size by the size of the tables which is around 33k. By removing the tables, we can also now support the deflate algorithm with window sizes up to 32k, where previously it only supported up to 8k. The default window size for the quick algorithm remains at 8k. Now, when switching between compression levels (ie: 6->1) mid-stream the quick algorithm can take advantage of the initial 32k window size. 

I have found that most of the speed that the quick deflate method has over the fast deflate method is in the use of intrinics and in the emitting of the literal and match dist/length codes without using a huffman tree.

Additionally, the `zng_emit_dist` code that was moved from `compress_block` now only requires 2 calls to `send_bits` where as previously it required up to 4. In the future if we were to support `put_uint64` we could reduce this to 1.

The calculating of `compressed_len` and `bits_sent` have their own macros to reduce the amount of `#ifdef ZLIB_DEBUG` throughout the code. 

For some background see #539.


trees.h
```diff
@@ -3,7 +3,7 @@
 
 /* header created automatically with maketrees.c */
 
-ZLIB_INTERNAL const ct_data static_ltree[L_CODES+2] = {
+const ct_data ZLIB_INTERNAL static_ltree[L_CODES+2] = {
```
Dead2: Why these changes?
I know this is technically the correct order, but GCC doesn't care, we don't do this other places and does this really belong in this PR?
My personal opinion on this is to put these flags at the start of the line for easier readability, and for some attributes even on the preceding line to avoid cluttering it up with stuff that is ultimately not interesting (like HOT, COLD, ALLOC, etc).

trees.c
```diff
@@ -589,26 +590,153 @@ static void send_all_trees(deflate_state *s, int lcodes, int dcodes, int blcodes
...
+    sent_bits_add(s, 32);
     if (stored_len)
         memcpy(s->pending_buf + s->pending, (unsigned char *)buf, stored_len);
     s->pending += stored_len;
-#ifdef ZLIB_DEBUG
-    s->compressed_len = (s->compressed_len + 3 + 7) & (unsigned long)~7L;
-    s->compressed_len += (stored_len + 4) << 3;
-    s->bits_sent += 2*16;
-    s->bits_sent += stored_len<<3;
-#endif
```
Dead2, line 609: What happens to the ZLIB_DEBUG, is its functionality maintained in this new version?

Dead2:
This looks promising, also needs a rebase now.

trees.h
nmoinvaz: I'm not so concerned about the order. I had to change `static_dtree` and `base_length` from static to non-static. And I noticed this discrepancy. This instance was the only place where it had this order. There were two other instances in the same file with the const first order.

nmoinvaz: Let me know how you want it and I can change it.

trees.c
nmoinvaz: Yes, it is functionally maintained. `compr_bits_add` and `sent_bits_add` are defined as no-op if ZLIB_DEBUG is defined.

trees.h
Dead2: Lets go with attributes-first, it does make things more readable.

Dead2:
Please make a separate PR for the trees.h and maketrees.c changes, those have no bearing on this already-huge PR.

nmoinvaz:
Ok I have backed out the two changes to trees.h and maketrees.c. There are changes to those two files which are necessary - the removal of static keyword.

nmoinvaz:
Rebased.

nmoinvaz:
Rebased. It looks like CI instance needs to be re-run.

nmoinvaz:
Rebased.

Dead2:
Another rebase needed :)

nmoinvaz:
Rebased.

Dead2:
Compression performance is ~0.56% slower than before. Is there anything that can be done to improve that?

Compressed size is identical though, so that is good.

Library size is a lot smaller, as expected:
```
  85794    1216      32   87042   15402 libz-ng.so.1
```
Before:
```
 120013    1216      32  121261   1d9ad libz-ng.so.1
```

Dead2:
Maybe add the new `ZLIB_INTERNAL` functions (+ dependencies) in trees.c to a new .h file that can included in trees.c and deflate_quick.c, allowing the functions to be declared `static inline` instead? I suspect that is where the performance went.

I have been thinking of renaming trees.h and crc32.h to tables_huffman.h and tables_crc32.h, since they only contain tables anyway and easily come in conflict with the need for actual header files. (trees_p.h for example)

nmoinvaz:
> I have been thinking of renaming trees.h and crc32.h to tables_huffman.h and tables_crc32.h..

I like this idea. It would be good for a separate PR.

nmoinvaz:
@Dead2 I have split out the code into a separate file _tree_emit.h_ that is now included in both `deflate_quick.c` and `trees.c`. I added this in a separate commit, so we can easily check to see if it makes a difference and if not it can be reverted.

nmoinvaz:
> Compression performance is ~0.56% slower than before.

Was this overall or only for deflate_quick algorithm?

Dead2:
> Was this overall or only for deflate_quick algorithm?

Overall. deflate_quick lost a little more, the rest a little less.

Dead2:
### PR/530 (Current develop)
```
 120013    1216      32  121261   1d9ad libz-ng.so.1

 1      49.603%    0.215/0.219/0.222       0.132/0.135/0.137  20804915
 2      38.062%    0.420/0.426/0.430       0.130/0.135/0.138  15964205
 3      37.021%    0.520/0.527/0.533       0.130/0.133/0.135  15527698
 4      35.832%    0.595/0.604/0.610       0.128/0.130/0.132  15028898
 5      35.719%    0.650/0.661/0.667       0.121/0.128/0.132  14981562
 6      35.651%    0.788/0.801/0.811       0.125/0.129/0.132  14953264
 7      35.030%    1.433/1.445/1.455       0.123/0.128/0.130  14692564
 8      34.995%    2.218/2.235/2.247       0.120/0.128/0.131  14677992

 avg1   37.739%                0.865                   0.131
 tot                         103.761                  15.683
```

### PR/549 without inlining
```
  85794    1216      32   87042   15402 libz-ng.so.1

 1      49.603%    0.237/0.243/0.247       0.124/0.133/0.136  20804915
 2      38.062%    0.421/0.427/0.431       0.134/0.137/0.139  15964205
 3      37.021%    0.520/0.528/0.534       0.127/0.132/0.134  15527698
 4      35.832%    0.588/0.599/0.605       0.125/0.129/0.131  15028898
 5      35.719%    0.650/0.658/0.664       0.119/0.127/0.132  14981562
 6      35.651%    0.787/0.803/0.815       0.125/0.130/0.132  14953264
 7      35.030%    1.428/1.449/1.470       0.125/0.128/0.130  14692564
 8      34.995%    2.225/2.250/2.268       0.121/0.127/0.130  14677992

 avg1   37.739%                0.869                   0.130
 tot                         104.340                  15.632
```

### PR/549 with inlining
```
  86655    1216      32   87903   1575f libz-ng.so.1

 1      49.603%    0.226/0.231/0.235       0.127/0.134/0.136  20804915
 2      38.062%    0.420/0.426/0.430       0.134/0.136/0.138  15964205
 3      37.021%    0.519/0.527/0.531       0.129/0.133/0.135  15527698
 4      35.832%    0.588/0.599/0.605       0.125/0.129/0.133  15028898
 5      35.719%    0.645/0.655/0.660       0.122/0.130/0.133  14981562
 6      35.651%    0.793/0.801/0.807       0.123/0.129/0.132  14953264
 7      35.030%    1.430/1.446/1.457       0.125/0.128/0.131  14692564
 8      34.995%    2.225/2.246/2.258       0.125/0.128/0.131  14677992

 avg1   37.739%                0.866                   0.131
 tot                         103.948                  15.706
```

Looks like most of the speed loss has been regained.
-Deflate_quick is a little slower, I'd say this is fine.
-Deflate_fast seem to be same speed as before.
-Deflate_medium seem to be overall same speed as before, a little faster sometimes perhaps.
-Deflate_slow is a little slower overall, is there anything that could still be improved here?

nmoinvaz:
The only thing I can think of is to make bi_windup and bi_flush static inline as well, but it wasn’t that way before.

trees.c
```diff
@@ -733,52 +718,24 @@ static void compress_block(deflate_state *s, const ct_data *ltree, const ct_data
     unsigned dist;      /* distance of matched string */
     int lc;             /* match length or unmatched char (if dist == 0) */
     unsigned sx = 0;    /* running index in sym_buf */
-    int code;           /* the code to send */
-    int extra;          /* number of extra bits to send */
-
-    // Temp local variables
-    uint32_t bi_valid = s->bi_valid;
```
nmoinvaz, line 740: @Dead2 perhaps this is where the speed decrease came from. The `bi_valid` and `bi_buf` variables are no longer set at the top for the loop.

nmoinvaz:
I have made an additional commit with possible performance improvement. I don't see any other area where it can be improved given the change.

Dead2:
No, doesn't seem to help at all.
```
 1      49.603%    0.222/0.227/0.230       0.129/0.134/0.137  20804915
 2      38.062%    0.421/0.428/0.432       0.133/0.137/0.139  15964205
 3      37.021%    0.519/0.529/0.536       0.129/0.133/0.136  15527698
 4      35.832%    0.591/0.597/0.601       0.124/0.129/0.133  15028898
 5      35.719%    0.638/0.651/0.659       0.126/0.130/0.132  14981562
 6      35.651%    0.784/0.796/0.805       0.127/0.130/0.132  14953264
 7      35.030%    1.444/1.460/1.474       0.122/0.129/0.131  14692564
 8      34.995%    2.240/2.255/2.273       0.120/0.127/0.132  14677992

 avg1   37.739%                0.868                   0.131
 tot                         104.156                  15.720
```
Probably better to revert that last change and we'll pull it like it was.

nmoinvaz:
I have reverted that last change. I think I can possibly speed things up in a future PR by doing 64-bit bit buffer.

PR #552, opened 2020-03-09
Fixed codecov upload from forked repositories in GitHub Actions cmake.

nmoinvaz:


nmoinvaz:
I think I am wrong about this. Secret availability check uses environment variable from `env` section.

PR #554, opened 2020-03-13
deflate_medium speedup and some cleanup

Dead2:
This is basically the leftovers after my recent experiments with tr_tally in deflate_medium.

First we have a cleanup of bflush usage, and making sure we always initialize that value before we use it.
Second is a minor optimization of deflate_medium by reducing the size of a struct by half.
Third commit is just adding a comment.

### DIFF
--- a/deflate_medium.c
+++ b/deflate_medium.c
@@ -21,25 +21,25 @@ struct match {
 };
 
 static int emit_match(deflate_state *s, struct match match) {
-    int flush = 0;
+    int bflush = 0;
 
     /* matches that are not long enough we need to emit as literals */
     if (match.match_length < MIN_MATCH) {
         while (match.match_length) {
-            flush += zng_tr_tally_lit(s, s->window[match.strstart]);
+            bflush += zng_tr_tally_lit(s, s->window[match.strstart]);
             s->lookahead--;
             match.strstart++;
             match.match_length--;
         }
-        return flush;
+        return bflush;
     }
 
     check_match(s, match.strstart, match.match_start, match.match_length);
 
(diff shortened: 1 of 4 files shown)


deflate_medium.c
```diff
@@ -189,14 +190,15 @@ static void fizzle_matches(deflate_state *s, struct match *current, struct match
 }
 
 ZLIB_INTERNAL block_state deflate_medium(deflate_state *s, int flush) {
-    struct match current_match, next_match;
+    ALIGNED_(16) struct match current_match;
+    struct match next_match;
```
mtl1979: This doesn't look right... Either neither should be aligned or both should be aligned.

nmoinvaz: Looks good to me

nmoinvaz:
Looks good to me

deflate_medium.c
mtl1979: The ```struct match``` is 8 bytes (4x2 bytes), and aligning it on 16 byte boundary is overkill.

Dead2: I aligned it on a 16-byte boundary to force it to start at a new cacheline, since it is 8 bytes it only takes half of the cacheline, leaving room for the next struct to take up the second half of the cacheline. Thus both structs live in a single cacheline.
This is how I have seen this done other places, but I understand it can look a bit confusing, so maybe a comment would help?

mtl1979: It would be a lot clearer if it would be nested struct instead of two separate structs... But it definitely needs a comment.

Dead2: Comment added. Also indented the second struct to make it easier to read and to see that they are related.

Dead2:
Codecov shows a big change in coverage, not because of this PR but because of a fix for mismatched paths in codecov config.

Issue #553, opened 2020-03-10
Illegal AVX instructions on older Xeon systems


todd-richmond:
AVX instructions were added in pull #502 but cause illegal instructions on older Xeon systems. I had to disable with -DHAVE_AVX2_INTRIN=OFF even though they were enabled by default using cmake

model name	: Intel(R) Xeon(R) CPU E5-2407 0 @ 2.20GHz
cpuid level	: 13
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon pebs bts rep_good xtopology nonstop_tsc aperfmperf eagerfpu pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 cx16 xtpr pdcm pcid dca sse4_1 sse4_2 x2apic popcnt tsc_deadline_timer aes xsave avx lahf_lm arat xsaveopt pln pts dtherm pti retpoline tpr_shadow vnmi flexpriority ept vpid

mtl1979:
Can you test what is the exact instruction that is not implemented on that Xeon?

todd-richmond:
all I get is an illegal instruction exception when running ccache and gcc but it doesn't say what the instruction id is. Do you have a way to tell?

mtl1979:
It shouldn't throw exception when running either ccache or gcc because in that stage it's not executing any zlib-ng code... When running cmake it should only throw exception in the configuration stage as that is where it tests if the current processor supports AVX2... When I test if some specific instruction causes an error, I will surround it with fprintf statements, so I see what is the last text that gets printed...  You can use fprintf to print to standard error output so it won't interfere with regular output of minigzip and the other binaries.

gdevenyi:
> ause illegal instructions on older Xeon systems.

Do you have an actual output showing exactly the error you claim?

todd-richmond:
it definitely does. We build a complete 100% local toolchain with 150 or so components including gcc and everything else - nothing from the OS libraries. When narrowing down the offending component, I determined that zlib-ng was the only component necessary to cause a failure. A git checkout from before AVS instructions works fine and disabling the new cmake define also works. 

I just deleted my toolchain to build again from scratch, but this is the outcome of the tree I was trying to build

```
[...]

PR #555, opened 2020-03-14
Apply compiler flags only to source files that require them in CMake.

nmoinvaz:
This PR will only apply compiler flags for instruction set support to the files that require them instead of globally. This fixes an issue I encountered myself during development and also possibly something that was encountered in #553.


nmoinvaz:
Codecov reporting on GH before all checks are complete.

Dead2:
As mentioned, codecov updates its comment when new data is received, so just ignore the early report. :)

Dead2:
Somehow, according to codecov, with this change we no longer have coverage for _acle files. Did some test get botched in cmake, so these files are no longer compiled/used?
I didn't see any obvious mistakes.

CMakeLists.txt
```diff
@@ -622,22 +631,24 @@ if(WITH_OPTIM)
...
         endif()
         if(WITH_ACLE)
-            list(APPEND ZLIB_ARCH_SRCS ${ARCHDIR}/crc32_acle.c ${ARCHDIR}/insert_string_acle.c)
             add_definitions(-DARM_ACLE_CRC_HASH)
+            set(ACLE_SRCS ${ARCHDIR}/crc32_acle.c ${ARCHDIR}/insert_string_acle.c)
+            list(APPEND ZLIB_ARCH_SRCS ${ACLE_SRCS})
             # For ARM aarch64, we need to check WITH_NEON first
-            if("${ARCH}" MATCHES "arm" OR NOT WITH_NEON)
-                add_intrinsics_option("${ACLEFLAG}")
+            if(${ARCH} MATCHES "arm" OR NOT WITH_NEON)
```
mtl1979: Removing quotes might be unsafe if the variable contents have non-alphanumeric characters which is possible under ARM. Basically variable expansion happens before IF condition is evaluated and it can change the order the conditions are executed.

mtl1979:
@Dead2 codecov coverage doesn't make sense at all... almost like both functions were inlined.

CMakeLists.txt
nmoinvaz: Fixed

nmoinvaz:
It seems that GCC does not globally define `__ARM_NEON` or `__ARM_FEATURE_CRC32` anymore. I don't think we really need those since we already have our own `ARM_ACLE_CRC_HASH` and `ARM_NEON_ADLER32`.

nmoinvaz:
memcpy.h probably needs to be split up... but I don't think we would be able to inline the arch-specific functions.

mtl1979:
memcpy.h definitely needs to be split... We can't inline arch-specific functions that require specific ABI as that would require compiling .c file including the functions using same ABI.

What I was thinking earlier is that we would disable the problematic optimizations by default and enable them only if making native build. This would be only required by gcc and maybe clang as Visual C++ can handle vector types as return value of a function.

nmoinvaz:
What if we just don't inline the functions, it is an extra jmp, but who cares, it makes managing the arch-specific code much easier.

mtl1979:
gcc requires all intrinsic functions to be inlined, not inlining them will be an error. This means we will have to move the problematic functions to separate .c files.

nmoinvaz:
I think we will need something like functable for inflate chunk memset. It is a lot of work..

mtl1979:
We need to make sure we don't use optimized versions for too short lengths...

Dead2:
Codecov results are completely wrong now, because this PRs commit is based on develop from a month back.
As to the comments above, yes we have a lot of cleanup to do, but I think this is a first step in the right direction.

I have several times been thinking about these intrinsics that we could perhaps duplicate the important ones using inline asm. That maintains the readability of using intrinsics, but requires that we are able to duplicate them exactly, and gives us the benefit of not using actual intrinsics but rather asm that can be inlined into code without the instruction-set flags set.

nmoinvaz:
When #525 is merged, I can revise this PR along the same lines as whatever is decided there.

nmoinvaz:
I don't think we can do this without splitting up the memcpy.c by architecture as I have done with #560.

todd-richmond:
My build was able to complete successfully after pulling this CMakeLists.txt on CentOS 6

nmoinvaz:
Closing in favor of #560.

PR #557, opened 2020-03-18
Added CMake and CI configuration for inflate compilation options

nmoinvaz:
This PR adds CMake and CI configuration for `INFLATE_STRICT` and `INFLATE_ALLOW_INVALID_DISTANCE_TOOFAR_ARRR`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,6 +90,8 @@ option(WITH_NATIVE_INSTRUCTIONS
     "Instruct the compiler to use the full instruction set on this host (gcc/clang -march=native)" OFF)
 option(WITH_MAINTAINER_WARNINGS "Build with project maintainer warnings" OFF)
 option(WITH_CODE_COVERAGE "Enable code coverage reporting" OFF)
+option(WITH_INFLATE_STRICT "Build with strict inflate distance checking" OFF)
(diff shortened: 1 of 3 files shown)


Dead2:
Same comments are valid here as in #525, these options should not be displayed by default.

nmoinvaz:
When #525 is merged, I can revise this PR along the same lines as whatever is decided there.

nmoinvaz:
Rebased and I changed these to be advanced options.

CMakeLists.txt
```diff
@@ -103,9 +103,11 @@ elseif(BASEARCH_X86_FOUND)
     option(WITH_SSE4 "Build with SSE4" ON)
     option(WITH_PCLMULQDQ "Build with PCLMULQDQ" ON)
 endif()
+option(WITH_INFLATE_STRICT "Build with strict inflate distance checking" OFF)
+option(WITH_INFLATE_ALLOW_INVALID_DIST "Build with zero fill for inflate invalid distances" OFF)
```
Dead2: These should be advanced options (and moved up over the basearch bunch)

nmoinvaz: It is already part of `mark_as_advanced`

nmoinvaz: I have moved those two lines above basearch bunch as you suggested.

PR #560, opened 2020-03-24
Split up memcopy by architecture

nmoinvaz:
This PR is the final result of my changes to break out architecture specific code into separate files.

This code moves all of the memory chunk copying code into their own files and utilizes `functable` to reference them.


nmoinvaz:
Rebased and fixed issue with configure script.

nmoinvaz:
Using @neurolabusc pigz-bench-python _c_decompress.py_ script I was able to get these performance results:
```
Method                  ms      mb/s
pigzCloudflare.exe      5167    216.72
pigzng.exe              4127    271.35
pigznmoinvaz.exe        4077    274.67
```
I ran it a few times. Sometimes ng and my branch equaled more closely, but other times it was equivalent to what is seen above. Cloudflare was always the longest.

neurolabusc:
@nmoinvaz your decompression results match up with [my findings that zlib-ng provides superior decompression to Cloudflare or baseline zlib](https://github.com/neurolabusc/pigz-bench-python). 

On the other hand, I do need to update the graphs to reflect the recent improvements in zlib-ng compression speed. I was hoping to update the repository when [pigz built with zlib-ng supports compression level 1](https://github.com/zlib-ng/zlib-ng/issues/536). That would change the lower graph in my repository, as zlib-ng uses a different strategy at level 1 than classic zlib.

nmoinvaz:
I have fixed all the pigz issues for compression level 1 - it is just a matter of getting them merged into develop by @Dead2.

arch/arm/memchunk_neon.c
```diff
@@ -0,0 +1,129 @@
+/* memchunk_neon.h -- NEON inline functions to copy small data chunks.
```
mtl1979: wrong extension

arch/x86/memchunk_avx.c
```diff
@@ -0,0 +1,56 @@
+/* memchunk_avx.h -- AVX inline functions to copy small data chunks.
+ * For conditions of distribution and use, see copyright notice in zlib.h
```
mtl1979: Extension doesn't match filename here either...

arch/x86/memchunk_sse.c
```diff
@@ -0,0 +1,56 @@
+/* memchunk_sse.h -- SSE inline functions to copy small data chunks.
```
mtl1979: Extension does not match filename here either...

memchunk.c
```diff
@@ -0,0 +1,71 @@
+/* memchunk.h -- inline functions to copy small data chunks.
```
mtl1979: Extension doesn't match filename here either.

nmoinvaz: Fixed

arch/x86/memchunk_sse.c
nmoinvaz: Fixed

arch/x86/memchunk_avx.c
nmoinvaz: Fixed

arch/arm/memchunk_neon.c
nmoinvaz: Fixed

nmoinvaz:
Rebased.

nmoinvaz:
@Dead2 being able to split out compiler flags by file is important for performance testing. Because we don't know if our compiler flags for intrinsics support is changing other parts of the executable and leading to performance increase/decreases.

Dead2:
I agree completely.
I thought this PR was not ready yet, but I'll try to get some benchmarks done soon(TM). 

I am working on setting up a dedicated machine for benchmarks, where one core is isolated from other processes, most kernel threads and interrupts. Deviation in the benchmarks is looking like it might be half of what it used to be, but I am not quite finished with it yet.

memchunk_tpl.h
```diff
@@ -0,0 +1,166 @@
...
+        len -= sz;
+    }
+
+    /* Last, deal with the case when LEN is not a multiple of SZ. */
+    if (rem)
+        memcpy(out, from, rem);
+    out += rem;
+
+    return out;
+}
```
mtl1979: End of the function might be faster if added as inlined function so compiler does not need to emit jump from each if condition above... Code size in compiled form might increase a bit though. This also reduced size and complexity of this function in source code form. Inlined function should return "out + len" so it can be called and used with "return" directly.

nmoinvaz:
@mtl1979 any ideas what might be the issue with libpng CI?
```
 Testing /home/runner/work/zlib-ng/zlib-ng/libpng/pngtest.png:
 Pass 0: rwrwrwrwrwrwrwrwrw
 Pass 1: rwrwrwrwrwrwrwrwrw
 Pass 2: rwrwrwrwrwrwrwrw
 Pass 3: rwrwrwrwrwrwrwrwrwrwrwrwrwrwrwrwrwrw
/home/runner/work/zlib-ng/zlib-ng/libpng/pngtest.png: libpng warning: bad adaptive filter value
/home/runner/work/zlib-ng/zlib-ng/libpng/pngtest.png -> pngout.png: libpng read error

 Testing /home/runner/work/zlib-ng/zlib-ng/libpng/pngtest.png: FAIL
```

arch/x86/memchunk_sse.c
```diff
@@ -0,0 +1,54 @@
...
+static inline void chunkmemset_2(uint8_t *from, memchunk_t *chunk) {
+    *chunk = _mm_set1_epi16(*(int16_t *)from);
+}
+
+static inline void chunkmemset_4(uint8_t *from, memchunk_t *chunk) {
+    *chunk = _mm_set1_epi32(*(int32_t *)from);
+}
+
+static inline void chunkmemset_8(uint8_t *from, memchunk_t *chunk) {
+    *chunk = _mm_set1_epi64(*(int64_t *)from);
```
mtl1979: Should be _mm_set1_epi64x() or ```*(__m64 *)from``` which is MMX copy (Only for x86 32-bit)...

nmoinvaz: I had that before and changed it to see if that would fix libpng but apparently it does not.

memchunk_tpl.h
nmoinvaz: I think that might be better for a future optimization.

nmoinvaz:
This LGTM bot is annoying.

nmoinvaz:
Looks like problem is with AVX chunk implementation. Perhaps the chunk size is too big idk.

mtl1979:
@nmoinvaz You can test that by changing ```__m256``` to ```__m128```. That essentially downgrades it from AVX2 to AVX.

nmoinvaz:
It seems there are no `_mm128_set` functions.. Maybe `mm_broadcastb..` but how to cast and then load and store?

nmoinvaz:
Maybe `mm_broadcastb..`

nmoinvaz:
Got it working. Appears chunk size is too big like I suspected.

nmoinvaz:
Actually we already have SSE2 which is `_m128i`.  I willl just remove AX2. It can be a separate PR if anything.

nmoinvaz:
@Dead2 this is ready for review again.

memchunk_tpl.h
```diff
@@ -0,0 +1,166 @@
...
+        from += sizeof(memchunk_t);
+        --len;
+    }
+    return out;
+}
+
+/* Behave like chunkcopy, but avoid writing beyond of legal output. */
+uint8_t* CHUNKCOPY_SAFE(uint8_t *out, uint8_t const *from, unsigned len, uint8_t *safe) {
+    if ((safe - out) < (ptrdiff_t)sizeof(memchunk_t)) {
+        if (len & 8) {
```
mtl1979: Here, len modulo memchunk_t can be more than 15 (8+4+2+1), when using 256-bit vector.

nmoinvaz: Ah you are probably right.

mtl1979: You can still make another pull request with all the suggestions not currently on this one... Makes easier to review ;)

CMakeLists.txt
```diff
@@ -627,21 +627,14 @@ if(WITH_INFLATE_ALLOW_INVALID_DIST)
...
-# An alternative version of this macro would take a file argument, and set ${flag}
-# only for that file as opposed to ${NATIVEFLAG} globally, to limit side-effect of
-# using ${flag} globally.
 #
 macro(add_intrinsics_option flag)
     if(WITH_NATIVE_INSTRUCTIONS AND NATIVEFLAG)
-        if(NOT "${CMAKE_C_FLAGS} " MATCHES ".*${NATIVEFLAG} .*")
-            set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${NATIVEFLAG}")
-        endif()
+        set_property(SOURCE ${ARGN} PROPERTY COMPILE_FLAGS ${NATIVEFLAG})
```
dankegel, line 648: Do you want APPEND here?

CMakeLists.txt
```diff
@@ -627,21 +627,14 @@ if(WITH_INFLATE_ALLOW_INVALID_DIST)
...
     else()
-        if(NOT "${CMAKE_C_FLAGS} " MATCHES ".*${flag} .*")
-            set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${flag}")
-        endif()
+        set_property(SOURCE ${ARGN} PROPERTY COMPILE_FLAGS ${flag})
```
dankegel, line 650: and here?

CMakeLists.txt
mtl1979: APPEND might result in duplicate or conflicting flags. Most compile flags are set on directory level, not source file level, so this line overrides directory level flags.

CMakeLists.txt
mtl1979: As far as I understand, this is used to test functionality of single flag or group of flags, so any old value should be ignored.

CMakeLists.txt
```diff
@@ -722,32 +711,39 @@ if(WITH_OPTIM)
...
         endif()
         if(WITH_SSE2 AND HAVE_SSE2_INTRIN)
-            add_definitions(-DX86_SSE2)
-            list(APPEND ZLIB_ARCH_SRCS ${ARCHDIR}/slide_sse.c)
+            add_definitions(-DX86_SSE2 -DX86_SSE2_MEMCHUNK)
+            set(SSE2_SRCS ${ARCHDIR}/memchunk_sse.c ${ARCHDIR}/slide_sse.c)
+            list(APPEND ZLIB_ARCH_SRCS ${SSE2_SRCS})
             if(NOT ${ARCH} MATCHES "x86_64")
-                add_intrinsics_option("${SSE2FLAG}")
+                add_intrinsics_option("${SSE2FLAG}" ${SSE2_SRCS})
```
dankegel: configure seems to build memchunk_sse.c with different options than cmake.

One uses -msse4, the other doesn't.

What's up with that?  7e8593bd may have a thinko in arch/x86/Makefile.in...?

Found with https://github.com/dankegel/zlib-ng/tree/cmake-intrinsics-identical like so:
```
  sh test/pkgcheck.sh
```

And for the gory details:
```
  sh test/pkgcheck.sh > log 2>&1
  grep 'memchunk_sse.c$' < log  | grep -v fPIC > x; head -n1 < x | tr ' ' '\012' | grep . | sort > x1; tail -n1 < x | tr ' ' '\012' | grep . | sort > x2; diff -u x1 x2
```

CMakeLists.txt
nmoinvaz: I'm unsure probably not unless there is an issue.

CMakeLists.txt
nmoinvaz: Good find! I have fixed _arch/x86/Makefile.in_ to use SSE2 flag.

Dead2:
## Tests using configure, to check for regressions
### Baseline ee28753685d62117f8ef4ef8b94ba101a5cd1e96
```
  93054    1304      32   94390   170b6 libz-ng.so.1

 1      52.643%    0.173/0.177/0.180       0.102/0.111/0.118  16560094
 2      40.045%    0.343/0.348/0.353       0.104/0.115/0.118  12597075
 3      39.033%    0.441/0.450/0.454       0.105/0.111/0.115  12278729
 4      38.024%    0.492/0.496/0.500       0.103/0.109/0.113  11961356
 5      37.949%    0.540/0.547/0.552       0.100/0.109/0.113  11937808
 6      37.902%    0.683/0.691/0.694       0.103/0.107/0.113  11923043
 7      37.128%    1.440/1.446/1.449       0.100/0.108/0.113  11679311
 8      37.095%    2.272/2.288/2.294       0.096/0.110/0.116  11668933
 9      37.089%    2.596/2.609/2.616       0.099/0.109/0.113  11667081

 avg1   39.656%                1.006                   0.110
 tot                         135.812                  14.843
```
### PR/560 Split up memcopy by architecture 0e4ec9731e78172c5a9415592d49428e8c3fe05a
```
  95248    1472      32   96752   179f0 libz-ng.so.1

 1      52.643%    0.170/0.175/0.180       0.109/0.119/0.125  16560094
 2      40.045%    0.337/0.349/0.356       0.110/0.122/0.129  12597075
 3      39.033%    0.442/0.452/0.457       0.113/0.120/0.123  12278729
 4      38.024%    0.490/0.498/0.501       0.111/0.119/0.124  11961356
 5      37.949%    0.536/0.545/0.550       0.107/0.115/0.120  11937808
 6      37.902%    0.687/0.693/0.696       0.110/0.119/0.121  11923043
 7      37.128%    1.439/1.446/1.451       0.110/0.117/0.123  11679311
 8      37.095%    2.290/2.293/2.296       0.100/0.115/0.120  11668933
 9      37.089%    2.610/2.615/2.620       0.103/0.115/0.120  11667081

 avg1   39.656%                1.007                   0.118
 tot                         135.982                  15.909
```

No changes to compressed size.
It does look like things have gotten slower, both deflate and inflate seems to have slowed down, but it is especially noticeable in inflate where it is 7.3% slower.

Dead2:
## CMAKE Tests
### Baseline ee28753685d62117f8ef4ef8b94ba101a5cd1e96
```
  92883    1304      32   94219   1700b libz-ng.so.1

 1      52.643%    0.170/0.179/0.183       0.099/0.112/0.116  16560094
 2      40.045%    0.345/0.351/0.355       0.104/0.113/0.118  12597075
 3      39.033%    0.442/0.452/0.457       0.105/0.113/0.118  12278729
 4      38.024%    0.478/0.494/0.499       0.100/0.113/0.117  11961356
 5      37.949%    0.543/0.547/0.551       0.097/0.110/0.116  11937808
 6      37.902%    0.683/0.689/0.693       0.103/0.109/0.113  11923043
 7      37.128%    1.438/1.448/1.453       0.106/0.110/0.113  11679311
 8      37.095%    2.288/2.293/2.297       0.099/0.109/0.116  11668933
 9      37.089%    2.600/2.612/2.617       0.096/0.112/0.116  11667081

 avg1   39.656%                1.007                   0.111
 tot                         135.977                  14.995
```

### PR/560 Split up memcopy by architecture 0e4ec9731e78172c5a9415592d49428e8c3fe05a
```
  95253    1472      32   96757   179f5 libz-ng.so.1

 1      52.643%    0.166/0.176/0.183       0.116/0.122/0.126  16560094
 2      40.045%    0.343/0.348/0.353       0.116/0.124/0.129  12597075
 3      39.033%    0.442/0.454/0.458       0.113/0.120/0.126  12278729
 4      38.024%    0.478/0.491/0.496       0.111/0.117/0.121  11961356
 5      37.949%    0.530/0.541/0.547       0.104/0.117/0.124  11937808
 6      37.902%    0.682/0.688/0.692       0.107/0.115/0.120  11923043
 7      37.128%    1.441/1.448/1.453       0.110/0.118/0.123  11679311
 8      37.095%    2.283/2.293/2.298       0.113/0.116/0.120  11668933
 9      37.089%    2.606/2.615/2.618       0.106/0.116/0.123  11667081

 avg1   39.656%                1.006                   0.118
 tot                         135.813                  15.984
```

Cmake sees a speedup on deflate, but inflate gets slower here too.

PS: These tests were done using `gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-39)`, on a Haswell cpu with hyperthreading, turbo and some power saving disabled.

nmoinvaz:
@Dead2, I imagine it will be slower for both deflate and inflate, due to the compiler only enabling intrinsics in the files that use them. Also there is a performance hit due to the `functable.chunk*` calls.

nmoinvaz:
I have made an additional commit which grabs the chunksize in `inflateInit` and then uses it instead of continuously calling `functable.chunksize`.

dankegel:
I ran into something odd:
Changing default optimization in configure from -O2 to -O3 (to match zlib and cmake) crashes when combined with this pull request.

i.e. https://github.com/dankegel/zlib-ng/commits/cmake-intrinsics-files-v2-plus-O3 crashes on the GCC no opt configure builders, but https://github.com/dankegel/zlib-ng/commits/bug638-default-O3 doesn't.

I can't reproduce the crash on my local machine with ubuntu 20.04 (where gcc is 9.3.0-10ubuntu2), but I can with a container running ubuntu 18.04 (where gcc is 7.5.0-3ubuntu1~18.04).  So, the recipe to reproduce the crash is:

fire up ubuntu 18.04, then:
```
git clone https://github.com/zlib-ng/zlib-ng
cd zlib-ng
git remote add dank https:///github.com/dankegel/zlib-ng
git fetch dank
git checkout cmake-intrinsics-files-v2-plus-O3
mkdir btmp
cd btmp
../configure --warn --without-optimizations --without-new-strategies
make -j3
cd test
make testshared
```

Rebuilding with -ggdb in *FLAGS and rerunning with QEMU_RUN=valgrind finds:

Process terminating with default action of signal 11 (SIGSEGV)
General Protection Fault
at storechunk (memchunk.c:54)
by chunkmemset_c (memchunk_tpl.h:143)
by zng_inflate_fast (inffast.c:276)
by zng_inflate (inflate.c:817)
by test_large_inflate (example.c:445)
by main (example.c:988)

Changing -DUNALIGNED_OK to -UUNALIGNED_OK temporarily and rebuilding rescues the test.

This is on a plain old amd64 box ( i3-4130 ).

nmoinvaz:
If you have time to track it down it would be very helpful.

dankegel:
Evidently -O3 turns on vector instructions on ubuntu 18.04, but not ubuntu 20.04;
compiling with -Wa,-adhln shows movaps used on 18.04 but not 20.04.

mtl1979:
Adding "no-sse" target attribute to that function stops it from crashing even with -O3.

dankegel:
On both ubuntu 18.04 and 20.04:
```
ubuntu@demo-ubu1804:~/src/zlib-ng/btmp$ gcc -Q --help=target | grep sse
  -mfpmath=                   		sse
  -mno-sse4                   		[enabled]
  -msse                       		[enabled]
  -msse2                      		[enabled]
  -msse2avx                   		[disabled]
  -msse3                      		[disabled]
  -msse4                      		[disabled]
  -msse4.1                    		[disabled]
  -msse4.2                    		[disabled]
  -msse4a                     		[disabled]
  -msse5                      		
  -msseregparm                		[disabled]
  -mssse3                     		[disabled]
```
so those instructions are lurking, just waiting for the optimizer to use them.

dankegel:
This commit seemed necessary here:
https://github.com/dankegel/zlib-ng/commit/1c9d7f3707c831b372442a202a3fbfd71352d67d

mtl1979:
Personally I think we should just add a comment to the function and use memcpy() always... Using SSE/SSE2 instructions on a 64-bit target is not discouraged optimization in my opinion as all processors support it.

Dead2:
SSE + SSE2 is mandatory on x86-64, so the fact that compilers utilize that fact is unsurprising.

It looks to me like the UNALIGNED_OK check is false, since the compiler can choose to use movaps there, and movaps does not support unaligned access. I find it strange that the compiler would use movaps here though, since it is meant for floating point ("Move Aligned Packed Single-Precision Floating-Point Values"). If it had use movups, it would probably have worked fine.
I suspect we would have to use compiler intrinsics to do this reliably correctly.

I wonder whether gcc using movaps there is a compiler bug, but since we specify uint64_t, I think it is within its rights to assume it is 64-bit aligned.


Btw, it is not actually true that cmake uses -O3, that is an Ubuntu configuration and does not reflect a cmake default. RHEL and even Fedora default to -O2.
So if we were to change configure to -O3, we'd have to specify a cmake default of -O3 also.

mtl1979:
SSE engine is by default in floating point mode and switching to integer mode would require extra move between SSE registers... That would obviously have performance penalty... You can see it clearly in disassembly of some functions where the result of a load to vector register is used by integer SSE instruction.

dankegel:
Re O2 vs O3: for what it's worth, https://gitlab.kitware.com/cmake/cmake shows that the upstream default is O3:
```
./Modules/Compiler/GNU.cmake:  string(APPEND CMAKE_${lang}_FLAGS_RELEASE_INIT " -O3 -DNDEBUG")
```

Whether or not we make O3 the default in configure, we need to test with O3 on, or we'll miss out on some interesting bugs...

mtl1979:
Somebody wants to make a pull request for CI run to test -O3 and to include possible fixes?

dankegel:
memcpy optimizes nicely these days, doesn't it?

dankegel:
I rebased my "compare shared libraries generated by configure and cmake" branch on top of
this one and got it to pass for amd64 and powerpc.  It's failing on arm, possibly because 
configure builds more files with architecture-specific options than cmake does; see
https://github.com/dankegel/zlib-ng/tree/cmake-intrinics-files-v2-plus-bug638
Here's a direct link to the aarch64 failure:
https://github.com/dankegel/zlib-ng/runs/791713946

Feels like that should be addressed before this pull request lands?

mtl1979:
@dankegel For functable.c failure: The different flags are:
```
-DARM_NEON_SLIDEHASH -D__USE_LARGEFILE64 -Wno-implicit-fallthrough xxx -MD -MT -MF xxxxxxxxxxxxxxxxxxxxxxx
xxxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxxxxxx xxxxxxxxxxxxxxxxxxxxxxxxx -O2 xxx xxx xxx -march=armv8-a+crc+simd
```

I added all the x's so I could reorder the flags and see which ones are equal and which ones are missing... First one is cmake, second one is configure...

Second failure in slide_neon.c is due to configure version having it disabled.

configure
```diff
@@ -1363,10 +1363,10 @@ case "${ARCH}" in
                 if test $native -eq 0; then
                     ARCH="${ARCH}+simd"
                 fi
-                CFLAGS="${CFLAGS} -DARM_NEON_ADLER32"
-                SFLAGS="${SFLAGS} -DARM_NEON_ADLER32"
-                ARCH_STATIC_OBJS="${ARCH_STATIC_OBJS} adler32_neon.o"
-                ARCH_SHARED_OBJS="${ARCH_SHARED_OBJS} adler32_neon.lo"
+                CFLAGS="${CFLAGS} -DARM_NEON_ADLER32 -DARM_NEON_MEMCHUNK"
+                SFLAGS="${SFLAGS} -DARM_NEON_ADLER32 -DARM_NEON_MEMCHUNK"
```
mtl1979: No ```-DARM_NEON_SLIDEHASH``` here?

configure
```diff
@@ -1363,10 +1363,10 @@ case "${ARCH}" in
...
+                ARCH_STATIC_OBJS="${ARCH_STATIC_OBJS} adler32_neon.o memchunk_neon.o"
+                ARCH_SHARED_OBJS="${ARCH_SHARED_OBJS} adler32_neon.lo memchunk_neon.lo"
```
mtl1979: Missing ```slide_neon.o``` and ```slide_neon.lo```

turol:
@dankegel You've probably hit an issue similar to this: https://pzemtsov.github.io/2016/11/06/bug-story-alignment-on-x86.html

TL;DR: `UNALIGNED_OK` isn't. While x86 (and AMD64) can do unaligned access in machine code the same is not true of C(++).

mtl1979:
Like I said earlier, UNALIGNED_OK only works correctly when all vector instructions are disabled... For some source files, this is what we actually want and as such higher optimization levels could be considered as user error... We could add "-mno-sse" to those files, but that would only work on AMD64/x86_64 targets... 

For 32-bit x86 processors, even 64-bit variables can cause issues as largest general-purpose register is 32 bits and MMX-optimizations could be used for 64-bit variables... Disabling both SSE and MMX would solve that though...

dankegel:
UNALIGNED_OK appears to be overloaded... it used to mean 4-byte accesses on non-4-byte-aligned addresses by the main CPU?

mtl1979:
@dankegel Now we have UNALIGNED_OK and UNALIGNED64_OK...

nmoinvaz:
Rebased against latest develop

configure
nmoinvaz: Rebased and should be fixed.

configure
nmoinvaz: Rebased and should be fixed.

dankegel:
Rebased https://github.com/dankegel/zlib-ng/tree/cmake-intrinics-files-v2-plus-bug638
on top of this, now aarch64 passes; arm-linux-gnueabihf is still failing, though,
because configure builds all source files with some arch options, but cmake doesn't.
e.g. configure builds adler32.c like this:
```
2020-06-22T18:20:47.6201024Z arm-linux-gnueabihf-gcc -march=armv8-a+crc -O2 -O3 -DNDEBUG -std=c99 -Wall -D_LARGEFILE64_SOURCE=1 -DHAVE_VISIBILITY_HIDDEN -DHAVE_VISIBILITY_INTERNAL -DHAVE_BUILTIN_CTZ -DHAVE_BUILTIN_CTZLL -DARM_CPUID -mfloat-abi=hard -DUNALIGNED_OK -DUNALIGNED64_OK -DARM_ACLE_CRC_HASH -mfpu=neon -DARM_NEON_ADLER32 -DARM_NEON_MEMCHUNK -DARM_NEON_SLIDEHASH -I/home/runner/work/zlib-ng/zlib-ng/btmp1 -I/home/runner/work/zlib-ng/zlib-ng -c -o adler32.o /home/runner/work/zlib-ng/zlib-ng/adler32.c
```
but cmake builds it like
```
2020-06-22T18:20:39.3454955Z [12/93] /usr/bin/arm-linux-gnueabihf-gcc -DARM_ACLE_CRC_HASH -DARM_CPUID -DARM_NEON_ADLER32 -DARM_NEON_MEMCHUNK -DARM_NEON_SLIDEHASH -DHAVE_BUILTIN_CTZ -DHAVE_BUILTIN_CTZLL -DHAVE_VISIBILITY_HIDDEN -DHAVE_VISIBILITY_INTERNAL -DUNALIGNED64_OK -DUNALIGNED_OK -D_LARGEFILE64_SOURCE=1 -D__USE_LARGEFILE64 -I. -I../ -Wall -Wno-implicit-fallthrough -O3 -DNDEBUG   -std=c99 -MD -MT CMakeFiles/zlibstatic.dir/adler32.c.o -MF CMakeFiles/zlibstatic.dir/adler32.c.o.d -o CMakeFiles/zlibstatic.dir/adler32.c.o   -c ../adler32.c
```

mtl1979:
Now it's somebody else's turn to fix the gnueabihf discrepancy...

nmoinvaz:
@mtl1979 any thoughts on this error? Can't seem to find arm_acle.h. Doesn't appear to be any such file available on ARM windows.

```
crc32_acle.c
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.26.28801\bin\HostX86\ARM64\cl.EXE"' : return code '0x2'
Stop.
./arch/arm\crc32_acle.c(9): fatal error C1083: Cannot open include file: 'arm_acle.h': No such file or directory
```

@Dead2 I'm not sure there is much else left to do here except keep rebasing each time develop changes..

mtl1979:
@nmoinvaz ```#include <intrin.h>```

CMakeLists.txt
mscastanho: I had the same concern as @dankegel while reading this line. As it is today, I don't see a problem since `add_intrinsics_option` is only called once for each group of source files, so nothing is being overwritten. But the name of the macro may convey the wrong idea. At least for me, without reading the macro implementation, I would think it would just append to the list of flags. Maybe rename it to `set_intrinsics_option`?

CMakeLists.txt
```diff
@@ -682,37 +675,36 @@ if(WITH_OPTIM)
...
-                ${ARCHDIR}/slide_hash_power8.c)
-            set_source_files_properties(
-                ${ZLIB_POWER8_SRCS}
-                PROPERTIES COMPILE_FLAGS ${POWER8FLAG})
             list(APPEND ZLIB_ARCH_HDRS ${ARCHDIR}/power.h)
-            list(APPEND ZLIB_ARCH_SRCS ${ARCHDIR}/power.c ${ZLIB_POWER8_SRCS})
+            list(APPEND ZLIB_ARCH_SRCS ${ARCHDIR}/power.c)
+            set(POWER8_SRCS ${ARCHDIR}/adler32_power8.c ${ARCHDIR}/slide_hash_power8.c)
+            list(APPEND ZLIB_ARCH_SRCS ${POWER8_SRCS})
+            add_intrinsics_option("${POWER8FLAG}" ${POWER8_SRCS})
```
mscastanho: OK.

nmoinvaz:
It looks like my latest commit solved it.

CMakeLists.txt
dankegel: +1 on renaming to match function.

CMakeLists.txt
nmoinvaz: Yes this is just so I can call `add_intrinsics_option` more easily.

CMakeLists.txt
nmoinvaz: Good idea! Fixed.

nmoinvaz: We can also possibly remove the macro all together when we fix the issue with the native flag in another PR..

arch/arm/chunkset_neon.c
```diff
@@ -0,0 +1,132 @@
...
+#endif
+
+static inline void loadchunk(uint8_t const *s, chunk_t *chunk) {
+    *chunk = *(chunk_t *)s;
+}
+
+static inline void storechunk(uint8_t *out, chunk_t *chunk) {
+#ifdef _MSC_VER
+    /* Cast to chunk_t pointer to avoid compiler error on MSVC ARM */
+    chunk_t *target = (chunk_t *)chunk;
```
dankegel: Sorry if you already tried this, but couldn't you unconditionally do memcpy((chunk_t *)out, chunk, sizeof(chunk_t))?

Also, shouldn't that be target = (chunk_t *)out ?  If that is indeed a thinko, that's an argument for just doing the unconditional cast...

mtl1979: You are correct that it should be ```out``` instead of ```chunk```.

dankegel: next question: why didn't the test suite catch this? ... I guess we don't have a test worker for windows arm yet.

mtl1979: QEMU only works for Linux binaries.

mtl1979: QEMU only works for Linux binaries.

dankegel: I was hoping there was a CI provider somewhere that made win10 arm support easy, but no.
(Mozilla does it, but I guess they had to roll their own a bit: https://wiki.mozilla.org/CI_Automation/windows10_aarch64 )

dankegel:
I rebased https://github.com/dankegel/zlib-ng/tree/cmake-intrinics-files-v2-plus-bug638 on top of https://github.com/zlib-ng/zlib-ng/commit/05959455249f85edc1dea09071e7d81a10f6a218.  To make things easier, it only checks amd64, ppc, and aarch64... and it's green.  I had to change configure slightly to get the file order to match.

arch/arm/chunkset_neon.c
nmoinvaz: It will only be a matter of time before GitHub supports ARM especially with the recent macOS announcements.

chunkset_tpl.h
```diff
@@ -0,0 +1,166 @@
...
+    unsigned sz = sizeof(chunk);
+    if (len < sz) {
+        do {
+            *out++ = *from++;
+            --len;
+        } while (len != 0);
+        return out;
+    }
+
+#ifdef HAVE_CHUNKMEMSET_1
```
dankegel, line 103: If the code review bot keeps saying this method is too complex, you might try another layer
of macro, so this section just becomes
```
   USE_CHUNKMEMSET(1)
   USE_CHUNKMEMSET(2)
   ...
   USE_CHUNKMEMSET(8)
```

I dunno if that would fool the complexity checker, but it might improve readability...?

nmoinvaz: It is complaining about no header guard for this header. Such as `#ifdef CHUNKSET_TPL_H_`.. It is being used as a template so header guard doesn't apply, it can be used multiple times if somebody wanted.

nmoinvaz:
All the CI checks out now..

arch/arm/chunkset_neon.c
nmoinvaz: Fixed.

dankegel:
curse you coverage uploader!!

Dead2:
Tested using a configure build on x86-64 Haswell.
```
Level 2: /tmp/deflatebench-2.tmp 135.1 MiB / 141,626,880 B
Level 3: /tmp/deflatebench-3.tmp 105.1 MiB / 110,154,240 B
Level 4: /tmp/deflatebench-4.tmp 90.0 MiB / 94,417,920 B
Level 5: /tmp/deflatebench-5.tmp 90.0 MiB / 94,417,920 B
Level 6: /tmp/deflatebench-6.tmp 75.0 MiB / 78,681,600 B
Level 7: /tmp/deflatebench-7.tmp 60.0 MiB / 62,945,280 B
Level 8: /tmp/deflatebench-8.tmp 45.0 MiB / 47,208,960 B
Level 9: /tmp/deflatebench-9.tmp 45.0 MiB / 47,208,960 B

 Runs: 20
 Levels: 1-9
 Trimworst: 10
```

## Baseline 738a6e5b07c0495b8652c77b4879145c083ea49f
```
  93246    1304      32   94582   17176 libz-ng.so.1

 1      57.881%    1.754/1.779/1.790       1.082/1.098/1.106  163950540
 2      43.903%    1.735/1.750/1.763       0.559/0.567/0.572  62178636
 3      42.510%    1.615/1.631/1.644       0.414/0.423/0.438  46826710
 4      41.490%    1.576/1.591/1.606       0.358/0.364/0.368  39174432
 5      41.215%    1.707/1.730/1.749       0.330/0.354/0.363  38913920
 6      41.038%    1.699/1.726/1.737       0.286/0.300/0.305  32289128
 7      40.784%    1.899/1.909/1.915       0.223/0.236/0.243  25671760
 8      40.707%    1.829/1.843/1.849       0.168/0.173/0.178  19217358
 9      40.696%    2.020/2.042/2.047       0.164/0.175/0.181  19212387

 avg1   43.358%                1.778                   0.410
 tot                         160.016                  36.908

 1      57.881%    1.772/1.781/1.790       1.077/1.097/1.118  163950540
 2      43.903%    1.737/1.747/1.754       0.552/0.564/0.573  62178636
 3      42.510%    1.616/1.624/1.630       0.408/0.427/0.439  46826710
 4      41.490%    1.572/1.583/1.590       0.352/0.360/0.369  39174432
 5      41.215%    1.708/1.723/1.732       0.351/0.361/0.367  38913920
 6      41.038%    1.715/1.725/1.732       0.289/0.296/0.303  32289128
 7      40.784%    1.890/1.897/1.903       0.233/0.242/0.246  25671760
 8      40.707%    1.819/1.835/1.843       0.159/0.174/0.182  19217358
 9      40.696%    2.034/2.039/2.043       0.149/0.174/0.182  19212387

 avg1   43.358%                1.773                   0.411
 tot                         159.551                  36.952
```

## PR/560 24f8fc777a6957d05284a130e48a751515ca23ad
```
  95360    1472      32   96864   17a60 libz-ng.so.1

 1      57.881%    1.776/1.793/1.800       1.114/1.142/1.158  163950540
 2      43.903%    1.733/1.747/1.755       0.570/0.590/0.600  62178636
 3      42.510%    1.608/1.628/1.638       0.435/0.453/0.462  46826710
 4      41.490%    1.568/1.585/1.593       0.369/0.377/0.381  39174432
 5      41.215%    1.708/1.730/1.740       0.369/0.379/0.383  38913920
 6      41.038%    1.705/1.723/1.731       0.299/0.307/0.315  32289128
 7      40.784%    1.887/1.896/1.903       0.234/0.245/0.252  25671760
 8      40.707%    1.826/1.835/1.842       0.179/0.186/0.188  19217358
 9      40.696%    2.024/2.033/2.039       0.178/0.185/0.188  19212387

 avg1   43.358%                1.774                   0.429
 tot                         159.701                  38.638
```

Dead2:
Cmake version of the above tests

## Baseline 738a6e5b07c0495b8652c77b4879145c083ea49f w/CMAKE
```
  93075    1304      32   94411   170cb libz-ng.so.1

 1      57.881%    1.774/1.790/1.798       1.092/1.107/1.118  163950540
 2      43.903%    1.723/1.738/1.746       0.554/0.563/0.569  62178636
 3      42.510%    1.603/1.621/1.628       0.415/0.429/0.438  46826710
 4      41.490%    1.571/1.579/1.586       0.342/0.358/0.365  39174432
 5      41.215%    1.710/1.725/1.730       0.346/0.357/0.364  38913920
 6      41.038%    1.706/1.720/1.727       0.285/0.295/0.302  32289128
 7      40.784%    1.878/1.891/1.897       0.225/0.237/0.243  25671760
 8      40.707%    1.826/1.830/1.833       0.168/0.176/0.182  19217358
 9      40.696%    2.030/2.037/2.042       0.168/0.176/0.185  19212387

 avg1   43.358%                1.770                   0.411
 tot                         159.304                  36.981
```

## PR/560 24f8fc777a6957d05284a130e48a751515ca23ad w/CMAKE
```
  95365    1472      32   96869   17a65 libz-ng.so.1

 1      57.881%    1.749/1.777/1.790       1.113/1.134/1.146  163950540
 2      43.903%    1.747/1.754/1.762       0.576/0.590/0.600  62178636
 3      42.510%    1.612/1.625/1.632       0.421/0.443/0.452  46826710
 4      41.490%    1.576/1.585/1.592       0.362/0.369/0.378  39174432
 5      41.215%    1.718/1.728/1.738       0.357/0.370/0.380  38913920
 6      41.038%    1.713/1.725/1.730       0.300/0.311/0.316  32289128
 7      40.784%    1.890/1.897/1.904       0.230/0.247/0.257  25671760
 8      40.707%    1.832/1.838/1.843       0.170/0.181/0.191  19217358
 9      40.696%    2.035/2.040/2.044       0.177/0.184/0.190  19212387

 avg1   43.358%                1.774                   0.425
 tot                         159.689                  38.286
```

Dead2:
The output library is still 5 text-bytes bigger with cmake than with configure, but that is much much closer than before.
The performance is now maybe a bit slower with cmake than before, but is now more in line with configure.

So all in all, I think this looks great, and I think we should get this merged. Any objections?

Dead2:
Oh, forgot to say that inflate is ~4.5% slower, so that is not great. But these are important fixes, so in my opinion it is worth it.

mtl1979:
@Dead2 I already said that there was one small issue that can be resolved in subsequent pull request... After that we can add AVX2-optimized versions for some functions. Those should improve performance on more recent machines...

nmoinvaz:
Rebased.

PR #561, opened 2020-03-24
Avoid unnecessary include of windows.h from zbuild.h

pps83:
zbuild.h is included from every .c file of zlib-ng, which forces every translation unit to parse all windows system includes only to be able to typedef ssize_t. This change removes windows.h include from zbuild.h and ssize_t is instead defined in-line with equivalent defines from windows.h


mtl1979:
Seems clean to me, passes all relevant CI tests.

PR #567, opened 2020-04-25
Fix ARM64 build with Visual C++.

mtl1979:
* Correct header is arm64_neon.h, not arm_neon.h
* functable.c was missing disabling of NEON check, Visual C++ does not have getauxval()

### DIFF
--- a/arch/arm/adler32_neon.c
+++ b/arch/arm/adler32_neon.c
@@ -18,7 +18,11 @@
  */
 #include "adler32_neon.h"
 #if defined(__ARM_NEON__) || defined(__ARM_NEON)
-#include <arm_neon.h>
+#ifdef _M_ARM64
+#  include <arm64_neon.h>
+#else
+#  include <arm_neon.h>
+#endif
 #include "../../adler32_p.h"
 
 static void NEON_accum32(uint32_t *s, const unsigned char *buf, size_t len) {
--- a/functable.c
+++ b/functable.c
@@ -131,7 +131,9 @@ ZLIB_INTERNAL void slide_hash_stub(deflate_state *s) {
 #  endif
         functable.slide_hash = &slide_hash_sse2;
 #elif defined(ARM_NEON_SLIDEHASH)
+#  ifndef ARM_NOCHECK_NEON
     if (arm_cpu_has_neon)
+#  endif
         functable.slide_hash = &slide_hash_neon;
 #endif
 #ifdef X86_AVX2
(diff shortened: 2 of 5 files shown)


nmoinvaz (commented):
The indentation for # should be two spaces, but in these changes it is only one space.

arch/arm/adler32_neon.c
```diff
@@ -18,7 +18,11 @@
  */
 #include "adler32_neon.h"
 #if defined(__ARM_NEON__) || defined(__ARM_NEON)
-#include <arm_neon.h>
+#ifdef _M_ARM64
+# include <arm64_neon.h>
```
nmoinvaz: For example, should be two spaces between # and include.

nmoinvaz:
Why does `ARM_NOCHECK_NEON` exist?

arch/arm/adler32_neon.c
mtl1979: Indentation in these files doesn't conform to our own guidelines yet... Whole file needs to be reformatted... I can change the 1 space indentation to 2 space indentation, but that doesn't completely fix the formatting...

mtl1979:
@nmoinvaz Just like we don't check for SSE2 on x86_64, we don't need to check for NEON on ARM64 because it's always available when compiling with Visual C++, it's mandatory requirement.

functable.c
```diff
@@ -121,7 +121,9 @@ ZLIB_INTERNAL uint32_t adler32_stub(uint32_t adler, const unsigned char *buf, si
     functable.adler32 = &adler32_c;
 
 #if (defined(__ARM_NEON__) || defined(__ARM_NEON)) && defined(ARM_NEON_ADLER32)
+# ifndef ARM_NOCHECK_NEON
```
nmoinvaz: This should be two spaces instead of one, same with the #endif.

nmoinvaz (approved):
Looks good. I am wondering if it is possible to add a CI build configuration for MSVC ARM64. What commands are using to do the build?

mtl1979:
```vcvarsamd64_arm64.bat``` and ```nmake -f win32\Makefile.a64```

Dead2:
Needs rebase

nmoinvaz:
Looks like it has been rebased.

mtl1979:
@nmoinvaz Yeah... I rebased but Codecov was giving weird errors so I waited if it would fix itself before making a new comment...

PR #570, opened 2020-04-25
Develop zlib fixes

pps83:
Here some quick fixes I made when I tried to build and use zlib-ng. Perhaps you may find some of them valuable and cherry pick what you think should to your master branch


Dead2:
There are things I want here, but not the whole thing. But I'll let this wait until some of the bigger things are merged.

nmoinvaz:
Since recent changes, "Adding missing header guard" will mess up support for multiple longest_match in compare258.c.

nmoinvaz:
@Dead2 which commits were you interested in keeping? I see possibly:

* Remove deprecated `register` keyword
* Do not use word xor (iso646 legacy)

Dead2:
Seems important:
- `Do not use word xor (iso646 legacy)` for sure, but I think a temporary variable might still be better than calculating it twice, not sure we can trust the compiler to de-duplicate that.
- `Add cmake /build dir to gitignore` 

I am also interested in these:
- `Do note use thread local storage for functable` I never saw the need for this myself, and was never entirely convinced that it is actually needed. zlib has never been thread-safe after all..
- `Fixed casting warnings`
- `Fixed wrong forward declaration for read_buf function`

Uncertain:
- `Corrected use of extern keyword: removed from functions, added to variables`
- `Removed checks for Borland compiler`
- `Clean up windows defines, use _WIN32`

The register keyword has been deprecated from C++11, but not from C99 that zlib-ng targets. And last I checked, the register hints do give a good performance boost. So I am not overly inclined to remove those. The compilers have gotten more clever, but with many variables, they sometimes make the wrong ones into registers. These hints are unfortunately still helpful. We _could_ hide the register keyword behind a define that gets nulled if C++ is detected..

Dead2:
A problem here is really there are too many entirely different changes, so it makes it difficult to get reviews and everything has to wait for everything else. Smaller PRs typically go in much faster and get more and better reviews.

nmoinvaz:
I have submitted PR for:

* `Do not use word xor (iso646 legacy)`
* `Add cmake /build dir to gitignore`
* `Remove deprecated register keyword` (changed as you discussed above)
* `Fixed casting warnings`

These are already merged:

* `Removed checks for Borland compiler` #634 
* `Clean up windows defines, use _WIN32` #566

The following are no longer necessary because the code has changed:

* `Fixed forward declaration of static data arrays and moved them`
* `Added missing header guard to match_p.h`
* `Fixed wrong forward declaration for read_buf function`

nmoinvaz:
Any thoughts on `Fix __builtin_ctzl for builds with MS compiler`? I think it is a good fix and I have always thought that extra branch adds a performance hit. But @mtl1979 might have something else to say.

mtl1979:
If we ignore ctzl on Windows, we need to ignore it on all operating systems as otherwise we get penalty hit on CPUs where the instruction exists due to the workaround not being 1:1 replacement.

nmoinvaz:
I’m not quite sure I follow. BitScanForward should work fine on its own.

mtl1979:
@nmoinvaz _BitScanForward() doesn't map to just single intrinsic... Due to way it is written, it specifically tests the input value is not zero instead of CPU doing the test as micro-op.

BTW. MSVC 2019.23 will transparently rewrite BitScanForward calls as "ctz", as it has __builtin_ctz(). This means 0 as input will return 32, not 0.

nmoinvaz:
I think all the ones we want have been merged and I think we can close this.

Dead2:
Only things I see that might still be worth discussing is these:
* `Do not use thread local storage for functable`
* `Corrected use of extern keyword: removed from functions, added to variables`

Thoughts on whether these are things we should do or not?

nmoinvaz:
* Regarding thread local storage, perhaps we can put `__thread` behind something like `#define Z_THREAD_LS` so it can easily be turned off.
* We COULD remove the extern keyword from everything. However, I am wondering if it makes sense to just move all the function declarations from functable.h into their own header files if that is the case..

Dead2:
Hmm, `Z_THREAD_LS` sounds like a great idea. Perhaps even start off with it disabled by default, and see whether anyone actually raises their hands about it.
And my vote would be skipping the extern changes, they are not really useful to us, at least not at this point.

nmoinvaz:
I have added a PR for thread local storage.

I agree perhaps `extern` keyword can be removed at a later time or better yet we can have header files for everything.

nmoinvaz:
Actually for `extern` I am now wondering if we can use `ZEXTERN`, but that seems to change for DLL building so it might not be usable to us for this.

Dead2:
Most of this has been merged in some fashion or another. Closing.

pps83:
> Actually for `extern` I am now wondering if we can use `ZEXTERN`, but that seems to change for DLL building so it might not be usable to us for this.

[extern related commit](https://github.com/zlib-ng/zlib-ng/pull/570/commits/7ff4ca69808407eac52832f58bef86701e4c9bdb) is actually a real bugfix imo. There are two parts of it:
 - removing pointless `extern` from functions (all functions are extern by default, unless declared static). `crc_table_empty` variable doesn't need to be extern. Actually, It should be changed to static instead (it's not used outside of functable.c).
 - missing extern for a const array `static_ltree`. This is a bug, no idea why such thing is even ok in C. I think it won't even work in C++

Quick google search explains exactly this:
image

pps83:
`ZEXTERN` has completely different meaning, unrelated to `extern`.

PR #571, opened 2020-04-30
Added CI configurations for MSVC ARM and ARM64

nmoinvaz:
This PR adds CI build checks for MSVC ARM and ARM64 for both nmake and cmake. 

It does not test the built output for those platforms, since I think we would need a Windows ARM machine to test them against. I have not found a way to emulate them like we normally would with cmake/qemu.

It should be rebased after #567 is merged, then we can see where it is failing after that.


Dead2:
One of the new tests fails with `functable.c(124): error C2065: 'arm_cpu_has_neon': undeclared identifier`.

nmoinvaz:
That should be resolved after #567 is merged.

nmoinvaz:
Now I am getting an error when compiling MSVC ARM.

```
memcopy.h(53): error C3905: unaligned accesses are not supported for intrinsic type '__n128'
```
However on my machine it works without a problem when using the VS2015 x86 ARM Cross Tools Command Prompt. The easiest thing to do would be to disable `INFFAST_CHUNKSIZE` on `_M_ARM`.

@mtl1979 what do you think?

mtl1979:
@nmoinvaz For some reason it is trying to compile ACLE-related files without ACLE support.

Also...

```
static inline void storechunk(unsigned char* d, inffast_chunk_t c) {
    memcpy((inffast_chunk_t *)d, &c, sizeof(c));
}
```

Visual C++ requires explicit cast with memcpy().

nmoinvaz:
The error says the issue is with unaligned access. Are you saying `d` should be cast to `uint8_t*`?

mtl1979:
@nmoinvaz Cast to ```uint8_t*``` will not work as VC++ is too smart. The error is misleading, because ```__n128``` is defined as a struct with maximum alignment, not minimum alignment... Essentially it is a compiler ambiguity that has existed from at least VS2017.

nmoinvaz:
```
memcopy.h(53): error C3905: unaligned accesses are not supported for intrinsic type '__n128'
```
And..
```
functable.obj : error LNK2019: unresolved external symbol compare258_unaligned_16 referenced in function compare258_stub
functable.obj : error LNK2019: unresolved external symbol longest_match_unaligned_16 referenced in function longest_match_stub
```

mtl1979:
That first error was already addressed... It's a compiler bug that needs workaround...

nmoinvaz:
Where was it addressed? I did a rebase so it should have it?

nmoinvaz:
Ah in the comments! :-)

nmoinvaz:
I guess I'm not quite sure what to do about it.

mtl1979:
Adding implicit cast should not break it with gcc/clang... For the two remaining errors, I can't help until I try to fetch the commit and grep the sources...

nmoinvaz:
I fixed the last two errors. Looks like the Windows arm makefiles were out of date some what. Good thing we are adding them to CI!

nmoinvaz:
@mtl1979 I tried your code above, it resulted in failed fuzzing and other CIs. `this fault was caused by a dereference of a high value address`. I'm going to try:
```
static inline void storechunk(unsigned char* d, inffast_chunk_t c) {
    memcpy((inffast_chunk_t *)d, &c, 1);
}
```
To see if that fixes it.

mtl1979:
Casting a pointer should not garble the higher bits of the address... That most likely is caused by irrelevant bug.

nmoinvaz:
The full error can be found here: https://github.com/zlib-ng/zlib-ng/runs/704073458?check_suite_focus=true. I only copied a portion of it.

nmoinvaz:
Well that didn't work.. it compiled but a lot of other CI failed.

mtl1979:
Most likely it's bug in the sanitizer, not our code... I tried disassembling the code using compiler explorer and the generated code is correct and output from running is expected even if I force unaligned access.

https://godbolt.org/z/L1jm0E

nmoinvaz:
I am going to try something like this and see if it works then:
```
static inline void storechunk(unsigned char* d, inffast_chunk_t c) {
    inffast_chunk_t *dchunk = (inffast_chunk_t *)d;
    memcpy(dchunk, &c, 1);
}
```

mtl1979:
Obviously the third parameter can't be 1, as it would copy only one byte.

nmoinvaz:
Even still looks like a bunch of CI fail.. hmm

mtl1979:
Memory addresses in the register dump are way off... This suggests the sanitizer has failed to intercept some memory allocation calls or it tries to read outside of stack...

nmoinvaz:
Looks like all tests passing now.

mtl1979:
Hopefully the next release of Visual Studio 2019 will have a fix for the optimizer bug...

PR #572, opened 2020-05-01
Simplify generic hash function using knuth's multiplicative hash

nmoinvaz:
This PR changes the `UPDATE_HASH` function which is used as a fallback when architecture specific version is not available to use knuth's multiplicative hash.

Some background AFAIK: 

* Original zlib uses a hash function that relies on a simple xor hash involving an initial byte `s->ins_h`
* Zlib-ng uses the fallback function we currently have which I can only find reference to in d306c75d3bb36cba73aec9b3b3ca378e31d1799e. But I can't find any reference to it in Intel zlib where a similar commit [exists](https://github.com/jtkukunas/zlib/commit/43a5e7618b18f2bd502025121afb1c3510464447). But we also have a fallback for this which is the original zlib hash.
* LZ4 & others uses knuth multiplicative hash function (same as this PR)

Because of recent changes, we now pass in a `uint32_t` to the hash functions, which means doing shift and multiply on each byte is extra work. 

I did some testing to see which hash function results in more collisions. I check 100,000 random integers and counted the number of collisions per hash. Then I graphed the number of collisions by how many hashes collided by that number.

Figure_2

[This](https://gist.github.com/nmoinvaz/26daed85a714e7c55cb29e61a3ee82fe) is my python script I used to test with. Assuming my script is correct, what it would indicate is that both hash functions perform about the same.

The benefit to this PR would be the knuth hash is simpler and requires fewer instructions and we can get rid of the original zlib hash.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -330,9 +330,6 @@ int ZEXPORT PREFIX(deflateInit2_)(PREFIX3(stream) *strm, int level, int method, 
 
     s->hash_size = 1 << s->hash_bits;
     s->hash_mask = s->hash_size - 1;
-#if !defined(__x86_64__) && !defined(_M_X64) && !defined(__i386) && !defined(_M_IX86)
-    s->hash_shift =  ((s->hash_bits+MIN_MATCH-1)/MIN_MATCH);
-#endif
 
 #ifdef X86_PCLMULQDQ_CRC
     window_padding = 8;
@@ -1205,7 +1202,6 @@ static void lm_init(deflate_state *s) {
     s->match_length = s->prev_length = MIN_MATCH-1;
     s->match_available = 0;
     s->match_start = 0;
-    s->ins_h = 0;
 }
 
 #ifdef ZLIB_DEBUG
@@ -1291,7 +1287,6 @@ void ZLIB_INTERNAL fill_window(deflate_state *s) {
         /* Initialize the hash value now that we have some input: */
         if (s->lookahead + s->insert >= MIN_MATCH) {
             unsigned int str = s->strstart - s->insert;
-            s->ins_h = s->window[str];
             if (str >= 1)
                 functable.quick_insert_string(s, str + 2 - MIN_MATCH);
 #if MIN_MATCH != 3
--- a/deflate.h
+++ b/deflate.h
@@ -147,19 +147,9 @@ typedef struct internal_state {
 
     Pos *head; /* Heads of the hash chains or NIL. */
 
-    unsigned int  ins_h;             /* hash index of string to be inserted */
     unsigned int  hash_size;         /* number of elements in hash table */
     unsigned int  hash_bits;         /* log2(hash_size) */
     unsigned int  hash_mask;         /* hash_size-1 */
-
-#if !defined(__x86_64__) && !defined(_M_X64) && !defined(__i386) && !defined(_M_IX86)
-    unsigned int  hash_shift;
-#endif
-    /* Number of bits by which ins_h must be shifted at each input
-     * step. It must be such that after MIN_MATCH steps, the oldest
-     * byte no longer takes part in the hash key, that is:
-     *   hash_shift * MIN_MATCH >= hash_bits
-     */
 
     long block_start;
     /* Window position at the beginning of the current output block. Gets
--- a/deflate_fast.c
+++ b/deflate_fast.c
@@ -75,7 +75,6 @@ ZLIB_INTERNAL block_state deflate_fast(deflate_state *s, int flush) {
             } else {
                 s->strstart += s->match_length;
                 s->match_length = 0;
-                s->ins_h = s->window[s->strstart];
 #if MIN_MATCH != 3
                 functable.insert_string(s, s->strstart + 2 - MIN_MATCH, MIN_MATCH - 2);
 #else
--- a/deflate_medium.c
+++ b/deflate_medium.c
@@ -86,7 +86,6 @@ static void insert_match(deflate_state *s, struct match match) {
     } else {
         match.strstart += match.match_length;
         match.match_length = 0;
-        s->ins_h = s->window[match.strstart];
         if (match.strstart >= (MIN_MATCH - 2))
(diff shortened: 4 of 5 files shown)


mtl1979:
I'm pretty sure the Python script has a bug... See my comment in the gist...

Dead2:
@nmoinvaz Could you check out fibonacci hashing as well?

There are two 2-3 line code snippets on the following page, the 2-line is a bit naive, but the 3-line could possibly be better for our case. Both are quite fast.
https://probablydance.com/2018/06/16/fibonacci-hashing-the-optimization-that-the-world-forgot-or-a-better-alternative-to-integer-modulo/


I was also thinking we should check our usage of the following variables in the deflate struct.
Do we currently respect hash_size in all our different hashing?
Should we just set a static size using defines and remove those from the struct, or do we care enough about our memory usage to keep them? Some of the other popular compression libraries just use a static size that can be tweaked with defines, simplifying the code and probably increasing speed slightly.
```
    unsigned int  hash_size;         /* number of elements in hash table */
    unsigned int  hash_bits;         /* log2(hash_size) */
    unsigned int  hash_mask;         /* hash_size-1 */
```

nmoinvaz:
> Do we currently respect hash_size in all our different hashing?

From what I can tell it is respected - being the number of elements in the hash table. So when `memLevel == 8` then `hash_bits = 15` and `hash_size == 32k-1` and `hash_mask` is used to mod % hash values within `hash_size` range.

> Should we just set a static size using defines and remove those from the struct, or do we care enough about our memory usage to keep them?

It's not a bad idea. However, I think if we do that then `memLevel` is only used by `s->lit_bufsize`. Does `memLevel` still have value at that point? Might there be low memory applications that might still want to use it? I wonder if `memLevel` can currently be used to make hash table 64k by specifying `memLevel = 9`.

Dead2:
No, memlevel looses its meaning from that, and that is a very good point.
If we do that, then we'd probably want to remove memlevel from the init functions for the zlib-ng native api (and before the first release), and just ignore it in the zlib-compat api.
We should probably start documenting API differences too, to help porting efforts.

Honestly, the memlevel has never made sense to me other than for embedded or really old machines (286-era perhaps). Neither are platforms we care about.
We could still leave it as a define that someone who really wants to mess with it, for whatever reason, can change.
The only reason I can think to keep it small is to attempt to keep the whole of zlib hot code and data in L1 or L2 cache, but L1 cache is unlikely since it is very very small. L2 is often big enough on x86-64 cpus, but ARM and such could potentially gain some speed from keeping everything in cache. But there are already other compression algorithms more suited for that kind of use already. So I don't really think we should worry too much about that

I vote remove it. The side benefit of reducing the size of that struct is also good, as it improves cache-locality for the struct.

Dead2:
Thanks for testing fibonacci.
Didn't search that page for "knuth" before, but I did now, and look what I found:
`First of all how do we find out what this Fibonacci Hashing is? Rich Geldreich called it “Knuth’s multiplicative method,”`
So I guess the differences should be pretty minor.  Did you try to measure speed for each of the methods btw?

Dead2:
@nmoinvaz Did you try the second (improved) fibonacci hash he listed?

He writes:

"Going back to Fibonacci hashing, there is actually one simple change you can make to improve the bad pattern for the top bits: Shift the top bits down and xor them once. So the code changes to this:"
```
size_t index_for_hash(size_t hash) {
    hash ^= hash >> shift_amount;
    return (11400714819323198485llu * hash) >> shift_amount;
}
```

That same trick might possibly also be useful to us.

nmoinvaz:
Rebased

nmoinvaz:
@Dead2 I have run benchmark tests using Google benchmark and [here](https://gist.github.com/nmoinvaz/e0931461822440a5ab7aae077fa60889) are the results.

nmoinvaz:
I have updated my GitHub [gist](https://gist.github.com/nmoinvaz/26daed85a714e7c55cb29e61a3ee82fe) with all the latest results. I think Knuth is probably the best performer. This hash would only used when crc32 intrinics are not available.

Dead2:
This makes compression a little slower on my machine (x86_64), but since that codepath only gets removed code, it should be faster. So I attribute the slowdown to the changes in cache locality.

```
# HEAD
  91282    1280      32   92594   169b2 libz-ng.so.1

 1      49.603%    0.221/0.226/0.228       0.129/0.133/0.135  20804915
 2      38.062%    0.429/0.433/0.436       0.130/0.134/0.136  15964205
 3      37.021%    0.524/0.534/0.541       0.122/0.130/0.133  15527698
 4      35.832%    0.583/0.594/0.599       0.126/0.129/0.131  15028898
 5      35.719%    0.641/0.649/0.654       0.126/0.129/0.132  14981562
 6      35.651%    0.781/0.794/0.803       0.124/0.129/0.132  14953264
 7      35.030%    1.434/1.458/1.475       0.119/0.126/0.129  14692564
 8      34.995%    2.224/2.248/2.264       0.124/0.128/0.130  14677992

 avg1   37.739%                0.867                   0.130
 tot                         104.038                  15.565
```

```
# PR/572 688770a6773ea9a3d53334da7d20dac7886db650
  91218    1280      32   92530   16972 libz-ng.so.1

 1      49.603%    0.221/0.224/0.226       0.126/0.132/0.135  20804915
 2      38.062%    0.422/0.431/0.435       0.129/0.133/0.136  15964205
 3      37.021%    0.524/0.535/0.538       0.129/0.131/0.132  15527698
 4      35.832%    0.592/0.601/0.607       0.126/0.129/0.131  15028898
 5      35.719%    0.640/0.652/0.659       0.119/0.128/0.131  14981562
 6      35.651%    0.784/0.796/0.805       0.123/0.128/0.130  14953264
 7      35.030%    1.439/1.460/1.477       0.124/0.128/0.130  14692564
 8      34.995%    2.227/2.250/2.276       0.125/0.128/0.131  14677992

 avg1   37.739%                0.869                   0.130
 tot                         104.249                  15.548
```

PR #573, opened 2020-05-02
Added CI test for dual linking zlib-ng with system zlib.

nmoinvaz:


nmoinvaz:
Increase in code coverage must be due to PR #557

Dead2:
Needs rebase

nmoinvaz:
Rebased

CMakeLists.txt
```diff
@@ -81,6 +81,7 @@ check_include_file(sys/sdt.h   HAVE_SYS_SDT_H)
 option(WITH_GZFILEOP "Compile with support for gzFile related functions" OFF)
 option(ZLIB_COMPAT "Compile with zlib compatible API" OFF)
 option(ZLIB_ENABLE_TESTS "Build test binaries" ON)
+option(ZLIB_DUAL_LINK "Dual link tests against system zlib" OFF)
 option(WITH_SANITIZERS "Build with address sanitizer and all supported sanitizers other than memory sanitizer" OFF)
 option(WITH_MSAN "Build with memory sanitizer" OFF)
 option(WITH_FUZZERS "Build test/fuzz" OFF)
```
Dead2, line 87: I am thinking we should probably have a new PR rename the tests.
What do you think about TEST_DUAL_LINK and TEST_MSAN etc? Should make things more readable and obvious.

nmoinvaz: I would be in favor of that. Would you rename `ZLIB_ENABLE_TESTS` too?

nmoinvaz: For minizip I use something similar to:

`BUILD_TESTS` (for minigzip, example)
`BUILD_UNIT_TESTS` for (ctest)
`BUILD_FUZZ_TESTS` for (fuzzers)

https://github.com/nmoinvaz/minizip/blob/b39f7a0e896615833a5af9dc7134563917a3a227/CMakeLists.txt#L24

Dead2: TESTS_ENABLE? I'd support that. Although, do we actually need a master switch for tests if we have individual switches for different tests?

I'll vote for TEST_ as a prefix, because we have many test switches and BUILD_ could easily be used(confused?) instead of WITH_.
TEST_ prefix should also group the tests in the ccache list.

With TEST_ prefix, it is also a bit more human readable like "test sanitizers" does just that, although the master switch does kind of mess with that since you can enable the tests without them actually getting tested :)

mtl1979: Would make sense for the value of master switch to be default for individual test switches... Then one could easily enable or disable just one test.

Dead2: Problem is we have tests that are mutually exclusive too. MSAN and SANITIZERS don't mix, unfortunately.

mtl1979: That means we have to exclude those two and only allow them to be enabled individually.

PR #576, opened 2020-05-05
Make DYNAMIC_CRC_TABLE configurable in CMake

nmoinvaz:


nmoinvaz:
Actually it looks like this `DYNAMIC_CRC_TABLE` define no longer works correctly.

Dead2:
I am not sure that is something we want to support either, is there any reason to except for saving flash space in embedded?

nmoinvaz:
We used to use it at work to keep down executable size. But there are some benefits to not using it, namely you don't have to worry about calling initialization function to create the tables. I would be in favor of removing what is left of it (since it is broken now). Also having a separate tool that generates the crc tables  (makecrct) makes for nice separation of code.

Dead2:
Yeah, lets remove it. The code is complex enough without having complexity that is not really needed.

PR #578, opened 2020-05-05
Remove several NOT_TWEAK_COMPILER checks.

Dead2:
Gets rid of several instances of legacy code that is likely to bit-rot.

The compiled code (without the define set) should be identical, except for two cases where the code is no longer in local scope code blocks. As far as I can see, there is no reason to have those in such blocks. One of them has local variables, and that used to be a valid reason, but C99 that we target removed the requirement that variables have to be defined before code (not that many modern compilers cared, except for printing warnings about standards compliance).

Benchmarks show that these changes does not change performance.
To be precise, my benchmarks show an overall 0.0538% speedup, and that could potentially be because the compiler can optimize those blocks of code (and the before and after code) better, but that is well within the margin of error of my measurements, so it is hard to tell without analyzing the generated assembly code.

The text-size of the library remains the same as before.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -223,8 +223,8 @@ ZLIB_INTERNAL void slide_hash_c(deflate_state *s) {
             *q++ = (Pos)(m >= t ? m-t: NIL);
         }
     }
-
 #endif /* NOT_TWEAK_COMPILER */
+
     n = wsize;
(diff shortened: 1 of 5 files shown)


nmoinvaz (approved):
These are some good changes. Makes the code more readable.

PR #581, opened 2020-05-12
Move sanitizer detection to its own cmake file

nmoinvaz:


nmoinvaz:
Codecov is experiencing partially degraded service at this time.

PR #583, opened 2020-05-14
Check distance in deflate_quick using MAX_DIST(s)

nmoinvaz:
Added check_match to validate matches when using deflate_quick in debug mode.

This issue fixes some problems found in #536.

### DIFF
--- a/arch/x86/deflate_quick.c
+++ b/arch/x86/deflate_quick.c
@@ -64,7 +64,7 @@ ZLIB_INTERNAL block_state deflate_quick(deflate_state *s, int flush) {
             hash_head = functable.quick_insert_string(s, s->strstart);
             dist = s->strstart - hash_head;
 
-            if (dist > 0 && (dist-1) < (s->w_size - 1)) {
+            if (dist > 0 && dist < MAX_DIST(s)) {
                 match_len = functable.compare258(s->window + s->strstart, s->window + hash_head);
 
                 if (match_len >= MIN_MATCH) {
@@ -73,6 +73,8 @@ ZLIB_INTERNAL block_state deflate_quick(deflate_state *s, int flush) {
 
                     if (match_len > MAX_MATCH)
                         match_len = MAX_MATCH;
+
+                    check_match(s, s->strstart, hash_head, match_len);
 
                     zng_tr_emit_dist(s, static_ltree, static_dtree, match_len - MIN_MATCH, dist);
                     s->lookahead -= match_len;


nmoinvaz:
The same check for `dist < MAX_DIST(s)` can be found in:
https://github.com/zlib-ng/zlib-ng/blob/06173b6a49ae4a331758e009885e1aa1829cdd19/deflate_slow.c#L60

https://github.com/zlib-ng/zlib-ng/blob/06173b6a49ae4a331758e009885e1aa1829cdd19/deflate_medium.c#L218

https://github.com/zlib-ng/zlib-ng/blob/06173b6a49ae4a331758e009885e1aa1829cdd19/deflate_fast.c#L50

nmoinvaz:
Rebased.

Dead2:
This reduces compression a little, is that correct?
In my test it goes from 49.603% to 49.642%

nmoinvaz:
I imagine that is correct because allowable distances is lessened. But the code is the same as the other deflate algorithms.

nmoinvaz:
I can double check it.

Dead2:
The comments on MAX_DIST says it is limited in order to simplify code, especially with regards to 16-bit platforms.
This makes me think we should in theory be able to increase MAX_DIST, *unless* inflate would also need to be changed (something we don't have control over). But I suspect that inflate would work fine, since the window size is part of the spec after all. With that in mind, I wonder why this would be needed, what part of the code does not handle this properly?

nmoinvaz:
I had run into an error in pigz. Let me see if I can track it down again. #536

nmoinvaz:
Here is the error I get:
```
pigzCloudflare.exe: skipping: ./temp\pigzng.exe1_asl16.nii.gz: corrupted -- invalid deflate data (invalid distance code)
```
I also tried `dist < wsize - 1` to see if it was an off by one error. Fails in inflate:

image

Dead2:
Ah, so the same kind of thing is used in inflate as well, that is unfortunate (for deflate-algo in general).
Then I agree :)

nmoinvaz:
image

nmoinvaz:
Fails when dist is 89.

nmoinvaz:
When I use _deflate_medium_ with `MAX_DIST = wsize` it works. 

I don't think it is bad that this was merged, because it adds the match check and makes all the code similar for if we do want to change `MAX_DIST`. But I am going to try and investigate it more.

Dead2:
Yes, perhaps add an issue for investigating whether this can be improved upon?
And I agree that it is better that all the algorithms are as similar as possible, and perhaps we can improve compression for all of them later if we find a way to do so.

nmoinvaz:
Actually I think I might have been wrong in my last comment. _deflate_medium_ doesn't hit the same issue, but it might given the right input. MAX_DIST(s) is the `wsize - 262 (MIN_LOOKAHEAD)`. So if anything we save 262 bytes? Not much. I think we should just keep using `MAX_DIST` here.

nmoinvaz:
Because _deflate_fast_ is closer to _deflate_quick_, if I change it to ` if (hash_head != NIL && s->strstart - hash_head < s->w_size - 1) {`, then it suffers the same fate. So it doesn't look to be related to _deflate_quick_.

Dead2:
Well, if in those 262 bytes there is a match for a big string starting at the current position, it could save up to 258 bytes for each time that occurs. Those files might be rare, IDK.

PR #584, opened 2020-05-15
Set binary mode for stdin and stdout in switchlevels.

nmoinvaz:
I have added an additional commit for formatting changes.

### DIFF
--- a/test/switchlevels.c
+++ b/test/switchlevels.c
@@ -22,7 +22,8 @@
 #endif
 
 static int read_all(unsigned char *buf, size_t size) {
-    for (size_t total_read = 0; total_read < size;) {
+    size_t total_read = 0;
+    while (total_read < size) {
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Codecov is messed up in GitHub comments. For the last few PRs it has shown -5% which are all based on latest develop. Then you go into view the full report and it looks ok.

Dead2:
Is there any special reason you move the variables to the top?
We have decided that we target C99, and C99 has removed that requirement (it also allows // comments btw).

nmoinvaz:
No special reason I just thought it looked more readable. I work on other projects that are started pre-C99 so I guess it is muscle memory. I can revert those changes if you want.

Dead2:
I don't especially care either way. 
Could you rebase this to not depend on another PR?

nmoinvaz:
Ok should be rebased now. I don't think this was dependent upon another PR.

PR #585, opened 2020-05-16
Fixed zero length stored block left open when using Z_SYNC_FLUSH.

nmoinvaz:
This fixes an issue found from running tests in #536. Inflate would error with `minigzip: lcet10x.txt.gz: invalid stored block lengths`. I have included ctests unit tests using _switchlevels_. This PR uses commits from #583 and #584 which are necessary for proper testing of the initial issue.

Background:

_Z_SYNC_FLUSH_ writes a stored block of zero length to help inflate - with no end of block code. What was previously happening was this zero length block was being written and the _block_open_ was left set to 1 by _zng_tr_emit_tree_ - so when the stream was switched to _deflate_quick_ (when used in switchlevels) it would not emit the beginning of the block when it started.

### DIFF
--- a/trees.c
+++ b/trees.c
@@ -597,11 +597,12 @@ void ZLIB_INTERNAL zng_tr_stored_block(deflate_state *s, char *buf, unsigned lon
     put_short(s, (uint16_t)~stored_len);
     cmpr_bits_add(s, 32);
     sent_bits_add(s, 32);
-    if (stored_len)
+    if (stored_len) {
         memcpy(s->pending_buf + s->pending, (unsigned char *)buf, stored_len);
-    s->pending += stored_len;
-    cmpr_bits_add(s, stored_len << 3);
-    sent_bits_add(s, stored_len << 3);
+        s->pending += stored_len;
+        cmpr_bits_add(s, stored_len << 3);
+        sent_bits_add(s, stored_len << 3);
+    }
 }
 
 /* ===========================================================================
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1185,6 +1185,22 @@ if (ZLIB_ENABLE_TESTS)
             "-DCOMMAND=${GH_536_COMMAND}"
             -DINPUT=${CMAKE_CURRENT_SOURCE_DIR}/test/data/lcet10.txt
             -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/run-and-redirect.cmake)
+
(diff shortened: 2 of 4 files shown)


trees.c
```diff
@@ -599,6 +599,8 @@ void ZLIB_INTERNAL zng_tr_stored_block(deflate_state *s, char *buf, unsigned lon
     sent_bits_add(s, 32);
     if (stored_len)
         memcpy(s->pending_buf + s->pending, (unsigned char *)buf, stored_len);
+    else
+        s->block_open = 0;
     s->pending += stored_len;
```
mtl1979: If stored_len is 0 above, this line is no-op, maybe should be moved inside if after memcpy().

nmoinvaz: That is intentional. It should only close the block opened by _zng_tr_emit_tree_ if the length is zero. Apparently _Z_SYNC_FLUSH_ writes a stored block of zero length to help inflate. What was previously happening was this zero length block was being written, however since the block was still opened, _deflate_quick_  (when used in switchlevels) would not emit the beginning of the block when it started deflate.

nmoinvaz: It was only an issue that would be seen in conjunction with `block_open` which uses the flag.

nmoinvaz: However, next three lines after the if else statement can be moved into first conditional of if statement.

mtl1979: I meant the line "s->pending += stored_len;" which was last line before the comment... If lines further down should also moved inside the first part of "if" block, then you can do it before I review again...

nmoinvaz: I can also move the setting of _block_open_ to 1 or 0 back to only _deflate_quick_. What do you think?

mtl1979: Considering deflate_stored does some weird stuff with empty blocks, I rather see all the modifications in deflate_quick.

nmoinvaz: I made the changes. Thanks for the reivew. I think this looks better now.

nmoinvaz:
I have moved toggling of _block_open_ back to _deflate_quick_ based on review since there is this weird case with zero length stored block with no end of block code.

nmoinvaz:
Rebased

Dead2:
Could you rebase this to not depend on other PRs?
Cherry-picked and tested just the last commit in this PR, and it does not affect compressed size, but it seems to result in a slight speedup of non-deflate_quick algorithms. (~0.082% speedup overall)

nmoinvaz:
I can probably only remove the first commit. The other one about "binary mode for stdint and stdout in switchlevels" is necessary for the unit test in this.

Dead2:
Ah, ofc. Wait until that one is merged then, I am waiting for it to pass CI.

Dead2:
Rebase please :)

nmoinvaz:
Rebased.

Issue #515, opened 2020-02-07
Benchmark/compare -O vs -O2 vs -O3 on a variety of platforms.


gdevenyi:
Following up on #507, Fedora defaults cmake to -O2, while upstream cmake defaults to -O3. Testing by @Dead2 noted that there are indeed performance differences between the settings and it is not currently clear which is optimal.

nmoinvaz:
Do we still need to do this since everything has been harmonized?

Dead2:
Probably not necessary anymore

PR #588, opened 2020-05-23
Add Adler32 SSSE3 and AVX2 variants

nmoinvaz:
These implementations are based on work on a Cloudflare zlib fork by Brian Bockelman @bbockelm.

- https://github.com/bbockelm/zlib/tree/adler32_vectorized
- https://sft.its.cern.ch/jira/browse/ROOT-8465
- https://indico.cern.ch/event/754335/contributions/3166227/attachments/1733712/2803236/ROOT_I_O_compression_algorithms-DIANA1.pdf

The SSSE3 detection code for CMake and configure is from https://github.com/mtl1979/zlib-ng/commit/7d2b9e9555c1271999027b6e7d2372e255c08ea8.  

I ran benchmarks against the Adler32 variants using Google benchmark:

image

https://gist.github.com/nmoinvaz/1bf9e4098fe6399a14f829957c2123f8


nmoinvaz:
I know that the code for both variants need to be cleaned up. I am looking for reviewers to help decide the best way to do that.

nmoinvaz:
Codefactor is quite annoying in configure script..

mtl1979:
I found quite a lot of possible optimizations in my version... If I drop support for SSE3 then the code gets a lot cleaner...

nmoinvaz:
Rebased.

CMakeLists.txt
```diff
@@ -148,6 +149,7 @@ if(${CMAKE_C_COMPILER} MATCHES "icc" OR ${CMAKE_C_COMPILER} MATCHES "icpc" OR ${
         if(BASEARCH_X86_FOUND)
             set(AVX2FLAG "-mavx2")
             set(SSE2FLAG "-msse2")
+            set(SSE3FLAG "-msse3")
```
mtl1979: Shouldn't this be ssse3?

nmoinvaz: Good catch. I must have copied the wrong line. I have fixed it.

arch/x86/adler32_avx.c
```diff
@@ -0,0 +1,103 @@
...
+    if (UNLIKELY(buf == NULL))
+        return 1L;
+
+    /* in case short lengths are provided, keep it somewhat fast */
+    if (UNLIKELY(len < 16))
+        return adler32_len_16(adler, buf, len, sum2);
+
+    uint32_t ALIGNED_(32) s1[8], s2[8];
+    memset(s1, '\0', sizeof(uint32_t)*7); s1[7] = adler; // TODO: would a masked load be faster?
+    memset(s2, '\0', sizeof(uint32_t)*7); s2[7] = sum2;
```
mtl1979: Because these are aligned, writing 8 zeroes should be faster than writing 7 zeroes. this is same as xor with itself to clear register contents.

nmoinvaz: It is zeroing 28 bytes (4 * 7). Would 32 bytes be faster?

mtl1979: 32 bytes is xor + store, just two instructions. Clearing 28 bytes requires multiple smaller stores.

nmoinvaz: Ok fixed. Thanks!

nmoinvaz:
Two test runs using @neurolabusc's pigz-bench-python:

image
image

neurolabusc:
@nmoinvaz this is very exciting. This should also accelerate decompression, which you can test with my `c_decompress.py`. Since the Adler32 computation is identical for compression and decompression, and decompression is faster (though single threaded) I suspect the proportional speedup will be even clearer for decompression. I would be interested to see what @Dead2 's single-threaded benchmark shows for this PR.

neurolabusc:
@nmoinvaz does the .gz format use [Adler32 or only CRC32](https://stackoverflow.com/questions/388595/why-use-deflate-instead-of-gzip-for-text-files-served-by-apache)? I know that Adler32 is used by the .z format (compress), but not sure it is used in .gz (gzip). If gzip does not employ Adler32, you will need to generate .z format files. With pigz, you would need to add `-z` during compression: 
```
  -z, --zlib           Compress to zlib (.zz) instead of gzip format
```
and @Dead2 would need to modify minigzip to support zlib format to see a benefit in his single-threaded suite.

nmoinvaz:
Looks like gzip format uses CRC32 in footer. The body of gzip format is DEFLATE. Zlib stream uses Adler32 in footer. The body of Zlib stream is also DEFLATE. So the speed may only improve when using Zlib stream. To produce zlib stream `windowBits` must be 15 or `MAX_WBITS`.

So you may be right that I was not producing zlib stream in my initial results.

Here are the results for `c_decompress` across multiple runs with `-z` argument:

```
Method                       ms       mb/s
pigzCloudflare.exe           6619   207.46
pigzng.exe                   5750   238.80
pigznmoinvaz.exe             5789   237.20

Method                       ms       mb/s
pigzCloudflare.exe           6591   208.33
pigzng.exe                   5891   233.10
pigznmoinvaz.exe             5685   241.52

Method                       ms       mb/s
pigzCloudflare.exe           6235   220.24
pigzng.exe                   5889   233.18
pigznmoinvaz.exe             5612   244.68
```

On a side note it appears zlib is calling `adler32` function just to set initial hash value which is always 1 when we call `functable.adler32(0L, NULL, 0);` in multiple places. We can easily just set it to `1` or in our `adler32_stub` return 1 when buf is NULL.

Dead2:
@nmoinvaz Such a "return 1 when buf is NULL" should probably be in a define to avoid confusion, but sounds like a good idea.

nmoinvaz:
You mean like `#define ADLER32_INITIAL_VALUE (1)` and then `strm->adler32 = ADLER32_INITIAL_VALUE;`?

Dead2:
Yes, exactly like that. Better name than the one I was thinking of too (ADLER32_HASH_EMPTY).

nmoinvaz:
Here is a test run using `-z` command line arg:
image

nmoinvaz:
It is really hard to tell since results are not consistent (at least not on my machine). For example, here I copied pigzCloudflare and had it run at the beginning and end and results are a bit strange. @neurolabusc

image

@Dead2's results may more consistent results, but `minigzip` would need to be modified to produce zlib format like what was mention earlier by @neurolabusc.

neurolabusc:
@nmoinvaz I think that parallel workloads are often a bit less stable. I created a [single threaded](https://github.com/neurolabusc/zlib-bench-python) benchmark that you should provide more reliable results. It uses the [Silesia corpus](http://sun.aei.polsl.pl/~sdeor/index.php?page=silesia) - which appears to show odd behavior at the extreme compression levels. You can easily adapt this to compile your custom versions of zlib-ng. Unfortunately, it does not test the Cloudflare zlib, as the minigzip in that distribution does not support the `-k` option. Regardless, I think you have now caught up to CloudFlare and going forward you will want to compare different variants of zlib-ng.

nmoinvaz:
Nice work @neurolabusc! I have made some comments on the initial commit of your benchmark that might be worth taking a look at. You should also be able to test level 1 with `minigzip` without a problem. It is only with pigz that I have noticed `deflate_quick` has a problem..

Here are my results (using *develop* branch):

image

```
CompressMethod  Level   Min     Mean    Max     mb/s    %
gzip.exe        1       4414    4619    4824    48      36.50
gzip.exe        2       4654    4895    5135    46      35.44
gzip.exe        3       5301    5311    5321    40      34.46
gzip.exe        4       5557    5570    5582    38      33.52
gzip.exe        5       6794    6797    6801    31      32.62
gzip.exe        6       8992    9004    9016    24      32.19
gzip.exe        7       10554   10574   10594   20      32.05
gzip.exe        8       15777   15852   15927   13      31.93
gzip.exe        9       20301   20404   20508   10      31.91
CompressMethod  Level   Min     Mean    Max     mb/s    %
minigzgcc.exe   1       2819    2839    2859    75      47.70
minigzgcc.exe   2       3616    3623    3630    59      35.54
minigzgcc.exe   3       3947    3963    3979    54      34.22
minigzgcc.exe   4       4289    4293    4298    49      32.95
minigzgcc.exe   5       4624    4735    4846    46      32.67
minigzgcc.exe   6       5218    5362    5506    41      32.51
minigzgcc.exe   7       7287    7323    7360    29      33.61
minigzgcc.exe   8       10168   10197   10227   21      33.52
minigzgcc.exe   9       12692   12778   12864   17      33.51
CompressMethod  Level   Min     Mean    Max     mb/s    %
minigzmsvc.exe  1       1829    1840    1851    116     47.70
minigzmsvc.exe  2       2581    2586    2591    82      35.54
minigzmsvc.exe  3       3010    3017    3025    70      34.22
minigzmsvc.exe  4       3754    3762    3771    56      32.95
minigzmsvc.exe  5       3758    3776    3794    56      32.67
minigzmsvc.exe  6       4324    4329    4333    49      32.51
minigzmsvc.exe  7       5820    6222    6624    36      33.61
minigzmsvc.exe  8       8446    8464    8483    25      33.52
minigzmsvc.exe  9       10955   11001   11047   19      33.51
```

I am using [gzip for Windows](http://gnuwin32.sourceforge.net/packages/gzip.htm). The single threaded benchmark does look more reliable.

neurolabusc:
@nmoinvaz the poor showing of gcc relative to msvc is surprising. I updated my scripts with your comments and added an optional test of CloudFlare. Surprisingly, on my Linux system zlib-ng performs better with Clang than GCC (mirroring your findings) while the reverse is true for CloudFlare zlib. I know that zlib-ng defaults to -O2 while CloudFlare zlib used -O3, so [perhaps this explains the observed pattern](https://github.com/zlib-ng/zlib-ng/issues/515). For your Windows data, the difference in compilers is really dramatic.

mtl1979:
gcc is generally bad at optimizing code without explicitly setting target processor... This is because AMD and Intel have slight differences in how some instructions behave and gcc generally optimizes for AMD because that means same code works for processors made by both manufacturers. That however results in speed penalty on Intel processors.

When I worked with fixing optimization issues on gcc, I noticed that register intensive code makes gcc spill registers to memory instead of reordering the code so register spill would not be necessary.

Increasing optimization level can help to decrease speed penalty in some cases but it might also introduce more speed penalties due to gcc picking wrong optimization for some code blocks.

nmoinvaz:
Rebased

PR #589, opened 2020-05-23
Use standard int types in zlib-ng api

nmoinvaz:
See last commits.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -249,14 +249,14 @@ ZLIB_INTERNAL void slide_hash_c(deflate_state *s) {
 }
 
 /* ========================================================================= */
-int ZEXPORT PREFIX(deflateInit_)(PREFIX3(stream) *strm, int level, const char *version, int stream_size) {
+int32_t ZEXPORT PREFIX(deflateInit_)(PREFIX3(stream) *strm, int32_t level, const char *version, int32_t stream_size) {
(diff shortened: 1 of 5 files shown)


Dead2:
Are these changes compatible with the zlib-compat api on all the arches we support?

I know int is 16bit on 16bit platforms, but we don't support those anyway. But wasn't there another arch/platform with a similar issue?

nmoinvaz:
I wouldn't know about 16bit platforms.. @mtl1979 might though.

nmoinvaz:
And this was another instance I was unsure about what to do:

image

Should we convert to using `size_t`? I don't think it would be compatible.. we would need a shim for zlib API users..

mtl1979:
We already have shims for some functions, so adding more isn't issue.

What comes to changing int types, only PPC is strict about them, and we already have PPC build in CI to catch those issues.

mtl1979:
@nmoinvaz size_t is not standard. It can be 32-bit even on 64-bit targets. We use _z suffix for functions that use size_t instead on uint64_t.

nmoinvaz:
I have changed `zng_gz_header::time` back to unsigned long.

nmoinvaz:
I think I will wait on this until after #602.

nmoinvaz:
Rebased

Dead2:
CI fails

mtl1979:
Conflicting declarations of inflateMark

Dead2:
Lets get this out there :)

PR #596, opened 2020-05-25
Add a workflow for clang and gcc static analysis

gdevenyi:


.github/workflows/analyze.yml
```diff
@@ -0,0 +1,39 @@
+name: CI Static Analysis
+on: [push, pull_request]
+jobs:
+  gcc:
```
nmoinvaz: Name to GCC and Clang casing to be consistent with job names in other workflows..?

gdevenyi: Renamed and added versions to be more clear.

gdevenyi:
One last thing I could do here is I could add "--status-bugs" to the clang build, which would cause the CI to error if there are any clang-identified bugs.

Dead2:
@gdevenyi I think that would be best, since otherwise nobody will be checking the logs.

gdevenyi:
clang static analysis will now throw errors if anything is detected. Looking into how to make GCC do it but its at compiler level rather than a wrapper.

gdevenyi:
It looks like I don't have a way to make the GCC analyzer fail out on detecting a problem.

So, this is ready for merge.

It will start failing test builds until the outstanding clang warnings are fixed.

turol:
Adding `-Werror` should cause analyzer warnings to become errors. A side effect is that this will make all warnings into errors. But you should fix those anyway. Alternatively there might be some variant of `-Werror=` which could make only analyzer warnings into errors.

gdevenyi:
Thanks @turol I looked at that, unfortunately it will cause the build to fail on the first analyzer warning rather than providing a report on the whole build.

mtl1979:
I've seen people use scan-build even with gcc to get better static analyzer reports... Basically it uses gcc to compile the files, but clang to analyze the output...

gdevenyi:
@mtl1979 technically, that's already happening for my scan-build run, because I believe CC is being exported by the workflow, scan-build does the analysis but GCC ends up doing the building.

mtl1979:
Looking at the changes, it looks like in the first path gcc does both compiling and static analysis, and in the second path, scan-build is wrapped around cmake that use clang as actual compiler...

gdevenyi:
Disabled extra maintainer warnings and code coverage on analyzer runs to avoid duplicating logging output and hiding analyzer outputs.

gdevenyi:
clang analyzer now properly fails when scan-build detects bugs.

Dead2:
Looks pretty good to me

PR #597, opened 2020-05-25
Remove toofar checks

Dead2:
Remove toofar checks now that we always hash 4 bytes, and the minimum
match length is thus also 4. (MIN_MATCH is still 3 due to it being used for calculations of other needed numbers, and they fail when set to 4, those still need to get fixed).

If MIN_MATCH was 4, we could instead hide these behind an ifdef.

Thoughts on whether we want to support an actual MIN_MATCH=3 mode in the future? It would need a different UPDATE_HASH algo in that case, and would sacrifice a lot of speed.

Not supporting MIN_MATCH=3 could affect compression of some png files, that is why for example optipng recommends setting TOO_FAR higher than the default.
It would likely not be an issue for just saving a png, but when you try to use optipng/pngcrush for maximum compression we would not be able to reduce size as much as stock zlib.

### DIFF
--- a/deflate_medium.c
+++ b/deflate_medium.c
@@ -261,11 +261,6 @@ ZLIB_INTERNAL block_state deflate_medium(deflate_state *s, int flush) {
                 else
                     fizzle_matches(s, &current_match, &next_match);
             }
-
-            /* short matches with a very long distance are rarely a good idea encoding wise */
-            /* distances 8193–16384 take 12 extra bits, distances 16385–32768 take 13 extra bits */
-            if (next_match.match_length == 3 && (next_match.strstart - next_match.match_start) > 12000)
-                    next_match.match_length = 1;
             s->strstart = current_match.strstart;
 
         } else {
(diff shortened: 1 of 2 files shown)


Dead2:
Oh, forgot to add the reference to optipng
http://optipng.sourceforge.net/pngtech/too_far.html

nmoinvaz:
I am not too familiar with that code but it sounds ok. It seems like the `TOO_FAR` is fine tuning the algorithm in a way that is similar to `s->level` check in `longest_match`.

Dead2:
Well, the theory behind TOO_FAR is that short matches (3 bytes) might end up not saving space if the distance is too high. But it is not quite that simple, and it seems to end up being positive/negative almost randomly. (Likely depending on the frequency of that distance, IIRC)

nmoinvaz:
In that case it does seem unnecessary given we hash 4 bytes. Yeah, I don't know that we would want to support an actual `MIN_MATCH=3` use case because the performance suffers..

Dead2:
Oh, yeah, I forgot to say that since we already hash 4 bytes, this PR does not change the compressed output at all, since the checks will always be false.

PR #600, opened 2020-05-27
Remove IPos typedef.

nmoinvaz:
This typedef seems like legacy compiler optimization from K&R days and leads to more casting warnings.


Dead2:
Looks good to me.

Issue #605, opened 2020-05-28
deflate_quick invalid stored block lengths error


Dead2:
@nmoinvaz I found a problem with one of the recent changes to deflate_quick

minigzip attempts a decompress and reports`./minigzip: /tmp/zlib-testfil.gz: invalid stored block lengths`
Same file with gzip: `gzip: /tmp/zlib-testfil.gz: invalid compressed data--format violated`

I bisected this to 343596fc98a0d7a6283dbe5d25abefca18439184.

The way I catched this was to set BUFLEN to 1MIB, like this:
`CFLAGS="-DBUFLEN=1048576" ./configure --warn && make`
etc, and this triggered with my testdata.

With the default buflen, it does not get any errors.
I suspect this is a problem that could occur in normal systems as well, and might be the one @neurolabusc is currently getting as well

nmoinvaz:
Nice find! I will take a look at it and see if changing it back fixes @neurolabusc’s issue. 

We could also add a CI instance for large BUFLEN..

Dead2:
I was thinking the same thing :)

nmoinvaz:
It appears to be different as I am still getting the original error @neurolabusc reported:
```
pigzCloudflare.exe: skipping: ./temp\pigznmoinvaz.exe1_asl32.nii.gz: corrupted -- invalid deflate data (invalid distance code)
```

nmoinvaz:
I'm also not sure why it would be emitting out a `STORED` block in _deflate_quick_ also `Z_SYNC_FLUSH` isn't used in _minigzip_ either..

Are you able to reproduce this on any publicly available files? I wasn't able to reproduce it on my `earth32.bmp` which is 32MB.

mtl1979:
It would be useful to know what are the two lengths that don't match... it should write the block length first, then bit negated version as "checksum". As far as I know, block length for stored block when using deflate_quick should always be 0, so the negated version is 0xFFFF.

Dead2:
@nmoinvaz 
I managed to trim down the 40MiB file to 160KiB while still preserving the error. At that point removing 32KiB from the beginning or end would also remove the error.
https://mirror.circlestorm.org/buggy-testfile.gz

Dead2:
The above is the test data compressed by gzip.
I uploaded the faulty one that was compressed by deflate_quick as well here:
https://mirror.circlestorm.org/buggy-testfile-deflate_quick.gz

nmoinvaz:
[...]

PR #606, opened 2020-05-29
Fixed invalid deflate quick output when flush_pending didn't flush

nmoinvaz:
See #605 for background.

> Fixed invalid deflate quick output when flush_pending didn't flush due to s->strm->avail_out == 0. We check that there is input data available to process and output space to write to, which allows us to continue processing as long as we can keep flushing successfully.


nmoinvaz:
I'm getting the error again.

PR #607, opened 2020-05-29
Support larger BUFLEN sizes by dynamically allocating buf.

nmoinvaz:


mtl1979:
This seems to have revealed another issue somewhere... No SSE2 and non-optimized versions are failing...

nmoinvaz:
Rebased for CI status..

Dead2:
Is this alignment related somehow?
I really don't see how, but it is the only thing that has made any sense to me since this was first posted.

mtl1979:
@Dead2 Maybe it's still using SSE even though SSE2 is disabled. Clearly using malloc() is no-go as it does not align on 16 byte boundary.

Dead2:
But we cannot require that an external buffer (from zlib's perspective) is aligned, and we cannot expect it to be either.
Actually that sounds like something we should explicitly do in tests, for example do our external-to-zlib allocs 16-byte aligned and then skip the first 3 bytes to ensure that we are certainly not aligned in any sense, to provoke any possible such error.

nmoinvaz:
I can rebase after #560 and see if that fixes the problem.

nmoinvaz:
Rebased.

test/minigzip.c
```diff
@@ -102,9 +102,9 @@ void gz_compress(FILE *in, gzFile out) {
 #endif
     /* Clear out the contents of buf before reading from the file to avoid
        MemorySanitizer: use-of-uninitialized-value warnings. */
-    memset(buf, 0, sizeof(buf));
+    memset(buf, 0, BUFLEN);
```
nmoinvaz: I could optionally use `calloc` instead.

nmoinvaz: I have added that as a separate commit.

nmoinvaz:
It looks like the previous errors went away due to the latest merges in `develop`. It could have been fixed by #560 or #657, but most likely #560.

Issue #595, opened 2020-05-24
clang-11 -Weverything at 1e280ca517976f464116a460729cd0a34b5dd682


gdevenyi:
```
$ clang-11 --version
clang version 11.0.0-++20200523063809+304b0ed4039-1~exp1~20200523164413.1701
$ CC=clang-11 cmake -DCMAKE_C_FLAGS_RELEASE:STRING='-O3 -DNDEBUG -Weverything' .
make -j40 2> clang-warnings.log
```

[clang-warnings.log](https://github.com/zlib-ng/zlib-ng/files/4674049/clang-warnings.log)

nmoinvaz:
Cleaning up every single warning will take ages.

gdevenyi:
Sure, I realize this is a huge amount of warnings, possibly many of these spurious. But I hope at least some of these might provide beneifical hints.

Would there be interest in getting this into the CI?

nmoinvaz:
In CI we do `WITH_MAINTAINER_WARNINGS` option.

PR #609, opened 2020-05-29
Add -Weverything for maintainer builds on clang

gdevenyi:
Ref #595


Dead2:
I am not quite sure I want this many warnings in CI, for one it makes it harder to spot any serious ones, and it also prevents us from enabling -Werror for CI builds. If all those warnings had been fixed first, that might help though.
-Weverything is a good exercise to do sometimes though, to see whether there are low-hanging code-quality improvements that can be done.

gdevenyi:
How about a seperate CI config for -Weverything with my analyzer PR?

nmoinvaz:
I think the level of warnings that the `WITH_MAINTAINER_WARNINGS` provides is sufficient in most cases. If there is a PR to actually fix the warnings that is more valuable, but I will likely never go into the CI job and search for warnings to fix.

gdevenyi:
Okay no problem.

PR #610, opened 2020-05-29
Fix specification of warnings for intel compilers

gdevenyi:
I got a warning testing on intel that -W3 and -W4 weren't supported, that's because intel's compiler is stupidly inconsistent with its settings across platforms:
https://software.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/error-handling-1/warnings-errors-and-remarks.html

I updated the cmake based on the latest documentation.


Dead2:
Thanks :)

PR #613, opened 2020-05-29
Cache chocolatey packages on Windows for CMake workflow.

nmoinvaz:
This should cache chocolatey packages on Windows.
*  It prevents download errors on Windows GCC builds we have been getting
 * Speed up Windows GCC CI builds because don't have to redownload. 

This is the expected output when loading from cache:

```
Chocolatey v0.10.15
Installing the following packages:
mingw
By installing you accept licenses for the packages.

mingw v7.3.0 (forced) [Approved]
mingw package files install completed. Performing other installation steps.
File appears to be downloaded already. Verifying with package checksum to determine if it needs to be redownloaded.
Hashes match.
Hashes match.
```

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -341,6 +341,17 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y ${{ matrix.packages }}
 
+    - name: Set environment variables (Windows)
+      run:  echo "::set-env name=temp::$($env:TEMP)"
+
+    - name: Cache packages (Windows)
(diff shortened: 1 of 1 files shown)


Dead2:
I have no idea how to see whether this works or not.

nmoinvaz:
I believe if you re-run the workflow for the PR it should show in the interface the message shown in the description, but I don't have permissions to re-run zlib-ng workflows.

You can see at the beginning of the CI instance:

image

You can also see at the end of the run:

image

When it succeeds it should show:

image

PR #614, opened 2020-05-29
Run compression level tests without open mode.

nmoinvaz:
We currently don't test each compression level without an open mode. I have added those test cases, and removed some duplicate ones that are now covered.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1001,37 +1001,18 @@ if(ZLIB_ENABLE_TESTS)
             -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/run-and-redirect.cmake)
     endforeach()
 
-    macro(minigzip_deflate_file target name path)
-        set(GZ_COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:${target}>
-            ${CMAKE_CURRENT_SOURCE_DIR}/${path})
-        add_test(NAME ${target}-${name}-file-compr
-            COMMAND ${CMAKE_COMMAND}
-            "-DCOMMAND=${GZ_COMMAND}"
-            "-DSUCCESS_EXIT=0;1"
-            -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/run-and-redirect.cmake)
-    endmacro()
-    macro(minigzip_inflate_file target name path)
-        set(GZ_COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:${target}>
-            -d ${CMAKE_CURRENT_SOURCE_DIR}/${path})
-        add_test(NAME ${target}-${name}-file-uncompr
-            COMMAND ${CMAKE_COMMAND}
-            "-DCOMMAND=${GZ_COMMAND}"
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1095,10 +1073,14 @@ if (ZLIB_ENABLE_TESTS)
         endif()
     endmacro()
 
-    set(OPEN_MODES -f -h -R -F -T)
+    set(OPEN_MODES -f -h -R -F -T -A)
     set(COMPRESSION_LEVELS -0 -1 -6 -9)
```
Dead2: Just a thought while I was reading this, maybe this should test level 3 instead of 1? level 3 proceeds further than 1 does. The others are all tested at their max.

Or should both min and max be tested? (-0 -1 -2 -3 -4 -6 -7 -9)   PS: 2 is in this because it is minimum deflate_fast if deflate_quick is disabled..  That only excludes 5 and 8. I would guess this would increase coverage of the algos a tiny bit.

nmoinvaz: I have updated the compression levels to the ones you mentioned. I have no preference which ones we test; the more the merry.

CMakeLists.txt
```diff
@@ -1095,10 +1073,14 @@ if (ZLIB_ENABLE_TESTS)
...
-    foreach(OPEN_MODE ${OPEN_MODES})
-        foreach(COMPRESSION_LEVEL ${COMPRESSION_LEVELS})
+    set(OPEN_MODES -f -h -R -F -T -A)
+    set(COMPRESSION_LEVELS -0 -1 -2 -3 -4 -6 -7 -9)
+
+    foreach(COMPRESSION_LEVEL ${COMPRESSION_LEVELS})
+        minigzip_stdio_cmp(minigzip "lcet10" "test/data/lcet10.txt" ${COMPRESSION_LEVEL})
+        minigzip_stdio_cmp(minigzip "fireworks" "test/data/fireworks.jpg" ${COMPRESSION_LEVEL})
+        minigzip_stdio_cmp(minigzip "paper-100k" "test/data/paper-100k.pdf" ${COMPRESSION_LEVEL})
+        foreach(OPEN_MODE ${OPEN_MODES})
```
Dead2: Reading the test logs is kind of a mess now, could that be improved by reordering the levels of loops?

Would this be better?
```
for (file)
    for (mode)
        for (level)
```
Not sure how to do two parameters for the file loop though.

The extra non-mode tests above could possibly also be removed by including an empty ("") mode in the list, at least I assume that would work in cmake lists.

nmoinvaz: I have simplified it with the `for (file)` method. I tried using the empty string or even a string with a single space in it, but the way CMake parses arg lists was having difficulty with it.

The ordering should look something like this now:

image

nmoinvaz:
It might be too many tests now with all the different compression levels etc. Close to 1200. CI is a bit slower..

mtl1979:
There is never too many tests... Even with the old count, there was just 2 failures out of about 500 tests with some commits... That was still 2 too many failures...

nmoinvaz:
I changed it to do `-0 -1 -3 -6 -9`. It now has around 300 tests.

PR #615, opened 2020-05-30
Fixed wrong 64-bit casting in deflatePrime causing bits to be lost.

nmoinvaz:
The casting to higher integer type should happen to `value` first so no loss of bits.

Even _madler/zlib_ has the same casting, it looks like since it always used < 32-bit it was ok. Probably should have casted `value` first anyways.

image

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -581,7 +581,7 @@ int ZEXPORT PREFIX(deflatePending)(PREFIX3(stream) *strm, uint32_t *pending, int
 /* ========================================================================= */
 int ZEXPORT PREFIX(deflatePrime)(PREFIX3(stream) *strm, int bits, int value) {
     deflate_state *s;
-    int put;
+    int32_t put;
 
     if (deflateStateCheck(strm))
         return Z_STREAM_ERROR;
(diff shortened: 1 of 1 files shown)


deflate.c
```diff
@@ -593,7 +593,7 @@ int ZEXPORT PREFIX(deflatePrime)(PREFIX3(stream) *strm, int bits, int value) {
         put = BIT_BUF_SIZE - s->bi_valid;
         if (put > bits)
             put = bits;
-        s->bi_buf |= (uint64_t)((value & ((1 << put) - 1)) << s->bi_valid);
+        s->bi_buf |= (((uint64_t)value & (((uint64_t)1 << put) - 1)) << s->bi_valid);
```
mtl1979: For constants, you should use UINT64_C() macro instead...

nmoinvaz:
This is partially what is causing #536.

deflate.c
nmoinvaz: Fixed. Thanks!

nmoinvaz:
Moved to #618.

PR #616, opened 2020-05-31
Fixed multi-line assembly macro in UPDATE_HASH for MSVC

nmoinvaz:
https://docs.microsoft.com/en-us/cpp/assembler/inline/defining-asm-blocks-as-c-macros
```
insert_string_tpl.h(44,5): error : cannot use more than one symbol in memory operand
insert_string_sse.c(28,13): message : expanded from macro 'UPDATE_HASH'
```

### DIFF
--- a/arch/x86/insert_string_sse.c
+++ b/arch/x86/insert_string_sse.c
@@ -23,12 +23,10 @@
 #else
 #  ifdef _MSC_VER
 #    define UPDATE_HASH(s, h, val) {\
-        __asm {\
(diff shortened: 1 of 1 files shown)


Dead2:
Did not know this was a problem.
I would like another person besides me to review this though, since I do not know right from wrong here.

mtl1979:
Looks good to me...

nmoinvaz:
I encountered it when trying to compile using ClangCl.

PR #617, opened 2020-06-01
Fixed incorrect deflate header bits being sent

nmoinvaz:
```
    Arithmetic overflow: 32-bit value is shifted, then cast to 64-bit value.
      Results might not be an expected value.
```


nmoinvaz:
Moved to #618.

PR #619, opened 2020-06-03
Move trace statements to emitting functions

nmoinvaz:
This also moves match length check into `check_match`. Also none of the `compare256` functions should be able to return more than 258..


Dead2:
Is it intentional that it no longer checks match length in a release compile, only for debug?

nmoinvaz:
Yes that is intentional. `compare258` functions should never return more than 258 anyways. This should really a debug mode check. None of the other deflate functions have it. They must have put the check in when writing it initially and never took it out after deflate_quick's _compare258_ was perfected.

nmoinvaz:
Rebased

Issue #621, opened 2020-06-04
Unable to compile minigzip with USE_MMAP defined


neurolabusc:
On my Ubuntu 20.04 Ryzen system with gcc version 9.3.0, I am unable to compile with the USE_MMAP option for the minigzip project. I think this is because the [caddr_t is a legacy BSD type associated ... and it should never be used in modern code](https://stackoverflow.com/questions/6381526/what-is-the-significance-of-caddr-t-and-when-is-it-used)

```
export CXXFLAGS="-DUSE_MMAP=1" 
export CFLAGS="-DUSE_MMAP=1" 
cmake  ..
```

results in errors
```
/home/chris/Desktop/z/gz/test/minigzip.c: In function ‘gz_compress_mmap’:
/home/chris/Desktop/z/gz/test/minigzip.c:129:5: error: unknown type name ‘caddr_t’
  129 |     caddr_t buf;    /* mmap'ed buffer for the entire input file */
      |     ^~~~~~~
```

I can compile the project if I define caddr_t in minigzip.c:
```
typedef	char		*caddr_t;	/* ? type */ 
```

mtl1979:
Linux is fork of BSD (as is MacOS), so this is compiler issue, not issue with zlib-ng itself... 

We can't add workarounds for every possible broken compiler just because someone has claimed something should not be used. This is because we would need to know exact compiler and C library version where the deprecation happened, otherwise we end up fixing the same third-party bug over and over.

I checked system headers of Ubuntu 18.04 and they still use caddr_t extensively.

As for the type itself, it can be replaced by either "char *", "unsigned char *" or "void *" depending on surrounding code, so there is no universal fix.

nmoinvaz:
This change has been merged. @neurolabusc can you confirm it works?

neurolabusc:
Fixed. Thanks!

PR #622, opened 2020-06-04
Use char* instead of caddr_t

mtl1979:
See #621. Type caddr_t was removed in recent version of sys/types.h

### DIFF
--- a/test/minigzip.c
+++ b/test/minigzip.c
@@ -126,7 +126,7 @@ int gz_compress_mmap(FILE *in, gzFile out) {
     int len;
     int err;
     int ifd = fileno(in);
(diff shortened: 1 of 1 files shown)


test/minigzip.c
```diff
@@ -126,7 +126,7 @@ int gz_compress_mmap(FILE *in, gzFile out) {
     int len;
     int err;
     int ifd = fileno(in);
-    caddr_t buf;    /* mmap'ed buffer for the entire input file */
+    char *buf;    /* mmap'ed buffer for the entire input file */
```
nmoinvaz: Comment should be aligned?

mtl1979: Fixed.

PR #623, opened 2020-06-04
Added help messages to minigzip and switchlevels.

nmoinvaz:


test/minigzip.c
```diff
@@ -250,20 +250,20 @@ void file_uncompress(char *file, int keep) {
...
- *   -h : compress with Z_HUFFMAN_ONLY
- *   -R : compress with Z_RLE
- *   -F : compress with Z_FIXED
- *   -T : stored raw
- *   -A : auto detect type
- *   -0 to -9 : compression level
- */
+void show_help(void)
+{
+    printf("Usage: minigzip [-c] [-d] [-k] [-f|-h|-R|-F|-T] [-A] [-1 to -9] [files...]\n\n" \
```
mtl1979: Should be "-0 to -9" to be consistent with the text below.

test/switchlevels.c
```diff
@@ -107,10 +107,11 @@ static int compress_chunk(PREFIX3(stream) *strm, int level, int size, int last)
...
 }
 
-/* ===========================================================================
- * Usage:  switchlevels [-w bits] level1 size1 [level2 size2 ...]
- *   -w : window bits
- */
+void show_help(void)
+{
+    printf("Usage: switchlevels [-w bits] level1 size1 [level2 size2 ...]\n\n" \
+           "  -w : window bits\n\n");
```
mtl1979: This should mention valid range for window bits...

test/minigzip.c
```diff
@@ -276,6 +276,12 @@ int main(int argc, char *argv[]) {
     char *level = "6";
     char *type = "b";
 
+
+    if (argc <= 1) {
```
mtl1979: argc cannot be 0, it's always at least 1, as the process name is the first argument.

test/switchlevels.c
```diff
@@ -121,6 +122,12 @@ int main(int argc, char **argv) {
     int window_bits = MAX_WBITS + 16;
     PREFIX3(stream) strm;
 
+
+    if (argc <= 1) {
```
mtl1979: argc cannot be 0, it's always at least 1, as the process name is the first argument.

nmoinvaz: Fixed.

test/minigzip.c
nmoinvaz: Fixed.

test/minigzip.c
nmoinvaz: Fixed.

test/switchlevels.c
nmoinvaz: Fixed.

Dead2:
Could we catch `--help` here as well?
For many (like me) it is a built in reflex to run a program like that if you are uncertain about the parameters, since for some programs it can be problematic to just run it without parameters at all (since it will then do stuff, and not just print the help text).

mtl1979:
@Dead2 I was thinking of suggesting that too, because ```-h``` is already in use.

nmoinvaz:
I thought it was kind of odd that we would support single dash `-` in all instances but one `--help`.

mtl1979:
@nmoinvaz That was actually the reason why I didn't suggest adding it... It would have required to add double dash version for all command-line switches and/or changing the existing ```-h```.

Dead2:
There are other programs out there that don't print help with -h and only with --help. I'd leave it out of the parameter help text and just support it silently.
An alternative is displaying the helptext for any unknown parameter, but then we should probably support `--` (no more parameters after this, treat them as files).

nmoinvaz:
Ok I added support for `-help` and `--help`.

test/minigzip.c
```diff
@@ -276,6 +275,11 @@ int main(int argc, char *argv[]) {
     char *level = "6";
     char *type = "b";
 
+    if ((argc == 1) || (argc == 2 && strstr(argv[1], "-help"))) {
```
mtl1979: strstr() is a hack, should use strcmp, otherwise someone could pass in any junk (including real filenames) containing "-help" anywhere... This defeats the purpose even though it would be nice for it to catch any unknown command-line flag, but even that requires checking the first character which strstr() doesn't.

nmoinvaz: Seems like adding help text is more trouble than it is worth..

mtl1979: Nah... It's just some functions should never be used... strstr() is one of them...

nmoinvaz: Ok it is changed.

nmoinvaz:
Ok now only supports `--help`.

PR #625, opened 2020-06-04
Disable unaligned access of > 4 bytes on 32-bit arm

### LINKED ISSUE #624: longest_match_unaligned_64 trips the alignment trap on armv7
Tested on a Cortex A8 & a Cortex A9 (Linux, GCC 7.5, @ 1a3d80677fa3bfce74f521c278040a2a8bfb04ad).

My feeble understanding of the ARM specs would indicate that it makes sense (those are 32-bit processor, this is more that word-sized) ;).

Even if the kernel is setup in fixup mode, the performance penalty is not worth it. Which means, ideally, this shouldn't be compiled in on 32-bit arm, even with `UNALIGNED_OK`.

I'm not sure I see a better solution than adding another preprocessor define (say, `UNALIGNED64_OK`), which means buildsystem tweaks :/. Iff this limitation (can't load > word size) happened to be true on every supported 32-bit arch, something like a `__SIZEOF_POINTER__ == 8` check would do the trick (with the caveat that this is non-standard, I can vouch for it with GCC & Clang, but that's it), but I don't even know if that assumption is sound to begin with ;).

(FWIW, I currently simply disable the `HAVE_BUILTIN_CTZLL` configure check on my end, and everything's peachy :)).

NiLuJe:
Fix #624

I'm not a fan of CMake, so, hopefully I haven't mangled that too much ;).


nmoinvaz:
Looks good. I think you are safe to squash/combine those commits into one.

Dead2:
Looks good to me.

NiLuJe:
(Rebased w/ ppc64le handling, where I assume this ought to be enabled ;)).

PR #636, opened 2020-06-10
test: bring back default behavior of minigzip

### LINKED ISSUE #635: minigzip test failing
As part of `make test`, `minigzip` is tested with a simple `hello world` string.
 https://github.com/zlib-ng/zlib-ng/blob/0b39a43490aea6024cc57bb577be58c675142d6d/test/Makefile.in#L50

But this test seems to be broken right now (commit 0b39a43490aea):
```
$ echo "hello world" | ./minigzip | ./minigzip -d
Usage: minigzip [-c] [-d] [-k] [-f|-h|-R|-F|-T] [-A] [-0 to -9] [files...]

  -c : write to standard output
  -d : decompress
  -k : keep input files
  -f : compress with Z_FILTERED
  -h : compress with Z_HUFFMAN_ONLY
  -R : compress with Z_RLE
  -F : compress with Z_FIXED
  -T : stored raw
  -A : auto detect type
  -0 to -9 : compression level
```

Doing a bisect shows that the problem was introduced by 18da0db0d9e8663683

Apart from the minigzip issue, the test should have failed since it is not doing what is expected.

dankegel:
Fixes #635 

Also:
- makes sure 'make test' fails if minigzip aborts
- fixes old quoting error that prevented temp file names from being unique as intended
- uses a distinguishable exit code on bad commandline, as minigzip_stdio_cmp treats 1 as expected and ok.  Chose value used by BSD for bad usage.

### DIFF
--- a/test/Makefile.in
+++ b/test/Makefile.in
@@ -46,26 +46,28 @@ fuzzer:
 endif
 
 teststatic: check_cross_dep
-	@TMPST=tmpst_$$; \
-	if echo hello world | ${QEMU_RUN} ../minigzip$(EXE) | ${QEMU_RUN} ../minigzip$(EXE) -d && ${QEMU_RUN} ../example$(EXE) $$TMPST ; then \
+	@TMPST=tmpst_$$$$; \
+	HELLOST=tmphellost_$$$$; \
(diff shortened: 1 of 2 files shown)


mscastanho (commented):
LGTM =)

nmoinvaz:
I have modified #473 to also do command line parameter the same way with regard to `--help`.

Issue #633, opened 2020-06-09
invalid instruction exception on mac


malytomas:
My application crashes with EXC_BAD_INSTRUCTION (code=EXC_I386_INVOP) on this line:
https://github.com/zlib-ng/zlib-ng/blob/53f9257a24362386e75504836eaeb2788360b1e8/memcopy.h#L133

callstack:
```
chunkmemset_1
chunkmemset
zng_inflate_fast
inflate
png_zlib_inflate
```

Running on iMac (mid 2011)
macos High Sierra (version 10.13.6)

cmake output:
```
-- The C compiler identification is AppleClang 9.1.0.9020039
-- The CXX compiler identification is AppleClang 9.1.0.9020039

-- Using CMake version 3.14.3
-- ZLIB_HEADER_VERSION: 1.2.11
-- ZLIBNG_HEADER_VERSION: 1.9.9
-- Arch detected: 'x86_64'
-- Basearch of 'x86_64' has been detected as: 'x86'
-- Architecture supports unaligned reads
-- Architecture-specific source files: arch/x86/x86.c;arch/x86/slide_avx.c;arch/x86/insert_string_sse.c;arch/x86/deflate_quick.c;arch/x86/fill_window_sse.c;arch/x86/slide_sse.c;arch/x86/crc_folding.c
-- The following features have been enabled:

 * CMAKE_BUILD_TYPE, Build type: Release (selected)
 * ZLIB_COMPAT, Provide a zlib-compatible API
 * WITH_OPTIM, Build with optimisation
 * WITH_NEW_STRATEGIES, Use new strategies
 * AVX2_SLIDEHASH, Support AVX2-optimized slide_hash, using "-mavx2"
 * SSE42_CRC, Support CRC hash generation using the SSE4.2 instruction set, using "-msse4"
 * SSE42_DEFLATE_QUICK, Support SSE4.2-accelerated quick compression
 * PCLMUL_CRC, Support CRC hash generation using PCLMULQDQ, using "-mpclmul"

-- The following features have been disabled:

 * WITH_GZFILEOP, Compile with support for gzFile-related functions
 * WITH_SANITIZERS, Build with address sanitizer and all supported sanitizers other than memory sanitizer
 * WITH_MSAN, Build with memory sanitizer
 * WITH_FUZZERS, Build test/fuzz
 * WITH_MAINTAINER_WARNINGS, Build with project maintainer warnings
 * WITH_CODE_COVERAGE, Enable code coverage reporting
[...]

PR #639, opened 2020-06-11
Install identical bits for configure and cmake build systems.

### LINKED ISSUE #638: Hard to verify that cmake and configure do the same thing
The two build systems do not produce exactly the same output, and it's hard to know if that is benign or a real problem.  But it turns out to be a real problem; see e.g. https://github.com/zlib-ng/zlib-ng/issues/553

dankegel:
This passes test/pkgcheck.sh on x86_64, but I have not yet added
a workflow to run that test on CI, so this is a WIP.

It also conflicts with the much more interesting https://github.com/zlib-ng/zlib-ng/pull/560
so I expect to rebase on top of that once it lands.

The most notable iffy bit about this pull request is that, despite my intention to only change the cmake build, I ended up changing configure to default to -O3 to match cmake's behavior (and zlib's).

Fixes #638 in preparation for fixing #633 (which is similar to #553).


CMakeLists.txt
```diff
@@ -720,34 +723,20 @@ if(WITH_OPTIM)
...
-            list(APPEND ZLIB_ARCH_SRCS ${ARCHDIR}/compare258_sse.c)
-            add_feature_info(SSE42_COMPARE258 1 "Support SSE4.2 optimized compare258, using \"${SSE4FLAG}\"")
-        endif()
         if(WITH_SSE2 AND HAVE_SSE2_INTRIN)
             add_definitions(-DX86_SSE2)
             list(APPEND ZLIB_ARCH_SRCS ${ARCHDIR}/slide_sse.c)
             if(NOT ${ARCH} MATCHES "x86_64")
-                add_intrinsics_option("${SSE2FLAG}")
+                add_intrinsics_option("${SSE2FLAG}" ${ARCHDIR}/slide_sse.c)
                 add_feature_info(FORCE_SSE2 FORCE_SSE2 "Assume CPU is SSE2 capable")
```
Dead2, line 740: Just some thoughts while reviewing this:

FORCE_SSE2 should also apply SSE2 to every compiled file, meaning we require the cpu to be SSE2 capable and that the compiler is free to utilize SSE2 and any preceding instruction set anywhere in the code. (This is only a base instruction set, we will still add AVX2 for files that need that for example).

WITH_SSE2 should not however, that should only apply to slide_sse.c for example.

I don't think FORCE_SSE2 should depend on WITH_SSE2 though, since the latter enables SSE2-specific functions, and the former enables the instruction set (but the X85_NOCHECK_SSE2 define still needs to depend on both being active though).

Looks like this distinction has been lost at some point (or I planned for this and forgot about it long ago, not sure).
I don't think FORCE_SSE2 currently exists in configure, but I think that would be good thing to add there without it requiring a lot of code.

I also think FORCE_SSE2 should default to ON on x86 (and permanently ON on x86_64), and that anyone who wants to support older cpus will be required to disable it manually. Those cpus are quite old by now, default to ON and get the speedups for free by default.

This might be a good fit for a separate PR though, but I will accept it as a separate commit in this PR if you want to implement it.

CMakeLists.txt
```diff
@@ -757,19 +746,38 @@ if(WITH_OPTIM)
...
+            add_intrinsics_option("${SSSE3FLAG}" ${ARCHDIR}/adler32_ssse3.c)
             add_feature_info(SSSE3_ADLER32 1 "Support SSSE3-accelerated adler32, using \"${SSSE3FLAG}\"")
-            add_intrinsics_option("${SSSE3FLAG}")
         endif()
         if(WITH_PCLMULQDQ AND HAVE_PCLMULQDQ_INTRIN)
             add_definitions(-DX86_PCLMULQDQ_CRC)
             list(APPEND ZLIB_ARCH_SRCS ${ARCHDIR}/crc_folding.c)
-            add_intrinsics_option("${PCLMULFLAG}")
+            add_intrinsics_option("${PCLMULFLAG}" ${ARCHDIR}/crc_folding.c)
+            add_intrinsics_option("${SSE4FLAG}" ${ARCHDIR}/crc_folding.c)
```
Dead2, line 756: Should the SSE4FLAG be applied in the else() part below?

dankegel:
Thank you for the review.  Closing in favor of https://github.com/zlib-ng/zlib-ng/pull/642 and a future pull request.

PR #647, opened 2020-06-18
Add optimized adler32 for POWER and new adler32 tests

mscastanho:
This PR adds an optimized adler32 implementation for POWER processors. The work was originally developed by @racardoso (madler/zlib#458). In addition, we provide a new set of tests for adler32 that applies to all archs. 

To measure the performance improvement, we used [Chromium's zlib_bench.cc](https://cs.chromium.org/chromium/src/third_party/zlib/contrib/bench/) with input files from [jsnell/zlib-bench](https://github.com/jsnell/zlib-bench/tree/master/corpus).

Benchmark was run on a POWER9 machine and the results below show compression and decompression throughput in MB/s using zlib wrapper, for all compression levels:

```
======== File 'html' ======== 

 Wrapper 'zlib'

          |    compression throughput (MB)     |   decompression throughput (MB)    |
  | level | default | adler32-power |     gain | default | adler32-power |     gain |
  |     1 |    90.2 |          94.8 |   +5.10% |   180.1 |         202.9 |  +12.66% |
  |     2 |    67.4 |          70.9 |   +5.19% |   212.6 |         244.9 |  +15.19% |
  |     3 |    51.2 |          53.0 |   +3.52% |   224.5 |         260.2 |  +15.90% |
  |     4 |    48.1 |          49.9 |   +3.74% |   230.7 |         267.2 |  +15.82% |
  |     5 |    43.4 |          44.8 |   +3.23% |   233.0 |         272.0 |  +16.74% |
  |     6 |    35.7 |          36.7 |   +2.80% |   235.1 |         274.8 |  +16.89% |
  |     7 |    23.3 |          23.7 |   +1.72% |   230.4 |         268.4 |  +16.49% |
  |     8 |    16.2 |          16.4 |   +1.23% |   230.4 |         268.5 |  +16.54% |
  |     9 |    15.9 |          16.1 |   +1.26% |   230.3 |         268.7 |  +16.67% |

======== File 'executable' ======== 

 Wrapper 'zlib'

          |    compression throughput (MB)     |   decompression throughput (MB)    |
  | level | default | adler32-power |     gain | default | adler32-power |     gain |
  |     1 |    99.7 |         106.3 |   +6.62% |   210.7 |         242.3 |  +15.00% |
  |     2 |    65.3 |          68.3 |   +4.59% |   202.0 |         230.6 |  +14.16% |
  |     3 |    54.4 |          56.6 |   +4.04% |   209.0 |         239.7 |  +14.69% |
  |     4 |    48.9 |          50.6 |   +3.48% |   214.7 |         247.3 |  +15.18% |
  |     5 |    44.7 |          46.2 |   +3.36% |   216.8 |         249.0 |  +14.85% |
  |     6 |    37.9 |          39.0 |   +2.90% |   218.2 |         251.7 |  +15.35% |
  |     7 |    22.2 |          22.5 |   +1.35% |   220.4 |         254.6 |  +15.52% |
  |     8 |    14.5 |          14.7 |   +1.38% |   220.9 |         255.2 |  +15.53% |
  |     9 |     9.4 |           9.5 |   +1.06% |   221.2 |         255.7 |  +15.60% |

======== File 'jpeg' ======== 

 Wrapper 'zlib'

          |    compression throughput (MB)     |   decompression throughput (MB)    |
  | level | default | adler32-power |     gain | default | adler32-power |     gain |
  |     1 |    59.5 |          62.3 |   +4.71% |   176.6 |         198.5 |  +12.40% |
  |     2 |    29.1 |          29.8 |   +2.41% |   870.2 |        1953.8 | +124.52% |
  |     3 |    29.1 |          29.8 |   +2.41% |   869.2 |        1954.6 | +124.87% |
  |     4 |    24.7 |          25.1 |   +1.62% |   869.4 |        1952.0 | +124.52% |
  |     5 |    24.7 |          24.8 |   +0.40% |   870.2 |        1949.1 | +123.98% |
  |     6 |    24.7 |          25.1 |   +1.62% |   869.4 |        1951.8 | +124.50% |
  |     7 |    28.7 |          29.4 |   +2.44% |   870.1 |        1954.9 | +124.68% |
  |     8 |    28.7 |          29.4 |   +2.44% |   869.3 |        1955.3 | +124.93% |
  |     9 |    28.7 |          29.4 |   +2.44% |   869.3 |        1954.4 | +124.82% |

======== File 'pngpixels' ======== 

 Wrapper 'zlib'

          |    compression throughput (MB)     |   decompression throughput (MB)    |
  | level | default | adler32-power |     gain | default | adler32-power |     gain |
  |     1 |   144.7 |         160.0 |  +10.57% |   301.6 |         370.5 |  +22.84% |
  |     2 |   122.8 |         134.4 |   +9.45% |   343.0 |         433.6 |  +26.41% |
  |     3 |    87.0 |          92.9 |   +6.78% |   377.8 |         489.3 |  +29.51% |
  |     4 |    76.2 |          80.6 |   +5.77% |   375.3 |         484.1 |  +28.99% |
  |     5 |    57.2 |          59.7 |   +4.37% |   407.4 |         540.1 |  +32.57% |
  |     6 |    30.5 |          31.2 |   +2.30% |   440.8 |         599.4 |  +35.98% |
  |     7 |    16.9 |          17.2 |   +1.78% |   439.7 |         597.5 |  +35.89% |
  |     8 |     5.3 |           5.3 |   +0.00% |   448.8 |         613.7 |  +36.74% |
  |     9 |     2.4 |           2.5 |   +4.17% |   451.0 |         618.9 |  +37.23% |


```

### DIFF
--- a/adler32_p.h
+++ b/adler32_p.h
@@ -80,4 +80,14 @@ static inline uint32_t adler32_len_16(uint32_t adler, const unsigned char *buf, 
     return adler | (sum2 << 16);
 }
 
+static inline uint32_t adler32_len_64(uint32_t adler, const unsigned char *buf, size_t len, uint32_t sum2) {
+    while (len >= 16) {
+        len -= 16;
+        DO16(adler, sum2, buf);
+        buf += 16;
+    }
+    /* Process tail (len < 16).  */
+    return adler32_len_16(adler, buf, len, sum2);
+}
+
 #endif /* ADLER32_P_H */
--- a/adler32_p.h
+++ b/adler32_p.h
@@ -12,6 +12,12 @@
 #define NMAX 5552
 /* NMAX is the largest n such that 255n(n+1)/2 + (n+1)(BASE-1) <= 2^32-1 */
 
+#define DO1(sum1, sum2, buf, i)  {(sum1) += buf[(i)]; (sum2) += (sum1);}
+#define DO2(sum1, sum2, buf, i)  {DO1(sum1, sum2, buf, i); DO1(sum1, sum2, buf, i+1);}
+#define DO4(sum1, sum2, buf, i)  {DO2(sum1, sum2, buf, i); DO2(sum1, sum2, buf, i+2);}
+#define DO8(sum1, sum2, buf, i)  {DO4(sum1, sum2, buf, i); DO4(sum1, sum2, buf, i+4);}
+#define DO16(sum1, sum2, buf)    {DO8(sum1, sum2, buf, 0); DO8(sum1, sum2, buf, 8);}
+
 /* use NO_DIVIDE if your processor does not do division in hardware --
    try it both ways to see which is faster */
 #ifdef NO_DIVIDE
--- a/arch/power/adler32_power8.c
+++ b/arch/power/adler32_power8.c
@@ -1,0 +1,165 @@
+/* Adler32 for POWER8 using VSX instructions.
+ * Copyright (C) 2020 IBM Corporation
+ * Author: Rogerio Alves <rcardoso@linux.ibm.com>
+ * For conditions of distribution and use, see copyright notice in zlib.h
+ *
+ * Calculate adler32 checksum for 16 bytes at once using POWER8+ VSX (vector)
+ * instructions.
+ *
+ * If adler32 do 1 byte at time on the first iteration s1 is s1_0 (_n means
+ * iteration n) is the initial value of adler - at start  _0 is 1 unless
+ * adler initial value is different than 1. So s1_1 = s1_0 + c[0] after
+ * the first calculation. For the iteration s1_2 = s1_1 + c[1] and so on.
+ * Hence, for iteration N, s1_N = s1_(N-1) + c[N] is the value of s1 on
+ * after iteration N.
+ *
+ * Therefore, for s2 and iteration N, s2_N = s2_0 + N*s1_N + N*c[0] +
+ * N-1*c[1] + ... + c[N]
+ *
+ * In a more general way:
+ *
+ * s1_N = s1_0 + sum(i=1 to N)c[i]
+ * s2_N = s2_0 + N*s1 + sum (i=1 to N)(N-i+1)*c[i]
+ *
+ * Where s1_N, s2_N are the values for s1, s2 after N iterations. So if we
+ * can process N-bit at time we can do this at once.
+ *
+ * Since VSX can support 16-bit vector instructions, we can process
+ * 16-bit at time using N = 16 we have:
+ *
+ * s1 = s1_16 = s1_(16-1) + c[16] = s1_0 + sum(i=1 to 16)c[i]
+ * s2 = s2_16 = s2_0 + 16*s1 + sum(i=1 to 16)(16-i+1)*c[i]
+ *
+ * After the first iteration we calculate the adler32 checksum for 16 bytes.
+ *
+ * For more background about adler32 please check the RFC:
+ * https://www.ietf.org/rfc/rfc1950.txt
+ */
+
+#ifdef POWER8_VSX_ADLER32
+
+#include <altivec.h>
+#include "zbuild.h"
+#include "zutil.h"
+#include "adler32_p.h"
+
+#define DO1(s1,s2,buf,i)  {(s1) += buf[(i)]; (s2) += (s1);}
+#define DO2(s1,s2,buf,i)  {DO1(s1,s2,buf,i); DO1(s1,s2,buf,i+1);}
+#define DO4(s1,s2,buf,i)  {DO2(s1,s2,buf,i); DO2(s1,s2,buf,i+2);}
+#define DO8(s1,s2,buf,i)  {DO4(s1,s2,buf,i); DO4(s1,s2,buf,i+4);}
+#define DO16(s1,s2,buf)   {DO8(s1,s2,buf,0); DO8(s1,s2,buf,8);}
+
+/* Vector across sum unsigned int (saturate).  */
+inline vector unsigned int vec_sumsu(vector unsigned int __a, vector unsigned int __b) {
+    __b = vec_sld(__a, __a, 8);
+    __b = vec_add(__b, __a);
+    __a = vec_sld(__b, __b, 4);
+    __a = vec_add(__a, __b);
+
+    return __a;
+}
+
+uint32_t adler32_power8(uint32_t adler, const unsigned char* buf, size_t len) {
+    uint32_t s1 = adler & 0xffff;
+    uint32_t s2 = (adler >> 16) & 0xffff;
+
+    /* in case user likes doing a byte at a time, keep it fast */
+    if (UNLIKELY(len == 1))
+        return adler32_len_1(s1, buf, s2);
+
+    /* If buffer is empty or len=0 we need to return adler initial value.  */
+    if (UNLIKELY(buf == NULL))
+        return 1;
+
+    /* This is faster than VSX code for len < 64.  */
+    if (len < 64) {
+        while (len >= 16) {
+            len -= 16;
+            DO16(s1,s2,buf);
+            buf += 16;
+        }
+    } else {
+        /* Use POWER VSX instructions for len >= 64. */
+        const vector unsigned int v_zeros = { 0 };
+        const vector unsigned char v_mul = {16, 15, 14, 13, 12, 11, 10, 9, 8, 7,
+             6, 5, 4, 3, 2, 1};
+        const vector unsigned char vsh = vec_splat_u8(4);
+        const vector unsigned int vmask = {0xffffffff, 0x0, 0x0, 0x0};
+        vector unsigned int vs1 = { 0 };
+        vector unsigned int vs2 = { 0 };
+        vector unsigned int vs1_save = { 0 };
+        vector unsigned int vsum1, vsum2;
+        vector unsigned char vbuf;
+        int n;
+
+        vs1[0] = s1;
+        vs2[0] = s2;
+
+        /* Do length bigger than NMAX in blocks of NMAX size.  */
+        while (len >= NMAX) {
+            len -= NMAX;
+            n = NMAX / 16;
+            do {
+                vbuf = vec_xl(0, (unsigned char *) buf);
+                vsum1 = vec_sum4s(vbuf, v_zeros); /* sum(i=1 to 16) buf[i].  */
+                /* sum(i=1 to 16) buf[i]*(16-i+1).  */
+                vsum2 = vec_msum(vbuf, v_mul, v_zeros);
+                /* Save vs1.  */
+                vs1_save = vec_add(vs1_save, vs1);
+                /* Accumulate the sums.  */
+                vs1 = vec_add(vsum1, vs1);
+                vs2 = vec_add(vsum2, vs2);
+
+                buf += 16;
+            } while (--n);
+            /* Once each block of NMAX size.  */
+            vs1 = vec_sumsu(vs1, vsum1);
+            vs1_save = vec_sll(vs1_save, vsh); /* 16*vs1_save.  */
+            vs2 = vec_add(vs1_save, vs2);
+            vs2 = vec_sumsu(vs2, vsum2);
+
+            /* vs1[0] = (s1_i + sum(i=1 to 16)buf[i]) mod 65521.  */
+            vs1[0] = vs1[0] % BASE;
+            /* vs2[0] = s2_i + 16*s1_save +
+               sum(i=1 to 16)(16-i+1)*buf[i] mod 65521.  */
+            vs2[0] = vs2[0] % BASE;
+
+            vs1 = vec_and(vs1, vmask);
+            vs2 = vec_and(vs2, vmask);
+            vs1_save = v_zeros;
+        }
+
+        /* len is less than NMAX one modulo is needed.  */
+        if (len >= 16) {
+            while (len >= 16) {
+                len -= 16;
+
+                vbuf = vec_xl(0, (unsigned char *) buf);
+
+                vsum1 = vec_sum4s(vbuf, v_zeros); /* sum(i=1 to 16) buf[i].  */
+                /* sum(i=1 to 16) buf[i]*(16-i+1).  */
+                vsum2 = vec_msum(vbuf, v_mul, v_zeros);
+                /* Save vs1.  */
+                vs1_save = vec_add(vs1_save, vs1);
+                /* Accumulate the sums.  */
+                vs1 = vec_add(vsum1, vs1);
+                vs2 = vec_add(vsum2, vs2);
+
+                buf += 16;
+            }
+            /* Since the size will be always less than NMAX we do this once.  */
+            vs1 = vec_sumsu(vs1, vsum1);
+            vs1_save = vec_sll(vs1_save, vsh); /* 16*vs1_save.  */
+            vs2 = vec_add(vs1_save, vs2);
+            vs2 = vec_sumsu(vs2, vsum2);
+        }
+        /* Copy result back to s1, s2 (mod 65521).  */
+        s1 = vs1[0] % BASE;
+        s2 = vs2[0] % BASE;
+    }
+
+    /* Process tail (len < 16).and return  */
+    return adler32_len_16(s1, buf, len, s2);
+}
+
+#endif /* POWER8_VSX_ADLER32 */
--- a/arch/power/adler32_power8.c
+++ b/arch/power/adler32_power8.c
@@ -43,12 +43,6 @@
 #include "zutil.h"
 #include "adler32_p.h"
 
-#define DO1(s1,s2,buf,i)  {(s1) += buf[(i)]; (s2) += (s1);}
-#define DO2(s1,s2,buf,i)  {DO1(s1,s2,buf,i); DO1(s1,s2,buf,i+1);}
-#define DO4(s1,s2,buf,i)  {DO2(s1,s2,buf,i); DO2(s1,s2,buf,i+2);}
-#define DO8(s1,s2,buf,i)  {DO4(s1,s2,buf,i); DO4(s1,s2,buf,i+4);}
-#define DO16(s1,s2,buf)   {DO8(s1,s2,buf,0); DO8(s1,s2,buf,8);}
-
 /* Vector across sum unsigned int (saturate).  */
 inline vector unsigned int vec_sumsu(vector unsigned int __a, vector unsigned int __b) {
     __b = vec_sld(__a, __a, 8);
--- a/arch/power/adler32_power8.c
+++ b/arch/power/adler32_power8.c
@@ -66,91 +66,86 @@ uint32_t adler32_power8(uint32_t adler, const unsigned char* buf, size_t len) {
         return 1;
 
     /* This is faster than VSX code for len < 64.  */
-    if (len < 64) {
+    if (len < 64)
+        return adler32_len_64(s1, buf, len, s2);
+
+    /* Use POWER VSX instructions for len >= 64. */
+    const vector unsigned int v_zeros = { 0 };
+    const vector unsigned char v_mul = {16, 15, 14, 13, 12, 11, 10, 9, 8, 7,
+         6, 5, 4, 3, 2, 1};
+    const vector unsigned char vsh = vec_splat_u8(4);
+    const vector unsigned int vmask = {0xffffffff, 0x0, 0x0, 0x0};
+    vector unsigned int vs1 = { 0 };
+    vector unsigned int vs2 = { 0 };
+    vector unsigned int vs1_save = { 0 };
+    vector unsigned int vsum1, vsum2;
+    vector unsigned char vbuf;
+    int n;
+
+    vs1[0] = s1;
+    vs2[0] = s2;
+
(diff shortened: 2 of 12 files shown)


arch/power/adler32_power8.c
```diff
@@ -0,0 +1,192 @@
...
+ * s1 = s1_16 = s1_(16-1) + c[16] = s1_0 + sum(i=1 to 16)c[i]
+ * s2 = s2_16 = s2_0 + 16*s1 + sum(i=1 to 16)(16-i+1)*c[i]
+ *
+ * After the first iteration we calculate the adler32 checksum for 16 bytes.
+ *
+ * For more background about adler32 please check the RFC:
+ * https://www.ietf.org/rfc/rfc1950.txt
+ */
+
+#include <altivec.h>
```
nmoinvaz, line 41: maybe wrap in a `#ifdef POWER8_VSX_ADLER32`?

functable.c
```diff
@@ -53,6 +53,9 @@ extern uint32_t adler32_ssse3(uint32_t adler, const unsigned char *buf, size_t l
 #ifdef X86_AVX2_ADLER32
 extern uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_t len);
 #endif
+#ifdef POWER8
```
nmoinvaz: `#ifdef POWER8_VSX_ADLER32`

arch/power/adler32_power8.c
```diff
@@ -0,0 +1,192 @@
...
+#include "zbuild.h"
+#include "zutil.h"
+
+/* Largest prime smaller than 65536.  */
+#define BASE 65521U
```
nmoinvaz: Some of these are already defined in `adler32_p.h`. Include that file.

arch/power/adler32_power8.c
```diff
@@ -0,0 +1,192 @@
...
+
+#define DO1(s1,s2,buf,i)  {(s1) += buf[(i)]; (s2) += (s1);}
+#define DO2(s1,s2,buf,i)  {DO1(s1,s2,buf,i); DO1(s1,s2,buf,i+1);}
+#define DO4(s1,s2,buf,i)  {DO2(s1,s2,buf,i); DO2(s1,s2,buf,i+2);}
+#define DO8(s1,s2,buf,i)  {DO4(s1,s2,buf,i); DO4(s1,s2,buf,i+4);}
+#define DO16(s1,s2,buf)   {DO8(s1,s2,buf,0); DO8(s1,s2,buf,8);}
+
+/* Vector across sum unsigned int (saturate).  */
+inline vector unsigned int vec_sumsu (vector unsigned int __a, vector unsigned int __b) {
+  __b = vec_sld(__a, __a, 8);
```
nmoinvaz: This file uses 2 spaces for tabing instead of 4 that is used everywhere else.

arch/power/adler32_power8.c
```diff
@@ -0,0 +1,192 @@
...
+  /* If buffer is empty or len=0 we need to return adler initial value.  */
+  if (buf == NULL)
+      return 1;
+
+  unsigned int s1 = adler & 0xffff;
+  unsigned int s2 = (adler >> 16) & 0xffff;
+
+  /* in case user likes doing a byte at a time, keep it fast */
+  if (len == 1) {
+      s1 += buf[0];
```
nmoinvaz: Use `adler32_len_1`

arch/power/adler32_power8.c
```diff
@@ -0,0 +1,192 @@
...
+          s1 -= BASE;
+      s2 += s1;
+      if (s2 >= BASE)
+          s2 -= BASE;
+      return (s2 << 16) | s1;
+  }
+
+  /* Keep it fast for short length buffers. */
+  if (len < 16) {
+      while (len--) {
```
nmoinvaz: Use `adler32_len_16`

arch/power/adler32_power8.c
```diff
@@ -0,0 +1,192 @@
...
+uint32_t adler32_power8 (uint32_t adler, const unsigned char* buf, size_t len) {
+  /* If buffer is empty or len=0 we need to return adler initial value.  */
+  if (buf == NULL)
+      return 1;
+
+  unsigned int s1 = adler & 0xffff;
+  unsigned int s2 = (adler >> 16) & 0xffff;
+
+  /* in case user likes doing a byte at a time, keep it fast */
+  if (len == 1) {
```
nmoinvaz: ` if (UNLIKELY(len == 1))`

arch/power/adler32_power8.c
```diff
@@ -0,0 +1,192 @@
...
+      if (s1 >= BASE)
+          s1 -= BASE;
+      s2 += s1;
+      if (s2 >= BASE)
+          s2 -= BASE;
+      return (s2 << 16) | s1;
+  }
+
+  /* Keep it fast for short length buffers. */
+  if (len < 16) {
```
nmoinvaz: ` if (UNLIKELY(len < 16))`

arch/power/adler32_power8.c
```diff
@@ -0,0 +1,192 @@
...
+inline vector unsigned int vec_sumsu (vector unsigned int __a, vector unsigned int __b) {
+  __b = vec_sld(__a, __a, 8);
+  __b = vec_add(__b, __a);
+  __a = vec_sld(__b, __b, 4);
+  __a = vec_add(__a, __b);
+
+  return __a;
+}
+
+uint32_t adler32_power8 (uint32_t adler, const unsigned char* buf, size_t len) {
```
nmoinvaz: No space after function name.

arch/power/adler32_power8.c
```diff
@@ -0,0 +1,192 @@
...
+/* NMAX is the largest n such that 255n(n+1)/2 + (n+1)(BASE-1) <= 2^32-1.  */
+
+#define DO1(s1,s2,buf,i)  {(s1) += buf[(i)]; (s2) += (s1);}
+#define DO2(s1,s2,buf,i)  {DO1(s1,s2,buf,i); DO1(s1,s2,buf,i+1);}
+#define DO4(s1,s2,buf,i)  {DO2(s1,s2,buf,i); DO2(s1,s2,buf,i+2);}
+#define DO8(s1,s2,buf,i)  {DO4(s1,s2,buf,i); DO4(s1,s2,buf,i+4);}
+#define DO16(s1,s2,buf)   {DO8(s1,s2,buf,0); DO8(s1,s2,buf,8);}
+
+/* Vector across sum unsigned int (saturate).  */
+inline vector unsigned int vec_sumsu (vector unsigned int __a, vector unsigned int __b) {
```
nmoinvaz: No space after function name.

arch/power/adler32_power8.c
```diff
@@ -0,0 +1,192 @@
...
+      s1 = vs1[0] % BASE;
+      s2 = vs2[0] % BASE;
+   }
+
+  /* Process tail (len < 16).  */
+  while (len--) {
+      s1 += *buf++;
+      s2 += s1;
+  }
+  s1 %= BASE;
```
nmoinvaz: You can use `MOD(s1)` here and elsewhere.

arch/power/adler32_power8.c
```diff
@@ -0,0 +1,192 @@
...
+              buf += 16;
+          }
+          /* Since the size will be always less than NMAX we do this once.  */
+          vs1 = vec_sumsu(vs1, vsum1);
+          vs1_save = vec_sll(vs1_save, vsh); /* 16*vs1_save.  */
+          vs2 = vec_add(vs1_save, vs2);
+          vs2 = vec_sumsu(vs2, vsum2);
+      }
+      /* Copy result back to s1, s2 (mod 65521).  */
+      s1 = vs1[0] % BASE;
```
nmoinvaz: Ie. `MOD(s1)`

configure
```diff
@@ -1392,8 +1392,8 @@ case "${ARCH}" in
 
         if test $without_optimizations -eq 0; then
             if test $HAVE_POWER8 -eq 1; then
-                ARCH_STATIC_OBJS="${ARCH_STATIC_OBJS} power.o slide_hash_power8.o"
-                ARCH_SHARED_OBJS="${ARCH_SHARED_OBJS} power.lo slide_hash_power8.lo"
+                ARCH_STATIC_OBJS="${ARCH_STATIC_OBJS} adler32_power8.o power.o slide_hash_power8.o"
+                ARCH_SHARED_OBJS="${ARCH_SHARED_OBJS} adler32_power8.lo power.lo slide_hash_power8.lo"
                 POWERFLAGS="-DPOWER_FEATURES -DPOWER8"
```
nmoinvaz: Add `-DPOWER_VSX_ADLER32`

functable.c
```diff
@@ -213,6 +216,10 @@ ZLIB_INTERNAL uint32_t adler32_stub(uint32_t adler, const unsigned char *buf, si
     if (x86_cpu_has_avx2)
         functable.adler32 = &adler32_avx2;
 #endif
+#ifdef POWER8
```
nmoinvaz: Add `#ifdef POWER8_VSX_ADLER32`

functable.c
```diff
@@ -213,6 +216,10 @@ ZLIB_INTERNAL uint32_t adler32_stub(uint32_t adler, const unsigned char *buf, si
...
+    if (power_cpu_has_arch_2_07)
+        functable.adler32 = &adler32_power8;
```
nmoinvaz, line 221: Maybe the function should be `adler32_vsx`?  Named after the instruction set rather than the cpu..

arch/power/adler32_power8.c
```diff
@@ -0,0 +1,192 @@
...
+
+  return __a;
+}
+
+uint32_t adler32_power8 (uint32_t adler, const unsigned char* buf, size_t len) {
+  /* If buffer is empty or len=0 we need to return adler initial value.  */
+  if (buf == NULL)
+      return 1;
+
+  unsigned int s1 = adler & 0xffff;
```
nmoinvaz: Used standard int types, `uint32_t`.

arch/power/adler32_power8.c
```diff
@@ -0,0 +1,192 @@
...
+  __b = vec_add(__b, __a);
+  __a = vec_sld(__b, __b, 4);
+  __a = vec_add(__a, __b);
+
+  return __a;
+}
+
+uint32_t adler32_power8 (uint32_t adler, const unsigned char* buf, size_t len) {
+  /* If buffer is empty or len=0 we need to return adler initial value.  */
+  if (buf == NULL)
```
nmoinvaz: Other versions of adler32 have `UNLIKELY(buf == NULL) and it is put below the check for `len == 1`.

image

nmoinvaz (changes requested):
Nice work! I suggested some changes here and there.

functable.c
mscastanho: The requirements for this optimization are the same as for the slide_hash one, in which I used only POWER8. I don't have a strong opinion about either naming scheme. I'd just like to keep them consistent. I see that ARM uses macros like `ARM_NEON_<FUNCTION>` for most optimized functions. x86 also does that for some, but for `slide_hash`, for example, it uses only `X86_SSE2`. Which one do you think we should maintain for POWER? If the `<PROC>_<FEATURE>_<FUNCTION>` is preferred I can switch here and maybe submit a quick PR to update slide_hash.

arch/power/adler32_power8.c
mscastanho: Done. Thanks for the hint!

arch/power/adler32_power8.c
mscastanho: Fixed.

functable.c
racardoso: We choose to use `adler32_power8` because we intend to not support old machines < P8. However, this code may works on P7 (not sure). Should work on P9.

arch/power/adler32_power8.c
mscastanho: Done.

arch/power/adler32_power8.c
mscastanho: Done.

arch/power/adler32_power8.c
mscastanho: Done.

arch/power/adler32_power8.c
mscastanho: Done.

arch/power/adler32_power8.c
mscastanho: Fixed.

arch/power/adler32_power8.c
mscastanho: Fixed.

arch/power/adler32_power8.c
mscastanho: Not really. `MOD(s1)` would be replaced by `s1 %= MOD(s1)`, but here we need to take the remainder of the value stored at the last element inside a vector register and save it to a regular one, so I don't think we can use `MOD()` here.

functable.c
mscastanho: We also opted for the `*_powerX` naming scheme because that better reflects a set of features on POWER. The set of VSX instructions is usually incremented with each new version of the Power ISA, so the best way to distinguish an specific feature/instruction set is by the ISA version, which usually maps to a processor version (e.g. P7, P8, P9). `adler32_power8` indicates that P8 (ISA 2.07) is the baseline, although it also runs on later processor versions, as the instruction sets are incremental. So I believe that keeping the `powerX` suffix would be best.

arch/power/adler32_power8.c
mscastanho: Done.

arch/power/adler32_power8.c
mscastanho: Done.

arch/power/adler32_power8.c
mscastanho: Fixed.

functable.c
nmoinvaz: Yeah I know some of them need to be cleaned up. I think the `<PROC>_<FEATURE>_<FUNCTION>` method is the best to be consistent.

arch/power/adler32_power8.c
nmoinvaz: Ahh

functable.c
nmoinvaz: Ok sounds good.

functable.c
mscastanho: Done.

configure
mscastanho: Done.

arch/power/adler32_power8.c
mscastanho: Done.

functable.c
mscastanho: I agree. Just switched to `POWER8_VSX_ADLER32`

mscastanho:
Changes:
- Rebase against current `develop`
- Use `POWER8_VSX_ADLER32` macro as suggested
- Fixed an invalid memory access on ppc and ppc64 found when running with `WITH_SANITIZERS`

arch/power/adler32_power8.c
```diff
@@ -0,0 +1,176 @@
...
+ */
+
+#ifdef POWER8_VSX_ADLER32
+
+#include <altivec.h>
+#include "zbuild.h"
+#include "zutil.h"
+#include "adler32_p.h"
+
+#define DO1(s1,s2,buf,i)  {(s1) += buf[(i)]; (s2) += (s1);}
```
nmoinvaz: This looks like the same #defines in adler32.c. Consider moving them into adler32_p.h and using them in both places? Or maybe this can be in another commit..

arch/power/adler32_power8.c
```diff
@@ -0,0 +1,176 @@
...
+    if (UNLIKELY(buf == NULL))
+        return 1;
+
+    /* Keep it fast for short length buffers. */
+    if (UNLIKELY(len < 16))
+        return adler32_len_16(s1, buf, len, s2);
+
+    /* This is faster than VSX code for len < 64.  */
+    if (len < 64) {
+        while (len >= 16) {
```
nmoinvaz, line 124: Consider creating an `adler32_len_64` and using it like `adler32_len_16`. This could also be another commit.

nmoinvaz (approved):
LGTM. I suggested some changes. which can be done now or in the future

arch/power/adler32_power8.c
mscastanho: Done.

arch/power/adler32_power8.c
mscastanho: Agreed. Done.

Dead2:
Looks great :)

mscastanho:
Rebased against current develop.

adler32_p.h
```diff
@@ -74,4 +80,20 @@ static inline uint32_t adler32_len_16(uint32_t adler, const unsigned char *buf,
...
 }
 
+static inline uint32_t adler32_len_64(uint32_t adler, const unsigned char *buf, size_t len, uint32_t sum2) {
+    while (len >= 16) {
+        len -= 16;
+        DO16(adler, sum2, buf);
+        buf += 16;
+    }
+    /* Process tail (len < 16).  */
+    while (len--) {
```
nmoinvaz: Perhaps it is possible to do `return adler32_len_16(adler, buf, sum2);` here? I'm not sure..

arch/power/adler32_power8.c
```diff
@@ -0,0 +1,165 @@
...
+    /* in case user likes doing a byte at a time, keep it fast */
+    if (UNLIKELY(len == 1))
+        return adler32_len_1(s1, buf, s2);
+
+    /* If buffer is empty or len=0 we need to return adler initial value.  */
+    if (UNLIKELY(buf == NULL))
+        return 1;
+
+    /* Keep it fast for short length buffers. */
+    if (UNLIKELY(len < 16))
```
nmoinvaz: Do you still need this? Since we already do `len < 64` below..

arch/power/adler32_power8.c
```diff
@@ -0,0 +1,165 @@
...
+        vs1_save = vec_sll(vs1_save, vsh); /* 16*vs1_save.  */
+        vs2 = vec_add(vs1_save, vs2);
+        vs2 = vec_sumsu(vs2, vsum2);
+    }
+    /* Copy result back to s1, s2 (mod 65521).  */
+    s1 = vs1[0] % BASE;
+    s2 = vs2[0] % BASE;
+
+    /* Process tail (len < 16).  */
+    while (len--) {
```
nmoinvaz: I wonder if it is possible to use `adler32_len_16` here too.. and the same goes for the other adler32 implementations..

nmoinvaz (commented):
I made some more feedback. Things are very good. Some of your code comments made me think about other things which I noted.

adler32_p.h
mtl1979: If we drop support for NO_DIVIDE, we can simplify these a lot as then we don't need to worry about CHOP().

nmoinvaz: That might be a good idea for a separate PR.

mtl1979: Problem with using adler32_len_16 is that when/if NO_DIVIDE is set, the result can overflow. Modulo 65521 is essentially high 16 bits multiplied by 15 and added to low 16 bits... The result will be less than 2*65521, so it can do just one substraction to get same result as real nodulo.

mscastanho: yes, I believe those are equivalent. The only difference is that `adler32_len_16` does a 
```
if (adler >= BASE)
      adler -= BASE
```
in the end, while that snippet does a `MOD(adler)`. But since the maximum increment to `adler` in the function will be fairly small (the sum of each byte in the buf, so <= 64*255 ?) then in this case both operations will be effectively the same.

arch/power/adler32_power8.c
mscastanho: Not really. By reusing `adler32_len_16` inside `adler32_len_64`, this check will be done implicitly. I'll remove this.

I'm just not sure how rare are < 64 byte chunks, i.e. if we should use `UNLIKELY` for `len < 64`.

adler32_p.h
nmoinvaz: Yeah I'm not certain either. It might be same rarity though -- I think at least in `read_buf` buf size is typically multiple of 1024 at least...

arch/power/adler32_power8.c
nmoinvaz: Yeah I'm not certain either. It might be same rarity though -- I think at least in `read_buf` buf size is typically multiple of 1024 at least... so `UNLIKELY` might be a good idea too.

arch/power/adler32_power8.c
mscastanho: For `adler32_power8` it can indeed. For other implementations, we would only need to make sure all do `MOD(adler)` before calling `adler32_len_16`, since it does a subtraction instead of a proper mod, which could lead to problems if the value previously accumulated in `adler` was too large. But I'd prefer to leave that change to a separate PR.

nmoinvaz: Sounds good

mscastanho:
Added suggestions from @nmoinvaz

nmoinvaz (approved):
Looks great. No more comments. :-)

Dead2:
Thanks!

mscastanho:
Thanks @nmoinvaz, @Dead2!

PR #648, opened 2020-06-19
Add GitHub Actions workflow for creating releases.

nmoinvaz:
Just tag a commit with the version number and a description and it will create a GitHub Actions release. 

Here is an example of the output:
https://github.com/nmoinvaz/zlib-ng/releases/tag/1.0.0.8

Remember to do `git push --tags`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,79 @@
+name: CI Release
+on:
+  push:
+    tags:
+    - '*'
+jobs:
+  ci-cmake:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        name: [
+          Windows MSVC Win32,
+          Windows MSVC Win32 Compat,
+          Windows MSVC Win64,
+          Windows MSVC Win64 Compat
+        ]
+        include:
+          - name: Windows MSVC Win32
+            os: windows-latest
+            compiler: cl
+            cmake-args: -A Win32
+            deploy-name: win32
+
+          - name: Windows MSVC Win32 Compat
+            os: windows-latest
+            compiler: cl
+            cmake-args: -A Win32 -DZLIB_COMPAT=ON
+            deploy-name: win32-compat
+
+          - name: Windows MSVC Win64
+            os: windows-latest
+            compiler: cl
+            cmake-args: -A x64
+            deploy-name: win64
+
+          - name: Windows MSVC Win64 Compat
+            os: windows-latest
+            compiler: cl
+            cmake-args: -A x64 -DZLIB_COMPAT=ON
+            deploy-name: win64-compat
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v1
+
+    - name: Set environment variables
+      shell: bash
+      run:  echo "::set-env name=tag::${GITHUB_REF/refs\/tags\//}"
+
+    - name: Generate project files
+      run: |
+        cmake . ${{ matrix.cmake-args }} -DCMAKE_BUILD_TYPE=Release -DWITH_GZFILEOP=ON -DZLIB_ENABLE_TESTS=ON -DCMAKE_INSTALL_PREFIX=out
+      env:
+        CC: ${{ matrix.compiler }}
+        CI: true
+
+    - name: Compile source code
+      run: |
+        cmake --build . --config Release --target install
+
+    - name: Package release (Windows)
+      if: runner.os == 'Windows'
+      run: |
+        cd out
+        7z a -tzip ../zlib-ng-${{ matrix.deploy-name }}.zip bin include lib ../LICENSE.md ../README.md
+
+    - name: Upload release (Windows)
+      uses: svenstaro/upload-release-action@v1-release
+      if: runner.os == 'Windows'
+      with:
+        asset_name: zlib-ng-${{ matrix.deploy-name }}.zip
+        file: zlib-ng-${{ matrix.deploy-name }}.zip
+        tag: ${{env.tag}}
+        repo_token: ${{ secrets.GITHUB_TOKEN }}
+        overwrite: true
+      env:
+        GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,7 +51,7 @@ jobs:
 
     - name: Generate project files
       run: |
-        cmake . ${{ matrix.cmake-args }} -DCMAKE_BUILD_TYPE=Release -DWITH_GZFILEOP=ON -DZLIB_ENABLE_TESTS=ON -DCMAKE_INSTALL_PREFIX=out
+        cmake . ${{ matrix.cmake-args }} -DCMAKE_BUILD_TYPE=Release -DWITH_GZFILEOP=ON -DZLIB_ENABLE_TESTS=ON -DCMAKE_INSTALL_PREFIX=out -DINSTALL_UTILS=ON
       env:
         CC: ${{ matrix.compiler }}
         CI: true
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,6 +103,7 @@ elseif(BASEARCH_X86_FOUND)
     option(WITH_SSE4 "Build with SSE4" ON)
     option(WITH_PCLMULQDQ "Build with PCLMULQDQ" ON)
 endif()
+option(INSTALL_UTILS "Copy minigzip and minideflate during install" OFF)
 
 mark_as_advanced(FORCE
     ZLIB_DUAL_LINK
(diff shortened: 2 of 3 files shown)


gdevenyi:
I think what you've done here can be done with cpack as part of the cmake build. Should be able to produce, deb, rpm and tar.gz

nmoinvaz:
That is an interesting idea. I've never used `cpack` before, but I'm also not sure we want to create binary installers.

gdevenyi:
https://cmake.org/cmake/help/latest/manual/cpack-generators.7.html lists generators, and a little tutorial:
https://gitlab.kitware.com/cmake/community/-/wikis/doc/cpack/Packaging-With-CPack

Dead2:
I want us to make a windows binary release. It should include both dll, minigzip and include-files needed for compiling against zlib-ng. Both a native zlib-ng and a zlib-compat package.

If we get the zlib-compat package actually ABI compatible on windows, people could perhaps use our dll in place of zlib.dll in programs/games they use but do not have source code for. I am not sure whether we are close to being able to do that or not. Having automated testing for that would be nice.

Linux etc should not get binary packages, those would require a lot more variations and would likely increase our maintenance load by a lot. We should document the `cpack` workflow though, so people can build their own packages and get them installed the proper way (`make install` is never a good idea on linux, unless you install to a separate folder, because uninstallation is difficult since you don't know what files were installed and where).

nmoinvaz:
@Dead2 did you see the output that this produces from the link in the description? The assets already have everything you mentioned, broken into compat versus non-compat for each platform.

I'm not sure about having a windows installer executable, since libraries are not something are generally installed like on Linux. 

Are you saying you don't want to distribute for the rest of the platforms any binaries?

Dead2:
Yes, I saw the release you linked, and while that is neat, it will become a pain to maintain and is unfortunately not really a good idea. Linux and most OSes other than windows have good ways to install programs and libraries from repos, and when you install something into those systems without going through those systems, it can very quickly become a mess.
Especially for libraries, this can become a problem, or can even break the entire system so it won't boot any more.
For that reason and others, it is also not customary to provide binaries for *nix.

Windows is different, in good and bad ways, but it has different expectations and ways of doing stuff. And the sum of this is that you also make it less likely you will break your whole system.

Yes, I am saying I don't want to distribute binaries for any other OS than windows.
And No, no installer for windows, where would you even install zlib-ng to on windows? There is no standard place for that like there is in *NIX.

We can provide spec files that can help people build packages though, and/or document the cpack method.

mtl1979:
In Windows, the only place to install .dll files is under C:\Windows\WinSxS ... For development files (import library, static library and headers), the correct place would be inside the appropriate Visual C++ directory as it won't try to search anywhere else... This means we would need one version/copy for each supported Visual C++ version.

nmoinvaz:
I think there is some confusion. `cpack` is only for creating binary installers - this is not what I am currently using. I can remove the non-Windows platform distributables. Is there anything wrong with what we are distributing for Windows ones?

mtl1979:
\share\pkgconfig\zlib.pc is useless for Windows... It even has wrong paths...

nmoinvaz:
> If we get the zlib-compat package actually ABI compatible on windows, people could perhaps use our dll in place of zlib.dll in programs/games they use but do not have source code for. I am not sure whether we are close to being able to do that or not. Having automated testing for that would be nice.

Doesn't the `libpngtest` CI workflow cover a lot of the compatibility testing?

mtl1979:
We could just overwrite the external headers from zlib-ng with equivalent ones from stock zlib and test how far the compile goes... Essentially that would test if some function signatures are incompatible, but wouldn't test for missing macros...

Dead2:
cpack creates installers, yes, so we could document how to use cpack to create an installer after they compile it (or drop it entirely). That is a big difference to actually distributing binary installers/packages.

ABI compatability is not the same as API compatability. If someone replaces the zlib.dll/lzip.so file with one from our compat-mode, it is not likely to work (unfortunately), especially since our deflate struct is bigger than the original zlib one is.
If someone recompiles their program against one of our dlls+header files however, it should work fine.
So I do not think the libpngtest covers the ABI-part of this.

Replacing the libz.so file of a system would require it to be ABI-compatible with the exact variant installed on the system and that all other programs are compiled against, otherwise it is likely that the whole system becomes completely useless and in need of a reinstall (or a manual rescue session).


We should provide the source code.
And we can provide the compiled windows dll files (compat/native, 32/64bit) along with their respective header files.
No installers.
There is a reason why we warn against using ´make install´, it is super-easy to break the system, and then who is to blame?
Better to leave the distros (or third-party repos) to make rpms/debs that have been through QA on each respective distro and version, before end-users install them.

nmoinvaz:
Would using `minigzip` (compiled against zlib_shared) from madler/zlib using our zlib-compat dll, be a sufficient test?

Dead2:
@nmoinvaz Only for the parts of the of the api that minigzip uses, and that is a very small part unfortunately.
Therefore, we should rather have a proper ABI-comparison tool in CI, those at least take care of function signatures, but I am not sure they can notice things like struct size being different (nginx for example includes a fallback when it encounters a zlib that uses a struct bigger than what it pre-allocs for zlib, and yes that is rather fragile).

nmoinvaz:
Maybe that was the comparison tool that @dankegel had mention once before?

dankegel:
Yes.  libabigail comes with an abi comparator. I will put together a pull request that shows how it could be used.

nmoinvaz:
> We should provide the source code.

GitHub Actions automatically adds the source for the commit/tag to the release.

dankegel:
IMHO a source release is completely sufficient. 

Windows developers will get binaries via vcpkg.

nmoinvaz:
I have updated the yaml to support only Windows binaries if that is still wanted. Once the result is reviewed etc I can update this PR. See this release tag:
https://github.com/nmoinvaz/zlib-ng/releases/tag/1.0.0.12

Dead2:
Why does the yaml list all those other distros etc if it only does windows releases?
Makes it confusing to review..

nmoinvaz:
Like I said in my last comment, that once the result is reviewed (the binaries at the tagged release link), then I can update the PR. Because there was talk of not doing binaries at all and doing source code releases only. So I didn't want to go to that extra work if that is what is ultimately decided, but I have updated the PR now.

dankegel:
Removing the linux builds was a good move.

Binary windows releases have some chance of being relevant.
But given the rise of vcpkg, conan, cmake's external project support, and meson's wrapdb, 
it's really tempting to drop all binary releases.
If I'm wrong, and people want them bad enough, it'll be easy to add them later.

Dead2:
@nmoinvaz Sorry, I didn't understand that you wanted the zip files reviewed first, I thought you meant the yaml when you said "result" 😄

The yaml looks a lot better now.

.zip output looks pretty good, but I think we can delete everything under /share, since man files and pkg files are AFAIK useless on windows, unless you use cygwin, but then that .zip file is probably not what you want anyhow.
We should however include the license file and some kind of documentation, just readme.md renamed to .txt and unix2dos'ed or something else?

dankegel:
I routinely use .pc files on Windows, but it'd be ok to delete them until we have a test to verify they work.

nmoinvaz:
@Dead2 did you take a look at the zips at https://github.com/nmoinvaz/zlib-ng/releases/tag/1.0.0.12? They don't have any share folders. 

I can add the license. I don't think it is necessary to rename to .txt. Markdown format and extension is popular enough. Also I think unix2dos is Linux only maybe?

Dead2:
@nmoinvaz Oh, no, I only checked the one in the link above.

Hmm, notepad only recently got the ability to open files with unix-style fileendings, but come to think of it, does perhaps git already convert them on checkout?

What text editors open markdown files? I was not aware that they could.

nmoinvaz:
Yes, git automatically converts line endings to Windows if you have it configured that way. I am using Windows 10.2004 and it opens fine (however that is the latest public version of 10). 

VSCode/Atom can which is what a lot of developers are using now-a-days.

One thing to consider, even if we convert to different line endings, the file still has markdown styling.

Dead2:
Most of the markdown looks quite good in a regular text editor (like Nano), as long as
it uses a fixed-width font. The CI badges do not look so nice though, not sure what we can do about those easily.

Honestly, I think it is OK, it is not awesome but I do not see it as a problem/blocker. And it can be improved on by a later PR.

nmoinvaz:
@Dead2 please review the artifacts at this tag:
https://github.com/nmoinvaz/zlib-ng/releases/tag/1.0.0.13

Do you think we should have cmake option for both `INSTALL_MINIGZIP` and `INSTALL_MINIDEFLATE` or just a single cmake option?

Dead2:
@nmoinvaz I honestly think one for both is fine. `INSTALL_UTILS` or something like that perhaps?

Contents of the .zip files look good to me.

nmoinvaz:
I have changed it to `INSTALL_UTILS` cmake option.
See https://github.com/nmoinvaz/zlib-ng/releases/tag/1.0.0.15

PR #656, opened 2020-06-25
Quick fixes to make POWER code more consistent

mscastanho:
Currently unaligned reads are not being properly enabled on native powerpc64le machines when using cmake (configure is fine). In such cases `cmake/detect-arch.c` is returning an abbreviated CPU string:
https://github.com/zlib-ng/zlib-ng/blob/738a6e5b07c0495b8652c77b4879145c083ea49f/cmake/detect-arch.c#L35-L44

While the check to enable the feature is expecting a full form:
https://github.com/zlib-ng/zlib-ng/blob/738a6e5b07c0495b8652c77b4879145c083ea49f/CMakeLists.txt#L272
https://github.com/zlib-ng/zlib-ng/blob/738a6e5b07c0495b8652c77b4879145c083ea49f/CMakeLists.txt#L290

To make it consistent with `configure`, switch to using the longer form everywhere.

Also, as discussed previously with @nmoinvaz (https://github.com/zlib-ng/zlib-ng/pull/647#discussion_r442969611), I'm also adding a new `POWER8_VSX_SLIDEHASH` macro to keep the naming scheme consistent. This is just a code organization improvement and does not add any functional changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -705,6 +705,7 @@ if(WITH_OPTIM)
             add_definitions(-DPOWER8)
             add_definitions(-DPOWER_FEATURES)
             add_definitions(-DPOWER8_VSX_ADLER32)
+            add_definitions(-DPOWER8_VSX_SLIDEHASH)
             set(ZLIB_POWER8_SRCS
                 ${ARCHDIR}/adler32_power8.c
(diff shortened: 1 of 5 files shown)


nmoinvaz (approved):
Looks good

Issue #654, opened 2020-06-23
minigzip fails if compiled with USE_MMAP=1 and creating compression level 1


neurolabusc:
minigzip compiled with USE_MMAP=1 creates corrupted gz files for compression level 1.

```
export CFLAGS="-DUSE_MMAP=1" 
git clone https://github.com/zlib-ng/zlib-ng
cd zlib-ng/
./configure
make
./minigzip -1 libz-ng.a 
gzip -d libz-ng.a.gz
     gzip: libz-ng.a.gz: invalid compressed data--format violated
./minigzip -d libz-ng.a.gz 
    ./minigzip: libz-ng.a.gz: invalid stored block lengths
```

nmoinvaz:
This should be resolved now that #657 is merged. We can reopen if needed..

PR #657, opened 2020-06-26
Fixes to deflate_quick algorithm when compiling with USE_MMAP

nmoinvaz:
This fixes the bug reported in #654 with deflate_quick when compiling with `USE_MMAP`

* Fixed avail_out == 0 conditional not returning `need_more` in deflate_quick.
* Fixed ending block when returning `need_more` caused problems with inflate.

So instead of ending the block each time the function returns, we check upon start to see if it is Z_FINISH, and if it is, and the last block not yet started it will end the previous  block and start the last block.

I have tested this against pigz-bench-python and with USE_MMAP and the x-ray file from silesia corpus.

### DIFF
--- a/deflate_quick.c
+++ b/deflate_quick.c
@@ -28,7 +28,7 @@ extern const ct_data static_dtree[D_CODES];
 
 #define QUICK_START_BLOCK(s, last) { \
     zng_tr_emit_tree(s, STATIC_TREES, last); \
-    s->block_open = 1; \
+    s->block_open = 1 + last; \
     s->block_start = s->strstart; \
 }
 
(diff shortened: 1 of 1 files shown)


Dead2:
```

 Runs: 20
 Levels: 1
 Trimworst: 10
```

```
# Baseline 738a6e5b07c0495b8652c77b4879145c083ea49f
  93246    1304      32   94582   17176 libz-ng.so.1

 1      57.881%    1.754/1.779/1.790       1.082/1.098/1.106  163950540

 1      57.881%    1.772/1.781/1.790       1.077/1.097/1.118  163950540
```

```
## PR/657 bb5dba370012cbf1c322c3e8cfb5eda2671dc2ee
  93454    1304      32   94790   17246 libz-ng.so.1

 1      57.870%    1.770/1.793/1.810       1.079/1.100/1.114  163918126
```

This makes a big change to compressed output as well, the output is now 32414 bytes smaller.
Are those changes well enough understood?

nmoinvaz:
Yes, I think the changes only relate to when the start and end blocks are written. The initial problem with `deflate_quick` was partially related to the last block not being written properly in some cases. So I had changed it to write start and end blocks each time it entered and exit the function. This probably caused it to write more frequently than it should have. However, that had some problems as inflate did not like seeing an end block in certain instances when the function does not return `block_done` or `finish_done`. So instead, it is now changed that at the top of the loop it will write the end of previous block if the function is processing the last block and start of block hasn't been written. 

And in my last commit I thought about a potential problem, where an end block would be written during last block processing if it returns `finish_started`. I have tested it against both tests which previously produced problems for `deflate_quick`.

PR #659, opened 2020-06-28
Make shared library produced by cmake and configure identical

### LINKED ISSUE #638: Hard to verify that cmake and configure do the same thing

dankegel:
zlib shared libraries really ought to be identical regardless of whether you build them with cmake or configure.
This pull request verifies that, and makes it so, on four platforms, at least for plain old and --zlib-compat builds.

Here's what it took (beyond the static library fix landed recently):
- canonicalize CHOST in configure
- rearrange source file order in cmake and configure to match (at least for i686, amd64, aarch64, ppc, and ppc64le)
- don't link shared libraries explicitly with -lc (shouldn't be needed; caused differences between configure and cmake)
- on mac, do like configure, and don't use @rpath
- update pkgcheck.sh to verify that cmake and configure create identical shared libraries

Along the way, it turned out that the configure sections for 32 and 64 bit x86 were
essentially identical, so this pull request merges them to reduce chance of
accidental divergence later.

Fixes #638 and #666 

Known issues left for future pull requests:
- There may still be differences between cmake and configure on architectures not yet tested by github/workflows/pkgcheck.yaml
- --native and -DWITH_NATIVE_INSTRUCTIONS don't do quite the same thing (see #697)


dankegel:
Rebased.

dankegel:
This is ready to go, I think.

nmoinvaz:
I think the problem with configure script for PPC detection when cross compile should be fixed.

And configure and cmake should be set to -O2. 

That should fix some workarounds that were needed I believe, also you won’t need to check CMakeCache.txt.

configure
```diff
@@ -1173,11 +1173,22 @@ case "${ARCH}" in
...
-            CFLAGS="${CFLAGS} -DX86_CPUID -DX86_SSE2 -DX86_SSE2_CHUNKSET -DX86_SSE42_CRC_HASH"
-            SFLAGS="${SFLAGS} -DX86_CPUID -DX86_SSE2 -DX86_SSE2_CHUNKSET -DX86_SSE42_CRC_HASH"
+            CFLAGS="${CFLAGS} -DX86_CPUID"
+            SFLAGS="${SFLAGS} -DX86_CPUID"
+            ARCH_STATIC_OBJS="${ARCH_STATIC_OBJS} x86.o"
+            ARCH_SHARED_OBJS="${ARCH_SHARED_OBJS} x86.lo"
+
+            # if(WITH_SSE4 AND (HAVE_SSE42CRC_INLINE_ASM OR HAVE_SSE42CRC_INTRIN))
+            CFLAGS="${CFLAGS} -DX86_CPUID -DX86_SSE42_CRC_HASH"
+            SFLAGS="${SFLAGS} -DX86_CPUID -DX86_SSE42_CRC_HASH"
```
Dead2: Duplicates -DX86_CPUID from the above block?

test/pkgcheck.sh
```diff
@@ -40,24 +100,59 @@ cd btmp1
...
-    fi
-    ;;
   esac
-  ../configure
+  # Use same optimization level as cmake did.
+  CFLAGS="$(awk -F= '/CMAKE_C_FLAGS_RELEASE:STRING=/ {print $2}' < ../btmp2/CMakeCache.txt)"
+  export CFLAGS
+  # Hack: oddly, given CHOST=powerpc-linux-gnu, configure concludes uname is gnu,
+  # causing it to set LDSHAREDFLAGS to a different value than cmake uses.
+  # So override it when appropriate.  FIXME
```
Dead2: Perhaps fix this powerpc-detection in configure instead?

test/pkgcheck.sh
```diff
@@ -40,21 +43,35 @@ cd btmp1
...
   make install
 cd ..
 
-if diff --exclude '*.so*' --exclude '*.a' -Nur pkgtmp1 pkgtmp2
+repack_ar() {
+  if ! cmp --silent pkgtmp1/usr/local/lib/libz$suffix.a pkgtmp2/usr/local/lib/libz$suffix.a
+  then
+    echo "Warning: libz$suffix.a does not match.  Assuming ar needs -D option.  Unpacking..."
+    cd pkgtmp1; ar x usr/local/lib/libz$suffix.a; rm usr/local/lib/libz$suffix.a; cd ..
+    cd pkgtmp2; ar x usr/local/lib/libz$suffix.a; rm usr/local/lib/libz$suffix.a; for a in *.c.o; do b=$(echo $a | sed 's/\..*//'); mv $a $b.o; done; cd ..
```
mscastanho: The inner part of that for loop could be simplified to just `mv $a ${a%%.*}.o`. But to be honest I'm not sure if that's portable to other shells too. If not, just disconsider.

.github/workflows/pkgcheck.yml
```diff
@@ -8,14 +8,30 @@ jobs:
...
             packages: ninja-build
 
+          - name: Ubuntu GCC AARCH64
+            os: ubuntu-latest
+            chost: aarch64-linux-gnu
+            compiler: aarch64-linux-gnu-gcc
+            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-aarch64.cmake
+            packages: ninja-build qemu gcc-aarch64-linux-gnu libc-dev-arm64-cross
+
+          - name: Ubuntu GCC PPC
```
mscastanho: I believe we should switch this to PPC64LE. This is the ABI followed by recent POWER processors and I believe not many recent distros still support ppc or ppc64, so if we need to choose one, ppc64le is the way to go since it will cover most of the current ppc user base.

test/pkgcheck.sh
```diff
@@ -1,10 +1,67 @@
...
+$ sudo apt install ninja-build diffoscope qemu gcc-arm-linux-gnueabihf libc-dev-armel-cross
+$ export CHOST=arm-linux-gnueabihf
+$ export CMAKE_ARGS="-DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-arm.cmake -DCMAKE_C_COMPILER_TARGET=${CHOST}"
+
+aarch64:
+$ sudo apt install ninja-build diffoscope qemu gcc-aarch64-linux-gnu libc-dev-arm64-cross
+$ export CHOST=aarch64-linux-gnu
+$ export CMAKE_ARGS="-DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-aarch64.cmake -DCMAKE_C_COMPILER_TARGET=${CHOST}"
+
+ppc:
```
mscastanho: Some as I commented above. Probably better to use ppc64le.

test/pkgcheck.sh
mscastanho: I believe the issue is on `configure:26`:
https://github.com/zlib-ng/zlib-ng/blob/9abe04b926ef688efa3d4f464209246a22afb01d/configure#L25-L28
When `CHOST` is something like `powerpc-linux-gnu`, it will match the second pattern, which will select the last string after the `-`. But this issue happens with any `CHOST` value in the form `first-second-third`, such as `arm-linux-gnueabihf` and `aarch64-linux-gnu`. All these match the second pattern.

That change may deserve a separate PR though. For now I believe setting `--uname=linux` will suffice.

configure
```diff
@@ -69,7 +69,7 @@ fi
 # set defaults before processing command line options
 LDCONFIG=${LDCONFIG-"ldconfig"}
 LDFLAGS=${LDFLAGS-"-L."}
-LDSHAREDLIBC="${LDSHAREDLIBC--lc}"
+LDSHAREDLIBC="${LDSHAREDLIBC}"
```
mscastanho, line 74: Looks good. If somebody actually needs `-lc` they can just set it before calling configure.

mscastanho (commented):
Some comments here and there but overall the PR looks good. My main suggestion is to use ppc64le instead of ppc in the CI tests.

mscastanho:
Also an additional comment in the `Usage` string of `test/pkgcheck.sh` saying that the example commands are for Ubuntu/Debian-like distros would be nice. Since the instructions use `apt` and the names of the dependency packages may change on other distros. Something simple like `For example, on Ubuntu:`.

test/pkgcheck.sh
mscastanho: Just opened a new issue about this so it doesn't get forgotten: https://github.com/zlib-ng/zlib-ng/issues/666

configure
dankegel: Thanks.

.github/workflows/pkgcheck.yml
dankegel: Happy to add ppc64le, but isn't a lot of embedded powerpc going to be 32 bit for the foreseeable future?  So maybe keep the 32 bit one... I have a Powermac G4, and I'm biased :-)

test/pkgcheck.sh
dankegel: I added a commit to address this.

test/pkgcheck.sh
dankegel: That syntax looks strange to me :-)  But yes, that seems portable and would be shorter.  If I get a vote or two 
for the %% syntax, I'll switch to it.

configure
dankegel: Agreed.

test/pkgcheck.sh
dankegel: Added.

dankegel:
Resolved except maybe for the %% and the "on ubuntu" wording in test/pkgcheck.sh comments.
Tomorrow morning I'll push an update for those and anything else that comes up tonight.

nmoinvaz:
@dankegel have you encountered the issue with `-mfpu=neon` being set on all files when Neon is present in configure, but in CMake we only set it per file that uses neon?

dankegel:
Yes.  That's why I wimped out and only test aarch64, not 32 bit arm...

nmoinvaz:
See #669 and #668. Maybe you can cherry pick those and see if they solve the problems?

.github/workflows/pkgcheck.yml
mscastanho: Agreed. The more the merrier =)

.github/workflows/pkgcheck.yml
```diff
@@ -8,14 +8,37 @@ jobs:
...
+            compiler: powerpc-linux-gnu-gcc
+            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-powerpc.cmake
+            packages: ninja-build qemu gcc-powerpc-linux-gnu libc-dev-powerpc-cross
+
+          - name: Ubuntu GCC PPC64LE
+            os: ubuntu-latest
+            chost: powerpc64le-linux-gnu
+            compiler: powerpc64le-linux-gnu-gcc
+            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-powerpc64le.cmake
+            packages: ninja-build qemu gcc-powerpc64le-linux-gnu libc6-dev-ppc64-powerpc-cross
```
mscastanho: @dankegel I believe the glibc package you need here is actually `libc6-dev-ppc64el-cross`

.github/workflows/pkgcheck.yml
```diff
@@ -8,14 +8,37 @@ jobs:
...
+            compiler: aarch64-linux-gnu-gcc
+            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-aarch64.cmake
+            packages: ninja-build qemu gcc-aarch64-linux-gnu libc-dev-arm64-cross
+
+          - name: Ubuntu GCC PPC
+            os: ubuntu-latest
+            chost: powerpc-linux-gnu
+            compiler: powerpc-linux-gnu-gcc
+            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-powerpc.cmake
+            packages: ninja-build qemu gcc-powerpc-linux-gnu libc-dev-powerpc-cross
```
mscastanho: glibc packages on Ubuntu seem to start with `libc6`. So this should be `libc6-dev-powerpc-cross`

.github/workflows/pkgcheck.yml
```diff
@@ -8,14 +8,37 @@ jobs:
...
             os: ubuntu-latest
             compiler: gcc
             packages: ninja-build
 
+          - name: Ubuntu GCC AARCH64
+            os: ubuntu-latest
+            chost: aarch64-linux-gnu
+            compiler: aarch64-linux-gnu-gcc
+            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-aarch64.cmake
+            packages: ninja-build qemu gcc-aarch64-linux-gnu libc-dev-arm64-cross
```
mscastanho: `libc6-dev-arm64-cross` here?

.github/workflows/pkgcheck.yml
```diff
@@ -8,14 +8,37 @@ jobs:
       fail-fast: false
       matrix:
         name: [
-          Ubuntu GCC
+          Ubuntu GCC,
+          Ubuntu GCC AARCH64,
+          Ubuntu GCC PPC
```
mscastanho: @dankegel Shouldn't you add `Ubuntu GCC PPC64LE` to this list too?

test/pkgcheck.sh
```diff
@@ -1,10 +1,72 @@
...
+$ export CHOST=aarch64-linux-gnu
+$ export CMAKE_ARGS="-DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-aarch64.cmake -DCMAKE_C_COMPILER_TARGET=${CHOST}"
+
+ppc (32 bit big endian):
+$ sudo apt install ninja-build diffoscope qemu gcc-powerpc-linux-gnu libc-dev-powerpc-cross
+$ export CHOST=powerpc-linux-gnu
+$ export CMAKE_ARGS="-DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-powerpc.cmake"
+
+ppc64le:
+$ sudo apt install ninja-build diffoscope qemu gcc-powerpc64le-linux-gnu libc6-dev-ppc64-powerpc-cross
```
mscastanho: Should be `libc6-dev-ppc64el-cross`. And also replace `libc` by `libc6` on the other package names.

mscastanho (commented):
@dankegel just some small comments about package names. I think the builds passed because the glibc cross packages are in the `recommends` list for the gcc package, but having the correct package names is ideal.
Also, your fix for the `CHOST` issue looks good. Now `uname` is correctly being detected as `linux` for those failing cases.

.github/workflows/pkgcheck.yml
dankegel: Yes, looks like the libc- one is an alias, but it's always less confusing to use the real package name.

.github/workflows/pkgcheck.yml
dankegel: thanks.  yesterday was not my best day :-)

.github/workflows/pkgcheck.yml
dankegel: I used same package name as in cmake.yml.  Evidently the one missing the 6 is an alias, and it pulls in the right package.  Both files probably should use the real package name.

.github/workflows/pkgcheck.yml
dankegel: yep.

dankegel:
cherry-picking #668 and #669 still leaves us with configure applying -march=armv8-a+crc hither and yon for
arm-linux-gnueabihf, e.g. on adler32.c.

nmoinvaz:
@dankegel are there also problems with configure using powerpc? I noticed we don't have a CI test for powerpc for configure..

dankegel:
powerpc seems ok.  arm is more attached to -march flags than powerpc is, I think...

dankegel:
The cmake O2/O3 issue seems beyond the goals of this pull request; my strategy of letting the wookie, er cmake, win seems like enough for the moment.

And I'm not sure this pull request should be blocked waiting for the 32 bit -march disagreement between cmake and configure to be solved...?  A later pull request can expand the list of architectures tested by pkgcheck.

Other than those, I think the pull request now covers the feedback received, and it's ready for merging.

configure
```diff
@@ -1173,11 +1175,22 @@ case "${ARCH}" in
...
-            CFLAGS="${CFLAGS} -DX86_CPUID -DX86_SSE2 -DX86_SSE2_CHUNKSET -DX86_SSE42_CRC_HASH"
-            SFLAGS="${SFLAGS} -DX86_CPUID -DX86_SSE2 -DX86_SSE2_CHUNKSET -DX86_SSE42_CRC_HASH"
+            CFLAGS="${CFLAGS} -DX86_CPUID"
+            SFLAGS="${SFLAGS} -DX86_CPUID"
+            ARCH_STATIC_OBJS="${ARCH_STATIC_OBJS} x86.o"
+            ARCH_SHARED_OBJS="${ARCH_SHARED_OBJS} x86.lo"
 
-            ARCH_STATIC_OBJS="${ARCH_STATIC_OBJS} x86.o chunkset_sse.o insert_string_sse.o slide_sse.o"
-            ARCH_SHARED_OBJS="${ARCH_SHARED_OBJS} x86.lo chunkset_sse.lo insert_string_sse.lo slide_sse.lo"
+            # if(WITH_SSE4 AND (HAVE_SSE42CRC_INLINE_ASM OR HAVE_SSE42CRC_INTRIN))
```
mtl1979, line 1183: Short-circuiting the test should be reverted...

configure
```diff
@@ -1173,11 +1175,22 @@ case "${ARCH}" in
...
+            CFLAGS="${CFLAGS} -DX86_SSE42_CRC_HASH"
```
mtl1979: Assuming SSE4.2 is available is still bad as a lot of AMD processors don't have SSE4.2 as it was originally available on only Intel processors. I don't know when the test was removed as even my previous development machine didn't have SSE4.2 and zlib-ng worked on it using configure...

dankegel: I think there was never a check in configure; I pulled in the check from CMakeLists.txt and
added it as a comment in configure.  I admit that's a bit confusing.  I suspect zlib-ng checks
for sse4.2 at runtime before using it...?

nmoinvaz:
I think we should wait until some of the cmake/configure issues are resolved and then add more configurations to the pkgcheck. Sometimes these things take time. I have had PRs that take 3-4 months so I know how it is.

dankegel:
h'ok.  Landing the other three pull requests first is fine.

dankegel:
Let's do this :-)

mtl1979:
This overlaps with #676 ?

dankegel:
Only trivially.  git will do the right thing.  (I think.)

In retrospect, I suppose I should have waited until #676 was merged before saying "let's do this", to avoid that question.

(Merge could happen in either order, but if you merge this one first, #676 would be a no-op.)

mtl1979:
As long as @Dead2 knows the correct order of merging pull requests...

nmoinvaz:
@dankegel thank you for adding the new architectures. Are there any remaining outstanding issues that we haven't cataloged? I had also noticed the issue with `LDSHAREDLIBC` too during test.  And I also see that the line `LDFLAGS=${LDFLAGS-"-L."}` adds `-L.` which is not added in CMake.

config.sub
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
nmoinvaz: Is it possible to do this inline in `configure` without a separate file?

config.sub
```diff
@@ -0,0 +1,13 @@
...
+# Canonicalize CHOST.  Like
+# https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub
+# but accepts a much smaller set of machine specifications,
+# and is under the zlib license rather than the GPL.
+# See also https://bugs.gentoo.org/651908
```
nmoinvaz: I don't know that we need to mention GPL here.

test/pkgcheck.sh
```diff
@@ -1,9 +1,81 @@
...
+
+if ! test -f "configure"
+then
+  echo "Please run from top of source tree"
+  exit 1
+fi
+
+# Tell GNU's ld etc. to use Jan 1 1970 when embedding timestamps
+# Probably only needed on older systems (ubuntu 14.04?)
+export SOURCE_DATE_EPOCH=0
```
nmoinvaz: Do we need this `SOURCE_DATE_EPOCH`? Maybe not..?

test/pkgcheck.sh
```diff
@@ -40,24 +112,55 @@ cd btmp1
...
+    cd pkgtmp1; ar x usr/local/lib/libz$suffix.a; rm usr/local/lib/libz$suffix.a; cd ..
+    cd pkgtmp2; ar x usr/local/lib/libz$suffix.a; rm usr/local/lib/libz$suffix.a; for a in *.c.o; do mv $a ${a%%.c.o}.o; done; cd ..
+  fi
+}
+
+case $(uname) in
+Darwin)
+  # Alas, dylibs still have an embedded hash or something,
+  # so nuke it.
+  # FIXME: find a less fragile way to deal with this.
```
nmoinvaz: Is there something we can do about this? How many bytes were changed?

nmoinvaz (commented):
I have given some feedback on a few things.

config.sub
dankegel: Yes, but config.sub is traditional, and if somebody needs one of the arches not yet in this config.sub,
they can drop in the gnu one and get unblocked.  So there's utility in keeping it a separate file.

config.sub
dankegel: Yeah, that could confuse license scanners.

test/pkgcheck.sh
dankegel: I believe I needed it on FreeBSD 12.1, Debian 8, and Ubuntu 14.04. 

See also https://reproducible-builds.org/docs/source-date-epoch/ for what this does.

test/pkgcheck.sh
dankegel: Per the dd command, 16 bytes.

One does not do anything about Apple tools, one endures them :-)

dankegel:
-L. appears to do nothing and could be omitted.  I looked through git history and found it may have originally been
added in f6194ef3, but even then, it was a no-op (unless there is some strange ld out there
that searches the library search path even when you provide the filename of the library).

Likewise, the -L.. could be omitted from test/Makefile.in.

But as it does not cause configure and cmake to generate different bits,
I hesitate to remove it as part of this pull request... perhaps it could be a followon.

test/pkgcheck.sh
mtl1979: Would be nice to have some link to site that describes what those bytes actually are... Mach-O file format reference doesn't have clear definition what starts at byte offset 1337 as the file header isn't that long...

dankegel: Yes, that would be nice.  However, I don't know how to dig it up myself, and I don't know that it's worth holding up this pull request for.

dankegel:
OK, I updated config.sub's comment to explain what it does better, and to
mention using autotools' config.sub as a workaround in case this one is too limited.

I also rebased.

I think that (and my replies above) resolves all the feedback received today sufficiently.
If the magic dd for macos seems unacceptable, I can remove macos support from pkgcheck.sh.

Please merge #676 first, then I'll rebase.

test/pkgcheck.sh
nmoinvaz: I'm not sure if this is current but `strip` seems to have a `-no_uuid` flag that will strip out all `LC_UUID` load commands.
https://www.manpagez.com/man/1/strip/

dankegel: That sounds handy.  Also, evidently ld has a -no_uuid option.  I'll see if those can replace the dd!

dankegel: So far no joy.  It was (and maybe is) a good theory, though.

I just refreshed the mac support a bit, turned up an interesting bit so I left it as a separate commit.

dankegel: Wait, there is joy in mudville.  Stand by...

dankegel: I got it to go green, but there's a catch.  See https://github.com/zlib-ng/zlib-ng/issues/681

test/pkgcheck.sh
```diff
@@ -40,24 +118,57 @@ cd btmp1
...
-    if grep -i fedora /etc/os-release > /dev/null
-    then
-        # Note: Fedora patches cmake to use -O2 in release, which
-        # does not match the -O3 configure sets :-(
-        export CFLAGS="-O2 -DNDEBUG"
-    fi
-    ;;
   esac
-  ../configure
+  # Use same optimization level as cmake did.
```
nmoinvaz: We probably won’t need this after #674 is merged.

dankegel:
h'ok, it passes on mac now, too, without the dastardly dd hack.

How do people feel about the way the mac AVX2 problem was
solved by overloading WITH_NATIVE_INSTRUCTIONS ?
I could have introduced a new option, but WITH_NATIVE_INSTRUCTIONS
is such a close match.

dankegel:
Is that code coverage decrease real?

Dead2:
This is now ready for rebase.

mscastanho:
@dankegel I noticed that this leaves directories `btmp1`, `btmp2`, `pkgtmp1` and `pkgtmp2` after running the script. Maybe add them to `make clean`? Another option would be to erase them after the check is done, but it might be useful to be able to check them in case of error. This could easily be part of a future clean up PR as well.

dankegel:
@mscastanho I added clean and .gitignore entries for pkgcheck's detritus, and made pkgcheck clean up after itself except on failure.

dankegel:
Rebased and ready for final review and landing.

dankegel:
The coverage changes seem spurious, but they're caused by something, and this pull request probably has to wait until it's understood.  It's mildly annoying that they only show up in the CI of the pull request, not the CI of the branch before creating the pull request...

dankegel:
Let's land #692 first, then I'll rebase and figure out that coverage problem.

config.sub
nmoinvaz: Is it possible to merge this file into the configure script?

nmoinvaz: Is it possible to merge this file into the configure script?

configure
```diff
@@ -1118,8 +1120,8 @@ ARCH_SHARED_OBJS=''
 
 # Set ARCH specific FLAGS
 case "${ARCH}" in
-    # x86 specific optimizations
-    i386 | i486 | i586 | i686)
+    # x86/amd64 specific optimizations
+    i386 | i486 | i586 | i686 |x86_64)
```
nmoinvaz, line 1124: I assume with these changes now we always do SSE2 check for x64 instead of assuming it exists (like CMake) and that you have merged the x86 and x64 sections in configure?

nmoinvaz (commented):
I have made only two comments. It looks much smaller indeed!

nmoinvaz:
It seems I already asked the question about config.sub. I think it would be better if it was merged into configure in some way. That way configure script is more self contained. 

An alternative might be to create a “configure” directory same as the “CMake” directory and put it there but I’m not that fond of that idea either.

configure
dankegel: Yup.

test/pkgcheck.sh
dankegel: Gone.

dankegel:
I removed the questionable set of CFLAGS in test/pkgcheck.sh, and moved config.sub out of the top level directory.

I think it really is good to go now...?

nmoinvaz (approved):
I think this commit makes the code better - I like how it no longer checks the CMakeCache.txt. Having the file in the _tools_ directory is a nicer too. @Dead2 will be able to speak to if he wants it there or in a _configure_ directory. Otherwise, LGTM

dankegel:
Can't have it in a 'configure' subdirectory, there's a bit of a name clash there :-)

dankegel:
Push the button @Dead2 :-)

Dead2:
I am in no condition to review stuff right now, with 5 holes in my abdomen and recovering from surgery that did not go quite as planned (it turned out OK, but means a much longer and more painful recovery). I hope to be able to read without my eyes constantly crossing in a couple of days, after that I can probably get some more testing and reviewing done.

That said, this PR has been reviewed multiple times by me and others, and although I see only 1 current review approval, I'll merge this now so we can keep things moving along and get the rest rebased on top of this as well. The other PRs will have to wait a bit longer though, unless there are more-or-less trivial changes that get at least TWO review approvals in addition to me.

dankegel:
Thank you and get well soon!

PR #660, opened 2020-06-28
Modify cmake to apply native flag when using intrinsic flags

nmoinvaz:


mtl1979:
A lot of failures with ```_mm_shuffle_epi8()``` which requires SSSE3.

nmoinvaz:
Now `_mm_extract_epi32`...

mtl1979:
@nmoinvaz ```_mm_extract_epi32()``` is SSE4.1

nmoinvaz:
I don't really want to add all new infrastructure for it.. SSE4.2 check is good enough imo.

mtl1979:
My old PC did have SSE4.1, but no SSE4.2 ;)

nmoinvaz:
It looks like I used the wrong flag name.

mtl1979:
As far as I remember, before AMD's Bulldozer SSE4.2 was pretty much Intel thing... AMD processors had SSE4a (and XOP only for Bulldozer). After that there has been no common instruction set between AMD and Intel... (Even AVX implementations have differences)

Dead2:
Hmm, can't we set the native-flag as a global flag used on all files, instead of adding it to all-the-places? (I don't remember).
Might also want to do the native flag before all the instruction-set flags.

Also, x86 instruction sets depend on each other, so if -msse42 is set, that should implicitly set -msse41 (and -mssse3 -msse3 -msse2 -msse). Not sure what went wrong above though..

nmoinvaz:
@Dead2 you are right. If I set `CMAKE_C_FLAGS` then it will apply it to everything.

I have made the requested changes.

Dead2:
Looks awesome.

Dead2:
@mtl1979 AMD has been lagging behind on instruction sets, yes. And their suggested instruction sets were never accepted by Intel, so Intels won out. AMD suggested SSE5 for example, but that was never accepted or published, although its enhancements was later implemented as XOP, FMA and F16C.

However, AVX and AVX2 are very much supported by recent AMD processors.
Remember that FMA is not part of AVX even though some mistakenly say it is, in reality they are related but they exist behind different cpuid flags. Intel wanted FMA4 then changed to FMA3, while AMD wanted FMA3 then changed to FMA4. Interesting reading about that mess of history here: https://en.wikipedia.org/wiki/FMA_instruction_set#History

Btw, the only old AMD-only instruction set still supported in recent cpus is SSE4A, and that is only a very small addition to the standard SSE4 instructions, consisting only of a few instructions.

In short, instruction sets are generally a mess, but Intel and AMD are generally pretty well compatible, the biggest difference is in the implementations, so some instructions are faster or slower, but that also changes from cpu generation to generation. 😄

mtl1979:
@Dead2 Code compiled for AMD without XOP (Zen and later don't have XOP anymore) works on Intel, but it wastes one instruction that is not necessary on Intel processors but is required on AMD... When I find the link to that issue, I can post it, but it's been a while since I did investigate gcc issues.... This pretty much means that gcc optimizes for AMD by default, to be the most compatible, but on Intel the code is suboptimal unless specifically optimized for Intel processors instead.

PR #661, opened 2020-06-28
Fixed bad shift operation warning in deflatePrime.

nmoinvaz:
```
  CID 293475 (#2-4 of 4): Bad bit shift operation (BAD_SHIFT)
  In expression 1UL << put, left shifting by more than 63 bits has undefined behavior.
```

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -586,14 +586,17 @@ int32_t ZEXPORT PREFIX(deflatePrime)(PREFIX3(stream) *strm, int32_t bits, int32_
     if (deflateStateCheck(strm))
         return Z_STREAM_ERROR;
     s = strm->state;
-    if (bits < 0 || bits > BIT_BUF_SIZE ||
+    if (bits < 0 || bits > BIT_BUF_SIZE || bits > (sizeof(value) << 3) ||
         s->sym_buf < s->pending_out + ((BIT_BUF_SIZE + 7) >> 3))
         return Z_BUF_ERROR;
     do {
         put = BIT_BUF_SIZE - s->bi_valid;
         if (put > bits)
             put = bits;
-        s->bi_buf |= (((uint64_t)value & ((UINT64_C(1) << put) - 1)) << s->bi_valid);
+        if (s->bi_valid == 0)
+            s->bi_buf = (uint64_t)value;
+        else
+            s->bi_buf |= (((uint64_t)value & ((UINT64_C(1) << put) - 1)) << s->bi_valid);
         s->bi_valid += put;
         zng_tr_flush_bits(s);
         value >>= put;
--- a/test/example.c
+++ b/test/example.c
@@ -805,50 +805,96 @@ void test_deflate_pending(unsigned char *compr, size_t comprLen) {
 }
 
 /* ===========================================================================
- * Test deflatePrime() with small buffers
+ * Test deflatePrime() wrapping gzip around deflate stream
  */
-void test_deflate_prime(unsigned char *compr, size_t comprLen) {
+void test_deflate_prime(unsigned char *compr, size_t comprLen, unsigned char *uncompr, size_t uncomprLen) {
     PREFIX3(stream) c_stream; /* compression stream */
+    PREFIX3(stream) d_stream; /* decompression stream */
     int err;
-    int bits = 0;
-    int values = 0;
     size_t len = strlen(hello)+1;
+    uint32_t crc = 0;
 
 
     c_stream.zalloc = zalloc;
     c_stream.zfree = zfree;
     c_stream.opaque = (voidpf)0;
 
-    // windowBits can also be -8..-15 for raw deflate,The default value is 15
+    /* Raw deflate windowBits is -15 */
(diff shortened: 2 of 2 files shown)


mtl1979:
If ```s->bi_valid == 0```, then it should just copy and zero-extend the value instead of trying to mask and shift... The original calculation did overflow for obvious reason.

deflate.c
```diff
@@ -590,7 +590,7 @@ int32_t ZEXPORT PREFIX(deflatePrime)(PREFIX3(stream) *strm, int32_t bits, int32_
         s->sym_buf < s->pending_out + ((BIT_BUF_SIZE + 7) >> 3))
         return Z_BUF_ERROR;
     do {
-        put = BIT_BUF_SIZE - s->bi_valid;
+        put = (BIT_BUF_SIZE-1) - s->bi_valid;
```
mtl1979: This essentially reduces the bit buffer size by one bit, which is a bad thing...

deflate.c
```diff
@@ -590,7 +590,7 @@ int32_t ZEXPORT PREFIX(deflatePrime)(PREFIX3(stream) *strm, int32_t bits, int32_
...
         if (put > bits)
             put = bits;
         s->bi_buf |= (((uint64_t)value & ((UINT64_C(1) << put) - 1)) << s->bi_valid);
```
mtl1979:
```
if (s->bi_valid == 0)
    s->bi_buf = (uint64_t)value;
else
    s->bi_buf |= (((uint64_t)value & ((UINT64_C(1) << put) - 1)) << s->bi_valid);
```

nmoinvaz: Ok it is fixed now. Thanks.

deflate.c
nmoinvaz: Fixed with your changes.

nmoinvaz:
Changes made.

deflate.c
```diff
@@ -593,7 +593,10 @@ int32_t ZEXPORT PREFIX(deflatePrime)(PREFIX3(stream) *strm, int32_t bits, int32_
         put = BIT_BUF_SIZE - s->bi_valid;
         if (put > bits)
             put = bits;
-        s->bi_buf |= (((uint64_t)value & ((UINT64_C(1) << put) - 1)) << s->bi_valid);
+        if (s->bi_valid == 0)
+            s->bi_buf = (uint64_t)value;
```
mtl1979, line 597: Ignoring ```bits``` is safe here because below we essentially set ```s->bi_valid```  to value of ```put``` (actually ```0 + put```) and the test ```put > bits``` above is outside the ```if``` block... Any extra bits will be shifted away when adding more bits to ```s->bi_buf```.

nmoinvaz: @mtl1979, was there something you wanted me to do or were you just commenting?

mtl1979: @nmoinvaz I was just wondering if it needs a comment to explain why it ignores ```bits``` when ```s->bi_valid``` is 0.

nmoinvaz: I don't really think so personally.

Dead2:
Looks good to me, but I wish we had a test for this. Is that a possibility here?

nmoinvaz:
I have made some improvements to `deflatePrime` testing. There is not a possibility for `bits` param being 64 because `value` param is 32-bit so the problem this issue solves would never be hit. But it had been fixed to silence coverity.

Dead2:
This currently fails all checks. Unfortunately I don't know why, because github actions says
`We are currently unable to download the log. Please try again later.`

mtl1979:
```/home/runner/work/zlib-ng/zlib-ng/test/example.c:1055:5: error: too few arguments to function ‘test_deflate_prime’```

File: test/example.c [hunk 4]
```diff
@@ -805,50 +806,96 @@ void test_deflate_pending(unsigned char *compr, size_t comprLen) {
 }
 
 /* ===========================================================================
- * Test deflatePrime() with small buffers
+ * Test deflatePrime() wrapping gzip around deflate stream
  */
-void test_deflate_prime(unsigned char *compr, size_t comprLen) {
+void test_deflate_prime(unsigned char *compr, size_t comprLen, unsigned char *uncompr, size_t uncomprLen) {
```
mtl1979, line 811: This is bad... Changes in function signature break build.

Dead2:
This should also fix the newly found coverity issue at line 599 `value >>= put;` (It basically detects this issue in four variants now)

Dead2:
When you fix the current compile error, please merge the (now 2, would-be 3) test-commits into one. 😄

nmoinvaz:
They are not test commits, I have broken them out by the logic changes. However, I have merged the first two into one commit to reduce the number of commits. Hopefully that looks good to you.

test/example.c
nmoinvaz: Fixed thanks.

Dead2:
My mistake then, I only reviewed the whole thing and then found out there were multiple commits, and assumed the last one was a fix for the next-to-last one. Sorry.

Btw, is Github working really crappy for you lately? Click on "commits" or "files changed" and it never finishes loading until you refresh, same when you want to return to "conversation". Has made it a royal pain to review things the past couple of weeks. Might be something on my end, but not sure what would cause that particular problem, kind of like ajax-requests just get lost on the way.

mtl1979:
@Dead2 GitHub status page is showing a lot of orange and red for last few days...

PR #665, opened 2020-06-30
Fixed unsigned integer overflow in adler32_avx and adler32_ssse3 when len is zero.

nmoinvaz:
This is an ASAN error reported by oss-fuzz.
```
adler32_avx.c:108:15: runtime error: unsigned integer overflow: 0 - 1 cannot be represented in type 'size_t' (aka 'unsigned long')
```
What is strange is that oss-fuzz reported "early exit" as the error, but when I ran the test case they provided it reported this.


Dead2:
LGTM

PR #667, opened 2020-07-01
Remove some extra ci configurations.

nmoinvaz:
Some of these "compounded" configurations I think are unnecessary.
And one or two are duplicates.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -15,22 +15,14 @@ jobs:
           Ubuntu GCC No SSE2,
           Ubuntu GCC No SSE4,
           Ubuntu GCC No PCLMULQDQ,
-          Ubuntu GCC No Opt,
-          Ubuntu GCC Compat,
           Ubuntu GCC Compat No Opt,
-          Ubuntu GCC Compat,
           Ubuntu GCC ARM SF,
           Ubuntu GCC ARM SF No ACLE,
-          Ubuntu GCC ARM SF No Opt,
-          Ubuntu GCC ARM SF Compat,
           Ubuntu GCC ARM SF Compat No Opt,
           Ubuntu GCC ARM HF,
           Ubuntu GCC ARM HF No NEON,
-          Ubuntu GCC ARM HF No Opt,
-          Ubuntu GCC ARM HF Compat,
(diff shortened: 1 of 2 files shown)


.github/workflows/cmake.yml
```diff
@@ -18,20 +18,16 @@ jobs:
...
           Ubuntu GCC ARM SF,
           Ubuntu GCC ARM SF No ACLE,
           Ubuntu GCC ARM SF No Opt,
           Ubuntu GCC ARM SF Compat,
-          Ubuntu GCC ARM SF Compat No Opt,
           Ubuntu GCC ARM HF,
           Ubuntu GCC ARM HF No NEON,
           Ubuntu GCC ARM HF No Opt,
           Ubuntu GCC ARM HF Compat,
-          Ubuntu GCC ARM HF Compat No Opt,
```
Dead2, line 31: I feel like we could rather do "Compat No Opt" to replace both "Compat" and "No Opt", removing two instead of one.
The two options do not interact much, if at all, so should be fine to test together.

Dead2: Same with the SF variant and AARCH64 variant.

.github/workflows/configure.yml
```diff
@@ -13,17 +13,14 @@ jobs:
...
-          Ubuntu GCC ARM SF Compat No Opt,
           Ubuntu GCC ARM HF,
           Ubuntu GCC ARM HF No Opt,
+          Ubuntu GCC ARM HF No NEON,
           Ubuntu GCC ARM HF Compat,
-          Ubuntu GCC ARM HF Compat No Opt,
-          Ubuntu GCC ARM HF Compat No Neon,
-          Ubuntu GCC AARCH64 Compat No Opt,
+          Ubuntu GCC AARCH64 No Opt,
           Ubuntu GCC AARCH64 Compat,
```
Dead2: Same in configure as in CMake

.github/workflows/cmake.yml
nmoinvaz: Done.

.github/workflows/configure.yml
nmoinvaz: Done.

PR #668, opened 2020-07-01
Fixed issues with float abi in cmake

nmoinvaz:
* Apply float abi flag to all files in cmake (same as configure).
* Don't override `-mfloat-abi` if set via `-DCMAKE_C_FLAGS` in cmake.
* Only set `-mfloat-abi=softfp` if compiler ends in eabi in cmake. (same as configure).

I'm pretty sure this will fix the issues mentioned in #478. It will allow being able to set `-mfloat-abi` manually via `CMAKE_C_FLAGS`. Also when the compiler does not end in "eabi" or "eabihf" it won't try to set a float type. This is the same behavior as configure.


nmoinvaz:
CMake instances have been slow and freezing but things looked good my fork's run.

PR #671, opened 2020-07-02
Fixed configure applying -march flags globally for ARM. 

nmoinvaz:
This should fix the issue reported in #659 where cmake and configure don't apply `-march` equally.


Dead2:
Looks good to me

PR #674, opened 2020-07-02
Default to optimization level 2 in cmake for release mode.

nmoinvaz:
This will sync up configure and cmake settings in preparation for #659.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -268,6 +268,12 @@ else()
     endif()
 endif()
 
+# Replace optimization level 3 added by default with level 2
+if(NOT MSVC AND NOT CMAKE_C_FLAGS MATCHES "([\\/\\-]O)3")
+    string(REGEX REPLACE "([\\/\\-]O)3" "\\12"
+        CMAKE_C_FLAGS_RELEASE ${CMAKE_C_FLAGS_RELEASE})
+endif()
+
 # Set architecture alignment requirements
 if(BASEARCH_ARM_FOUND OR (BASEARCH_PPC_FOUND AND "${ARCH}" MATCHES "powerpc64le") OR BASEARCH_X86_FOUND)
     if(NOT DEFINED UNALIGNED_OK)


CMakeLists.txt
```diff
@@ -268,6 +268,11 @@ else()
     endif()
 endif()
 
+# Remove optimization level 3 added by default by cmake
+string(REPLACE "-O3" "" CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE}")
+# Set optimization level 2 for release mode
+set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -O2")
```
Dead2: Should this be made conditional perhaps? Only add -O2 if it does not already have it?
Also, can this still be overridden by command-line?

I seem to remember that there is a way to set the defaults (for when `cmake .` is run the first time), would that be more appropriate perhaps? I am not sure, but I suspect this would also change what is in the cache if run a second time.

nmoinvaz: I could do something like this:
```
if(NOT CMAKE_C_FLAGS MATCHES "([\\/\\-]O)[0-9d]")
    # Remove optimization level 3 added by default by cmake
    string(REGEX REPLACE "([\\/\\-]O)3" "\\2"
        CMAKE_C_FLAGS_RELEASE ${CMAKE_C_FLAGS_RELEASE})
else()
    # Optimization level provided by C flags, remove it from release flags
    string(REGEX REPLACE "([\\/\\-]O)[0-9d]" ""
        CMAKE_C_FLAGS_RELEASE ${CMAKE_C_FLAGS_RELEASE})
endif()
```

Or I can just reset the default release mode C flags:

```
if(MSVC)
    set(CMAKE_C_FLAGS_RELEASE "/O2 /DNDEBUG")
else()
    set(CMAKE_C_FLAGS_RELEASE "-O2 -DNDEBUG")
endif()
```
In this instance anybody can set `-DCMAKE_C_FLAGS=/O3` and this is added after the `CMAKE_C_FLAGS_RELEASE`:

image

Which do you prefer?

nmoinvaz: I have made some changes.

nmoinvaz:
I have made some changes:

- To overwrite optimization level, it can be set with `-DCMAKE_C_FLAGS=-O3`. 
- To overwrite all optimization flags the specific configuration flags variable can be set, ie: `-DCMAKE_C_FLAGS_RELEASE=-O1`.

I have not found any other good way.

Dead2:
Shouldn't that check be able to detect -O2, -O1, -O0 -Os, -O and -Ofast as well? Not just -O3?
Maybe replace the `3` with `[a-z0-9]{0,5}` or something?

Intel also supports -Od btw.

nmoinvaz:
If somebody specifies -Od then the -O3 will not exist and the replace will do nothing..

Dead2:
Ah, I forgot the NOT there :)
Makes more sense now.

dankegel:
This seems... strange.  I do not know if it will go over well.

Dead2:
@dankegel -O3 is incompatible with UNALIGNED_OK (at least in rough terms)

An alternative could be to disable vectorization using function attributes on each of the functions that depend on UNALIGNED_OK.
(Or the heavy hammer, disable using gcc command-line parameter to disable it per compilation-unit.)

We probably should document this and warn about it.

PR #676, opened 2020-07-03
configure: fix typo (.o in shared library)

dankegel:
Linking a non-pic .o into a shared library could be bad.
In this case it seems to have been harmless, but it should be fixed anyway.


mtl1979:
Good catch ;)

nmoinvaz (approved):

mtl1979 (approved):

Dead2:
Nice :)

Issue #658, opened 2020-06-28
crc32_combine_{gen,op} originated in zlib-1.2.12?, but are listed as 1.2.9 in zlib.map?


dankegel:
Running test/pkgcheck.sh from 
https://github.com/dankegel/zlib-ng/tree/cmake-intrinics-files-v2-plus-bug638
after editing it to not pass --no-added-syms to abidiff fails with:
```
+ abidiff test/zlib-1.2.11-x86_64-linux-gnu.abi zlib-built.abi
Function symbols changes summary: 0 Removed, 31 Added function symbols not referenced by debug info
Variable symbols changes summary: 0 Removed, 4 Added variable symbols not referenced by debug info

31 Added function symbols not referenced by debug info:

... 
  crc32_combine_gen64
  crc32_combine_gen@@ZLIB_1.2.9
  crc32_combine_op@@ZLIB_1.2.9
```

Strictly speaking, zlib.so (when built in compat mode) probably shouldn't offer any
new symbols, as that might make it hard for users to flip back to classic
zlib.  But if they're not in zlib.h, trouble is unlikely.

Those symbols came from https://github.com/madler/zlib/commit/41d86c73b21 in zlib,
which was merged along with zlib.map changes as https://github.com/zlib-ng/zlib-ng/commit/cc36e352bafd42a2cdd

That means they're alien functions from the future :-) as far as zlib-1.2.11 is concerned...
so they probably shouldn't be exported as ZLIB_1.2.9.  Maybe 1.2.12?

Also, those symbols weren't in zlib.h in 1.2.11, so maybe they should be protected with
```
#if defined(ZLIB_INTERNAL) || !defined(ZLIB_COMPAT)
...
#endif 
```
in zlib.h until zlib actually releases a new version?

I guess it depends on which version of zlib we're trying to be compatible with, and how compatible --zlib-compat really is.

nmoinvaz:
We could do `#if ZLIB_VERNUM > 0x12b0 || !defined(ZLIB_COMPAT)`?

nmoinvaz:
Great work! Were those crc32_combine functions the only ones that were different?

dankegel:
They were the only ones that jumped out at me so far, but I have not yet really looked at the map file and symbol versions.

nmoinvaz:
@dankegel is this part of your PR, or did you want me to make one?

dankegel:
It's not part of my PR, and is fair game.

[...]

PR #682, opened 2020-07-05
Fixed extra symbols added to ABI when zlib-compat specified.

nmoinvaz:
See #658 for background.

Abi diff before:
```
31 Added function symbols not referenced by debug info:

  adler32_avx2
  adler32_c
  adler32_ssse3
  chunkcopy_c
  chunkcopy_safe_c
  chunkcopy_safe_sse2
  chunkcopy_sse2
  chunkmemset_c
  chunkmemset_safe_c
  chunkmemset_safe_sse2
  chunkmemset_sse2
  chunksize_c
  chunksize_sse2
  chunkunroll_c
  chunkunroll_sse2
  compare258_c
  compare258_unaligned_16
  compare258_unaligned_32
  compare258_unaligned_64
  compare258_unaligned_avx2
  compare258_unaligned_sse4
  crc32_combine_gen64
  crc32_combine_gen@@ZLIB_1.2.9
  crc32_combine_op@@ZLIB_1.2.9
  longest_match_c
  longest_match_unaligned_16
  longest_match_unaligned_32
  longest_match_unaligned_64
  longest_match_unaligned_avx2
  longest_match_unaligned_sse4
  zlibng_version

4 Added variable symbols not referenced by debug info:

  zlibng_string
  zng_deflate_copyright
  zng_errmsg
  zng_inflate_copyright
```

After:

```
Functions changes summary: 0 Removed, 0 Changed, 0 Added function
Variables changes summary: 0 Removed, 0 Changed, 0 Added variable
Function symbols changes summary: 0 Removed, 3 Added function symbols not referenced by debug info
Variable symbols changes summary: 0 Removed, 0 Added variable symbol not referenced by debug info
3 Added function symbols not referenced by debug info:

  crc32_combine_gen64@@ZLIB_1.2.12
  crc32_combine_gen@@ZLIB_1.2.12
  crc32_combine_op@@ZLIB_1.2.12
```

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -52,7 +52,7 @@
 #include "deflate_p.h"
 #include "functable.h"
 
-const char zng_deflate_copyright[] = " deflate 1.2.11.f Copyright 1995-2016 Jean-loup Gailly and Mark Adler ";
+const char PREFIX(deflate_copyright)[] = " deflate 1.2.12.f Copyright 1995-2016 Jean-loup Gailly and Mark Adler ";
 /*
   If you use the zlib library in a product, an acknowledgment is welcome
   in the documentation of your product. If for some reason you cannot
--- a/zutil.c
+++ b/zutil.c
@@ -9,7 +9,7 @@
 #  include "gzguts.h"
 #endif
 
-const char * const zng_errmsg[10] = {
+const char * const PREFIX(z_errmsg)[10] = {
     (const char *)"need dictionary",     /* Z_NEED_DICT       2  */
     (const char *)"stream end",          /* Z_STREAM_END      1  */
     (const char *)"",                    /* Z_OK              0  */
@@ -23,7 +23,7 @@ const char * const zng_errmsg[10] = {
 };
 
 const char zlibng_string[] =
-    " zlib-ng 1.9.9 forked from zlib 1.2.11 ";
+    " zlib-ng 1.9.9 forked from zlib 1.2.12.f ";
 
 #ifdef ZLIB_COMPAT
 const char * ZEXPORT zlibVersion(void) {
--- a/zutil.h
+++ b/zutil.h
@@ -33,15 +33,16 @@
 #else
 #  include "zlib-ng.h"
 #endif
+#include "zbuild.h"
 
 typedef unsigned char uch; /* Included for compatibility with external code only */
 typedef uint16_t ush;      /* Included for compatibility with external code only */
 typedef unsigned long ulg;
 
-extern const char * const zng_errmsg[10]; /* indexed by 2-zlib_error */
+extern const char * const PREFIX(z_errmsg)[10]; /* indexed by 2-zlib_error */
 /* (size given to avoid silly warnings with Visual C++) */
 
-#define ERR_MSG(err) zng_errmsg[Z_NEED_DICT-(err)]
+#define ERR_MSG(err) PREFIX(z_errmsg)[Z_NEED_DICT-(err)]
 
 #define ERR_RETURN(strm, err) return (strm->msg = ERR_MSG(err), (err))
 /* To be used only when the state is known to be valid */
@@ -122,7 +123,7 @@ extern const char * const zng_errmsg[10]; /* indexed by 2-zlib_error */
 
 /* MS Visual Studio does not allow inline in C, only C++.
    But it provides __inline instead, so use that. */
-#if defined(_MSC_VER) && !defined(inline) && !defined(__cplusplus) 
+#if defined(_MSC_VER) && !defined(inline) && !defined(__cplusplus)
 #  define inline __inline
 #endif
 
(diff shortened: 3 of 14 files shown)


zutil.h
```diff
@@ -38,10 +38,10 @@ typedef unsigned char uch; /* Included for compatibility with external code only
 typedef uint16_t ush;      /* Included for compatibility with external code only */
 typedef unsigned long ulg;
 
-extern const char * const zng_errmsg[10]; /* indexed by 2-zlib_error */
+extern const char * const PREFIX(z_errmsg)[10]; /* indexed by 2-zlib_error */
 /* (size given to avoid silly warnings with Visual C++) */
 
-#define ERR_MSG(err) zng_errmsg[Z_NEED_DICT-(err)]
+#define ERR_MSG(err) PREFIX(z_errmsg)[Z_NEED_DICT-(err)]
```
nmoinvaz, line 45: Since now using `PREFIX` inside zutil.h, we have to always include zbuild.h before including zutil.h in any source files.

zutil.c
```diff
@@ -22,16 +22,7 @@ const char * const zng_errmsg[10] = {
...
-const char zlibng_string[] =
-    " zlib-ng 1.9.9 forked from zlib 1.2.11 ";
-
-#ifdef ZLIB_COMPAT
-const char * ZEXPORT zlibVersion(void) {
-    return ZLIB_VERSION;
-}
-#endif
-
-const char * ZEXPORT zlibng_version(void) {
```
nmoinvaz: zlibng_version no longer exposed when zlib-compat mode. Is this ok?

zutil.c
```diff
@@ -22,16 +22,7 @@ const char * const zng_errmsg[10] = {
     (const char *)""
 };
 
-const char zlibng_string[] =
```
nmoinvaz, line 25: String not used for anything..

zutil.c
Dead2: I think we should not completely hide that we are in fact zlib-ng and not zlib, so we should make it possible for programs to detect this, at least at compile-time.

We need to return ZLIB_VERSION in compat-mode, in case the external application checks the zlib-version and does things depending on that version, but we should use ZLIBNG_VERSION in native mode.

I am also not sure about this change of function name, does zlib call it zVersion instead of zlibVersion?

zutil.c
Dead2: We should expose this string to external applications somehow, don't you think?

deflate.c
```diff
@@ -52,7 +52,7 @@
 #include "deflate_p.h"
 #include "functable.h"
 
-const char zng_deflate_copyright[] = " deflate 1.2.11.f Copyright 1995-2016 Jean-loup Gailly and Mark Adler ";
+ZLIB_INTERNAL const char zng_deflate_copyright[] = " deflate 1.2.11.f Copyright 1995-2016 Jean-loup Gailly and Mark Adler ";
```
Dead2: Does this still prevent the compiler from optimizing it out? It should remain in the finished binary.
We should probably add our own string to this.

Should probably say 1.2.12.f now.

zutil.h
nmoinvaz: @Dead2 Any thoughts on this change?

Dead2: It looks OK. Though I was also wondering whether it might be just as well to include zbuild.h from zutil.h.

zutil.c
nmoinvaz: `zVersion` is a preprocessor define. It would only expose `zlibVersion` if in `ZLIB_COMPAT` and `zlibng_Version` otherwise.

zutil.h
nmoinvaz: I have included `zbuild.h`.

zutil.c
nmoinvaz: I have added these functions back.

zutil.c
nmoinvaz: I suppose. I'm not sure the usefulness of it or the copyright strings, but it is what it is..

nmoinvaz: I have added back the string.

deflate.c
nmoinvaz: I have updated the version numbers to 1.2.12.f. I removed `ZLIB_INTERNAL` and added `PREFIX` instead.

nmoinvaz:
I have resolved all the comments.

The output from abidiff should now be:
```
+ abidiff test/zlib-1.2.11-x86_64-linux-gnu.abi zlib-built.abi
Functions changes summary: 0 Removed, 0 Changed, 0 Added function
Variables changes summary: 0 Removed, 0 Changed, 0 Added variable
Function symbols changes summary: 0 Removed, 4 Added function symbols not referenced by debug info
Variable symbols changes summary: 0 Removed, 1 Added variable symbol not referenced by debug info

4 Added function symbols not referenced by debug info:

  crc32_combine_gen64@@ZLIB_1.2.12
  crc32_combine_gen@@ZLIB_1.2.12
  crc32_combine_op@@ZLIB_1.2.12
  zlibng_version

1 Added variable symbol not referenced by debug info:

  zlibng_string
```

PR #689, opened 2020-07-09
Remove inflate compilation options from CMake script

nmoinvaz:
Instead I now just pass them through CMAKE_C_FLAGS in CI yaml. No reason to support two extra options if we don’t have to.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -219,7 +219,7 @@ jobs:
           - name: Ubuntu Clang Inflate Strict
             os: ubuntu-latest
             compiler: clang
-            cmake-args: -DWITH_INFLATE_STRICT=ON
+            cflags: -DINFLATE_STRICT
             packages: llvm-6.0
             gcov-exec: llvm-cov-6.0 gcov
             codecov: ubuntu_clang_inflate_strict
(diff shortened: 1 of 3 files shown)


nmoinvaz:
Rebased

gdevenyi:
I kinda feel like this should remain as an option, rather than embedding in the CFLAGS.

Perhaps hide behind "advanced options"?

mtl1979:
I think this should remain as an option too...

nmoinvaz:
I think there will be some options that dont need to be CMake options because they are rarely used or only used for testing, like -DUSE_MMAP.

mtl1979:
Passing things through CMAKE_C_FLAGS is bad practice, as it is cache variable, that should not be used by external code... The correct external variable is CFLAGS.

nmoinvaz:
Ok I have modified CI to use `CFLAGS`. In other PRs I plan to also:
* Add CI instance with `-DUSE_MMAP` using `CFLAGS` #690 
* Remove `ZLIB_DUAL_LINK` cmake option and make it a ctest if zlib package is found.

Dead2:
I feel like these two options might be useful in some very rare corner-cases, such as attempting to recover data from a compressed stream/file that has been corrupted and compliance checking of a compressed stream/file. I admit that those cases are likely very very rare, but it is a possibility, and that is also why I haven't removed those from the code altogether.

Maybe we should put those and any really special-purpose options like those into a separate options table in the readme file, to deter users from thinking those are good to have enabled.

nmoinvaz:
I think the interface/option set looks cleaner without them. We could have a table in the wiki with all the advanced compiler defines like USE_MMAP, etc and remove mention of them in the readme altogether.

nmoinvaz:
Btw these options are already in the "Advanced Options" table in readme. 

I suppose if you really want to keep the options you can close this PR.

PR #691, opened 2020-07-10
Use mainline codecov repository

nmoinvaz:
It is no longer necessary to use my fork since codecov has brought their repository up to speed with the necessary changes including retry.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -333,10 +333,8 @@ jobs:
 
     - name: Install codecov.io tools
       if: matrix.codecov
-      # Need to use special fork of codecov-python for Windows
-      # https://github.com/codecov/codecov-python/pull/169
       run: |
-        python -u -m pip install git+https://github.com/nmoinvaz/codecov-python.git@master
+        python -u -m pip install git+https://github.com/codecov/codecov-python.git
 
     - name: Generate project files
       # Shared libaries turned off for qemu ppc* and sparc & reduce code coverage sources


nmoinvaz:
Looks like it is still not ready for prime time I will have to look at the differences when I have time.

Dead2:
Maybe we should close this for now?
I am thinking about setting up a proper proxy for codecov btw, with an https certificate and everything.
It looks to me like the problem is routing from the CI machines to Codecov (for whatever reason), nothing else really makes any sense to me, but it is very hard to debug such problems in such an environment. There is a slight chance the routing might work to my server when it is flaky/broken directly to Codecov, and it might fix the problem for us if we are lucky.

nmoinvaz:
Even my fork has been having problems lately. I think in the past week or so there have been more failures to upload coverage reports in general. 

Proxy might be a possible solution.

Another one is to have all the matrixes use gcovr to generate xml coverage files and save as artifacts. Then a final step would download all artifacts and upload them at once to codecov.

nmoinvaz:
What makes the problem even more mysterious is that I believe Codecov is using AWS which should have a decent connection.

nmoinvaz:
It is likely that the problems with codecov are nothing to do with my branch versus mainline, because we have been on mine this whole time and still having problems. I suggest we move to mainline codecov; I don't want to maintain a separate fork.

nmoinvaz:
At least this PR is now all green!

Dead2:
Hmm, I guess we could try it out.

I also have https://codecov.circlestorm.org set up now as a forward proxy, so we could also try setting CODECOV_URL to that in the fallback attempt in tools/codecov-upload.sh

Dead2:
Still fails, and it is unfortunately harder to see when it fails due to codecov since it now fails after 5min instead of 4 hrs, so you now need to investigate each CI failure to see what failed. =/ 
Wish there was a "yellow fail"/soft fail and not just the red complete fail.

nmoinvaz:
There is a `--tries` argument that we don't use in the codecov upload script. My fork defaulted to 8. Mainline defaults to 3. But also mainline does not wait any amount of time between requests, where as my forked waited some time between them.

PR #692, opened 2020-07-12
Make static libraries produced by cmake and configure identical.

### LINKED ISSUE #681: configure and cmake disagree on CPU feature checks
When doing native builds, cmake actually checks whether CPU features
are present, whereas configure doesn't.

This leads to the cmake and configure builds producing different
output on the Mac CI build, which uses a Sandy Bridge processor
which supports AVX but not AVX2.

https://github.com/dankegel/zlib-ng/tree/bug638-identical-builds-tmpsssh
disables that particular runtime check and achieves identical
output from cmake and configure builds... on that one cpu.

Does CMakeLists.txt need an option for "restrict to features present on current running CPU" that
defaults to false, to match configure's behavior?

dankegel:
This is the first half of https://github.com/zlib-ng/zlib-ng/pull/659; 
a later pull request will fix shared libraries.

This fixes #681, exposed by zlib-ng's current mac builders, which lack AVX2.

Also:
- test/pkgcheck.sh: also compare .a, check compat mode, and handle mac better
- CMakeLists.txt: get Version right in compat mode .pc file
- .github/workflows/pkgcheck.sh: test on more systems, including mac

### DIFF
--- a/.github/workflows/pkgcheck.yml
+++ b/.github/workflows/pkgcheck.yml
@@ -8,28 +8,90 @@ jobs:
       fail-fast: false
       matrix:
         name: [
-          Ubuntu GCC
+          Ubuntu GCC,
+          Ubuntu GCC -m32,
+          Ubuntu GCC ARM HF,
+          Ubuntu GCC AARCH64,
+          Ubuntu GCC PPC,
+          Ubuntu GCC PPC64LE,
+          macOS Clang
         ]
         include:
           - name: Ubuntu GCC
             os: ubuntu-latest
             compiler: gcc
-            packages: ninja-build
+
+          - name: Ubuntu GCC -m32
+            os: ubuntu-latest
+            compiler: gcc
+            packages: gcc-multilib
+            cmake-args: -DCMAKE_C_FLAGS=-m32
+            cflags: -m32
+            ldflags: -m32
+
+          - name: Ubuntu GCC ARM HF
+            os: ubuntu-latest
+            chost: arm-linux-gnueabihf
+            compiler: arm-linux-gnueabihf-gcc
+            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-arm.cmake -DCMAKE_C_COMPILER_TARGET=arm-linux-gnueabihf
+            packages: qemu gcc-arm-linux-gnueabihf libc6-dev-armhf-cross
+
+          - name: Ubuntu GCC AARCH64
+            os: ubuntu-latest
+            chost: aarch64-linux-gnu
+            compiler: aarch64-linux-gnu-gcc
+            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-aarch64.cmake
+            packages: qemu gcc-aarch64-linux-gnu libc6-dev-arm64-cross
+
+          - name: Ubuntu GCC PPC
+            os: ubuntu-latest
+            chost: powerpc-linux-gnu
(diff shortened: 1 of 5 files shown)


Dead2:
Looks good to me

.github/workflows/pkgcheck.yml
```diff
@@ -24,12 +70,29 @@ jobs:
       if: runner.os == 'Linux' && matrix.packages
       run: |
         sudo apt-get update
-        sudo apt-get install -y ${{ matrix.packages }}
+        sudo apt-get install -y --no-install-recommends ${{ matrix.packages }}
```
nmoinvaz: Consider adding `ninja-build` here instead of everytime above in the matrix. Also under macOS we install ninja as part of the install packages step below by default.

CMakeLists.txt
```diff
@@ -458,8 +458,15 @@ endif()
 
 # Macro to check if source compiles when cross-compiling
 # or runs when compiling natively
+#
+# Note: because we want to match the behavior
```
nmoinvaz: I'm not sure if this extra CMake comment is really necessary. It seems like this change actually fixes a bug in non-native compiles.

nmoinvaz (commented):
I have only made two feedback comments. Thanks for splitting this into a new PR, it makes it easier to review and probably for @Dead2 to approve. This is looking good!

nmoinvaz:
I sort of see what you mean about the `check_c_source_compile_or_run ` check now. I wonder if it is even necessary to have this macro and perhaps we should only be calling `check_c_source_compiles`. It doesn't seem really necessary to check that it runs, because everything is checked in functable if it is supported.

CMakeLists.txt
nmoinvaz: In your comment you mention that configure doesn't make runtime checks like this. But I looked at the try() function in the configure check and it appears to check to run it see that the exit code is zero. Did you mean something else? Sorry I am trying to understand.

dankegel: aye, try() verifies that the given command runs... but the given command is a compile command; it does not also run the output of the compile command.

nmoinvaz: If that is the case then maybe we should only be using `check_c_source_compiles` in CMake. What do you think?

dankegel: I think it's reasonable to let the native flag drive slightly divergent bits for now, since it's already a "here's a flag to shoot yourself in the foot" kind of thing, and there's no divergence in behavior.

dankegel: (Users who turn on the native flag want the very tightest, and who are we to deny them?
But if this lands, it would be reasonable to file a bug against configure, saying it's missing
a chance to omit dead code in the --native case.)

nmoinvaz (approved):

dankegel:
At least judging by the inline "Codecov Report" above, the reported coverage reduction appears to be entirely on macos, which is interesting... could it be because the AVX2 code is now compiled in but not run on our current Mac builders?
If that's the explanation, the coverage failure is expected and benign.

Not sure what to make of the more detailed coverage change report for "match_tpl.h", though.

mscastanho (approved):
@dankegel Thanks for addressing the directory cleanup. This looks good to me.

Dead2:
Two review approvals? Wow, you must really want this in. 😉  (I wish I could see two approvals on every PR)
Thanks for the help with reviewing this, my reviewing capacity has been low this past week and will probably vary a bit in the coming weeks.

And thanks for this cleanup, things are looking better week by week.

PR #693, opened 2020-07-12
Remove ARM crc instruction set from NEONFLAG. 

nmoinvaz:
The crc instruction set is not used for neon source files. It may have been necessary back when the NEONFLAG was applied to all source files in CMake. Configure script does not apply crc instruction set when setting neon flags.


nmoinvaz:
Code coverage doesn't seem to be related to this change..

PR #695, opened 2020-07-13
Allow for CMake to run tests in parallel.

nmoinvaz:
- Seperate minigzip and switchlevels set of tests into test-compress.cmake.
- Create run-and-compare.cmake for tests that need to run a command and compare output.

This should resolve #646.

PR #698, opened 2020-07-16
Improvements to ctests

nmoinvaz:
* Change ctest to test all the files in test/data
   * We no longer need to specify each file by name.
   * Anybody who wants to do tests can just drop their files in there.
* Improved minigzip test configurations in ctest.
  * For example, we don't need to be testing level -9 with -R (RLE) because those are two different algorithm configurations
* Added corpora repository to test/data to test for each run in CI CMake
  * This tests all the corpora in https://github.com/nmoinvaz/corpora
  * Tests artificial, calgary, canterbury, large, miscellaneous, silesia, snappy corpora 
  * Downside is that each CI instance takes about 2 minutes longer because it is running about 650 tests.

PR #695 should be merged before this one.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -370,7 +370,7 @@ jobs:
       if: runner.os != 'Windows' || contains(matrix.name, 'ARM') == false
       run: |
         cd ${{ matrix.build-dir || '.' }}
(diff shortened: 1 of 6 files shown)


Dead2:
Looks awesome, have not tested but CI looks good.

Issue #700, opened 2020-07-17
compat mode not yet ABI compatible with classic zlib on 64 bit machines


dankegel:
Noticed while fixing #699.

https://github.com/dankegel/zlib-ng/tree/abicheck-tmpsssh is a draft of a new script, tmp/abicheck.sh,
that compares zlib and zlib-ng --compat ABIs (as well as makes sure zlib-ng non-compat ABI remains stable).

It complains that quite a few functions have arguments of the wrong size, e.g.
```
  [C]'function uLong adler32(uLong, const Bytef*, uInt)' at adler32.c:88:1 has some indirect sub-type changes:
    return type changed:
    parameter 1 of type 'typedef uLong' changed:
      underlying type 'unsigned long int' at types.h:41:1 changed:
        entity changed from 'unsigned long int' to compatible type 'typedef __uint32_t' at types.h:41:1
          type name changed from 'unsigned long int' to 'unsigned int'
          type size changed from 64 to 32 bits
```

Evidently ZLIB_COMPAT needs to be sprinkled around the codebase a little more, a la
https://github.com/zlib-ng/zlib-ng/blob/develop/crc32.c#L64

mtl1979:
I can make a pull request to fix this...

dankegel:
While looking into this, I noticed 32-bit-only problems, and filed https://github.com/zlib-ng/zlib-ng/issues/705 for them.

(In short, the recent commit
"For gzseek, gzoffset, gzopen, adler32_combine, crc32_combine and crc32_combine_gen, export 32-bit and 64-bit versions for zlib-compatible api and only 64-bit version (without 64 suffix) for zlib-ng native api",
https://github.com/zlib-ng/zlib-ng/commit/6dcc7bf815c803fe7712e449b3c8fee8f3ab6dda#diff-ce3e5ac484a4ca73e6533421950dbe15
seems to introduce z_off64_t problems when building --zlib-compat on -m32 and/or real 32 bit machines.)

mtl1979:
@nmoinvaz might need to make pull request with partial revert to fix that issue... We do still support 32-bit processors.

dankegel:
Y'know, I should probably file a separate bug for the 32 bit problems... done.

mtl1979:
@dankegel Indeed...

nmoinvaz:
@dankegel anything more to do on this or can we close it out? Seems like all the PRs have been merged..

dankegel:
[...]

PR #701, opened 2020-07-17
Fix signature of adler32(), adler32_combine(), adler32_combine64() and adler32_z() in compat mode.

mtl1979:
* See #700, compat API needs to use "unsigned long" instead of "uint32_t"

### DIFF
--- a/adler32.c
+++ b/adler32.c
@@ -82,7 +82,7 @@ uint32_t adler32_c(uint32_t adler, const unsigned char *buf, size_t len) {
 
 #ifdef ZLIB_COMPAT
 unsigned long ZEXPORT PREFIX(adler32_z)(unsigned long adler, const unsigned char *buf, size_t len) {
-    return (unsigned long) functable.adler32((uint32_t) adler, buf, len);
+    return (unsigned long)functable.adler32((uint32_t)adler, buf, len);
 }
 #else
 uint32_t ZEXPORT PREFIX(adler32_z)(uint32_t adler, const unsigned char *buf, size_t len) {
@@ -93,7 +93,7 @@ uint32_t ZEXPORT PREFIX(adler32_z)(uint32_t adler, const unsigned char *buf, siz
 /* ========================================================================= */
 #ifdef ZLIB_COMPAT
 unsigned long ZEXPORT PREFIX(adler32)(unsigned long adler, const unsigned char *buf, unsigned int len) {
-    return (unsigned long) functable.adler32((uint32_t) adler, buf, len);
+    return (unsigned long)functable.adler32((uint32_t)adler, buf, len);
 }
 #else
 uint32_t ZEXPORT PREFIX(adler32)(uint32_t adler, const unsigned char *buf, uint32_t len) {
@@ -129,11 +129,11 @@ static uint32_t adler32_combine_(uint32_t adler1, uint32_t adler2, z_off64_t len
 /* ========================================================================= */
 #ifdef ZLIB_COMPAT
 unsigned long ZEXPORT PREFIX(adler32_combine)(unsigned long adler1, unsigned long adler2, z_off_t len2) {
-    return (unsigned long) adler32_combine_((uint32_t) adler1, (uint32_t) adler2, len2);
+    return (unsigned long)adler32_combine_((uint32_t)adler1, (uint32_t)adler2, len2);
 }
 
 unsigned long ZEXPORT PREFIX4(adler32_combine)(unsigned long adler1, unsigned long adler2, z_off64_t len2) {
-    return (unsigned long) adler32_combine_((uint32_t) adler1, (uint32_t) adler2, len2);
+    return (unsigned long)adler32_combine_((uint32_t)adler1, (uint32_t)adler2, len2);
 }
 #else
 uint32_t ZEXPORT PREFIX4(adler32_combine)(uint32_t adler1, uint32_t adler2, z_off64_t len2) {
--- a/adler32.c
+++ b/adler32.c
@@ -80,14 +80,26 @@ uint32_t adler32_c(uint32_t adler, const unsigned char *buf, size_t len) {
     return adler | (sum2 << 16);
 }
 
+#ifdef ZLIB_COMPAT
+unsigned long ZEXPORT PREFIX(adler32_z)(unsigned long adler, const unsigned char *buf, size_t len) {
+    return (unsigned long) functable.adler32((uint32_t) adler, buf, len);
+}
+#else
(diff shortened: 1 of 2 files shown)


zlib.h
```diff
@@ -1645,7 +1645,7 @@ ZEXTERN void ZEXPORT gzclearerr(gzFile file);
    library.
 */
 
-ZEXTERN uint32_t ZEXPORT adler32(uint32_t adler, const unsigned char *buf, uint32_t len);
+ZEXTERN unsigned long ZEXPORT adler32(unsigned long adler, const unsigned char *buf, unsigned int len);
```
dankegel, line 1648: This seems a bit mysterious.  Won't it be wrong now on 64 bits in the non-compat case?

dankegel (commented):
There are a number of other functions that need similar fixes, see output of the script in the bug report.

zlib.h
dankegel: Seems like we need a type that is ZNG_uLong which expands to 32 bits for ng, and unsized for compat?

mtl1979: Non-compat case will not use this header, it uses zlib-ng.h instead.

mtl1979:
@dankegel I did split the changes as several pull requests, one for each source file. See #702 for changes related to ```crc32.c```.

zlib.h
mtl1979: Using ZNG_uLong isn't enough as some compilers will issue warning or error if function parameters or return type is truncated. Explicit zero-extending is for readability mostly as it makes it clear for human reader and obvious for compiler when inlining -- and static analysis -- that the zero-extending is intended behavior and not oversight.

zlib.h
```diff
@@ -1664,7 +1664,7 @@ ZEXTERN uint32_t ZEXPORT adler32(uint32_t adler, const unsigned char *buf, uint3
      if (adler != original_adler) error();
 */
 
-ZEXTERN uint32_t ZEXPORT adler32_z (uint32_t adler, const unsigned char *buf, size_t len);
+ZEXTERN unsigned long ZEXPORT adler32_z (unsigned long adler, const unsigned char *buf, size_t len);
```
nmoinvaz: Prefer no space after function name..

adler32.c
```diff
@@ -116,8 +128,8 @@ static uint32_t adler32_combine_(uint32_t adler1, uint32_t adler2, z_off64_t len
 
 /* ========================================================================= */
 #ifdef ZLIB_COMPAT
-uint32_t ZEXPORT PREFIX(adler32_combine)(uint32_t adler1, uint32_t adler2, z_off_t len2) {
-    return adler32_combine_(adler1, adler2, len2);
+unsigned long ZEXPORT PREFIX(adler32_combine)(unsigned long adler1, unsigned long adler2, z_off_t len2) {
+    return (unsigned long) adler32_combine_((uint32_t) adler1, (uint32_t) adler2, len2);
```
nmoinvaz: Prefer no spaces after casts..

adler32.c
```diff
@@ -80,14 +80,26 @@ uint32_t adler32_c(uint32_t adler, const unsigned char *buf, size_t len) {
     return adler | (sum2 << 16);
 }
 
+#ifdef ZLIB_COMPAT
+unsigned long ZEXPORT PREFIX(adler32_z)(unsigned long adler, const unsigned char *buf, size_t len) {
+    return (unsigned long) functable.adler32((uint32_t) adler, buf, len);
```
nmoinvaz: No space after casts?

nmoinvaz (commented):
Looks pretty good except for some formatting.

zlib.h
mtl1979: I didn't want to mix style changes with code changes as it is frowned upon...

mtl1979:
Mixing formatting and code changes is frowned upon and that's why I didn't change them at the same time... I tried to be consistent with the style in related crc32.c changes partially already merged in...

dankegel:
That makes abidiff happier, but it still objects to adler32_combine64:

```
  [C]'function uLong adler32_combine64(uLong, uLong, off64_t)' at adler32.c:136:1 has some indirect sub-type changes:
    parameter 1 of type 'typedef uLong' changed:
      typedef name changed from uLong to uint32_t at stdint-uintn.h:26:1
      underlying type 'unsigned long int' changed:
        entity changed from 'unsigned long int' to compatible type 'typedef __uint32_t' at types.h:42:1
          type name changed from 'unsigned long int' to 'unsigned int'
          type size changed from 64 to 32 (in bits)
    parameter 2 of type 'typedef uLong' changed:
      typedef name changed from uLong to uint32_t at stdint-uintn.h:26:1
      underlying type 'unsigned long int' changed:
        entity changed from 'unsigned long int' to compatible type 'typedef __uint32_t' at types.h:42:1
          type name changed from 'unsigned long int' to 'unsigned int'
          type size changed from 64 to 32 (in bits)
```

mtl1979:
@dankegel Updated...

dankegel:
Still present.  Check the last three lines of adler32.c...

mtl1979:
@dankegel Updated once again ;)

nmoinvaz (approved):

dankegel (changes requested):
I think this code path in zlib.h, which only shows up when you try to compile a program that uses zlib-ng's compat mode from source (and so doesn't show up in the output of abidiff), has the problem, too?
```
--- a/zlib.h
+++ b/zlib.h
@@ -1795,7 +1795,7 @@ ZEXTERN int ZEXPORT gzgetc_(gzFile file);  /* backward compatibility */
      ZEXTERN z_off_t ZEXPORT gzseek64(gzFile, z_off_t, int);
      ZEXTERN z_off_t ZEXPORT gztell64(gzFile);
      ZEXTERN z_off_t ZEXPORT gzoffset64(gzFile);
-     ZEXTERN uint32_t ZEXPORT adler32_combine64(uint32_t, uint32_t, z_off_t);
+     ZEXTERN unsigned long ZEXPORT adler32_combine64(unsigned long, unsigned long, z_off_t);
      ZEXTERN unsigned long ZEXPORT crc32_combine64(unsigned long, unsigned long, z_off_t);
      ZEXTERN void ZEXPORT crc32_combine_gen64(uint32_t *op, z_off64_t);
 #  endif
```
(note also the inconsistent use of z_off64_t there; more on that in #700 ).

Other than that, I don't see any remaining problems.

mtl1979:
@dankegel Updated ;)

dankegel:
LGTM

nmoinvaz (approved):

PR #704, opened 2020-07-17
Reintroduce support for ZLIB_CONST in compat mode.

### LINKED ISSUE #703: Wrong default for constness in compat mode
* ```next_in``` and ```msg``` inside ```zstream``` should not be const by default in compat mode

mtl1979:
* In compat mode, the default should be next_in and msg without const qualified in zstream.

Fixes #703.

### DIFF
--- a/compress.c
+++ b/compress.c
@@ -42,7 +42,7 @@ int ZEXPORT PREFIX(compress2)(unsigned char *dest, z_size_t *destLen, const unsi
 
     stream.next_out = dest;
     stream.avail_out = 0;
-    stream.next_in = (const unsigned char *)source;
+    stream.next_in = C_CONST_UCP(source);
     stream.avail_in = 0;
 
     do {
--- a/compress.c
+++ b/compress.c
@@ -42,7 +42,7 @@ int ZEXPORT PREFIX(compress2)(unsigned char *dest, z_size_t *destLen, const unsi
 
     stream.next_out = dest;
     stream.avail_out = 0;
-    stream.next_in = C_CONST_UCP(source);
+    stream.next_in = (z_const unsigned char *)source;
     stream.avail_in = 0;
 
     do {
--- a/deflate.c
+++ b/deflate.c
@@ -468,7 +468,7 @@ int32_t ZEXPORT PREFIX(deflateSetDictionary)(PREFIX3(stream) *strm, const uint8_
     avail = strm->avail_in;
     next = strm->next_in;
     strm->avail_in = dictLength;
-    strm->next_in = C_CONST_UCP(dictionary);
+    strm->next_in = (z_const unsigned char *)dictionary;
     fill_window(s);
     while (s->lookahead >= MIN_MATCH) {
         str = s->strstart;
@@ -484,7 +484,7 @@ int32_t ZEXPORT PREFIX(deflateSetDictionary)(PREFIX3(stream) *strm, const uint8_
     s->lookahead = 0;
     s->prev_length = MIN_MATCH-1;
     s->match_available = 0;
-    strm->next_in = C_CONST_UCP(next);
+    strm->next_in = (z_const unsigned char *)next;
     strm->avail_in = avail;
     s->wrap = wrap;
     return Z_OK;
--- a/deflate.c
+++ b/deflate.c
@@ -468,7 +468,7 @@ int32_t ZEXPORT PREFIX(deflateSetDictionary)(PREFIX3(stream) *strm, const uint8_
     avail = strm->avail_in;
     next = strm->next_in;
     strm->avail_in = dictLength;
-    strm->next_in = (const unsigned char *)dictionary;
+    strm->next_in = C_CONST_UCP(dictionary);
     fill_window(s);
     while (s->lookahead >= MIN_MATCH) {
         str = s->strstart;
@@ -484,7 +484,7 @@ int32_t ZEXPORT PREFIX(deflateSetDictionary)(PREFIX3(stream) *strm, const uint8_
     s->lookahead = 0;
     s->prev_length = MIN_MATCH-1;
     s->match_available = 0;
-    strm->next_in = next;
+    strm->next_in = C_CONST_UCP(next);
     strm->avail_in = avail;
     s->wrap = wrap;
     return Z_OK;
--- a/gzwrite.c
+++ b/gzwrite.c
@@ -212,7 +212,7 @@ static size_t gz_write(gz_state *state, void const *buf, size_t len) {
             return 0;
 
         /* directly compress user buffer to file */
-        state->strm.next_in = (const unsigned char *)buf;
+        state->strm.next_in = C_CONST_UCP(buf);
         do {
             unsigned n = (unsigned)-1;
             if (n > len)
--- a/gzwrite.c
+++ b/gzwrite.c
@@ -212,7 +212,7 @@ static size_t gz_write(gz_state *state, void const *buf, size_t len) {
             return 0;
 
         /* directly compress user buffer to file */
-        state->strm.next_in = C_CONST_UCP(buf);
+        state->strm.next_in = (z_const unsigned char *) buf;
         do {
             unsigned n = (unsigned)-1;
             if (n > len)
--- a/infback.c
+++ b/infback.c
@@ -129,17 +129,17 @@ int32_t ZEXPORT PREFIX(inflateBackInit_)(PREFIX3(stream) *strm, int32_t windowBi
  */
 int32_t ZEXPORT PREFIX(inflateBack)(PREFIX3(stream) *strm, in_func in, void *in_desc, out_func out, void *out_desc) {
     struct inflate_state *state;
-    T_CONST_UCP(next);          /* next input */
-    unsigned char *put;         /* next output */
-    unsigned have, left;        /* available input and output */
-    uint32_t hold;              /* bit buffer */
-    unsigned bits;              /* bits in bit buffer */
-    unsigned copy;              /* number of stored or match bytes to copy */
-    unsigned char *from;        /* where to copy match bytes from */
-    code here;                  /* current decoding table entry */
-    code last;                  /* parent table entry */
-    unsigned len;               /* length to copy for repeats, bits to drop */
-    int32_t ret;                /* return code */
+    z_const unsigned char *next; /* next input */
+    unsigned char *put;          /* next output */
+    unsigned have, left;         /* available input and output */
+    uint32_t hold;               /* bit buffer */
+    unsigned bits;               /* bits in bit buffer */
+    unsigned copy;               /* number of stored or match bytes to copy */
+    unsigned char *from;         /* where to copy match bytes from */
+    code here;                   /* current decoding table entry */
+    code last;                   /* parent table entry */
+    unsigned len;                /* length to copy for repeats, bits to drop */
+    int32_t ret;                 /* return code */
     static const uint16_t order[19] = /* permutation of code lengths */
         {16, 17, 18, 0, 8, 7, 9, 6, 10, 5, 11, 4, 12, 3, 13, 2, 14, 1, 15};
 
--- a/infback.c
+++ b/infback.c
@@ -129,7 +129,7 @@ int32_t ZEXPORT PREFIX(inflateBackInit_)(PREFIX3(stream) *strm, int32_t windowBi
  */
 int32_t ZEXPORT PREFIX(inflateBack)(PREFIX3(stream) *strm, in_func in, void *in_desc, out_func out, void *out_desc) {
     struct inflate_state *state;
-    const unsigned char *next;  /* next input */
+    T_CONST_UCP(next);          /* next input */
     unsigned char *put;         /* next output */
     unsigned have, left;        /* available input and output */
     uint32_t hold;              /* bit buffer */
--- a/inffast.c
+++ b/inffast.c
@@ -65,7 +65,7 @@ static inline uint64_t load_64_bits(const unsigned char *in, unsigned bits) {
 void ZLIB_INTERNAL zng_inflate_fast(PREFIX3(stream) *strm, unsigned long start) {
     /* start: inflate()'s starting value for strm->avail_out */
     struct inflate_state *state;
-    T_CONST_UCP(in);            /* local strm->next_in */
+    z_const unsigned char *in;  /* local strm->next_in */
     const unsigned char *last;  /* have enough input while in < last */
     unsigned char *out;         /* local strm->next_out */
     unsigned char *beg;         /* inflate()'s initial strm->next_out */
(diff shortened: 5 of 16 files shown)


nmoinvaz:
Isn't `ZLIB_CONST` set to `const` ? So maybe these should be like `ZLIB_CONST unsigned int x = 0`?

mtl1979:
@nmoinvaz ```z_const``` is set to ```const``` when ```ZLIB_CONST``` is defined to any value... Value of ```ZLIB_CONST``` is ignored.

nmoinvaz:
Maybe we should be using `z_const` instead of `#ifdef ZLIB_CONST`.. what do you think?

mtl1979:
```ZLIB_CONST``` and ```z_const``` were deprecated in our own API, so ```z_const``` is not defined anymore ever, but instead is implicitly assumed to equal to ```const```, not empty string, so using ```z_const``` wouldn't work correctly in non-compat mode.

**Comment by nmoinvaz** [comment]
Maybe we need a #define in zbuild.h for internal const. Maybe like a PREFIX() but form CONST().. The only thing I don't like is all the #if else scattered throughout.

**Comment by mtl1979** [comment]
@nmoinvaz using something akin PREFIX() would get ugly as sometimes the cast needs to be ```unsigned char *``` and sometimes ```char *```... I could perhaps add two macros to hide the casts...

**Comment by nmoinvaz** [comment]
#define CONST(x) const x
Else 
#define CONST(x) x
..

**Comment by mtl1979** [comment]
Non-compat version would need to no-op the cast completely...

**Comment by mtl1979** [comment]
@nmoinvaz Now it's somewhat cleaner... I used three macros to avoid cases when macros would be used inside parentheses as that  is not supported on all compilers...

**Comment by nmoinvaz** [comment]
That does look better. I was thinking more like this that can be used for any type..
```
#ifdef ..
#define CONST(x) const x
#else
#define CONST(x) x
#endif
...
stream.next_in = (CONST(unsigned char*))source;
...
CONST(unsigned char) in;                /* local strm->next_in */
```

At least that way the type is visible in the variable declaration.. thoughts? Or are you saying that would not work?

**Comment by mtl1979** [comment]
@nmoinvaz Like I said above, preprocessor macros don't work inside (). This is because some preprocessors pass through text inside () verbatim to compiler instead of parsing it... Developer can use the feature to disambiguate between preprocessor macro and function name.

**Comment by nmoinvaz** [comment]
Ok I understand now that is the reason for the extra macros. But in my example x is the variable type not the variable name. I think it should still be possible to change it to pass in the variable type instead of name.. what do you think is that look any better?

**Comment by mtl1979** [comment]
You can't pass in variable type as it would make using the macro in casts impossible... It would essentially only work in declaration of variables or struct members where it is always surrounded by just white-space.

**Comment by Dead2** [comment]
Looks pretty good, but as far as I can tell, this removes const from zlib-ng native as well.
I think each `#ifdef` should also check whether compat is enabled, as we do not want to support non-const in the native mode, only for compat mode. This also avoids problems when someone compiles the source without using our cmake/makefiles.

I am not sure I like the names of the macros though, they do not seem to tell me what they do and what their difference is, and I actually mistook one for the other during my review of the code and misremembered another, so this can easily be confused.

I'd like to see more descriptive names used.
These are from the PR currently:
```
#define CONSTU(x) (const unsigned char *) x
#define CONSTUP(x) const unsigned char * x
#define CONST(x)  (const char *) x
```

The below suggestion is meant to provoke thought and discussion, not names I really want, but these contain more information in the name, making it easier to remember what they actually mean and do.
```
#define C_CONST_UCP(x) (const unsigned char *) x
#define T_CONST_UCP(x)  const unsigned char * x
#define C_CONST_CP(x)  (const char *) x
// C_ / T_ = Cast / Type
// U = Unsigned
// CP = Char Pointer
```

**Comment by mtl1979** [comment]
@Dead2 This shouldn't remove "const" from native builds, it should do the opposite... it forces the types to have "const" qualifier...

**Comment by Dead2** [comment]
@mtl1979 It looks to me like it depends on ZLIB_CONST being set, and it never gets set anywhere in our current code/buildsystem and it does not get set in this PR either as far as I can see.. Am I overlooking something?

**Comment by mtl1979** [comment]
It only depends on ZLIB_CONST in compatibility mode due to order of the preprocessor tests...

**Comment by Dead2** [comment]
You are right, I overlooked the above compat test, and the fact that we have a separate zconf-ng.h.in file. Good 👍

**Comment by mtl1979** [comment]
I'm still open for suggestions for the final names of the macros.... It was just a quick hack to test how much code needed to be adapted...

**Comment by joshtriplett** [comment]
I ran into this as well, when trying to use zlib-ng in zlib-compat mode. I'd like to see this change. Thanks for the fix!

**Comment by joshtriplett** [comment]
Also, on what platforms can you not write `(z_const unsigned char *)` as a cast? I've never seen a preprocessor that wouldn't do a substitution there.

**Comment by mtl1979** [comment]
@joshtriplett I mainly use gcc on Ubuntu Linux and Visual C++ on Windows...

Only place where "(" works correctly is right-hand side of macro definition. This is because "(" acts in preprocessor input like double quotes on compiler, it starts string literal in preprocessor that ends with ")" and is parsed as single token. Only difference is that you can put "(" inside the string literal to start another string literal and the result can be passed to another preprocessor macro as input.

**Comment by joshtriplett** [comment]
@mtl1979 I'm not talking about a macro that takes arguments, just a simple `#define` with no arguments:

```
/tmp$ cat test.h 
#define z_const const

x = (z_const unsigned char *)y;
void func(z_const unsigned char *arg);
/tmp$ gcc -E test.h 
# 1 "test.h"
# 1 "<built-in>"
# 1 "<command-line>"
# 31 "<command-line>"
# 1 "/usr/include/stdc-predef.h" 1 3 4
# 32 "<command-line>" 2
# 1 "test.h"


x = (const unsigned char *)y;
void func(const unsigned char *arg);
```

```
/tmp$ cat test.h 
#define z_const

x = (z_const unsigned char *)y;
void func(z_const unsigned char *arg);
/tmp$ gcc -E test.h 
# 1 "test.h"
# 1 "<built-in>"
# 1 "<command-line>"
# 31 "<command-line>"
# 1 "/usr/include/stdc-predef.h" 1 3 4
# 32 "<command-line>" 2
# 1 "test.h"


x = ( unsigned char *)y;
void func( unsigned char *arg);
```

**Comment by Dead2** [comment]
Regarding the macro names, can we get a show of hands for/against the current ones or any suggestions for other naming schemes?

**Comment by mtl1979** [comment]
@joshtriplett By default, gcc uses GNU extensions, which might alter the way preprocessor behaves. When we build zlib-ng, we ask it to compile using specific standard and that disables GNU extensions.

**Comment by mtl1979** [comment]
@Dead2 Obviously we can't use macro names that would possibly conflict with already defined macros... If there is no more suggestions for names, I will not change the current ones to random ones just for the sake of experimenting...

**Comment by joshtriplett** [comment]
@mtl1979 Here's the same test again in C89 mode:

```
/tmp$ cat test.h 
#define z_const const

x = (z_const unsigned char *)y;
void func(z_const unsigned char *arg);
/tmp$ gcc -std=c89 -E test.h 
# 1 "test.h"
# 1 "<built-in>"
# 1 "<command-line>"
# 31 "<command-line>"
# 1 "/usr/include/stdc-predef.h" 1 3 4
# 32 "<command-line>" 2
# 1 "test.h"


x = (const unsigned char *)y;
void func(const unsigned char *arg);
```

It even works in `-traditional-cpp` mode. On what platform is the preprocessor *not* handling this?

**Comment by mtl1979** [comment]
@joshtriplett I tested the code on Ubuntu only and got weird errors until I changed to the current style... There was no point in testing on other operating systems, because it already failed on Linux with default gcc version available on 18.04 (x86_64).

Obviously CI tests on all other operating systems and compilers to make sure the code that works on Linux and default gcc version doesn't break on other compilers and operating systems.

**Comment by joshtriplett** [comment]
@mtl1979 I checked via godbolt, and it works with *ancient* GCC as well, and with other compilers. I don't think the problems you were seeing were related to an interaction between parentheses and the C preprocessor, unless there was something more going on. Is it possible that there was a file using `z_const` that didn't include the definition of `z_const`?

(In any case, the fix is much more important than the naming and style of the fix, and it'd always be possible to CI-test further improvements later. Just mystified because I've never heard of a preprocessor issue like what you're describing.)

**Comment by mtl1979** [comment]
@joshtriplett I don't know what the preprocessor did, but compiler did throw an error. I've been programmer for over 30 years but never seen error like that... Getting the patch to work was back then more important than trying to capture the preprocessor output and figure out what went wrong inside the preprocessor...

**Comment by nmoinvaz** [comment]
@Dead2 I'm not that keen about the macro names. 

@mtl1979 What if we did a typedef instead of a macro? Something like `zng_const`?

**Comment by mtl1979** [comment]
@nmoinvaz Using typedef would not be much cleaner as it would still require three versions.

**Comment by nmoinvaz** [comment]
How about just an #ifdef ZLIB_COMPAT #define zng_const const?

**Comment by mtl1979** [comment]
Using ZLIB_COMPAT wouldn't work as ZLIB_COMPAT does not exist in zlib... It has to match 1-to-1 with what zlib itself has...

**Comment by mtl1979** [comment]
I think I figured out one of the weird errors I was getting... Prototype of in_func for inflateBack() was still incorrect in zlib.h... It also needed z_const instead of normal const.

**Comment by nmoinvaz** [comment]
I'm not understanding why the full type needs to be in the preprocessor definition. It seems like "const" is the only difference and that can be specified as a #define. Is there something I'm missing?

**Comment by mtl1979** [comment]
@nmoinvaz We don't want to introduce simple macro in non-compat mode because that would suggest that using it is optional when it is mandatory. We also don't want to use anything that leaks to the binary as a symbol.

If we use simple zng_const, then types like "zng_const char * const" work incorrectly as they should be "const char *" or "char *" instead in compat mode... Those types are distinct in standard C and mixing them will cause error.

**Comment by Dead2** [comment]
Ok, I thought about this, and figured we should try to stay as close to stock zlib as possible if there is no tangible gain.
So I checked, and zlib just uses z_const, example from zlib/deflate.c:415:
` strm->next_in = (z_const Bytef *)dictionary;`

Also zlib does this in zutil.h, a little different from how this PR does it.
```
extern z_const char * const z_errmsg[10]; /* indexed by 2-zlib_error */
/* (size given to avoid silly warnings with Visual C++) */

#define ERR_MSG(err) z_errmsg[Z_NEED_DICT-(err)]
```

So I suggest just using `z_const`, that is a lot simpler in any case and it stays close to stock zlib.
And if someone reports a problem caused by it, we will have a better idea of exactly what needs to be fixed and document it as such.

@mtl1979 Would you mind making a similar PR using `z_const` that we can test with and better decide between the two?

**Comment by mtl1979** [comment]
@Dead2 I already tried with simple z_const and it failed to compile, because implicit conversion of pointers to const are not allowed like I already explained earlier.

**Comment by mtl1979** [comment]
Added it as separate commit to test on CI...

**Comment by Dead2** [comment]
No failing CI (except for codecov uploads of course).


Local native build gives no warnings and did not change compiled size, output compression size or speed.

Local zlib-compat build gives some warnings though.
```
[ 91%] Building C object CMakeFiles/infcover.dir/inftrees.c.o
/root/github/zlib-ng/test/minideflate.c: In function ‘deflate_params’:
/root/github/zlib-ng/test/minideflate.c:81:27: warning: assignment discards ‘const’ qualifier from pointer target type [enabled by default]
         c_stream.next_in  = (const uint8_t *)read_buf;
                           ^
/root/github/zlib-ng/test/minideflate.c: In function ‘inflate_params’:
/root/github/zlib-ng/test/minideflate.c:166:27: warning: assignment discards ‘const’ qualifier from pointer target type [enabled by default]
         d_stream.next_in  = (const uint8_t *)read_buf;
                           ^
/root/github/zlib-ng/test/infcover.c: In function ‘cover_back’:
/root/github/zlib-ng/test/infcover.c:497:5: warning: passing argument 2 of ‘inflateBack’ from incompatible pointer type [enabled by default]
     ret = PREFIX(inflateBack)(&strm, pull, NULL, push, NULL);
     ^
[ 92%] Linking C executable CVE-2003-0107
In file included from /root/github/zlib-ng/test/infcover.c:21:0:
/root/github/zlib-ng/zlib.h:1069:21: note: expected ‘in_func’ but argument is of type ‘unsigned int (*)(void *, const unsigned char **)’
 ZEXTERN int ZEXPORT inflateBack(z_stream *strm, in_func in, void *in_desc, out_func out, void *out_desc);
                     ^
/root/github/zlib-ng/test/infcover.c:502:5: warning: passing argument 2 of ‘inflateBack’ from incompatible pointer type [enabled by default]
     ret = PREFIX(inflateBack)(&strm, pull, NULL, push, &strm);
     ^
In file included from /root/github/zlib-ng/test/infcover.c:21:0:
/root/github/zlib-ng/zlib.h:1069:21: note: expected ‘in_func’ but argument is of type ‘unsigned int (*)(void *, const unsigned char **)’
 ZEXTERN int ZEXPORT inflateBack(z_stream *strm, in_func in, void *in_desc, out_func out, void *out_desc);
                     ^
/root/github/zlib-ng/test/infcover.c:505:5: warning: passing argument 2 of ‘inflateBack’ from incompatible pointer type [enabled by default]
     ret = PREFIX(inflateBack)(&strm, pull, &strm, push, NULL);
     ^
In file included from /root/github/zlib-ng/test/infcover.c:21:0:
/root/github/zlib-ng/zlib.h:1069:21: note: expected ‘in_func’ but argument is of type ‘unsigned int (*)(void *, const unsigned char **)’
 ZEXTERN int ZEXPORT inflateBack(z_stream *strm, in_func in, void *in_desc, out_func out, void *out_desc);
                     ^
/root/github/zlib-ng/test/infcover.c: In function ‘try’:
/root/github/zlib-ng/test/infcover.c:572:9: warning: passing argument 2 of ‘inflateBack’ from incompatible pointer type [enabled by default]
         ret = PREFIX(inflateBack)(&strm, pull, NULL, push, NULL);
         ^
In file included from /root/github/zlib-ng/test/infcover.c:21:0:
/root/github/zlib-ng/zlib.h:1069:21: note: expected ‘in_func’ but argument is of type ‘unsigned int (*)(void *, const unsigned char **)’
 ZEXTERN int ZEXPORT inflateBack(z_stream *strm, in_func in, void *in_desc, out_func out, void *out_desc);
                     ^
```

Other than those warnings, this is looking very good IMO.

mtl1979:
I had funeral today, so I had little time to finish the patch... I will fix the warnings and push again...

nmoinvaz:
PR is looking better. Sorry to hear about the funeral. Hope you are doing ok.

mtl1979:
@nmoinvaz We all want one thing, zlib-ng to be ready for first release... Even though it was funeral of someone really close to me, I do still my best to finish what I have started already... There is quite a lot of different combinations to test to make sure this patch is complete, so I decided it was best to test two of the most common combinations before pushing on GitHub and let CI test as many it can, and afterwards go through the logs. 

For some reason "minideflate" is not built when using configure, so I obviously missed it, because I primarily use configure when testing patches as it requires less manual cleaning up between different variations. 

As "infcover" requires own configure flag and make target, I didn't have enough time this morning to test it... I had to actually look up the correct configure flag from the sources, because it wasn't obvious from output of "configure --help". It's been quite a while since I built it last time.

PR #706, opened 2020-07-21
Various SystemZ DFLTCC fixes

iii-i:
We have found and fixed a number of issues with the hardware acceleration.


nmoinvaz:
With regard to _deflate_quick_ not working due to `windowBits = 13`, you should be able to `#ifdef` where `windowBits = 13` for DFLTCC in deflate.c. There is no requirement that _deflate_quick_ use 8k window. 32k window will be slower than 8k window, but I'm not sure the difference with 32k window in comparison to _deflate_fast_.

iii-i:
A simple test shows that on SystemZ deflate_quick with 32k window is still faster than deflate_fast with 32k window.

iii-i:
Btw, I would mark this as "bug", not "enhancement".
"Fix DFLTCC not flushing EOBS when creating raw streams" fixes a serious issue, which leads to inability to decompress.

deflate.c
```diff
@@ -620,13 +621,13 @@ int32_t ZEXPORT PREFIX(deflateParams)(PREFIX3(stream) *strm, int32_t level, int3
...
-    DEFLATE_PARAMS_HOOK(strm, level, strategy);  /* hook for IBM Z DFLTCC */
+    DEFLATE_PARAMS_HOOK(strm, level, strategy, &hook_flush);  /* hook for IBM Z DFLTCC */
     func = configuration_table[s->level].func;
 
-    if ((strategy != s->strategy || func != configuration_table[level].func) &&
+    if ((strategy != s->strategy || func != configuration_table[level].func || hook_flush != Z_NO_FLUSH) &&
         s->last_flush != -2) {
         /* Flush the last buffer: */
-        int err = PREFIX(deflate)(strm, Z_BLOCK);
+        int err = PREFIX(deflate)(strm, RANK(hook_flush) > RANK(Z_BLOCK) ? hook_flush : Z_BLOCK);
```
Dead2: I think this deserves a comment to avoid future confusion, since it is not obvious.
Alternately an if-else block with comment.

deflate.c
```diff
@@ -1052,8 +1053,6 @@ int32_t ZEXPORT PREFIX(deflate)(PREFIX3(stream) *strm, int32_t flush) {
 
     if (flush != Z_FINISH)
         return Z_OK;
-    if (s->wrap <= 0)
-        return Z_STREAM_END;
```
Dead2, line 1056: This and all the below changes to deflate.c look like they could change compressed output, and need to be tested for that. Have you observed any changes in output on non-DFLTCC builds?
If this results in any changes, we need to understand them completely, to make sure they do not cause regressions on certain input data.

deflate_p.h
```diff
@@ -33,7 +33,10 @@ static inline int zng_tr_tally_lit(deflate_state *s, unsigned char c) {
     s->sym_buf[s->sym_next++] = c;
     s->dyn_ltree[c].Freq++;
     Tracevv((stderr, "%c", c));
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wtype-limits"
```
nmoinvaz: Why these three lines for GCC only? Is there another way this can be done through casting?

deflate.c
iii-i: ok

deflate.c
iii-i: No, this should not change anything for non-DFLTCC builds, and I did not observe changes in compressed data.

wrap can be -1/-2 ("done"), 0 (raw), 1 (zlib) or 2 (gzip).
For wrap values 1 and 2 everything stays exactly as it used to be.
For -1/-2 and 0 (wrap == 0 is the reason why this is needed), there is now an additional `flush_pending(strm)` call. In non-DFLTCC builds there is nothing to flush, so this is a no-op.

deflate_p.h
iii-i: I've seen this warning only with GCC (and not with Clang).

I tried to cast `c` to `unsigned int` in the assertion, but GCC is smart enough to see through that.
So `c`'s parameter type must be changed to `unsigned int` or some such, but I thought there must've been some good reason to make it `unsigned char` in the first place.

mtl1979: Casts don't work in macro parameters as in most cases macro parameters need to be a single token. This is also why in some macros the macro parameters are assigned to variables before use, or there is extra parentheses when accessing macro parameters. In preprocessor, token starts with "(" and ends with ")" or is completely alphanumeric.... 

Some documents say preprocessor tokens can contain also "_" but the result is undefined if there is ambiguity, as you can pretty much rename or split any preprocessor token by defining another token whose name is substring of the name of the first token.

When preprocessor parses a token that results in a string literal, space (ASCII 32) is appended to the output unless the token is followed by "##", which is concatenation operator in preprocessor.

iii-i:
- Added explanation to deflateParams.
- Added fixes for a number of additional issues found with libFuzzer.
- Added the corresponding libFuzzer fuzz target to examples directory.

Dead2:
Both CI fuzzing and static analysis failed, wonder why.. Rerunning

Dead2:
@iii-i Would you please rebase this? The merged #716 should fix at least one of the CI failures, perhaps both.
Other than that, this looks very good and also shows no changes on my x86-64 test machine. 👍

iii-i:
I think fuzzing failed because of the call sequence fuzzer. It's not 100% trivial to build (it's not a single file and it requires protoc). For now I rebased and pulled it out. I'll try to figure out how to build it in zlib-ng context and submit it separately.

.github/workflows/cmake.yml
```diff
@@ -210,6 +211,14 @@ jobs:
...
             codecov: ubuntu_gcc_s390x
 
+          - name: Ubuntu GCC S390X DFLTCC
+            os: ubuntu-latest
+            compiler: s390x-linux-gnu-gcc
+            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-s390x.cmake -DWITH_DFLTCC_DEFLATE=ON -DWITH_DFLTCC_INFLATE=ON
+            packages: qemu gcc-s390x-linux-gnu libc-dev-s390x-cross
+            ldflags: -static
+            codecov: ubuntu_gcc_s390x
+
```
Dead2: I feel like S390X w/DFLTCC should also be tested with  `-DZLIB_COMPAT=ON`, and possibly also with ` -DWITH_SANITIZERS=ON`, what do you think?

iii-i: Sure, let's give it a try. I normally develop it with ASAN on real hardware, let's see how this will look like in QEMU.

Dead2:
Looks good to me

nmoinvaz:
I think I would prefer if the commits related to fixing warnings were moved into a separate PR. They seem to be unrelated to DFLTCC fixes.

iii-i:
* Split fixes for warnings into https://github.com/zlib-ng/zlib-ng/pull/726.

iii-i:
* Fixed stray tabs in `dfltcc_deflate.c`.

arch/s390/dfltcc_deflate.c
```diff
@@ -101,31 +105,38 @@ int ZLIB_INTERNAL dfltcc_deflate(PREFIX3(streamp) strm, int flush, block_state *
...
+     * that DFLTCC has buffered some output in the parameter block and needs to
+     * be called again in order to flush it.
      */
-    if (flush == Z_FINISH && strm->avail_in == 0 && !param->cf) {
-        if (param->bcf) {
-            /* A block is still open, and the hardware does not support closing
-             * blocks without adding data. Thus, close it manually.
-             */
+    if (strm->avail_in == 0 && !param->cf) {
+        /* A block is still open, and the hardware does not support closing
```
nmoinvaz, line 125: Deflate structure has a property called `block_open` . Can `param->cf` be replaced with it? Currently it is only used for `deflate_quick`.

nmoinvaz:
@iii-i can you document the extent of the DFLTCC implementation?  For example right now we have "Deflate hooks for IBM Z DFLTCC" in the README.md but I know that that isn't the extent of the DFLTCC changes. Yes there are hooks, but the DFLTCC implementation does more than that. Can you make a bullet list of what the DFLTCC implementation does? We can put it in the readme or wiki page..

nmoinvaz:
Also thank you for breaking out those commits into a separate PR. Makes it easier to review.

iii-i:
Sure. I think it's worth keeping such documentation closer to the source code. I'll send a separate PR with README.md updates.

arch/s390/dfltcc_deflate.c
iii-i: `param->cf` (Continuation Flag) is more like "did hardware buffer some output bits". `param->bcf`(Block Continuation Flag) is close to `block_open`, but it's set and read by hardware, so no full replacement is possible. However, it might be worth considering syncing the two before and after DFLTCC call.

iii-i: Btw, the instruction is finally publicly documented (this wasn't the case when I first sent the patches). Here is our latest official SystemZ architecture document - http://publibfp.dhe.ibm.com/epubs/pdf/a227832c.pdf. DFLTCC is in Chapter 26 (page 26-16).

iii-i: I have to correct myself: BCF is only read (and never set) by the hardware. I must have been thinking about CF when writing this. Still, the main point holds: since it's a hardware-defined bit, it cannot be fully thrown away.

Dead2: @iii-i Does this mean there are more changes pending? Your last comment was a bit unclear on that.

iii-i: @Dead2 No, not in the context of this PR. This might be an interesting cleanup though.

Issue #666, opened 2020-06-30
Configure silently misbehaves when CHOST is a debian tuple rather than a GNU triplet


mscastanho:
I believe the issue is on `configure:26`:
https://github.com/zlib-ng/zlib-ng/blob/9abe04b926ef688efa3d4f464209246a22afb01d/configure#L25-L28
When `CHOST` is something like `powerpc-linux-gnu`, it will match the second pattern, which will select the last string after the `-`. But this issue happens with any `CHOST` value in the form `first-second-third`, such as `arm-linux-gnueabihf` and `aarch64-linux-gnu`. All these match the second pattern.

_Originally posted by @mscastanho in https://github.com/zlib-ng/zlib-ng/pull/659_

dankegel:
A bit more detail:

This has also been mentioned at http://james.hiebert.name/blog/work/2012/05/23/Gentoo's-libz-mess.html

Evidently CHOST is a variable named used by Gentoo, see https://wiki.gentoo.org/wiki/Changing_the_CHOST_variable
https://bugs.gentoo.org/651908 has a list of all the CHOST values known to gentoo as of 2018.

Debian talks about this, too; see https://wiki.debian.org/Multiarch/Tuples  Evidently Debian uses "target tuples"
and distinguishes them slightly from "gnu triplets", and gives rationale.

autoconf/build-aux/config.guess is the canonical source of gnu triplets:
http://git.savannah.gnu.org/gitweb/?p=autoconf.git;a=blob_plain;f=build-aux/config.guess
libtool has a list of known gnu triplets:
http://git.savannah.gnu.org/cgit/libtool.git/tree/doc/PLATFORMS
clang has a documented class for manipulating gnu triplets, including a method to normalize e.g. debian tuples into canonical triplets:
http://llvm.org/doxygen/Triple_8h_source.html

It's likely that this bug should be entitled "configure silently misbehaves when CHOST is a debian tuple rather than a gnu triplet".

Happily, autoconf also provides a script to canonicalize gnu triplets:
https://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub
And indeed:
```
$ ./config.sub powerpc-linux-gnu
powerpc-unknown-linux-gnu
```
So it's possible configure just needs to run CHOST through config.sub before using it.

dankegel:
It's likely that this bug should be entitled "configure silently misbehaves when CHOST is a debian tuple rather than a gnu triplet".

PR #709, opened 2020-08-02
Fix inconsistent wildcards in CHOST canonicalization script

NiLuJe:
Should prevent mangling actually sane four-part CHOSTs,
while not regressing #666


NiLuJe:
For a few testcases, see basically any of the [crosstool-ng samples](https://github.com/crosstool-ng/crosstool-ng/tree/master/samples) ;).

NiLuJe:
This was breaking system detection in the configure script, which lead to, among other things, failing to detect Linux as Linux, breaking the soname & symbol versioning.

PR #710, opened 2020-08-02
Honor CFLAGS when linking

NiLuJe:
It's the GCC driver, honor CFLAGS to make stuff consistent.
(Allows one to actually get an LTO build w/ debug info, for
instance).

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -265,7 +265,7 @@ $(OBJG): %.o: $(SRCDIR)/%.c
 
 $(SHAREDTARGET): $(PIC_OBJS) $(DEFFILE) $(RCOBJS)
 ifneq ($(SHAREDTARGET),)
-	$(LDSHARED) $(LDSHAREDFLAGS) $(LDFLAGS) -o $@ $(DEFFILE) $(PIC_OBJS) $(RCOBJS) $(LDSHAREDLIBC)
+	$(LDSHARED) $(CFLAGS) $(LDSHAREDFLAGS) $(LDFLAGS) -o $@ $(DEFFILE) $(PIC_OBJS) $(RCOBJS) $(LDSHAREDLIBC)
 ifneq ($(STRIP),)
 	$(STRIP) $@
 endif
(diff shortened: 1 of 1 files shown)


Dead2:
Well, I am for this change, but please note that LTO might not be compatible with how we compile files with different cflags. At least most gcc versions that support LTO only respect the cflag instruction sets passed to the linker, and that breaks when we want different instruction sets per file. I know GCC was working on making LTO remember these settings from the compile phase per file, but I am not aware of whether this has made it to the latest release or not.

We might need to detect an LTO parameter and warn the user or some such to avoid runtime crashes. An alternative is per-function gcc-specific arguments, but that quickly gets messy. LTO with native could work though, or LTO without any instruction-set specific code enabled would also work.

joshtriplett:
Another approach would be to just put the desired link-time flags in `LDFLAGS`.

Dead2:
@joshtriplett It would not help in the LTO case mentioned by the OP, since you can't just choose one combination from mutually exclusive flags combinations, unless you either combine it with a native-style configuration (that would not be able to be redistributed in for example a distro, unless you require all users to have all instruction sets currently present on the buildsystem), or you would need to shed all support for intrinsics.

For non-LTO stuff, the logical way to go is LDFLAGS of course.

joshtriplett:
@Dead2 That's true for instruction-set-specific flags. I was thinking of things like optimization level flags and similar, for which LDFLAGS would work.

But yeah, in general, I don't think LTO is especially compatible with compiling different files with different instruction sets. That's something that might need raising upstream with LLVM and GCC.

NiLuJe:
This is generally no longer an issue since GCC 5... except for stuff like debug information, which is what tripped me up here ;):

> Command-line optimization and target options are now streamed on a per-function basis and honored by the link-time optimizer. This change makes link-time optimization a more transparent replacement of per-file optimizations. It is now possible to build projects that require different optimization settings for different translation units (such as -ffast-math, -mavx, or -finline). Contrary to earlier GCC releases, the optimization and target options passed on the link command line are ignored.

> Note that this applies only to those command-line options that can be passed to optimize and target attributes. Command-line options affecting global code generation (such as -fpic), warnings (such as -Wodr), optimizations affecting the way static variables are optimized (such as -fcommon), debug output (such as -g), and --param parameters can be applied only to the whole link-time optimization unit. In these cases, it is recommended to consistently use the same options at both compile time and link time.

(https://gcc.gnu.org/gcc-5/changes.html)

Dead2:
@NiLuJe Ah, I was not aware that it had gone in in GCC 5. I guess that only makes it a problem with GCC 4, and hopefully not too many are building with LTO on that version.

So it is either this way, or adding an LTO parameter to CMAKE and supporting it officially. (I would have added that already, had I known GCC5 was fine.)

NiLuJe:
Yeah, LTO on GCC < 5 (I'd even go so far as saying < 7), is... not pretty ^^.

NiLuJe:
TL;DR: All in all, this approach should be *mostly harmless* ;).

Issue #708, opened 2020-08-01
loadchunk in chunkcopy_neon can generate unaligned 64bit loads on ARMv7


NiLuJe:
Me again with some more sneaky ARMv7 issues ;).

Since https://github.com/zlib-ng/zlib-ng/commit/e40d88adc9e8180969c6c56a0deb4ec69c3ec92b, `loadchunk` (via `chunkcopy_neon`) will happily generate unaligned 64bit loads:

```
Program received signal SIGBUS, Bus error.
chunkcopy_neon (out=0x9dc53 "", from=0x9dc4b "", len=4) at /home/niluje/Kindle/CrossTool/Build_PW2/zlib-ng/chunkset_tpl.h:23
23      /home/niluje/Kindle/CrossTool/Build_PW2/zlib-ng/chunkset_tpl.h: No such file or directory.
(gdb) bt full
#0  chunkcopy_neon (out=0x9dc53 "", from=0x9dc4b "", len=4) at /home/niluje/Kindle/CrossTool/Build_PW2/zlib-ng/chunkset_tpl.h:23
        chunk = {0, 0, 0, 0, 0, 0, 0, 0, 0, 48, 25, 64, 245, 219, 9, 0}
#1  0x40175206 in chunkcopy_stub (out=0x9dc53 "", from=0x9dc4b "", len=5) at /home/niluje/Kindle/CrossTool/Build_PW2/zlib-ng/functable.c:296
No locals.
#2  0x4017610a in zng_inflate_fast (strm=0xfff6, strm@entry=0x9db80, start=start@entry=0) at /home/niluje/Kindle/CrossTool/Build_PW2/zlib-ng/inffast.c:274
        state = 0x9dbe8
        in = 0x8db22 "\325\312\332\312\332^\211dmggie\017\276\333*Y\222\254H\340\202X\376S"
        last = 0x10000 "\177ELF\001\001\001"
        out = 0xff9e <error: Cannot access memory at address 0xff9e>
        beg = 0x1 <error: Cannot access memory at address 0x1>
        end = 0x10000 "\177ELF\001\001\001"
        safe = 0xadee4 "\016"
        wsize = 646016
        whave = 1075288925
        wnext = 711776
        window = 0xadc4c "\006"
        hold = 48550880
        bits = 23
        lcode = <optimized out>
        dcode = 0xae9b4
        lmask = 0
        dmask = 0
        here = <optimized out>
        op = 65437
        len = 5
        dist = 8
[...]

PR #711, opened 2020-08-02
Fix unaligned double word accesses in chunkset_c on ARMv7

NiLuJe:
c.f., https://github.com/zlib-ng/zlib-ng/issues/708#issuecomment-667622465

### DIFF
--- a/chunkset.c
+++ b/chunkset.c
@@ -5,10 +5,13 @@
 #include "zbuild.h"
 #include "zutil.h"
 
-#ifdef UNALIGNED_OK
+// We need sizeof(chunk_t) to be 8, no matter what.
+#if defined(UNALIGNED64_OK)
 typedef uint64_t chunk_t;
+#elif defined(UNALIGNED_OK)
+typedef struct chunk_t { uint32_t u32[2]; } chunk_t;
 #else
-typedef uint8_t chunk_t[8];
+typedef struct chunk_t { uint8_t u8[8]; } chunk_t;
 #endif
 
 #define HAVE_CHUNKMEMSET_1
@@ -16,18 +19,24 @@ typedef uint8_t chunk_t[8];
 #define HAVE_CHUNKMEMSET_8
 
 static inline void chunkmemset_1(uint8_t *from, chunk_t *chunk) {
-#ifdef UNALIGNED_OK
+#if defined(UNALIGNED64_OK)
     *chunk = 0x0101010101010101 * (uint8_t)*from;
+#elif defined(UNALIGNED_OK)
+    chunk->u32[0] = 0x01010101 * (uint8_t)*from;
+    chunk->u32[1] = chunk->u32[0];
 #else
     memset(chunk, *from, sizeof(chunk_t));
 #endif
 }
 
 static inline void chunkmemset_4(uint8_t *from, chunk_t *chunk) {
-#ifdef UNALIGNED_OK
+#if defined(UNALIGNED64_OK)
     uint32_t half_chunk;
     half_chunk = *(uint32_t *)from;
     *chunk = 0x0000000100000001 * (uint64_t)half_chunk;
+#elif defined(UNALIGNED_OK)
+    chunk->u32[0] = *(uint32_t *)from;
+    chunk->u32[1] = chunk->u32[0];
 #else
     uint8_t *chunkptr = (uint8_t *)chunk;
     memcpy(chunkptr, from, 4);
@@ -36,8 +45,12 @@ static inline void chunkmemset_4(uint8_t *from, chunk_t *chunk) {
 }
 
 static inline void chunkmemset_8(uint8_t *from, chunk_t *chunk) {
-#ifdef UNALIGNED_OK
+#if defined(UNALIGNED64_OK)
     *chunk = *(uint64_t *)from;
+#elif defined(UNALIGNED_OK)
+    uint32_t* p = (uint32_t *)from;
+    chunk->u32[0] = p[0];
+    chunk->u32[1] = p[1];
 #else
     memcpy(chunk, from, sizeof(chunk_t));
 #endif
@@ -48,11 +61,13 @@ static inline void loadchunk(uint8_t const *s, chunk_t *chunk) {
 }
 
 static inline void storechunk(uint8_t *out, chunk_t *chunk) {
-#ifdef UNALIGNED_OK
+#if defined(UNALIGNED64_OK)
     *(uint64_t *)out = *chunk;
+#elif defined(UNALIGNED_OK)
+    ((uint32_t *)out)[0] = chunk->u32[0];
+    ((uint32_t *)out)[1] = chunk->u32[1];
 #else
-    /* Cast to chunk_t pointer to avoid compiler error on MSVC ARM */
-    memcpy((chunk_t *)out, chunk, sizeof(chunk_t ));
+    memcpy(out, chunk, sizeof(chunk_t));
 #endif
 }
 
--- a/deflate.h
+++ b/deflate.h
@@ -351,9 +351,14 @@ static inline void put_uint32_msb(deflate_state *s, uint32_t dw) {
  * IN assertion: there is enough room in pending_buf.
  */
 static inline void put_uint64(deflate_state *s, uint64_t lld) {
-#if defined(UNALIGNED_OK)
+#if defined(UNALIGNED64_OK)
     *(uint64_t *)(&s->pending_buf[s->pending]) = lld;
     s->pending += 8;
+#elif defined(UNALIGNED_OK)
+    *(uint32_t *)(&s->pending_buf[s->pending]) = lld & 0xffffffff;
+    s->pending += 4;
+    *(uint32_t *)(&s->pending_buf[s->pending]) = (lld >> 32) & 0xffffffff;
+    s->pending += 4;
 #else
     put_byte(s, (lld & 0xff));
     put_byte(s, ((lld >> 8) & 0xff));


NiLuJe:
~~Does not yet implement the recommendation from https://github.com/zlib-ng/zlib-ng/issues/708#issuecomment-667706892~~

chunkset.c
```diff
@@ -5,7 +5,7 @@
 #include "zbuild.h"
 #include "zutil.h"
 
-#ifdef UNALIGNED_OK
+#ifdef UNALIGNED64_OK
 typedef uint64_t chunk_t;
```
nmoinvaz, line 10: We can do:
```
#if defined(UNALIGNED64_OK)
typedef uint64_t chunk_t;
#elif defined(UNALIGNED_OK)
typedef uint32_t chunk_t;
#else
typedef uint8_t chunk_t;
#endif
```

But the functions below would also need to support `chunk_t` being `uint32_t`.

NiLuJe:
Re: https://github.com/zlib-ng/zlib-ng/issues/708#issuecomment-667706892

Would something like that work?

```diff
diff --git a/deflate.h b/deflate.h
index 153ec25..6e3d431 100644
--- a/deflate.h
+++ b/deflate.h
@@ -97,7 +97,7 @@ typedef struct tree_desc_s {
 typedef uint16_t Pos;
 
 /* A Pos is an index in the character window. We use short instead of int to
- * save space in the various tables. 
+ * save space in the various tables.
  */
 
 typedef struct internal_state {
@@ -355,14 +355,10 @@ static inline void put_uint64(deflate_state *s, uint64_t lld) {
     *(uint64_t *)(&s->pending_buf[s->pending]) = lld;
     s->pending += 8;
 #else
-    put_byte(s, (lld & 0xff));
-    put_byte(s, ((lld >> 8) & 0xff));
-    put_byte(s, ((lld >> 16) & 0xff));
-    put_byte(s, ((lld >> 24) & 0xff));
-    put_byte(s, ((lld >> 32) & 0xff));
-    put_byte(s, ((lld >> 40) & 0xff));
-    put_byte(s, ((lld >> 48) & 0xff));
-    put_byte(s, ((lld >> 56) & 0xff));
+    *(uint32_t *)(&s->pending_buf[s->pending]) = lld & 0xffffffff;
+    s->pending += 4;
+    *(uint32_t *)(&s->pending_buf[s->pending]) = (lld >> 32) & 0xffffffff;
+    s->pending += 4;
 #endif
 }
```

chunkset.c
NiLuJe: True, but I'd have to actually grok what's happening in there ^^.

I'll take a closer look at it after dinner, but no promises ;).

nmoinvaz:
It looks right. But you would want something like this:
```
#if defined(UNALIGNED64_OK)
     *(uint64_t *)(&s->pending_buf[s->pending]) = lld;
     s->pending += 8;
#elif defined(UNALIGNED_OK)
     *(uint32_t *)(&s->pending_buf[s->pending]) = lld & 0xffffffff;
     s->pending += 4;
     *(uint32_t *)(&s->pending_buf[s->pending]) = (lld >> 32) & 0xffffffff;
     s->pending += 4;
#else
     put_byte(s, (lld & 0xff));
     put_byte(s, ((lld >> 8) & 0xff));
     put_byte(s, ((lld >> 16) & 0xff));
     put_byte(s, ((lld >> 24) & 0xff));
     put_byte(s, ((lld >> 32) & 0xff));
     put_byte(s, ((lld >> 40) & 0xff));
     put_byte(s, ((lld >> 48) & 0xff));
     put_byte(s, ((lld >> 56) & 0xff));
#endif
```

NiLuJe:
@nmoinvaz: Oh, yeah, of course. :facepalm: 

EDIT: In 188c55e

chunkset.c
NiLuJe: Does that look okay? Not *quite* confident about `chunkmemset_4`, but the rest is trivial enough (maybe :D).

```diff
diff --git a/chunkset.c b/chunkset.c
index 121f3a4..423c69e 100644
--- a/chunkset.c
+++ b/chunkset.c
@@ -5,8 +5,10 @@
 #include "zbuild.h"
 #include "zutil.h"
 
-#ifdef UNALIGNED64_OK
+#if defined(UNALIGNED64_OK)
 typedef uint64_t chunk_t;
+#elif defined(UNALIGNED_OK)
+typedef uint32_t chunk_t;
 #else
 typedef uint8_t chunk_t[8];
 #endif
@@ -16,18 +18,24 @@ typedef uint8_t chunk_t[8];
 #define HAVE_CHUNKMEMSET_8
 
 static inline void chunkmemset_1(uint8_t *from, chunk_t *chunk) {
-#ifdef UNALIGNED64_OK
+#if defined(UNALIGNED64_OK)
     *chunk = 0x0101010101010101 * (uint8_t)*from;
+#elif defined(UNALIGNED_OK)
+    *chunk = 0x01010101 * (uint8_t)*from;
 #else
     memset(chunk, *from, sizeof(chunk_t));
 #endif
 }
 
 static inline void chunkmemset_4(uint8_t *from, chunk_t *chunk) {
-#ifdef UNALIGNED64_OK
+#if defined(UNALIGNED64_OK)
     uint32_t half_chunk;
     half_chunk = *(uint32_t *)from;
     *chunk = 0x0000000100000001 * (uint64_t)half_chunk;
+#elif defined(UNALIGNED_OK)
+    uint16_t half_chunk;
+    half_chunk = *(uint16_t *)from;
+    *chunk = 0x00010001 * (uint32_t)half_chunk;
 #else
     uint8_t *chunkptr = (uint8_t *)chunk;
     memcpy(chunkptr, from, 4);
@@ -36,8 +44,10 @@ static inline void chunkmemset_4(uint8_t *from, chunk_t *chunk) {
 }
 
 static inline void chunkmemset_8(uint8_t *from, chunk_t *chunk) {
-#ifdef UNALIGNED64_OK
+#if defined(UNALIGNED64_OK)
     *chunk = *(uint64_t *)from;
+#elif defined(UNALIGNED_OK)
+    *chunk = *(uint32_t *)from;
 #else
     memcpy(chunk, from, sizeof(chunk_t));
 #endif
@@ -48,8 +58,10 @@ static inline void loadchunk(uint8_t const *s, chunk_t *chunk) {
 }
 
 static inline void storechunk(uint8_t *out, chunk_t *chunk) {
-#ifdef UNALIGNED64_OK
+#if defined(UNALIGNED64_OK)
     *(uint64_t *)out = *chunk;
+#elif defined(UNALIGNED_OK)
+    *(uint32_t *)out = *chunk;
 #else
     /* Cast to chunk_t pointer to avoid compiler error on MSVC ARM */
     memcpy((chunk_t *)out, chunk, sizeof(chunk_t ));
```

A very quick bsdtar test on a random tarball appears successful with this, FWIW ;).

nmoinvaz: In `chunkmemset_4`:
```
+#elif defined(UNALIGNED_OK)
+    uint16_t half_chunk;
+    half_chunk = *(uint16_t *)from;
+    *chunk = 0x00010001 * (uint32_t)half_chunk;
```
Should just be:
```
+#elif defined(UNALIGNED_OK)
+    *chunk = *(uint32_t *)from;
```
And for `chunkmemset_8`, I don't think it should exist if defined `UNALIGNED64_OK` is not defined and neither should be `HAVE_CHUNKMEMSET_8`. So wrap those in `#ifdef UNALIGNED64_OK`..

NiLuJe: Should be taken care of w/ e09ef09, thanks for the pointers (ha!)!

NiLuJe:
@nmoinvaz: Oops ^^. I think I got everything in 6516925. It builds, and bsdtar still doesn't implode ;).

nmoinvaz (approved):
LGTM

chunkset.c
```diff
@@ -36,8 +44,12 @@ static inline void chunkmemset_4(uint8_t *from, chunk_t *chunk) {
 }
 
 static inline void chunkmemset_8(uint8_t *from, chunk_t *chunk) {
-#ifdef UNALIGNED_OK
+#if defined(UNALIGNED64_OK)
     *chunk = *(uint64_t *)from;
+#elif defined(UNALIGNED_OK)
+    uint64_t full_chunk = *(uint64_t *)from;
```
mtl1979: This is still possibly 64-bit unaligned read.

NiLuJe: The fact that it's *declared* as an uint64_t on the stack *should* mean it's properly aligned, though, right?

NiLuJe: Oh, wait, the *read*, not the store. :D. :facepalm:

mtl1979: The problem is ```from``` which is not guaranteed to be on stack and even if it is on stack, it might not be aligned on 64-bit boundary. Stack can be aligned on either 32-bit or 64-bit boundary depending on system.

On release builds, optimizer might not generate any code for this line, but debug builds must preserve all local variables that are used.

NiLuJe: Although, `from` itself is an ` uint8_t *`, so it's *probably* never going to be an issue?

(My brain hurts >_<).

In any case, GCC appears to be chunking that in two 32bits loads anyway:

```
3980   │ static inline void chunkmemset_8(uint8_t *from, chunk_t *chunk) {
3981   │ #if defined(UNALIGNED64_OK)
3982   │     *chunk = *(uint64_t *)from;
3983   │ #elif defined(UNALIGNED_OK)
3984   │     uint64_t full_chunk = *(uint64_t *)from;
3985   │     3bb2:   682c        ldr r4, [r5, #0]
3986   │     3bb4:   686b        ldr r3, [r5, #4]
```

NiLuJe: I guess I can make that explicit by doing it this way instead (i.e., relying on good old pointer arithmetic rules)?

```C
    uint32_t* p = (uint32_t *)from;
    *chunk[0] = *p;
    *chunk[1] = *(p+1);
```

```
3979   │ static inline void chunkmemset_8(uint8_t *from, chunk_t *chunk) {
3980   │ #if defined(UNALIGNED64_OK)
3981   │     *chunk = *(uint64_t *)from;
3982   │ #elif defined(UNALIGNED_OK)
3983   │     uint32_t* p = (uint32_t *)from;
3984   │     *chunk[0] = *p;
3985   │     3bb2:   682c        ldr r4, [r5, #0]
3986   │     *chunk[1] = *(p+1);
3987   │     3bb4:   686b        ldr r3, [r5, #4]
```

mtl1979: Being ```uint8_t *``` only implies that the pointer is assumed to be unaligned. When you cast it to wider type, you essentially create temporary variable that must be assigned to another variable with same alignment restrictions.... Basically you can only assign the value of memory it references to, to vector of ```uint8_t``` or another memory location pointed by another variable of type ```uint8_t *```. Assigning values explicitly to registers was removed from recent gcc versions, even though compiler itself might temporarily use registers to store the variable contents.

Basically this code relies on compiler generating the correct code instead of writing it as two 32-bit copies explicitly.

mtl1979: Or even...

```
uint32_t* p = (uint32_t *)from;
chunk[0] = p[0];
chunk[1] = p[1];
```

NiLuJe: This has a no impact on register usage once inlined in `chunkcopy_c`:

```
4066   │     uint64_t full_chunk = *(uint64_t *)from;
4067   │     3c0e:   680c        ldr r4, [r1, #0]
4068   │     3c10:   684b        ldr r3, [r1, #4]
4069   │     ((uint32_t *)out)[0] = *chunk[0];
4070   │     3c12:   6004        str r4, [r0, #0]
4071   │     ((uint32_t *)out)[1] = *chunk[1];
4072   │     3c14:   6043        str r3, [r0, #4]
```

```
4067   │     *chunk[0] = *p;
4068   │     3c0e:   680c        ldr r4, [r1, #0]
4069   │     *chunk[1] = *(p+1);
4070   │     3c10:   684b        ldr r3, [r1, #4]
4071   │     ((uint32_t *)out)[0] = *chunk[0];
4072   │     3c12:   6004        str r4, [r0, #0]
4073   │     ((uint32_t *)out)[1] = *chunk[1];
4074   │     3c14:   6043        str r3, [r0, #4]
```

NiLuJe: And a minor one (for the better?) in `CHUNKMEMSET`:

```
6140   │     uint64_t full_chunk = *(uint64_t *)from;
6141   │     4af8:   682a        ldr r2, [r5, #0]
6142   │     4afa:   686b        ldr r3, [r5, #4]
6143   │     *chunk[0] = (uint32_t)(full_chunk & 0xffffffff);
6144   │     4afc:   60ba        str r2, [r7, #8]
6145   │     *chunk[1] = (uint32_t)(full_chunk >> 32);
6146   │     4afe:   613b        str r3, [r7, #16]
```

```
6143   │     *chunk[0] = *p;
6144   │     4af8:   682b        ldr r3, [r5, #0]
6145   │     4afa:   60bb        str r3, [r7, #8]
6146   │     *chunk[1] = *(p+1);
6147   │     4afc:   686b        ldr r3, [r5, #4]
6148   │     4afe:   613b        str r3, [r7, #16]
```

NiLuJe: @mtl1979: Gah, I missed your comments, but it looks like we had mostly the same idea at around the same time ;).

NiLuJe: Now that I'm caught up, is there a preference towards the array index notation vs. the pointer arithmetic one?

mtl1979: When one of the variables is type[], then it makes more sense to use array notation on both sides even if the other variable is type* as it is more consistent and easier to read. Pointer arithmetics should only be used when all variables are type*, or when using operator++ etc. to advance the pointer itself.

chunkset.c
```diff
@@ -5,8 +5,10 @@
...
 #include "zutil.h"
 
-#ifdef UNALIGNED_OK
+#if defined(UNALIGNED64_OK)
 typedef uint64_t chunk_t;
+#elif defined(UNALIGNED_OK)
+typedef uint32_t chunk_t[2];
 #else
 typedef uint8_t chunk_t[8];
 #endif
```
mtl1979, line 15: This is completely wrong... size of chunk_t is always correct, but the type in function prototype/parameter list is not... Mixing arrays with non-arrays in typedef makes the code confusing and prone to errors.

We don't want uint32_t*[2] as the function parameter type. we want uint32_t* when UNALIGNED64_OK not defined.
We don't want uint8_t*[8] as the function parameter type. we want uint8_t* when neither UNALIGNED_OK or UNALIGNED64_OK is defined..

chunkset.c
```diff
@@ -5,12 +5,13 @@
 #include "zbuild.h"
 #include "zutil.h"
 
+// We need sizeof(chunk_t) to be 8, no matter what.
 #if defined(UNALIGNED64_OK)
 typedef uint64_t chunk_t;
 #elif defined(UNALIGNED_OK)
-typedef uint32_t chunk_t[2];
+typedef struct { uint32_t u32[2]; } chunk_t;
```
mtl1979: If we really want to support pre-C11 compilers, then the start of the line should be modified as "typedef struct chunk_t {"... so the struct itself has same name as the variable type the typedef declares.

Same below ;)

nmoinvaz (approved):
Looks good.

chunkset.c
Dead2: We have stated that we target C99 compilers, so we should make sure we keep it compatible.
I don't see anything really interesting (for us) in C11 anyways.

mtl1979: That's why I commented how this particular line should be modified -- here -- instead of continuing the discussion on the original location. Sometimes long conversations start to wander around and it becomes hard to follow which line(s) are being discussed.

NiLuJe: Done! (Sorry, was away on vacation ;)).

Dead2:
Looks good to me.

PR #716, opened 2020-08-14
Update clang static analyzer to version 12

gdevenyi:
It looks like clang-tools-11 has disappeared from the llvm repo


Dead2:
Thanks 👍

Issue #705, opened 2020-07-18
compat mode not yet ABI compatible with classic zlib on 32 bit machines


dankegel:
While looking into #700, I noticed a few z_off64_t problems when testing with -m32, see https://github.com/dankegel/zlib-ng/runs/883759283 :
```
  [C]'function int gzbuffer(gzFile, unsigned int)' at gzlib.c:245:1 has some indirect sub-type changes:
    parameter 1 of type 'typedef gzFile' has sub-type changes:
      underlying type 'gzFile_s*' changed:
        in pointed to type 'struct gzFile_s' at zlib.h:1761:1:
          type size changed from 128 to 96 bits
          1 data member change:
           type of 'off64_t gzFile_s::pos' changed:
             underlying type 'typedef __off64_t' at types.h:140:1 changed:
               underlying type 'typedef __int64_t' changed:
                 entity changed from 'typedef __int64_t' to compatible type 'long int'
                   type name changed from 'long long int' to 'long int'
                   type size changed from 64 to 32 bits

  [C]'function off64_t gzseek64(gzFile, off64_t, int)' at gzlib.c:289:1 has some indirect sub-type changes:
    return type changed:
    parameter 2 of type 'typedef off64_t' changed:

  [C]'function off64_t gztell64(gzFile)' at gzlib.c:370:1 has some indirect sub-type changes:
    return type changed:
```

On a 32 bit box (ubuntu 16.04), I also saw:
```
  [C]'function uLong adler32_combine64(uLong, uLong, off64_t)' at adler32.c:135:1 has some indirect sub-type changes:
    parameter 3 of type 'typedef off64_t' changed:
      underlying type 'typedef __off64_t' at types.h:131:1 changed:
        underlying type 'typedef __quad_t' changed:
          entity changed from 'typedef __quad_t' to compatible type 'long int'
            type name changed from 'long long int' to 'long int'
            type size changed from 64 to 32 bits
```

Evidently zconf.h never sets Z_LARGE64 anymore since the recent commit
"For gzseek, gzoffset, gzopen, adler32_combine, crc32_combine and crc32_combine_gen, export 32-bit and 64-bit versions for zlib-compatible api and only 64-bit version (without 64 suffix) for zlib-ng native api",
https://github.com/zlib-ng/zlib-ng/commit/6dcc7bf815c803fe7712e449b3c8fee8f3ab6dda#diff-ce3e5ac484a4ca73e6533421950dbe15
[...]

PR #717, opened 2020-08-15
test/abicheck.sh: new script to verify abi compatibility with older versions

### LINKED ISSUE #699: No test currently verifies whether zlib-ng ABI is stable
Many consumers of zlib-ng assume that each new minor release (or, let's face it, each new commit)
yields a library that is ABI-compatible with the previous one, i.e. that they can just install
a prebuilt zlib-ng shared library (say, from homebrew or apt) and their program will
still run properly without being recompiled.

This is important enough that there are websites that track the ABI compatibility of libraries like zlib over time, e.g. https://abi-laboratory.pro/?view=timeline&l=zlib

zlib-ng should have a test that checks the ABI against that provided by a prior reference commit, and fails if there are any incompatible changes.

And when building with --zlib-compat, that prior reference commit should probably be the most recent release of original zlib.

dankegel:
Verifies that zlib-ng's ABI has not changed since the reference commit
(indicated by variables ABI_URL and ABI_COMMIT in the script).
If --zlib-compat is given, the reference commit is zlib's 1.2.11;
otherwise, it's zlib-ng's 1d2504ddc489 (for now).
    
Ignores new symbols entirely, as they probably don't break backwards compatibility.

Ignores warnings listed in test/abi/ignore, currently, just those related to internal_state or z_stream*.
    
If --refresh is given, actually checks out the reference commit,
builds it, and stores its ABI description into an .abi file;
otherwise just uses the .abi file saved in git for that CHOST.
--refresh_if is similar, but only refreshes if the .abi file is not present already.
    
Fixes https://github.com/zlib-ng/zlib-ng/issues/699
    
Known issues:
    - although it supports -m32 (if in both CFLAGS and LDFLAGS), it doesn't yet support -32 in CONFIGURE_ARGS.
    - only includes abi definitions for x86_64 and armhf; the rest can be added in a followon commit (see --refresh).
    - skips -m32 tests (loudly) because that's broken at moment, see #705

### DIFF
--- a/test/abicheck.sh
+++ b/test/abicheck.sh
@@ -1,0 +1,164 @@
+#!/bin/sh
+set -ex
+TESTDIR="$(cd $(dirname "$0"); pwd)"
+
+usage() {
+    cat <<_EOF_
+Usage: $0 [--zlib-compat][--refresh][--refresh-if]
+
+Build shared library with -ggdb, then compare its ABI to the stable
+ABI, and abort if differences found.
+
+Options:
+--zlib-compat  - check the ABI of the zlib-compatible flavor of zlib-ng.
+--refresh      - build the reference library and extract its ABI rather than using a stored ABI file.
+--refresh-if   - refresh only if ABI file not present.
+
+Obeys CHOST, CONFIGURE_ARGS, CFLAGS, and LDFLAGS.
+
+Requires libabigail (on Ubuntu, install package abigail-tools).
+_EOF_
+}
+
+# Print the multiarch tuple for the current (non-cross) machine,
+# or the empty string if unavailable.
+detect_chost() {
+    dpkg-architecture -qDEB_HOST_MULTIARCH ||
+     $CC -print-multiarch ||
+     $CC -print-search-dirs | sed 's/:/\n/g' | grep -E '^/lib/[^/]+$' | sed 's%.*/%%' ||
+     true
+}
+
+if ! test -f "configure"
+then
+  echo "Please run from top of source tree"
+  exit 1
+fi
+
+suffix="-ng"
+CONFIGURE_ARGS_NG="$CONFIGURE_ARGS"
+refresh=false
+refresh_if=false
+for arg
+do
+  case "$arg" in
+  --zlib-compat)
+    suffix=""
+    CONFIGURE_ARGS_NG="$CONFIGURE_ARGS_NG --zlib-compat"
+    ;;
+  --refresh)
+    refresh=true
+    ;;
+  --refresh_if)
+    refresh_if=true
+    ;;
+  --help)
+    usage
+    exit 0
+    ;;
+  *)
+    echo "Unknown arg '$arg'"
+    usage
+    exit 1
+    ;;
+  esac
+done
+
+# Choose reference repo and commit
+if test "$suffix" = ""
+then
+  # Reference is zlib 1.2.11
+  ABI_GIT_REPO=https://github.com/madler/zlib.git
+  ABI_GIT_COMMIT=v1.2.11
+else
+  # Reference should be the tag for zlib-ng 2.0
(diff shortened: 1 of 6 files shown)


nmoinvaz:
What is the difference between `pkgcheck` and `abicheck`?

test/abicheck.sh
```diff
@@ -0,0 +1,152 @@
...
+  make -j2
+  cd ..
+  # Find shared library, extract its abi
+  dylib1=$(find btmp1 -type f -name '*.dylib*' -print -o -type f -name '*.so.*' -print)
+  abidw $dylib1 > "$ABIFILE"
+  # Maintainers may wish to check $ABIFILE into git when a new
+  # target is added, or when a major release happens that is
+  # intended to change the ABI.  Alternately, this script could
+  # just always rebuild the reference source, and dispense with
+  # caching abi files in git (but that would slow builds down).
```
nmoinvaz, line 134: Do you know how long this would add?

test/abicheck.sh
```diff
@@ -0,0 +1,152 @@
...
+cd ..
+# Find shared library, extract its abi
+dylib2=$(find btmp2 -type f -name '*.dylib*' -print -o -type f -name '*.so.*' -print)
+abidw $dylib2 > btmp2/zlib${suffix}-built.abi
+
+# Compare it to the reference
+# FIXME: use --no-added-syms for now, but we probably want to be more strict.
+if abidiff --no-added-syms --suppressions test/abi/ignore "$ABIFILE" btmp2/zlib${suffix}-built.abi
+then
+  echo "pkgcheck-abi-check PASS"
```
nmoinvaz: I noticed multiple times it called `pkgcheck-abi-check`, maybe shorten it to `abicheck`? So in `pkgcheck.sh` it says "pkgcheck: PASS". Maybe in `abicheck.sh` it should say "abicheck: PASS". There are a couple of other instances above too.

nmoinvaz (commented):
Overall looks good!

dankegel:
pkgcheck compares output of build systems on same source tree, and says which files differ.

Abicheck compares shared libraries generated by different source trees, and says which variable or function went incompatible.

test/abicheck.sh
Dead2: @dankegel *nudge*

dankegel:
I'm employed now, so will probably look at things on the weekend.

test/abicheck.sh
nmoinvaz: @Dead2 perhaps we should move ahead with this PR.

dankegel: I have an updated branch with fixes, will try to update this Mr tomorrow.

dankegel:
If you're impatient, you can get a peek at a rebased and improved version at https://github.com/dankegel/zlib-ng/commits/abicheck-tmpsssh

Some tests have been queued overnight... I wonder if that's a sign of throttling, in which case maybe checking in those .abi files might help.

nmoinvaz:
I would rerun the jobs if you can. It is probably just GHA acting up.

dankegel:
The three failures are spurious (coverage upload).  Ready for re-review.

Dead2:
I feel like this tool should get some documentation, either in perhaps a `.github/workflows/pkgcheck.txt` file or in the wiki. (The former is probably more robust since it lives in git, the wiki only lives on github but is more convenient to edit.

We need to make sure it is simple and foolproof to maintain this. Personally I would not know how (yet), and that makes it a hurdle if I need how to learn to use abigail for example. If we have a guide telling us when/why things should be done and step-by-step instructions on how to do those, it would make it a lot easier to maintain in the future.

dankegel:
OK, added https://github.com/dankegel/zlib-ng/blob/bug699-abi-checker/test/abicheck.md 
How's that look?

PR #719, opened 2020-08-16
Harmonize the CPU architecture preprocessor definitions.

nmoinvaz:


Dead2:
Since this gates the _arch_ features check function and its dependants, and does not really enable support for the arch itself, I think perhaps _FEATURES might be a more descriptive name.
An even more descriptive name might be such as X86_RUNTIME_CHECKS, but idk, it might be more easily misunderstood.

nmoinvaz:
Ok I have renamed it to _FEATURES as you suggested.

Dead2:
Looks good to me.

PR #721, opened 2020-08-16
zlib.h: Use unsigned long for totals, for compatibility with zlib on 64-bit Windows

joshtriplett:
Stock zlib uses unsigned long for total_in and total_out. On 64-bit
Windows, that isn't the same as size_t.

### DIFF
--- a/test/example.c
+++ b/test/example.c
@@ -16,6 +16,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <inttypes.h>
+#include <stdint.h>
 
 #define TESTFILE "foo.gz"
 
@@ -452,7 +453,7 @@ void test_large_inflate(unsigned char *compr, size_t comprLen, unsigned char *un
     CHECK_ERR(err, "inflateEnd");
 
     if (d_stream.total_out != 2*uncomprLen + diff) {
-        fprintf(stderr, "bad large inflate: %zu\n", d_stream.total_out);
+        fprintf(stderr, "bad large inflate: %" PRIu64 "\n", (uint64_t)d_stream.total_out);
         exit(1);
     } else {
         printf("large_inflate(): OK\n");
--- a/test/fuzz/example_large_fuzzer.c
+++ b/test/fuzz/example_large_fuzzer.c
@@ -109,7 +109,7 @@ void test_large_inflate(unsigned char *compr, size_t comprLen, unsigned char *un
     CHECK_ERR(err, "inflateEnd");
 
     if (d_stream.total_out != 2 * uncomprLen + diff) {
-        fprintf(stderr, "bad large inflate: %zu\n", d_stream.total_out);
+        fprintf(stderr, "bad large inflate: %" PRIu64 "\n", (uint64_t)d_stream.total_out);
         exit(1);
     }
 }
--- a/test/infcover.c
+++ b/test/infcover.c
@@ -10,6 +10,8 @@
 #include <string.h>
 #undef NDEBUG
 #include <assert.h>
+#include <inttypes.h>
+#include <stdint.h>
 
 /* get definition of internal structure so we can mess with it (see pull()),
    and so we can call inflate_trees() (see cover5()) */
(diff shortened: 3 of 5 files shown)


joshtriplett:
Re-pushed to use `unsigned long` directly; `z_size_t` isn't defined there, and in any case, these fields are defined directly as `unsigned long` in stock zlib, even if `size_t` exists.

mtl1979:
Looks good to me...

joshtriplett:
Assuming this passes CI, I'd love to get this merged, to unblock running CI for a [downstream project](https://github.com/rust-lang/libz-sys) on Windows.

mtl1979:
@joshtriplett You need two passed reviews from long-time contributors...

joshtriplett:
@mtl1979 Ah, good to know, thank you.

nmoinvaz:
It looks like this introduces some warnings on Windows GCC CI instance in the tests.

mtl1979:
@nmoinvaz Fixing the tests should be separated as own pull request in my opinion.

nmoinvaz:
They could be a separate commit also.

mtl1979:
@nmoinvaz Glancing quickly at the affected file, it seems this pull request reveals bigger issue in that file, not just the change of type for total_in and total_out.

joshtriplett:
Looking over the warnings, it looks like there are two separate issues:

1) That GCC on Windows doesn't know about `%zu` to print `size_t` values, and issues multiple warnings every time it's used. That's not really related to this patch; it affects *every* use of `%zu`. That's hard to fix portably, without relying on definitions that some platforms don't have. Current versions of Windows/MSVC do understand `%zu`, but much older versions don't; those older versions also don't understand `%llu`, and don't provide `PRIu64`.
2) Prints of `total_in` and `total_out` shouldn't use `%zu` now, they should use `%lu`, but only when building in zlib-compat mode. The easiest fix for that is to cast to a type that'll always be at least as big as *either* `unsigned long` *or* `size_t`, and print that. Again hard to fix portably, though, for the same reason as (1).

Fortunately, both warnings only occur in tests, not in any public function, and this should not affect users of zlib-ng in practice.

The most portable fix I know of involves detecting if `inttypes.h` is available, including it if available, and then conditionally defining `PRIu64` if not already defined, to `I64u` on Microsoft compilers and `llu` everywhere else. Then cast to `uint64_t` to print (which itself may require defining on older compilers if `stdint.h` isn't available).

Note that every warning that would fix already exists prior to this commit.

joshtriplett:
@mtl1979 @nmoinvaz Turns out that zlib-ng already counts on the availability of `inttypes.h` and `stdint.h`, so I used those to fix the warnings. With the additional commit, this *should* fix the warnings in CI.

joshtriplett:
Well, that now produces warnings like this:
```
test/example.c:456:25: warning: ISO C does not support the 'I64' ms_printf length modifier [-Wformat=]
         fprintf(stderr, "bad large inflate: %" PRIu64 "\n", (uint64_t)d_stream.total_out);
                         ^~~~~~~~~~~~~~~~~~~~~~
test/example.c:456:25: warning: ISO C does not support the 'I64' ms_printf length modifier [-Wformat=]
```
If it doesn't like `zu`, and it doesn't like what `PRIu64` expands to, then I don't think there's any reasonable way for that test code to avoid warnings. With the added patch, however, the code should work correctly on all platforms, even if it produces warnings on Windows with GCC.

joshtriplett:
Would it be reasonable to merge this PR as a net improvement, and leave further warning fixes to someone with access to a Windows system? (This PR does not add any new warnings; the warnings were already present.)

nmoinvaz (approved):
Looks good.

nmoinvaz:
I have approved this PR. To fix those other warnings it looks like we may need something in CMake such as:
```
if(MINGW) 
    add_definition(-D__USE_MINGW_ANSI_STDIO=1)
endif()
```
For the test apps at least.

See: https://stackoverflow.com/questions/44382862/how-to-printf-a-size-t-without-warning-in-mingw-w64-gcc-7-1

But perhaps it might be outside of the scope of this PR.

Dead2:
Looks good to me.

@nmoinvaz Could that define perhaps be added to a header file, like zbuild/zutil perhaps?
Kind of stupid that MSVC does not support selecting C99 standard for example, to silence warnings like that at least, ISO C is getting kind of old by now..

mtl1979:
@Dead2 MSVC does support selecting C standard, but Windows's gcc lags behind for compatibility with older compiler versions that didn't support selecting the standard.

Dead2:
@mtl1979 Really? I found no MSVC compiler switch to select C99, only to select between C++ versions and one to force C no matter what file extension.

mtl1979:
@Dead2 When Microsoft does something, it doesn't do things the same way as gcc for other operating systems... It doesn't separate C and C++ because it is basically just C++ compiler that supports disabling some C++ features for compatibility.

The range of features Visual C++ supports is based on what toolkit version Visual Studio is configured to use. There is minimum standard version it supports and maximum standard version it supports...

Dead2:
@mtl1979 Yes, and that was my point exactly.. MSVC does not support C properly, just on a best-effort basis, and often poorly at that.

mtl1979:
@Dead2 Even with C++, MSVC works on best-effort basis... The standard version option doesn't match the standards 1:1... Some features can't be disabled (or enabled for all targets).

PR #724, opened 2020-08-19
 Move zero check for insert_string count to fill_window.

nmoinvaz:

PR #725, opened 2020-08-19
Simplify and optimize insert_string template

Dead2:
This moves unnecessary code out of the loop in insert_string, and adds a compiler hint to both quick_insert_string and insert_string.

Result is ~0.4% faster in my benchmarks.
See benchmarks in comment below.

PS: I recommend looking at/reviewing the two commits separately, it is split up to make it easier to understand why the changes were made.

### DIFF
--- a/insert_string_tpl.h
+++ b/insert_string_tpl.h
@@ -61,7 +61,7 @@ ZLIB_INTERNAL Pos QUICK_INSERT_STRING(deflate_state *const s, const uint32_t str
  *    (except for the last MIN_MATCH-1 bytes of the input file).
  */
 ZLIB_INTERNAL Pos INSERT_STRING(deflate_state *const s, const uint32_t str, uint32_t count) {
-    Pos head, idx, ret = 0;
+    Pos head = 0, idx;
     uint8_t *strstart = s->window + str;
     uint8_t *strend = strstart + count - 1; /* last position */
     uint32_t hash_mask = s->hash_mask;
@@ -88,12 +88,6 @@ ZLIB_INTERNAL Pos INSERT_STRING(deflate_state *const s, const uint32_t str, uint
         }
     }
 
-    if (strstart == strend) {
-        if (head != idx)
-            ret = head;
-        else
-            ret = idx;
-    }
-    return ret;
+    return head;
 }
 #endif
--- a/insert_string_tpl.h
+++ b/insert_string_tpl.h
@@ -45,7 +45,7 @@ ZLIB_INTERNAL Pos QUICK_INSERT_STRING(deflate_state *const s, const uint32_t str
     hm = h & s->hash_mask;
 
     head = s->head[hm];
-    if (head != str) {
+    if (LIKELY(head != str)) {
         s->prev[str & s->w_mask] = head;
         s->head[hm] = str;
     }
@@ -61,7 +61,7 @@ ZLIB_INTERNAL Pos QUICK_INSERT_STRING(deflate_state *const s, const uint32_t str
  *    (except for the last MIN_MATCH-1 bytes of the input file).
  */
 ZLIB_INTERNAL Pos INSERT_STRING(deflate_state *const s, const uint32_t str, uint32_t count) {
-    Pos idx, ret = 0;
+    Pos head, idx, ret = 0;
     uint8_t *strstart = s->window + str;
     uint8_t *strend = strstart + count - 1; /* last position */
     uint32_t hash_mask = s->hash_mask;
@@ -81,15 +81,18 @@ ZLIB_INTERNAL Pos INSERT_STRING(deflate_state *const s, const uint32_t str, uint
         UPDATE_HASH(s, h, val);
         hm = h & hash_mask;
 
-        Pos head = s->head[hm];
-        if (head != idx) {
+        head = s->head[hm];
+        if (LIKELY(head != idx)) {
             s->prev[idx & s->w_mask] = head;
             s->head[hm] = idx;
-            if (strstart == strend)
-                ret = head;
-        } else if (strstart == strend) {
-            ret = idx;
         }
+    }
+
+    if (strstart == strend) {
+        if (head != idx)
+            ret = head;
+        else
+            ret = idx;
     }
     return ret;
 }
--- a/insert_string_tpl.h
+++ b/insert_string_tpl.h
@@ -60,13 +60,12 @@ ZLIB_INTERNAL Pos QUICK_INSERT_STRING(deflate_state *const s, const uint32_t str
  *    input characters and the first MIN_MATCH bytes of str are valid
  *    (except for the last MIN_MATCH-1 bytes of the input file).
  */
-ZLIB_INTERNAL Pos INSERT_STRING(deflate_state *const s, const uint32_t str, uint32_t count) {
-    Pos head = 0, idx;
+ZLIB_INTERNAL void INSERT_STRING(deflate_state *const s, const uint32_t str, uint32_t count) {
     uint8_t *strstart = s->window + str;
     uint8_t *strend = strstart + count - 1; /* last position */
     uint32_t hash_mask = s->hash_mask;
 
-    for (idx = str; strstart <= strend; idx++, strstart++) {
+    for (Pos idx = str; strstart <= strend; idx++, strstart++) {
         uint32_t val, hm, h = 0;
 
 #ifdef UNALIGNED_OK
@@ -81,13 +80,11 @@ ZLIB_INTERNAL Pos INSERT_STRING(deflate_state *const s, const uint32_t str, uint
         UPDATE_HASH(s, h, val);
         hm = h & hash_mask;
 
-        head = s->head[hm];
+        Pos head = s->head[hm];
         if (LIKELY(head != idx)) {
             s->prev[idx & s->w_mask] = head;
             s->head[hm] = idx;
         }
     }
-
-    return head;
 }
 #endif
--- a/functable.c
+++ b/functable.c
@@ -15,11 +15,11 @@
 #endif
 
 /* insert_string */
-extern Pos insert_string_c(deflate_state *const s, const uint32_t str, uint32_t count);
+extern void insert_string_c(deflate_state *const s, const uint32_t str, uint32_t count);
 #ifdef X86_SSE42_CRC_HASH
-extern Pos insert_string_sse4(deflate_state *const s, const uint32_t str, uint32_t count);
+extern void insert_string_sse4(deflate_state *const s, const uint32_t str, uint32_t count);
 #elif defined(ARM_ACLE_CRC_HASH)
-extern Pos insert_string_acle(deflate_state *const s, const uint32_t str, uint32_t count);
+extern void insert_string_acle(deflate_state *const s, const uint32_t str, uint32_t count);
 #endif
 
 /* quick_insert_string */
@@ -144,7 +144,7 @@ ZLIB_INTERNAL void cpu_check_features(void)
 }
 
 /* stub functions */
-ZLIB_INTERNAL Pos insert_string_stub(deflate_state *const s, const uint32_t str, uint32_t count) {
+ZLIB_INTERNAL void insert_string_stub(deflate_state *const s, const uint32_t str, uint32_t count) {
     // Initialize default
 
     functable.insert_string = &insert_string_c;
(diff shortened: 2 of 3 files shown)


Dead2:
## Deflatebench config
 Runs: 70
 Levels: 1-9
 Trimworst: 40

Level 2: /tmp/deflatebench-2.tmp 135.1 MiB / 141,626,880 B
Level 3: /tmp/deflatebench-3.tmp 105.1 MiB / 110,154,240 B
Level 4: /tmp/deflatebench-4.tmp 90.0 MiB / 94,417,920 B
Level 5: /tmp/deflatebench-5.tmp 90.0 MiB / 94,417,920 B
Level 6: /tmp/deflatebench-6.tmp 75.0 MiB / 78,681,600 B
Level 7: /tmp/deflatebench-7.tmp 60.0 MiB / 62,945,280 B
Level 8: /tmp/deflatebench-8.tmp 45.0 MiB / 47,208,960 B
Level 9: /tmp/deflatebench-9.tmp 45.0 MiB / 47,208,960 B

## Baseline 97c115946f4162599773bef28dd1cc92c0459bf4
```
  93327    1272      32   94631   171a7 libz-ng.so.1

 1      57.870%    1.835/1.869/1.882       1.190/1.230/1.243  163918126
 2      43.903%    1.812/1.849/1.859       0.620/0.639/0.648  62178636
 3      42.510%    1.713/1.724/1.731       0.464/0.481/0.489  46826710
 4      41.490%    1.673/1.685/1.693       0.382/0.404/0.412  39174432
 5      41.215%    1.818/1.834/1.843       0.370/0.402/0.410  38913920
 6      41.038%    1.820/1.834/1.840       0.314/0.332/0.340  32289128
 7      40.784%    1.998/2.006/2.010       0.248/0.265/0.271  25671760
 8      40.707%    1.925/1.944/1.951       0.187/0.197/0.203  19217358
 9      40.696%    2.141/2.157/2.165       0.176/0.193/0.200  19212387

 avg1   43.357%                1.878                   0.460
 tot                         507.061                 124.285
```

## PR #725 03076155b4188137678d3d9c8bc75ab9b0fc4669
```
  93199    1272      32   94503   17127 libz-ng.so.1

 1      57.870%    1.844/1.868/1.880       1.173/1.217/1.234  163918126
 2      43.903%    1.824/1.841/1.852       0.626/0.639/0.646  62178636
 3      42.510%    1.700/1.718/1.723       0.451/0.481/0.488  46826710
 4      41.490%    1.650/1.665/1.675       0.387/0.404/0.411  39174432
 5      41.215%    1.807/1.821/1.829       0.385/0.400/0.408  38913920
 6      41.038%    1.802/1.821/1.828       0.316/0.333/0.339  32289128
 7      40.784%    1.982/2.000/2.008       0.254/0.266/0.270  25671760
 8      40.707%    1.930/1.942/1.947       0.183/0.196/0.203  19217358
 9      40.696%    2.146/2.156/2.162       0.179/0.196/0.202  19212387

 avg1   43.357%                1.870                   0.459
 tot                         504.987                 123.968
```

insert_string_tpl.h
```diff
@@ -82,20 +82,16 @@ ZLIB_INTERNAL Pos INSERT_STRING(deflate_state *const s, const uint32_t str, uint
...
-        Pos head = s->head[hm];
-        if (head != idx) {
+        head = s->head[hm];
+        if (LIKELY(head != idx)) {
             s->prev[idx & s->w_mask] = head;
             s->head[hm] = idx;
-            if (strstart == strend)
-                ret = head;
-        } else if (strstart == strend) {
-            ret = idx;
```
mtl1979, line 96: This was originally added to avoid problems when sliding the window, removing it is EBADIDEA.

Dead2:
For a fun comparison, this is a benchmark of develop from last week if. The commits since then, plus this PR totals at ~0.84% faster, and zlib-ng.so is also getting smaller, so things are moving nicely in the right direction.
## Old Baseline 8832d7db7241194fa68509c96c092f3cf527ccce
```
  93335    1272      32   94639   171af libz-ng.so.1

 1      57.870%    1.863/1.880/1.891       1.181/1.215/1.234  163918126
 2      43.903%    1.825/1.854/1.863       0.614/0.631/0.640  62178636
 3      42.510%    1.700/1.724/1.734       0.460/0.476/0.483  46826710
 4      41.490%    1.681/1.699/1.707       0.387/0.399/0.407  39174432
 5      41.215%    1.825/1.849/1.857       0.372/0.396/0.405  38913920
 6      41.038%    1.834/1.845/1.851       0.323/0.332/0.337  32289128
 7      40.784%    1.998/2.014/2.021       0.253/0.261/0.269  25671760
 8      40.707%    1.935/1.949/1.955       0.185/0.194/0.199  19217358
 9      40.696%    2.154/2.164/2.170       0.185/0.195/0.201  19212387

 avg1   43.357%                1.887                   0.456
 tot                         509.364                 122.992
```

insert_string_tpl.h
Dead2: What part of it? And what commit are you commenting on, the first or the second?

insert_string_tpl.h
```diff
@@ -82,20 +82,16 @@ ZLIB_INTERNAL Pos INSERT_STRING(deflate_state *const s, const uint32_t str, uint
...
         }
     }
-    return ret;
+
+    return head;
```
mtl1979: ```head``` should not be returned if there was no updating of the hash tables as it can be garbage in case of recent window slide.

Dead2: 1. If count >= 1 (it always is), the for loop will set head.
2. strstart is always == strend after the for loop has completed, since we never exit it until they are equal.
3. If head != idx, we return head, if head == idx, we return idx (might as well return head since they are the same)

The only corner-case I see is with count==0, but that that is already handled specially above.

I don't see how this can currently return 0, can you give an example of that?

Dead2: Sorry, fixed typo, had = instead of >= in point 1 of my previous comment..

mtl1979: In the for loop, s->head[hm] can be garbage as when the window is slided, the hash table is not cleared, so anything that is still in window is no longer valid. It would be really slow to clear whole hash table and recalculate all chains, so on first call to insert_string() the old value was discarded. On subsequent calls there is no more garbage.

If we don't ignore the initial contents after window slide, we can get future matches in for example deflate_medium and that will cause assertion failure in debug build.

Dead2: But how does that actually happen in this function? Please walk me through it, because I don't see how.
Has this been lost and is already a bug?

mtl1979: As far as I understand it, after window is slided, the next call to insert_string is from "next match", not "current match" which should be at the start of the window... When it tries to compare the matches, it gets zero offset, because the previous call to insert_match before the slide was also from the "next match". As such the current and previous match become same offset on next check of same hash chain.

If we set the previous offset as 0 (return value of insert_string) when the new and old values match inside insert_string, we actually truncate the hash chain as 0 means end of chain. This doesn't cause assertion failure as the "previous" value in the hash chain is now correct as it was not replaced (the previous value was slided in fill_window, but is guaranteed to be valid).

Dead2: I understand how this would be beneficial, but I cannot see how the current code will ever return `0`. As far as I can see, it always returns either `head` or `idx` (same as head), or with count=0 it returns `s->prev[str & s->w_mask];`.

So my question is how do you trace through this function in a way that `ret` ends up being `0`? IE: How does this function detect that the windows has been slided and a `0` should be returned?

And if it currently cannot return `0` (as it looks like to me), is this actually a hidden bug?

mtl1979: I haven't followed all the pull requests that you have merged in, so there might be a previous regression that got unnoticed.

insert_string_tpl.h
```diff
@@ -61,7 +61,7 @@ ZLIB_INTERNAL Pos QUICK_INSERT_STRING(deflate_state *const s, const uint32_t str
  *    (except for the last MIN_MATCH-1 bytes of the input file).
  */
 ZLIB_INTERNAL Pos INSERT_STRING(deflate_state *const s, const uint32_t str, uint32_t count) {
-    Pos idx, ret;
+    Pos head, idx, ret=0;
```
nmoinvaz: Consider whitespace between operator ie `ret = 0`

insert_string_tpl.h
```diff
@@ -61,7 +61,7 @@ ZLIB_INTERNAL Pos QUICK_INSERT_STRING(deflate_state *const s, const uint32_t str
  *    (except for the last MIN_MATCH-1 bytes of the input file).
  */
 ZLIB_INTERNAL Pos INSERT_STRING(deflate_state *const s, const uint32_t str, uint32_t count) {
-    Pos head, idx, ret=0;
+    Pos head=0, idx;
```
nmoinvaz: Whitespace between operator?

insert_string_tpl.h
nmoinvaz: @mtl1979 I am not following your explanation very well. I thought we don't use hash-chains aka rolling hash because we are only ever hashing 4 bytes at a time.

nmoinvaz (commented):
Breaking it into two commits made it easier for me to review. Thanks for that. The logic changes made sense to me. Only have minor formatting suggestion.

insert_string_tpl.h
mtl1979: I'm not talking about linear hashing chain, I'm talking about how we store indexes that have same hash value. That is essential when we want to know if two indexes have same hash value or not... If we store same index for both the "latest" and "previous" hash tables, then we are basically wasting the "previous" hash table.

When the hash tables are initialized, "previous" hash table should be all zeroes, until the "latest" hash table is populated, when the "old" value of "latest" will be moved to "previous" hash table.... as on first iteration (before first window slide) the old value is 0, there is no problem.

nmoinvaz: That is a great explanation but I'm still failing to see how it applies to this code change.

Last time I looked, the return value from _insert_string_ is not even used by anything.

mtl1979: So we should change the function signature so we don't return anything?

insert_string_tpl.h
Dead2: Fixed for `head`

insert_string_tpl.h
Dead2: Since `ret=0` gets removed in the next commit and is a purely cosmetic change, I hope it is ok if I leave that as-is. 😄

insert_string_tpl.h
Dead2: Looks like we could do that, since the return value is only ever used when calling quick_insert_string.
Doing some tests now to see whether it has any effect on benchmarks, it does reduce compiled text size by 64, so it should help a bit.

mtl1979: I think we have split fixing the function to enough separate pull requests, so we should do one final cleanup on this function and then we don't touch it before first release...

Dead2:
Removing the return value did nothing for the speed in my benchmarks, but as usual that could just be cache-alignments changing and messing up.

I decided to do a combo test with the other insert_string optimization PR though, and this is the combined result:
## PR #724 + #725
```

 1      57.870%    1.829/1.862/1.876       1.176/1.221/1.238  163918126
 2      43.903%    1.803/1.835/1.845       0.608/0.640/0.652  62178636
 3      42.510%    1.693/1.709/1.716       0.463/0.482/0.493  46826710
 4      41.490%    1.648/1.668/1.676       0.389/0.404/0.412  39174432
 5      41.215%    1.801/1.817/1.826       0.377/0.400/0.410  38913920
 6      41.038%    1.800/1.818/1.827       0.321/0.335/0.341  32289128
 7      40.784%    1.985/2.004/2.012       0.248/0.263/0.272  25671760
 8      40.707%    1.930/1.943/1.948       0.190/0.198/0.203  19217358
 9      40.696%    2.143/2.157/2.164       0.186/0.195/0.200  19212387

 avg1   43.357%                1.868                   0.460
 tot                         504.378                 124.157
```
Good reduction of code size and overall speed improvement of close to 0.53%.
Most of the speedup is found in deflate_fast and deflate_medium (levels 2-6).

insert_string_tpl.h
Dead2: The nice thing now is quick_insert_string and insert_string are very close to identical and much cleaner.

Dead2:
Rebased on top of current #develop (with #724).
I would appreciate some quick re-reviews to make sure I didn't mess anything new up.

nmoinvaz (approved):
Looks good

mtl1979:
Looks better now...

PR #730, opened 2020-08-20
MinGW cross-compile toolchain files for cmake under Linux

mtl1979:
These toolchain files set minimal variables for successful compile on Ubuntu, tested on 18.04.4 ...
Does not yet have full support for running tests, but most tests pass if Ubuntu is running inside WSL.


Dead2:
Looks good to me

nmoinvaz:
Maybe in a separate PR we can add it to CI.

mtl1979:
I'm not sure what would be the best way to make the tests run on clean Linux without WSL... maybe add support for wine or something... Not really my area of expertise...

Issue #729, opened 2020-08-20
Cannot cross compile to Windows


mati865:
Tested in Ubuntu 18.04 and 20.04 containers.
That's caused by https://github.com/zlib-ng/zlib-ng/blob/1a46fb9ee1b6926eb8635d32d7f7c1c023a79625/cmake/detect-arch.cmake#L20 which according to [official docs](https://cmake.org/cmake/help/v3.18/variable/CMAKE_LANG_COMPILER_TARGET.html) is not generally set:
> This variable may only be set in a toolchain file specified by the CMAKE_TOOLCHAIN_FILE variable.

<details>
<summary>The error:</summary>

```
$ cmake -DCMAKE_SYSTEM_NAME="Windows" -DCMAKE_C_COMPILER="i686-w64-mingw32-gcc" ..
-- Using CMake version 3.16.3
-- ZLIB_HEADER_VERSION: 1.2.11
-- ZLIBNG_HEADER_VERSION: 1.9.9
-- The C compiler identification is GNU 9.3.0
-- Check for working C compiler: /usr/bin/i686-w64-mingw32-gcc
-- Check for working C compiler: /usr/bin/i686-w64-mingw32-gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Arch not recognized, falling back to cmake arch: ''
-- Basearch '' not recognized, defaulting to 'x86'.
-- Basearch of '' has been detected as: 'x86'
-- Architecture supports unaligned reads
-- Architecture supports unaligned reads of > 4 bytes
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for stdarg.h
-- Looking for stdarg.h - found
-- Looking for stdint.h
-- Looking for stdint.h - found
-- Looking for stddef.h
-- Looking for stddef.h - found
-- Looking for sys/sdt.h
-- Looking for sys/sdt.h - not found
-- Looking for unistd.h
-- Looking for unistd.h - found
-- Check size of off64_t
-- Check size of off64_t - done
-- Looking for fseeko
-- Looking for fseeko - found
-- Looking for strerror
-- Looking for strerror - found
-- Performing Test HAVE_ATTRIBUTE_VISIBILITY_HIDDEN
-- Performing Test HAVE_ATTRIBUTE_VISIBILITY_HIDDEN - Failed
[...]

PR #732, opened 2020-08-20
Fix unquoted variables in cmake if statements

Dead2:
This was exposed by #729, but does not fix the real problems there.


mtl1979:
LGTM.

nmoinvaz (approved):

PR #733, opened 2020-08-21
Add optional support for thread local storage.

nmoinvaz:
It is possible to predefine `ZTHREAD` to disable thread local storage.
```
cmake . -DCMAKE_C_FLAGS=-DZTHREAD=
```

I used `ZTHREAD` to coincide with `ZEXPORT` and `ZEXTERN`.

See #570 for previous discussion.

### DIFF
--- a/functable.c
+++ b/functable.c
@@ -126,7 +126,7 @@ extern int32_t longest_match_unaligned_avx2(deflate_state *const s, Pos cur_matc
 #endif
 #endif
 
-ZLIB_INTERNAL __thread struct functable_s functable;
+ZLIB_INTERNAL Z_TLS struct functable_s functable;
 
 ZLIB_INTERNAL void cpu_check_features(void)
 {
@@ -416,7 +416,7 @@ ZLIB_INTERNAL int32_t longest_match_stub(deflate_state *const s, Pos cur_match) 
 }
 
 /* functable init */
-ZLIB_INTERNAL __thread struct functable_s functable = {
+ZLIB_INTERNAL Z_TLS struct functable_s functable = {
     insert_string_stub,
     quick_insert_string_stub,
     adler32_stub,
--- a/functable.h
+++ b/functable.h
@@ -24,6 +24,6 @@ struct functable_s {
     uint8_t* (* chunkmemset_safe)   (uint8_t *out, unsigned dist, unsigned len, unsigned left);
 };
 
-ZLIB_INTERNAL extern __thread struct functable_s functable;
+ZLIB_INTERNAL extern Z_TLS struct functable_s functable;
 
 #endif
--- a/zbuild.h
+++ b/zbuild.h
@@ -8,7 +8,6 @@
 #  else
     typedef long ssize_t;
 #  endif
-#  define __thread __declspec(thread)
 #endif
 
 #if defined(ZLIB_COMPAT)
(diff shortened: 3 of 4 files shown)


nmoinvaz:
Do we want it named `ZTHREAD` or `Z_THREAD`. I mean there are a bunch of other #define's in zlib headers that use `Z_*` - more do than don't. Perhaps on another PR we can synchronize everything.

gdevenyi:
Should this be a cmake advanced feature rather than a explicit define?

nmoinvaz:
I thought about that but I don't want to keep stuffing CMake with all these advanced options if they won't hardly be used. I already made that mistake with the advanced inflate options to be honest.

mtl1979:
ZTHREAD is OK for me... I agree with @nmoinvaz That very rarely used optional features, especially ones that are enabled by default, don't warrant using cmake's advanced features... On the other hand, if we ever want to reverse so that TLS support is disabled by default, then we might want to use something like USE_TLS, so different keywords for different compilers are automatically used correctly.

Dead2:
I'd kind of like to see this disabled by default, so if someone wants it enabled they will have to explicitly do that.
zlib/zlib-ng is not supposed to be thread safe after all, so even having this makes little sense to me.
If someone can come up with a (realistic) regression test that fails with this disabled, I'd like to see that of course.

Only legitimate use-case I can think of right now is if someone wanted to do multi-threaded hashing using zlib-ng, but is that really something we should support by default? IIRC inflate and deflate still won't be able to work multi-threaded.

Or do you disagree?

If we keep it enabled, I would have liked to have a regression test for it in CI, since we don't really have any idea whether it works or not.

mtl1979:
Multi-threaded hashing would make sense with image manipulation... Things like Qt use libpng extensively and is multi-threaded by design...

Dead2:
@mtl1979 Yes, but does it fail with zlib-ng without TLS? If it does, then yes, that would be a regression and we should keep it.
If it does not, then I don't see the point of introducing it.

mtl1979:
@Dead2 It's not so much about failing, but making sure it always calls the right variation... There is small moment when it is possible that it calls non-optimized version if previous thread hasn't returned from feature detection yet.

Dead2:
Yeah, my question is about whether that is actually our problem or should the program calling zlib/zlib-ng make sure it gets it right? Can we say we are partially thread-safe and get away with not dealing with any more of that, or does that open the door to 'we need completely-thread-safe' (explicitly or implied).

mtl1979:
@Dead2 My personal opinion is that partially thread-safe is enough... Most likely completely thread-safe deflate/inflate is wishful thinking and not realistic...

Dead2:
Maybe rename it to Z_TLS?
IMO I think we should move towards the Z_ prefix. And TLS makes more sense than THREAD in this case, also THREAD is bound to make someone think that we support threads 😛

nmoinvaz:
I have renamed it to `Z_TLS`. And @Dead2 you said you want it disabled by default, so I have done that. If somebody wants it they can use:
```
cmake . -DCMAKE_C_FLAGS=-DZ_TLS=__thread
```

Dead2:
I see no changes in output, no change to performance, but compiled code size goes down a lot, and data goes down as well.

## Baseline a19bbdeec892b1ff8210bd8c3b679cca77ad042d
```
```
## PR #733 44fdfcf36f6b4c03e202afc6a3647eba03b8fd79
```
  91255    1232      32   92519   16967 libz-ng.so.1
```

PR #707, opened 2020-07-26
Use unaligned 32-bit and 64-bit compare in longest_match

nmoinvaz:
- Use unaligned 32-bit and 64-bit compare based on best match length when searching for matches.
- Move TRIGGER_LEVEL to match_tpl.h since it is only used in longest match.
- Use early return inside match loops instead of cont variable.
- Added back 2 byte check for platforms that don't supported unaligned access.

PR #734, opened 2020-08-21
Change to static hash-table size

Dead2:
First commit removes hash_bits, hash_size and hash_mask from the deflate struct, and replaces them with defines.
Second commit doubles the size of the hash-table, this often leads to slightly better compression on all levels (sometimes slightly worse on the higher levels) and gives a very nice performance boost (looks to be because we AND the hash by 0xFFFF instead of by 0x7FFF, and that seems to be a lot more cpu-friendly).

I need help analyzing this, since it changes output data we need to be sure it is not exposing any new bugs.
I have not detected any invalid data in my tests, but I also cannot explain the sometimes increased output size.
Double the hash table size should decrease collisions (my understanding of the better compression), but will not make any new ones.
So, is the decreased compression because we detect a longer-distance match that would have otherwise been overwritten by a newer one?

PS: This is not the same as increasing memory-level, since that also affects the lit-buffer, and that cannot be increased beyond 16K entries without adding extra code to prevent invalid data.

See benchmarks in separate comment.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -178,12 +178,10 @@ static const config configuration_table[10] = {
 
 
 /* ===========================================================================
- * Initialize the hash table (avoiding 64K overflow for 16 bit systems).
- * prev[] will be initialized on the fly.
+ * Initialize the hash table. prev[] will be initialized on the fly.
  */
 #define CLEAR_HASH(s) do {                                                                \
-    s->head[s->hash_size - 1] = NIL;                                                      \
-    memset((unsigned char *)s->head, 0, (unsigned)(s->hash_size - 1) * sizeof(*s->head)); \
+    memset((unsigned char *)s->head, 0, HASH_SIZE * sizeof(*s->head)); \
   } while (0)
 
 /* ===========================================================================
@@ -192,11 +190,11 @@ static const config configuration_table[10] = {
  * keep the hash table consistent if we switch back to level > 0 later.
  */
 ZLIB_INTERNAL void slide_hash_c(deflate_state *s) {
-    unsigned n;
     Pos *p;
+    unsigned n;
     unsigned int wsize = s->w_size;
 
-    n = s->hash_size;
+    n = HASH_SIZE;
     p = &s->head[n];
 #ifdef NOT_TWEAK_COMPILER
     do {
@@ -320,16 +318,6 @@ int32_t ZEXPORT PREFIX(deflateInit2_)(PREFIX3(stream) *strm, int32_t level, int3
     s->w_size = 1 << s->w_bits;
     s->w_mask = s->w_size - 1;
 
-#ifdef X86_SSE42_CRC_HASH
-    if (x86_cpu_has_sse42)
-        s->hash_bits = (unsigned int)15;
-    else
-#endif
-        s->hash_bits = (unsigned int)memLevel + 7;
-
-    s->hash_size = 1 << s->hash_bits;
-    s->hash_mask = s->hash_size - 1;
-
 #ifdef X86_PCLMULQDQ_CRC
     window_padding = 8;
 #endif
@@ -337,7 +325,7 @@ int32_t ZEXPORT PREFIX(deflateInit2_)(PREFIX3(stream) *strm, int32_t level, int3
     s->window = (unsigned char *) ZALLOC_WINDOW(strm, s->w_size + window_padding, 2*sizeof(unsigned char));
     s->prev   = (Pos *)  ZALLOC(strm, s->w_size, sizeof(Pos));
     memset(s->prev, 0, s->w_size * sizeof(Pos));
-    s->head   = (Pos *)  ZALLOC(strm, s->hash_size, sizeof(Pos));
+    s->head   = (Pos *)  ZALLOC(strm, HASH_SIZE, sizeof(Pos));
 
     s->high_water = 0;      /* nothing written to s->window yet */
 
@@ -734,7 +722,7 @@ unsigned long ZEXPORT PREFIX(deflateBound)(PREFIX3(stream) *strm, unsigned long 
 
     /* if not default parameters, return conservative bound */
     if (DEFLATE_NEED_CONSERVATIVE_BOUND(strm) ||  /* hook for IBM Z DFLTCC */
-            s->w_bits != 15 || s->hash_bits != 8 + 7)
+            s->w_bits != 15 || HASH_BITS < 15)
         return complen + wraplen;
 
     /* default settings: return tight bound for that case */
@@ -1127,7 +1115,7 @@ int32_t ZEXPORT PREFIX(deflateCopy)(PREFIX3(stream) *dest, PREFIX3(stream) *sour
 
     ds->window = (unsigned char *) ZALLOC_WINDOW(dest, ds->w_size + window_padding, 2*sizeof(unsigned char));
     ds->prev   = (Pos *)  ZALLOC(dest, ds->w_size, sizeof(Pos));
-    ds->head   = (Pos *)  ZALLOC(dest, ds->hash_size, sizeof(Pos));
+    ds->head   = (Pos *)  ZALLOC(dest, HASH_SIZE, sizeof(Pos));
     ds->pending_buf = (unsigned char *) ZALLOC(dest, ds->lit_bufsize, 4);
 
     if (ds->window == NULL || ds->prev == NULL || ds->head == NULL || ds->pending_buf == NULL) {
@@ -1137,7 +1125,7 @@ int32_t ZEXPORT PREFIX(deflateCopy)(PREFIX3(stream) *dest, PREFIX3(stream) *sour
 
     memcpy(ds->window, ss->window, ds->w_size * 2 * sizeof(unsigned char));
     memcpy((void *)ds->prev, (void *)ss->prev, ds->w_size * sizeof(Pos));
-    memcpy((void *)ds->head, (void *)ss->head, ds->hash_size * sizeof(Pos));
+    memcpy((void *)ds->head, (void *)ss->head, HASH_SIZE * sizeof(Pos));
     memcpy(ds->pending_buf, ss->pending_buf, (unsigned int)ds->pending_buf_size);
 
     ds->pending_out = ds->pending_buf + (ss->pending_out - ss->pending_buf);
--- a/deflate.h
+++ b/deflate.h
@@ -68,6 +68,10 @@
 #define FINISH_STATE 666    /* stream complete */
 /* Stream status */
 
+#define HASH_BITS 15        /* log2(HASH_SIZE) */
+#define HASH_SIZE 32768     /* number of elements in hash table */
+#define HASH_MASK 0x7FFF    /* HASH_SIZE-1 */
+
 
 /* Data structure describing a single value and its code string. */
 typedef struct ct_data_s {
@@ -145,10 +149,6 @@ typedef struct internal_state {
      */
 
     Pos *head; /* Heads of the hash chains or NIL. */
-
-    unsigned int  hash_size;         /* number of elements in hash table */
-    unsigned int  hash_bits;         /* log2(hash_size) */
-    unsigned int  hash_mask;         /* hash_size-1 */
 
     long block_start;
     /* Window position at the beginning of the current output block. Gets
--- a/deflate.h
+++ b/deflate.h
@@ -68,9 +68,9 @@
 #define FINISH_STATE 666    /* stream complete */
 /* Stream status */
 
-#define HASH_BITS 15        /* log2(HASH_SIZE) */
-#define HASH_SIZE 32768     /* number of elements in hash table */
-#define HASH_MASK 0x7FFF    /* HASH_SIZE-1 */
+#define HASH_BITS    16u           /* log2(HASH_SIZE) */
+#define HASH_SIZE 65536u           /* number of elements in hash table */
+#define HASH_MASK (HASH_SIZE - 1u) /* HASH_SIZE-1 */
 
 
 /* Data structure describing a single value and its code string. */
--- a/insert_string.c
+++ b/insert_string.c
@@ -15,7 +15,7 @@
  *    previous key instead of complete recalculation each time.
  */
 #define UPDATE_HASH(s, h, val) \
-    h = ((val * 2654435761U) >> (32 - s->hash_bits));
+    h = ((val * 2654435761U) >> (32 - HASH_BITS));
 
 #define INSERT_STRING       insert_string_c
 #define QUICK_INSERT_STRING quick_insert_string_c
--- a/insert_string.c
+++ b/insert_string.c
@@ -14,8 +14,10 @@
  *    input characters, so that a running hash key can be computed from the
  *    previous key instead of complete recalculation each time.
  */
+#define HASH_SLIDE 16  // Number of bits to slide hash
+
 #define UPDATE_HASH(s, h, val) \
-    h = ((val * 2654435761U) >> (32 - HASH_BITS));
+    h = ((val * 2654435761U) >> HASH_SLIDE);
 
 #define INSERT_STRING       insert_string_c
 #define QUICK_INSERT_STRING quick_insert_string_c
--- a/match_tpl.h
+++ b/match_tpl.h
@@ -41,10 +41,9 @@ ZLIB_INTERNAL int32_t LONGEST_MATCH(deflate_state *const s, Pos cur_match) {
     bestcmp_t scan_end, scan_start;
 
     /*
-     * The code is optimized for HASH_BITS >= 8 and MAX_MATCH-2 multiple
-     * of 16. It is easy to get rid of this optimization if necessary.
+     * The code is optimized for MAX_MATCH-2 multiple of 16.
      */
-    Assert(s->hash_bits >= 8 && MAX_MATCH == 258, "Code too clever");
+    Assert(MAX_MATCH == 258, "Code too clever");
 
     /*
      * Do not waste too much time if we already have a good match
--- a/arch/arm/slide_neon.c
+++ b/arch/arm/slide_neon.c
@@ -46,7 +46,7 @@ static inline void slide_hash_chain(Pos *table, unsigned int entries, uint16_t w
 ZLIB_INTERNAL void slide_hash_neon(deflate_state *s) {
     unsigned int wsize = s->w_size;
 
-    slide_hash_chain(s->head, s->hash_size, wsize);
+    slide_hash_chain(s->head, HASH_SIZE, wsize);
     slide_hash_chain(s->prev, wsize, wsize);
 }
 #endif
--- a/arch/power/slide_hash_power8.c
+++ b/arch/power/slide_hash_power8.c
@@ -48,7 +48,7 @@ void ZLIB_INTERNAL slide_hash_power8(deflate_state *s) {
     unsigned int n;
     Pos *p;
 
-    n = s->hash_size;
+    n = HASH_SIZE;
     p = &s->head[n];
     slide_hash_power8_loop(s,n,p);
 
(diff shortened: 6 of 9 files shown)


Dead2:
## Deflatebench config
activated multiple generated file mode. Source: silesia-small.tar
Level 1: /tmp/deflatebench-1.tmp 210.1 MiB / 220,308,480 B
Level 2: /tmp/deflatebench-2.tmp 105.1 MiB / 110,154,240 B
Level 3: /tmp/deflatebench-3.tmp 90.0 MiB / 94,417,920 B
Level 4: /tmp/deflatebench-4.tmp 75.0 MiB / 78,681,600 B
Level 5: /tmp/deflatebench-5.tmp 75.0 MiB / 78,681,600 B
Level 6: /tmp/deflatebench-6.tmp 60.0 MiB / 62,945,280 B
Level 7: /tmp/deflatebench-7.tmp 45.0 MiB / 47,208,960 B
Level 8: /tmp/deflatebench-8.tmp 30.0 MiB / 31,472,640 B
Level 9: /tmp/deflatebench-9.tmp 30.0 MiB / 31,472,640 B

 Runs: 70
 Levels: 1-9
 Trimworst: 40

## Baseline 0c8b66925af3ad3c07112b32c8e7eb7f80b29ff4
```

 1      57.869%    1.421/1.451/1.462       0.921/0.943/0.957  127491314
 2      43.903%    1.395/1.425/1.433       0.483/0.496/0.503  48360972
 3      42.509%    1.452/1.466/1.473       0.387/0.413/0.423  40136545
 4      41.492%    1.364/1.386/1.394       0.317/0.332/0.340  32646413
 5      41.214%    1.500/1.515/1.521       0.309/0.334/0.342  32427914
 6      41.038%    1.440/1.457/1.463       0.253/0.263/0.270  25831228
 7      40.784%    1.489/1.502/1.508       0.187/0.198/0.204  19253811
 8      40.706%    1.285/1.294/1.299       0.116/0.130/0.135  12811154
 9      40.695%    1.426/1.439/1.444       0.122/0.131/0.136  12807943

 avg1   43.357%                1.437                   0.360
 tot                         388.051                  97.226
```

## PR #734 81987abb20e3f7c7e6aa59ce4dd921801a05b582
```
  92863    1272      32   94167   16fd7 libz-ng.so.1

 1      57.781%    1.395/1.435/1.450       0.902/0.953/0.966  127296397
 2      43.877%    1.351/1.367/1.374       0.472/0.494/0.502  48332706
 3      42.509%    1.368/1.402/1.412       0.405/0.414/0.421  40136072
 4      41.488%    1.356/1.373/1.381       0.296/0.334/0.343  32643562
 5      41.212%    1.482/1.496/1.504       0.311/0.331/0.338  32425995
 6      41.037%    1.425/1.436/1.443       0.252/0.268/0.272  25831160
 7      40.784%    1.443/1.451/1.458       0.183/0.196/0.203  19253737
 8      40.706%    1.245/1.256/1.261       0.116/0.129/0.135  12811166
 9      40.695%    1.379/1.401/1.406       0.118/0.129/0.135  12807941

 avg1   43.343%                1.402                   0.361
 tot                         378.542                  97.422
```

So we can see here a whopping 2.45% faster compression and a 0.31% reduction in compiled code size.

deflate.c
```diff
@@ -182,8 +182,8 @@ static const config configuration_table[10] = {
  * prev[] will be initialized on the fly.
  */
 #define CLEAR_HASH(s) do {                                                                \
-    s->head[s->hash_size - 1] = NIL;                                                      \
-    memset((unsigned char *)s->head, 0, (unsigned)(s->hash_size - 1) * sizeof(*s->head)); \
+    s->head[HASH_SIZE - 1] = NIL;                                                      \
+    memset((unsigned char *)s->head, 0, (unsigned)(HASH_SIZE - 1) * sizeof(*s->head)); \
```
mtl1979: This seems to be workaround for platforms where "unsigned int" is only 16 bits... Maybe we should clean this up as we only support platforms where it is 32 bits.

Dead2: Yeah, I was thinking the same thing, but I was a little uncertain what would be the correct replacement, and wanted to avoid any extra complicating changes.
Would removing the `= NIL` line and remove the `- 1`in memset to do memset on the whole `HASH_SIZE` be the correct thing to do?

Doing so decreases text size by another 48 bytes, and it does not change the compressed output size.

mtl1979: UINT32_C(HASH_SIZE) should be best as it avoids the cast completely and just changes the constant type to unsigned... This makes it possible for memset to use largest possible vector size when vector instructions are allowed.

Dead2: I think that only works on numbers. Might try it on the define directly though.

mtl1979: It should depend on evaluation order of the preprocessor... That's why I wrote "should" instead of "will"...

Dead2: I my case `UINT32_C(HASH_SIZE)` evaluates to `HASH_SIZEU` 😄. I might just add an `u` to both the new numeric defines.

mtl1979: Now we know the evaluation order ;)

insert_string.c
```diff
@@ -15,7 +15,7 @@
  *    previous key instead of complete recalculation each time.
  */
 #define UPDATE_HASH(s, h, val) \
-    h = ((val * 2654435761U) >> (32 - s->hash_bits));
+    h = ((val * 2654435761U) >> (32 - HASH_BITS));
```
nmoinvaz: The thing I worry about is that previously `s->hash_bits` was always 15, but now it will be 16. When I had done my testing, 15 produced the best results.

deflate.c
```diff
@@ -734,7 +724,7 @@ unsigned long ZEXPORT PREFIX(deflateBound)(PREFIX3(stream) *strm, unsigned long
 
     /* if not default parameters, return conservative bound */
     if (DEFLATE_NEED_CONSERVATIVE_BOUND(strm) ||  /* hook for IBM Z DFLTCC */
-            s->w_bits != 15 || s->hash_bits != 8 + 7)
+            s->w_bits != 15 || HASH_BITS != 15)
```
nmoinvaz: This will always be hit because `HASH_BITS` = 16 != 15.

insert_string.c
mtl1979: Even changing to static 15 bits should improve performance as compiler can optimize code better when it knows the parameter can be evaluated at compile time.

deflate.c
Dead2: Hmm, maybe `HASH_BITS < 15` would be better.

insert_string.c
Dead2: Actually it was usually 15 on x86, but on other platforms it respected the memory-level setting, so it could be much lower than that too.

Also, using a static 15 did not improve performance measurably. 16 means shift by two whole bytes, so I assume that is a lot easier for the cpu to do than shifting by odd non-byte-aligned lengths.

nmoinvaz: @Dead2 are you testing using that hash function with all the other optimizations turned off?

nmoinvaz (commented):
Overall I like the change. It simplifies code in multiple areas.

deflate.c
Dead2: Fixed.

deflate.c
Dead2: Fixed.

insert_string.c
Dead2: I'll do some tests where I change only this instance to 15 and 16 to compare the results. Perhaps not using HASH_BITS is better here.

nmoinvaz: I would also be okay with hard coding it to 15 in this instance.

deflate.h
```diff
@@ -68,6 +68,10 @@
 #define FINISH_STATE 666    /* stream complete */
 /* Stream status */
 
+#define HASH_BITS    16u    /* log2(HASH_SIZE) */
+#define HASH_SIZE 65536u    /* number of elements in hash table */
+#define HASH_MASK 0xFFFF    /* HASH_SIZE-1 */
```
mtl1979: 0xFFFF is essentially same as 65535u as hexadecimal constants are always unsigned. For consistency, maybe this also should be decimal.

match_tpl.h
```diff
@@ -44,7 +44,7 @@ ZLIB_INTERNAL int32_t LONGEST_MATCH(deflate_state *const s, Pos cur_match) {
      * The code is optimized for HASH_BITS >= 8 and MAX_MATCH-2 multiple
      * of 16. It is easy to get rid of this optimization if necessary.
      */
-    Assert(s->hash_bits >= 8 && MAX_MATCH == 258, "Code too clever");
+    Assert(HASH_BITS >= 8 && MAX_MATCH == 258, "Code too clever");
```
mtl1979: The first test is now always true.

mtl1979 (changes requested):

insert_string.c
Dead2: ## (32 - 16)
```
 1      43.877%    3.427/3.434/3.438       1.005/1.023/1.029  96665297
 2      43.129%    1.788/1.792/1.799       0.492/0.495/0.502  47508837
 3      42.509%    1.711/1.726/1.734       0.412/0.424/0.428  40136252
 4      41.962%    1.540/1.546/1.553       0.338/0.347/0.358  33016216
 5      41.114%    1.962/1.965/1.968       0.334/0.346/0.354  32349446
 6      40.847%    1.948/1.953/1.957       0.263/0.272/0.279  25711547
 7      40.784%    1.624/1.633/1.639       0.203/0.208/0.213  19253843
 8      40.706%    1.375/1.378/1.381       0.130/0.135/0.140  12811177
 9      40.695%    1.525/1.529/1.535       0.126/0.135/0.143  12807944

 avg1   41.736%                1.884                   0.376
 tot                          84.773                  16.920
```
## (32 - 15)
```
 1      43.891%    3.586/3.598/3.605       1.012/1.021/1.029  96696009
 2      43.135%    1.867/1.878/1.886       0.493/0.502/0.509  47515070
 3      42.510%    1.792/1.804/1.807       0.422/0.426/0.429  40136867
 4      41.968%    1.585/1.598/1.603       0.348/0.351/0.354  33021210
 5      41.119%    2.035/2.040/2.042       0.337/0.344/0.347  32353240
 6      40.848%    2.012/2.017/2.020       0.260/0.270/0.276  25711774
 7      40.784%    1.678/1.685/1.691       0.199/0.203/0.206  19253840
 8      40.706%    1.418/1.420/1.422       0.129/0.135/0.139  12811158
 9      40.695%    1.569/1.570/1.572       0.126/0.135/0.139  12807944

 avg1   41.740%                1.957                   0.376
 tot                          88.043                  16.931
```

In my tests 16 is always faster, and almost always results in smaller output.
16 ends up taking 3.7% less time than 15 in these tests, that is quite significant IMO.

match_tpl.h
Dead2: Wasn't it always true before as well?  s->hash_bits = memory_level+7, that should equate to a range of 8 to 15.
And MAX_MATCH is also always 258.

I thought we might as well keep the assert in case someone decides to mess with the defined values for whatever reason, I guess it'd be optimized out by the compiler anyways.

deflate.h
Dead2: Isn't it more normal to represent masks in hex though?

mtl1979: The difference is that when you specify it in decimal, the storage width is explicit...

insert_string.c
Dead2: ## (32 - 8)
8 is `s->hash_bits` with lowest possible `memory_level` selected
```
 1      50.134%    3.976/3.983/3.993       1.059/1.069/1.082  110450356
 2      48.086%    2.310/2.323/2.333       0.488/0.520/0.538  52968586
 3      44.115%    4.362/4.371/4.381       0.419/0.428/0.435  41652257
 4      46.038%    2.379/2.392/2.408       0.349/0.361/0.369  36223316
 5      43.682%    4.191/4.194/4.196       0.355/0.357/0.362  34369631
 6      41.058% 12.843/12.852/12.859       0.274/0.281/0.287  25844010
 7      40.827% 11.023/11.032/11.038       0.196/0.203/0.209  19273855
 8      40.708%    8.216/8.225/8.232       0.136/0.138/0.140  12811785
 9      40.696%    8.415/8.420/8.424       0.119/0.129/0.133  12807979

 avg1   43.927%                6.421                   0.387
 tot                         288.960                  17.433
```
Compression did not hurt as much as I had expected, but wow look at that performance dive!

Dead2: I'd be fine with both 15 and 16, but at least in my tests 16 looks like a clear winner.
The compression advantage might be data-dependant though, but the performance advantage should be universal.

deflate.h
Dead2: I am a bit uncertain what you mean.
Is this what you mean?
`#define HASH_MASK 65535u`

At least for me that seems less clear, and I do not see what is the actual benefit from it.

mtl1979: You can also write it as ```(HASH_SIZE - 1u)``` and it will be more clear and still valid, compiler is smart enough to convert it to 65535u. 0xFFFF without width specifier is same as UINT16_C(65535) as it is smallest integer type that can hold the value.

```- 1u``` works and does not trigger error as 65536 doesn't fit in uint16_t and as such is stored as uint32_t. When two types are different size in expression, left hand side size wins.

File: match_tpl.h (hunk 5 above)
Comment (nmoinvaz) [reply match_tpl.h]: I thought we might want to take it out as well. The longest match function doesn't matter much with regard to hash bits. It is only the hash function. And we support multiple hash functions. I don't know if it really matters what the hash bits are in this function and why the code would be too clever if it was changed.

The second MAX_MATCH condition I understand why it is there.

Comment (mtl1979) [reply match_tpl.h]: I think it might be historic relic from old code and was not changed when the code was modified so the s->hash_bits is always at least 8.

File: deflate.h (hunk 4 above)
Comment (Dead2) [reply deflate.h]: Done.

File: match_tpl.h (hunk 5 above)
Comment (Dead2) [reply match_tpl.h]: So, remove the whole assert and comment, or only the HASH_BITS part?

Comment (mtl1979) [reply match_tpl.h]: Only the HASH_BITS part...

File: insert_string.c (hunk 2 above)
Comment (Dead2) [reply insert_string.c]: Do we want `>> (32 - 16)` or just `>> 16`? (alternatively `>> 17` if -15).
I don't see any real good reason to base the shift on 32, is it somehow more helpful that way when _you_ read it?

Comment (mtl1979) [reply insert_string.c]: That would pretty much require fourth opinion then... lol...

File: match_tpl.h (hunk 5 above)
Comment (Dead2) [reply match_tpl.h]: Done.

File: match_tpl.h [hunk 6]
```diff
@@ -41,10 +41,10 @@ ZLIB_INTERNAL int32_t LONGEST_MATCH(deflate_state *const s, Pos cur_match) {
     bestcmp_t scan_end, scan_start;
 
     /*
-     * The code is optimized for HASH_BITS >= 8 and MAX_MATCH-2 multiple
-     * of 16. It is easy to get rid of this optimization if necessary.
+     * The code is optimized for MAX_MATCH-2 multiple of 16.
+     * It is easy to get rid of this optimization if necessary.
```
mtl1979: I would vote the line containing "easy to get rid of" to be removed also... We need to check the first few bytes separately...

insert_string.c
Dead2: Well, myself I see no reason for the 32 to be there anymore, it did when it was based on `s->hash_bits` though, so I am thinking it can now be removed.
I think it makes more sense to just put how many bits you actually shift the value by.

mtl1979: I think the rationale is same as with using (65536 - 1) instead of 65535... Basically you are counting backwards because you can't precalculate with variables or expressions containing preprocessor tokens.

nmoinvaz: I think if were are going to go with 16 then I would just leave it the way it is: `(32 - HASH_BITS)`.  Performance tests look good.

Dead2: Well, I am thinking this really has nothing to do with HASH_BITS, does it?
And if someone changes HASH_BITS, this also changes. As you saw from my (32-8) benchmark above, I don't think that is is a good idea.

If I understand correctly, this in a way does the opposite of the HASH_MASK that comes right after. The extreme would be if HASH_BITS was 32, then we would be shifting the whole value away, but HASH_MASK would keep the whole string of zeroes. Shifting 16 bytes and then having HASH_MASK=16 means we keep everything that is left after the shift.
If both were 17, then we are left with 15bits of data, but we keep 17 bits. These examples are extremes, but they help illustrate what I mean about them doing completely opposite things. This goes for the lower bit shifts as well, if 15 was optimal but we shift it less, and keep less, then the optimal part was never shifted down to the part we wanted to keep.

So I'd vote to have this be hardcoded to `>> 16` or `>> 15` directly (or a local define), as that tells you exactly what it does and it doesn't get affected by tinkerers.
And I feel 16 is the way to go since it gives a huge speed boost for the legacy-c code, even if it gives worse compression in corner cases. (I haven't found any myself).

match_tpl.h
Dead2: Removed

nmoinvaz (approved):
Nice!

Dead2:
Okay, updated benchmarks

## Baseline a19bbdeec892b1ff8210bd8c3b679cca77ad042d
```

 1      57.869%    1.430/1.451/1.464       0.934/0.952/0.961  127491314
 2      43.903%    1.404/1.427/1.435       0.477/0.499/0.507  48360972
 3      42.509%    1.449/1.464/1.472       0.396/0.414/0.423  40136545
 4      41.492%    1.376/1.388/1.395       0.317/0.336/0.344  32646413
 5      41.214%    1.495/1.514/1.520       0.318/0.335/0.345  32427914
 6      41.038%    1.433/1.457/1.463       0.260/0.267/0.273  25831228
 7      40.784%    1.490/1.502/1.507       0.184/0.198/0.204  19253811
 8      40.706%    1.282/1.293/1.298       0.116/0.129/0.135  12811154
 9      40.695%    1.426/1.437/1.442       0.119/0.130/0.136  12807943

 avg1   43.357%                1.437                   0.362
 tot                         388.004                  97.793
```
## PR #734 833287176486ef095a20ca52a8c5e15a5be0bfef
```
  92903    1272      32   94207   16fff libz-ng.so.1

 1      57.781%    1.411/1.429/1.436       0.914/0.948/0.962  127296397
 2      43.877%    1.342/1.362/1.371       0.478/0.497/0.504  48332706
 3      42.509%    1.388/1.403/1.412       0.397/0.413/0.420  40136072
 4      41.488%    1.342/1.374/1.381       0.319/0.333/0.339  32643562
 5      41.212%    1.479/1.499/1.508       0.317/0.331/0.337  32425995
 6      41.037%    1.424/1.438/1.445       0.248/0.265/0.271  25831160
 7      40.784%    1.441/1.450/1.456       0.183/0.197/0.202  19253737
 8      40.706%    1.247/1.255/1.259       0.121/0.131/0.135  12811166
 9      40.695%    1.391/1.400/1.405       0.121/0.130/0.135  12807941

 avg1   43.343%                1.401                   0.361
 tot                         378.308                  97.395
```
~2.5% faster and little reduction in compressed size.


## PR #707 + #734
This one is the combination of this PR and PR #707
```
  94263    1272      32   95567   1754f libz-ng.so.1

 1      57.781%    1.402/1.424/1.438       0.906/0.944/0.959  127296397
 2      43.874%    1.350/1.370/1.379       0.470/0.494/0.501  48329132
 3      42.490%    1.408/1.421/1.429       0.397/0.411/0.420  40118068
 4      41.472%    1.373/1.400/1.409       0.319/0.334/0.343  32630864
 5      41.212%    1.496/1.505/1.513       0.320/0.336/0.343  32425995
 6      41.037%    1.403/1.423/1.432       0.246/0.265/0.272  25831160
 7      40.784%    1.372/1.386/1.391       0.183/0.195/0.202  19253737
 8      40.706%    1.176/1.187/1.193       0.118/0.129/0.133  12811166
 9      40.695%    1.309/1.319/1.324       0.119/0.129/0.135  12807941

 avg1   43.339%                1.382                   0.360
 tot                         373.077                  97.095
```
Totals to ~3.75% faster on average, and little extra improvement in compressed size.

PR #735, opened 2020-08-23
Rename Z defines for consistency

nmoinvaz:


nmoinvaz:
Rebased

Dead2:
Does this cause problems with stock `zlib.h` compatibility in any way?

I also wonder whether we should be using the WINAPIV call type for all functions in zlib-ng native. I suspect the WINAPI version was chosen at an early point before WINAPIV existed (or was needed), and they couldn't break that compatibility once chosen.

Using the same call method everywhere might be a good thing IMO, should we change the default for our native builds?

mtl1979:
@Dead2 Only benefit for using WINAPIV is that it supports variadic parameters, otherwise the two are the same feature-wise.

nmoinvaz:
Rebased.

nmoinvaz:
@Dead2 I have kept some few references to `ZLIB_INTERNAL` for zlib.h compatibility.

PR #737, opened 2020-08-25
Rename headers.

nmoinvaz:


CMakeLists.txt
```diff
@@ -822,20 +822,20 @@ set(ZLIB_PUBLIC_HDRS
...
+    inffixed_tbl.h
     inflate.h
     inflate_p.h
     inftrees.h
     insert_string_tpl.h
     match_tpl.h
-    trees.h
+    trees_tbl.h
     trees_emit.h
     trees_p.h
```
Dead2: Maybe rename `trees_p.h` to `trees.h` too? It was named that because the `trees.h` name was already taken.

mtl1979: Point of having _p suffix in header basename is that it discourages external code to use it when including zlib-ng as subproject.

mtl1979: FYI These filenames are not in alphabetical order anymore...

Dead2: trees_p.h contains the bits that are not private to trees.c, and it needs to be included from the tool maketrees.c currently.
So the name is in my opinion wrong and breaks with how we treat the _p file naming scheme in the rest of the codebase.

Also if we wanted to discourage external users, then almost all our header files would need to have the _p suffix.

mtl1979: If trees_p.h contains non-private bits, they should be moved to trees.h instead...

The point of having _p suffix only applies when there is two files, one without the suffix, and one with the suffix.

Dead2: That is the point I am making. Now that there is no trees.h, trees_p.h can finally be renamed to trees.h like it should have been. The only thing stopping that was that trees.h was an auto-generated file for the tables.

mtl1979: @Dead2 I'm not even trying to understand all the complex details... As long as we agree there is still something to fix, then everything is good...

nmoinvaz: I have renamed trees_p.h to trees.h and also fixed the alphabetical ordering. 

@Dead2 should crc32_p.h be renamed crc32.h and adler32_p.h be renamed adler32.h?

nmoinvaz: Well i have done it, I can remove the commits if necessary.

Dead2: Well, how we used it in the past has been to have `_p.h` contain C functions that needs to live in a header file for whatever reason. And .h for defines, function declarations and tables.
`crc32_p.h` follows that way of doing it, but `adler32_p.h` has both functions and defines.

So should `adler32_p.h` be split or should we change how we do this? IDK, I am open for suggestions.
`_tpl.h` and `_tbl.h` are both great.
`_p.h` is maybe not, should we put "included functions" into a file with a a new and better suffix perhaps? `_func.h` comes to mind, would that be a good place to put them?

Another alternative is using `_p.h` for all non-externally visible header files, but would require renaming almost all header files, and I honestly don't think that solves anything for us, and I think most programmers are able to figure out they need to include zlib-ng.h.

Open to suggestions, but my vote is currently on `_func.h` (or a better name if someone comes up with one).

mtl1979: I see it as *_p.h having everything that is not/should not be externally visible but shared with multiple files in our source tree, including defines and static inline functions. This keeps us safe when external project has some "global" define that might conflict with what we use, as it will trigger an error/warning during subproject build.

Like I said already earlier, we should only use _p suffix when there is already header file with same basename but without the suffix... Having it on all non-public header files serve no purpose.

Dead2:
Looks good to me

nmoinvaz:
@Dead2 did you want me to remove the last 2 commits that rename _p.h?

Dead2:
Hmm, yeah, I guess we should keep functions in _p, to avoid them being included multiple places etc.

nmoinvaz:
Ok done. Did you want the trees one too? Or is this good now?

Dead2:
Looks great

PR #742, opened 2020-09-06
Enable gzfileops by default when compiling using Cmake or configure

Dead2:
Enable gzfileop by default to simplify transition to zlib-ng.
Shared libraries should include gzfileop, but we still let those that build zlib-ng themselves the option of disabling it.

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -12,7 +12,7 @@ jobs:
         sudo apt-get install -y gcc-10
     - name: Generate project files
       run: |
-        cmake . -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DWITH_GZFILEOP=ON -DWITH_FUZZERS=OFF -DWITH_CODE_COVERAGE=OFF -DWITH_MAINTAINER_WARNINGS=OFF
+        cmake . -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DWITH_FUZZERS=OFF -DWITH_CODE_COVERAGE=OFF -DWITH_MAINTAINER_WARNINGS=OFF
       env:
         CC: gcc-10
         CFLAGS: "-fanalyzer -Wanalyzer-double-fclose -Wanalyzer-double-free -Wanalyzer-exposure-through-output-file -Wanalyzer-file-leak -Wanalyzer-free-of-non-heap -Wanalyzer-malloc-leak -Wanalyzer-null-argument -Wanalyzer-null-dereference -Wanalyzer-possible-null-argument -Wanalyzer-possible-null-dereference -Wanalyzer-stale-setjmp-buffer -Wanalyzer-tainted-array-index -Wanalyzer-unsafe-call-within-signal-handler -Wanalyzer-use-after-free -Wanalyzer-use-of-pointer-in-stale-stack-frame"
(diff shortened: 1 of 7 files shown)


nmoinvaz (approved):
For some reason I thought this would be more complicated than that. Nice!

Dead2:
I don't know how to do this for the MSVC makefiles though.

mtl1979:
@Dead2 Top of the nmake Makefiles list the default values, you just need to change the line saying `WITH_GZFILEOP =`.

Dead2:
@mtl1979 Yeah, I got that much, but I am not sure what it should be set to. 1? True? (upper/lowercase true?)

mtl1979:
@Dead2 `WITH_GZFILEOP = yes` for consistency.... Basically anything non-empty means it is enabled...

Dead2:
Added MSVC changes

Dead2:
Looks like the ARM MSVC tests fail with gzfileops=yes, I guess our CI tests of those has not been good enough.

mtl1979:
@Dead2 Looks like ARM nmake files are missing ZLIB_COMPAT variable and rely on WITH_GZFILEOP to decide what header to use (contents of SUFFIX variable is wrong).
Contents of DEFFILE and SUFFIX variables should rely on contents of both ZLIB_COMPAT and WITH_GZFILEOP...

win32/Makefile.arm
```diff
@@ -36,7 +36,7 @@ RCFLAGS = /dARM /r
 DEFFILE = zlib.def
 RCFILE = zlib1.rc
 RESFILE = zlib1.res
-WITH_GZFILEOP =
+WITH_GZFILEOP = yes
```
mtl1979: Should have `ZLIB_COMPAT =` below this line...

win32/Makefile.a64
```diff
@@ -37,7 +37,7 @@ RCFLAGS = /dARM64 /r
 DEFFILE = zlib.def
 RCFILE = zlib1.rc
 RESFILE = zlib1.res
-WITH_GZFILEOP =
+WITH_GZFILEOP = yes
```
mtl1979: Should have `ZLIB_COMPAT =` below this line...

Dead2:
@mtl1979 Doesn't look like it fixed the failures. I might revert the change for Windows, and let someone who knows nmake fix it instead.

mtl1979:
@Dead2 That's because you didn't fix the conditional block below to change SUFFIX depending on contents of ZLIB_COMPAT.

Dead2:
@mtl1979 Exactly, I'll leave that for someone else to fix then.

mtl1979:
@Dead2 But I just delegated it back to you... lol...

Dead2:
@mtl1979 I still don't know how to fix it, so that won't help.
It also doesn't really belong in this PR since it would be a bugfix and not a simple switch of default settings.

mtl1979:
@Dead2 nmake Makefiles are not the only places where .def files are used... If you change the defaults, even the Windows builds that use configure or cmake start to fail... the nmake Makefiles were intentionally simplified because there was only two possibilities out of four combinations that were needed... Basically in compat mode gz functions were enabled, but in zlib-ng native they were disabled....

Dead2:
I don't see how .def files relate to the compilation failing because of not finding zlib-ng.h, sounds to me like a different/additional problem.

mtl1979:
.def file lists all the symbols that are exported... if you change the preprocessor macros so that more or less symbols are available, then the .def file needs to be updated or otherwise it will fail when linking, instead of when compiling...

PR #743, opened 2020-09-06
Rename variable to correctly indicate register usage

Dead2:
This PR renames the variable `zmm_wsize` to `ymm_wsize` to indicate the correct
register usage, YMM (256bit) instead of ZMM (512bit) register usage.


mtl1979 (approved):
I noticed that discrepancy a long time ago... Back then I thought it was low priority to fix it, but it might be a good thing to do before first release.

PR #744, opened 2020-09-09
Remove FORCE_STATIC and FORCE_STORED

Dead2:
- FORCE_STORED is kind of useless, unless you really need to make a valid deflate stream and your cpu/memory is too small to do compression, and in that case zlib-ng is not a good candidate in the first place.
- FORCE_STATIC is also not likely to be useful.

Both of them only represent just one line of code change, and are unlikely to break, but I am tempted to remove them nevertheless, just to reduce the amount of different possible configs.

### DIFF
--- a/trees.c
+++ b/trees.c
@@ -674,13 +674,9 @@ void Z_INTERNAL zng_tr_flush_block(deflate_state *s, char *buf, uint32_t stored_
         opt_lenb = static_lenb = stored_len + 5; /* force a stored block */
     }
 
-#ifdef FORCE_STORED
-    if (buf != NULL) { /* force stored block */
-#else
     if (stored_len+4 <= opt_lenb && buf != NULL) {
-        /* 4: two words for the lengths */
-#endif
-        /* The test buf != NULL is only necessary if LIT_BUFSIZE > WSIZE.
+        /* 4: two words for the lengths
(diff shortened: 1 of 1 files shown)


trees.c
```diff
@@ -674,12 +674,9 @@ void Z_INTERNAL zng_tr_flush_block(deflate_state *s, char *buf, uint32_t stored_
         opt_lenb = static_lenb = stored_len + 5; /* force a stored block */
     }
 
-#ifdef FORCE_STORED
-    if (buf != NULL) { /* force stored block */
-#else
     if (stored_len+4 <= opt_lenb && buf != NULL) {
         /* 4: two words for the lengths */
-#endif
+
```
mtl1979: No need for empty line, both comment blocks refer to parts of same statement, contents of comment blocks should be merged.

nmoinvaz:
It looks like `FORCE_STATIC` existed in zlib 0.9 and before that it was called `FORCE_METHOD` because it would force static trees on level 3. I'm not sure the need for `FORCE_STATIC` anyways since `Z_FIXED` deflate strategy can be used to control it. 

A quick search of GitHub doesn't appear anybody is using `FORCE_STORED` and no really relevant finds for `FORCE_STATIC` either.

nmoinvaz (approved):

trees.c
Dead2: Fixed

PR #746, opened 2020-09-11
Add support for ZLIB_COMPÅT in nmake Makefiles for ARM/AArch64.

mtl1979:
This should fix compiling when ZLIB_COMPAT is enabled on nmake command-line.


Dead2:
Could you add a commit that enables gz_fileops by default?
That way this change gets better testing by the CI, and it brings nmake in sync with the rest.

PR #747, opened 2020-09-11
Allow forcing UNALIGNED[64]_OK off in cmake.

Dead2:
- Make it possible to disable UNALIGNED[64]_OK in cmake.
- Add a CI test with -O3 and unaligned reads disabled (-O3 also requires UNALIGNED_OK to be false).

PS: I recommend to disable showing whitespace differences when reviewing this.


nmoinvaz:
I think this should also be added to README.md.

nmoinvaz:
What was the idea behind this pull request? Previously it was possible to disable unaligned reads by using -DUNALIGNED_OK=OFF. It does make the CMake clearer..

mtl1979:
@nmoinvaz The new option turns off both with just specifying one option instead of two...

Dead2:
The motivation behind this PR was actually to increase CI code coverage.
But due to vectorizing, we probably cannot allow -O3 and UNALIGNED[64]_OK to be enabled at the same time, so making the option more available is perhaps a first step towards that.
Not sure whether we should just document that, fail at configure time, silently fix it or perhaps always pass the no-vectorize parameter to gcc when compiling files that contain UNALIGNED[64]_OK checks.

Even better would be if we could make a short self-test (in cmake/configure) that checks whether unaligned reads are actually safe and throw an error if they are not. Not sure whether that is feasible though.

mtl1979:
Simple loop that moves a lot of unaligned data using wider variable than 8-bits (16, 32 or 64 bits) should be enough to cause CPU trap with -O3. When cross-compiling, it will not always work as qemu doesn't always trap with unaligned access and just corrupts the data.

PR #748, opened 2020-09-12
coveralls.io support

Dead2:
This is (for now) a hack to add partial coveralls.io support to our CI runs, so we can decide whether to improve that support or drop it.

Have a good look at https://coveralls.io/builds/33411175 to see what the data looks like.
Since codecov.io is unable to fix their systems, hopefully coveralls.io works better..

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -9,7 +9,7 @@ jobs:
       matrix:
         name: [
           Ubuntu GCC,
-          Ubuntu GCC OSB -O1,
+          Ubuntu GCC OSB,
           Ubuntu GCC Link Zlib,
           Ubuntu GCC No AVX2,
           Ubuntu GCC No SSE2,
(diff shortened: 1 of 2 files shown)


nmoinvaz:
Do they support these coverage commits?

nmoinvaz:
I also noticed Windows and macOS runs missing. Perhaps that is why the coverage % is higher. Maybe we can use `gcovr` (python) instead of `lcov`?

Dead2:
@nmoinvaz Yeah, this is (so far) just a hack so we can see what our coverage data looks like in their system. We would have to find a way to get coverage for windows and macos too of course, if we decide to go this way.

So far though, I like their stability.
I don't love the coverage visualization, and I am confused about files saying 100% coverage when they don't show any coverage for entire functions..

nmoinvaz:
Yeah the font on coveralls.io is not really that great. And there is no sunburst graph that I can find.

Dead2:
The interesting thing here is that this commit gets a MUCH better coverage score (+10%) with codecov.io.
I suspect it picks up the lcov.info file and perhaps it is better capable of processing that?
Could codecov.io perhaps be fed with a pre-processed lcov.info file instead of the raw gcov files?

Could that be what makes codecov.io mess up our stats? The huge influx of files from multiple CI jobs at the same time, so processing collides on their end somehow (missing locks causing overwrites?). Because that would explain why functions suddenly disappear or appear in coverage data, even when nothing changed (such as with the readme.md PR). Perhaps it can also explain the 500 errors and timeouts..

PR #749, opened 2020-09-12
Improve coverage data accuracy

Dead2:
Disable optimization by default when doing coverage tests


nmoinvaz (approved):
Seems ok to me. I thought at one point we had `-O0` C flag for coverage, but perhaps that got removed.

PR #754, opened 2020-09-13
Added ubuntu mingw cross-compile CI instances to cmake GHA.

nmoinvaz:
This might be useful for #729. One thing to note is that the toolchain cmake files seem to be for Windows while this is for Ubuntu. Not sure which platform we care about with regard to mingw.


Dead2:
Nice :)

nmoinvaz:
@mtl1979 should I also update the toolchain files to use these values? Or should we have separate toolchain files for mingw for windows and linux?

mtl1979:
MinGW doesn't use toolchain files under Windows as all the paths are detected by the MinGW Makefile generator.

nmoinvaz:
Ok, in that case I have updated the toolchain files.

nmoinvaz:
I think by now we probably have the most comprehensive set of CI environments out of any project on GitHub.

cmake/toolchain-mingw-i686.cmake
```diff
@@ -1,13 +1,16 @@
-SET(CMAKE_SYSTEM_NAME Windows)
+set(CMAKE_SYSTEM_NAME Linux)
```
mtl1979: This is wrong, MinGW is not Linux...

cmake/toolchain-mingw-x86_64.cmake
```diff
@@ -1,13 +1,16 @@
-SET(CMAKE_SYSTEM_NAME Windows)
+set(CMAKE_SYSTEM_NAME Linux)
```
mtl1979: This is wrong, MinGW is not Linux...

nmoinvaz: Fixed.

cmake/toolchain-mingw-i686.cmake
nmoinvaz: Fixed.

mtl1979: Basically what this line does is for example that it changes the resource file format to PE instead of ELF and use windres to embed them to executable.

PR #756, opened 2020-09-14
Optionally include gzguts.h with WITH_GZFILEOP flag.

nmoinvaz:


Dead2:
What is the motivation here?

nmoinvaz:
I saw gzguts.h was being included even though `WITH_GZFILEOPS` is turned off. Ideally none of those source files should be necessary to be in the source tree with that setting nor compiled in, in any way.

nmoinvaz:
Rebased

PR #757, opened 2020-09-14
Move gzclose() to gzlib.c. 

nmoinvaz:
Modern linkers should be smart enough to link it in only if it is used. I checked the size of minigzip.exe with and without `NO_GZCOMPRESS` and whether or not `gzclose` was in gzclose.c or gzlib.c and there was no difference in MSVC. Perhaps this was done for some ancient linkers?


Dead2 (approved):

nmoinvaz:
Rebased

PR #758, opened 2020-09-14
Fixed match_byte casting warning in compare256_unaligned_64_static.

nmoinvaz:
I encountered this warning on MSVC.

```
compare258.c(158,36): warning C4244: 'return': conversion from 'uint64_t' to 'uint32_t', possible loss of data
```

### DIFF
--- a/compare258.c
+++ b/compare258.c
@@ -155,7 +155,7 @@ static inline uint32_t compare256_unaligned_64_static(const unsigned char *src0,
 
         if (diff) {
             uint64_t match_byte = __builtin_ctzll(diff) / 8;
-            return len + match_byte;
+            return len + (uint32_t)match_byte;
         }
 
         src0 += 8, src1 += 8, len += 8;


compare258.c
```diff
@@ -155,7 +155,7 @@ static inline uint32_t compare256_unaligned_64_static(const unsigned char *src0,
 
         if (diff) {
             uint64_t match_byte = __builtin_ctzll(diff) / 8;
-            return len + match_byte;
+            return len + (uint32_t)match_byte;
```
Dead2, line 158: Why not cast the output from builtin_ctzll instead?
Not that it really matters, the compiler should do the same thing but might be a little more readable.

mtl1979: Changing the type of match_byte doesn't make it more readable as you then need to enclose the whole expression in quotes to make sure it casts the end result and not just the intrinsic. Casting return value or parameter of intrinsic function is invalid as it is essentially "always inlined" assembler instruction, not C function. Any cast involving intrinsic is always optimized away.

Dead2: You are right

PR #762, opened 2020-09-20
Remove gcov-7 install since warnings on gcov-8 no longer occur 

nmoinvaz:
This is a Windows CI issue. It appears that the error messages that once occured on Windows CI no longer do. It might have been due to a version mismatch between gcc & gcov on the GitHub Windows CI instances.


Dead2:
LGTM

PR #766, opened 2020-09-20
Fix compiling on FreeBSD

tpoechtrager:


test/minigzip.c
```diff
@@ -33,8 +33,12 @@
 #endif
 
 #ifndef UNALIGNED_OK
```
Dead2, line 35: I am actually not sure why this is under UNALIGNED_OK check originally (I suspect it is checking the wrong thing), but I also think FreeBSD should not be in that category, at least not in all cases.
x86/x86_64 should define UNALIGNED_OK, probably also on FreeBSD if it does not already, and in that case this does not work like intended.

For this PR, I suggest adding a separate ifdef check for FreeBSD.
Then in a separate PR we can look into changing this whole block to behave the same way as in zutil_p.h.

mtl1979: @dead2 Originally it was only used and required for PowerPC where UNALIGNED_OK is not defined.

Dead2: @mtl1979 Aha. That might not work great when you run Linux on ppc then.
In this case, we should probably try to detect the OS or feature instead to determine whether including malloc.h is needed.

mtl1979: @Dead2 I haven't tried compiling the recent code on my Ubuntu/PPC, but I might when I have changed the hard disk to SSD drive...

nmoinvaz: I agree, not sure why we even have `UNALIGNED_OK` check in here..

michaelortmann: if all "obsolete" memalign()s were switched to aligned_alloc() that `if __FreeBSD__` would not be necessary and we would only need to include stdlib.h instead of malloc.h for all or nearly all systems. then again, its C11, so maybe no option for zlib-ng. anyway we probably also want to add `__OpenBSD__` and the other BSDs here, because they also dont know about malloc.h.

mtl1979: Common sense says we should check in configure phase if aligned_alloc() is available and use it, instead of guessing which operating system (and version) and compiler supports it...

Dead2: @tpoechtrager Do you mind doing the changes suggested in the first comment above?

Dead2: @tpoechtrager *bump*

Dead2: > if all "obsolete" memalign()s were switched to aligned_alloc() that `if __FreeBSD__` would not be necessary and we would only need to include stdlib.h instead of malloc.h for all or nearly all systems. then again, its C11, so maybe no option for zlib-ng. anyway we probably also want to add `__OpenBSD__` and the other BSDs here, because they also dont know about malloc.h.

I looked into this, and unfortunately `aligned_alloc()` is C11. We target C99, so dropping the others might not be so easy.
But we could perhaps support `aligned_alloc()` in addition, and prefer it if it is available.
Any takers for making a PR that does this?

I don't know whether we can get FreeBSD and/or OpenBSD added to CI? I guess it would be useful, but don't know whether it is doable in GA.

nmoinvaz: FreeBSD is [not supported](https://github.com/actions/runner/issues/385) natively in GHA or through self-hosted runner, but it seems some people have tried to do FreeBSD virtualization in macOS: https://github.com/vmactions/freebsd-vm.

valpackett: Also you can use https://builds.sr.ht or http://cirrus-ci.com for native CI

test/minigzip.c
```diff
@@ -33,8 +33,12 @@
...
+#ifdef __FreeBSD__
+#  include <stdlib.h>
+#else
 #  include <malloc.h>
```
valpackett, line 39: Maybe use `stdlib.h` unconditionally?

https://stackoverflow.com/questions/12973311/difference-between-stdlib-h-and-malloc-h

Or if it's used for something Linux specific like `memalign`, `ifdef __linux__` it and use `stdlib.h` otherwise?

mtl1979: @myfreeweb Current version of this file doesn't use memalign() for allocating buffers, but it might be faster if it would use aligned allocation instead of malloc().

valpackett: btw, there's the standardized `posix_memalign` that should be used for this purpose..

mtl1979: posix_memalign() is not always available... Even if something is standardized, it doesn't mean all compilers honor that standard...

nmoinvaz:
Just to test I was able to use the GHA _vmactions_  to run FreeBSD, but because it is essentially a VM within a VM, I'm not sure it would be a good candidate for incorporating it into our GHA workflow. 
https://github.com/nmoinvaz/zlib-ng/commit/1b69748e29e468a32f6cc396525c3138c270492a

And the fact that it takes at least 10 minutes for each run.
https://github.com/nmoinvaz/zlib-ng/actions/runs/529340881

Only change I had to make was to ensure I had the proper packages installed `cmake bash gmake gcc` and edit `testCVEinputs.sh`.

I also setup a VM locally using FreeBSD 12 and had no compilation errors.

past-due:
@nmoinvaz FreeBSD 12+ fixed this, but it's still an issue with earlier versions. Example: #873 

Also, Cirrus CI supports FreeBSD 11 and 12 virtual machines (free for open source): https://cirrus-ci.org/guide/FreeBSD/

An example config is available here: https://github.com/Warzone2100/warzone2100/blob/master/.cirrus.yml

nmoinvaz:
Cirrus CI might be an option - up until now we have been GHA-only. At one point we used Travis CI (which also supports FreeBSD) but it was really slow. When we made the switch GHA integration with GH was superior, but now both have GH integrations.

I'm not sure the status of this PR, it may be abandoned. But it looks like what is needed is `configure` and `CMake` checks to see if `posix_memalign` or `aligned_alloc` exist, and then set a preprocessor for it if exists and use it.

Dead2:
A small fix for compilation on FreeBSD is already in stable, closing this.

PR #774, opened 2020-09-24
Add 'diffoscope' for MacOS in pkgcheck.yml

mtl1979:
Add missing brew package on MacOS for 'diffoscope'.
Disable ABI check for MacOS as the runner doesn't contain abigail.

### DIFF
--- a/.github/workflows/pkgcheck.yml
+++ b/.github/workflows/pkgcheck.yml
@@ -74,7 +74,7 @@ jobs:
     - name: Install packages (macOS)
       if: runner.os == 'macOS'
       run: |
-        brew install ninja ${{ matrix.packages }}
+        brew install ninja diffoscope ${{ matrix.packages }}
 
     - name: Compare builds
       run: |
@@ -97,6 +97,8 @@ jobs:
         LDFLAGS: ${{ matrix.ldflags }}
 
     - name: Check ABI
+      # macOS runner does not contain abigail
+      if: runner.os != 'macOS'
       run: |
         sh test/abicheck.sh --refresh_if
       env:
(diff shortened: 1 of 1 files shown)


mtl1979:
Seems like we should disable abicheck for MacOS as there is no abigail for MacOS...

nmoinvaz:
I'm ok with just testing it on just one platform. @dankegel thoughts?

dankegel:
Well, this is kind of an argument for splitting it out into its own yml; you could also skip abicheck on mac with a conditional.

It would make me uncomfortable to not run pkgcheck.sh on mac...

mtl1979:
I'm not sure if we should or shouldn't make another pull request to fix the issue with abicheck. I already did split this to two pull requests just for pkgcheck.

nmoinvaz:
Add if statement to yaml to check for macOS.
```
- name: Check ABI
   # macOS does not contain abigail
   if: runner.os != 'macOS'
...
- name: Check ABI (compat)
   # macOS does not contain abigail
   if: runner.os != 'macOS'
```

mtl1979:
@nmoinvaz Done...

nmoinvaz:
Perhaps in the future we can change ABI check to use this project: https://github.com/lvc/abi-compliance-checker which supports macOS, Windows, and Linux. @dankegel

dankegel:
Why would you want to switch away from libabigail?

Looks like gnutls switched *to* libabigail recently:
https://gitlab.com/gnutls/gnutls/-/merge_requests/972

mtl1979:
@dankegel obviously because the MacOS runner in GitHub doesn't have abigail.

dankegel:
Guess I wasn't paying attention.  Sure, should be easy to switch (or maybe use the other tool on the other platforms).

mtl1979:
@dankegel I guess I can leave that for another pull request, now pkgcheck at least works on MacOS.

mtl1979:
codecov seems stuck, but I think this is ready for merging when @dead2 has time...

PR #779, opened 2020-09-28
Fix constness in DFLTCC code

iii-i:
Use z_const instead of const in order to fix compiler warnings.

### DIFF
--- a/arch/s390/dfltcc_detail.h
+++ b/arch/s390/dfltcc_detail.h
@@ -46,16 +46,16 @@ typedef enum {
 #define DFLTCC_FACILITY 151
 
 static inline dfltcc_cc dfltcc(int fn, void *param,
-                               unsigned char **op1, size_t *len1, const unsigned char **op2, size_t *len2, void *hist) {
+                               unsigned char **op1, size_t *len1, z_const unsigned char **op2, size_t *len2, void *hist) {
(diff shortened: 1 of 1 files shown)


mtl1979:
This still causes compiler warnings, should use `z_const` in place of `const` to completely fix the issue.

See https://github.com/zlib-ng/zlib-ng/pull/779/checks?check_run_id=1178744879

arch/s390/dfltcc_detail.h
```diff
@@ -46,16 +46,16 @@ typedef enum {
 #define DFLTCC_FACILITY 151
 
 static inline dfltcc_cc dfltcc(int fn, void *param,
-                               unsigned char **op1, size_t *len1, const unsigned char **op2, size_t *len2, void *hist) {
+                               unsigned char **op1, size_t *len1, unsigned char **op2, size_t *len2, void *hist) {
```
mtl1979: Don't remove `const` completely, replace it with `z_const` as that makes the code work with and without zlib compatibility mode.

mtl1979 (changes requested):

arch/s390/dfltcc_detail.h
iii-i: You're right, `next_in` is not `unsigned char *`, it's `z_const unsigned char *`. Same story with `msg`. I'll amend the commit. Thanks!

Issue #592, opened 2020-05-24
gcc-10 static analysis at 1e280ca517976f464116a460729cd0a34b5dd682


gdevenyi:
```
$ gcc --version
gcc-10 (Ubuntu 10.1.0-2ubuntu1~18.04) 10.1.0
$ CC=gcc-10 cmake -DCMAKE_C_FLAGS_RELEASE:STRING='-O3 -DNDEBUG -fanalyzer -Wanalyzer-double-fclose -Wanalyzer-double-free -Wanalyzer-exposure-through-output-file -Wanalyzer-file-leak -Wanalyzer-free-of-non-heap -Wanalyzer-malloc-leak -Wanalyzer-null-argument -Wanalyzer-null-dereference -Wanalyzer-possible-null-argument -Wanalyzer-possible-null-dereference -Wanalyzer-stale-setjmp-buffer -Wanalyzer-tainted-array-index -Wanalyzer-unsafe-call-within-signal-handler -Wanalyzer-use-after-free -Wanalyzer-use-of-pointer-in-stale-stack-frame' .
make -j 40 2> gcc-analyzer.log
```

output:
```
In file included from /home/gdevenyi/scratch/zlib-ng/arch/x86/deflate_quick.c:29:
/home/gdevenyi/scratch/zlib-ng/arch/x86/../../trees_emit.h:75:13: warning: ‘bi_flush’ defined but not used [-Wunused-function]
   75 | static void bi_flush(deflate_state *s) {
      |             ^~~~~~~~
In file included from /home/gdevenyi/scratch/zlib-ng/arch/x86/deflate_quick.c:29:
/home/gdevenyi/scratch/zlib-ng/arch/x86/../../trees_emit.h:75:13: warning: ‘bi_flush’ defined but not used [-Wunused-function]
   75 | static void bi_flush(deflate_state *s) {
      |             ^~~~~~~~
/home/gdevenyi/scratch/zlib-ng/test/example.c: In function ‘test_gzio’:
/home/gdevenyi/scratch/zlib-ng/test/example.c:199:16: warning: pointer targets in passing argument 1 of ‘strcmp’ differ in signedness [-Wpointer-sign]
  199 |     if (strcmp(uncompr, hello) != 0) {
      |                ^~~~~~~
      |                |
      |                unsigned char *
In file included from /home/gdevenyi/scratch/zlib-ng/zutil.h:22,
                 from /home/gdevenyi/scratch/zlib-ng/deflate.h:13,
[...]

PR #780, opened 2020-09-28
Fixed dereference of possibly null head variable in test_deflate_set_header

nmoinvaz:
See #592 for full warning.


nmoinvaz:
@gdevenyi now that this fixes all the static analysis warnings is there a way to make CI error out on static analysis warnings?

gdevenyi:
Yes, I can make a PR to turn that on.

PR #786, opened 2020-10-03
Abort CI tests on sanitizer error.

nmoinvaz:


mtl1979:
`gz_uncompress` in minigzip doesn't free `buf` before calling `error` in case `buf` is non-null.

PR #794, opened 2020-10-20
Add support for compiling to WebAssembly using Emscripten.

PiotrSikora:
Add support for compiling to WebAssembly using Emscripten.

Tested with ./configure:

  CHOST=wasm32 emconfigure ./configure --static
  make

and CMake:

  emcmake cmake . -DCMAKE_C_COMPILER_TARGET=wasm32 -DBUILD_SHARED_LIBS=OFF
  cmake --build .

Signed-off-by: Piotr Sikora <piotrsikora@google.com>


mtl1979:
Does it work using configure without any changes?

nmoinvaz:
Nice! It would be good to add a CI configuration for CMake Git Hub Actions.

PiotrSikora:
> Does it work using configure without any changes?

`./configure` fallbacks to `arch/generic` instead of `arch/x86` like CMake, so this works just fine:
```
CC=emcc CHOST=wasm32 ./configure --static --zlib-compat && make
```
Perhaps the correct fix would be to make CMake fallback to `arch/generic` as well, but that's a breaking change.

mtl1979:
@PiotrSikora I think the main reason why using specific arch instead of generic one is better is when the arch requires specific flags for compile to work correctly... This way for example it can raise error if incompatible option is used.

PiotrSikora:
> @PiotrSikora I think the main reason why using specific arch instead of generic one is better is when the arch requires specific flags for compile to work correctly... This way for example it can raise error if incompatible option is used.

Sure, but this results in inconsistent behavior between `./configure` and CMake for the unsupported architectures.

nmoinvaz:
Is it possible to do sudo apt-get install enscripten? If so it can be added as “packages: enscripten” in the yaml

mtl1979:
@nmoinvaz There is "emscripten" package at least in Ubuntu 18.04

PiotrSikora:
@nmoinvaz @mtl1979 That package is way too old, Ubuntu has Emscripten 1.22.0, which is from 2014, whereas Emscripten and WebAssembly are very actively developed nowadays, so using it is pretty much pointless.

mtl1979:
@PiotrSikora Didn't comment anything about usability of the package... The best would be if Emscripten developers would try to get newer package added to Ubuntu repositories... Seems like GitHub requires quite a few workarounds before the newer package works...

PiotrSikora:
@mtl1979 I agree it would be much better if we could simply install `emscripten` package, but since that's not the case, we have to deal with what we have now. It seems to work now.

mtl1979:
@PiotrSikora I'm all about adding new supported platforms and compilers to zlib-ng build system, especially ones that are still in use... As the workarounds are only needed for GitHub CI, it's not such a big issue... When you finish experimenting, you should squash all the commits...

PiotrSikora:
@mtl1979 squashed. Thanks!

.github/workflows/cmake.yml
```diff
@@ -372,6 +378,21 @@ jobs:
...
 
+    - name: Install Emscripten
+      if: contains(matrix.name, 'Emscripten')
+      run: |
+        git clone https://github.com/emscripten-core/emsdk.git
+        cd emsdk
+        ./emsdk install 1.39.6-upstream
+        ./emsdk activate 1.39.6-upstream
+        source ./emsdk_env.sh
+        echo "${PATH}" >> $GITHUB_PATH
```
mtl1979: When adding whole path, it might be better not to append and instead replace, as it might add some paths twice and that might have unexpected results. We can assume that PATH equals to contents of file pointed by GITHUB_PATH at start of current CI step.

PiotrSikora: Done (replaced with `echo "${PATH}" > $GITHUB_PATH`).

mtl1979:
LGTM

PiotrSikora:
This should be good to merge, I don't think that I can fix the coverage.

mtl1979:
@Dead2 will merge after two long-time contributors approve...

mtl1979:
We could make emcc generate JavaScript wrappers for the binaries and use nodejs to test the generated binaries, but I think that is best to implement in another pull request...

PiotrSikora:
(I removed duplicate `-DBUILD_SHARED_LIBS=OFF` since it's hardcoded later on for all targets)

nmoinvaz:
I wonder if the CI changes would be better as it is own workflow, because it skips testing, code coverage, has to install emscripten from scratch, etc. @Dead2 might have an opinion on this.

PiotrSikora:
Could someone re-run `CMake / Ubuntu MinGW i686`? It timed out in one test.

mtl1979:
@nmoinvaz Adding testing of WebAssembly requires extensive changes to parts unrelated to Emsripten, so it's better to leave to next pull request... Basically we need emcc to create JavaScript wrappers that nodejs can execute as WebAssembly code can't be run directly from command-line.

nmoinvaz:
I wasn't suggesting adding testing. I was suggesting that perhaps since there are enough large changes in the cmake yaml that it should be broken out into its own yaml workflow.

mtl1979:
@nmoinvaz I know, I was suggesting adding testing, as then we know the compiled binary actually works as expected...

Dead2:
I agree that this should probably be in a separate CI workflow. Could you fix that @PiotrSikora or do you need any assistance?

PiotrSikora:
@Dead2 done.

PiotrSikora:
> Basically we need emcc to create JavaScript wrappers that nodejs can execute as WebAssembly code can't be run directly from command-line.

`.js` wrappers are already generated for the test targets, and using `emcmake` sets `CMAKE_CROSSCOMPILING_EMULATOR` to `node` binary that is installed as part of Emscripten SDK, so `ctest` is aware of it, but there is some mangling happening, and `ctest` tries to execute e.g. `"/path/to/emsdk/node/12.18.1_64bit/bin/node";/path/to/zlib-ng/build/minigzip.js;-c;-k`, which it cannot find (note the `;` between emulator and test executable). This is probably easy to fix, but I don't see where this mangling is happening.

mtl1979:
@PiotrSikora ";" is list separator... It's not passing it as space-delimited like it should.

nmoinvaz:
@PiotrSikora this usually happens when converting a list to a string in CMake. Is it possible to just call cmake with `-DCMAKE_CROSSCOMPILING_EMULATOR=node` without using `emcmake`? `CMAKE_CROSSCOMPILING_EMULATOR` is what we use for `qemu` so we know it works for that.

PiotrSikora:
@nmoinvaz that works, thanks!

It looks that most of the tests are failing:
```
26% tests passed, 39 tests failed out of 53

Total Test time (real) =   6.09 sec

The following tests FAILED:
          6 - minigzip-fireworks-R (Failed)
          7 - minigzip-fireworks-h (Failed)
          8 - minigzip-fireworks-T (Failed)
          9 - minigzip-fireworks-0 (Failed)
         10 - minigzip-fireworks-1 (Failed)
         11 - minigzip-fireworks-4 (Failed)
         12 - minigzip-fireworks-5 (Failed)
         13 - minigzip-fireworks-F (Failed)
         14 - minigzip-fireworks-6 (Failed)
         15 - minigzip-fireworks-9 (Failed)
         16 - minigzip-fireworks-f (Failed)
         17 - minigzip-lcet10-R (Failed)
         18 - minigzip-lcet10-h (Failed)
         19 - minigzip-lcet10-T (Failed)
         20 - minigzip-lcet10-0 (Failed)
         21 - minigzip-lcet10-1 (Failed)
         22 - minigzip-lcet10-4 (Failed)
         23 - minigzip-lcet10-5 (Failed)
         24 - minigzip-lcet10-F (Failed)
         25 - minigzip-lcet10-6 (Failed)
         26 - minigzip-lcet10-9 (Failed)
         27 - minigzip-lcet10-f (Failed)
         28 - minigzip-paper-100k-R (Failed)
         29 - minigzip-paper-100k-h (Failed)
         30 - minigzip-paper-100k-T (Failed)
         31 - minigzip-paper-100k-0 (Failed)
         32 - minigzip-paper-100k-1 (Failed)
         33 - minigzip-paper-100k-4 (Failed)
         34 - minigzip-paper-100k-5 (Failed)
         35 - minigzip-paper-100k-F (Failed)
         36 - minigzip-paper-100k-6 (Failed)
         37 - minigzip-paper-100k-9 (Failed)
         38 - minigzip-paper-100k-f (Failed)
         39 - minigzip-detect-text-A (Failed)
         40 - minigzip-detect-binary-A (Failed)
         47 - GH-361 (Failed)
         48 - GH-364 (Failed)
         49 - GH-382 (Failed)
         53 - GH-751 (Failed)
Errors while running CTest
```
I imagine that at least some of it is because of an argument passing and/or stdin/stdout interactions, but unfortunately I don't have time to dig further into this right now.

Let me know how do you want to proceed.

mtl1979:
@PiotrSikora I think it's better to hold back merging until you have more time to investigate why the tests fail... Because we are quite close to first public release, it makes no sense to merge in work-in-progress code...

Dead2:
@PiotrSikora When I said separate workflow, I meant separate workflow files too 😄 
`configure-emscripten.yaml` and `cmake-emscripten.yaml` for example.

I think the tests should probably be fixed before we merge this, otherwise we only have compile tests and those don't really tell us whether things work. And I will hold this until after Stable release.
I like this though, and I hope you fix the remaining issues 👍

Dead2:
Please rebase and split out into separate workflow file.
Hope you have time to look into the failing tests as well 😄

dzenanz:
This is probably interesting for [itk.js](https://insightsoftwareconsortium.github.io/itk-js/index.html). Ping @thewtex.

merceyz:
Would it be possible to enable the x86 SSE instruction set when that is enabled in emscripten?
https://emscripten.org/docs/porting/simd.html#compiling-simd-code-targeting-x86-sse-instruction-set

EDIT:
The ARM version can be built with emscripten so can at least benefit from its instruction set
```sh
CHOST="arm" CFLAGS="-static -msimd128" LDFLAGS="-static -msimd128" emconfigure ./configure --warn --zlib-compat --static
make -j2
```

EDIT 2:
The x86 build works if I remove the cpu checks in https://github.com/zlib-ng/zlib-ng/blob/071e4da02377b05f5f4a5a1db9f07ea3507a5d9f/arch/x86/x86.c and just hardcode the supported instruction sets

hangedfish:
LGTM. It work for WASI-SDK toolchains.

nmoinvaz:
It looks like Emscripten [doesn't support](https://github.com/emscripten-core/emscripten/issues/16755#issuecomment-1102732849) binary data via `stdin/stdout` very well. That is why most of the tests that aren't using `test_compress.cmake` succeed. 

The only way I can see to fix the tests would be to add the ability to specify an output file to `minigzip` and possibly `minideflate`, via another argument say `-o`, but the linux `gzip` command doesn't originally support this. One problem with that is that `minigzip --help` suggests it might one day support multiple files, and `gzip` itself might, so having a singular `-o` argument might not make much sense. 

Another option is simply disable those tests if doing Emscripten. All the `gtest_zlib` tests would run.

nmoinvaz:
Merged in #1329.

PR #797, opened 2020-10-31
Fixed format specifier warning in build_tree call to Tracev.

nmoinvaz:
`stored_len` is `uint32_t` not `unsigned long`
```
Format specifies type 'unsigned long' but the argument has type 'uint32_t' (aka 'unsigned int')
```


nmoinvaz:
CI failing only due to #798.

PR #801, opened 2020-11-02
Remove unnecessary name array from matrix.

nmoinvaz:


nmoinvaz:
CI failing only due to #798.

PR #804, opened 2020-11-04
Separate sanitizers so they can run independently

nmoinvaz:
This issue is due to something I've noticed on Ubuntu and Clang, where running the ASAN and UBSAN sanitizer checks at the same time causes linker issues or causes the sanitizers not to show errors. This PR will allow you to separate each sanitizer so they can run individually.


configure
```diff
@@ -617,6 +637,31 @@ if test $with_msan -eq 1; then
     echo >> configure.log
 fi
 
+if test $with_ubsan -eq 1; then
+    echo -n "Checking for undefined behavior sanitizer... " | tee -a configure.log
+    sanitizers=""
+    for san in  array-bounds bool bounds builtin enum float-cast-overflow float-divide-by-zero function integer-divide-by-zero local-bounds null nonnull-attribute object-size pointer-overflow return returns-nonnull-attribute shift shift-base shift-exponent signed-integer-overflow undefined unsigned-integer-overflow unsigned-shift-base vla-bound vptr; do
```
mtl1979: Two spaces after `in`?

configure
```diff
@@ -617,6 +637,31 @@ if test $with_msan -eq 1; then
...
+        if try $CC -c $CFLAGS $test.c -fsanitize=$san ; then
+            if test -n "$sanitizers"; then
```
mtl1979, line 628: Inconsistent style... Should be space on both sides of `;`.

configure
nmoinvaz: Fixed.

configure
nmoinvaz: Fixed.

CMakeLists.txt
```diff
@@ -304,10 +306,10 @@ endif()
 # Set code coverage compiler flags
 if(WITH_CODE_COVERAGE)
     if(CMAKE_C_COMPILER_ID MATCHES "Clang")
-        set(CMAKE_C_FLAGS "-O0 ${CMAKE_C_FLAGS} -coverage")
+        set(CMAKE_C_FLAGS "-O1 ${CMAKE_C_FLAGS} -coverage")
     elseif(__GNUC__)
         # Some versions of GCC don't support -coverage shorthand
-        set(CMAKE_C_FLAGS "-O0 ${CMAKE_C_FLAGS} -ftest-coverage -fprofile-arcs -fprofile-values")
+        set(CMAKE_C_FLAGS "-O1 ${CMAKE_C_FLAGS} -ftest-coverage -fprofile-arcs -fprofile-values")
```
Dead2: Why -O1 instead of -O0?
This was changed to -O0 because the coverage data was lower quality with the -O1 we used previously, with lines being combined etc.

.github/workflows/cmake.yml
```diff
@@ -11,22 +11,22 @@ jobs:
...
-          - name: Ubuntu GCC OSB -O1 No Unaligned
+          - name: Ubuntu GCC OSB No Unaligned
             os: ubuntu-latest
             compiler: gcc
-            cmake-args: -DWITH_SANITIZERS=ON -DWITH_UNALIGNED=OFF
+            cmake-args: -DWITH_UBSAN=ON -DWITH_UNALIGNED=OFF
             build-dir: ../build
             build-src-dir: ../zlib-ng
             codecov: ubuntu_gcc_osb
-            cflags: -O1 -g3
```
Dead2: Related to the other comment.

CMakeLists.txt
Dead2: Ref: https://gcc.gnu.org/onlinedocs/gcc-4.8.4/gcc/Gcov-and-Optimization.html

nmoinvaz: I mentioned it in our group chat:

I am also getting this warning. I think it should be changed back to -O1:
```
warning: the object size sanitizer has no effect at -O0, but is explicitly enabled: -fsanitize=array-bounds,bool,bounds,builtin,enum,float-divide-by-zero,function,integer-divide-by-zero,null,nonnull-attribute,object-size,pointer-overflow,return,returns-nonnull-attribute,shift,shift-base,shift-exponent,signed-integer-overflow,undefined,unsigned-integer-overflow,vla-bound,vptr,leak [-Winvalid-command-line-argument]
```

Dead2: Or perhaps we should separate coverage and sanitation?
It is quite annoying when the coverage visualization is obviously wrong to the human eye because of the optimizer combining things, and those are only the obvious instances. Makes it hard to trust it when using it for analyzing actual  code usage across platforms.

mtl1979: I agree with @Dead2 that coverage and sanitation should be isolated to avoid conflicting requirements... I'm just worried we end up with a lot more CI runs (again)...

nmoinvaz: Yeah, likely double the number of runs.

Dead2: Do we need to actually run the tests for the sanitizer runs, or can those be left out?

An alternative would be that each run does two compiles, one for testing/coverage and one for sanitizers. That cuts down on the number of CI runs, and with it the long queue and setup-times.

nmoinvaz: Most of the time is spent in the unit tests, very little in the start up from what I can tell. The unit tests need to be run for both sanitizer and coverage.

Dead2: Queue time does not seem to get accounted anywhere, but it is often significant, not sure how much can be saved.
On Ubuntu, it seems that setup takes at least 13seconds, but does not include provisioning and "boot", not sure whether it includes git checkout.
So doing each after each other would save an undeterminate amount of time.

A workaround for the sanitizer problem could perhaps be to disable the "object-size" checker, and rather add that to non-O0 test runs.

mtl1979: I thought doing two compiles in same CI run messes up caching compiler output...It would also require cleaning up build tree when doing build inside source directory, or wiping the separate build directory...

nmoinvaz:
What do you think about having one flag such as `-DWITH_SANITIZER=Address` or `-DWITH_SANITIZER=Memory` or something similar? Instead of 4 separate flags?

mtl1979:
@nmoinvaz When the choices are mutually exclusive, having just one flag makes sense, but it also needs to make sure user doesn't pass something random as the flag contents, so it should abort with error if it can't understand... Also, the flag option should be all capitals, as cmake is partially case insensitive and lower case letters are reserved for internal variable names...

CMakeLists.txt
nmoinvaz: @Dead2 are you suggesting only adding _object-size_ checker when `-O0` not found in `C_FLAGS`? That would be a simple solution.

nmoinvaz:
I was thinking something along the lines of what LLVM-project uses (ie: `-DLLVM_USE_SANITIZER=MemoryWithOrigins`)

mtl1979:
@nmoinvaz CamelCase might work for multiple words, but the matching should still be case insensitive and first character must be capitalized letter.

CMakeLists.txt
Dead2: @mtl1979 Doing two compiles should not be a problem for comiler caches..  I was thinking the sanitizer could be built in a separate directory, thus also removing the need for testing out-of-sourcedir builds separately.

@nmoinvaz I guess that would work, yes.

.github/workflows/cmake.yml
nmoinvaz: Fixed.

CMakeLists.txt
nmoinvaz: Ok it is done.

nmoinvaz:
Now it supports only a single parameter `-DWITH_SANITIZER=Memory` or `--with-sanitizer=memory`

README.md
```diff
@@ -111,12 +111,11 @@ Build Options
...
-| WITH_MSAN                | --with-msan              | Build with memory sanitizer                                                           | OFF     |
+| WITH_GZFILEOP            | --without-gzfileops      | Compile with support for gzFile related functions                                     | ON      |
 | WITH_OPTIM               | --without-optimizations  | Build with optimisations                                                              | ON      |
 | WITH_NEW_STRATEGIES      | --without-new-strategies | Use new strategies                                                                    | ON      |
 | WITH_NATIVE_INSTRUCTIONS | --native                 | Compiles with full instruction set supported on this host (gcc/clang -march=native)   | OFF     |
-| WITH_SANITIZERS          | --with-sanitizers        | Build with address sanitizer and all supported sanitizers other than memory sanitizer | OFF     |
+| WITH_SANITIZER           | --with-sanitizer         | Build with sanitizer (memory, address, undefined)                                     | OFF     |
```
mtl1979, line 118: "Memory, Address, Undefined" -- first character of each option value should be upper case like everywhere else as these are string literals.

nmoinvaz: Maybe @Dead2 can chime in here, I prefer to have it different case for different systems. Or would prefer values be all lowercase.

mtl1979: Gotta have separate documentation for cmake and configure... lol...

Dead2: Honestly, I have no preference here, this is fine with me.

Dead2 (approved):
Looks good to me

Dead2:
This needs a rebase now

nmoinvaz:
Rebased.

nmoinvaz:
@Dead2 can you please re-run CI to make sure it is not a fluke. Nothing in this PR should really change the outcome of the tests. Codecov does look good now though.

Dead2:
Re-run started

nmoinvaz:
Looks like macOS CI build is still using AppleClang. Will need to do `brew install gcc` and `export CC=gcc-10`.

PR #806, opened 2020-11-10
Separate crc32 and crc32_combine tables

nmoinvaz:
This is so the crc32_combine tables are not included when not used if statically linking. Should reduce code size by 4k when statically linking minizip.


crc32_comb.c
```diff
@@ -0,0 +1,108 @@
+/* crc32.c -- compute the CRC-32 of a data stream
```
mtl1979: Wrong filename in header.

nmoinvaz: Fixed.

Dead2:
Finally had time to get some testing done, and everything looks good, thanks.

PR #808, opened 2020-11-14
Add AArch64 feature detection support for FreeBSD

valpackett:
Also fix a typo in CMake output.


arch/arm/armfeature.c
```diff
@@ -3,13 +3,21 @@
...
 #elif defined(_WIN32)
 #  include <winapifamily.h>
 #endif
 
 static int arm_has_crc32() {
 #if defined(__linux__) && defined(HWCAP2_CRC32)
     return (getauxval(AT_HWCAP2) & HWCAP2_CRC32) != 0 ? 1 : 0;
+#elif defined(__FreeBSD__) && defined(__aarch64__)
+    return getenv("QEMU_EMULATING") == NULL
+      && ID_AA64ISAR0_CRC32_VAL(READ_SPECIALREG(id_aa64isar0_el1)) >= ID_AA64ISAR0_CRC32_BASE;
```
mtl1979, line 20: There's some reports that READ_SPECIALREG() needs temporary variable to store the result...

valpackett: It's defined using the GNU block expression extension, defining its own temporary variable on the inside:

```c
#define	READ_SPECIALREG(reg)						\
({	uint64_t _val;							\
	__asm __volatile("mrs	%0, " __STRING(reg) : "=&r" (_val));	\
	_val;								\
})
```

mtl1979: I've skimmed through the relevant parts of FreeBSD sources and there seems to be quite questionable changes... I'm not sure this code works on all FreeBSD versions currently in use.... It might work on latest, but it might also contain unnecessary workarounds...

Older FreeBSD versions used uppercase `ID_AA64ISAR0_EL1` but at some point it was changed to lowercase seen in this patch.

valpackett: `define ID_AA64ISAR0_CRC32_VAL ID_AA64ISAR0_CRC32` is the relevant support for 12.x, since the `_VAL` defines are only coming in 13.

MSRs in assembly are not case sensitive, i.e. even this works:

```diff
--- i/arch/arm/armfeature.c
+++ w/arch/arm/armfeature.c
@@ -17,7 +17,7 @@ static int arm_has_crc32() {
     return (getauxval(AT_HWCAP2) & HWCAP2_CRC32) != 0 ? 1 : 0;
 #elif defined(__FreeBSD__) && defined(__aarch64__)
     return getenv("QEMU_EMULATING") == NULL
-      && ID_AA64ISAR0_CRC32_VAL(READ_SPECIALREG(id_aa64isar0_el1)) >= ID_AA64ISAR0_CRC32_BASE;
+      && ID_AA64ISAR0_CRC32_VAL(READ_SPECIALREG(iD_aA64iSaR0_El1)) >= ID_AA64ISAR0_CRC32_BASE;
 #elif defined(ARM_NOCHECK_ACLE)
     return 1;
 #else
```

mtl1979: The issue is not case sensitivity, the issue is possible conflict with C macro with similar name, but in all uppercase. That macro is defined as series of bit shifts depending on passed in macro parameters.

I only went back to 2019 while going back changes to relevant files, as it was enough to see there is issues that either need to be added as comments or otherwise addressed. If not, we can't review the patch without installing multiple versions of FreeBSD on real hardware as the code can't be run inside emulator like QEMU...

valpackett: > possible conflict with C macro with similar name

Right, that's why lowercase is preferred now. I'm saying there are no backwards compatibility concerns with using lowercase.

For reference, https://phabricator.services.mozilla.com/D55386 is what we committed to Mozilla NSS. (And it's still [the current code](https://searchfox.org/mozilla-central/rev/ba567fa55c6d8ea5a0d9cd8f8b4bbe3149829134/security/nss/lib/freebl/blinit.c) of course.)

> we can't review the patch without installing

Huh, that's surprising. Very rarely do maintainers care about actually checking BSD related patches, especially on non-x86, it's usually all on trust :)

mtl1979: Some people read Donald Duck or other comics, I read FreeBSD source code... I'm not x86 guy, I started with Z80 and since studied quite a lot of different processors, including 68000, 32-bit PowerPC (running BeOS), both 32-bit and 64-bit ARM...

PR #812, opened 2020-12-03
Fixed macOS CI instance to use GCC

nmoinvaz:
macOS GCC CI instance has not been using GCC. Instead it has been using AppleClang because on macOS the `gcc` command points to Apple's clang compiler.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -297,8 +297,10 @@ jobs:
 
           - name: macOS GCC
             os: macos-latest
-            compiler: gcc
+            compiler: gcc-10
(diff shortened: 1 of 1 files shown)


nmoinvaz:
The `CI Fuzz` having issues, appears to be Google's fault.

nmoinvaz:
Rebased.

nmoinvaz:
@dankegel do you know the issue with `CI Pkgcheck - macOS Clang`? Seems there is some kind of mismatch on `/usr/lib/libSystem.B.dylib `. It seems to be unrelated to this commit - could be a change to GitHub CI.

nmoinvaz:
The issue with pkgcheck is not related to this PR.

PR #819, opened 2021-01-04
Improvements to check_match

nmoinvaz:
- Print hex values of mismatch characters (previously it printed the ascii characters which only work well for ascii text not binary files).
- Check that `match` is not 0
- Check that `match` is not the same as `start`


mtl1979:
One CI test is failing, possibly because of dependency on fix in #820.

Issue #815, opened 2020-12-16
Oss-fuzz issue 24294


Dead2:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24294

mtl1979:
Commit range in the regression report is too broad... it doesn't identify single commit that introduced triggering the assertion failure.

nmoinvaz:
Appears to be an issue with `invalid match` in `deflate_slow`. Turning off `WITH_OPTIM` doesn't help.

nmoinvaz:
I believe `s->prev_length` needs to be reset (maybe to 0?) in `fill_window`. What happens is `fill_window` is called and `s->prev_match` is off between deflate calls.

![image](https://user-images.githubusercontent.com/1364432/103111414-a27dd800-4601-11eb-9d17-f22e26c7f402.png)

Results in compressed length of 22016 bytes.

nmoinvaz:
Here is another option:

![image](https://user-images.githubusercontent.com/1364432/103111541-03f27680-4603-11eb-986f-d4152faea55a.png)

This results in compressed length of 21911 bytes. Basically what it does, is it doesn't fill in the window until it has written the last/prev match so long as it is at least > MIN_MATCH. 

With the oss-fuzz use case, it was using level 7.

|Level|7|9|
|-|-|-|
|Gzip|21969 b|21917 b|. 
|Method 1|21998 b|21958 b|
|Method 2|21890 b|21848 b|
|Master (bad output)|21911 b |21862 b|

nmoinvaz:
Looks like Method 2 fails on two CI instances though: https://github.com/nmoinvaz/zlib-ng/runs/1606988187?check_suite_focus=true.

nmoinvaz:
This has been resolved and the 2 open OSS-fuzz issues have been fixed.

PR #820, opened 2021-01-04
Fixed previous match length not reset when match start reset.

nmoinvaz:
See #815 
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=24294

Using test case from OSS-Fuzz:

- match of length 25 is found at `wsize-1` 
- `prev_match` and `prev_length` is set
- `fill_window` is called and `memcpy` occurs which moves buffer
- `s->match_start` is reset to 0
- `check_match` fails since match start no longer points to start of match

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -1246,7 +1246,12 @@ void Z_INTERNAL fill_window(deflate_state *s) {
          */
         if (s->strstart >= wsize+MAX_DIST(s)) {
             memcpy(s->window, s->window+wsize, (unsigned)wsize);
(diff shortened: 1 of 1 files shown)


mtl1979:
LGTM

Dead2:
Looks good to me too.

nmoinvaz:
This appears to also have fixed the other OSS-Fuzz issue:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=22003#c6

PR #821, opened 2021-01-11
Disable semantic interposition in ELF shared libraries

Dead2:
Disallow semantic interposition in ELF shared libraries if supported by the compiler.
This disallows calls to our own exported functions being replaced by LD_PRELOAD, thus
avoiding the potential bugs and allowing the compiler to optimize better.

The best/most detailed description of this compiler parameter I have found is here:
https://fedoraproject.org/wiki/Changes/PythonNoSemanticInterpositionSpeedup

I have not been able to reliably measure a performance difference with this, and that is
likely because we rarely call our own exported functions, and never in any hot areas.

This does give a little size reduction for the shared library.

```
Configure before
 106832    1368      32  108232   1a6c8 libz-ng.so.1.9.9

Configure with PR
 106504    1296      32  107832   1a538 libz-ng.so.1.9.9
 ```

```
Cmake before
 111456    1368      32  112856   1b8d8 libz-ng.so.1.9.9

Cmake with PR
 111064    1296      32  112392   1b708 libz-ng.so.1.9.9
 ```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -364,6 +364,17 @@ elseif(WITH_SANITIZER STREQUAL "Memory")
 elseif(WITH_SANITIZER STREQUAL "Undefined")
     add_undefined_sanitizer()
 endif()
+
+#
+# Check whether compiler supports -fno-semantic-interposition parameter
+#
+set(CMAKE_REQUIRED_FLAGS "-fno-semantic-interposition")
+check_c_source_compiles(
+    "int main() { return 0; }"
+    HAVE_NO_INTERPOSITION
+)
+set(CMAKE_REQUIRED_FLAGS)
+
 #
 # Check if we can hide zlib internal symbols that are linked between separate source files using hidden
 #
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -856,6 +856,18 @@ else
...
 fi
 
+# Check for -fno-semantic-interposition compiler support
+echo "" > test.c
+  cat > $test.c <<EOF
+int main() { return 0; }
+EOF
+if test "$gcc" -eq 1 && ($cc $CFLAGS -fno-semantic-interposition -c $test.c) >> configure.log 2>&1; then
+  echo "Checking for -no-semantic-interposition... Yes." | tee -a configure.log
+  SFLAGS="${SFLAGS} -fno-semantic-interposition"
```
nmoinvaz: Should this be `${SFLAGS}` or `$SFLAGS`. The check for hidden attribute below uses `$SFLAGS`. Maybe it doesn't matter?

CMakeLists.txt
```diff
@@ -468,6 +468,14 @@ if(MSVC)
     add_definitions(-D_CRT_NONSTDC_NO_DEPRECATE)
 endif()
 
+# Check whether compiler supports -fno-semantic-interposition parameter
```
nmoinvaz: Perhaps this block should be moved near check for `attribute(hidden)` in cmake?

nmoinvaz (approved):
I only had minor questions, otherwise looks good!

CMakeLists.txt
Dead2: That was what I initially wanted, but I thought it depended on check_c_source_compile_or_run, now that I have finished it I see that it is not needed. Will rewrite to use check_c_source_compiles() instead, and move it up.

configure
Dead2: I got that from the tests above this one actually. It doesn't really matter, but I'll remove the curly braces for simplicity.
Ref: https://unix.stackexchange.com/questions/4899/var-vs-var-and-to-quote-or-not-to-quote

Dead2: Fixed

CMakeLists.txt
Dead2: Fixed

nmoinvaz (approved):
LGTM

PR #825, opened 2021-01-15
Rename cmake variables to match those used in redhat-based distros.

Dead2:
Cmake has no standard variable names for this purpose, but redhat and derived distros have standardized on these names.
Using these names means that at least on redhat/fedora the package spec file can call a `%cmake` macro that also sets all those variables automatically to match the distro preferences. So this makes it easier to make compliant packages.

This naming style also appears in examples in both official cmake documentation and also in some ubuntu documentation, but I don't know whether ubuntu has any naming policy for these.

So while these might not be standard everywhere, I don't see any downside to renaming them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,11 +36,11 @@ message(STATUS "ZLIBNG_HEADER_VERSION: ${ZLIBNG_HEADER_VERSION}")
 
 project(zlib VERSION ${ZLIB_HEADER_VERSION} LANGUAGES C)
 
-set(INSTALL_BIN_DIR "${CMAKE_INSTALL_PREFIX}/bin" CACHE PATH "Installation directory for executables")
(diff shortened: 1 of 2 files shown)


mtl1979:
Ubuntu has no policies about a lot of things as it isn't base distribution like RedHat, because Ubuntu is based on Debian... We should be talking about how RedHat and Debian do things, not how RedHat and Ubuntu do things...

But... I'm all for making things as clean and friendly (both user and developer PoV) as possible...

gdevenyi:
Instead of coding these manually, can we take advantage of this cmake standard module? https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

Dead2:
@gdevenyi Hmm, I have never seen those variables used for cmake, only for configure. For reference, I have packaged 100+ different applications into rpms.
I am not convinced GNUInstallDirs actually makes it easier or more portable in any way, except that it would follow a standard that few(?) others follow in the cmake case.

What this PR currently implements is about making it easier for package creators by using names that they are used to seeing from other packages, and that in the redhat-derived case means they don't even have to think about them.

PR #832, opened 2021-01-26
Use a version of Xcode that supports ZERO_AR_DATE.

nmoinvaz:
This should fix CI issues with macOS pkgcheck builds. We have to use an older version of Xcode for now with CMake. I think in Xcode 12.2 Apple made a change to how deterministic builds should be created by removed `ZERO_AR_DATE` environment variable from `AppleClang` and requiring the use of `libtool -D`. Our problem is that CMake does not have good support for using `libtool` as of 3.19.

I've tried the methods mentioned in the article below, but the only one that worked was changing the version of Xcode.
https://gitlab.kitware.com/cmake/cmake/-/issues/19852

### DIFF
--- a/.github/workflows/pkgcheck.yml
+++ b/.github/workflows/pkgcheck.yml
@@ -67,6 +67,15 @@ jobs:
       run: |
         brew install ninja diffoscope ${{ matrix.packages }}
 
+    - name: Select Xcode version (macOS)
+      # Use a version of Xcode that supports ZERO_AR_DATE until CMake supports
+      # AppleClang linking with libtool using -D argument
+      # https://gitlab.kitware.com/cmake/cmake/-/issues/19852
+      if: runner.os == 'macOS'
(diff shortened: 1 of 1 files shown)


dankegel:
LGTM, please also file a bug against zlib-ng of the ilk "zlib-ng does not support reproducible builds with Xcode >= 12.2 yet"

nmoinvaz:
Thanks. I'm not quite sure it is a zlib-ng bug; it is more like a shortcoming of CMake.

dankegel:
Maybe file a cmake bug, too, then... if this is upstream, I'd like it to be visible.

nmoinvaz:
A CMake bug for it already exists in the description. I also put it in the workflow.
https://gitlab.kitware.com/cmake/cmake/-/issues/19852.

dankegel:
I bet my code review was as thorough as my description reading :->

Dead2:
Nice! 👍

Issue #811, opened 2020-11-30
nginx: "gzip filter failed to use preallocated memory" alerts appeared in logs when using zlib-ng as drop-in replacement


HansBull:
As the title says, being on Ubuntu 20.04 amd64. This is likely a nginx problem (they made a fix for a similar problem with a Intel gzip version: https://git.kizub.ru/mirrors/nginx/-/commit/f1c4853ea1a981073b067cd1356febd08b86f898 ), but I consider it nevertheless worth reporting here.

mtl1979:
Preallocating memory has been discouraged since early days of zlib... Some applications do it nevertheless. As long as the application developer knows how much smaller than necessary their allocation is, they can adjust the code to allocate more.

nmoinvaz:
Nginx issue is also mentioned in #4.

mtl1979:
I don't know how many open issues we need to track if nginx and other libraries/programs that misuse zlib API fail on zlib-ng... Obviously those will fail on all optimized forks of zlib, not just Intel or zlib-ng.

Dead2:
@HansBull I have released the patch I personally use to make nginx compile cleanly with zlib-ng in native mode, as part of that it also fixes the allocation sizes.
Please see https://github.com/zlib-ng/patches

PR #835, opened 2021-01-30
Porting guide and readme fixes.

Dead2:
This is a the start of a porting guide.

I am initially thinking of pushing this, and then let others make PRs to improve upon it.
I know it lacks all kinds of detail, but I think it might be easier to add those in PRs.


gdevenyi:
Maybe add a reference about buffer sizing re #811?

PORTING.md
```diff
@@ -0,0 +1,47 @@
...
+
+Advantages:
+- Does not conflict with other zlib implementations, and can co-exist as a
+  system library along with zlib.
+- In certain places zlib-ng native uses more appropriate data types, removing
+  the need for some workarounds in the API compared to zlib.
+
+Disadvantages:
+- Requires minor changes to applications to use the prefixed zlib-ng
+  function calls and structs. Usually this means a small prefix has to be added.
```
nmoinvaz: Maybe add the name of the prefix?

- Usually this means a small prefix _zng__ has to be added.

PORTING.md
```diff
@@ -0,0 +1,47 @@
...
+recompiled against the zlib-ng headers and libs to ensure full compatability.
+
+Compile against the zlib.h provided by zlib-ng.
+
+Advantages:
+- Easy to port to, since it only requires a recompile of the application and
+  no changes to the application code.
+
+Disadvantages:
+- Can conflict with a system-installed zlib, as that can often be linked in
```
nmoinvaz, line 27: > Can conflict with system-installed zlib..

- _When installed it_ can conflict with system-installed zlib..?

Maybe the rest of it would need to be reworded too.

README.md
```diff
@@ -190,17 +190,18 @@ Advanced Build Options
...
 | WITH_ACLE                       | --without-acle        | Build with ACLE intrinsics                                          | ON                     |
 | WITH_NEON                       | --without-neon        | Build with NEON intrinsics                                          | ON                     |
 | WITH_POWER8                     |                       | Build with POWER8 optimisations                                     | ON                     |
-| WITH_DFLTCC_DEFLATE             | --with-dfltcc-deflate | Use DEFLATE COMPRESSION CALL instruction for compression on IBM Z   | OFF                    |
-| WITH_DFLTCC_INFLATE             | --with-dfltcc-inflate | Use DEFLATE COMPRESSION CALL instruction for decompression on IBM Z | OFF                    |
-| WITH_UNALIGNED                  |                       | Allow optimizations that use unaligned reads if safe on current arch| ON                    |
+| WITH_DFLTCC_DEFLATE             | --with-dfltcc-deflate | Build with DFLTCC intrinsics for compression on IBM Z               | OFF                    |
```
nmoinvaz, line 202: I think the `add_option`  descriptions in _CMakeLists.txt_ needs to be updated too.

PORTING.md
```diff
@@ -0,0 +1,47 @@
...
+in a single application or system-wide.
+
+Please note that zlib-ng in zlib-compat mode is API-compatible but not
+ABI-compatible, meaning that you cannot simply replace the zlib library/dll
+files and expect the application to work. The application will need to be
+recompiled against the zlib-ng headers and libs to ensure full compatability.
+
+Compile against the zlib.h provided by zlib-ng.
+
+Advantages:
```
nmoinvaz: Consider making Advantages/Disadvantages bold.

PORTING.md
```diff
@@ -0,0 +1,47 @@
...
+----------------
+Zlib-ng can be compiled in zlib-compat mode, suitable for zlib-replacement
+in a single application or system-wide.
+
+Please note that zlib-ng in zlib-compat mode is API-compatible but not
+ABI-compatible, meaning that you cannot simply replace the zlib library/dll
+files and expect the application to work. The application will need to be
+recompiled against the zlib-ng headers and libs to ensure full compatability.
+
+Compile against the zlib.h provided by zlib-ng.
```
nmoinvaz: Consider "zlib.h" being italics

PORTING.md
```diff
@@ -0,0 +1,47 @@
...
+
+zlib-ng native mode
+-------------------
+Zlib-ng in native mode is suitable for co-existing with the standard zlib
+library, allowing applications to implement support and testing separately.
+
+The zlib-ng native has implemented some modernization and simplifications
+in its API, intended to make life easier for application developers.
+
+Compile against zlib-ng.h.
```
nmoinvaz: Consider "zlib-ng.h" being italics

nmoinvaz (approved):
The porting guide looks good. Nice work. I have made some trivial suggestions.

nmoinvaz:
In addition to the buffer size changes, also mentioned in #4 is that it might be good idea to add how to detect existence of zlib-ng over stock zlib. You can say something like:

- It is possible to detect the existence of zlib-ng using `#ifdef ZLIBNG_VERSION` in the zlib-ng.h/zlib.h header.

PORTING.md
Dead2: If I understand correctly, it can still conflict if zlib-ng is statically linked in app/lib and zlib is also for some reason linked in by something else. As I understand, the exact interactions when two conflicting definitions are loaded into the same process, it results in an undefined condition where some functions can be selected from zlib-ng and some from zlib.

Luckily we implemented --fno-semantic-interposisiton, so at least that makes sure zlib-ng will never call zlib functions directly, but the application can still get confused and mix what it uses.

nmoinvaz: Ok sounds good, I was thinking it would only conflict if installed like the README.md suggests.

Dead2:
Updated to fix all comments

PORTING.md
```diff
@@ -0,0 +1,56 @@
...
+Zlib-ng can be compiled in zlib-compat mode, suitable for zlib-replacement
+in a single application or system-wide.
+
+Please note that zlib-ng in zlib-compat mode is API-compatible but not
+ABI-compatible, meaning that you cannot simply replace the zlib library/dll
+files and expect the application to work. The application will need to be
+recompiled against the zlib-ng headers and libs to ensure full compatability.
+
+Zlib-ng can be distinguished from other zlib implementations at compile-time
+by checking the existence of `#ifdef ZLIBNG_VERSION` in the zlib.h header
```
nmoinvaz: Might read better as "checking for the existence of the `ZLIBNG_VERSION` predefined macro in the zlib.h header." (period)

PR #838, opened 2021-01-31
Fixes clang tools in static analysis CI

nmoinvaz:
I have changed it to just use `clang-tools` as I don't want to update this every time GHA changes the version of Ubuntu that is the latest. It will now just use whatever `clang-tools` version Ubuntu is distributing.

I have also cleaned up the YAML a bit to make it easier to read.

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -1,15 +1,16 @@
 name: CI Static Analysis
 on: [push, pull_request]
 jobs:
-  GCC-10:
+  GCC:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
     - name: Install packages (Ubuntu)
       run: |
-        sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
-        sudo apt-get update
         sudo apt-get install -y gcc-10
+
     - name: Generate project files
       run: |
         cmake . -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DWITH_FUZZERS=OFF -DWITH_CODE_COVERAGE=OFF -DWITH_MAINTAINER_WARNINGS=OFF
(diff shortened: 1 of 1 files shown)


.github/workflows/analyze.yml
```diff
@@ -1,39 +1,44 @@
...
     steps:
-    - uses: actions/checkout@v1
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
     - name: Install packages (Ubuntu)
       run: |
-        sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
-        sudo apt-get update
         sudo apt-get install -y gcc-10
```
Dead2, line 12: Is it intentional that you kept gcc-10 here?
I guess there is no gcc-latest or anything like that? Fedora/redhat has an 'alternatives' system that makes symlinks depending on versions installed, not sure whether ubunto has anything like that.

mtl1979: Ubuntu does have "alternatives" system... there is no "latest" package, but there is base package for any package with multiple versions... so basically adding "gcc gcc-10" will add both the base package and version "10"...

gdevenyi: The ubuntu "alternatives" system is woefully underused, and does not come configured for gcc or clang, so its useless for this task.

nmoinvaz: I kept `gcc-10` because using `gcc` (the default is 7.5.0) will complain about all of the `CFLAGS` that are specified. 
```
cc: error: unrecognized command line option ‘-fanalyzer’
cc: error: unrecognized command line option ‘-Wanalyzer-double-fclose’
cc: error: unrecognized command line option ‘-Wanalyzer-double-free’
cc: error: unrecognized command line option ‘-Wanalyzer-exposure-through-output-file’
cc: error: unrecognized command line option ‘-Wanalyzer-file-leak’
cc: error: unrecognized command line option ‘-Wanalyzer-free-of-non-heap’
cc: error: unrecognized command line option ‘-Wanalyzer-malloc-leak’
cc: error: unrecognized command line option ‘-Wanalyzer-null-argument’; did you mean ‘-Wunused-dummy-argument’?
cc: error: unrecognized command line option ‘-Wanalyzer-null-dereference’; did you mean ‘-Wno-null-dereference’?
cc: error: unrecognized command line option ‘-Wanalyzer-possible-null-argument’
cc: error: unrecognized command line option ‘-Wanalyzer-possible-null-dereference’; did you mean ‘-Wno-null-dereference’?
cc: error: unrecognized command line option ‘-Wanalyzer-stale-setjmp-buffer’
cc: error: unrecognized command line option ‘-Wanalyzer-tainted-array-index’
cc: error: unrecognized command line option ‘-Wanalyzer-unsafe-call-within-signal-handler’
cc: error: unrecognized command line option ‘-Wanalyzer-use-after-free’
cc: error: unrecognized command line option ‘-Wanalyzer-use-of-pointer-in-stale-stack-frame’
```
Using the _alternatives_ system, you have to install `gcc-10` as an alternative using another command - and then it will set up the alias.

PR #840, opened 2021-02-04
.github: Test with old gcc and clang on oldest supported ubuntu

dankegel:
That's ubuntu 16.04 (earliest version on github), gcc-5.5, and clang-6.0.

Would like to show compiler version at the beginning of the "generate project files" step, but couldn't figure out how to do it without errors on windows.

FIXME: github actions doesn't support ubuntu 12.04 directly, so for now, use 16.04.
A future pull request should probably add 12.04 via qemu or the like.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -8,6 +8,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - name: Ubuntu 16.04 GCC
+            os: ubuntu-16.04
+            compiler: gcc
+
           - name: Ubuntu GCC
             os: ubuntu-latest
             compiler: gcc
@@ -242,6 +246,11 @@ jobs:
             codecov: ubuntu_gcc_mingw_x86_64
              # Limit parallel test jobs to prevent wine errors
             parallels-jobs: 3
+
+          - name: Ubuntu 16.04 Clang
+            os: ubuntu-16.04
+            compiler: clang-6.0
+            # note: github preinstalls compilers now, see https://github.com/actions/virtual-environments/pull/369/
 
           - name: Ubuntu Clang
             os: ubuntu-latest
--- a/.github/workflows/configure.yml
+++ b/.github/workflows/configure.yml
@@ -13,6 +13,11 @@ jobs:
             compiler: gcc
             configure-args: --warn
 
+          - name: Ubuntu 16.04 GCC
+            os: ubuntu-16.04
+            compiler: gcc
+            configure-args: --warn
+
           - name: Ubuntu GCC OSB
             os: ubuntu-latest
             compiler: gcc


.github/workflows/configure.yml
```diff
@@ -13,6 +13,11 @@ jobs:
             compiler: gcc
             configure-args: --warn
 
+          - name: Ubuntu 16.04 GCC
+            os: ubuntu-16.04
```
nmoinvaz, line 17: This uses oldest ubuntu image.. 
Can we instead do `packages: gcc-5` and `compiler: gcc-5`?

.github/workflows/cmake.yml
```diff
@@ -209,6 +213,11 @@ jobs:
             packages: wine-stable gcc-mingw-w64
             codecov: ubuntu_gcc_mingw_x86_64
 
+          - name: Ubuntu 16.04 Clang
+            os: ubuntu-16.04
+            compiler: clang-6.0
```
nmoinvaz, line 252: I think we can use `ubuntu-latest` so long as we have `package: clang-6.0`, correct? Named `Ubuntu Clang 6.0`?

dankegel: The "Minimum requirements" specifies operating systems, and this is a token effort of
testing that zlib supports clang 6.0 *on ubuntu 16.04*.

.github/workflows/configure.yml
dankegel: An explicit goal of this PR is to test on the oldest supported operating system.

Dead2:
I am uncertain about this one, pros vs cons, this implementation or a changed one?
Would like more thoughts from others.

mtl1979:
@Dead2 Even though Ubuntu 20.04 is LTS release, it won't get all the fixes as the roadmap for 20.04 was locked a long time before the issues were found... 20.10 that is not LTS release will get those fixes, but GitHub doesn't support non-LTS releases... This leaves us to either use 18.04 or wait for 22.04, but that will still take over a year... 

Every LTS release will drop support for oldest release (major version number) of packages and will add support for one more recent package version. Some packages might get renamed or replaced with another package with different name. Currently for example gcc-7 is the oldest gcc release that can be installed on latest Ubuntu.

nmoinvaz:
@Dead2 I think we can wait and see until after Dan rebases on master.

dankegel:
Rebasing looks rather good.  Once the mingw fixes land I can rebase it again...

.github/workflows/cmake.yml
```diff
@@ -243,6 +247,11 @@ jobs:
              # Limit parallel test jobs to prevent wine errors
             parallels-jobs: 3
 
+          - name: Ubuntu 16.04 Clang
+            os: ubuntu-16.04
+            compiler: clang-6.0
+            # note: github preinstalls compilers now, see https://github.com/actions/virtual-environments/pull/369/
```
nmoinvaz, line 253: I'm not sure if the comment about preinstalling the compilers is necessary. It is good to know off hand though.

nmoinvaz:
Looks ok to me. Adds 3 new CI runs.

nmoinvaz (approved):

PR #841, opened 2021-02-13
Added build requirements to readme

nmoinvaz:


README.md
```diff
@@ -120,6 +120,30 @@ Build Options
...
+Build Requirements
+------------------
+
+Supported build systems:
+
+* CMake 3.5.1
+* Configure script (bash/gmake)
+* NMake
+
+Supported C99 compatible compilers:
```
Dead2: This wording makes it seem like no other compilers except the ones listed will work.

Dead2:
Supported wording also makes it seem like we don't support newer releases, and only the ones listed. Perhaps use "Minumum supported" or something like that?

mtl1979:
@Dead2 Obviously we don't support newer releases as those can introduce breaking changes. We can test how easy it is to support any breaking change, but adding the support for newer releases means we will also add that specific version to supported list only if the fix is backwards safe.

README.md
mtl1979: The build scripts might have limited support for other compilers, but we can't really guarantee the compilation finishes successfully or that the resulting binaries work as expected.

nmoinvaz:
I have updated the wording. Let me know what you think.

Dead2:
This is still misleading though, the list still reads as these are the minimums, don't bother trying older versions. IF you read the text above the list, you might consider it could be possible anyway but I would read that as a no-go.

I primarily test using GCC 4.8.5, and this says minimum is 5.5, so that is obviously not a correct minimum.
So this list implies that zlib-ng cannot be compiled on Centos 7, but that is obviously incorrect.

Maybe change this to "Recommended minimum compiler versions", that instead implies that older versions could work but with a reduced feature-set.


And perhaps change this part:
```
Supported build systems:

* CMake 3.5.1 or newer
* Configure script (bash/gmake)
* NMake
```

mtl1979:
@Dead2 gcc 4.8.5 might work with some platforms, but at least ARM requires newer version due to variable scope bug causing ICE.

Dead2:
@mtl1979 Then that fits well with the recommended minimum version.

mtl1979:
@Dead2 Possibly... But with for example, MSVC it gets really ugly... 2013 works, 2015 works, 2017 and 2019 some versions are dud, 2017.9 is what I currently use...

Dead2:
We cannot maintain exhaustive lists of compiler versions that are known to work, that would also mean checking every new release and keep testing all versions.

If necessary, we can however list versions that are known to be broken. That might be better placed on the wiki, with a link to that from the readme file.

mtl1979:
@Dead2 At work we use build matrix that has 10+ compiler versions and 3 different Ubuntu versions + Windows 10... Basically tests from clang/gcc 6.x up to latest available version, using latest binary package version of each release.

Like I've mentioned before, I have Visual C++ 2015, 2017.9 and 2019 installed on same machine and MinGW (32-bit and 64-bit) installed on external hard disk (salvaged from my previous development machine).

gdevenyi:
I can confirm things work fine on icc

nmoinvaz:
I think maybe this would be better for the Wiki instead of the readme

nmoinvaz:
I have moved this information to the [wiki](https://github.com/zlib-ng/zlib-ng/wiki). If anybody has any edits to the wording please feel free to make them. I just wanted to make sure that the information was captured somewhere publicly.

PR #843, opened 2021-02-13
Fixed qemu test runs for Ubuntu 20.

nmoinvaz:
These changes are compatible for Ubuntu 18 too.


Dead2:
From some of the ARM tests:
```
-- Performing Test HAVE_NO_INTERPOSITION
CMake Warning at cmake/detect-sanitizer.cmake:49 (message):
  Leak sanitizer is not supported when cross compiling
  CMakeLists.txt:361 (add_address_sanitizer)
```

nmoinvaz:
MinGW tests fails due to code coverage detection, but this PR should be okay to at least fix the qemu issues. I will make another PR to fix MinGW.

PR #844, opened 2021-02-14
Upload cmake and configure errors when build fails.

nmoinvaz:
Uploads cmake/configure logs on failure..
https://github.com/zlib-ng/zlib-ng/actions/runs/565032267

image


Dead2:
Two questions;
- Can the deletion of these be automated? Thinking about quota if we keep adding big logs all the time. Pastebin-ish 30day availability would be ideal for example.
- Would these show up as downloads for a release as well?

nmoinvaz:
Yes we can specify the retention time. It defaults to 90 days.

It will not show up in GitHub Releases.

nmoinvaz:
I have changed it to 30 days and rebased.

PR #845, opened 2021-02-15
Fixed mingw CI builds

nmoinvaz:
These commits fix issues with the mingw builds when CI was upgraded to Ubuntu 20.04.


dankegel:
LGTM

PR #847, opened 2021-02-18
Reduce number of mingw x86_64 parallel test runs due to wine errors

nmoinvaz:


nmoinvaz:
All checks passed. Christmas miracle!

PR #848, opened 2021-02-23
Reduce number of mingw i686 parallel test runs due to wine errors.

nmoinvaz:
I recently saw another instance but relating to x86. Putting it to 2 for both MinGW instances is probably best.


nmoinvaz:
Ha. Looks like it still fails with `2`. I will try to switch both to `1`.

Issue #697, opened 2020-07-15
configure's --native and cmake's -DWITH_NATIVE_INSTRUCTIONS don't do quite the same thing, but it's ok?


dankegel:
This is a followon to #681 

configure's --native still links in some code that is dead because it's not supported by build machine.

cmake's  -DWITH_NATIVE_INSTRUCTIONS avoids that by doing a configure-time run check.

Because *runtime* runtime checks keep the behavior the same, this discrepancy will probably not matter to end users.

It could affect users who are insanely focused on small code size (and/or possibly small working set size / fitting everything in cache) or who expect configure and cmake to always produce identical output.

So... it's one of those little tiny bit of tech debt lint that will just hang around until it bites someone.
If this affects you, please speak up.

nmoinvaz:
Is this still the case? I reviewed both `-DWITH_NATIVE_INSTRUCTIONS` and `--native` and they appear very similar.

dankegel:
Are the .so's bit for bit identical?

nmoinvaz:
Yes I think so.

Beyond Compare:

![image](https://user-images.githubusercontent.com/1364432/108781869-6f7f9400-751f-11eb-89cd-f33063ef1283.png)

CMake:

![image](https://user-images.githubusercontent.com/1364432/108782169-f3d21700-751f-11eb-98fe-53cbcaf2cdbd.png)

Configure:

![image](https://user-images.githubusercontent.com/1364432/108782373-50353680-7520-11eb-9b27-c65ea8523b6a.png)

dankegel:
As long as they have the same sha256sum, cool.  (Couldn't tell if you checked that.)

nmoinvaz:
BeyondCompare did binary comparison and it says "Binary same" in the bottom left, but here is `sha256sum` output:

![image](https://user-images.githubusercontent.com/1364432/108785363-1cf5a600-7526-11eb-9f1e-dc96237efc3b.png)

dankegel:
Great, wasn't familiar with that tool.

If I figure out what I was talking about, and it's still real, I'll reopen.

nmoinvaz:
Actually I thought about it and figured I'd try it out in CI just to make sure. I can reproduce it.

PR #849, opened 2021-02-23
Only set CPU feature flags in configure script if not using native instructions

nmoinvaz:
macOS dylib was not bit identical when using configure script with `-native` flag and cmake script with `-DWITH_NATIVE_INSTRUCTIONS=ON`. 

This was due to the fact that configure script would use `-mnative -maxv2` for compile checks while cmake would use only `-mnative`. I have also added a CI test for this case and now upload build errors for this workflow. 

See #697 for more background info.

### DIFF
--- a/configure
+++ b/configure
@@ -232,6 +232,15 @@ case `$cc -v 2>&1` in
   *clang*) gcc=1 ;;
 esac
 
+if test $native -eq 1; then
+  avx2flag=""
+  sse2flag=""
+  ssse3flag=""
+  sse4flag=""
+  sse42flag=""
+  pclmulflag=""
+fi
+
 if test $build32 -eq 1; then
   CFLAGS="${CFLAGS} -m32"
   SFLAGS="${SFLAGS} -m32"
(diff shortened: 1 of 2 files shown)


nmoinvaz:
We don't use `pkgcheck` workflow or `configure` script for code coverage so nothing should be affecting the codecov results. Probably another issue with codecov..

configure
```diff
@@ -232,6 +232,14 @@ case `$cc -v 2>&1` in
...
 esac
 
+if test $native -eq 0; then
+  avx2flag="-mavx2"
+  sse2flag="-msse2"
+  ssse3flag="-mssse3"
+  sse4flag="-msse4"
+  sse42flag="-msse4.2"
+  pclmulflag="-mpclmul"
+fi
```
Dead2, line 242: I would rather see this reversed.
So the flags are set appropriately initially, but if $native is enabled, only then zero the variables.
I think that would make more sense.

nmoinvaz: Fixed.

PR #863, opened 2021-02-26
CodeFactor fixes for configure script 

nmoinvaz:
https://www.codefactor.io/repository/github/zlib-ng/zlib-ng/file/develop/configure


Dead2:
LGTM
But will postpone this to after stable release, just in case it breaks anything.

nmoinvaz:
Sounds good to me.

PR #865, opened 2021-03-08
Remove misleading manpage and update README.md.

Dead2:
- Remove misleading manpage.
- Added a related projects section at the end.
- Added blank line after header where missing.
- Added extra blank line before header to make them easier to spot in plain-text mode.
- Changed line-length for Contributing section, to make it more readable as plain-text.


nmoinvaz:
Pkgcheck CI is failing...
Need to remove this from CMakeLists.txt
```
if(NOT SKIP_INSTALL_FILES AND NOT SKIP_INSTALL_ALL)
    install(FILES zlib.3 DESTINATION "${MAN_INSTALL_DIR}/man3" RENAME zlib${SUFFIX}.3)
endif()
```

Dead2:
@nmoinvaz Thanks, completely forgot to add that to the commit.

nmoinvaz:
Configure nows seems to be failing as well..

nmoinvaz:
Makefile.in.. just search for ".3" and you will see all the references.

nmoinvaz:
Also need to remove zlib.3 from INDEX.md.

Dead2:
Configure:
```
# Cleanup these files if building outside source tree
    @if [ ! -f zlib.3 ]; then rm -f zlib.3.pdf Makefile; fi
```
I assume this checks for zlib.3 not existing to mean it is outside the source tree. We'll need a different file to check against, ideas? Maybe just check for `deflate.c` , or perhaps `README.md`?

nmoinvaz:
I think `README.md` would work as a good check.

mtl1979:
The point of the filename is that it only exists in root of zlib source tree, not in any other source tree or build directory.

PR #872, opened 2021-03-14
[CMake] Add `zlibstatic` target alias when `BUILD_SHARED_LIBS` is "OFF"

past-due:
Add `zlibstatic` alias when `BUILD_SHARED_LIBS` is "OFF".

IMO, a matter of debate / preference whether this is desirable for inclusion. Figured I'd offer it up for consideration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -900,6 +900,10 @@ if(NOT DEFINED BUILD_SHARED_LIBS)
 else()
     add_library(zlib ${ZLIB_ALL_SRCS})
 
+    if(NOT BUILD_SHARED_LIBS)
+        add_library(zlibstatic ALIAS zlib)
+    endif()
+
     set(ZLIB_INSTALL_LIBRARIES zlib)
 endif()
 


CMakeLists.txt
```diff
@@ -900,6 +900,11 @@ if(NOT DEFINED BUILD_SHARED_LIBS)
 else()
     add_library(zlib ${ZLIB_ALL_SRCS})
 
+    if(NOT BUILD_SHARED_LIBS)
+		# Add zlibstatic alias
```
nmoinvaz: Please replace tabs with spaces and make sure indentation is correct.

nmoinvaz:
This is probably a good idea. Nice on using `ALIAS`.

CMakeLists.txt
nmoinvaz: I don't think the comment is necessary.

past-due: Force-pushed the changes. 👍

nmoinvaz:
@Dead2 notice that codecov diff is -0.61% (for a change that doesn't affect source code), perhaps `threshold` setting should be 1%, like the example on codecov settings page shows.

Dead2:
@nmoinvaz Well, in this case `ubuntu_gcc` did report big changes in coverage, for whatever reason, and `ubuntu_gcc_osb` seems to have failed to report in its coverage at all. So the report says -432 lines of coverage, and that is too much to ignore if it were to be real.
I wish we could hit an [Ignore] button on that coverage failure though, after manual review.

So, the big question now is whether we want to get this into todays release. It looks safe enough to me.

PR #874, opened 2021-03-15
Add testing of minigzip/minideflate "--help" and invalid parameter

Dead2:
This adds some very basic testing of minigzip/minideflate and switchlevels.
Currently these tools pull down our average coverage numbers, so might as well get that up a bit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1198,6 +1198,9 @@ if(ZLIB_ENABLE_TESTS)
             -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/run-and-redirect.cmake)
     endforeach()
 
+    # Run tests targeting tools
+    include(cmake/test-tools.cmake)
+
     if(NOT WIN32 AND ZLIB_COMPAT)
         add_executable(CVE-2003-0107 test/CVE-2003-0107.c)
         target_link_libraries(CVE-2003-0107 zlib)
--- a/cmake/test-tools.cmake
+++ b/cmake/test-tools.cmake
@@ -1,0 +1,35 @@
+# test-tools.cmake -- Tests targeting tool coverage
+
+# Test --help and invalid parameters for our tools
+set(TEST_COMMAND ${MINIGZIP_COMMAND} "--help")
+add_test(NAME minigzip-help
+    COMMAND ${CMAKE_COMMAND}
+    "-DCOMMAND=${TEST_COMMAND}"
+    -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/run-and-redirect.cmake)
+
+set(TEST_COMMAND ${MINIGZIP_COMMAND} "--invalid")
+add_test(NAME minigzip-invalid
+    COMMAND ${CMAKE_COMMAND}
+    "-DCOMMAND=${TEST_COMMAND}"
+    -DSUCCESS_EXIT=64
+    -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/run-and-redirect.cmake)
+
+set(TEST_COMMAND ${MINIDEFLATE_COMMAND} "--help")
+add_test(NAME minideflate-help
+    COMMAND ${CMAKE_COMMAND}
+     "-DCOMMAND=${TEST_COMMAND}"
+     -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/run-and-redirect.cmake)
+
+set(TEST_COMMAND ${MINIDEFLATE_COMMAND} "--invalid")
+add_test(NAME minideflate-invalid
+    COMMAND ${CMAKE_COMMAND}
+    "-DCOMMAND=${TEST_COMMAND}"
+    -DSUCCESS_EXIT=64
+    -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/run-and-redirect.cmake)
+
+set(TEST_COMMAND ${SWITCHLEVELS_COMMAND} "--help")
+add_test(NAME switchlevels-help
+    COMMAND ${CMAKE_COMMAND}
+     "-DCOMMAND=${TEST_COMMAND}"
+     -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/run-and-redirect.cmake)
+


Dead2:
I noticed that we only ever test minigzip with the `-c` parameter (output to stdout), IIRC that should be a remnant from before we supported `-k` (keep input file). Perhaps time to revert that to testing without -c by default, this should also be a little faster, benefiting some of the CI runners especially (we should test both though, so we don't want to convert all of them).
This might also help improve zlib-ng coverage, since they use different calls into the zlib-ng API.

Another thing not currently being tested in minigzip is the multiple input-file support.

I tried to look into what it would take to add tests without `-c`, but as I am still not very good with cmake, I confused myself and had to revert the changes, this PR is what survived that effort.

Btw, I think our cmake/* files should have a file-level description at the top. Especially run-and-compare.cmake and run-and-redirect.cmake would benefit from a little documentation.

Dead2:
And of course Github Actions decides to start failing to install Ubuntu right now..

Dead2:
`apt-get update` errors out with
```
Err:31 https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_20.04  Release
  Certificate verification failed: The certificate is NOT trusted. The received OCSP status response is invalid.  Could not handshake: Error in the certificate verification. [IP: 195.135.221.134 443]
  ```

**Comment by nmoinvaz** [comment]
I agree with you on all the points you brought up. Sometimes -c is needed to redirect output to a unique file name if multiple tests use the same source file. I think a bunch of the tests do just with different args. But we can still use -k where appropriate. 

Something is wrong with GHA sounds like.. of all days!

**Comment by Dead2** [comment]
@nmoinvaz Any idea whether we would be able (permissions-wise) to run this workaround before running apt-get update?
```
echo >> /etc/apt/apt.conf.d/99verify-peer.conf "Acquire { https::Verify-Peer false }"
```
This would disable the OCSP verification, thus speeding up updates a little and not fail on broken OCSP. There is a minor security risk with it, but I don't really think that is something to worry about for us. Connections would still be encrypted though.

**Comment by nmoinvaz** [comment]
We do have the ability to run `sudo` but maybe that requires some other kind of permissions. It might be best to just wait. They probably have it fixed with-in a day. We may be able to forego `sudo apt-get update` altogether but I have not tried. Most of the packages we use are old so the system should know about them already?

**Comment by Dead2** [comment]
Btw, regarding compressing files with unique names, we could possibly look into implementing support for this gzip parameter:
```
-S, --suffix=SUF  use suffix SUF on compressed files
```

A little disappointing that gzip does not have a `-o` parameter for selecting output filename.

**Comment by nmoinvaz** [comment]
Actually that might only be for compressed files (source) not uncompressed (output). Another idea I had thought of was making a `file(COPY)` of the original source file to a random name before the operation - that is also an option which would not require `-c`.

**Comment by Dead2** [comment]
@nmoinvaz True, but once the compressed file has the suffix, the decompressed file should also get the suffix automatically without the parameter.

**Comment by mtl1979** [comment]
@Dead2 Personally I think it would be bad if minigzip had more features than full gzip, but on the other hand, it would be easier to implement both ```-S```  and ```-o``` at the same time.

**Review by nmoinvaz — APPROVED** [review]

**Comment by nmoinvaz** [comment]
@Dead2 it may be better to remove `sudo apt-get update` if we can get away with it. I remember now that there have also been random times in the past where it has failed before.  

I may not be understanding the `-S` option, would probably have to look at the gzip source code.

**Comment by Dead2** [comment]
@mtl1979 Yes, I do not want us to add `-o`. We could add it as an undocumented feature, but I feel that it would still not be a good idea when it comes to compatibility.

Btw, we could perhaps add fake support for -n/-N/-v, ie that we just print a warning about it not being implemented, instead of erroring out. Those might be useful for scripts if minigzip is impersonating gzip on some platform. Just a thought for now.

**Comment by Dead2** [comment]
@nmoinvaz 
```
# touch test.txt

# gzip -S .foo.gz test.txt

# ls -l test.txt.foo.gz
-rw-r--r-- 1 root root 29 Mar 15 16:56 test.txt.foo.gz

# gunzip test.txt.foo.gz

# ls -l test.txt.foo
-rw-r--r-- 1 root root 0 Mar 15 16:56 test.txt.foo
```

**Comment by mtl1979** [comment]
@Dead2 What came to my mind is that we could make our own version of gzip that would have more features than minigzip, but still use zlib-ng to do the actual decompression/compression... That way we could have more useful features without renaming whole minigzip and changing its purpose.

gzip has field in header to store original filename, but it's not fully supported... Sometimes we don't want the original filename to be preserved and sometimes it would be useful to have it preserved on decompression.

**Comment by Dead2** [comment]
Why did this fail on macOS gcc I wonder..

Only one test failed:
```
test 102
        Start 102: minigzip-news-R

102: Test command: /usr/local/Cellar/cmake/3.19.6/bin/cmake "-DTARGET=/Users/runner/work/zlib-ng/zlib-ng/minigzip" "-DCOMPRESS_ARGS=-k;-c;-R" "-DDECOMPRESS_ARGS=-d;-c" "-DINPUT=/Users/runner/work/zlib-ng/zlib-ng/test/data/corpora/calgary/news" "-DOUTPUT=/Users/runner/work/zlib-ng/zlib-ng/test/data/corpora/calgary/news-minigzip-news-R.gz" "-P" "/Users/runner/work/zlib-ng/zlib-ng/cmake/test-compress.cmake"

105/637 Test #102: minigzip-news-R ............................................................................................................***Failed    0.58 sec

CMake Error at cmake/test-compress.cmake:182 (message):
  Compare minigzip decompress gzip failed: 1
The following tests FAILED:
	102 - minigzip-news-R (Failed)
```

nmoinvaz:
I have seen that happen from time to time. It is never on the same file, but always on macOS.

mtl1979:
@Dead2 Few days ago I was thinking we need to dump expected size and the actual size for failed tests, and it might be even necessary to dump the hex values of differing bytes... Sometimes it's just a simple timeout, sometimes it might be a glitch in the virtual machine. It can even be two tests trying to read or write same file, even though that should already be handled...

Dead2:
It failed for both runs too, since it runs everything twice for PR+Push.

I re-ran them, and now it failed on `minigzip-news-6` for one of them, but the other succeeded.
I fail to see how, but smells like a timing-related bug. Related to a tmp-file name would be my first thought, but AFAIK we don't have conflicts for that, or do we?

mtl1979:
When it compares the files with expected output, the filename used when normalizing line endings isn't unique...

PR #875, opened 2021-03-15
Fix bi_valid handling in deflate_quick()

iii-i:
The attached test started failing after commit ad89d5131b29 ("Don't
write end of last block when returning finish_started."): either with
"bi_buf not flushed" message in debug builds, or by producing corrupted
output in release builds.
    
The problem is that we must not return finish_started when bi_buf is
not empty, because the bits there will be lost. Fix by checking that
bi_valid is not 0.


test/deflate_quick_bi_valid.c
```diff
@@ -0,0 +1,85 @@
...
+    PREFIX3(stream) strm;
+    memset(&strm, 0, sizeof(strm));
+
+    int ret = PREFIX(deflateInit2)(&strm, 1, Z_DEFLATED, 31, 1, Z_FILTERED);
+    if (ret != Z_OK) {
+        fprintf(stderr, "deflateInit2() failed with code %d\n", ret);
+        return EXIT_FAILURE;
+    }
+
+    z_const Bytef next_in[554] =
```
mtl1979: New code should not use ```Bytef```... It's deprecated in zlib-ng.

iii-i: Will change to `unsigned char`.

Dead2:
@iii-i FYI, the `macOS GCC` test is randomly failing, that is a known problem and not your fault.

nmoinvaz:
The changes look pretty straight forward. Having a test case helps validate the edge case. Based on the description of `finish_started` it makes sense that it should need to check `bi_valid`. Nice work tracking this down!

image

iii-i:
Thanks for reviewing! I've rebased and pushed the `unsigned char` change.

Dead2:
Github is struggling today, so CI is not running (scarily it says "All checks have passed" as if nothing is wrong).
Hopefully they get it fixed shortly.

Dead2:
GHA is up again, but the missing checks are still missing, and Github pretends that everything is fine..  Can't even re-run them without a forced re-push.

Dead2:
@iii-i I pushed a dummy commit so the CI would run. 😄 
Please do a force push without it whenever you get back online.

Dead2:
Thanks! 😄

PR #877, opened 2021-03-16
Added comments for cmake test scripts.

nmoinvaz:
I have added comments to these scripts I created. Hopefully these make it easy to follow.

### DIFF
--- a/cmake/test-compress.cmake
+++ b/cmake/test-compress.cmake
@@ -1,3 +1,28 @@
+# test-compress.cmake -- Runs a test against an input file to make sure that the specified
+#   targets are able to to compress and then decompress it successfully. Optionally verify
+#   the results with gzip. Output files are generated with unique names to prevent parallel
+#   tests from corrupting one another. Default target arguments are compatible with minigzip.
+
+# Copyright (C) 2021 Nathan Moinvaziri
+# Licensed under the Zlib license, see LICENSE.md for details
+
+# that test a specific input file for compression or decompression.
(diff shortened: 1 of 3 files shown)


cmake/test-compress.cmake
```diff
@@ -1,3 +1,28 @@
...
+
+# Required Variables
+#   INPUT                   - Input file to test
+#   TARGET or               - Command to run for both compress and decompress
+#     COMPRESS_TARGET and   - Command to run to compress input file
+#     DECOMPRESS_TARGET     - Command to run to decompress output file
+
+# Optional Variables
+#   COMPRESS_ARGS           - Arguments to pass for compress command (default: -c -k)
+#   DECOMPRESS_ARGS         - Arguments to pass to decompress command (default: -d -c)
```
Dead2, line 19: Why are these sent as semicolon-separated lists from CMakeLists.txt?
Ex line 1218: `"-DCOMPRESS_ARGS=-c;-k;-4"`

nmoinvaz: That is the equivalent of a list in CMake or at least how it stores them.

nmoinvaz: It is better for it to be a list in the event that it tries to double quote each argument when calling the command. "-c" "-k" versus "-c -k".

Dead2: Ah, that makes sense then. Thanks for the explanation. 👍

PR #878, opened 2021-03-16
Fix defining HAVE_SYS_SDT_H with CMake

iii-i:
DFLTCC probe points are never activated when compiling with CMake. This
is because check_include_file() only sets an internal CMake variable,
but not a C define [1].

Fix as the link suggests - by adding an explicit add_definitions.

[1] https://cmake.org/pipermail/cmake/2015-June/060830.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -314,13 +314,16 @@ if(WITH_NATIVE_INSTRUCTIONS AND DEFINED NATIVEFLAG)
 endif()
 
 #
-# Check for stndard/system includes
+# Check for standard/system includes
 #
 check_include_file(sys/types.h HAVE_SYS_TYPES_H)
 check_include_file(stdarg.h    HAVE_STDARG_H)
 check_include_file(stdint.h    HAVE_STDINT_H)
 check_include_file(stddef.h    HAVE_STDDEF_H)
 check_include_file(sys/sdt.h   HAVE_SYS_SDT_H)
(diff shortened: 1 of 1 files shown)


mtl1979:
Good catch with the spelling mistake... LGTM

Dead2:
This makes me wonder whether any of the other checks also need defines.. Anyone have the time to do a quick check on that?

mtl1979:
@Dead2 ```HAVE_STDARG_H``` is the only other one that is currently handled...

I'm pretty sure the rest are not handled, because the source files don't use those defines to check if the header can be included.

nmoinvaz:
This is what I do in minizip-ng. https://github.com/zlib-ng/minizip-ng/blob/1593060ecf093411f1f20f655e50992913c6d5b9/CMakeLists.txt#L122-L131

nmoinvaz:
`HAVE_STDARG_H` and `HAVE_UNISTD_H` are set in zconf.h.

https://github.com/zlib-ng/zlib-ng/blob/cb791930b26b3c275b99e0db5fd7f0db35179cc4/CMakeLists.txt#L971-L985

`HAVE_SYS_TYPES_H`, `HAVE_STDDEF_H`, and `HAVE_STDDEF_H` are not used - these are remnants of zlib's CMakeLists.txt:

https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/CMakeLists.txt#L23-L25

PR #879, opened 2021-03-16
Remove unused include checks from CMake. 

nmoinvaz:
See my comment here for background info:
https://github.com/zlib-ng/zlib-ng/pull/878#issuecomment-800413242


mtl1979:
Keeping them for future/past use confused more than removing them ;)

Issue #869, opened 2021-03-10
Test failures on IBM Z


iii-i:
Hi,

I ran the zlib-ng test suite in multiple configurations on IBM Z, and found a number of failures (even without DFLTCC!).
I will try to investigate all of them and post them here.

**Failure 1:**

```
cmake . -DZLIB_COMPAT=ON -DCMAKE_C_COMPILER=gcc '-DCMAKE_C_FLAGS=-O2 -g -m31'
...
The following tests FAILED:
	  2 - example (SEGFAULT)
	 46 - CVE-2003-0107 (SEGFAULT)
	 47 - infcover (SEGFAULT)
	 49 - GH-364 (Failed)
	 50 - GH-382 (Failed)
	 51 - GH-536-segfault (Failed)
	 52 - GH-536-incomplete-read (Failed)
	 53 - GH-536-zero-stored-block (Failed)
```

This is happening for a long time now, I couldn't even find a suitable bisect base.

**Failure 2:** In the following [external test](https://github.com/mephi42/raw_zlib/blob/acec52919e4f47000e646ab80ba81bcf836a8385/raw_zlib/test/test.py#L426):

```
cmake . -DZLIB_COMPAT=ON -DWITH_DFLTCC_DEFLATE=ON -DCMAKE_C_COMPILER=gcc '-DCMAKE_C_FLAGS=-O2 -g -m64'
...
FAIL: test_deflate_params (__main__.TestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/iii/.local/lib/python3.7/site-packages/raw_zlib/test/test.py", line 453, in test_deflate_params
    self._check_inflate(dest, compressed_size, plain)
  File "/home/iii/.local/lib/python3.7/site-packages/raw_zlib/test/test.py", line 422, in _check_inflate
    self.assertEqual(raw_zlib.Z_STREAM_END, err)
AssertionError: 1 != -3
```

Bisected to https://github.com/zlib-ng/zlib-ng/commit/bc5915e2dec741236df51d513229f882430962f3 (PR: https://github.com/zlib-ng/zlib-ng/pull/772).

I think I saw more yesterday, but right now can reproduce only these two. I'll edit the description if more pop up.

Best regards,
Ilya

iii-i:
[...]

PR #880, opened 2021-03-17
Fix block_open handling in deflate_quick()

iii-i:
While tracking down #869, I stumbled upon another deflate_quick issue :-(

---

The attached test fails with "inflate() failed", because the deflate
stream that it produces ends up being corrupted. Bisect points to the
commit e7bb6db09a18 ("Replace hash_bits, hash_size and hash_mask with
defines."), but it's most likely a coincidence.

In any case, the reason is that if we happen to simultaneously exhaust
all the buffers (in, out and bi), we return finish_started without
writing the end of block symbol, which will never happen afterwards.

Fix by adding another check to the tricky condition: if we are in the
middle of a block, return need_more instead of finish_started.


Dead2:
@iii-i Great work on finding this, I'll release a hotfix for this.

Do you have a scripted fuzzing setup that finds these? If so, is it something you can/want to share?
Getting a local fuzzing rig set up has been on my todo ever since we got oss-fuzz in CI, but I have never had the time to read up on it.

iii-i:
Yeah, that's essentially https://github.com/zlib-ng/zlib-ng/commit/08274c765af9d038d5e692dc4024eeaea8313bf6 + some new modifications. I should probably open source the full repo where I develop this, and then we can think whether it would make sense to integrate this into zlib-ng.

Dead2:
@iii-i That looks nice indeed, but as I said I have no experience with fuzzing, so I would not know how to build or run that thing 😄

Getting it into zlib-ng/zlib-ng or perhaps better, a separate zlib-ng/fuzzing (or similar) repo would be great, and would make it easier for everyone to contribute both code and testing.

nmoinvaz:
I am getting these compiler warnings now on `MSVC`:
```
  deflate_quick_block_open.c
C:\Users\Nathan\Source\zlib-ng\test\deflate_quick_block_open.c(62,69): warning C4244: '=': conversion from '__int64' to
 'uint32_t', possible loss of data [C:\Users\Nathan\Source\zlib-ng\deflate_quick_block_open.vcxproj]
C:\Users\Nathan\Source\zlib-ng\test\deflate_quick_block_open.c(73,1): warning C4244: 'initializing': conversion from '_
_int64' to 'uint32_t', possible loss of data [C:\Users\Nathan\Source\zlib-ng\deflate_quick_block_open.vcxproj]
```

iii-i:
Weird that clang and/or gcc didn't catch it. There indeed are a few casts missing in this test. I'll send a PR tomorrow.

PR #882, opened 2021-03-17
Fix .pc file generation in CMakeLists.txt

viccie30:
Two words were swapped in a variable name, leading to empty `libdir` and `includedir` variables in the generated .pc files.


Dead2:
Good catch, thanks 😄

PR #884, opened 2021-03-17
[WIP] Use portable code for aligned memory allocation

### LINKED ISSUE #873: Compilation fails on FreeBSD < 12
Using commit: aeffa9bbf4ef114ebccddfe45a38c6128437333d

FreeBSD 11.4:
-- The C compiler identification is Clang 10.0.0
-- The CXX compiler identification is Clang 10.0.0

```
FAILED: 3rdparty/zlib-ng/CMakeFiles/zlib.dir/zutil.c.o 
/usr/bin/cc -DHAVE_BUILTIN_CTZ -DHAVE_BUILTIN_CTZLL -DHAVE_VISIBILITY_HIDDEN -DHAVE_VISIBILITY_INTERNAL -DUNALIGNED64_OK -DUNALIGNED_OK -DWITH_GZFILEOP -DX86_AVX2 -DX86_AVX2_ADLER32 -DX86_AVX_CHUNKSET -DX86_FEATURES -DX86_PCLMULQDQ_CRC -DX86_SSE2 -DX86_SSE2_CHUNKSET -DX86_SSE2_SLIDEHASH -DX86_SSE42_CMP_STR -DX86_SSE42_CRC_HASH -DX86_SSE42_CRC_INTRIN -DX86_SSSE3 -DX86_SSSE3_ADLER32 -D_LARGEFILE64_SOURCE=1 -D__USE_LARGEFILE64 -I3rdparty/zlib-ng -I../../3rdparty/zlib-ng -fstack-protector-strong -Wall -Wno-implicit-fallthrough -O2 -g -DNDEBUG -fPIC -std=c99 -MD -MT 3rdparty/zlib-ng/CMakeFiles/zlib.dir/zutil.c.o -MF 3rdparty/zlib-ng/CMakeFiles/zlib.dir/zutil.c.o.d -o 3rdparty/zlib-ng/CMakeFiles/zlib.dir/zutil.c.o -c ../../3rdparty/zlib-ng/zutil.c
In file included from ../../3rdparty/zlib-ng/zutil.c:8:
In file included from ../../3rdparty/zlib-ng/zutil_p.h:11:
/usr/include/malloc.h:3:2: error: "<malloc.h> has been replaced by <stdlib.h>"
#error "<malloc.h> has been replaced by <stdlib.h>"
 ^
In file included from ../../3rdparty/zlib-ng/zutil.c:8:
../../3rdparty/zlib-ng/zutil_p.h:21:20: warning: implicit declaration of function 'memalign' is invalid in C99 [-Wimplicit-function-declaration]
    return (void *)memalign(64, size);
                   ^
../../3rdparty/zlib-ng/zutil_p.h:21:12: warning: cast to 'void *' from smaller integer type 'int' [-Wint-to-void-pointer-cast]
    return (void *)memalign(64, size);
           ^
2 warnings and 1 error generated.
```

Although interestingly, zero issues on FreeBSD 12.2. In case it matters (for the warnings), 12.2 has a slightly newer version of Clang:
-- The C compiler identification is Clang 10.0.1
-- The CXX compiler identification is Clang 10.0.1

concatime:
This is a quick reborn of https://github.com/zlib-ng/zlib-ng/issues/766.
@Dead2, can you review this approach? I focused on standards instead of operating systems.
It should probably fix https://github.com/zlib-ng/zlib-ng/issues/873, will see.

We now have two standard “backends” and a fallback one:
 - Set C standard to 11 for [`aligned_alloc`](https://en.cppreference.com/w/c/memory/aligned_alloc);
 - Set POSIX to 2001 (`-D _POSIX_C_SOURCE=200112L`) for [`posix_memalign`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/posix_memalign.html);
 - Otherwise fallback to non-standard `memalign`;

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,7 @@ include(FeatureSummary)
 
 include(cmake/detect-arch.cmake)
 include(cmake/detect-coverage.cmake)
+include(cmake/detect-malloc.cmake)
 include(cmake/detect-sanitizer.cmake)
 
 if(CMAKE_TOOLCHAIN_FILE)
--- a/cmake/detect-malloc.cmake
+++ b/cmake/detect-malloc.cmake
@@ -1,0 +1,32 @@
+# Neither Windows nor macOS support ISO C11 `aligned_alloc`.
+#  https://social.msdn.microsoft.com/Forums/vstudio/4617cd56-4c8d-4717-a96b-348ad3b2f149
+#  https://developer.apple.com/forums/thread/81413
+# macOS supports POSIX.1-2001 `posix_memalign`.
+#  https://opensource.apple.com/source/Libc/Libc-825.25/gen/posix_memalign.3.auto.html
+# Android supports `aligned_alloc` from API 28.
+#  https://github.com/aosp-mirror/platform_bionic/commit/cae21a9b53a10f0cba79bf6783c4a5af16228fed
+# Android supports `posix_memalign` from API 17 (but is available on API 16 via
+# libandroid_support.a, which is linked by default when targeting API 16).
+#  https://github.com/aosp-mirror/platform_bionic/commit/85aad909560508410101c18c6ecc6633df39c596
+#  https://github.com/aosp-mirror/platform_bionic/commit/e219cefc173bf93b8ff710431784e5de30ffab8f
+if(WIN32)
+    return()
+elseif(APPLE)
+    add_definitions(-D MAL_IMPL=1)
+elseif(ANDROID)
+    # TODO: if API level >= 28, MAL_IMPL=2
+    # TODO: elif API level >= 17, MAL_IMPL=1
+    # TODO: else, MAL_IMPL=0
+elseif(NOT CMAKE_C_STANDARD EQUAL 99)
+    return()
+endif()
+
+include(CMakePushCheckState)
+
+cmake_push_check_state(RESET)
+set(CMAKE_REQUIRED_DEFINITIONS -D _POSIX_C_SOURCE=200112L)
+check_function_exists(posix_memalign HAVE_POSIX_MEMALIGN)
+if(HAVE_POSIX_MEMALIGN)
+    add_definitions(-D _POSIX_C_SOURCE=200112L)
+endif()
+cmake_pop_check_state()
--- a/test/minigzip.c
+++ b/test/minigzip.c
@@ -12,8 +12,9 @@
  * real thing.
  */
 
-#define _POSIX_SOURCE 1  /* This file needs POSIX for fdopen(). */
-#define _POSIX_C_SOURCE 200112  /* For snprintf(). */
+#ifndef _POSIX_C_SOURCE
+#define _POSIX_C_SOURCE 200112L /* fdopen, snprintf */
+#endif
 
 #include "zbuild.h"
 #ifdef ZLIB_COMPAT
--- a/zutil_p.h
+++ b/zutil_p.h
@@ -5,20 +5,34 @@
 #ifndef ZUTIL_P_H
 #define ZUTIL_P_H
 
-#ifdef __APPLE__
-#  include <stdlib.h>
-#else
-#  include <malloc.h>
+#if !defined(MAL_IMPL) && !defined(_WIN32)
+#  if __STDC_VERSION__+0 >= 201112L
+#    define MAL_IMPL 2 /* ISO C11 `aligned_alloc` */
+#  elif _POSIX_C_SOURCE+0 >= 200112L
+#    define MAL_IMPL 1 /* POSIX.1-2001 optional `posix_memalign` */
+#  else
+#    pragma message("No portable function for aligned memory allocation found, defaulting to memalign()")
+#  endif
 #endif
 
-/* Function to allocate 16 or 64-byte aligned memory */
+#include <stddef.h> /* NULL, size_t */
+#if MAL_IMPL+0 != 0
+#  include <stdlib.h> /* aligned_alloc, free, posix_memalign */
+#else
+#  include <malloc.h> /* _aligned_malloc, _aligned_free, free, memalign */
+#endif
+
+/* Function to allocate 64-byte aligned memory */
 static inline void *zng_alloc(size_t size) {
 #if defined(_WIN32)
-    return (void *)_aligned_malloc(size, 64);
-#elif defined(__APPLE__)
-    return (void *)malloc(size);     /* MacOS always aligns to 16 bytes */
+    return _aligned_malloc(size, 64);
+#elif MAL_IMPL == 2
+    return aligned_alloc(64, size);
+#elif MAL_IMPL == 1
+    void *ptr;
+    return posix_memalign(&ptr, 64, size) ? NULL : ptr;
 #else
-    return (void *)memalign(64, size);
+    return memalign(64, size);
 #endif
 }
 
--- a/configure
+++ b/configure
@@ -732,7 +732,24 @@ EOF
     echo "Checking for fseeko... No." | tee -a configure.log
   fi
 fi
+echo >> configure.log
 
+cat > $test.c <<EOF
+#define _POSIX_C_SOURCE 200112L
+#include <stdlib.h>
+int main(void) {
+  void *ptr;
+  posix_memalign(&ptr, 64, 10);
+  return EXIT_SUCCESS;
+}
+EOF
+if try $CC $CFLAGS -o $test $test.c $LDSHAREDLIBC; then
+  echo "Checking for posix_memalign... Yes." | tee -a configure.log
+  CFLAGS="${CFLAGS} -D _POSIX_C_SOURCE=200112L"
+  SFLAGS="${SFLAGS} -D _POSIX_C_SOURCE=200112L"
+else
+  echo "Checking for posix_memalign... No." | tee -a configure.log
+fi
 echo >> configure.log
 
 # check for strerror() for use by gz* functions


concatime:
FreeBSD 11 works with `cmake -D CMAKE_C_STANDARD=11`: https://builds.sr.ht/~iemaghni/job/462260

concatime:
FreeBSD 11 works with `cmake -D 'CMAKE_C_FLAGS=-D _POSIX_C_SOURCE=200112L'`: https://builds.sr.ht/~iemaghni/job/462268

Dead2:
This is an interesting approach for sure.

It does make me wonder whether one of these new methods could replace or be preferred before the apple one. Are there any macos users in the house? 😄

nmoinvaz:
I would prefer to see `configure` and `CMake` (see `check_function_exists`) check for available functions, rather than doing `#ifdef` for each platform. It becomes a mess as more platforms are encountered. Then all you need to do is `#ifdef HAVE_ALIGNED_ALLOC`.

concatime:
I don’t have, neither plan to, a MacOS system.
- Neither `std::aligned_alloc` from C++17 nor `aligned_alloc` from C11 are supported! [Source](https://developer.apple.com/forums/thread/81413).
- `posix_memalign` is present. [Source](https://stackoverflow.com/a/22876707).

concatime:
> I would prefer to see `configure` and `CMake` (see `check_function_exists`) check for available functions, rather than doing `#ifdef` for each platform. It becomes a mess as more platforms are encountered. Then all you need to do is `#ifdef HAVE_ALIGNED_ALLOC`.

This should not be required. We know for a fact that `aligned_alloc` exists in C11, except for MacOS, which would use `posix_memalign`. I can rework my PR to include this.

nmoinvaz:
I was saying it would be good to have a unified way of detecting support for those functions, and I think it should be done in the build scripts versus doing the platform `#ifdef` checks.

nmoinvaz:
I have experience with minizip-ng on this same thing, I kept getting commit requests for people needing to change the platform `#ifdef` checks for their particular variant. As soon as I moved the check to the build script, that stopped happening.

concatime:
But my PR focuses on standard, not platforms. The platforms should either support C11 or POSIX. We don’t care about FreeBSD or OpenBSD or any of that. What we can do on the build system is ensure one of C11 or a forced value to `_POSIX_C_SOURCE` if the system is not Windows.
One question: Is `aligned_alloc` supported in C11 Windows?

Dead2:
One issue here is that we target C99, and not C11.

concatime:
And I think POSIX 2001 is "next generation" enough, even 2008.

mtl1979:
It's also common knowledge that gcc is not POSIX-compliant by default... so requiring either C11 or POSIX is just recipe for disaster...

POSIX is pretty much deprecated standard as far as gcc goes.

concatime:
> It's also common knowledge that gcc is not POSIX-compliant by default... so requiring either C11 or POSIX is just recipe for disaster...
> 
> POSIX is pretty much deprecated standard as far as gcc goes.

As far as I understand, POSIX is relevant to libc, not the compiler. So Glibc/musl instead of GCC. But C11 is revelant to the compiler of course.

mtl1979:
Even though Linux is BSD-based, we decided to avoid any legacy BSD functions... As such we should also try to avoid POSIX functions if there is GNU equivalent...

concatime:
> Even though Linux is BSD-based

Since when?

> we decided to avoid any legacy BSD functions...

Make sens to avoid legacy functions, BSD or not.

> As such we should also try to avoid POSIX functions if there is GNU equivalent...

I disagree. We should avoid GNU equivalent and instead focus on POSIX. By doing this, we support *BSD, musl-based distros, Android (bionic libc) and even partially MacOS.

mtl1979:
@concatime We have limited Android support... we can't support all older SDK versions due to missing functions... What comes to other BSD variants (including MacOS, FreeBSD, OpenBSD, NetBSD, Haiku), they are being handled currently...

Dead2:
I think we should support the Posix alternative.
When it comes to what to prefer over the other when both exist, I really don't currently have a preference.

If Posix support needs a compiler parameter to be set, then this would need to be detected and added by configure and cmake. So I see what @nmoinvaz means about doing the detection there.
When it comes to doing the entire detection and selection of malloc type in cmake/configure, I think that should probably be a separate PR that targets Next.

mtl1979:
@Dead2 I still haven't decided what alternatives we should support... It's too early to exclude anything... Even if we have C99 as recommended standard, I think we should not exclude newer standards if doing so doesn't complicate the code too much...

concatime:
I recommend picking the ISO C11 `aligned_alloc` over POSIX.1-2001 `posix_memalign` when available (except for MacOS). I updated my PR to reflect this.

concatime:
So, we should edit the build systems to reflect this logic:
- if Windows, define `MAL_IMPL=4`
    Unsurprisingly, Windows supports neither ISO C11 `aligned_alloc` nor POSIX.1-2001 `posix_memalign`. So the build system should handle this edge case.
    https://social.msdn.microsoft.com/Forums/vstudio/4617cd56-4c8d-4717-a96b-348ad3b2f149
- else if macOS, define `MAL_IMPL=2`
    Sadly, macOS does not support ISO C11 `aligned_alloc` (nor ISO C++17 `std::aligned_alloc`), but supports POSIX.1-2001 `posix_memalign`. So we should ensure that the POSIX version is picked even if C11 is used or `_POSIX_C_SOURCE` is undefined by forcing the value of the implementation.
    https://developer.apple.com/forums/thread/81413
    https://opensource.apple.com/source/Libc/Libc-825.25/gen/posix_memalign.3.auto.html
- else if Android, it depends on the value of `ANDROID_NATIVE_API_LEVEL`/`ANDROID_PLATFORM` from CMake Android NDK:
    - if api level >= 28, define `MAL_IMPL=1`
      https://github.com/aosp-mirror/platform_bionic/commit/cae21a9b53a10f0cba79bf6783c4a5af16228fed#diff-36695d8552ae920ef06523ad8a71a9742db13f3f1d8281f917d5994cea1e03e1R82
    - else if api level >= 17, define `MAL_IMPL=2`
      https://github.com/android/ndk/issues/1339#issuecomment-676821211
      https://github.com/aosp-mirror/platform_bionic/commit/e219cefc173bf93b8ff710431784e5de30ffab8f#diff-36695d8552ae920ef06523ad8a71a9742db13f3f1d8281f917d5994cea1e03e1R80
      https://github.com/aosp-mirror/platform_bionic/commit/85aad909560508410101c18c6ecc6633df39c596
    - else, define `MAL_IMPL=3`
- else if C version < 11,
    - if `posix_memalign` is provided by libc when setting `-D _POSIX_C_SOURCE=200112L`, add `-D _POSIX_C_SOURCE=200112L` to cflags

nmoinvaz:
I have made a PR that shows kind of what I had in mind with the build system check. If you want to take from it feel free.

concatime:
Now, FreeBSD builds w/o any added CMake CLI flags: https://builds.sr.ht/~iemaghni/job/462711

CMakeLists.txt
```diff
@@ -42,20 +42,50 @@ set(INC_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/include" CACHE PATH "Installation d
...
+include(CMakePushCheckState)
+include(CheckCCompilerFlag)
 include(CheckCSourceCompiles)
 include(CheckCSourceRuns)
-include(CheckCCompilerFlag)
-include(CMakeDependentOption)
+include(CheckFunctionExists)
+include(CheckIncludeFile)
+include(CheckSymbolExists)
+include(CheckTypeSize)
```
Dead2: Why reorder these? This makes it more difficult to review, since we cannot easily spot any actual changes.
CMakePushCheckState certainly looks interesting though.

Just to be clear, we prefer to have each commit do one thing  (We should put this in the wiki, I guess).
So code-moving/reordering/cleanup in a separate commit/PR to make reviews/audits easier.
Changes like CMakePushCheckState is the kind of thing I'd like to se a separate PR for, and that it changes the old code to use it as well, to keep an common code style.

The more involved the changes of this PR, the less likely I can merge it into the mostly-fixes 2.0 branch. It can still make the 2.1 branch of course, but that does not have an estimated release-date yet.

My preference would be this: (I can still be persuaded otherwise though)
PR1. A minimal fix that enables building on FreeBSD, but does not touch other platforms.
PR2. Rewrite cmake to use CMakePushCheckState.
PR3. Rewrite aligned malloc detection/selection (Effectively this PR).

CMakeLists.txt
```diff
@@ -42,20 +42,50 @@ set(INC_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/include" CACHE PATH "Installation d
...
+        cmake_push_check_state(RESET)
+        set(CMAKE_REQUIRED_DEFINITIONS -D _POSIX_C_SOURCE=200112L)
+        check_function_exists(posix_memalign HAVE_POSIX_MEMALIGN)
+        if(HAVE_POSIX_MEMALIGN)
+            add_definitions(-D _POSIX_C_SOURCE=200112L)
+        endif()
+        cmake_pop_check_state()
+    endif()
+endif()
+
```
Dead2: I am thinking this block of code could go into `cmake/detect-malloc.cmake`, where we are free to use a lot of space without making the main file more complex.

Dead2:
Not sure whether it is a bug in the code or a bug with coverage measurements, but it doesn't look like posix_memalign() or memalign() gets used. It might be due to compiler builtins being used instead perhaps, and gcov being unable to detect that.

CMakeLists.txt
concatime: I just ordered them in alphabetical order. This way, each `include()` has only one possible place. I can revert this and simply add `include(CMakePushCheckState)` at the end.

EDIT: Done.

CMakeLists.txt
concatime: Makes sens. ~~Will do this once I receive a response for https://github.com/zlib-ng/zlib-ng/pull/884#discussion_r596821634~~ Done.

concatime:
I never used codecov. So folks, I need your help to fix the check “codecov/patch”.

zutil_p.h
```diff
@@ -5,30 +5,44 @@
 #ifndef ZUTIL_P_H
 #define ZUTIL_P_H
 
-#ifdef __APPLE__
-#  include <stdlib.h>
+#if !defined(MAL_IMPL) && !defined(_WIN32)
+#  if __STDC_VERSION__+0 >= 201112L
+#    define MAL_IMPL 2 /* ISO C11 `aligned_alloc` */
```
nmoinvaz, line 10: Having the `MAL_IMPL` with number codes I think hurts readability. Perhaps we can use something more descriptive and also get away from using numbers to represent the different methods. I would be in favor of not having an intermediary `MAL_IMPL`.

zutil_p.h
```diff
@@ -5,30 +5,44 @@
...
 #else
-    return (void *)memalign(64, size);
+    return memalign(64, size);
 #endif
 }
 
 /* Function that can free aligned memory */
 static inline void zng_free(void *ptr) {
-#if defined(_WIN32)
+#ifdef _WIN32
```
nmoinvaz: This kind of formatting clean up should happen in a separate commit.

cmake/detect-malloc.cmake
```diff
@@ -0,0 +1,32 @@
+# Neither Windows nor macOS support ISO C11 `aligned_alloc`.
```
nmoinvaz, line 1: This is almost the same thing as before but in comment form. Where when a new platform/sdk comes along are there going to be new commits for everybody wanting to add in their own URL for the docs. I guess that it is separated in its own .cmake helps.

zutil_p.h
```diff
@@ -5,30 +5,44 @@
...
     _aligned_free(ptr);
 #else
     free(ptr);
 #endif
 }
 
-#endif
+#endif /* ZUTIL_P_H */
```
nmoinvaz: Same, cleanup should happen in separate commit.

test/minigzip.c
```diff
@@ -12,8 +12,9 @@
  * real thing.
  */
 
-#define _POSIX_SOURCE 1  /* This file needs POSIX for fdopen(). */
-#define _POSIX_C_SOURCE 200112  /* For snprintf(). */
+#ifndef _POSIX_C_SOURCE
+#define _POSIX_C_SOURCE 200112L /* fdopen, snprintf */
```
nmoinvaz, line 16: Should it also specify `posix_memalign` in the comment?

cmake/detect-malloc.cmake
```diff
@@ -0,0 +1,32 @@
...
+#  https://developer.apple.com/forums/thread/81413
+# macOS supports POSIX.1-2001 `posix_memalign`.
+#  https://opensource.apple.com/source/Libc/Libc-825.25/gen/posix_memalign.3.auto.html
+# Android supports `aligned_alloc` from API 28.
+#  https://github.com/aosp-mirror/platform_bionic/commit/cae21a9b53a10f0cba79bf6783c4a5af16228fed
+# Android supports `posix_memalign` from API 17 (but is available on API 16 via
+# libandroid_support.a, which is linked by default when targeting API 16).
+#  https://github.com/aosp-mirror/platform_bionic/commit/85aad909560508410101c18c6ecc6633df39c596
+#  https://github.com/aosp-mirror/platform_bionic/commit/e219cefc173bf93b8ff710431784e5de30ffab8f
+if(WIN32)
```
nmoinvaz, line 12: This should be `MSVC` I think.

nmoinvaz (changes requested):

nmoinvaz:
Heading in the right direction. Don't get discouraged.

zutil_p.h
concatime: But I need a way to force a certain implementation from the build system. For Apple, I need to explicitly use the POSIX (MAL_IMPL=1) implementation. For Android, depending on the API level, I need to explicitly use either C11 (MAL_IMPL=2) or POSIX (MAL_IMPL=1) or the fallback `memalign()` (MAL_IMPL=0). If you have a cleaner way, I’m hearing.

nmoinvaz: How about just using separate names for them? `HAVE_POSIX_MEMALIGN` or `HAVE_ALIGNED_ALLOC`, etc..?

concatime: But then you would have something like this:
```
#if !defined(USE_ALIGNED_ALLOC) && !defined(USE_POSIX_MEMALIGN) && !defined(USE_MEMALIGN) && !defined(WIN32)
...
```
which is way too long. Also, I don’t have to set `HAVE_ALIGNED_ALLOC`, `HAVE_POSIX_MEMALIGN` and `HAVE_MEMALIGN` by using my method. I want the code to be platform (unix) agnostic (except for Windows) and only the build systems should problematic platforms like macOS and to a lesser extent Android (only if we decided to support older API levels). For example, the build system should not have to handle FreeBSD because it supports portable functions, nor Linux nor OpenBSD, nor Haiku, etc.

cmake/detect-malloc.cmake
concatime: If these new "platform/sdk" support portable functions (e.g. Fuchsia), then no action is required in the build system.

zutil_p.h
nmoinvaz: It might be longer but it is also more readable. I'm not sure that this auto-detection based on `_POSIX_C_SOURCE` or `__STDC_VERSION__` is necessary. The build system should just determine what is available and set `HAVE_XYZ_ALLOC`. `zng_alloc` would use the first available.

concatime: But we don’t need `HAVE_...` for standard functions (ISO C or POSIX)! If a system does not implement standard functions, then they should be handled in the build system, not the code. I don’t want my code to be dependent on the presence of a build system. That’s why I have auto-detection based on `__STDC_VERSION__` and `_POSIX_C_SOURCE`.

concatime:
Relevant informations found:

> This function was introduced in ISO C11 and hence **may have better portability to modern non-POSIX systems** than posix_memalign.

> This function was introduced in POSIX 1003.1d. Although this function is **superseded by aligned_alloc**, it is more portable to older POSIX systems that do not support ISO C11.

> The memalign function is **obsolete** and aligned_alloc or posix_memalign should be used instead.

https://www.gnu.org/software/libc/manual/html_node/Aligned-Memory-Blocks.html

That’s why we should use standard functions instead of GNU ones.

(emphasis mine)

zutil_p.h
mtl1979: You can't detect by value of ```_POSIX_C_SOURCE``` as it is defined as minimum supported version, not maximum... The build system must detect what is the maximum supported value for the compiler and then add it to compiler command line.

concatime: @mtl1979 I already add to compiler command: https://github.com/zlib-ng/zlib-ng/pull/884/files#diff-f673bdec1661e9b36c026093309afc4ce1086fbb900f9697fb6ecaaa95c27386R30.

mtl1979: @concatime I see conflicting definitions of ```_POSIX_C_SOURCE``` in configure now... In other projects we only define ```-D_GNU_SOURCE -D_DEFAULT_SOURCE``` which selects the latest supported ```_POSIX_C_SOURCE``` value on all modern compilers... The define name was recently changed, so both should be and can be safely defined.

cmake/detect-malloc.cmake
nmoinvaz: Hmm. I might be wrong actually.

Dead2:
This needs a rebase, and probably some redesign due to recent changes.

Dead2:
I think this is already fixed in develop.

PR #885, opened 2021-03-17
Cmake: Accept custom install dirs in various formats from command line.

### LINKED ISSUE #881: Please don't hard-code LIB_INSTALL_DIR
In Gentoo Linux x86_64 libs need to go into `/usr/lib64` but the current CMakeLists.txt file hard-codes `"${CMAKE_INSTALL_PREFIX}/lib" ` which will make installation in Gentoo fail because `/usr/lib` is reserved for x86_32 libraries.

Dead2:
Add capability for letting command-line parameters of various formats override the installation directories.

This also removes the now unused MAN_INSTALL_DIR.

Fixes #881

### DIFF
--- a/cmake/detect-install-dirs.cmake
+++ b/cmake/detect-install-dirs.cmake
@@ -1,0 +1,55 @@
+# detect-install-dirs.cmake -- Detect install directory parameters
+# Copyright (C) 2021 Hans Kristian Rosbach
+# Licensed under the Zlib license, see LICENSE.md for details
+
+# Determine installation directory for executables
+if (DEFINED BIN_INSTALL_DIR)
+    set(BIN_INSTALL_DIR "${BIN_INSTALL_DIR}" CACHE PATH "Installation directory for executables" FORCE)
+elseif (DEFINED INSTALL_BIN_DIR)
+    set(BIN_INSTALL_DIR "${INSTALL_BIN_DIR}" CACHE PATH "Installation directory for executables" FORCE)
+elseif (DEFINED CMAKE_INSTALL_FULL_BINDIR)
+    set(BIN_INSTALL_DIR "${CMAKE_INSTALL_FULL_BINDIR}" CACHE PATH "Installation directory for executables" FORCE)
+elseif (DEFINED CMAKE_INSTALL_BINDIR)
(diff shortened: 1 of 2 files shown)


gdevenyi:
Should this be documented somewhere in addition to the code changes? (which methods are supported?)

cmake/detect-install-dirs.cmake
```diff
@@ -0,0 +1,55 @@
...
+if (DEFINED PKGCONFIG_INSTALL_DIR)
+    set(PKGCONFIG_INSTALL_DIR "${PKGCONFIG_INSTALL_DIR}" CACHE PATH "Installation directory for pkgconfig (.pc) files" FORCE)
+elseif (DEFINED INSTALL_PKGCONFIG_DIR)
+    set(PKGCONFIG_INSTALL_DIR "${INSTALL_PKGCONFIG_DIR}" CACHE PATH "Installation directory for pkgconfig (.pc) files" FORCE)
+elseif (DEFINED CMAKE_INSTALL_FULL_PKGCONFIGDIR)
+    set(PKGCONFIG_INSTALL_DIR "${CMAKE_INSTALL_FULL_PKGCONFIGDIR}" CACHE PATH "Installation directory for pkgconfig (.pc) files" FORCE)
+elseif (DEFINED CMAKE_INSTALL_PKGCONFIGDIR)
+    set(PKGCONFIG_INSTALL_DIR "${LIB_INSTALL_DIR}/${CMAKE_INSTALL_PKGCONFIGDIR}" CACHE PATH "Installation directory for pkgconfig (.pc) files" FORCE)
+else()
+    set(PKGCONFIG_INSTALL_DIR "${LIB_INSTALL_DIR}/pkgconfig" CACHE PATH "Installation directory for pkgconfig (.pc) files")
```
nmoinvaz, line 54: `LIB_INSTALL_DIR` is this correct? The other two use `CMAKE_INSTALL_PREFIX` on their `else()` statements.

nmoinvaz:
I'm not a Linux package wizard, but I only noted one thing - otherwise I see nothing obvious.

cmake/detect-install-dirs.cmake
Dead2: Yes, it is supposed to be referenced to the LIB folder, but well spotted 😉
Of course it also means it must come after `LIB_INSTALL_DIR` has been set above.

nmoinvaz (approved):

Dead2:
@gdevenyi I guess it should. Perhaps on the wiki?

PR #868, opened 2021-03-09
Increase size of gzip output buffer 

nmoinvaz:
On the [Understanding zlib](https://www.euccas.me/zlib/) page, it shows the graph below and from it we can infer that I/O performance may be reduced by the small size of the output buffer. In the case of the *gz* family of functions, the size of the output buffer is defined by `GZBUFSIZE` which by default is 8KB.

Every time `deflate` fills the pending buffer, it needs to be called several more times to output it all to the output buffer.

We should increment the size of the `GZBUFSIZE` to be a multiple of the size of the pending buffer. And for developers who do not use the *gz* family of functions but instead call `deflate` directly, we should document the fact that zlib performs better when the output buffer size is at least as large as the pending buffer. In zlib-ng the size of the pending buffer is 64kb or twice the size of the default zlib library of 32kb.

Using *silesia.tar* you can see the number of calls to `deflate` that happen based on the value of `GZBUFSIZE`:

|GZBUFSIZE|Total # of `deflate` calls|
|-|-|
|8k|24380|
|32k|11345|
|64k|6524|
|128k|3574|
|192k|2435|
|256k|1853|

DFLTCC code that was introduced in zlib-ng is [compiled with a `GZBUFSIZE` value of ~254KB](https://github.com/zlib-ng/zlib-ng/blob/51566828e52803a1355837660ac2afd585ac05e4/CMakeLists.txt#L672). We could remove this carve out in CMake if we wanted.

## ZLIB-NG (GZBUFSIZE=8KB)

```
 Tool: minigzip-8k.exe Levels: 1-9
 Runs: 70         Trim worst: 40

 Level   Comp   Comptime min/avg/max/stddev   Compressed size
 1     47.618%      0.975/0.994/1.013/0.011       100,930,174
 2     35.518%      1.598/1.638/1.671/0.020        75,282,951
 3     34.198%      1.946/1.974/2.016/0.023        72,485,720
 4     32.928%      2.343/2.389/2.437/0.027        69,793,995
 5     32.662%      2.515/2.556/2.618/0.034        69,228,861
 6     32.507%      2.944/2.986/3.056/0.032        68,901,208
 7     32.255%      3.922/3.977/4.075/0.049        68,366,778
 8     32.167%      6.134/6.224/6.384/0.081        68,180,776
 9     32.156%      8.592/8.713/8.930/0.122        68,156,160

 avg1  34.668%                        3.495
 tot                                943.543       661,326,623
 ```

[...]

PR #890, opened 2021-03-18
Added low memory configuration option to CMake and configure.

nmoinvaz:
This is based on discussion from #868.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,6 +76,7 @@ option(ZLIB_ENABLE_TESTS "Build test binaries" ON)
 option(ZLIB_DUAL_LINK "Dual link tests against system zlib" OFF)
 option(WITH_FUZZERS "Build test/fuzz" OFF)
 option(WITH_OPTIM "Build with optimisation" ON)
+option(WITH_REDUCED_MEM "Reduced memory usage for special cases (reduces performance)" ON)
 option(WITH_NEW_STRATEGIES "Use new strategies" ON)
 option(WITH_NATIVE_INSTRUCTIONS
     "Instruct the compiler to use the full instruction set on this host (gcc/clang -march=native)" OFF)
@@ -109,6 +110,7 @@ option(INSTALL_UTILS "Copy minigzip and minideflate during install" OFF)
 
 mark_as_advanced(FORCE
     ZLIB_DUAL_LINK
+    WITH_REDUCED_MEM
     WITH_ACLE WITH_NEON
     WITH_DFLTCC_DEFLATE
     WITH_DFLTCC_INFLATE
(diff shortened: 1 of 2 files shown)


gdevenyi:
Should this option perhaps provide a definition of what low-memory means in the zlib context? I doubt it means a 1GB VM for example?

configure
```diff
@@ -155,6 +156,7 @@ case "$1" in
       echo '    [--without-neon]            Compiles without ARM Neon SIMD instruction set' | tee -a configure.log
       echo '    [--with-dfltcc-deflate]     Use DEFLATE CONVERSION CALL instruction for compression on IBM Z' | tee -a configure.log
       echo '    [--with-dfltcc-inflate]     Use DEFLATE CONVERSION CALL instruction for decompression on IBM Z' | tee -a configure.log
+      echo '    [--with-low-mem]            Build for low memory environments (reduced performance)' | tee -a configure.log
```
Dead2: Would it be clearer if the text was something like this?
"Reduced memory usage for special cases (reduces performance)"

Dead2: Likewise for the other messages of course.

Dead2: @nmoinvaz ^

nmoinvaz: Sorry. I'm not really a fan of this option but I can continue to add it. 

Should I rename the option to `--with-reduced-mem` too?

Dead2: Oh, I didn't know that.

I agree that this option should generally not be used except in certain cases (such as a very very busy nginx proxy for example). It is an useful option, just for a very narrow few niches.

So I think that it should be tagged as an advanced option in cmake.
And perhaps it should not show in the summary unless enabled (Also goes for a few of the other options we'd rather not people use unless they really know what they are doing).

Dead2: And yes, it would be smart to rename the option as well 😄

nmoinvaz:
Rebased.

configure
nmoinvaz: Done.

nmoinvaz:
@Dead2 this has problem in CI. `WITH_REDUCED_MEM` is enabled by default.

CMakeLists.txt
```diff
@@ -76,6 +76,7 @@ option(ZLIB_ENABLE_TESTS "Build test binaries" ON)
 option(ZLIB_DUAL_LINK "Dual link tests against system zlib" OFF)
 option(WITH_FUZZERS "Build test/fuzz" OFF)
 option(WITH_OPTIM "Build with optimisation" ON)
+option(WITH_REDUCED_MEM "Reduced memory usage for special cases (reduces performance)" ON)
```
nmoinvaz, line 79: @Dead2 This should be set to OFF.

Dead2:
@nmoinvaz Good thing it only went to develop so far then 😄, are you making a PR to fix it?

nmoinvaz:
Yes I can

PR #891, opened 2021-03-18
Fix GNUInstallDirs includedir detection

viccie30:
CMake uses `INCLUDEDIR` instead of `INCDIR`.


Dead2:
Oh, I missed those two. 🤦‍♂️

PR #772, opened 2020-09-23
Fixed unsigned integer overflow ASAN error when hash_head > s->strstart.

nmoinvaz:
```
  zlib-ng/deflate_medium.c:244:47: runtime error: unsigned integer overflow: 58442 - 58452 cannot be represented in type 'unsigned int'
```

PR #897, opened 2021-03-20
Restore hash_head != 0 checks

iii-i:
This is the simplest fix I could come up with, but I have to admit I didn't fully understand where exactly violating `hash_head != 0` leads to trouble. But the comments in `deflate_slow`, `deflate_medium`, `deflate_fast` and `LONGEST_MATCH` repeat that this invariant is important :-) I also haven't checked the performance implications of this fix. @nmoinvaz maybe you could come up with a better solution?

---

Commit bc5915e2dec7 ("Fixed unsigned integer overflow ASAN error when
hash_head > s->strstart.") removed hash_head != 0 checks in fast,
medium and slow deflate, because it improved performance [1].

Unfortunately, the attached test started failing after that.
Apparently, as the comments suggest, the code implicitly relies on
matches with the beginning of the window being skipped. So restore the
check.

[1] https://github.com/zlib-ng/zlib-ng/pull/772#issuecomment-710760300


nmoinvaz:
It is probably good that the `hash_head != 0` check is at the end of the if statement for performance reasons since I don't think that  case is hit very often. It makes since that we don't want to match the beginning of the window. Perhaps I incorrectly assumed that `strstart == 0` when `hash_head == 0`.

This reminds of https://github.com/zlib-ng/zlib-ng/commit/446c6974fec0d3b8a55eef7ce36a249a9eb601ea#diff-7e5fd0aa55941ed3e3c9c564a713b3841f20f7defc549277cdf94f6c90882af8. I think we should revert that commit and now check against this PR to see if that commit is warranted anymore. Because it was also intended to fix the same issue.

Dead2:
I had a very close look at this the other day when you told us about this potential bug, and this was the only potential fix I could come up with, but I too had no success explaining why it matters.

@nmoinvaz Will you do the testing of that PR reversion? Sounds like a good idea to investigate its necessity.

Dead2 (approved):
LGTM

nmoinvaz:
@Dead2 I verified it and it still needed.

PR #898, opened 2021-03-20
Only perform auto-detection of floating point ABI on ARM arch.

nmoinvaz:
`-mfloat-abi` does not apply to `aarch64` so we need to check for `arm` arch before doing auto-detection.


Dead2 (approved):
LGTM

PR #900, opened 2021-03-20
Detect unsupported VS at configure time (and not at compile time)

tbeu:
I successfully run CMake configuration on recent develop for VS2010. If I open the generated VS solution in VS2010 I get a many compiler errors. Such kind of compiler version requirements should not occur such late during compile time, but at configuration time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -160,6 +160,11 @@ if("${CMAKE_C_COMPILER}" MATCHES "icc" OR "${CMAKE_C_COMPILER}" MATCHES "icpc" O
         message(STATUS "Ignoring WITH_NATIVE_INSTRUCTIONS; not supported on this configuration")
     endif()
 elseif(MSVC)
+    # Minimum supported MSVC version is 1800 = Visual Studio 12.0/2013
+    # See also https://cmake.org/cmake/help/latest/variable/MSVC_VERSION.html
+    if(MSVC_VERSION VERSION_LESS 1800)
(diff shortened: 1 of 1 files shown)


Dead2 (approved):
LGTM

CMakeLists.txt
```diff
@@ -160,6 +160,11 @@ if("${CMAKE_C_COMPILER}" MATCHES "icc" OR "${CMAKE_C_COMPILER}" MATCHES "icpc" O
         message(STATUS "Ignoring WITH_NATIVE_INSTRUCTIONS; not supported on this configuration")
     endif()
 elseif(MSVC)
+    # Minimum supported MSVC version is 1800 = Visual Studio 12.0/2013
+    # See also https://cmake.org/cmake/help/latest/variable/MSVC_VERSION.html
+    if(MSVC_VERSION VERSION_LESS 1800)
+        message(SEND_ERROR "MSVC_VERSION ${MSVC_VERSION} is not a supported Visual Studio compiler version. Please use Visual Studio 2013 or any later version.")
```
nmoinvaz: My only suggestion is to use a shorter error message. Maybe something like:
```
Unsupported Visual Studio compiler version (requires 2013 or later). 
```

Dead2: I agree, that would be better.

Dead2: @tbeu Would you be ok with making the above change and update the PR?

tbeu: ✅

Issue #901, opened 2021-03-20
PVS-Studio report


pavel-pimenov:
http://oracle.fly-server.ru/pvs-studio/zlib-ng
PVS-Studio Version:	7.12.46137.2843
Total Warnings (GA):	103
Total Warnings (OP):	5
Total Warnings (64):	10

nmoinvaz:
It might be worth rerunning.

nmoinvaz:
I have fixed a few more issues that are now available in `develop` branch. Please rerun and reopen if you have more concerns.

PR #904, opened 2021-03-20
Remove extra check that always is false in compare256_c_static.

nmoinvaz:
This may have been a copy & paste artifact from `compare256_unaligned_16_static`. See #901.


Dead2 (approved):
LGTM

PR #906, opened 2021-03-20
Only run checks for intrinsics if optimizations are enabled.

nmoinvaz:
Checks for instrinics supports and intrinsics flags are still being done if `-DWITH_OPTIM=OFF`. I have also cleaned it up and moved the intrinsics checks to `detect-instrinsics.cmake`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,7 @@ include(FeatureSummary)
 include(cmake/detect-arch.cmake)
 include(cmake/detect-install-dirs.cmake)
 include(cmake/detect-coverage.cmake)
+include(cmake/detect-intrinsics.cmake)
 include(cmake/detect-sanitizer.cmake)
 
 if(CMAKE_TOOLCHAIN_FILE)
@@ -139,23 +140,11 @@ if(CMAKE_C_COMPILER_ID MATCHES "Intel")
         set(WARNFLAGS "-w3")
         set(WARNFLAGS_MAINTAINER "-w3 -Wcheck -Wremarks")
         set(WARNFLAGS_DISABLE "")
-        if(BASEARCH_X86_FOUND)
-            set(AVX2FLAG "-mavx2")
-            set(SSE2FLAG "-msse2")
-            set(SSSE3FLAG "-mssse3")
-            set(SSE4FLAG "-msse4.2")
-        endif()
     else()
         set(WARNFLAGS "/W3")
         set(WARNFLAGS_MAINTAINER "/W5")
         set(WARNFLAGS_DISABLE "")
-        if(BASEARCH_X86_FOUND)
-            set(AVX2FLAG "/arch:AVX2")
-            set(SSE2FLAG "/arch:SSE2")
-            set(SSSE3FLAG "/arch:SSSE3")
-            set(SSE4FLAG "/arch:SSE4.2")
         endif()
-    endif()
     if(WITH_NATIVE_INSTRUCTIONS)
         message(STATUS "Ignoring WITH_NATIVE_INSTRUCTIONS; not supported on this configuration")
     endif()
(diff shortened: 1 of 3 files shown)


nmoinvaz:
Still have to do `configure` script..

Dead2:
I like the cmake changes a lot.

The configure changes I am a little unsure about, I feel that they make the file less readable.
Perhaps do the similar thing and make functions?
Should make sure that they are the portable kind though: [with or without function prefix](https://unix.stackexchange.com/questions/73750/difference-between-function-foo-and-foo)

nmoinvaz:
@Dead2 done. The `check_*_intrinsics` functions are sorted alphabetically like in CMake. Also have added a commit that will print out which static/shared object files are used (like CMake does) for verification.

Dead2:
@nmoinvaz Wow, nice! Will do a full review at a later point, but looks good.

configure
```diff
@@ -1466,6 +1470,8 @@ esac
 
 echo "ARCH: ${ARCH}"
 echo "Using arch directory: ${ARCHDIR}"
+echo "Architecture-specific static object files:${ARCH_STATIC_OBJS}"
+echo "Architecture-specific shared object files:${ARCH_SHARED_OBJS}"
```
mtl1979, line 1562: For readability, there should be space after `:`.

nmoinvaz: Those variables already have a space at the beginning based on the way they are generated.

configure
```diff
@@ -1262,6 +1259,10 @@ case "${ARCH}" in
         CFLAGS="${CFLAGS} ${floatabi}"
         SFLAGS="${SFLAGS} ${floatabi}"
 
+        if test $without_optimizations -eq 0; then
+            check_neon_intrinsics
+        fi
```
mtl1979, line 1337: We should only check if NEON intrinsics are available if the processor might support NEON or VFP. There is a lot of ARM processors that support neither. We currently support VFPv3 and VFPv4 when compiling with gcc or Visual C++.

CMakeLists.txt
```diff
@@ -1261,7 +1125,6 @@ if(ZLIB_ENABLE_TESTS)
 
     add_simple_test_executable(deflate_quick_bi_valid)
     add_simple_test_executable(deflate_quick_block_open)
-    add_simple_test_executable(hash_head_0)
```
mtl1979: This edit seems out-of-place... Has nothing to do with optimizations.

mtl1979:
Needs rebase.

configure
mtl1979: At least on modern x86, the lines are really long, over 200 characters. Even with wide-screen aspect ratio, that is very long.

CMakeLists.txt
nmoinvaz: Fixed.

configure
nmoinvaz: I think that would have to be an improvement for another PR. I don't even own a ARM machine.

CMakeLists.txt
nmoinvaz: Fixed.

nmoinvaz:
Rebased.

configure
```diff
@@ -1210,13 +1179,17 @@ case "${ARCH}" in
             ARCH_STATIC_OBJS="${ARCH_STATIC_OBJS} x86.o"
             ARCH_SHARED_OBJS="${ARCH_SHARED_OBJS} x86.lo"
 
+            check_avx2_intrinsics
```
Dead2, line 1182: Maybe these could have () at the end? Makes it clear it is a function being called.

Dead2:
Needs rebase

configure
nmoinvaz: Fixed.

nmoinvaz:
Rebased and added those `()` to configure script.

Dead2:
CI fails with:
```
../configure: line 1337: syntax error near unexpected token `fi'
../configure: line 1337: `        fi'
```

Seems I was wrong in suggesting (), as I find no references to documentation using those. I have always used them myself, but I guess that is actually not portable or even fully functional after all. Learned something there. Sorry about that.

nmoinvaz:
Ok I have switched it back.

nmoinvaz:
Nice!!

Issue #914, opened 2021-03-24
better compat with zlib?


zhangn1985:
in zlib's zconf.h
```h
#ifndef ZEXTERN
#  define ZEXTERN extern
#endif
#ifndef ZEXPORT
#  define ZEXPORT
#endif
#ifndef ZEXPORTVA
#  define ZEXPORTVA
#endif
```
thus some programs need ZEXTERN will fail to build with zlib-ng, even build zlib-ng with `ZLIB_COMPAT`

here is an example:
https://github.com/libretro/hatari/blob/master/libretro/utils/unzip/zlib.h

nmoinvaz:
It seems old versions of minizip use these. It should only be added to zlib-compat API imo.

PR #915, opened 2021-03-25
Alias ZEXTERN, ZEXPORT and ZEXPORTVA to Z_EXTERN, Z_EXPORT and Z_EXPORTVA respectively.

mtl1979:
See #914.

I chained the defines so any changes to zlib-ng defines are mirrored to stock zlib defines if not overwritten.

### DIFF
--- a/zconf.h.in
+++ b/zconf.h.in
@@ -102,6 +102,18 @@
 #  define Z_EXPORTVA
 #endif
 
+/* For backwards compatibility */
+
+#ifndef ZEXTERN
+#  define ZEXTERN Z_EXTERN
+#endif
+#ifndef ZEXPORT
+#  define ZEXPORT Z_EXPORT
+#endif
+#ifndef ZEXPORTVA
+#  define ZEXPORTVA Z_EXPORTVA
+#endif
+
 /* Fallback for something that includes us. */
 typedef unsigned char Byte;
 typedef Byte Bytef;


Dead2 (approved):
LGTM

Dead2:
While I now kind of regret it, we do include the other backwards compatible types and defines in zconf-ng.h too, so I guess we should add these there too.
I think we will rather plan to deprecate and remove them again later, once more programs have converted.

Dead2:
Maybe our porting document should contain a list of all the defines and types that are deprecated and going away, along with the replacement define/type for each of them. Makes it easier for people to do a search for them.

nmoinvaz:
The reason I don't think it should go in `zconf-ng.h` is because that API is not intended for it to be zlib-compatible. I think the purpose of the zlib-compat API should be to aid adoption, but not the zlib-ng API. I would be in favor of removing the deprecated types from zlib-ng API sooner rather than later.

Dead2:
We could start by adding the (gcc) deprecated attribute to the typedefs, that would trigger a warning whenever someone uses one of them.
https://gcc.gnu.org/onlinedocs/gcc-3.2/gcc/Type-Attributes.html

And we could add pragmas trigger warnings when someone uses the defines.
https://stackoverflow.com/questions/2681259/how-to-deprecate-a-macro-in-gcc

Dead2:
Not having them does mean it would be harder for applications to implement support for both zlib and zlib-ng though.

mtl1979:
The way this pull request implements the defines is that if someone tries to define using the old names on command-line, it will most likely fail. As such it only allows using the defines as they are defined using the new names in this same file... I was pondering possibility to default value of the new defines based on old defines possibly specified in command line, but that would have made the patch a lot more complex.

We currently don't have macro that hides the differences between compilers in how they declare deprecated macros, functions etc., so marking them as deprecated in non-compat mode would have to be separate pull request.

nmoinvaz (approved):

Issue #909, opened 2021-03-22
Provide Windows ARM and ARM64 binaries in releases


MichalPetryka:
Currently releases provide only x86 and x64 Windows binaries, since Windows 10 also supports ARM CPUs, it would be nice to also have builds for those provided.

nmoinvaz:
GitHub Actions does not currently support Windows ARM.

MichalPetryka:
You can build for ARM from x64 though, and as I can see the actions do that?
![image](https://user-images.githubusercontent.com/35800402/112038143-cb4f3400-8b42-11eb-80b9-b4e5992a65ee.png)

nmoinvaz:
I forgot about that. We do cross compile so it may be possible to deliver binaries, but we don't run any tests because there is no Windows ARM platform in CI.

mtl1979:
AARCH64 compiler is still unreliable for Windows... I have been working with Microsoft for long before it was introduced. There is some cases where ARM and AARCH64 compilers won't trigger error even when they should and some intrinsics are still missing.

nmoinvaz:
This should be resolved for the next release.

PR #916, opened 2021-03-26
Added Windows ARM and ARM64 binaries in releases. 

nmoinvaz:
See #909.
For an example of the output:
https://github.com/nmoinvaz/zlib-ng/releases/tag/1.2.0.0


Dead2 (approved):
Looks good. Did you check that these produce seemingly correct files in a release?

nmoinvaz:
Yes I did. The link is in the description. But I don't have a Windows ARM machine to verify it completely.

PR #918, opened 2021-03-26
Check support for ARM float-abi compiler flag auto-detected value. 

nmoinvaz:
See #911. Here we check to see that the compiler supports our auto-detected floating point ABI flag support before using it.


Dead2:
Unfortunately this fails the Configure GCC ARM SF test, but it looks like you might be close to a working solution.
Btw, all your current PRs are based on an old commit. :)

ggardet:
Maybe you could do the hard/soft test out of the gnueabihf/gnueabi condition.

nmoinvaz:
@Dead2 that CI test is resolved now.

configure
```diff
@@ -1252,15 +1252,30 @@ case "${ARCH}" in
...
-                floatabi="-mfloat-abi=softfp" ;;
+                if try $CC -c $CFLAGS $test.c -mfloat-abi=softfp; then
+                  floatabi="-mfloat-abi=softfp"
+                fi
+            ;;
         esac
-        CFLAGS="${CFLAGS} ${floatabi}"
-        SFLAGS="${SFLAGS} ${floatabi}"
+        if [ -z $floatabi ]; then
+          echo "ARM floating point arch not auto-detected" | tee -a configure.log
```
Dead2: Perhaps this should be a warning/error instead?
Is this a "valid" state (assuming 100% working soft/hard-float detection)?

nmoinvaz: I would rather it just proceed normally instead of full stop but I'm not an ARM expert.

mtl1979: Earlier I was thinking it should try both versions if the last part of the triplet doesn't mention which one the compiler should support and then finally revert to having NEON optimizations disabled.

Last time I checked in Linaro toolkit for ARM, compiler that only supports either soft or hard float, has only one version of the internal headers, but it is possible to have compiler that has both versions of internal headers.

Most newer processors should use soft float, but older processors might work with hard float. NEON is basically overlapping set of older vector instruction set which is supported by original ARM compiler for Visual C++.

nmoinvaz: I could add an additional check for hard float-abi when it is `.*gnueabi` but do we even want to be doing detection based on triplet? We can just check for both and the first one it finds then it uses it.

mtl1979: As far as I know having "hf" in the triplet is kind of Debianism... Not depending on it makes it possible to support other Linux versions better... As I said, hard float is more common on ARMv7 and soft float is more common on ARMv8. We (should) support ARMv6 upwards when vector optimizations are enabled.

nmoinvaz: I have removed the triplet check.

Dead2:
Looks good to me, care to rebase and squash?

nmoinvaz:
Done.

Dead2:
I see macos started failing again, didn't we already fix this?

nmoinvaz:
No we did not. I did the research in our team chat but never finished tracking it down.

Dead2:
You are right, I was thinking about that discussion and probably thought of the win32 wine fixes as well.

CI in the cloud sure is problematic at times.. (due to changes they implement where we have no control and/or warning)

PR #921, opened 2021-03-27
Remove unnecessary malloc.h header include

nmoinvaz:
This comes out of the discussion from [#766](https://github.com/zlib-ng/zlib-ng/pull/766/files), where we discovered it doesn't make sense to have wrapped malloc.h include in `ifndef UNALIGNED_OK`. `malloc()` is already included from `stdlib.h` include.

### DIFF
--- a/test/fuzz/minigzip_fuzzer.c
+++ b/test/fuzz/minigzip_fuzzer.c
@@ -32,10 +32,6 @@
 #  include <sys/stat.h>
 #endif
 
-#ifndef UNALIGNED_OK
-#  include <malloc.h>
-#endif
-
 #if defined(_WIN32) || defined(__CYGWIN__)
 #  include <fcntl.h>
 #  include <io.h>
--- a/test/minigzip.c
+++ b/test/minigzip.c
@@ -32,10 +32,6 @@
 #  include <sys/stat.h>
 #endif
 
-#ifndef UNALIGNED_OK
-#  include <malloc.h>
-#endif
-
 #if defined(_WIN32) || defined(__CYGWIN__)
 #  include <fcntl.h>
 #  include <io.h>


iii-i:
```
malloc() is already included from stdlib.h include.
```

I vaguely remember seeing some situations where this wasn't the case on MSVC. And their doc still says:

https://docs.microsoft.com/de-de/cpp/c-runtime-library/reference/malloc?view=msvc-160

```
Routine: malloc
Required header: <stdlib.h> and <malloc.h>
```

nmoinvaz:
I compile against MSVC and haven't run into a problem. But even so, it would have failed for who didn't have `UNALIGNED_OK` set in the original code.

mtl1979:
I have all supported versions of MSVC except 2013 installed on my development PC... If something is going to break with MSVC, most likely it will be with oldest supported versions... Older versions didn't include Windows SDK by default, so there was basically two sets of system headers, one with less fixes and one that was regularly updated.

iii-i:
You are right, making `# include <malloc.h>` depend on `UNALIGNED_OK` makes no sense. It should either be unconditionally paired with `#include <stdlib.h>` or be skipped altogether.

mtl1979:
In recent days our use of memory allocation functions and malloc.h has changed drastically, so a lot of code needs to be re-evaluated. If malloc.h is still needed, we need to know exact compiler versions that still need including it and adjust the code accordingly... Unconditionally including it is unacceptable workaround.

nmoinvaz:
Rebased.

nmoinvaz:
Actually on macOS I did get an error because of malloc.h includes when compiling with `-DUNALIGNED_OK=OFF`. `__APPLE__` uses `<stdlib.h>`.
```
[ 82%] Building C object CMakeFiles/minigzip.dir/test/minigzip.c.o
/Users/runner/work/zlib-ng/zlib-ng/test/minigzip.c:36:12: fatal error: malloc.h: No such file or directory
   36 | #  include <malloc.h>
      |            ^~~~~~~~~~
```

Issue #925, opened 2021-04-08
SIGBUS error while inflate on armv7


githublulz:
Hello,
I got a SIGBUS error on Android device when inflating compressed data. Any ideas ? 
Original zlib inflate function work fine with the same code (without zng_ prefix). Deflate function work also fine with zlib-ng.

Static libs are built with Android NDK with following args :
```
cmake -B${ABI_BUILD_DIR} \
        -DANDROID_ABI=armeabi-v7a \
	-DWITH_GZFILEOP=OFF \
	-DCMAKE_C_FLAGS="-fdata-sections -ffunction-sections" \
        -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY=${STAGING_DIR}/lib/${ABI} \
        -DANDROID_PLATFORM=android-${MINIMUM_API_LEVEL} \
        -DZLIB_ENABLE_TESTS=OFF \
	-DCMAKE_BUILD_TYPE=Release \
	-DANDROID_TOOLCHAIN=clang \
	-DANDROID_STL=c++_shared \
	-DCMAKE_TOOLCHAIN_FILE=${ANDROID_NDK}/build/cmake/android.toolchain.cmake \
	-DCMAKE_INSTALL_PREFIX=."
```

Stack-trace : 
```
libc    : Fatal signal 7 (SIGBUS), code 1 (BUS_ADRALN), fault addr 0xff9cf895 in tid 16771 (binary), pid 16771 (binary)
04-08 19:45:11.273 16777 16777 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
04-08 19:45:11.273 16777 16777 F DEBUG   : Build fingerprint: 'samsung/a40eea/a40:9'
04-08 19:45:11.273 16777 16777 F DEBUG   : Revision: '4'
04-08 19:45:11.273 16777 16777 F DEBUG   : ABI: 'arm'
04-08 19:45:11.273 16777 16777 F DEBUG   : pid: 16771, tid: 16771, name: binary  >>> ./binary <<<
04-08 19:45:11.273 16777 16777 F DEBUG   : signal 7 (SIGBUS), code 1 (BUS_ADRALN), fault addr 0xff9cf895
04-08 19:45:11.274 16777 16777 F DEBUG   :     r0  00000001  r1  00000002  r2  0000002f  r3  ae393ac1
[...]

PR #927, opened 2021-04-10
[ARM] Use temporary variable when loading more than 8 bits in chunkmemset_neon().

mtl1979:
* ARMv7 does not support unaligned loads to vector registers from memory wider than 8 bits.

See #925.

### DIFF
--- a/arch/arm/chunkset_neon.c
+++ b/arch/arm/chunkset_neon.c
@@ -23,11 +23,15 @@ static inline void chunkmemset_1(uint8_t *from, chunk_t *chunk) {
 }
 
 static inline void chunkmemset_2(uint8_t *from, chunk_t *chunk) {
-    *chunk = vreinterpretq_u8_s16(vdupq_n_s16(*(int16_t *)from));
+    int16_t tmp;
+    memcpy(&tmp, from, 2);
+    *chunk = vreinterpretq_u8_s16(vdupq_n_s16(tmp));
 }
 
 static inline void chunkmemset_4(uint8_t *from, chunk_t *chunk) {
-    *chunk = vreinterpretq_u8_s32(vdupq_n_s32(*(int32_t *)from));
+    int32_t tmp;
+    memcpy(&tmp, from, 4);
+    *chunk = vreinterpretq_u8_s32(vdupq_n_s32(tmp));
 }
 
 static inline void chunkmemset_8(uint8_t *from, chunk_t *chunk) {


arch/arm/chunkset_neon.c
```diff
@@ -23,11 +23,15 @@ static inline void chunkmemset_1(uint8_t *from, chunk_t *chunk) {
 }
 
 static inline void chunkmemset_2(uint8_t *from, chunk_t *chunk) {
-    *chunk = vreinterpretq_u8_s16(vdupq_n_s16(*(int16_t *)from));
+    int16_t tmp;
```
nmoinvaz, line 26: Consider this instead:
```
int16_t tmp;
#if defined(UNALIGNED_OK)
    tmp = *(int16_t*)from;
#else
    memcpy(&tmp, from, 2);
#endif
*chunk = vreinterpretq_u8_s16(vdupq_n_s16(tmp));
```
Would need something similar for `chunkmemset_4`.

mtl1979: `UNALIGNED_OK` is wrong test as it is used for copies between memory and general-purpose registers, not copies between memory and vector registers... NEON is not the only vector instruction set that causes exception with unaligned loads and stores wider than 8 bits, even SSE has restriction even though it has unaligned version for some instructions.

nmoinvaz: It is still unaligned read/load regardless of source/destination right? I think `UNALIGNED_OK` should cover both unaligned support for intrinsic and also architecture unaligned support.

mtl1979: @nmoinvaz That would be seriously bad idea... For the same reason we introduced separate define for 64-bit integers.

nmoinvaz: I don't get it but it's ok.

mtl1979: @nmoinvaz 

1. Disabling `UNALIGNED_OK` would unnecessarily slow down irrelevant parts of the code
2. It would redefine purpose of `UNALIGNED_OK` the second time
3. Adding another define for just one part of the code is premature
4. Point of using memcpy is to forbid the code optimizer to optimize away the temporary variable as "register" keyword is now deprecated.

nmoinvaz: Good explanation I got it now thank you. My only suggestion then is to put in the commit message that `memcpy` is being used to prevent temporary variable from being optimized away.

mtl1979: @nmoinvaz Using memcpy() to alias pointer types of different bit width has been used elsewhere already... 

It usually works better than shifts due to some platforms being both big and little endian or using floating point registers for integer vector elements (vector registers are by default in floating point mode). 

Optimizer of gcc is also smart enough to optimize away redundant byte swaps using shifts.

mtl1979:
This is how the disassembly of the two relevant functions look now with gcc trunk:
```
chunkmemset_2(unsigned char*, __simd128_uint8_t*):
 ldrh r3, [r0, #0]
 vdup.16 q8, r3
 vst1.64 {d16-d17}, [r1 :64]
 bx lr
chunkmemset_4(unsigned char*, __simd128_uint8_t*):
 ldr r3, [r0, #0]
 vdup.32 q8, r3
 vst1.64 {d16-d17}, [r1 :64]
 bx lr
```

`bx lr` is same as `ret` on x86 assembly. `r0` is `from`, `r1` is `chunk` and `r3`is `tmp`. I disabled inlining so the whole function doesn't get optimized away.

nmoinvaz (approved):
Assembly looks exactly like what we would want. Description in commit has good explanation. Nice job!

kdrag0n (approved):
This fixes the issue for me without changing the generated code on 64-bit ARMv8.

PR #929, opened 2021-04-15
Improvements to minigzip fuzzer coverage

nmoinvaz:
The first commit I double checked my assumptions here:
https://jsfiddle.net/use0f6yv/
But using the same random input value to determine both values will miss some combinations.


mtl1979:
Obviously modulo without division didn't work, this is same as using "binary and" without bit shift.

When extracting two or more numbers from same memory location, the bit indices for each number must not overlap.

Dead2:
What is the status of this?

nmoinvaz:
It is ready.

PR #931, opened 2021-04-15
Use __attribute__ (constructor) for platform check features

kitaisreal:
Currently calls to check_features are not thread safe, they produce errors under thread sanitizer. Use __attribute__ (constructor) for platform check features.

### DIFF
--- a/arch/arm/arm.h
+++ b/arch/arm/arm.h
@@ -8,6 +8,4 @@
 extern int arm_cpu_has_neon;
 extern int arm_cpu_has_crc32;
 
-void Z_INTERNAL arm_check_features(void);
-
 #endif /* ARM_H_ */
--- a/arch/arm/armfeature.c
+++ b/arch/arm/armfeature.c
@@ -59,7 +59,7 @@ static inline int arm_has_neon() {
 Z_INTERNAL int arm_cpu_has_neon;
 Z_INTERNAL int arm_cpu_has_crc32;
 
-void Z_INTERNAL arm_check_features(void) {
+static void __attribute__((constructor)) arm_check_features(void) {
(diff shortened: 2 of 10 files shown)


NiLuJe:
`constructor` functions are optimized out of static libraries by default (because nothing references them), so this'd break platform detection on static builds.

mtl1979:
We also support other compilers than gcc, where attributes are declared differently and also have different set of supported attributes.

Dead2:
Why would this be needed though? zlib/zlib-ng are not thread-safe.

mtl1979:
@Dead2 Some parts of zlib-ng are thread-safe, but not all... Basically you can't compress or decompress from multiple threads using same buffers, but you can use different buffers.

Dead2:
So much for trusting what a random person on the internet said. I checked myself, and zlib does indeed promise to be thread safe.
So yeah, I guess we do need to solve this. But the proposed solution in this PR is not an universal solution we can use.

nmoinvaz:
I was thinking perhaps we include functable in deflate_struct and inflate_struct instead of as a global variable. It would incur a minor cost for each instance but at least then it would be safer. This would also play nicely with fast-zlib changes which change the hashing functions based on compression level.

mtl1979:
@nmoinvaz Whole point of func_table being global variable is so that crc32() and adler32() can also be used outside deflate/inflate.

Dead2:
@kitaisreal Would you please run a check of Develop and see whether this is still necessary?

kitaisreal:
@Dead2 we use this __attribute__ (constructor) and there are no more issues from thread sanitizer from our CI system. I am not sure if this can be ported for Windows.

PR #936, opened 2021-04-24
Don't directly include asm/hwcap.h; fix compilation on musl aarch64

joshtriplett:
sys/auxv.h includes the appropriate headers to provide the HWCAP
constants, on both glibc and musl, which makes it unnecessary to include
asm/hwcap.h directly. And on musl, asm/hwcap.h doesn't exist.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -640,6 +640,34 @@ endif()
 if(WITH_OPTIM)
     if(BASEARCH_ARM_FOUND)
         add_definitions(-DARM_FEATURES)
+        if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+            check_c_source_compiles(
+                "#include <sys/auxv.h>
+                int main() {
+                    return (getauxval(AT_HWCAP2) & HWCAP2_CRC32);
+                }"
+                ARM_AUXV_HAS_CRC32
+            )
+            if(ARM_AUXV_HAS_CRC32)
+                add_definitions(-DARM_AUXV_HAS_CRC32)
+            else()
+                message(STATUS "HWCAP2_CRC32 not present in sys/auxv.h; cannot detect support at runtime.")
+            endif()
+            if(NOT "${ARCH}" MATCHES "aarch64")
+                check_c_source_compiles(
+                    "#include <sys/auxv.h>
+                    int main() {
+                      return (getauxval(AT_HWCAP) & HWCAP_NEON);
+                    }"
+                    ARM_AUXV_HAS_NEON
+                )
+                if(ARM_AUXV_HAS_NEON)
+                    add_definitions(-DARM_AUXV_HAS_NEON)
+                else()
+                    message(STATUS "HWCAP_NEON not present in sys/auxv.h; cannot detect support at runtime.")
+                endif()
+            endif()
+        endif()
         list(APPEND ZLIB_ARCH_HDRS ${ARCHDIR}/arm.h)
         list(APPEND ZLIB_ARCH_SRCS ${ARCHDIR}/armfeature.c)
         if(WITH_ACLE AND NOT MSVC)
--- a/arch/arm/armfeature.c
+++ b/arch/arm/armfeature.c
@@ -2,7 +2,6 @@
 
 #if defined(__linux__)
 #  include <sys/auxv.h>
-#  include <asm/hwcap.h>
 #elif defined(__FreeBSD__) && defined(__aarch64__)
 #  include <machine/armreg.h>
 #  ifndef ID_AA64ISAR0_CRC32_VAL
--- a/arch/arm/armfeature.c
+++ b/arch/arm/armfeature.c
@@ -14,7 +14,7 @@
 #endif
 
 static int arm_has_crc32() {
-#if defined(__linux__) && defined(HWCAP2_CRC32)
+#if defined(__linux__) && defined(ARM_AUXV_HAS_CRC32)
     return (getauxval(AT_HWCAP2) & HWCAP2_CRC32) != 0 ? 1 : 0;
 #elif defined(__FreeBSD__) && defined(__aarch64__)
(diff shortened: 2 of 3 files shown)


mtl1979:
We explicitly include some redundant headers to trap differences between compilers. We don't officially support every combination there is (compiler + platform + C run-time library).

joshtriplett:
@mtl1979 In this case, this isn't a difference between compilers, it's a difference between musl and glibc; musl doesn't seem to support directly including the kernel headers here. It'd be helpful to have support for aarch64-linux-musl, for the purposes of building statically compiled binaries for aarch64.

I've tested that with this one change included, it's possible to compile zlib-ng for an aarch64-unknown-linux-musl target.

mtl1979:
@joshtriplett I assumed musl and glibc are both C run-time libraries... GCC usually use glibc by default, Visual C++ use its own C libary, etc...

In my personal opinion, we should support combinations that enough people use... This excludes cases that are complex to include without causing possible conflicts, including duplicate names, or names that are used for two different purposes.

Dead2:
As long as this does not trigger any regressions, I honestly don't see any problem with including the change. Supporting musl would be nice, even if it is not something we target or spend time on, especially if it is this simple.

joshtriplett:
Both glibc and musl seem to most commonly use gcc as their compiler (though both can be compiled with clang as well).

I completely agree that not all combinations are reasonable to support. I'm hoping that since aarch64-linux-musl works with just this one line changed, this would be OK to add.

mtl1979:
@joshtriplett Changing header filename is usually easier to accept than removing one header... This is because we assume that breaking change in one specific header file is likely break build of other projects than zlib-ng too and backfire more likely.

Personally, I have worked quite a while with Linaro gcc team and Microsoft, so I have had to fix quite a lot of header related issues.

joshtriplett:
Can you clarify what problem you're concerned about with this change?

Would you prefer that I wrap this in an `#ifndef` so that it includes the header when not on musl?

mtl1979:
@Dead2 It's not about triggering regression, it's about opening Pandora's box... Basically this is same as Microsoft moving some intrinsics to combined header and restricting their use to only AArch64, even though ARMv8 has supported them for both 32-bit and 64-bit... In some untested configurations, hard fail is better than unknown behavior.

@joshtriplett If we remove the header check for all C run-time library possibilities, we basically assert that we support all possible C run-time libraries even if they are not API compatible with glibc. I'm not sure if we should do the existence check on configure step or in just this source file. I'm inclined to accept both solutions, but it's just my opinion.

joshtriplett:
@mtl1979 Ah, I see.

Supporting *just* "glibc and musl" rather than "every possible libc" seems reasonable. I do think that'd be better done using cmake detection and compilation tests, rather than by including headers like this that exist in some implementations and not in others.

As far as I can tell, musl intentionally doesn't provide any musl-specific defines, so there's no way to detect musl in this context and only stop including the header in that case. Given that, would you consider it reasonable to drop this particular header include to fix musl?

mtl1979:
@joshtriplett I think the best solution is to move the conformity test to CMakeLists.txt and configure as build test... Basically we only need to make sure the function itself and all the used defines are declared correctly. This allows us to also trap cases when someone tries to use too old version of C run-time library, for example on Android.

joshtriplett:
@mtl1979 When you say "all the used defines", do you mean the `HWCAP_*` and `HWCAP2_*` defines? As far as I can tell, both of the used defines are checked with `defined(...)` around the code that uses them, presumably to prevent detecting them on systems that don't define those hwcaps. Are you wanting to drop that code, and prevent compilation on platforms that don't have those defines available? Or am I not understanding what you're proposing?

nmoinvaz:
I think it is fine also as long as it doesn't cause regression.

mtl1979:
@joshtriplett Failing as early as possible is the best... If I remember correctly ARM and AArch64 have slightly different names for the defines, so one has to check for the correct ones depending on if the target system is 32-bit or 64-bit.

joshtriplett:
I've added an additional commit that adds compile-time detection to make sure we have the necessary hwcap defines available to do runtime detection. That way, if there's a missing definition, the user will get a message from cmake when doing initial configuration, rather than a compile-time warning or silently missing detection at runtime.

I hope this addresses the concern; now, the code will compile with more C libraries, but if the C library doesn't provide the necessary definitions, the user will get a useful message from cmake.

mtl1979:
@joshtriplett The second commit fixes it for cmake, but not when using configure...

joshtriplett:
Pending CI confirmation, hopefully the third commit will do the same for configure.

joshtriplett:
Looks like it passed CI. So, this should cover both configure and cmake, and provide better configuration-time detection of whether runtime detection will work.

Dead2:
Looks pretty good to me, but I find the defines are a bit long and awkward `ARM_HAVE_CRC32_IN_AUXV`
What about simply `ARM_AUXV_CRC32`?

mtl1979:
@Dead2 I agree that the defines could be as short as possible while still being descriptive... Should definitely follow the same pattern as existing ones. 

It's also matter of debate if the commits should be squashed before merging.

nmoinvaz:
Well for includes we have always started them with `HAVE_*` but for architecture features it always started with `ARM_*`.

joshtriplett:
I can certainly modify the names. I do think it makes sense to put the `AUXV` before the hwcap name, so that the names have the same prefix. Something like `ARM_AUXV_HAS_CRC32` and `ARM_AUXV_HAS_NEON`, perhaps?

As for squashing, please feel free to squash-merge if you prefer, or if you'd like, I can rebase into 2 commits or 1 commit. Let me know what you'd prefer and I'd be happy to structure it accordingly.

joshtriplett:
I pushed a new version which combines the cmake and configure commits, and which renames the definitions.

Issue #722, opened 2020-08-17
arm-unknown-linux-gnueabi failure: "You must enable NEON instructions (e.g. -mfloat-abi=softfp -mfpu=neon) to use arm_neon.h"


joshtriplett:
When trying to do a build (via CI) on arm-unknown-linux-gnueabi:

```
In file included from /project/src/zlib-ng/arch/arm/adler32_neon.c:11:0:
/usr/lib/gcc-cross/arm-linux-gnueabi/5/include/arm_neon.h:31:2: error: #error You must enable NEON instructions (e.g. -mfloat-abi=softfp -mfpu=neon) to use arm_neon.h
 #error You must enable NEON instructions (e.g. -mfloat-abi=softfp -mfpu=neon) to use arm_neon.h
```
(Followed by a cascade of errors about not finding SIMD types and functions.)

nmoinvaz:
Do you have a simplified GitHub Actions workflow that this can be reproduced in?

nmoinvaz:
The detection in CMake has been improved since this was opened. Is it possible to retry with the latest version?

joshtriplett:
I can no longer reproduce this with current zlib-ng.

PR #938, opened 2021-04-24
Fix build of optimized functions with LTO

viccie30:
When building zlib-ng with LTO enabled, at least GCC doesn't respect the flags of each source file and just throws all the flags together. See https://gcc.gnu.org/onlinedocs/gccint/LTO-object-file-layout.html.

Because of this, flags don't get applied correctly to the arch-specific files, leading to problems like #722.

CMake does not allow disabling LTO per source file, so in `CMakeLists.txt` I just added `-fno-lto` if the compiler is GCC-like and native mode is not enabled.
For the configure script I wrote the same logic, but using an extra variable in `arch/*/Makefile.in`.


viccie30:
I tested this patch using `sbuild` on all supported Debian architectures (amd64, arm64, armel, armhf, i386, mips64el, mipsel, ppc64el, and s390x) on stable, testing, unstable, and experimental, both with and without LTO and for static and shared builds. 

For all combinations, except mips64el and mipsel on experimental due to dependency breakage, the builds succeed and all tests pass (using `qemu-user-static` for the non-x86 architectures).

Dead2:
I think this override should perhaps not be active if the user enables Native mode, native is obviously useless for redistribution such as in distros, but it can be useful for those who compile locally and want the (possibly) fastest end result.

When native mode is enabled, all files are compiled with the same flags, essentially enabling all instruction sets the local cpu supports, so that should allow it to work with LTO.

viccie30:
I agree, that's why both in `configure` and `CMakeLists.txt` the no lto flag is only defined if native mode is disabled

Dead2:
Hah, indeed it is, I didn't notice that. 👍

viccie30:
It took me a few tries to get that right as well, but I think I did the right way round

Dead2 (approved):

PR #939, opened 2021-04-29
Decompress gzip compressed archive to a separate file to prevent corruption issues on macOS.

nmoinvaz:
I'm not 100% sure why this works but it appears to on over 40,000 runs. The steps are run sequential but some how on macOS writing to the same content to the same filename gets corrupted.


Dead2 (approved):
LGTM

PR #942, opened 2021-05-02
Fix cmake detection of risc-v (32 and 64 bit)

### LINKED ISSUE #941: riscv64 support in cmake is broken
Currently if architecture is not detected it falls back to x86, with SSE2 and other stuff enabled, which is incorrect for riscv64.

I'm not sure what would be the best course of actions here - either to fallback to generic or add riscv64 (or maybe both) though as I don't know what were intentions behind fallback to x86.

Civil:
Add a case to detect risc-v architectures.

Fixes #941


cmake/detect-arch.cmake
```diff
@@ -85,6 +85,12 @@ elseif("${ARCH}" MATCHES "parisc")
 elseif("${ARCH}" MATCHES "rs6000")
     set(BASEARCH "rs6000")
     set(BASEARCH_RS6000_FOUND TRUE)
+elseif("${ARCH}" MATCHES "riscv32")
+    set(BASEARCH "riscv32")
+    set(BASEARCH_RISCV32_FOUND TRUE)
+elseif("${ARCH}" MATCHES "riscv64")
+    set(BASEARCH "riscv64")
+    set(BASEARCH_RISCV64_FOUND TRUE)
```
Dead2: These should likely have a common basearch, like `risc` or `riscv` or something like that.
Like basearch `x86` has both 32 and 64bit, and same for basearch `arm`.

Having a common basearch is useful because often the same feature/instruction-set/etc tests are common, or they can even use much of the same accelerated code.

Civil: Should be fixed in recent force-push.

nmoinvaz (approved):
LGTM

Dead2:
Detection of the architecture as 32- or 64-bit is still useful, so I really didn't want you to remove that (from the .c file). Only basearch should be the common for the two.
Would you please revert the change to the .c file (and modify the if in .cmake)? 👍

You can have a look at how the x86 detection is done, with differing arches but also having a common basearch. Both are used in CMakeLists.txt for different things.

Civil:
@Dead2 done.

I've also modified it a bit, as after a second thought, detecting riscv64 by pointer size seems to be wrong, and by https://github.com/riscv/riscv-toolchain-conventions correct way should be to check for `__riscv_xlen`

Dead2:
@Civil Are you sure you pushed the update? It looks unchanged to me. But not entirely sure whether github is messing with me or not.

cmake/detect-arch.c
```diff
@@ -93,6 +93,10 @@
 #elif defined(__THW_RS6000)
     #error archfound rs6000
 
+// RISC-V
+#elif defined(__riscv)
+    #error archfound riscv
+
```
mtl1979, line 102: This is oversimplified.

Civil:
@Dead2 indeed, sorry about that.

Dead2 (approved):

nmoinvaz (approved):

cmake/detect-arch.c
mtl1979: Test for `__riscv` should be outer test like in the previous examples when two related preprocessor macros are checked...
Test for `__riscv_xlen` should be inner test, essentially outputting nothing if the value is neither `32` or `64`.

Civil: Since your previous comment I've pushed another commit and currently check looks like that:
```
#elif defined(__riscv) && __riscv_xlen == 64
    #error archfound riscv64
#elif defined(__riscv) && __riscv_xlen == 32
    #error archfound riscv32
```

And it esentially do what you say, however as two checks, instead of one. If you think that it would be cleaner, I can obviously rewrite it as:
```
#elif defined(__riscv)
    #if __riscv_xlen == 64
        #error archfound riscv64
    #elif __riscv_xlen == 32
        #error archfound riscv32
```

mtl1979: I think the style should be consistent with the other similar checks in the same file...

Civil: I've force-pushed commit again. Please have a look if it is better now.

mtl1979 (approved):

PR #949, opened 2021-05-06
upgrade links to HTTPS

vszakats:
> http://infozip.sourceforge.net/ is sadly not having HTTPS access enabled. Shoutout to somebody with admin access for this > project: It would be nice to enable it and thus allowing secure access to these pages via https://infozip.sourceforge.io/ . The > option has been there [for a while now](https://sourceforge.net/blog/introducing-https-for-project-websites/).

also:
- follow permanent redirects
- add ending slashes


vszakats:
Note: This patch doesn't involve code, so 'CodeFactor' failures are all pre-existing issues and can be disregarded.

Dead2 (approved):
LGTM

PR #950, opened 2021-05-08
Don't check for SSSE3 on non-x86 architectures.

Dead2:
The SSSE3 test had no architecture check like all the others, this adds that.

### DIFF
--- a/configure
+++ b/configure
@@ -994,8 +994,9 @@ EOF
 esac
 
 # Check for SSSE3 intrinsics
-
-cat > $test.c << EOF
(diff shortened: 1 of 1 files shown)


mtl1979 (approved):

nmoinvaz:
This is already resolved in #906.

Dead2:
@nmoinvaz True, but that is a significantly bigger change, and I was hoping to get this simple fix into 2.0.3. What do you think?

nmoinvaz:
That is ok.

PR #828, opened 2021-01-18
Incorporate fast-zlib changes.

nmoinvaz:
This is a PR that seeks to incorporate the fast zlib changes mentioned in #97. Fast-zlib changes require rolling hash to be used. It also performs best when hash mask is 32k-1 instead of 64k-1 which is used for integer hashing method we currently use. 

I don't expect these changes to be incorporated into the first release. We might also decide to not incorporate them at all due to the added complexity. But this is the first step toward determining that. And in the future should anybody want to investigate this again they can use this PR as a starting base.

# Benchmarks

## MADLER

```
 Tool: minigzip-madler.exe Levels: 1-9
 Runs: 70         Trim worst: 40

 Level   Comp   Comptime min/avg/max/stddev  Decomptime min/avg/max/stddev  Compressed size
 1     36.448%      2.421/2.433/2.442/0.006        0.907/0.919/0.925/0.004       77,255,275
 2     35.381%      2.675/2.689/2.696/0.006        0.897/0.907/0.916/0.006       74,991,864
 3     34.426%      3.306/3.321/3.332/0.007        0.880/0.900/0.908/0.008       72,967,657
 4     33.498%      3.540/3.556/3.566/0.008        0.876/0.891/0.896/0.006       71,002,237
 5     32.627%      4.840/4.858/4.872/0.007        0.876/0.888/0.892/0.004       69,156,120
 6     32.188%      6.789/6.822/6.842/0.015        0.870/0.880/0.886/0.004       68,224,174
 7     32.051%      8.181/8.214/8.234/0.013        0.865/0.877/0.882/0.005       67,934,314
 8     31.935%   12.677/12.719/12.745/0.019        0.863/0.872/0.877/0.003       67,688,380
 9     31.916%   16.200/16.248/16.282/0.022        0.861/0.872/0.877/0.004       67,647,801

 avg1  33.385%                        6.762                          0.889
 tot                               1825.836                        240.135      636,867,822
 ```

 ## ZLIB-NG

 ```
 Tool: minigzip-ng.exe Levels: 1-9
 Runs: 70         Trim worst: 40

 Level   Comp   Comptime min/avg/max/stddev  Decomptime min/avg/max/stddev  Compressed size
 1     47.618%      1.840/1.859/1.865/0.006        0.681/0.688/0.693/0.003      100,930,315
 2     35.519%      2.050/2.065/2.074/0.007        0.703/0.715/0.719/0.005       75,286,316
[...]

PR #953, opened 2021-05-10
Separate MIN_MATCH into properly separated values.

Dead2:
- STD_MIN_MATCH is 3, and is used to stay compatible with the deflate standards.
- WANT_MIN_MATCH is set to 4, and is used to indicate to the code what the minimum match length we intend to accept is.

Having WAN_MIN_MATCH set to 4 also means that we now skip the last round of longest_match in the sliding window. This is good because we currently hash 4 bytes, so checking for a match when there is only 3 bytes left is pointless.

The performance difference with this PR appears to be better (as it should), but the difference is slight (around +0.05-0.10% faster), so it is within the margin of error.

This PR also removes the partial, dead, broken and untested support for STD_MIN_MATCH != 3.
As far as I understand, setting this to something else would cause the output data to deviate from the deflate standard and produce an incompatible output stream, unless the tree tables are also regenerated to match the changed offset. And STD_MAX_MATCH would probably have to be reduced by the same number as well. I am not 100% certain of this, but I think it is easier to experiment with this kind of thing now that the variables are split.

I am not 100% certain about the correctness, but the current iteration makes sense to me and the compressed output is identical to pre-PR in my tests.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -443,7 +443,7 @@ int32_t Z_EXPORT PREFIX(deflateSetDictionary)(PREFIX3(stream) *strm, const uint8
     s->block_start = (int)s->strstart;
     s->insert = s->lookahead;
     s->lookahead = 0;
-    s->prev_length = STD_MIN_MATCH - 1;
+    s->prev_length = 0;
     s->match_available = 0;
     strm->next_in = (z_const unsigned char *)next;
     strm->avail_in = avail;
@@ -1160,7 +1160,7 @@ static void lm_init(deflate_state *s) {
     s->block_start = 0;
     s->lookahead = 0;
     s->insert = 0;
-    s->prev_length = STD_MIN_MATCH - 1;
+    s->prev_length = 0;
     s->match_available = 0;
     s->match_start = 0;
 }
--- a/deflate.c
+++ b/deflate.c
@@ -167,7 +167,7 @@ static const config configuration_table[10] = {
 /* 8 */ {32, 128, 258, 1024, deflate_slow},
 /* 9 */ {32, 258, 258, 4096, deflate_slow}}; /* max compression */
 
-/* Note: the deflate() code requires max_lazy >= MIN_MATCH and max_chain >= 4
+/* Note: the deflate() code requires max_lazy >= STD_MIN_MATCH and max_chain >= 4
  * For deflate_fast() (levels <= 3) good is ignored and lazy has a different
  * meaning.
  */
@@ -431,19 +431,19 @@ int32_t Z_EXPORT PREFIX(deflateSetDictionary)(PREFIX3(stream) *strm, const uint8
     strm->avail_in = dictLength;
     strm->next_in = (z_const unsigned char *)dictionary;
     fill_window(s);
-    while (s->lookahead >= MIN_MATCH) {
+    while (s->lookahead >= STD_MIN_MATCH) {
         str = s->strstart;
-        n = s->lookahead - (MIN_MATCH-1);
+        n = s->lookahead - (STD_MIN_MATCH-1);
         functable.insert_string(s, str, n);
         s->strstart = str + n;
-        s->lookahead = MIN_MATCH-1;
+        s->lookahead = STD_MIN_MATCH-1;
         fill_window(s);
     }
     s->strstart += s->lookahead;
     s->block_start = (int)s->strstart;
     s->insert = s->lookahead;
     s->lookahead = 0;
-    s->prev_length = MIN_MATCH-1;
+    s->prev_length = STD_MIN_MATCH-1;
     s->match_available = 0;
     strm->next_in = (z_const unsigned char *)next;
     strm->avail_in = avail;
@@ -1160,7 +1160,7 @@ static void lm_init(deflate_state *s) {
     s->block_start = 0;
     s->lookahead = 0;
     s->insert = 0;
-    s->prev_length = MIN_MATCH-1;
+    s->prev_length = STD_MIN_MATCH-1;
     s->match_available = 0;
     s->match_start = 0;
 }
@@ -1224,17 +1224,16 @@ void Z_INTERNAL fill_window(deflate_state *s) {
         s->lookahead += n;
 
         /* Initialize the hash value now that we have some input: */
-        if (s->lookahead + s->insert >= MIN_MATCH) {
+        if (s->lookahead + s->insert >= STD_MIN_MATCH) {
             unsigned int str = s->strstart - s->insert;
             if (str >= 1)
-                functable.quick_insert_string(s, str + 2 - MIN_MATCH);
-#if MIN_MATCH != 3
-#error Call insert_string() MIN_MATCH-3 more times
+                functable.quick_insert_string(s, str + 2 - STD_MIN_MATCH);
+#if STD_MIN_MATCH != 3
             while (s->insert) {
                 functable.quick_insert_string(s, str);
                 str++;
                 s->insert--;
-                if (s->lookahead + s->insert < MIN_MATCH)
+                if (s->lookahead + s->insert < STD_MIN_MATCH)
                     break;
             }
 #else
@@ -1250,7 +1249,7 @@ void Z_INTERNAL fill_window(deflate_state *s) {
             }
 #endif
         }
-        /* If the whole input has less than MIN_MATCH bytes, ins_h is garbage,
+        /* If the whole input has less than STD_MIN_MATCH bytes, ins_h is garbage,
          * but this is not important since only literal bytes will be emitted.
          */
     } while (s->lookahead < MIN_LOOKAHEAD && s->strm->avail_in != 0);
@@ -1259,8 +1258,8 @@ void Z_INTERNAL fill_window(deflate_state *s) {
      * written, then zero those bytes in order to avoid memory check reports of
      * the use of uninitialized (or uninitialised as Julian writes) bytes by
      * the longest match routines.  Update the high water mark for the next
-     * time through here.  WIN_INIT is set to MAX_MATCH since the longest match
-     * routines allow scanning to strstart + MAX_MATCH, ignoring lookahead.
+     * time through here.  WIN_INIT is set to STD_MAX_MATCH since the longest match
+     * routines allow scanning to strstart + STD_MAX_MATCH, ignoring lookahead.
      */
     if (s->high_water < s->window_size) {
         unsigned int curr = s->strstart + s->lookahead;
@@ -1291,8 +1290,6 @@ void Z_INTERNAL fill_window(deflate_state *s) {
     Assert((unsigned long)s->strstart <= s->window_size - MIN_LOOKAHEAD,
            "not enough room for search");
 }
-
-
 
 #ifndef ZLIB_COMPAT
 /* =========================================================================
--- a/deflate.c
+++ b/deflate.c
@@ -1228,15 +1228,6 @@ void Z_INTERNAL fill_window(deflate_state *s) {
             unsigned int str = s->strstart - s->insert;
             if (str >= 1)
                 functable.quick_insert_string(s, str + 2 - STD_MIN_MATCH);
-#if STD_MIN_MATCH != 3
-            while (s->insert) {
-                functable.quick_insert_string(s, str);
-                str++;
-                s->insert--;
-                if (s->lookahead + s->insert < STD_MIN_MATCH)
-                    break;
-            }
-#else
             unsigned int count;
             if (UNLIKELY(s->lookahead == 1)) {
                 count = s->insert - 1;
@@ -1247,7 +1238,6 @@ void Z_INTERNAL fill_window(deflate_state *s) {
                 functable.insert_string(s, str, count);
                 s->insert -= count;
             }
-#endif
         }
         /* If the whole input has less than STD_MIN_MATCH bytes, ins_h is garbage,
          * but this is not important since only literal bytes will be emitted.
--- a/deflate.c
+++ b/deflate.c
@@ -433,17 +433,17 @@ int32_t Z_EXPORT PREFIX(deflateSetDictionary)(PREFIX3(stream) *strm, const uint8
     fill_window(s);
     while (s->lookahead >= STD_MIN_MATCH) {
         str = s->strstart;
-        n = s->lookahead - (STD_MIN_MATCH-1);
+        n = s->lookahead - (STD_MIN_MATCH - 1);
         functable.insert_string(s, str, n);
         s->strstart = str + n;
-        s->lookahead = STD_MIN_MATCH-1;
+        s->lookahead = STD_MIN_MATCH - 1;
         fill_window(s);
     }
     s->strstart += s->lookahead;
     s->block_start = (int)s->strstart;
     s->insert = s->lookahead;
     s->lookahead = 0;
-    s->prev_length = STD_MIN_MATCH-1;
+    s->prev_length = STD_MIN_MATCH - 1;
     s->match_available = 0;
     strm->next_in = (z_const unsigned char *)next;
     strm->avail_in = avail;
@@ -1160,7 +1160,7 @@ static void lm_init(deflate_state *s) {
     s->block_start = 0;
     s->lookahead = 0;
     s->insert = 0;
-    s->prev_length = STD_MIN_MATCH-1;
+    s->prev_length = STD_MIN_MATCH - 1;
     s->match_available = 0;
     s->match_start = 0;
 }
--- a/deflate_fast.c
+++ b/deflate_fast.c
@@ -76,11 +76,8 @@ Z_INTERNAL block_state deflate_fast(deflate_state *s, int flush) {
                 s->strstart += match_len;
             } else {
                 s->strstart += match_len;
-#if STD_MIN_MATCH != 3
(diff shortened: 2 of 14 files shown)


deflate_fast.c
```diff
@@ -95,7 +92,7 @@ Z_INTERNAL block_state deflate_fast(deflate_state *s, int flush) {
         if (UNLIKELY(bflush))
             FLUSH_BLOCK(s, 0);
     }
-    s->insert = s->strstart < MIN_MATCH-1 ? s->strstart : MIN_MATCH-1;
+    s->insert = s->strstart < (WANT_MIN_MATCH - 1) ? s->strstart : (STD_MIN_MATCH - 1);
```
mtl1979: I'm not 100% confident this is correct... it is guard for making sure there is at least 1 byte available before the string due to hashing offset.

deflate_medium.c
```diff
@@ -85,13 +85,10 @@ static void insert_match(deflate_state *s, struct match match) {
...
         match.match_length = 0;
-        if (match.strstart >= (MIN_MATCH - 2))
-#if MIN_MATCH != 3
-            functable.insert_string(s, match.strstart + 2 - MIN_MATCH, MIN_MATCH - 2);
-#else
-            functable.quick_insert_string(s, match.strstart + 2 - MIN_MATCH);
-#endif
-        /* If lookahead < MIN_MATCH, ins_h is garbage, but it does not
+        if (match.strstart >= (WANT_MIN_MATCH - 2))
+            functable.quick_insert_string(s, match.strstart + 2 - STD_MIN_MATCH);
```
mtl1979, line 90: Not 100% confident this is correct...

deflate_fast.c
Dead2: Yes, this is very much a request for comments PR, I am not sure about these either.
Please submit suggestions.

deflate_medium.c
```diff
@@ -280,7 +277,7 @@ Z_INTERNAL block_state deflate_medium(deflate_state *s, int flush) {
         if (UNLIKELY(bflush))
             FLUSH_BLOCK(s, 0);
     }
-    s->insert = s->strstart < MIN_MATCH-1 ? s->strstart : MIN_MATCH-1;
+    s->insert = s->strstart < (WANT_MIN_MATCH - 1) ? s->strstart : (STD_MIN_MATCH - 1);
```
mtl1979: Not 100% confident this is correct either...

deflate_quick.c
```diff
@@ -110,7 +110,7 @@ Z_INTERNAL block_state deflate_quick(deflate_state *s, int flush) {
         s->lookahead--;
     }
 
-    s->insert = s->strstart < MIN_MATCH-1 ? s->strstart : MIN_MATCH-1;
+    s->insert = s->strstart < (WANT_MIN_MATCH - 1) ? s->strstart : (STD_MIN_MATCH - 1);
```
mtl1979: Not 100% confident this is correct either...

deflate_slow.c
```diff
@@ -126,7 +126,7 @@ Z_INTERNAL block_state deflate_slow(deflate_state *s, int flush) {
         (void) zng_tr_tally_lit(s, s->window[s->strstart-1]);
         s->match_available = 0;
     }
-    s->insert = s->strstart < MIN_MATCH-1 ? s->strstart : MIN_MATCH-1;
+    s->insert = s->strstart < (WANT_MIN_MATCH - 1) ? s->strstart : (STD_MIN_MATCH - 1);
```
mtl1979: Not 100% confident this is correct either...

mtl1979 (changes requested):
The guards use both values, when they should only use one of them... Basically it makes sure there is at least 1 extra byte before the string, so the hashing functions have enough bytes to calculate the initial hash. I don't think WANT_MIN_MATCH should be used there.

Dead2:
I elected to use STD_MIN_MATCH-1 whenever we set s->strstart, because it is the smaller of the two and I interpreted it as a sort of reset, as the same thing is done in lm_init().
And if that is correct, it should not cause problems, while using WANT_MIN_MATCH-1 might cause problems if we switched to deflate_rle for example, where we don't hash and thus use STD_MIN_MATCH instead.

Feel free to experiment with changes based on this and suggest them if they work.
Using WANT_MIN_MATCH with insert_string did not seem to work for example, but I have not looked into why.
Most of the current changes are "safe" because they only check for one byte more lookahead.

mtl1979:
@Dead2 `WANT_MIN_MATCH - 1` doesn't work as it skips hashing one byte... `WANT_MIN_MATCH` should be only used at end of window, not at start of window...

deflate_medium.c
Dead2: What do you think is wrong?

Dead2:
> @Dead2 `WANT_MIN_MATCH - 1` doesn't work as it skips hashing one byte... `WANT_MIN_MATCH` should be only used at end of window, not at start of window...

Um, where? Do you mean deflate_medium.c:280 for example?

mtl1979:
@Dead2 When it compares with `s->strstart` to see if start of the string is too close to start of the window.

Dead2:
> @Dead2 When it compares with `s->strstart` to see if start of the string is too close to start of the window.

Can you point to an example of where this is?

mtl1979:
@Dead2 Pretty much everywhere I made a comment on specific line... Once per file where it sets `s->insert`.

deflate_medium.c
mtl1979: I don't think it should use `WANT_MIN_MATCH`. It only needs 1 extra byte (`2 - 3 = -1`), but `WANT_MIN_MATCH - 2 = 2`.

Dead2:
> @Dead2 Pretty much everywhere I made a comment on specific line... Once per file where it sets `s->insert`.

Ok, so Yes to the deflate_medium.c:280 then.

As I said I am uncertain about that one. `s->insert` is described as "bytes at end of window left to insert", so it might need -4 since we might be leaving that much remaining, unless it is a standards thing.

deflate_medium.c
Dead2: Yes, I think you are correct, running tests now to verify it does not break.

mtl1979:
@Dead2 I'm not sure that description of `s->insert` is correct... Basically it is pointer of the string that is currently compared against previous literal or match. As such current string can never be at index 0, so whenever the guard is checked, `s->strstart` should always be at least 1. If it isn't the guard I mentioned will fail and the condition falls back to using `s->strstart` directly. The guard is there to make sure `s->insert` never becomes negative.

deflate_fast.c
mtl1979:
```suggestion
    s->insert = s->strstart < (STD_MIN_MATCH - 1) ? s->strstart : (STD_MIN_MATCH - 1);
```

deflate_medium.c
```diff
@@ -85,13 +85,10 @@ static void insert_match(deflate_state *s, struct match match) {
...
         match.strstart += match.match_length;
         match.match_length = 0;
-        if (match.strstart >= (MIN_MATCH - 2))
-#if MIN_MATCH != 3
-            functable.insert_string(s, match.strstart + 2 - MIN_MATCH, MIN_MATCH - 2);
-#else
-            functable.quick_insert_string(s, match.strstart + 2 - MIN_MATCH);
-#endif
-        /* If lookahead < MIN_MATCH, ins_h is garbage, but it does not
+        if (match.strstart >= (WANT_MIN_MATCH - 2))
```
mtl1979:
```suggestion
        if (match.strstart >= (STD_MIN_MATCH - 2))
```

deflate_medium.c
mtl1979:
```suggestion
    s->insert = s->strstart < (STD_MIN_MATCH - 1) ? s->strstart : (STD_MIN_MATCH - 1);
```

Dead2:
Please verify that the changes are as suggested.
Testing shows no change to output.

Dead2:
Replacing that check with a plain `s->insert = 0;` actually passes tests and results in no change in output in my tests..

mtl1979:
@Dead2 It should only affect output if start of the window has a match. Basically index 0 and 1 of the window must be same. Of course the following 3 bytes must also match, but not the 4th byte.

Dead2:
@mtl1979 I see insert is only used when filling the window.
The fact that overriding that check makes no apparent change to anything suggests we should probably have a regression test that triggers a bug if the check is missing/wrong. Are you able to make such a test?

mtl1979:
@Dead2 It would have to compare the raw compressed stream and that can change for various reasons...

deflate.c
```diff
@@ -454,19 +454,19 @@ int32_t Z_EXPORT PREFIX(deflateSetDictionary)(PREFIX3(stream) *strm, const uint8
...
+        s->lookahead = STD_MIN_MATCH - 1;
         fill_window(s);
     }
     s->strstart += s->lookahead;
     s->block_start = (int)s->strstart;
     s->insert = s->lookahead;
     s->lookahead = 0;
-    s->prev_length = MIN_MATCH-1;
+    s->prev_length = STD_MIN_MATCH - 1;
     s->match_available = 0;
```
nmoinvaz, line 447: `s->prev_length` can be set to zero here and in the other place if we wanted.

deflate_medium.c
```diff
@@ -67,7 +67,7 @@ static void insert_match(deflate_state *s, struct match match) {
     /* Insert new strings in the hash table only if the match length
      * is not too large. This saves time but degrades compression.
      */
-    if (match.match_length <= 16* s->max_insert_length && s->lookahead >= MIN_MATCH) {
+    if (match.match_length <= 16* s->max_insert_length && s->lookahead >= WANT_MIN_MATCH) {
```
nmoinvaz: It would be nice if there was a space after the literal `16` (easier to read).

nmoinvaz:
> WANT_MIN_MATCH is set to 4, and is used to indicate to the code what the minimum match length we intend to accept is.

How is this different than what Cloudflare does?

nmoinvaz:
@Dead2 if `WANT_MIN_MATCH` == 4 then the branch in match_tpl.h `if (best_len < sizeof(uint32_t)) {` is not needed. Maybe right above it, it should be `#ifdef UNALIGNED_OK && WANT_MIN_MATCH < 4`?

deflate.c
```diff
@@ -1285,20 +1285,10 @@ void Z_INTERNAL fill_window(deflate_state *s) {
         s->lookahead += n;
 
         /* Initialize the hash value now that we have some input: */
-        if (s->lookahead + s->insert >= MIN_MATCH) {
+        if (s->lookahead + s->insert >= STD_MIN_MATCH) {
             unsigned int str = s->strstart - s->insert;
             if (str >= 1)
-                functable.quick_insert_string(s, str + 2 - MIN_MATCH);
-#if MIN_MATCH != 3
-#error Call insert_string() MIN_MATCH-3 more times
```
nmoinvaz, line 1232: Good that you removed that.

mtl1979:
@nmoinvaz I think more correct would be: `#if defined(UNALIGNED_OK) && (WANT_MIN_MATCH < 4)`

`#ifdef` should not be used when there is two or more conditions.

It's usually safer to enclose comparisons in parentheses if they are used in conditionals, as operator precedence can be unexpected.

deflate_medium.c
Dead2: Fixed

deflate.c
Dead2: Added

deflate_medium.c
Dead2: Added

Dead2:
> > WANT_MIN_MATCH is set to 4, and is used to indicate to the code what the minimum match length we intend to accept is.
> 
> How is this different than what Cloudflare does?

Not sure exactly what cloudflare does, but this lets us make WANT_MIN_MATCH configurable.
Before this PR it is 3, and that means we search for a 4 byte match when we only have 3 bytes left of the window, that is wasteful computation as it will never succeed, but it is not often so it has little effect on performance. It probably has a bigger impact on the smallest window sizes though.

Currently setting it to 3 or 4 makes no difference to the output because we hash 4 bytes anyways. It can be set to 5 or 10 if someone wanted to experiment (and it does work).

We could use WANT_MIN_MATCH to select between 3 or 4-byte hashing directly.

When we re-implement 3-byte hashing, things might change, and we could for example move WANT_MIN_MATCH into each of the deflate algorithms with separate values, such as 4 for quick, fast and medium, and 3 for slow, possibly. But we would have to look into how to support switching levels with differing hashing, perhaps hash both 3 and 4 bytes into separate buffers if WANT_MIN_MATCH is 3, otherwise only hash 4 bytes if all levels have it set to 4.

Dead2:
> @Dead2 if `WANT_MIN_MATCH` == 4 then the branch in match_tpl.h `if (best_len < sizeof(uint32_t)) {` is not needed. Maybe right above it, it should be `#ifdef UNALIGNED_OK && WANT_MIN_MATCH < 4`?

I looked at this a little, but was unable to really understand what you meant to change.
Did you mean that (if 4) it should not compile in any of the 3 first for loops and only keep the last one in the `#else` block at line 131? Wouldn't that be slower than using the UNALIGNED64_OK loop for example? There is so much going on in that function with custom types and ifdef sections that it is kind of hard to grok reliably.

deflate_slow.c
```diff
@@ -59,22 +59,22 @@ Z_INTERNAL block_state deflate_slow(deflate_state *s, int flush) {
...
-                match_len = MIN_MATCH-1;
+                match_len = STD_MIN_MATCH - 1;
             }
         }
         /* If there was a match at the previous step and the current
          * match is not better, output the previous match:
          */
-        if (s->prev_length >= MIN_MATCH && match_len <= s->prev_length) {
-            unsigned int max_insert = s->strstart + s->lookahead - MIN_MATCH;
+        if (s->prev_length >= WANT_MIN_MATCH && match_len <= s->prev_length) {
```
nmoinvaz, line 71: As you said, this will need to be `STD_MIN_MATCH` for fast-zlib changes because it uses a different hashing algorithm for deflate_slow. See #828.

nmoinvaz:
@Dead2 I can fix `longest_match` at a later stage if you prefer.

nmoinvaz:
Actually we will probably need that `longest_match` code for fast-zlib so likely nothing to be done there.

nmoinvaz:
Needs rebase.

deflate_slow.c
Dead2: To be clear; did you want me to change this or did you mean you would change this?

deflate_medium.c
```diff
@@ -283,7 +281,7 @@ Z_INTERNAL block_state deflate_medium(deflate_state *s, int flush) {
         if (UNLIKELY(bflush))
             FLUSH_BLOCK(s, 0);
     }
-    s->insert = s->strstart < MIN_MATCH-1 ? s->strstart : MIN_MATCH-1;
+    s->insert = s->strstart < (STD_MIN_MATCH - 1) ? s->strstart : (STD_MIN_MATCH - 1);
```
mtl1979, line 284: This should most likely use `MIN()` macro now.

deflate_fast.c
```diff
@@ -95,7 +92,7 @@ Z_INTERNAL block_state deflate_fast(deflate_state *s, int flush) {
         if (UNLIKELY(bflush))
             FLUSH_BLOCK(s, 0);
     }
-    s->insert = s->strstart < MIN_MATCH-1 ? s->strstart : MIN_MATCH-1;
+    s->insert = s->strstart < (STD_MIN_MATCH - 1) ? s->strstart : (STD_MIN_MATCH - 1);
```
mtl1979, line 95: This should most likely use `MIN()` macro now.

deflate_quick.c
```diff
@@ -110,7 +110,7 @@ Z_INTERNAL block_state deflate_quick(deflate_state *s, int flush) {
         s->lookahead--;
     }
 
-    s->insert = s->strstart < MIN_MATCH-1 ? s->strstart : MIN_MATCH-1;
+    s->insert = s->strstart < (STD_MIN_MATCH - 1) ? s->strstart : (STD_MIN_MATCH - 1);
```
mtl1979, line 113: This should most likely use `MIN()` macro now.

deflate_slow.c
mtl1979: I would say it depends on which pull request gets merged first.

deflate_slow.c
```diff
@@ -126,7 +126,7 @@ Z_INTERNAL block_state deflate_slow(deflate_state *s, int flush) {
         (void) zng_tr_tally_lit(s, s->window[s->strstart-1]);
         s->match_available = 0;
     }
-    s->insert = s->strstart < MIN_MATCH-1 ? s->strstart : MIN_MATCH-1;
+    s->insert = s->strstart < (STD_MIN_MATCH - 1) ? s->strstart : (STD_MIN_MATCH - 1);
```
mtl1979, line 129: This should most likely use `MIN()` macro now.

mtl1979 (changes requested):
Needs some cleanup ;)

Dead2:
I think adding MIN() does not belong in this PR, but I agree those should probably be converted in another PR.

PR #956, opened 2021-05-12
Fix compiler test to not succeed with mpicc

gsjaardema:
The compiler test for matching the intel compilers (icc, icpc, icl) was incorrectly matching the parallel compiler `mpicc`.  This causes the `-w3` flag to be added to `WARNFLAGS` and `WARNFLAGS_MAINTAINER` and the cmake config fails since `mpicc` may not recognize that flag if it is not intel-based.


CMakeLists.txt
```diff
@@ -134,7 +134,7 @@ if(WITH_GZFILEOP)
     add_definitions(-DWITH_GZFILEOP)
 endif()
 
-if("${CMAKE_C_COMPILER}" MATCHES "icc" OR "${CMAKE_C_COMPILER}" MATCHES "icpc" OR "${CMAKE_C_COMPILER}" MATCHES "icl")
+if(("${CMAKE_C_COMPILER}" MATCHES "icc" AND NOT "${CMAKE_C_COMPILER}" MATCHES "mpicc") OR "${CMAKE_C_COMPILER}" MATCHES "icpc" OR "${CMAKE_C_COMPILER}" MATCHES "icl")
```
mtl1979: I would rather see mpicc having own section before section for icc... That way this test doesn't become hard to read.

That new section would list correct flags that mpicc supports.

gsjaardema: mpicc is typically a wrapper around either gcc or clang or intel or ibm compilers.  Would need to be able to somehow determine the underlying compiler (which is already done by CMake:
```
s1041980:zlib-ng-2.0.2(master)> CC=mpicc cmake -Bbuild -DCMAKE_INSTALL_PREFIX=${INSTALL_PATH} -DZLIB_COMPAT=YES .
-- Using CMake version 3.19.8
-- ZLIB_HEADER_VERSION: 1.2.11
-- ZLIBNG_HEADER_VERSION: 2.0.2
-- The C compiler identification is Clang 8.0.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /opt/local/bin/mpicc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
```

So maybe the test needs to be done on whatever symbol is set for `C compiler identification`?   I will look closer to see how to make this cleaner.

mtl1979: As far as I know, MPI has own package for CMake, so overriding `CC` and `CXX` is discouraged.

nmoinvaz:
Another thing that is already fixed in #906.

gsjaardema:
@nmoinvaz Thanks.  That matches the fix I was just getting ready to change this to.

nmoinvaz:
Why close this? I thought you were going to make the changes? This patch would be more isolated and we could get it in sooner.

gsjaardema:
Sorry, I thought that it was fixed in #906.  I can make the change

CMakeLists.txt
gsjaardema: In all of the parallel MPI builds that we do (Trilinos, SEACAS, HDF5, NetCDF), we set the compiler to `mpicc` and `mpicxx`.  CMake does recognize that this is a parallel compiler and handles it from there...

CMakeLists.txt
```diff
@@ -134,7 +134,7 @@ if(WITH_GZFILEOP)
     add_definitions(-DWITH_GZFILEOP)
 endif()
 
-if("${CMAKE_C_COMPILER}" MATCHES "icc" OR "${CMAKE_C_COMPILER}" MATCHES "icpc" OR "${CMAKE_C_COMPILER}" MATCHES "icl")
+if("${CMAKE_C_COMPILER_ID}" MATCHES "Intel")
```
nmoinvaz: Should be `if(CMAKE_C_COMPILER_ID MATCHES "Intel")` like the comparisons for GNU further down. No need for double-quotes or `${}`

nmoinvaz:
@gsjaardema can you also squash the commits?

CMakeLists.txt
gsjaardema: I think either works, but if you prefer I will make the change.  Benefits both ways...

gsjaardema:
@nmoinvaz Sorry, I've just been editting directly in github, so not sure how to squash the commits from here...

gsjaardema:
Looks like the squash can be done at merge time -- https://docs.github.com/en/github/administering-a-repository/configuring-commit-squashing-for-pull-requests

nmoinvaz (approved):

mtl1979:
Editing directly on GitHub website is generally a bad idea... In the past we have required every contributor to squash pull requests themselves.

I don't have adequate permissions or capacity to do required changes to our commit policy or repository settings.

I will leave that kind of stuff to @Dead2 for now...

gsjaardema:
Yes, I agree. I typically only do it for what I assume will be a short minimal change to one file; however, in this case it didn't work out that way.

gsjaardema:
I've got this messed up...  Starting over...

PR #959, opened 2021-05-13
Fix compiler detection to avoid bad mpicc match

gsjaardema:
Redo of #956.  The previous test would match the MPI `mpicc` with the test for matching the intel compiler `icc`.  This test uses CMake determined `CMAKE_C_COMPILER_ID` which is a safer test.

Also fixes a test for the host machine.  `APPLE` was missing `CMAKE_HOST_APPLE`


gsjaardema:
Sorry, should have checked before assuming, but `APPLE` is a valid CMake variable, so does not need to be changed to `CMAKE_HOST_APPLE`.  I can remove that change if desired.

nmoinvaz:
It does not need APPLE at all because UNIX implies APPLE. So it would be better if APPLE was removed.

nmoinvaz:
LGTM thank you!

gsjaardema:
I think a maintainer has to click a button to allow the workflows to run for this PR.  I am a first-time contributor to this repository.

nmoinvaz:
@Dead2 will need to approve.

Dead2:
Did anyone actually test this with an Intel compiler to verify we don't have any regressions?

gsjaardema:
@Dead2 Yes, I have tested it with `CC=icc cmake ..` and with `CC=mpicc cmake ..` (where mpicc is based on an intel compiler) and both correctly identify that the compiler (or underlying compiler) is intel-based.  Configuration suceeds; build succeeds and tests complete.

scivision:
I tested this on Windows as well. Before this PR, GCC `mpicc` fails, and with this PR, it's fixed. Thank you!

PR #960, opened 2021-05-13
Handle HAVE_UNISTD_H defined to 0.

lemourin:
FFmpeg during the configure stage generates a config.h file with
```
#define HAVE_UNISTD_H 0
```
on windows. Then somewhere in FFmpeg's code there is:
```
#include "config.h"  // FFmpeg's config.h
#include <zlib.h>
```
which causes zlib.h to include unistd.h on windows. It is way easier to handle the issue here than in FFmpeg.


mtl1979:
`zconf.h.in` should not be modified by anything other than `configure` script... `configure` script depends on specific strings to exist in the file.

lemourin:
What if I care only about the cmake build system? Is there a way to upstream this fix?

mtl1979:
Most of the stuff on `zconf.h.in` should be compatible with stock zlib that zlib-ng declares to be compatible with... Whatever cmake does must be identical to what `configure` does, as the CI compares the outputs. Both make a copy of the file before making any edits.

lemourin:
Thanks for the pointers. I updated both configure and CMakeLists.txt.

Dead2 (approved):
LGTM

nmoinvaz:
@Dead2 can you re-run the CI?

nmoinvaz:
It seems like this change should also apply to `HAVE_STDARG_H`.

Dead2:
Triggered re-runs now.

CMakeLists.txt
```diff
@@ -1054,7 +1054,7 @@ endif()
 if(HAVE_UNISTD_H)
   SET(ZCONF_UNISTD_LINE "#if 1    /* was set to #if 1 by configure/cmake/etc */")
 else()
-  SET(ZCONF_UNISTD_LINE "#ifdef HAVE_UNISTD_H    /* may be set to #if 1 by configure/cmake/etc */")
+  SET(ZCONF_UNISTD_LINE "#if ~ HAVE_UNISTD_H + 1    /* may be set to #if 1 by configure/cmake/etc */")
```
nmoinvaz: Can you explain this logic a bit more?

configure
```diff
@@ -779,6 +779,8 @@ if try $CC -c $CFLAGS $test.c; then
   mv zconf${SUFFIX}.temp.h zconf${SUFFIX}.h
   echo "Checking for unistd.h... Yes." | tee -a configure.log
 else
+  sed < zconf${SUFFIX}.h "/^#ifdef HAVE_UNISTD_H.* may be/s/def HAVE_UNISTD_H\(.*\)/ ~ HAVE_UNISTD_H + 1\1/" > zconf${SUFFIX}.temp.h
```
nmoinvaz: What is happening here?

nmoinvaz (changes requested):
If these changes are going to be accepted they should be done for the other headers such as `HAVE_STDARG_H`.

CMakeLists.txt
mtl1979: If something is not defined, it equals either 0 or empty string...

lemourin: This has the exact same semantics as here [1]; it is only written slightly shorter.

[1]: https://github.com/zlib-ng/zlib-ng/blob/c91806258805d243080a934ebd169333d4f8ba46/zconf.h.in#L152

configure
lemourin: This is just as hard to parse as line 778 which I based it on. It does the same as cmake i.e when unistd.h is not available then replace `#ifdef HAVE_UNISTD_H` with `#ifdef ~ HAVE_UNISTD_H + 1`

CMakeLists.txt
mtl1979: I don't think shorter is better in this case... It is just confusing...

lemourin: IMO `-_LARGEFILE64_SOURCE - -1 == 1` is just as unreadable.

configure
nmoinvaz: I see. So `#ifdef ~ HAVE_UNISTD_H + 1` does not cover the case where `HAVE_UNISTD_H` is not defined?

lemourin:
I don't see `HAVE_STDARG_H` being used anywhere.

CMakeLists.txt
mtl1979: Possibly... But fixing both would change scope of this pull request...

mtl1979:
Some are obviously relics inherited from stock zlib... We already cleaned up CMakeLists.txt.

configure
lemourin: It does work with all compilers actually as expected i.e the code block is not enabled. https://godbolt.org/z/afdqonqah

lemourin:
So is the cleanup in scope of this PR?

mtl1979:
@lemourin Not in my opinion...

mtl1979:
I'm still not 100% sure if we should fix this in zlib-ng or if this is the best way to fix. Moving the fix to zconf.h.in and zconf-ng.h.in would also require editing two files and wouldn't be any cleaner solution as we would need to add extra lines before the current test that basically undefines the macro if it is set to 0.

The cons for fixing this in configure/CMakeLists.txt are pretty much style issues and having the workaround only for case when the macro is set to 0 outside zlib-ng build system. If the macro is set to 1 outside zlib-ng build system, it is essentially no-op and ignored.

lemourin:
The root cause of the issue is that macros are not scoped in C so in this case `HAVE_UNISTD_H` macro clashes with that of FFmpeg. The two libraries use the same macro with different semantics. 

A different solution could be to replace `HAVE_UNISTD_H` with just `0` in cmake/configure if `unistd.h` header is not detected.

I don't think undefining a macro is a good idea; it will likely break the users of zlib (i.e FFmpeg).

If this is not zlib-ng issue then where does the fix belong? I don't want to create yet another zlib fork :(

mtl1979:
@lemourin There is no easy answer... Obviously FFmpeg should support zlib-ng without any changes just like it's supposed to support original zlib. To summarize, any project that includes another project as dependency should be able to handle configuration of that subproject without any patches to official sources.

I'm in no position to blatantly reject pull requests, only @Dead2 can do that. But controversial changes like this require approval of two long-time collaborators and I'm not very keen in approving until I'm 100% confident there is no better alternative.

kepstin:
After doing some investigation with folks in the ffmpeg IRC channel, we've determined that classic zlib does *not* include any references to the `HAVE_UNISTD_H` define in the installed zconf.h file - the configure script for zlib replaces that with a static `#if 1` or `#if 0`, and then internally defines a prefixed define, `Z_HAVE_UNISTD_H`. (the same is true for `HAVE_STDARG_H`).

For compatibility, it would be best if zlib-ng followed suit here - there should be no references to the unprefixed `HAVE_UNISTD_H` or `HAVE_STDARG_H` defines in the installed `zconf.h` file.

Snippit of the installed `zconf.h` file from zlib 1.2.11:
```c
#if 1    /* was set to #if 1 by ./configure */
#  define Z_HAVE_UNISTD_H
#endif

```

Edit: Ah, I'm probably misunderstanding something about exactly which conditions cause that replacement to be performed. It looks like there are some cases where the zlib configure script does leave the `#ifdef HAVE_UNISTD_H` check in place, like the zlib-ng code does.

mtl1979:
@kepstin I don't see anywhere in zlib build system where it replaces anything with ```#if 0```, however it would be really clean if zlib-ng would do it that way.

kepstin:
> @kepstin I don't see anywhere in zlib build system where it replaces anything with `#if 0`, however it would be really clean if zlib-ng would do it that way.

Yeah, that does seem to be the case. That said, the configure script is pretty hard to read (or maybe i'm just not good at reading sed), and I'm not even sure how the `#if 1` substitution is being done. I think it would be a nice improvement if zlib-ng's build scripts could do an `#if 0` replacement where possible to reduce the cases where a stray unprefixed define check gets into an installed `zconf.h` file.

I suppose the reason for the way the check was done in the original zlib was to make it easier to statically include the zlib source files in another project (in which case the header wouldn't get installed).

mtl1979:
@kepstin I've used Linux since December 1996 and I still get hairballs trying to decipher sed scripts... Sometimes I just guess repeatedly what the script should be until the result is expected...

CMakeLists.txt
mtl1979:
```suggestion
  SET(ZCONF_UNISTD_LINE "#if 0    /* was set to #if 0 by configure/cmake/etc */")
```

configure
mtl1979:
```suggestion
  sed < zconf${SUFFIX}.h "/^#ifdef HAVE_UNISTD_H.* may be/s/def HAVE_UNISTD_H\(.*\) may be set to #if 1/ 0\1 was set to #if 0/" > zconf${SUFFIX}.temp.h
```

mtl1979 (approved):

lemourin:
Can this be merged? :)

mtl1979:
@Dead2 will merge when he is online...

nmoinvaz:
I think we should wait for a proper fix which is:
> there should be no references to the unprefixed HAVE_UNISTD_H or HAVE_STDARG_H defines in the installed zconf.h file.

mtl1979:
@nmoinvaz My personal opinion is that we should remove everything mentioning `HAVE_STDARG_H` as it is unused and doesn't affect compatibility with stock zlib. But that can be done in separate pull request. See #976.

PR #964, opened 2021-05-15
Fix build on NetBSD

kolbycrouch:
This fixes building on NetBSD for at least amd64.


mtl1979:
We can accept "incremental" patches, but it would be useful to have CI for NetBSD to verify this is the only fix/change that is required. Otherwise the patch is clean and I see no reason not to accept it... @Dead2 can merge it when he is online.

Dead2 (approved):
LGTM

PR #969, opened 2021-05-20
Open up develop branch for non-stable commits.

Dead2:
Lets open up develop for the next version.
I think this is a good time to remove the hex-based version define as well, as I foresee it might cause problems in the future due to its very limited ranges. We could only have 16 minor/revision versions for example, and I don't think it really brings much to the table feature-wise.

We should probably also define legal values for the defines as well. I suggest the following rules:
`ZLIBNG_VER_MAJOR` = single or multi-digit number.
`ZLIBNG_VER_MINOR` = single or multi-digit number.
`ZLIBNG_VER_REVISION` = single or multi-digit number.
`ZLIBNG_VER_SUBREVISION` = single or multi-digit number, or 'd' for development build and 'f' for externally modified build.

`ZLIBNG_VERSION` = `MAJOR`, `MINOR` and `REVISION` with . between them, optionally followed by a dot and an alphanumeric identifier. Same as used in `zlibng_string`.


mtl1979:
The hex-based version is useful when compiling against zlib-ng and when new version of zlib-ng changes behavior of existing functions or structures. This is important especially when some padding members are replaced with real struct member variables.

It would not be hard to add more bits for the parts of the hex value like was already some for some other projects.

Dead2:
If we add more bits for the hex values, we would have to rename it since it would be an incompatible change. And it would make it hard for applications to support both.

I don't see why you cannot do the same checks with `ZLIBNG_VER_MAJOR` and `ZLIBNG_VER_MINOR` though.

mtl1979:
@Dead2 It's a lot cleaner to use single variable, when some change is not done during change of major version, but instead on change of minor version... Obviously very few breaking changes should be made on change of revision or build number.

Dead2:
The removal and introduction of a new bigger hex-based value would cause even more complicated checks than just checking the two existing defines though.
Checking the value of two defines is not exactly much or difficult.

I agree that just checking the hex-based one would be simpler, but since we didn't replace that one with a bigger one before the first stable release, it has become a problem.

mtl1979:
@Dead2 The earlier you do the adding of bits, less problems it will cause... That's because adding more bits will make the value larger, not smaller.

nmoinvaz (approved):
LGTM. I have no opinion on the hex version number. OpenSSL also uses such a thing:
https://www.openssl.org/docs/man1.1.0/man3/OPENSSL_VERSION_NUMBER.html

mtl1979:
I think we have enough minor fixes already to warrant 2.0.4 before we skip to 2.1.0.

PR #970, opened 2021-05-21
OpenBSD build fix proposal

devnexen:


zutil.h
```diff
@@ -185,9 +185,15 @@ void Z_INTERNAL   zng_cfree(void *opaque, void *ptr);
 
 #elif defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__) || defined(__DragonFly__)
 #  include <sys/endian.h>
+#  if defined(__OpenBSD__)
+#  define ZSWAP16(q) __builtin_bswap16(q)
+#  define ZSWAP32(q) __builtin_bswap32(q)
+#  define ZSWAP64(q) __builtin_bswap64(q)
```
mtl1979: These should be `swap16`, `swap32` and `swap64` according to OpenBSD headers... We should not use compiler builtins if functions exist in standard headers. Might warrant separating OpenBSD as own `#elif` block.

devnexen: Yes there are but tried to use intrinsics rather than the macros (might be slower) but nevermind I ll change.

mtl1979: At least gcc should map the functions in the standard headers automatically to the intrinsics when optimizations are enabled.

devnexen: true

mtl1979 (approved):

Issue #971, opened 2021-05-22
Cross-Compiling with mingw gives libzlib.lib, even though ZLIB_COMPAT is ON (expected libz.lib)


Xaymar:
I'm trying to cross-compile zlib-ng for Windows using the provided toolchain while keeping zlib compatibility and found that the generated library files are all suffixed with `lib`. This breaks the expected compatibility with zlib, as the name should be `libz`, not `libzlib`. Below is my build command:

```
cmake -H. -Bbuild/build \
	-DCMAKE_TOOLCHAIN_FILE=./cmake/toolchain-mingw-x86_64.cmake \
	-DCMAKE_BUILD_TYPE=RELEASE -DZLIB_COMPAT=ON -DZLIB_ENABLE_TESTS=OFF -DBUILD_SHARED_LIBS=ON \
	-DCMAKE_INSTALL_PREFIX=./build/distrib/
cmake --build build/build --target install
# <snip> .lib stuff for msvc requirements
# Why do i need to do this with ZLIB_COMPAT=ON?
cp ./build/distrib/lib/libzlib.dll.a ./build/distrib/lib/libz.dll.a
```
Note that the generated zlib.pc file also has the link command `-lz`, not `-lzlib`. This appears to be the culprit: https://github.com/zlib-ng/zlib-ng/blob/develop/CMakeLists.txt#L996-L1001

nmoinvaz:
This should be resolved now in 2.0.4.

inkydragon:
This problem still exists.

https://github.com/zlib-ng/zlib-ng/blob/92107a92d3cfb11cc464c24ef2a73c2d7d05d2ba/CMakeLists.txt#L956-L975

Windows + cygwin + build flag:
+ NOT DEFINED BUILD_SHARED_LIBS: #L958, `libzlib`
+ `-DBUILD_SHARED_LIBS=ON`: #L958, `libzlib`
+ `-DBUILD_SHARED_LIBS=OFF`: #L968 -> #L972, `libz`
+ We will never execute line #L962~#L967

When do we need `zlib` as OUTPUT_NAME?

Or maybe you forgot to remove the prefix `lib`.
The library prefix is defined by the following two variables, both with the default value of `lib`.
+ `CMAKE_SHARED_LIBRARY_PREFIX`
+ `CMAKE_STATIC_LIBRARY_PREFIX`

xref: JuliaPackaging/Yggdrasil#3614

----

My proposal: replace #L957~#L974 with:
[...]

PR #972, opened 2021-05-22
Fix library name on Cygwin, MSYS and MinGW when using cmake

mtl1979:
See #971.


nmoinvaz:
Doesn't this line in madler/zlib mean that it is expecting zlib1.dll for all WIN32?
https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/CMakeLists.txt#L208-211

mtl1979:
@nmoinvaz I'm assuming `UNIX` is defined also when running Cygwin, MSYS and MinGW... All those have own names for the shared library. I can recheck the shared library name on hard disk of my old development PC, but it's too slow to compile code.

mtl1979:
@nmoinvaz I changed the code, not sure if we want filename zlibstatic-ng.lib on Windows when compiling with Visual C++...

Open for suggestions as long as it doesn't cause import library of dll to have same name as static library.

CMakeLists.txt
```diff
@@ -994,7 +994,24 @@ foreach(ZLIB_INSTALL_LIBRARY ${ZLIB_INSTALL_LIBRARIES})
 endforeach()
 
 if(WIN32)
-    set_target_properties(${ZLIB_INSTALL_LIBRARIES} PROPERTIES OUTPUT_NAME zlib${SUFFIX})
+    # Shared library
+    if(NOT DEFINED BUILD_SHARED_LIBS OR BUILD_SHARED_LIBS)
+        set_target_properties(zlib PROPERTIES OUTPUT_NAME zlib${SUFFIX})
+    endif()
+    # Static library
+    if(NOT DEFINED BUILD_SHARED_LIBS)
```
nmoinvaz, line 1002: Can you combine these four `set_target_properties` calls using `${ZLIB_INSTALL_LIBRARIES}` as the target?

nmoinvaz:
I'm not quite convinced that we need this because madler/zlib doesn't use this right? What am I missing?

CMakeLists.txt
mtl1979: I don't think so...  `${ZLIB_INSTALL_LIBRARIES}` can also contain the shared library.

I tried combining matching ones using `zlibstatic` as target name but it did throw error that aliases can't be used with `set_target_properties`.

mtl1979:
@nmoinvaz Stock zlib predates Cygwin, MSYS and MinGW, so its `CMakeLists.txt` wasn't fully tested with them. It used separate Makefile that did skip even `configure`.

CMakeLists.txt
nmoinvaz: What about using: `z$<$<C_COMPILER_ID:MSVC>:libstatic>${SUFFIX}`

mtl1979: That's pretty much unreadable to average developer...

PR #976, opened 2021-05-25
Remove unnecessary test for stdarg.h

mtl1979:
This style of test breaks if same test is done in project including zlib-ng.

zlib-ng doesn't include `stdarg.h` in any source files.


nmoinvaz (approved):

nmoinvaz:
I am seeing stdarg.h included in zconf.h & zlib.h & configure.. maybe those should be removed too?

nmoinvaz:
Actually looks like va_list is only used in one function: gzvprintf.

nmoinvaz:
Actually gzprintf too. But perhaps stdarg.h include can be moved to gzwrite.c instead of having it in zlib headers?

mtl1979:
@nmoinvaz We basically don't support anything that doesn't have `stdarg.h`... It's considered standard header nowadays.

nmoinvaz:
What do you think of removing it from zlib headers/config and moving it to gzwrite.c only where it is used?

mtl1979:
@nmoinvaz Doesn't compile if I try to remove include for `stdarg.h` as `va_list` is used in the headers.

PR #977, opened 2021-05-27
Some improvements to fuzzers

nmoinvaz:
* Use different fuzzer bits in `example_dict_fuzzer` for more input value combinations.
  *  This is similar to changes made in #929 but for `example_dict_fuzzer` instead of `minigzip_fuzzer`.
* Fixed gz mode for compression level not using ascii numeric value.
  *  This is a bug in `minigzip_fuzzer` writing the gz mode. Previously `minigzip_fuzzer` wouldn't have been testing all compression levels.
* Fixed trying to uncompress after compressing in gzip direct mode which is not supported by gz functions.
  * Changes made by #929 caused this issue. This should fix OSS-Fuzz issue https://oss-fuzz.com/testcase-detail/6194422837542912.

### DIFF
--- a/test/fuzz/example_dict_fuzzer.c
+++ b/test/fuzz/example_dict_fuzzer.c
@@ -42,13 +42,13 @@ void test_dict_deflate(unsigned char **compr, size_t *comprLen) {
 
     int method = Z_DEFLATED; /* The deflate compression method (the only one
                                 supported in this version) */
-    int windowBits = 8 + data[0] % 8; /* The windowBits parameter is the base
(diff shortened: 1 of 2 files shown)


Dead2:
I would like to see coverage diff for this one.
But it seems all the Clang-based coverage has stopped working..
```
Unexpected version: *804.
Invalid .gcno File!
```

nmoinvaz:
See #981. I don't think this PR will change code coverage very much if any at all. All the compression levels are already tested by our other tests. And during testing we only test these fuzzers with a limited number of files.

PR #978, opened 2021-05-28
Separate huff, rle, and stored deflate strategies into their own sources

nmoinvaz:
This PR creates 3 additional source files: `deflate_huff.c`, `deflate_rle.c`, and `deflate_stored.c`. By doing this the deflate strategies are better separated from deflate.c.


deflate_p.h
```diff
@@ -12,7 +12,37 @@
 /* Forward declare common non-inlined functions declared in deflate.c */
 
 #ifdef ZLIB_DEBUG
-void check_match(deflate_state *s, Pos start, Pos match, int length);
+/* ===========================================================================
+ * Check that the match at match_start is indeed a match.
+ */
+static void check_match(deflate_state *s, Pos start, Pos match, int length) {
```
mtl1979: Non-inline functions should not be defined in header files. 

If defined in header file, will make private copy of the function in each translation unit where included (and used).

deflate_stored.c
```diff
@@ -0,0 +1,194 @@
...
+ * Copyright (C) 1995-2013 Jean-loup Gailly and Mark Adler
+ * For conditions of distribution and use, see copyright notice in zlib.h
+ */
+
+#include "zbuild.h"
+#include "deflate.h"
+#include "deflate_p.h"
+#include "functable.h"
+
+unsigned read_buf(PREFIX3(stream) *strm, unsigned char *buf, unsigned size);
```
mtl1979: This should be moved to private header, so there is no risk of conflicting declarations.

Should be `Z_INTERNAL`.

deflate_stored.c
```diff
@@ -0,0 +1,194 @@
...
+    if (s->strm->avail_in > have && s->block_start >= (int)s->w_size) {
+        /* Slide the window down. */
+        s->block_start -= (int)s->w_size;
+        s->strstart -= s->w_size;
+        memcpy(s->window, s->window + s->w_size, s->strstart);
+        if (s->matches < 2)
+            s->matches++;           /* add a pending slide_hash() */
+        have += s->w_size;          /* more space now */
+        if (s->insert > s->strstart)
+            s->insert = s->strstart;
```
mtl1979: Codecov keeps complaining about this type of comparison... We should use MIN() macro to suppress the warning.

deflate_stored.c
```diff
@@ -0,0 +1,194 @@
...
+        have = (s->bi_valid + 42) >> 3;         /* number of header bytes */
+        if (s->strm->avail_out < have)          /* need room for header */
+            break;
+            /* maximum stored block length that will fit in avail_out: */
+        have = s->strm->avail_out - have;
+        left = (int)s->strstart - s->block_start;    /* bytes left in window */
+        if (len > (unsigned long)left + s->strm->avail_in)
+            len = left + s->strm->avail_in;     /* limit len to the input */
+        if (len > have)
+            len = have;                         /* limit len to the output */
```
mtl1979: This is essentially same as:
`len = MIN(len, have);`

Using the macro makes it fit nicely in single line.

deflate_stored.c
```diff
@@ -0,0 +1,194 @@
...
+        flush_pending(s->strm);
+
+        /* Update debugging counts for the data about to be copied. */
+        cmpr_bits_add(s, len << 3);
+        sent_bits_add(s, len << 3);
+
+        /* Copy uncompressed bytes from the window to next_out. */
+        if (left) {
+            if (left > len)
+                left = len;
```
mtl1979: Essentially same as:
`left = MIN(left, len);`

Using the macro will avoid Codecov warning.

deflate_stored.c
```diff
@@ -0,0 +1,194 @@
...
+            s->insert = s->strstart;
+        } else {
+            if (s->window_size - s->strstart <= used) {
+                /* Slide the window down. */
+                s->strstart -= s->w_size;
+                memcpy(s->window, s->window + s->w_size, s->strstart);
+                if (s->matches < 2)
+                    s->matches++;   /* add a pending slide_hash() */
+                if (s->insert > s->strstart)
+                    s->insert = s->strstart;
```
mtl1979: Essentially same as:
`s->insert = MIN(s->insert, s->strstart);`

deflate_stored.c
```diff
@@ -0,0 +1,194 @@
...
+            }
+            memcpy(s->window + s->strstart, s->strm->next_in - used, used);
+            s->strstart += used;
+            s->insert += MIN(used, s->w_size - s->insert);
+        }
+        s->block_start = (int)s->strstart;
+    }
+    if (s->high_water < s->strstart)
+        s->high_water = s->strstart;
```
mtl1979: Essentially same as:
`s->high_water = MAX(s->high_water, s->strstart);`

deflate_stored.c
```diff
@@ -0,0 +1,194 @@
...
+        s->strstart -= s->w_size;
+        memcpy(s->window, s->window + s->w_size, s->strstart);
+        if (s->matches < 2)
+            s->matches++;           /* add a pending slide_hash() */
+        have += s->w_size;          /* more space now */
+        if (s->insert > s->strstart)
+            s->insert = s->strstart;
+    }
+    if (have > s->strm->avail_in)
+        have = s->strm->avail_in;
```
mtl1979: Essentially same as:
`have = MIN(have, s->strm->avail_in);`

deflate_stored.c
```diff
@@ -0,0 +1,194 @@
...
+    if (have) {
+        read_buf(s->strm, s->window + s->strstart, have);
+        s->strstart += have;
+        s->insert += MIN(have, s->w_size - s->insert);
+    }
+    if (s->high_water < s->strstart)
+        s->high_water = s->strstart;
```
mtl1979: Essentially same as:
`s->high_water = MAX(s->high_water, s->strstart);`

deflate_rle.c
```diff
@@ -0,0 +1,81 @@
...
+                strend = s->window + s->strstart + MAX_MATCH;
+                do {
+                } while (prev == *++scan && prev == *++scan &&
+                         prev == *++scan && prev == *++scan &&
+                         prev == *++scan && prev == *++scan &&
+                         prev == *++scan && prev == *++scan &&
+                         scan < strend);
+                match_len = MAX_MATCH - (unsigned int)(strend - scan);
+                if (match_len > s->lookahead)
+                    match_len = s->lookahead;
```
mtl1979: Essentially same as:
`match_len = MIN(match_len, s->lookahead);`

Inconsistent use of `MIN()` macro and equivalent `if` comparison.

deflate_stored.c
```diff
@@ -0,0 +1,194 @@
...
+         */
+        len = MAX_STORED;       /* maximum deflate stored block length */
+        have = (s->bi_valid + 42) >> 3;         /* number of header bytes */
+        if (s->strm->avail_out < have)          /* need room for header */
+            break;
+            /* maximum stored block length that will fit in avail_out: */
+        have = s->strm->avail_out - have;
+        left = (int)s->strstart - s->block_start;    /* bytes left in window */
+        if (len > (unsigned long)left + s->strm->avail_in)
+            len = left + s->strm->avail_in;     /* limit len to the input */
```
mtl1979, line 53: Both `len` and `left` are defined as `unsigned int`, so when `left` is cast to larger type, the comparison wraps around as `len` can't be larger than what `unsigned int` can fit. 

Obviously `len` on line 54 might be implicitly zero-extended to `unsigned long`, but that makes the comparison (almost) always false.

deflate_stored.c
nmoinvaz: I didn't change any of those lines though.

deflate_p.h
nmoinvaz: We can make it inline. The function is only for debug builds anyways.

deflate_stored.c
mtl1979: Codecov keeps reminding until someone fixes them... I recently did fix some "newbie" style warnings elsewhere just to shut up the constant complaining.

deflate_p.h
mtl1979: Inlining should decrease binary size by few bytes and make some compilers (and indirectly linkers) happy that don't by default inline small functions (read: Visual C++).

mtl1979 (changes requested):
Needs some code and style cleanup. Further details are in the line comments.

deflate_stored.c
nmoinvaz: Fixed.

deflate_rle.c
nmoinvaz: Fixed.

deflate_stored.c
nmoinvaz: Fixed.

deflate_stored.c
nmoinvaz: Fixed.

deflate_stored.c
nmoinvaz: Fixed.

deflate_stored.c
nmoinvaz: Fixed.

deflate_stored.c
nmoinvaz: Fixed.

deflate_stored.c
nmoinvaz: Fixed.

deflate_stored.c
nmoinvaz: It is moved.

deflate_p.h
nmoinvaz: Done.

deflate_stored.c
nmoinvaz: I'm not sure why we can't use `MIN()` here. The `len` will be wrapped around if it is too big always.

mtl1979: MIN() as a macro only works with variables, not expressions... This is limitation of preprocessor. 

Instead of relying on implicit cast, which is non-standard (standard says cast to `int` if types are different), it should have explicit cast on both sides of comparison.

nmoinvaz: > MIN() as a macro only works with variables, not expressions... This is limitation of preprocessor.

`len = MIN(len, left + s->strm->avail_in);` compiles fine.

mtl1979: It does compile, but the result of second parameter can overflow... That is why the original code had cast to `unsigned long`...

Problem of the original code was that it assumed `len` was automatically zero-extended to `unsigned long` before comparison... Some compilers do it, but that is not standard behavior. Essentially it creates temporary variable that is discarded right after the comparison. I would say it's waste of registers as zero-extension can't be done for variable stored in memory.

nmoinvaz: Ok.

PR #980, opened 2021-05-29
Replace CMAKE_CURRENT_SOURCE_DIR to CMAKE_CURRENT_LIST_DIR

phprus:
Replace ```CMAKE_CURRENT_SOURCE_DIR``` to ```CMAKE_CURRENT_LIST_DIR``` to allow reuse script ```cmake/detect-arch.cmake``` in parent CMake project.

### DIFF
--- a/cmake/detect-arch.cmake
+++ b/cmake/detect-arch.cmake
@@ -25,7 +25,7 @@ else()
         run_result_unused
         compile_result_unused
         ${CMAKE_CURRENT_BINARY_DIR}
-        ${CMAKE_CURRENT_SOURCE_DIR}/cmake/detect-arch.c
+        ${CMAKE_CURRENT_LIST_DIR}/detect-arch.c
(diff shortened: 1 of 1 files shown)


Dead2:
Could you explain the use-case here? It is a little unclear what problem you are trying to solve here.

mtl1979:
@Dead2 The change allows using `include()` instead of `add_subdirectory()` when parent project wants to use functions declared by build system of zlib-ng (CMakeLists.txt and any files directly or indirectly included from it).

I'm not completely sure if we would want to allow that.

phprus:
This change will allow create a configuration like this:
```
project(BaseProject)

include(zlib-ng/cmake/detect-arch.cmake)

if(BASEARCH_X86_FOUND)
    # x86 project options
    ...
endif()

...

add_subdirectory(zlib-ng)

...

target_link_libraries(exe
    PRIVATE
        zlibstatic
)
```

To reuse the architecture detection code from zlib-ng.

PR #981, opened 2021-05-29
Use latest version of llvm-cov to prevent code coverage version mismatch

nmoinvaz:
See #977.
```
Unexpected version: *804.
Invalid .gcno File!
```


nmoinvaz:
I can probably count on my hand how many times code coverage check has succeeded.

Issue #979, opened 2021-05-29
SSE2/3/4/AVX2 optimisations cause crc32 error in pigz


cris-b:
This is on an Openmandriva 4.3 system building libz-ng and pigz from source (the system defaults to clang but have tried with gcc without change).

Recently we switched to zlib-ng as the system zlib, but this for me caused a problem with pigz (which we use as the default gzip).

I have to build zlib-ng with the following options (or WITH_OPTIM=OFF, but these are narrowed down):

-DWITH_SSSE3=OFF -DWITH_SSE4=OFF -DWITH_SSE2=OFF -DWITH_AVX2=OFF

or I get

pigz: skipping: pigz-2.5.tar.gz: corrupted -- crc32 mismatch

when uncompressing certain binaries.  (the pigz official source tar is one).

I've tried rebuilding zlib-ng with all compiler optimisations off (pgo/lto etc) and similar with pigz.  The only thing I can do to get it to work is with the above options off.

This on a haswell era Intel I7.

It should also be noted that uncompressing the archive using libarchive (via the libarchive tar implementation) which uses zlib-ng does not give an error.  

Have tried pigz 2.5/2.6, zlib-ng 2.0.2/2.0.3

gdevenyi:
Probably a good idea to provide configure and build logs for both zlib-ng and pigz as well as the versions and the actual complete output which contains the error.

cris-b:
[pigz.build.log](https://github.com/zlib-ng/zlib-ng/files/6565030/pigz.build.log)
[zlib-ng.build.log](https://github.com/zlib-ng/zlib-ng/files/6565032/zlib-ng.build.log)


output:

[crispin@yossarian pigz]$ wget http://www.zlib.net/pigz/pigz-2.6.tar.gz
--2021-05-29 17:57:25--  http://www.zlib.net/pigz/pigz-2.6.tar.gz
Resolving www.zlib.net... 85.187.148.2
Connecting to www.zlib.net|85.187.148.2|:80... connected.
HTTP request sent, awaiting response... 200 OK
Length: 106840 (104K) [application/x-gzip]
Saving to: ‘pigz-2.6.tar.gz’

pigz-2.6.tar.gz                                       100%[=======================================================================================================================>] 104.34K   170KB/s    in 0.6s    

2021-05-29 17:57:26 (170 KB/s) - ‘pigz-2.6.tar.gz’ saved [106840/106840]

[...]

PR #982, opened 2021-06-01
[CHUNKCOPY_SAFE] Fix off-by-one error

mtl1979:
* When chunk size is more than 8 bytes, the comparison logic failed if remaining buffer length was one less than chunk size.

1. For example when using AVX2 with buffer length of 31 bytes, it would write past end of buffer.
2. When `len` is less than remaining buffer length, `CHUNKCOPY_SAFE` now uses smaller chunks to avoid writing past of end of buffer.

### DIFF
--- a/chunkset_tpl.h
+++ b/chunkset_tpl.h
@@ -161,10 +161,11 @@ Z_INTERNAL uint8_t* CHUNKMEMSET(uint8_t *out, unsigned dist, unsigned len) {
 }
 
 Z_INTERNAL uint8_t* CHUNKMEMSET_SAFE(uint8_t *out, unsigned dist, unsigned len, unsigned left) {
+    len = MIN(len, left);
     if (left < (unsigned)(3 * sizeof(chunk_t))) {
+        uint8_t *from = out - dist;
         while (len > 0) {
-            *out = *(out - dist);
-            out++;
+            *out++ = *from++;
             --len;
         }
         return out;
--- a/chunkset_tpl.h
+++ b/chunkset_tpl.h
@@ -37,7 +37,8 @@ Z_INTERNAL uint8_t* CHUNKCOPY(uint8_t *out, uint8_t const *from, unsigned len) {
 
 /* Behave like chunkcopy, but avoid writing beyond of legal output. */
 Z_INTERNAL uint8_t* CHUNKCOPY_SAFE(uint8_t *out, uint8_t const *from, unsigned len, uint8_t *safe) {
-    if ((safe - out) < (ptrdiff_t)sizeof(chunk_t)) {
+    len = MIN(len, safe - out + 1);
+    if (len < sizeof(chunk_t)) {
         int32_t use_chunk16 = sizeof(chunk_t) > 16 && (len & 16);
         if (use_chunk16) {
             memcpy(out, from, 16);
--- a/inffast.c
+++ b/inffast.c
@@ -260,15 +260,11 @@ void Z_INTERNAL zng_inflate_fast(PREFIX3(stream) *strm, unsigned long start) {
                         out = functable.chunkcopy_safe(out, from, len, safe);
                     }
                 } else {
-                    /* Whole reference is in range of current output.  No range checks are
-                       necessary because we start with room for at least 258 bytes of output,
-                       so unroll and roundoff operations can write beyond `out+len` so long
-                       as they stay within 258 bytes of `out`.
-                    */
+                    /* Whole reference is in range of current output. */
                     if (dist >= len || dist >= state->chunksize)
-                        out = functable.chunkcopy(out, out - dist, len);
+                        out = functable.chunkcopy_safe(out, out - dist, len, safe);
                     else
-                        out = functable.chunkmemset(out, dist, len);
+                        out = functable.chunkmemset_safe(out, dist, len, safe - out + 1);
                 }
             } else if ((op & 64) == 0) {          /* 2nd level distance code */
                 here = dcode + here->val + BITS(op);


mtl1979:
@nmoinvaz

nmoinvaz:
It does not fix #979. I'm not sure I understand the off-by-one error. `if (31 < sizeof(chunk_t))` (only 31 bytes would be written) - 16, 8, 4, 2, 1 = 31. However I do like the idea of moving `MIN` to zbuild.h so it can be used by other sources.

mtl1979:
@nmoinvaz I think there might be error in code calling `CHUNKCOPY_SAFE` somewhere as otherwise the MIN() would be unnecessary... If I remove the MIN()... It fails with buffer size of 31, trying to write 32 bytes.

nmoinvaz:
I don't think there is any off by one issue. Having the `MIN()` is good. But to properly do the limit this is what code I was playing around with. It didn't make a difference with the #979 but it does prevent limiting of the output.

image

I haven't tested it very well. It always makes sure that `CHUNKCOPY` has a multiple of chunksize. Because it is possible in the original function for say `(safe - out) = 43`. The original function  would then write 32 * 2 or 64 bytes or past the safe buffer. It does probably decrease performance a bit because it has to do multiple memcpy's to get the initial alignment.

What do you think?

nmoinvaz:
It also has to be combined with this change in `chunkmemset`:

image

This may be the off by one issue you are noticing. I was able to find this one by adding these "asserts":

image

mtl1979:
I was thinking of adding
```
len = MIN(len, safe - out + 1);
```

to top of `CHUNKCOPY_SAFE`. That would make the comparison cleaner...

nmoinvaz:
@mtl1979 are you able to make the `MIN()` move into its own PR? There are many places in inflate code that could also benefit from it.

mtl1979:
@nmoinvaz I don't think separating the move of MIN() is hard...

mtl1979:
Rebased.

nmoinvaz:
I still don't understand why it is off by one.

mtl1979:
@nmoinvaz

If you have buffer of 31 bytes, it is indexed from 0 to 30... That is where the `- 1` comes from in one of the calling functions. I tested it by removing the `+ 1`inside `MIN()` and tests started failing. It is only relevant when both `len` and the safe length are less than size of `chunk_t` but not equal.

Safe length can calculated by formula 30 (end index) - 0 (start index) + 1 = 31.

In this case it comes to the comparison, which is `<` and not `<=`... It has to fall through if safe length is same as size of `chunk_t`.

Dead2:
Please rebase, so we get the new CI-tests too :)

nmoinvaz:
Benchmarks should be done as well.

mtl1979:
I'm pretty sure there will be speed drop, but @nmoinvaz is changing related code and I've already posted some benchmarks of improvements on that pull request... Basically if distance is less than pointer size, it is faster to do copy byte-by-byte instead of trying to use longer 2 (or 4) byte copies and unroll... If distance is at least 8 bytes on 64-bit machine, there is benefit in unrolling first.

nmoinvaz:
Even with that code, the performance suffers compared to HEAD. In my comment on #979 I mentioned that we could also do a detection based on whether or not the same buffer is used and in instances where same buffer is used, then we force safe versions, otherwise we don't use safe versions for the speed improvement.

mtl1979:
@nmoinvaz I didn't expect it to be magical solution that counteracts whole speed drop... What I was referring to is that this pull request is solely about fixing a single issue and that making improvements that counteract the speed drop would require waiting for merging your pull request first... Doesn't really make sense to do benchmarks until all possible improvements are made.

Obviously we can use both "unsafe" versions if remaining buffer length is at least three times the chunk size due to stricter requirement of chunkmemset. I'm just not sure if adding extra branches in speed critical part of the code is a good thing..

Dead2:
I am considering this PR for Stable as it is pretty small. But I want to postpone #987 to Next Devel since it is pretty much a refactoring of a lot of code and not a bugfix.
If there are smaller self-contained changes we should backport to Stable from #987, I'd like to hear about them.

We can live with a little speed degradation in Stable, it is not optimal but lets prioritize correctness and stability over speed.

mtl1979:
@Dead2 I can't really comment about #987 until it is finished... It started as a small patch like this one but now it's pretty much complete rewrite of the related functions. When it is finished, it might be even worth the effort trying to split it as two or more separate pull requests.

nmoinvaz:
Can you split the commit into two? One with the off by one error that you found. The other commit with the fix for inflateBack (using safe functions) which I found. On the second commit can you add me as a co-author since it took me several days to hunt it down and come up with possible solutions for?

After that I will try and benchmark each commit in this PR so we can qualify the speed degradation.

mtl1979:
@nmoinvaz Done.

nmoinvaz:
Thanks. Those are better commit messages too. Nice!

nmoinvaz:
## ZLIB-NG (HEAD)

```
 Tool: minigzip-ng-commit-0.exe Levels: 1-9

 1     48.014%      1.119/1.135/1.143/0.006        0.781/0.793/0.799/0.004      111,512,026
 2     35.391%      1.891/1.919/1.934/0.011        0.789/0.812/0.819/0.006       82,195,416
 3     34.039%      2.323/2.354/2.368/0.013        0.784/0.793/0.798/0.004       79,055,001
 4     32.751%      2.782/2.811/2.825/0.011        0.761/0.769/0.776/0.005       76,065,070
 5     32.484%      3.001/3.030/3.048/0.015        0.750/0.766/0.773/0.005       75,443,521
 6     32.318%      3.635/3.671/3.693/0.016        0.752/0.764/0.769/0.004       75,057,363
 7     32.061%      4.896/4.934/4.969/0.025        0.750/0.762/0.770/0.006       74,461,300
 8     31.978%      7.486/7.537/7.591/0.028        0.750/0.760/0.767/0.005       74,269,032
 9     31.961%   10.148/10.211/10.270/0.031        0.751/0.762/0.769/0.005       74,230,342

 avg1  34.555%                        4.178                          0.776
 tot                               1128.053                        209.449      722,289,071
```

## ZLIB-NG (COMMIT 1)

```
 Tool: minigzip-ng-commit-1.exe Levels: 1-9

 1     48.014%      1.123/1.142/1.150/0.007        0.795/0.811/0.817/0.006      111,512,026
 2     35.391%      1.881/1.902/1.920/0.012        0.843/0.856/0.864/0.006       82,195,416
 3     34.039%      2.295/2.341/2.360/0.015        0.832/0.848/0.855/0.006       79,055,001
 4     32.751%      2.755/2.783/2.803/0.013        0.801/0.815/0.822/0.006       76,065,070
 5     32.484%      2.968/3.008/3.032/0.016        0.799/0.814/0.820/0.005       75,443,521
 6     32.318%      3.604/3.641/3.668/0.020        0.801/0.815/0.825/0.006       75,057,363
 7     32.061%      4.847/4.904/4.943/0.027        0.804/0.814/0.821/0.005       74,461,300
 8     31.978%      7.464/7.510/7.569/0.035        0.802/0.814/0.823/0.005       74,269,032
 9     31.961%   10.110/10.193/10.277/0.052        0.796/0.812/0.820/0.008       74,230,342

 avg1  34.555%                        4.158                          0.822
 tot                               1122.726                        221.977      722,289,071
```
## ZLIB-NG (COMMIT 2)
```
 Tool: minigzip-ng-commit-2.exe Levels: 1-9

 1     48.014%      1.133/1.151/1.161/0.008        0.923/0.940/0.947/0.007      111,512,026
 2     35.391%      1.840/1.862/1.873/0.011        0.960/0.971/0.978/0.006       82,195,416
 3     34.039%      2.266/2.290/2.304/0.011        0.932/0.947/0.954/0.005       79,055,001
 4     32.751%      2.723/2.753/2.772/0.016        0.908/0.921/0.927/0.005       76,065,070
 5     32.484%      2.944/2.977/2.996/0.015        0.900/0.913/0.921/0.006       75,443,521
 6     32.318%      3.585/3.613/3.633/0.016        0.902/0.910/0.918/0.004       75,057,363
 7     32.061%      4.925/4.971/5.007/0.025        0.890/0.907/0.913/0.005       74,461,300
 8     31.978%      7.529/7.581/7.636/0.037        0.892/0.905/0.913/0.006       74,269,032
 9     31.961%   10.172/10.265/10.338/0.056        0.886/0.902/0.910/0.006       74,230,342

 avg1  34.555%                        4.163                          0.924
 tot                               1123.917                        249.539      722,289,071
```

mtl1979:
So about 19% slower... That sounds a lot, but I expected it as chunkmemset_safe() isn't currently optimized at all...

chunkset_tpl.h
```diff
@@ -163,10 +163,21 @@ Z_INTERNAL uint8_t* CHUNKMEMSET(uint8_t *out, unsigned dist, unsigned len) {
 Z_INTERNAL uint8_t* CHUNKMEMSET_SAFE(uint8_t *out, unsigned dist, unsigned len, unsigned left) {
     if (left < (unsigned)(3 * sizeof(chunk_t))) {
         uint8_t *from = out - dist;
-        while (len > 0) {
-            *out++ = *from++;
-            --len;
+        uint8_t *end = out + left - 1;
+        if (dist < sizeof(void *)) {
+            while (len > 0) {
```
nmoinvaz, line 168: I think this can be simplified as `while (out < end)`, or something similar.

nmoinvaz:
I don't think we will be able to optimize our way back to -19% with changes to chunkmemset or chunkcopy. Also it will start to overlap with what I have done in my PR. I had all these fixes/optimizations in my PR and never was able to get back to where we were.

I think the only way to get the speed back is to detect in inflate_fast if same buffer is used for "out" and "from" and if so then use safe functions and if not then use unsafe functions.

mtl1979:
@nmoinvaz I noticed that... Been doing benchmarks for last 2 hours or so and the difference is within error margin...

mtl1979:
On my benchmark, difference between unrolling and not unrolling in chunkmemset_safe() was 23 ms using 211957760 bytes long file.

nmoinvaz:
@mtl1979 you may be able to cherry pick https://github.com/zlib-ng/zlib-ng/pull/987/commits/f5325ab3b4ba03da2d713140e18e25f8e147c214 for the speed gains.

nmoinvaz:
I think this "off-by-one" bug should be its own PR with 1 commit and anything else should be a separate PR.

chunkset_tpl.h
```diff
@@ -37,7 +37,8 @@ Z_INTERNAL uint8_t* CHUNKCOPY(uint8_t *out, uint8_t const *from, unsigned len) {
 
 /* Behave like chunkcopy, but avoid writing beyond of legal output. */
 Z_INTERNAL uint8_t* CHUNKCOPY_SAFE(uint8_t *out, uint8_t const *from, unsigned len, uint8_t *safe) {
-    if ((safe - out) < (ptrdiff_t)sizeof(chunk_t)) {
+    len = MIN(len, safe - out + 1);
+    if (len < (ptrdiff_t)sizeof(chunk_t)) {
```
nmoinvaz: I don't think we need `(ptrdiff_t)` cast here.

mtl1979 (commented):
> @mtl1979 you may be able to cherry pick [f5325ab](https://github.com/zlib-ng/zlib-ng/commit/f5325ab3b4ba03da2d713140e18e25f8e147c214) for the speed gains.

I don't see any speed gain on my machine... The extra branch is in speed sensitive location, so it actually gets slower.

mtl1979:
> I think this "off-by-one" bug should be its own PR with 1 commit and anything else should be a separate PR.

Maybe... I just wanted to keep them together to test for any side-effects... If I separate, the "remaining" part must be merged in before the "off-by-one" fix. Then I can just rebase over for a really clean diff.

chunkset_tpl.h
mtl1979: `sizeof` should return `size_t` which is unsigned type... on the other hand, pointer math with no explicit cast will return `int` which is signed type. Unlike the standard says, modern compilers will actually cast both sides of comparison to same type if one side uses explicit cast as long as signedness match.

Dead2:
LGTM, but I am not so steady on the intricacies of the chunkset code, so I have not verified its correctness but rather its scope.

mtl1979:
@Dead2 From the start nobody was really sure about correctness of the chunkset code, they just assumed it was correct because it didn't crash or fail any CI tests back then... After adding pigz test, this is the first pull request that doesn't fail it and does not introduce new failures either. Codecov test fails only because this pull request replaces direct calls to two functions with indirect calls through another set of two functions.

It's really hard to spot off-by-one errors in any code and even more when there is guard bytes added by user application instead of the C library.

Dead2:
@mtl1979 I am aware. I meant that I'll need to wait for @nmoinvaz or someone else to approve as well. 😄

mtl1979:
@Dead2 I know... Controversial patches require two long-time collaborators... I already asked @nmoinvaz to recheck ;)

nmoinvaz (approved):

Dead2:
### Baseline 5cfb7f15e32bc16ec4cca82b9c094bb886456e4c
```
 111264    1296      32  112592   1b7d0 libz-ng.so.2

 1     57.780%      1.874/1.898/1.908/0.008        1.057/1.095/1.107/0.010       90,924,650
 2     43.876%      2.032/2.045/2.051/0.005        0.690/0.702/0.708/0.005       41,427,218
 3     42.490%      2.093/2.109/2.118/0.007        0.562/0.570/0.574/0.003       33,432,016
 4     41.467%      2.021/2.042/2.052/0.009        0.432/0.443/0.448/0.004       26,101,482
 5     41.212%      2.158/2.199/2.209/0.010        0.432/0.441/0.446/0.004       25,940,920
 6     41.039%      1.922/1.933/1.940/0.005        0.321/0.331/0.336/0.004       19,374,021
 7     40.781%      1.658/1.666/1.671/0.004        0.210/0.218/0.221/0.003       12,834,980
 8     40.706%      2.074/2.084/2.090/0.006        0.212/0.217/0.221/0.002       12,811,166
 9     40.695%      2.264/2.280/2.286/0.005        0.209/0.218/0.222/0.003       12,807,941

 avg1  43.339%                        2.028                          0.471
 tot                                547.633                        127.059      275,654,394
```

### PR #982 eb212936dd6da4222f566ea82f8ad48665aa614c
```
 111456    1296      32  112784   1b890 libz-ng.so.2

 1     57.780%      1.882/1.902/1.910/0.006        1.263/1.280/1.285/0.005       90,924,650
 2     43.876%      2.029/2.041/2.047/0.005        0.814/0.827/0.833/0.004       41,427,218
 3     42.490%      2.091/2.109/2.116/0.006        0.657/0.677/0.683/0.006       33,432,016
 4     41.467%      2.021/2.041/2.049/0.007        0.517/0.527/0.533/0.004       26,101,482
 5     41.212%      2.133/2.195/2.207/0.013        0.517/0.525/0.530/0.004       25,940,920
 6     41.039%      1.916/1.930/1.939/0.006        0.386/0.392/0.395/0.002       19,374,021
 7     40.781%      1.654/1.663/1.669/0.005        0.249/0.259/0.263/0.003       12,834,980
 8     40.706%      2.071/2.086/2.093/0.006        0.243/0.258/0.262/0.004       12,811,166
 9     40.695%      2.261/2.279/2.287/0.008        0.251/0.258/0.261/0.003       12,807,941

 avg1  43.339%                        2.027                          0.556
 tot                                547.381                        150.102      275,654,394
```

nmoinvaz:
I am getting some warnings now on MSVC:
```
C:\Users\Nathan\Source\zlib-ng\chunkset_tpl.h(40,35): warning C4244: '=': conversion from '__int64' to 'unsigned int',
possible loss of data [C:\Users\Nathan\Source\zlib-ng\build-x\zlib.vcxproj]
C:\Users\Nathan\Source\zlib-ng\inffast.c(267,88): warning C4244: 'function': conversion from '__int64' to 'unsigned int
', possible loss of data [C:\Users\Nathan\Source\zlib-ng\build-x\zlib.vcxproj]
C:\Users\Nathan\Source\zlib-ng\chunkset_tpl.h(40,35): warning C4244: '=': conversion from '__int64' to 'unsigned int',
possible loss of data [C:\Users\Nathan\Source\zlib-ng\build-x\zlib.vcxproj]
C:\Users\Nathan\Source\zlib-ng\chunkset_tpl.h(40,35): warning C4244: '=': conversion from '__int64' to 'unsigned int',
possible loss of data [C:\Users\Nathan\Source\zlib-ng\build-x\zlib.vcxproj]
```

mtl1979:
@nmoinvaz That's because I didn't change the function signatures to use `z_size_t` instead of `unsigned int`...

mtl1979:
Casting the pointer difference to `unsigned int` would have introduced possible wraparound bug, as the result is allowed to be `-1`. Any negative difference except `-1` is expected to crash.

nmoinvaz:
How is it allowed to be -1?

mtl1979:
@nmoinvaz Because `1` is added to the pointer difference, `-1` as difference means the passed-in buffer is full. Because it is now using the safe version of both chunkcopy and chunkmemset, it is allowed (even if it doesn't ever happen in normal circumstances) and essentially no-op. `-2` as difference would result in all bits set in safe length, which would most likely overflow.

PR #983, opened 2021-06-02
Use SET_BAD macro in inflateBack.

nmoinvaz:
It looks like some instances were missed when the macro was created. I noticed it when doing BeyondCompare with inflate.c and infback.c.


Dead2:
Nice. I must have completely forgotten to merge the infback changes 😄

PR #985, opened 2021-06-04
Added pigz to unit tests

nmoinvaz:
This PR adds:

* New GitHub Actions CI used to test `pigz`.
* New CMakeLists.txt that can be used to build `pigz` against zlib-ng or any other zlib variant.
  * Adds tests for `pigz` compression/decompression against all the files in `test/*` and `test/corpora/**` (similar to what zlib-ng does with minigzip)
    * This should also provide coverage for #536 since I have also added those brain images to the _corpora_ repository.
  * Adds test specifically for pigz-2.6.tar.gz mentioned in #979. This means I have added this .gz file to the repository which is ~100KB.


nmoinvaz:
If we add the files in https://github.com/neurolabusc/zlib-bench/tree/master/corpus to https://github.com/zlib-ng/corpora repository then we can also provide test coverage for old pigz issue #536. @Dead2 what do you think? Those files are 18mb and 34mb.

mtl1979:
I'm all for adding tests that make sure we don't introduce regressions or tests that confirm we have an crash issue.

nmoinvaz:
I might just create a new repository _corpora-lg_ for larger test corpora.

mtl1979:
@nmoinvaz I usually just clone the original repository and then rename it... That's a lot easier than trying to combine files from multiple repositories and preserving edit history to conform to copyrights.

nmoinvaz:
I have added a CI test for no threads to be able isolate threaded compression and also a CI test for no optimizations to isolate those as well.

nmoinvaz:
@mtl1979 it could be as simple as a GitHub CI action that will download the new repo (same as what is done with corpora). CMake script picks up everything in `test/data/**`.

mtl1979:
I'm not big fan of supporting pigz but if it helps finding issues involving corner cases, then I see no reason to "oppose"...

Keeping own "version" of repositories is always a good thing as the original repository might go offline and break our build system...

nmoinvaz:
If we ended up include the pigz sources in our repository then we will likely have to support it, that is why I pull the official sources from madler/pigz. We could fork that repo, but I don't want to have to keep it up to date.

nmoinvaz:
Regarding neuro images, I have decided to just added them to corpora repository (also I don't want to manage yet another repo). Even though it is an extra 80mb it will be good to have it tested every where because image data such as that is unique and good for RLE tests. So now this PR should provide coverage for  #536.

Dead2:
Looks good to me. Is this ready to merge?

nmoinvaz:
Yes it is. The CI tests that fail, fail because our code fails them.

PR #989, opened 2021-06-09
Fixed two bugs in zlib-ng build system.

ilyakurdyukov:


nmoinvaz:
First commit is also fixed in #906.

mtl1979:
@nmoinvaz Yeah... But that pull request is a draft and unlikely to be merged anytime soon...

nmoinvaz:
It was not a draft when it was written and only has a conflict because changes that it had were merged before it. The further we go along without merging it the more other people run into same issues.

mtl1979:
@nmoinvaz You are free to kick @Dead2 in the groin, because he is the one who decides what gets merged as soon as possible and what gets delayed to next big release... My personal observation is that smaller the pull request is, more likely it is going to get merged first...

As I've said before many times, for "controversial" pull requests that require two long-time contributors to review and are not written by myself, I try to review as soon as possible, so that they don't get delayed for too long.

I rarely ever do benchmarks, because that would require stopping all processes on this machine and that means everything stops in the company I work for, because I use my development PC to host daemons that are not ready yet for running on production servers.

Dead2:
I agree, I really want to get the next stable out and then open up develop and get the huge backlog merged in.
But in a stable release, I don't want to introduce bigger changes than necessary, as that introduces a higher level of risk.
Now we just need to get that inflate bug fixed so we can get the next stable out the door.

mtl1979:
> Now we just need to get that inflate bug fixed so we can get the next stable out the door.

If we leave all the possible speedups to next pull request, that inflate bug fix is pretty much ready now... There is enough safety checks now that the safe versions of the functions are actually safe.

PR #991, opened 2021-06-11
Open up develop branch for non-stable commits.

Dead2:
Start 2.1.0 develop cycle.
Replace ZLIBNG_VERNUM with a larger one with space for more accurate information.
Replace ZLIBNG_VER_SUBREVISION with ZLIBNG_VER_MODIFIED and ZLIBNG_VER_STATUS.


zlib.h
```diff
@@ -46,12 +46,13 @@
...
-#define ZLIBNG_VERNUM 0x2030
+#define ZLIBNG_VERSION "2.1.0.devel"
+#define ZLIBNG_VERNUM 0x02010000L   /* MMNNRRMS: major minor revision modified status */
 #define ZLIBNG_VER_MAJOR 2
-#define ZLIBNG_VER_MINOR 0
-#define ZLIBNG_VER_REVISION 3
-#define ZLIBNG_VER_SUBREVISION 0
+#define ZLIBNG_VER_MINOR 1
+#define ZLIBNG_VER_REVISION 0
+#define ZLIBNG_VER_MODIFIED 0       /* non-null if modified externally from zlib-ng */
```
mtl1979: Should be "non-zero".

zlib-ng.h
```diff
@@ -44,12 +44,13 @@
...
+#define ZLIBNG_VERSION "2.1.0.devel"
+#define ZLIBNG_VERNUM 0x02010000L   /* MMNNRRMS: major minor revision modified status */
 #define ZLIBNG_VER_MAJOR 2
-#define ZLIBNG_VER_MINOR 0
-#define ZLIBNG_VER_REVISION 3
-#define ZLIBNG_VER_SUBREVISION 0
+#define ZLIBNG_VER_MINOR 1
+#define ZLIBNG_VER_REVISION 0
+#define ZLIBNG_VER_MODIFIED 0       /* non-null if modified externally from zlib-ng */
+#define ZLIBNG_VER_STATUS 0         /* 0=devel, 1-E=beta, F=Release */
```
mtl1979, line 52: Might make more sense to have `ZLIBNG_VER_MODIFIED` as last element in version string.

Dead2: Yes, perhaps so.
I considered this back and forth when making the change, and both ways makes sense for different cases.
It kind of depends on what kind of modification we are talking, a patched version made by a distro? Makes sense to put it last. A more serious fork of zlib-ng? Might make sense to put it where it is now or even before major.

There really is no good way of mixing modified (from a fork) releases with official releases, for that it'd need to be an epoch thing due to the potential for API/ABI differences. Should we support such a thing? Probably not.

mtl1979: Personally I think we should not "support" more serious forks... Those should be handled the same way as difference between stock zlib and zlib-ng.

mtl1979: API/ABI breakakes can happen for various reasons... What matters is the question if the change will get upstreamed to zlib-ng or not... if it gets, then the change modifies `ZLIBNG_VER_MINOR` and any change to `ZLIBNG_VER_MODIFIED` is irrelevant.

This is my opinion as former package maintainer for RedHat.

Dead2:
Changed example.c to always print out the zlib-ng version, as it would likely not make sense to build it against stock zlib. Fixes the issue with the new ZLIBNG_VERNUM differing in length compared to the zlib one.

mtl1979:
@Dead2 I was kinda waiting for you to notice that issue... I didn't have any rush as I wanted to wait for CI to finish to see if there is any new warnings...

nmoinvaz:
LGTM. I like that you have kept the version number define and went with OpenSSL format. minizip-ng also uses openSSL format.

nmoinvaz (approved):

mtl1979 (approved):
LGTM.

PR #995, opened 2021-06-13
Change to C11 and add TLS support

Dead2:
In order to ensure functable consistency we need support for Thread Local Storage, C11 provides this but it is optional.
This PR adds detection of C11 and detection of a few of the most common types of defining TLS support for a variable.

### DIFF
--- a/zbuild.h
+++ b/zbuild.h
@@ -15,6 +15,20 @@
 #  endif
 #endif
 
+/* Determine compiler support for TLS */
+#ifndef Z_TLS
+#  if __STDC_VERSION__ >= 201112 && !defined __STDC_NO_THREADS__
+#    define Z_TLS _Thread_local
+#  elif defined(__GNUC__) || defined(__SUNPRO_C)
+#    define Z_TLS __thread
+#  elif defined(_WIN32) && (defined(_MSC_VER) || defined(__ICL))
+#    define Z_TLS __declspec(thread)
+#  else
+#    warning Unable to detect Thread Local Storage support.
+#    define Z_TLS
+#  endif
+#endif
+
 /* This has to be first include that defines any types */
 #if defined(_MSC_VER)
 #  if defined(_WIN64)
--- a/zbuild.h
+++ b/zbuild.h
@@ -1,6 +1,20 @@
 #ifndef _ZBUILD_H
 #define _ZBUILD_H
 
+/* Determine compiler version of C Standard */
+#ifdef __STDC_VERSION__
+#  if __STDC_VERSION__ >= 199901L
+#    ifndef STDC99
+#      define STDC99
+#    endif
+#  endif
+#  if __STDC_VERSION__ >= 201112L
+#    ifndef STDC11
+#      define STDC11
+#    endif
+#  endif
+#endif
+
 /* This has to be first include that defines any types */
 #if defined(_MSC_VER)
 #  if defined(_WIN64)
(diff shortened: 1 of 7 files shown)


Dead2:
NMake fails with `functable.h(27): error C2054: expected '(' to follow 'Z_TLS'`, any ideas as to how to resolve that one?
Edit: Fixed.

zbuild.h
```diff
@@ -1,6 +1,34 @@
...
+
+/* Determine compiler support for TLS */
+#ifndef Z_TLS
+#  if __STDC_VERSION__ >= 201112 && !defined __STDC_NO_THREADS__
+#    define Z_TLS _Thread_local
+#  elif defined(__GNUC__) || defined(__SUNPRO_C)
+#    define Z_TLS __thread
+#  elif defined(_WIN32) && (defined(_MSC_VER) || defined(__ICL))
+#    define thread_local __declspec(thread)
+#  else
```
phprus, line 26: Macro name is ```thread_local``` or ```Z_TLS```?

Dead2: Haha, thanks. I was blind to that mistake 😄

nmoinvaz:
@Dead2 I think perhaps all 3 nmake Makefiles need `-std=c11` in CFLAGS variable?

https://github.com/zlib-ng/zlib-ng/blob/1118fa8ecc0b5522738b8d7781c813d47f607394/win32/Makefile.msc#L25
https://github.com/zlib-ng/zlib-ng/blob/1118fa8ecc0b5522738b8d7781c813d47f607394/win32/Makefile.arm#L25
https://github.com/zlib-ng/zlib-ng/blob/1118fa8ecc0b5522738b8d7781c813d47f607394/win32/Makefile.a64#L25

nmoinvaz (commented):
I don't see anywhere, where these macros are used in the code. I think it would be a good idea to use them where needed to see if extra compilation issues happen when in use.

Overall this looks good and should be a good patch for people who need this.

zbuild.h
```diff
@@ -1,6 +1,34 @@
 #ifndef _ZBUILD_H
 #define _ZBUILD_H
 
+/* Determine compiler version of C Standard */
+#ifdef __STDC_VERSION__
+#  if __STDC_VERSION__ >= 199901L
+#    ifndef STDC99
+#      define STDC99
```
nmoinvaz, line 8: `STDC99` is also *defined* in zconf-ng.h.in and zconf.h, so do we also need to define `STDC11` there too?

Dead2:
> @Dead2 I think perhaps all 3 nmake Makefiles need `-std=c11` in CFLAGS variable?

Actually, I was thinking not.
Since it seems for some reason a lot of people are using older MSVC releases, and C11 support was only introduced in MSVC in 2019 or so, so it is very fresh compared to GCC/Clang etc.
With the workaround for TLS, it should be fine, at least until we introduce something else that requires C11.

> STDC99 is also defined in zconf-ng.h.in and zconf.h, so do we also need to define STDC11 there too?

STDC99 is no longer being defined in zconf, this commit removes that and moves it to zbuild instead. zconf gets included into applications including the zlib header file, so I think zbuild is more appropriate for this.

> I don't see anywhere, where these macros are used in the code.

STDC99 does get used in the code at least (gzguts), but there are no users of STDC11 yet.
Z_TLS macro does get used by functable.

I was thinking this might possibly get used for adding improved aligned malloc detection.

nmoinvaz:
I can take a look at nmake later today. I'm primarily on Windows.

Dead2:
> I can take a look at nmake later today. I'm primarily on Windows.

nmake works now, it was a stupid error on my part that made CI fail earlier. Or was there something else you were thinking about?

nmoinvaz:
No if you got it compiling then that's it.

nmoinvaz:
LGTM.

PR #996, opened 2021-06-13
Improve sanitizer selection and add support for selecting Thread sanitizer

Dead2:
Change WITH_SANITIZER to be a multi-option parameter (for ccmake etc).
Add support for selecting Thread sanitizer.


nmoinvaz:
I think the reason it was not a multiple option before was because some of the sanitizers don't play well together. I have experienced issues before both on this project and others where running multiple sanitizers fails to return output any errors.

Dead2:
@nmoinvaz This option does not allow selecting multiple at the same time, but it allows you to get a drop-down list in ccmake so you can select from the list of options instead of just a ON/OFF toggle (that probably doesn't even work, since ON does not map to any of the sanitizers).
Unfortunately it does not also give an error/warning if you select something that is not in the list 😄

nmoinvaz (approved):

Issue #418, opened 2019-09-19
Warnings in checksum_fuzzer.c and example.c


Dead2:
https://travis-ci.org/zlib-ng/zlib-ng/jobs/587160114
```
[ 73%] Linking C static library libz.a
[ 73%] Built target zlibstatic
Scanning dependencies of target checksum_fuzzer
[ 75%] Building C object CMakeFiles/checksum_fuzzer.dir/test/fuzz/checksum_fuzzer.c.o
/home/travis/build/zlib-ng/zlib-ng/test/fuzz/checksum_fuzzer.c:46:3: warning: Value stored to 'crc1' is never read
  crc1 = PREFIX(crc32_z)(crc1, data + offset, dataLen % buffSize);
  ^      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/travis/build/zlib-ng/zlib-ng/test/fuzz/checksum_fuzzer.c:48:3: warning: Value stored to 'crc2' is never read
  crc2 = PREFIX(crc32_z)(crc2, data, dataLen);
  ^      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/travis/build/zlib-ng/zlib-ng/test/fuzz/checksum_fuzzer.c:64:3: warning: Value stored to 'adler1' is never read
  adler1 = PREFIX(adler32_z)(adler1, data + offset, dataLen % buffSize);
  ^        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/travis/build/zlib-ng/zlib-ng/test/fuzz/checksum_fuzzer.c:66:3: warning: Value stored to 'adler2' is never read
  adler2 = PREFIX(adler32_z)(adler2, data, dataLen);
  ^        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
4 warnings generated.
[ 76%] Building C object CMakeFiles/checksum_fuzzer.dir/test/fuzz/standalone_fuzz_target_runner.c.o
[ 77%] Linking C executable checksum_fuzzer
[ 77%] Built target checksum_fuzzer
Scanning dependencies of target example64
[ 78%] Building C object CMakeFiles/example64.dir/test/example.c.o
[...]

PR #997, opened 2021-06-13
Added Z_UNUSED define for ignore unused variables.

nmoinvaz:
Previously mentioned in #418.
image

I have put it in zbuild.h for now and if we want to it can be moved. I can also rename it to `UNUSED()` if `Z_UNUSED` is not good name for zbuild.h.


mtl1979:
I think the name is acceptable as it follows our naming style. `zbuild.h` should contain all macros and defines that are supposed to be only used when building zlib-ng library itself or one of the test binaries in same repository. As such any other application or library that links against zlib-ng can define own version even with same name and it shouldn't ever conflict as no externally visible header of zlib-ng should include `zbuild.h`.

Dead2 (approved):
LGTM

PR #998, opened 2021-06-13
Fixed indentation in CMakeLists.txt.

nmoinvaz:
@mtl1979 I have made a separate PR for this since it has affected many parts of CMakeLists.txt.

PR #999, opened 2021-06-13
Show unavailable features in CMake feature summary.

nmoinvaz:
I know I was the one who combined these before, but I realized a while back it is useful to have it separate so that if a feature is not available on the platform, we can report it as disabled in the feature summary by setting its variable to `OFF`.

Please merge #998 first.


nmoinvaz:
Rebased.

PR #987, opened 2021-06-04
Improvements to inflate chunkset code

nmoinvaz:
This code makes some changes to the chunkset code and also fixes #979.

PR #1000, opened 2021-06-13
Reduce number of branches in partial chunk copy when no optimizations.

nmoinvaz:
Get rid of an extra branch for when optimizations are turned off. Because it `(len & 8)` would never be true if `(len < sizeof(chunk_t))`.

I chose to have a `CHUNK_SIZE` variable that we can reference - can't use `sizeof(chunk_t)` in `#if`. But if we want to have a `use_chunk8` variable, I would be okay to go with that route as well.

This is taken from my #987 PR.

### DIFF
--- a/arch/arm/chunkset_neon.c
+++ b/arch/arm/chunkset_neon.c
@@ -13,6 +13,8 @@
 
 typedef uint8x16_t chunk_t;
 
+#define CHUNK_SIZE 16
+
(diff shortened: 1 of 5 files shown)


mtl1979:
I think it's better to do it with preprocessor... That way compiler can optimize the code better.

Dead2 (approved):
LGTM

PR #1001, opened 2021-06-14
Also print compiler version when using configure

Dead2:
Since configure does not print out the compiler version, it has made it harder than necessary to debug some of the previously reported problems.
CMake does print compiler version already.
This commit makes configure also print out compiler version.

This is only run when a gcc-compatible compiler is detected, it might also have worked on more obscure compilers but I don't have any to test with.

### DIFF
--- a/configure
+++ b/configure
@@ -285,6 +285,16 @@ if test "$gcc" -eq 1 && ($cc $CFLAGS -c $test.c) >> configure.log 2>&1; then
   CC="$cc"
   CFLAGS="${CFLAGS} -std=c99"
 
+  # Detect compiler version
+  CCVER=$($CC --version)
+  if [ $? -eq 0 ]; then
+    IFS= read -r -a verarr <<< $CCVER
+    unset IFS
+    echo "Detected compiler version: ${verarr[0]}"
+  else
+    echo "Failed to detect compiler version."
+  fi
+
   # Re-check ARCH if the compiler is a cross-compiler.
   if $CC -print-multiarch 1> /dev/null 2>&1 && test -n "$($CC -print-multiarch)" 1> /dev/null 2>&1; then
       CC_ARCH=$($CC $CFLAGS -print-multiarch | sed 's/-.*//g')


mtl1979:
Fails CI.

nmoinvaz:
On macOS GCC it outputs:

Detected compiler version: Apple clang version 12.0.0 (clang-1200.0.32.29) Target: x86_64-apple-darwin19.6.0 Thread model: posix InstalledDir: /Applications/Xcode_12.4.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

Perhaps we only want the first line?

Dead2:
*sigh* another macos special case?
It was supposed to only print the first line, and I thought it was the most portable way to do it by using 'read' instead of for example 'head' or 'awk' that are both external and optional programs.
Ideas?

nmoinvaz:
@mtl1979 would probably know.

mtl1979:
@nmoinvaz Why would you think that I know that MacOS doesn't use ASCII 10 as line separator... It does use ASCII 13 instead ;)

Dead2:
@mtl1979 That is why I unset it, that sets it back to defaults according to the documentation.

mtl1979:
@Dead2 Default value and original value are not always the same. Even then, setting the field separator to empty string is wrong.

Dead2:
```The default value of IFS is space, tab, newline. (A three-character string.) If IFS is unset, it acts as though it were set to this default value. If IFS is set to an empty string (which is very different from unsetting it!) then no splitting will be performed.```

Setting IFS to an empty string effectively gives you a whole line.

**Comment by Dead2** [comment]
Verified to work on MacOS in CI run https://github.com/zlib-ng/zlib-ng/pull/1001/checks?check_run_id=2838696135
`Detected compiler version: Apple clang version 12.0.0 (clang-1200.0.32.29)`

**Comment by mtl1979** [comment]
@Dead2 It's always safer and cleaner to change IFS in subshell... that also means echoing any variable contents must be done in same subshell... When subshell exits, iFS is restored to previous value.

**Comment by Dead2** [comment]
@mtl1979 These make no reference to that being necessary, do you have a reference that explains that?
https://unix.stackexchange.com/questions/209123/understanding-ifs-read-r-line
https://mywiki.wooledge.org/IFS

**Comment by mtl1979** [comment]
@Dead2 Neither of those websites are credible sources. It's like saying everything written in Wikipedia is 100% fact, when it's 100% user contributions.

**Comment by Dead2** [comment]
Ok, I don't care that much about this one anyways.

**Comment by mtl1979** [comment]
```
       mapfile [-d delim] [-n count] [-O origin] [-s count] [-t] [-u fd] [-C callback] [-c quantum] [array]
       readarray [-d delim] [-n count] [-O origin] [-s count] [-t] [-u fd] [-C callback] [-c quantum] [array]
              Read lines from the standard input into the indexed array variable array, or from file descriptor fd if the -u option is supplied.  The variable MAPFILE is the default array.  Options, if supplied, have the following meanings:
              -d     The first character of delim is used to terminate each input line, rather than newline.
              -n     Copy at most count lines.  If count is 0, all lines are copied.
              -O     Begin assigning to array at index origin.  The default index is 0.
              -s     Discard the first count lines read.
              -t     Remove a trailing delim (default newline) from each line read.
              -u     Read lines from file descriptor fd instead of the standard input.
              -C     Evaluate callback each time quantum lines are read.  The -c option specifies quantum.
              -c     Specify the number of lines read between each call to callback.

              If -C is specified without -c, the default quantum is 5000.  When callback is evaluated, it is supplied the index of the next array element to be assigned and the line to be assigned to that element as  additional  arguments.   callback  is  evaluated
              after the line is read but before the array element is assigned.

              If not supplied with an explicit origin, mapfile will clear array before assigning to it.

              mapfile returns successfully unless an invalid option or option argument is supplied, array is invalid or unassignable, or if array is not an indexed array.
```

That is straight from bash manual... a lot cleaner way ;)

PR #1002, opened 2021-06-14
Fix MSVC warnings in hash_head_0.c test

iii-i:
@nmoinvaz I don't have MSVC at hand, but this should work - could you please give it a try?


nmoinvaz (approved):
This removes the warnings for me in MSVC.

PR #1009, opened 2021-06-19
[Power8] Add chunk*_power8.

mtl1979:
VSX optimized chunk copy and memset functions.

PR #1015, opened 2021-06-21
[functable] Add missing call to cpu_check_features().

mtl1979:
Separated from #1009.


nmoinvaz:
I think perhaps it is only necessary to call `cpu_check_features` in `chunksize()`. `chunksize` and `insert_string` are the first two functable entry points.

mtl1979:
@nmoinvaz I could comment each added line until tests fail again on PowerPC, but it would look really inconsistent if some of the stubs call the function and some don't... 

I don't know why PowerPC calls the functions in unexpected order or why other currently supported architectures don't fail the same tests.

mtl1979:
@Dead2

PR #1022, opened 2021-06-27
Fix Z_SOLO mode

berolinux:
Without this patch, #include <zlib.h> with Z_SOLO defined
(e.g. while building perl 5.34.0) fails because of use of
undefined types.


nmoinvaz:
I suggest only doing one PR and @Dead2 can cherry-pick the change into `stable` if he thinks it is necessary. I will just comment on the one targeting `develop` branch for now.

PR #1023, opened 2021-06-27
Fix Z_SOLO mode

berolinux:
Without this patch, #include <zlib.h> with Z_SOLO defined
(e.g. while building perl 5.34.0) fails because of use of
undefined types.

### DIFF
--- a/zlib.h
+++ b/zlib.h
@@ -1767,6 +1767,7 @@ Z_EXTERN int Z_EXPORT inflateBackInit_(z_stream *strm, int windowBits, unsigned 
                         inflateBackInit_((strm), (windowBits), (window), ZLIB_VERSION, (int)sizeof(z_stream))
 
 
+#ifndef Z_SOLO
 /* gzgetc() macro and its supporting function and exposed data structure.  Note
  * that the real internal state is much larger than the exposed structure.
  * This abbreviated structure exposes just enough for the gzgetc() macro.  The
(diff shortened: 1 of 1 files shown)


nmoinvaz (commented):
I think when setting `WITH_GZFILEOP` in cmake and `--without-gzfileops` configure, the build system should add `-DZ_SOLO`, because it is essentially the same thing.

And I wonder if instead of setting it as a preprocessor definition, we should write `Z_SOLO` in zconf.h.

nmoinvaz:
@Dead2 any thoughts on my last comment?

Dead2:
@nmoinvaz The `zconf.h` thing might be a good idea, though I am not 100% sure about any consequences it might have. I am sure a PR would get any cause for discussion going. 😉


Regarding `WITH_GZFILEOP`, I am not sure. Are they really doing the same thing, it did not look like it when I had a very quick look?
I meant to verify more thoroughly, because the two are really not the same when it comes to the actual idea behind them.

`Z_SOLO` is meant to disable functionality that requires functions linked in from an external library.
As far as I understand, in stock zlib this means it disables any features using file access as those functions are external.
We should of course honor this flag in compat mode.

Should we also support the flag in native mode?
How do we verify where we have calls to functions from external libraries? Have we introduced new dependencies?

So we could also get a situation where `Z_SOLO` and `WITH_GZFILEOP` diverge more than today. 🤷 

I accepted these two PRs so we have a starting point that we can build upon, because I suspect this could be a relatively complex issue.

PR #1031, opened 2021-07-22
Fix hangs on macOS due to loading of misaligned addresses in chunkmemset_8.

sergio-nsk:


chunkset_tpl.h
```diff
@@ -176,7 +176,18 @@ Z_INTERNAL uint8_t* CHUNKMEMSET(uint8_t *out, unsigned dist, unsigned len) {
...
+#endif
     len = MIN(len, left);
+    uint8_t *from = out - dist;
+    while (((uintptr_t)out & align_mask) && (len > 0)) {
+        *out++ = *from++;
+        --len;
+        --left;
+    }
     if (left < (unsigned)(3 * sizeof(chunk_t))) {
         uint8_t *from = out - dist;
```
mtl1979: Redeclaration of variable with same name.

chunkset_tpl.h
```diff
@@ -176,7 +176,18 @@ Z_INTERNAL uint8_t* CHUNKMEMSET(uint8_t *out, unsigned dist, unsigned len) {
 }
 
 Z_INTERNAL uint8_t* CHUNKMEMSET_SAFE(uint8_t *out, unsigned dist, unsigned len, unsigned left) {
+#ifdef UNALIGNED64_OK
+    static const uint32_t align_mask = 3;
```
mtl1979: When `UNALIGNED64_OK` is defined, no mask should be needed.

chunkset_tpl.h
```diff
@@ -176,7 +176,18 @@ Z_INTERNAL uint8_t* CHUNKMEMSET(uint8_t *out, unsigned dist, unsigned len) {
...
+#else
+    static const uint32_t align_mask = 7;
+#endif
```
mtl1979: `align_mask` should only be used when:
1. UNALIGNED64_OK is undefined
2. UNALIGNED_OK and UNALIGNED64_OK are both undefined

chunkset_tpl.h
sergio-nsk: Thanks. Forgot to remove it, since I am working with a different repo.

chunkset_tpl.h
sergio-nsk: I was getting odd addresses like 0x7fbc8800946d, thus I thought it would also hang in chunkmemset_4 and chunkmemset_2.

mtl1979: Unaligned addresses are totally normal on 64-bit little-endian machines, most big-endian machines need lower bits to be zero.
Some processors are both little- and big-endian, so in big-endian mode, addresses need to be aligned.
As exception of rule, some 32-bit little-endian processors need `UNALIGNED64_OK` undefined.

chunkset_tpl.h
```diff
@@ -176,9 +176,24 @@ Z_INTERNAL uint8_t* CHUNKMEMSET(uint8_t *out, unsigned dist, unsigned len) {
 }
 
 Z_INTERNAL uint8_t* CHUNKMEMSET_SAFE(uint8_t *out, unsigned dist, unsigned len, unsigned left) {
+#if !defined(UNALIGNED64_OK)
+#if !defined(UNALIGNED_OK)
```
Dead2: Please indent the second level ifs like this:
```
#if foo
#  if bar
    // foo and bar
#  else
    // foo
#  endif
#endif
```
(2 spaces between # and if)

Dead2:
LGTM after the indentation has been fixed.

PR #1032, opened 2021-08-09
IBM Z: Run DFLTCC tests on the self-hosted builder

iii-i:
Hello,

I've installed Github Actions Runner on the `zlib-ng01` z15 VM. Currently it points to [my zlib-ng fork](https://github.com/iii-i/zlib-ng/pull/3), but you should be able to easily reconfigure it to point to zlib-ng itself using the instructions that are part of this PR. With this in place, we'll have better coverage of DFLTCC code in zlib-ng CI.

Best regards,
Ilya

---

* Use the self-hosted builder instead of ubuntu-latest.
* Drop qemu-related settings from DFLTCC configurations.
* Install codecov only for the current user, since the self-hosted
  builder runs under a restricted non-root account.
* Use actions/checkout@v2 for configure checks, since for some reason
  actions/checkout@v1 cannot find git on the self-hosted builder.
* Update the testing section of the DFLTCC README.
* Add the infrastructure code for the self-hosted builder.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -253,23 +253,17 @@ jobs:
             codecov: ubuntu_gcc_s390x
 
           - name: Ubuntu GCC S390X DFLTCC
-            os: ubuntu-latest
-            compiler: s390x-linux-gnu-gcc
-            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-s390x.cmake -DWITH_DFLTCC_DEFLATE=ON -DWITH_DFLTCC_INFLATE=ON -DWITH_SANITIZER=Address
+            os: z15
+            compiler: gcc
+            cmake-args: -DWITH_DFLTCC_DEFLATE=ON -DWITH_DFLTCC_INFLATE=ON -DWITH_SANITIZER=Address
             asan-options: detect_leaks=0
-            packages: qemu qemu-user gcc-s390x-linux-gnu libc-dev-s390x-cross
-            qemu-run: qemu-s390x
-            ldflags: -static
             codecov: ubuntu_gcc_s390x
 
           - name: Ubuntu GCC S390X DFLTCC Compat
-            os: ubuntu-latest
-            compiler: s390x-linux-gnu-gcc
-            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-s390x.cmake -DZLIB_COMPAT=ON -DWITH_DFLTCC_DEFLATE=ON -DWITH_DFLTCC_INFLATE=ON -DWITH_SANITIZER=Undefined
+            os: z15
+            compiler: gcc
+            cmake-args: -DZLIB_COMPAT=ON -DWITH_DFLTCC_DEFLATE=ON -DWITH_DFLTCC_INFLATE=ON -DWITH_SANITIZER=Undefined
             asan-options: detect_leaks=0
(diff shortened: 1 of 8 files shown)


mtl1979:
Fails CI...

iii-i:
@mtl1979 That's expected - sorry, I should've been more explicit about this. The new builder is currently linked to my fork of zlib-ng, not to the main repo. A token is needed in order to link it to the main repo, which only the admin (@Dead2 I guess?) has access to. So he should either share the token with me, so that I could reconfigure the builder, or, preferably, reconfigure the builder himself - not only because I'm lazy, but also to make sure that the instructions I wrote are adequate.

mtl1979:
@iii-i Nobody can review the full pull request until @Dead2 makes the necessary changes... For controversial changes, at least two long-time developers need to review.

iii-i:
@Dead2 ping :-) Could you take a look please?

Dead2:
@iii-i I have not forgotten about this (or other PRs), it is just an extremely hectic period for me currently. Past 6 weeks I have barely been able to do any relaxing at all, and this will probably continue for a couple more weeks due to finishing up on the stuff currently going on plus we are moving offices starting next week. Hopefully things will be back to normal soon, and I can start doing the things I actually want to do again. 😄

In the mean time, anything small and approved by others I will probably merge, but anything bigger or things I think needs some extra scrutiny/testing from me, I am unfortunately unable to make the time for. Very sorry about that.

iii-i:
Rebased, added ephemeral runner support (https://github.blog/changelog/2021-09-20-github-actions-ephemeral-self-hosted-runners-new-webhooks-for-auto-scaling/).

Dead2:
@iii-i Looks like it works now. So, is it ready to merge?

iii-i:
@Dead2 Yes, I think so.

Regarding installing updates on the builder, once QEMU 6.2 is released, I will configure auto-updating https://github.com/multiarch/qemu-user-static there and send a corresponding PR.

Dead2:
@iii-i QEMU 6.2 has been released, and GHA emailed me that we should update our runner yesterday. Any chance of getting that updated?

~I am also contacting RedHat about getting a subscription so we can get system updates on that VM..~ OIC, there are warnings but it is actually bypassed

iii-i:
I'll send a PR with the GHA upgrade (it's strange that the runner does not receive upgrade notifications from the server, but oh well).
With QEMU, we'll have to wait for a new https://github.com/multiarch/qemu-user-static release.

Dead2:
@iii-i I don't think this is the reason it does not self-update for us, but the following was mentioned in the mail from Github.
```
Normally, self-hosted runners will auto-update during startup. When used in conjunction with the --ephemeral flag,
runner versions prior to 2.283.3 had an issue that would cause startup to fail during this auto-update process.
To ensure minimal disruption to your Actions workflows, we temporarily disabled automatic runner upgrades
for ephemeral runners 
```
I did not see any ephemeral flag in our config though, and I assume we would not need to use it anyway.

iii-i:
Ah, I see! [We actually do use it](https://github.com/zlib-ng/zlib-ng/blob/da7622bc8719c2b872a733bcddf1a0f42d8c7475/arch/s390/self-hosted-builder/fs/usr/bin/actions-runner#L37). It's needed because it helps ensuring that we start with a clean slate each time.

Dead2:
Mystery solved 💯 
The mail also says they will re-enable auto-updates in March or something, so we should be good then 👍

PR #1033, opened 2021-08-13
Reorganize inflate window layout

nmoinvaz:
This commit merges in the Intel inflate changes for crc32 pclmulqdq:
https://github.com/jtkukunas/zlib/commit/62b7616b0bad4ff088f1a73d0fa11159b6925fbb

Here is Jim Kukunas original notes from the commit:

> This commit significantly improves inflate performance by reorganizing the window buffer into a contiguous window and pending output buffer. The goal of this layout is to reduce branching, improve cache locality, and enable for the use of crc folding with gzip input.

> The window buffer is allocated as a multiple of the user-selected window size. In this commit, a factor of 2 is utilized.

> The layout of the window buffer is divided into two sections. The first section, window offset [0, wsize), is reserved for history that has already been output. The second section, window offset [wsize, 2 * wsize), is reserved for buffering pending output that hasn't been flushed to the user's output buffer yet.

> The history section grows downwards, towards the window offset of 0. The pending output section grows upwards, towards the end of the buffer. As a result, all of the possible distance/length data that may need to be copied is contiguous. This removes the need to stitch together output from 2 separate buffers.

> In the case of gzip input, crc folding is used to copy the pending output to the user's buffers.

The downsides:

* inflate uses ~~64kb of memory~~ twice as much window memory. If window is 32k, it uses 64k.
* ~~inflateBack no longer uses inflate_fast. With inflateBack you specify the window buffer and size which is different than what this commit does in that it uses a single memory allocation for window buffer and pending output buffer similar to deflate. We could possibly use a separate inflate_fast function or with separate parameters for inflateBack.~~
* For inflateBack we had to create a separate `zng_inflate_fast_back` function to preserve speed improvements when all we are supplied with is a buffer allocation only intended for use as the window.

Todo:

* ~~Make all CI work~~
* ~~Newer benchmarks~~
* ~~See if we can overcome inflateBack issue.~~

Old Benchmarks:

### ZLIB-NG
```
 Tool: minigzip-ng-develop.exe Levels: 1-9
 Runs: 70         Trim worst: 40

 Level   Comp   Comptime min/avg/max/stddev  Decomptime min/avg/max/stddev  Compressed size
 1     44.809%      1.015/1.022/1.030/0.005        0.660/0.670/0.674/0.003      104,067,822
 2     35.391%      1.678/1.700/1.710/0.009        0.673/0.686/0.692/0.005       82,195,416
[...]

PR #1034, opened 2021-08-14
Use ZFREE_STATE macro in inflateBackEnd instead of ZFREE.

nmoinvaz:
Some issues I noticed from my PR #1033.


nmoinvaz:
Nevermind, it only makes sense to fix this in the context of #1033 imo.

PR #1035, opened 2021-08-15
Include win directory in pigz even if not using threads.

nmoinvaz:
This PR always includes `win` directory in pigz when on Windows even if not using pthreads. This `win` directory contains source code the mimic the necessary linux functions. https://github.com/zlib-ng/pigzbench/tree/master/pigz/win
By not compiling with pthreads on Windows we can get to compiling pigz quicker for testing functional issues unrelated to threading.


Dead2 (approved):
LGTM

PR #1037, opened 2021-08-23
Fix UB in inffast.c when not using window

atdt:
When not using window, `window + wsize` applies a zero offset to a null pointer, which is undefined behavior.

### DIFF
--- a/inffast.c
+++ b/inffast.c
@@ -155,7 +155,7 @@ void Z_INTERNAL zng_inflate_fast(PREFIX3(stream) *strm, unsigned long start) {
     /* Detect if out and window point to the same memory allocation. In this instance it is 
        necessary to use safe chunk copy functions to prevent overwriting the window. If the 
        window is overwritten then future matches with far distances will fail to copy correctly. */
-    extra_safe = (out >= window && out + INFLATE_FAST_MIN_LEFT <= window + wsize);
+    extra_safe = (wsize != 0 && out >= window && out + INFLATE_FAST_MIN_LEFT <= window + wsize);
 
     /* decode literals and length/distances until end-of-block or not enough
        input data or output space */


mtl1979:
I don't see anything undefined in using null pointer in pointer arithmetics... As long as the variable size is well-defined (non-void pointer), the generated assembly will be correct, basically it either assembles to load + addition, or load effective address (LEA).

The real issue comes if the pointer is later dereferenced.

atdt:
IANALL (I am not a language lawyer, but..) this was discussed in https://reviews.llvm.org/D67122 and the consensus was that any arithmetic on null pointers is undefined in C, even if you add zero. UBSan flags it.

mtl1979:
@atdt Pointer arithmetics in C are a lot easier than with C++... As zlib and zlib-ng are both C, the code has no UB. Compiler doesn't know at compile-time that the pointer is NULL constant or has 0 offset, so it can't optimize away offset.

atdt:
@mtl1979 According to the discussion I linked to above, this is _only_ UB in C, and not in C++, where it would be well-defined.

I think the change improves readability, and it helps us adopt zlib-ng because otherwise we have to carry a patch that adds a `__attribute__((no_sanitize("undefined")))` annotation to this code.

mtl1979:
@atdt Well... I tested with two of the most common compilers and both return same predictable answer... I didn't try with Visual C++ as it is C++ compiler and not C compiler.

turol:
@mtl1979 That means nothing. UB is dangerous because some future compiler will likely start using it for optimization and your "working" code suddenly crashes or misbehaves.

mtl1979:
@turol Future compilers can't optimize the code differently unless they know for sure that both variables are zero. There is no chance for that to happen as it would already trigger a compiler warning when all warnings are enabled.

atdt:
@mtl1979 I'm confused. Are you opposed to this change?

Upstream LLVM very definitely thinks this is UB. There is [an explicit check for this in UBSan](https://github.com/llvm/llvm-project/blob/d480f968ad8b56d3ee4a6b6df5532d485b0ad01e/compiler-rt/lib/ubsan/ubsan_handlers.cpp#L786) which flags this code in zlib-ng. Why quibble?

mtl1979:
@atdt I'm saying it's false positive as there is no constant NULL or nullptr.

atdt:
It's undefined behavior because the C language specification does not define how the compiler ought to behave when you apply a zero offset to a null pointer. What current compilers can or cannot do with this code is not a very good argument.

mtl1979:
@atdt Offsets have nothing to do with C, so it's expected that C standard doesn't define it... C standard only covers arrays which have defined element size. C standard also says that undefined element size is assumed to equal to 1 byte, which would also apply to explicit NULL pointers, but not other pointers that point to address 0.

PR #1042, opened 2021-09-06
Make integration into bigger projects easier

dzenanz:


CMakeLists.txt
```diff
@@ -950,7 +950,9 @@ endif()
 
 foreach(ZLIB_INSTALL_LIBRARY ${ZLIB_INSTALL_LIBRARIES})
     target_include_directories(${ZLIB_INSTALL_LIBRARY} PUBLIC
-        ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
+        "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR};${CMAKE_CURRENT_SOURCE_DIR}>"
+        "$<INSTALL_INTERFACE:${CMAKE_INSTALL_PREFIX}/include>"
```
mtl1979: Using `${CMAKE_INSTALL_PREFIX}` is bad. See https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-relocatable-packages

dzenanz: Updated.

CMakeLists.txt
```diff
@@ -950,7 +950,9 @@ endif()
 
 foreach(ZLIB_INSTALL_LIBRARY ${ZLIB_INSTALL_LIBRARIES})
     target_include_directories(${ZLIB_INSTALL_LIBRARY} PUBLIC
-        ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_CURRENT_SOURCE_DIR})
+        "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR};${CMAKE_CURRENT_SOURCE_DIR}>"
+        "$<INSTALL_INTERFACE:include>"
+        )
```
nmoinvaz: I would prefer if closing brace wasn't on a separate line. That formatting is good if there is like a list of files and that list may change a lot but I don't believe this instance is that.

dzenanz: Done.

hjmjohnson (approved):
Modern CMake makes integration with other packages much easier.

dzenanz:
Why are majority of CI tests "cancelled"?

mtl1979:
@dzenanz Could be some issue with the servers hosting the CI runners... Usually tests only get cancelled if there is too many failures or another commit supersedes the current commit.

dzenanz:
Can tests be restarted? And this small patch hopefully merged?

dzenanz:
@Dead2 can this be merged too? It passes the tests, and is pretty small.

dzenanz:
Without this patch, generating ITK's solution fails:
```text
Configuring done
CMake Error in Modules/ThirdParty/ZLIB/src/itkzlib-ng/CMakeLists.txt:
  Target "zlib" INTERFACE_INCLUDE_DIRECTORIES property contains path:

    "C:/Dev/ITK-2019/Modules/ThirdParty/ZLIB/src/itkzlib-ng"

  which is prefixed in the build directory.


CMake Error in Modules/ThirdParty/ZLIB/src/itkzlib-ng/CMakeLists.txt:
  Target "zlib" INTERFACE_INCLUDE_DIRECTORIES property contains path:

    "C:/Dev/ITK-git/Modules/ThirdParty/ZLIB/src/itkzlib-ng"

  which is prefixed in the source directory.


Generating done
```
Until this is integrated, we will handle it applying this as a [custom patch](https://github.com/InsightSoftwareConsortium/ITK/pull/2803/commits/02993d5ba5d29930beab85ad70184ddba30cb33f).

Dead2:
Looks pretty good, just one question: Does this affect CMake minimum versions in any way?

dzenanz:
It looks like [generator expressions are present since CMake 3.0](https://cmake.org/cmake/help/v3.0/manual/cmake-generator-expressions.7.html). So a version bump is not necessary.

Dead2:
@dzenanz Excellent 👍

Dead2 (approved):

dzenanz:
Thanks!

PR #1048, opened 2021-10-10
COMP: Fix data loss warning

jhlegarreta:
Fix data loss warning.

Fixes:
```
itkzlib-ng/inflate.c(1209,24): warning C4267: '=': conversion from 'size_t' to 'unsigned long', possible loss of data
itkzlib-ng/inflate.c(1210,26): warning C4267: '=': conversion from 'size_t' to 'unsigned long', possible loss of data
```

### DIFF
--- a/inflate.c
+++ b/inflate.c
@@ -1206,8 +1206,8 @@ int32_t Z_EXPORT PREFIX(inflateSync)(PREFIX3(stream) *strm) {
     in = strm->total_in;
     out = strm->total_out;
     PREFIX(inflateReset)(strm);
-    strm->total_in = in;
-    strm->total_out = out;
+    strm->total_in = (z_size_t)in;
+    strm->total_out = (z_size_t)out;
     state->flags = flags;
     state->mode = TYPE;
     return Z_OK;


jhlegarreta:
Local system: Windows 10
Compiler: MSVC 2019 version 19.29.30133.0

Found when building ITK: https://github.com/InsightSoftwareConsortium/ITK

mtl1979:
Using `unsigned long` instead of `size_t` can cause unwanted truncation as we declare `total_in` and `total_out` as `size_t` in non-compat header. `unsigned long` is only 32 bits under 64-bit Windows, but `size_t` can be either 32 or 64 bits depending on compiler version.

We might want to use explicit cast to `z_size_t` instead on those two lines.

jhlegarreta:
Thanks for the review @mtl1979 .

> We might want to use explicit cast to z_size_t instead on those two lines.

Let me know if you want me to amend the commit that way.

mtl1979:
@jhlegarreta I think using cast is better than changing the type of the temporary variables as that avoids all issues when building the non-compat version, which should be the default version for most developers compiling from sources... "Compat" version still truncates, but that is something we can't fix to stay compatible with stock zlib.

jhlegarreta:
> @jhlegarreta I think using cast is better than changing the type of the temporary variables as that avoids all issues when building the non-compat version, which should be the default version for most developers compiling from sources... "Compat" version still truncates, but that is something we can't fix to stay compatible with stock zlib.

Changed in ad6171d.

mtl1979:
Cast to `unsigned long` is still wrong, it will truncate on non-compat version.

jhlegarreta:
> Cast to unsigned long is still wrong, it will truncate on non-compat version.

My bad; 977a2e5 casts to `z_size_t`. Sorry.

Dead2 (approved):

jhlegarreta:
Build failures seem unrelated:
```
Error: Could not find Xcode version that satisfied version spec: '12.1.1'
```

https://github.com/zlib-ng/zlib-ng/actions/runs/1326434359

e.g.
https://github.com/zlib-ng/zlib-ng/pull/1048/checks?check_run_id=3870178563#step:5:16

Dead2:
@jhlegarreta Yes, the 3 remaining Ubuntu 16.04 checks are also going to fail after maaany hours of waiting. That seems to be a bug in Github Actions regarding that specific distro version, not sure how that works but it has been like that for weeks now.

I have not seen the xcode errors before, but they probably just mean we need to update our CI scripts.

So I am ignoring those checks 😉

jhlegarreta:
> @jhlegarreta Yes, the 3 remaining Ubuntu 16.04 checks are also going to fail after maaany hours of waiting. That seems to be a bug in Github Actions regarding that specific distro version, not sure how that works but it has been like that for weeks now.

@Dead2 Does this have to do with the above?
https://github.blog/changelog/2021-04-29-github-actions-ubuntu-16-04-lts-virtual-environment-will-be-removed-on-september-20-2021/

Should those builds transition to 18.04 as the minimum version?

Dead2:
@jhlegarreta Yep, that is the reason. IMO, Github Actions should have failed quickly instead of spending lots of hours in queue before failing though.

PR #1049, opened 2021-10-11
IBM Z: Adjust compressBound() for DFLTCC (and several other fixes)

iii-i:
Hi,

a bug with `compressBound()` + DFLTCC was discovered in zlib. While porting the fix to zlib-ng, I also stumbled upon several other problems. This PR fixes them all.

Best regards,
Ilya


arch/s390/dfltcc_common.h
```diff
@@ -26,4 +26,8 @@ void Z_INTERNAL dfltcc_free_window(PREFIX3(streamp) strm, void *w);
 
 #define TRY_FREE_WINDOW dfltcc_free_window
 
+#define DEFLATE_BOUND_COMPLEN(source_len) \
+    ((3 + 5 + 5 + 4 + 19 * 3 + (286 + 30) * 7 + \
+      (source_len) * 16 + 15 + 7) >> 3)
```
Dead2: This bunch of magic numbers is begging for a comment detailing what the numbers mean and/or how to verify that these numbers are the correct ones after future changes.
(I know this is not a regression btw)

compress.c
```diff
@@ -3,14 +3,25 @@
...
 #endif
 
+/* ===========================================================================
+ *  Architecture-specific hooks.
+ */
+#ifdef S390_DFLTCC_DEFLATE
+#  include "arch/s390/dfltcc_common.h"
+#else
+/* Returns the upper bound on compressed data length based on uncompressed data length, assuming default settings.
+ * Zero means that arch-specific deflation code behaves identically to the regular zlib-ng algorithms. */
```
Dead2, line 21: 👍

compress.c
```diff
@@ -73,6 +84,11 @@ int Z_EXPORT PREFIX(compress)(unsigned char *dest, z_size_t *destLen, const unsi
    this function needs to be updated.
  */
 z_size_t Z_EXPORT PREFIX(compressBound)(z_size_t sourceLen) {
+    z_size_t complen = DEFLATE_BOUND_COMPLEN(sourceLen);
+
+    if (complen > 0)
+        return complen + 6;
```
Dead2: 6?

Perhaps remove the blank line and add a comment before these three lines regarding what and why this short-circuit?

test/example.c
```diff
@@ -514,7 +514,7 @@ void test_dict_deflate(unsigned char *compr, size_t comprLen) {
  */
 void test_dict_inflate(unsigned char *compr, size_t comprLen, unsigned char *uncompr, size_t uncomprLen) {
     int err;
-    uint8_t check_dictionary[10];
+    uint8_t check_dictionary[32768];
```
Dead2: Commit message says nothing about why this change is happening, care to elaborate?

iii-i: This is the maximum size that the `inflateGetDictionary()` can return according to the manual. I will add a `#define` for it.

compress.c
iii-i: This is `wraplen` for zlib. I thought about adding a `#define` here, but it's hardcoded in quite a few places already (I'll do it now anyway).

arch/s390/dfltcc_common.h
iii-i: I got this formula from our hardware team, I'll ask them what this all means. Most likely this is just the size of the worst deflate stream possible (2 bytes per symbol, largest DHT, etc).

test/example.c
Dead2: I was just surprised to see this one. 👍

compress.c
```diff
@@ -73,6 +84,11 @@ int Z_EXPORT PREFIX(compress)(unsigned char *dest, z_size_t *destLen, const unsi
    this function needs to be updated.
  */
 z_size_t Z_EXPORT PREFIX(compressBound)(z_size_t sourceLen) {
+    z_size_t complen = DEFLATE_BOUND_COMPLEN(sourceLen);
+
+    if (complen > 0)
+        /* Architecture-specific code provided an upper bound. */
+        return complen + ZLIB_WRAPLEN;
```
Dead2, line 91: I guess I was a bit unclear in my last feedback 😄 
I meant remove the blank line currently on line 88, and add comment between line 86/87.

But this works too, just add back the blank line 91/92. 👍

Dead2 (approved):

PR #1052, opened 2021-10-15
IBM Z: Run DFLTCC tests on the self-hosted builder

iii-i:
* Use the self-hosted builder instead of ubuntu-latest.
* Drop qemu-related settings from DFLTCC configurations.
* Install codecov only for the current user, since the self-hosted
  builder runs under a restricted non-root account.
* Use actions/checkout@v2 for configure checks, since for some reason
  actions/checkout@v1 cannot find git on the self-hosted builder.
* Update the testing section of the DFLTCC README.
* Add the infrastructure code for the self-hosted builder.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -253,23 +253,17 @@ jobs:
             codecov: ubuntu_gcc_s390x
 
           - name: Ubuntu GCC S390X DFLTCC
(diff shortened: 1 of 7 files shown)


iii-i:
Sorry for disturbance, this was supposed to be a test PR for my own fork :-(

PR #934, opened 2021-04-24
use memcpy/memcmp for unaligned reads

matijaskala:
that way any future optimizers won't make false
assumptions that those reads are actually aligned

compiler optimizes out memcpy/memcmp calls so
generated assembly code remains identical

PR #1053, opened 2021-10-23
AVX2 optimizations

KungFuJesus:
Since this is constant, anyway, we may as well use the variant that
doesn't add vector register pressure, has better ILP opportunities,
and has shorter instruction latency.

### DIFF
--- a/arch/x86/adler32_avx.c
+++ b/arch/x86/adler32_avx.c
@@ -50,9 +50,6 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
         {1, 1, 1, 1, 1, 1, 1, 1,  1, 1, 1, 1, 1, 1, 1, 1};
     __m256i dot3v = _mm256_load_si256((__m256i*)dot3);
 
-    // We will need to multiply by
-    char ALIGNED_(32) shift[16] = {5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
-    __m128i shiftv = _mm_load_si128((__m128i*)shift);
 
     while (len >= 32) {
        __m256i vs1 = _mm256_load_si256((__m256i*)s1);
@@ -77,7 +74,7 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
            __m256i v_short_sum2 = _mm256_maddubs_epi16(vbuf, dot2v);
            vs1 = _mm256_add_epi32(vsum1, vs1);
            __m256i vsum2 = _mm256_madd_epi16(v_short_sum2, dot3v);
-           vs1_0 = _mm256_sll_epi32(vs1_0, shiftv);
+           vs1_0 = _mm256_slli_epi32(vs1_0, 5);
            vsum2 = _mm256_add_epi32(vsum2, vs2);
            vs2   = _mm256_add_epi32(vsum2, vs1_0);
            vs1_0 = vs1;
--- a/arch/x86/adler32_avx.c
+++ b/arch/x86/adler32_avx.c
@@ -9,20 +9,24 @@
 #include "../../zutil.h"
 
 #include "../../adler32_p.h"
-#include <stdio.h>
 
 #include <immintrin.h>
 
 #ifdef X86_AVX2_ADLER32
 
-/* 64 bit horizontal sum, adapted from Agner Fog's
- * vector library. */
-static inline uint64_t hsum(__m256i x)
-{
+/* 64 bit horizontal sum, adapted from Agner Fog's vector library. */
+static inline uint64_t hsum(__m256i x) {
     __m256i sum1 = _mm256_shuffle_epi32(x, 0x0E);
     __m256i sum2 = _mm256_add_epi64(x, sum1);
     __m128i sum3 = _mm256_extracti128_si256(sum2, 1);
+#if defined(__x86_64__) || defined(_M_X64)
     return _mm_cvtsi128_si64(_mm_add_epi64(_mm256_castsi256_si128(sum2), sum3));
+#else
+    __m128i ret_vec = _mm_add_epi64(_mm256_castsi256_si128(sum2), sum3);
+    uint64_t ret_val;
+    _mm_storel_epi64((__m128i*)&ret_val, ret_vec);
+    return ret_val;
+#endif
 }
 
 Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_t len) {
@@ -44,17 +48,15 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
     if (UNLIKELY(len < 16))
         return adler32_len_16(adler, buf, len, sum2);
 
-    /* If we could shift over 128 bit lanes, a broadcast + shift would be better */
-    const __m256i sMask = _mm256_setr_epi32(0, 0, 0, 0, 0, 0, 0, -1);
+    const __m256i vs_mask = _mm256_setr_epi32(0, 0, 0, 0, 0, 0, 0, -1);
     __m256i vs1 = _mm256_set1_epi32(adler);
     __m256i vs2 = _mm256_set1_epi32(sum2);
-    vs1 = _mm256_and_si256(vs1, sMask);
-    vs2 = _mm256_and_si256(vs2, sMask);
+    vs1 = _mm256_and_si256(vs1, vs_mask);
+    vs2 = _mm256_and_si256(vs2, vs_mask);
 
     const __m256i dot1v = _mm256_set1_epi8(1);
-    const __m256i dot2v = _mm256_setr_epi8(32, 31, 30, 29, 28, 27, 26, 25, 24, 23, 22,
-                                           21, 20, 19, 18, 17, 16, 15, 14, 13, 12, 11,
-                                           10, 9, 8, 7, 6, 5, 4, 3, 2, 1);
+    const __m256i dot2v = _mm256_setr_epi8(32, 31, 30, 29, 28, 27, 26, 25, 24, 23, 22, 21, 20, 19, 18, 17, 16, 15,
+                                           14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1);
     const __m256i dot3v = _mm256_set1_epi16(1);
 
     while (len >= 32) {
@@ -67,7 +69,7 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
        while (k >= 32) {
            /*
               vs1 = adler + sum(c[i])
-              vs2 = sum2 + 16 vs1 + sum( (16-i+1) c[i] )
+              vs2 = sum2 + 32 vs1 + sum( (32-i+1) c[i] )
            */
            __m256i vbuf = _mm256_loadu_si256((__m256i*)buf);
            buf += 32;
@@ -86,66 +88,62 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
 
        /* The compiler is generating the following sequence for this integer modulus
         * when done the scalar way, in GPRs:
-            mov    $0x80078071,%edi // move magic constant into 32 bit register %edi
-            ...
-            vmovd  %xmm1,%esi // move vector lane 0 to 32 bit register %esi
-            mov    %rsi,%rax  // zero-extend this value to 64 bit precision in %rax
-            imul   %rdi,%rsi // do a signed multiplication with magic constant and vector element 
-            shr    $0x2f,%rsi // shift right by 47
-            imul   $0xfff1,%esi,%esi // do a signed multiplication with value truncated to 32 bits with 0xfff1 
-            sub    %esi,%eax // subtract lower 32 bits of original vector value from modified one above
-            ...
-            // repeats for each element with vpextract instructions
+        
+        adler = (s1_unpack[0] % BASE) + (s1_unpack[1] % BASE) + (s1_unpack[2] % BASE) + (s1_unpack[3] % BASE) +
+                (s1_unpack[4] % BASE) + (s1_unpack[5] % BASE) + (s1_unpack[6] % BASE) + (s1_unpack[7] % BASE);
 
-            This is tricky with AVX2 for a number of reasons:
-                1.) There's no 64 bit multiplication instruction, but there is a sequence to get there
-                2.) There's ways to extend vectors to 64 bit precision, but no simple way to truncate
-                    back down to 32 bit precision later (there is in AVX512) 
-                3.) Full width integer multiplications aren't cheap
-            */
+        mov    $0x80078071,%edi // move magic constant into 32 bit register %edi
+        ...
+        vmovd  %xmm1,%esi // move vector lane 0 to 32 bit register %esi
+        mov    %rsi,%rax  // zero-extend this value to 64 bit precision in %rax
+        imul   %rdi,%rsi // do a signed multiplication with magic constant and vector element 
+        shr    $0x2f,%rsi // shift right by 47
+        imul   $0xfff1,%esi,%esi // do a signed multiplication with value truncated to 32 bits with 0xfff1 
+        sub    %esi,%eax // subtract lower 32 bits of original vector value from modified one above
+        ...
+        // repeats for each element with vpextract instructions
 
-            // At this point, we have partial sums stored in vs1 and vs2.  There are AVX512 instructions that
-            // would allow us to sum these quickly (VP4DPWSSD).  For now, just unpack and move on.
-            /*uint32_t ALIGNED_(32) s1_unpack[8];
-            uint32_t ALIGNED_(32) s2_unpack[8];
+        This is tricky with AVX2 for a number of reasons:
+            1.) There's no 64 bit multiplication instruction, but there is a sequence to get there
+            2.) There's ways to extend vectors to 64 bit precision, but no simple way to truncate
+                back down to 32 bit precision later (there is in AVX512) 
+            3.) Full width integer multiplications aren't cheap
 
-            _mm256_store_si256((__m256i*)s1_unpack, vs1);
-            _mm256_store_si256((__m256i*)s2_unpack, vs2);
-            
+        We can, however, cast up to 64 bit precision on all 8 integers at once, and do a relatively
+        cheap sequence for horizontal sums. Then, we simply do the integer modulus on the resulting
+        64 bit GPR, on a scalar value
+        */
+
  
-            adler = (s1_unpack[0] % BASE) + (s1_unpack[1] % BASE) + (s1_unpack[2] % BASE) + (s1_unpack[3] % BASE) +
-                    (s1_unpack[4] % BASE) + (s1_unpack[5] % BASE) + (s1_unpack[6] % BASE) + (s1_unpack[7] % BASE);*/
+        /* Will translate to nops */
+        __m128i s1lo = _mm256_castsi256_si128(vs1);
+        __m128i s2lo = _mm256_castsi256_si128(vs2);
 
-            /* Will translate to nops */
-            __m128i s1lo = _mm256_castsi256_si128(vs1);
-            __m128i s2lo = _mm256_castsi256_si128(vs2);
+        /* Requires vextracti128 */
+        __m128i s1hi = _mm256_extracti128_si256(vs1, 1);
+        __m128i s2hi = _mm256_extracti128_si256(vs2, 1);
+        
+        /* Convert up to 64 bit precision to prevent overflow */
+        __m256i s1lo256 = _mm256_cvtepi32_epi64(s1lo);
+        __m256i s1hi256 = _mm256_cvtepi32_epi64(s1hi);
+        __m256i s2lo256 = _mm256_cvtepi32_epi64(s2lo);
+        __m256i s2hi256 = _mm256_cvtepi32_epi64(s2hi);
 
-            __m128i s1hi = _mm256_extracti128_si256(vs1, 1);
-            __m128i s2hi = _mm256_extracti128_si256(vs2, 1);
-            
-            /* Convert up to 64 bit precision to prevent overflow */
-            __m256i s1lo256 = _mm256_cvtepi32_epi64(s1lo);
-            __m256i s1hi256 = _mm256_cvtepi32_epi64(s1hi);
-            __m256i s2lo256 = _mm256_cvtepi32_epi64(s2lo);
-            __m256i s2hi256 = _mm256_cvtepi32_epi64(s2hi);
+        /* Sum vectors in existing lanes */
+        __m256i s1_sum = _mm256_add_epi64(s1lo256, s1hi256);
+        __m256i s2_sum = _mm256_add_epi64(s2lo256, s2hi256);
+        
+        /* In AVX2-land, this trip through GPRs will probably be unvoidable, as there's no cheap and easy
+         * conversion from 64 bit integer to 32 bit. This casting to 32 bit is cheap through GPRs 
+         * (just register aliasing), and safe, as our base is significantly smaller than UINT32_MAX */
+        adler = (uint32_t)(hsum(s1_sum) % BASE);
+        sum2 = (uint32_t)(hsum(s2_sum) % BASE);
 
-            /* sum vectors in existing lanes */
-            __m256i s1Sum = _mm256_add_epi64(s1lo256, s1hi256);
-            __m256i s2Sum = _mm256_add_epi64(s2lo256, s2hi256);
-            
-            /* In AVX2-land, this trip through GPRs will probably
-             * be unvoidable, as there's no cheap and easy conversion
-             * from 64 bit integer to 32 bit. This casting to 32 bit
-             * is cheap through GPRs (just register aliasing), and safe,
-             * as our base is significantly smaller than UINT32_MAX */
-            adler = (uint32_t)(hsum(s1Sum) % BASE);
-            sum2 = (uint32_t)(hsum(s2Sum) % BASE);
+        vs1 = _mm256_set1_epi32(adler);
+        vs2 = _mm256_set1_epi32(sum2);
 
-            vs1 = _mm256_set1_epi32(adler);
-            vs1 = _mm256_and_si256(vs1, sMask);
- 
-            vs2 = _mm256_set1_epi32(sum2);
-            vs2 = _mm256_and_si256(vs2, sMask);
+        vs1 = _mm256_and_si256(vs1, vs_mask);
+        vs2 = _mm256_and_si256(vs2, vs_mask);
     }
 
     /* Process tail (len < 16).  */
--- a/arch/x86/adler32_avx.c
+++ b/arch/x86/adler32_avx.c
@@ -33,27 +33,22 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
     if (UNLIKELY(len < 16))
         return adler32_len_16(adler, buf, len, sum2);
 
-    uint32_t ALIGNED_(32) s1[8], s2[8];
+    __m256i vsMask = _mm256_setr_epi32(0, 0, 0, 0, 0, 0, 0, -1);
+    __m256i vs1 = _mm256_set1_epi32(adler);
+    __m256i vs2 = _mm256_set1_epi32(sum2);
+    vs1 = _mm256_and_si256(vs1, vsMask);
+    vs2 = _mm256_and_si256(vs2, vsMask);
 
-    memset(s1, 0, sizeof(s1)); s1[7] = adler; // TODO: would a masked load be faster?
-    memset(s2, 0, sizeof(s2)); s2[7] = sum2;
-
-    char ALIGNED_(32) dot1[32] = \
-        {1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
-         1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1};
-    __m256i dot1v = _mm256_load_si256((__m256i*)dot1);
-    char ALIGNED_(32) dot2[32] = \
-        {32, 31, 30, 29, 28, 27, 26, 25, 24, 23, 22, 21, 20, 19, 18, 17,
-         16, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1};
-    __m256i dot2v = _mm256_load_si256((__m256i*)dot2);
-    short ALIGNED_(32) dot3[16] = \
-        {1, 1, 1, 1, 1, 1, 1, 1,  1, 1, 1, 1, 1, 1, 1, 1};
-    __m256i dot3v = _mm256_load_si256((__m256i*)dot3);
+    const __m256i dot1v = _mm256_set1_epi8(1);
+    const __m256i dot2v = _mm256_setr_epi8(32, 31, 30, 29, 28, 27,
+                                           26, 25, 24, 23, 22, 21,
+                                           20, 19, 18, 17, 16, 15,
+                                           14, 13, 12, 11, 10, 9,
+                                           8, 7, 6, 5, 4, 3, 2, 1);
+    __m256i dot3v = _mm256_set1_epi16(1);
 
 
     while (len >= 32) {
-       __m256i vs1 = _mm256_load_si256((__m256i*)s1);
-       __m256i vs2 = _mm256_load_si256((__m256i*)s2);
        __m256i vs1_0 = vs1;
 
        int k = (len < NMAX ? (int)len : NMAX);
@@ -88,15 +83,20 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
        _mm256_store_si256((__m256i*)s1_unpack, vs1);
        _mm256_store_si256((__m256i*)s2_unpack, vs2);
 
-       adler = (s1_unpack[0] % BASE) + (s1_unpack[1] % BASE) + (s1_unpack[2] % BASE) + (s1_unpack[3] % BASE) +
-               (s1_unpack[4] % BASE) + (s1_unpack[5] % BASE) + (s1_unpack[6] % BASE) + (s1_unpack[7] % BASE);
-       adler %= BASE;
-       s1[7] = adler;
+       uint64_t adler64 = ((uint64_t)s1_unpack[0] + (uint64_t)s1_unpack[1] + (uint64_t)s1_unpack[2] +
+                           (uint64_t)s1_unpack[3] + (uint64_t)s1_unpack[4] + (uint64_t)s1_unpack[5] +
+                           (uint64_t)s1_unpack[6] + (uint64_t)s1_unpack[7]) % BASE;
 
-       sum2 = (s2_unpack[0] % BASE) + (s2_unpack[1] % BASE) + (s2_unpack[2] % BASE) + (s2_unpack[3] % BASE) +
-              (s2_unpack[4] % BASE) + (s2_unpack[5] % BASE) + (s2_unpack[6] % BASE) + (s2_unpack[7] % BASE);
-       sum2 %= BASE;
-       s2[7] = sum2;
+       uint64_t sum264  = ((uint64_t)s2_unpack[0] + (uint64_t)s2_unpack[1] + (uint64_t)s2_unpack[2] +
+                           (uint64_t)s2_unpack[3] + (uint64_t)s2_unpack[4] + (uint64_t)s2_unpack[5] +
+                           (uint64_t)s2_unpack[6] + (uint64_t)s2_unpack[7]) % BASE;
+
+       adler = (uint32_t)adler64;
+       sum2 = (uint32_t)sum264;
+       vs1 = _mm256_set1_epi32(adler);
+       vs2 = _mm256_set1_epi32(sum2);
+       vs1 = _mm256_and_si256(vs1, vsMask);
+       vs2 = _mm256_and_si256(vs2, vsMask);
     }
 
     /* Process tail (len < 16).  */
--- a/arch/x86/adler32_avx.c
+++ b/arch/x86/adler32_avx.c
@@ -9,10 +9,21 @@
 #include "../../zutil.h"
 
 #include "../../adler32_p.h"
+#include <stdio.h>
 
 #include <immintrin.h>
 
 #ifdef X86_AVX2_ADLER32
+
+/* 64 bit horizontal sum, adapted from Agner Fog's
+ * vector library. */
+static inline uint64_t hsum(__m256i x)
+{
+    __m256i sum1 = _mm256_shuffle_epi32(x, 0x0E);
+    __m256i sum2 = _mm256_add_epi64(x, sum1);
+    __m128i sum3 = _mm256_extracti128_si256(sum2, 1);
+    return _mm_cvtsi128_si64(_mm_add_epi64(_mm256_castsi256_si128(sum2), sum3));
+}
 
 Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_t len) {
     uint32_t sum2;
@@ -33,20 +44,18 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
     if (UNLIKELY(len < 16))
         return adler32_len_16(adler, buf, len, sum2);
 
-    __m256i vsMask = _mm256_setr_epi32(0, 0, 0, 0, 0, 0, 0, -1);
+    /* If we could shift over 128 bit lanes, a broadcast + shift would be better */
+    const __m256i sMask = _mm256_setr_epi32(0, 0, 0, 0, 0, 0, 0, -1);
     __m256i vs1 = _mm256_set1_epi32(adler);
     __m256i vs2 = _mm256_set1_epi32(sum2);
-    vs1 = _mm256_and_si256(vs1, vsMask);
-    vs2 = _mm256_and_si256(vs2, vsMask);
+    vs1 = _mm256_and_si256(vs1, sMask);
+    vs2 = _mm256_and_si256(vs2, sMask);
 
     const __m256i dot1v = _mm256_set1_epi8(1);
-    const __m256i dot2v = _mm256_setr_epi8(32, 31, 30, 29, 28, 27,
-                                           26, 25, 24, 23, 22, 21,
-                                           20, 19, 18, 17, 16, 15,
-                                           14, 13, 12, 11, 10, 9,
-                                           8, 7, 6, 5, 4, 3, 2, 1);
-    __m256i dot3v = _mm256_set1_epi16(1);
-
+    const __m256i dot2v = _mm256_setr_epi8(32, 31, 30, 29, 28, 27, 26, 25, 24, 23, 22,
+                                           21, 20, 19, 18, 17, 16, 15, 14, 13, 12, 11,
+                                           10, 9, 8, 7, 6, 5, 4, 3, 2, 1);
+    const __m256i dot3v = _mm256_set1_epi16(1);
 
     while (len >= 32) {
        __m256i vs1_0 = vs1;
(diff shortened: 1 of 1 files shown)


arch/x86/adler32_avx.c
```diff
@@ -77,7 +77,7 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
            __m256i v_short_sum2 = _mm256_maddubs_epi16(vbuf, dot2v);
            vs1 = _mm256_add_epi32(vsum1, vs1);
            __m256i vsum2 = _mm256_madd_epi16(v_short_sum2, dot3v);
-           vs1_0 = _mm256_sll_epi32(vs1_0, shiftv);
+           vs1_0 = _mm256_slli_epi32(vs1_0, AVX2_SHIFTI);
```
mtl1979: Because `AVX2_SHIFTI` is ever used once, we might as well use the constant directly...

mtl1979 (changes requested):

arch/x86/adler32_avx.c
KungFuJesus: Good point.  I wasn't sure if people were going to be picky about using magic numbers or something.  For some compilers (including some but not all versions of GCC), they can get picky about the immediate being defined in code, even as a constant.  In C++ constexpr would obviously work, but if we're ok with hardcoded the immediate, LGTM.

mtl1979: The original code didn't explain it either, so there is no regression in readability... Using shifts for multiply or multiply for widening vector elements (in absence of single-vector horizontal add) is so common practice that it doesn't even warrant commenting in code that is not meant for beginner developers... For more complex intrinsic chains we prefer inline comments to make it clear what the code does.

A lot of the more recent intrinsics code for x86_64 (also known as AMD64) and some other architectures was originally written by me several years ago before I even finished reading the relevant intrinsics guides, so there might be some suboptimal intrinsics and as such it's primarily my territory to review, and accept or reject, any relevant changes. 

Currently @Dead2 is the only one who can trigger CI tests for new users and merge changes.

KungFuJesus: My apologies for all those forced pushes (though they shouldn't affect upstream).  My git workflow for clean PRs is a bit rusty.

KungFuJesus:
So, at least on newer GCCs, the compiler is recognizing the constant shift and using the immediate shift instruction.  So this makes basically zero difference on the compiled binary (but could on a crappier or older compiler).  I do have some micro optimizations that reduces a trip to the stack and back (s1 & s2) as well.  There's an obvious horizontal add optimization that _can_ be done if not for the goofy base that Mark Adler picks that we have to do integer modulus against.  Because NMAX is the maximum runs we can do before possibly overflowing the base, we have to do the modulo operation on every vector element.  This requires integer division (a no go for SIMD, obviously), or full 64 bit products (absent from AVX2), shifts, and subtractions, as the compiler is actually doing on the ALU.

Hopefully, I should be able to have something that does just this with AVX512, and it should be faster for a larger stream of bytes.

mtl1979:
@KungFuJesus AVX512 indeed has better support for single-vector horizontal add, which has been missing since AMD dropped XOP instructions... I haven't added support for it yet as most PCs currently in use still don't support it... Also, I'm not sure using 512-bit vectors will give much gain as adler32 hardly ever needs to checksum that long byte sequences.

KungFuJesus:
> @KungFuJesus AVX512 indeed has better support for single-vector horizontal add, which has been missing since AMD dropped XOP instructions... I haven't added support for it yet as most PCs currently in use still don't support it... Also, I'm not sure using 512-bit vectors will give much gain as adler32 hardly ever needs to checksum that long byte sequences.

I just purchased a Cascade Lake X CPU, so I should be able to have something that does this shortly.  Adler32 checksums tend to do things like checksum the entire zlib compressed row of bytes in a PNG.  I guess realistically for this to be _really_ a danger you'd need something larger than could possibly be rendered. However, as a generic checksum, there are cases where this could happen.  A thought _did_ occur to me, though.

Instead of rebasing to the prime number base for the adler checksum to each vector element, you really only need to rebase in such a way that the partial sums, when horizontally summed together, don't overflow 32 bit precision.  To do this, you could simply just do a modulo with 2^32/8, which is a power of 2.  Simply AND'ing with 2^29-1 and only doing the prime number modulus when it's a scalar in a GPR is probably feasible, do you agree?

KungFuJesus:
Ahh scratch that above idea, it was dumb. In order for that to work the base would need to be a multiple of that prime or we're losing the biproduct of the modular arithmetic.

Still, full 64 bit multiplication with AVX512 should allow me to do the same operations the compiler is generating on the ALU in GPRs for each vector element before the horizontal sum. Or even just summing into a higher precision and doing the modulus at the end could get us there.

KungFuJesus:
@mtl1979 I just pushed another commit into this PR that marginally has a real impact on efficiency.

mtl1979:
@KungFuJesus Masked loads require AVX512-capable processor, otherwise the set1 intrinsics can't be mapped to single instruction. Bitwise operations on vector registers can be really slow on older processors.

KungFuJesus:
> @KungFuJesus Masked loads require AVX512-capable processor, otherwise the set1 intrinsics can't be mapped to single instruction. Bitwise operations on vector registers can be really slow on older processors.

I'm not using masked loads (though avx does support them, mostly for float domains). The set1 intrinsic is typically shorthand for the vpbroadcast/w/d/q family of instructions. I'm applying a mask by creating full sized ymm registers so that after the relatively quick broadcast function, I can zero out all but the carry over from the last round of sums. It's shorthand for the s1[7] = adler behavior the code was already doing. In integer domain, on all AVX2 capable CPUs I've seen, bitwise masks are very low latency and can often be done on multiple execution ports, doing at least 2 per cycle.  Basically the reason I'm broadcasting to all 8 elements and then zeroing out the first 7 is because there's no idiom to set the last element of a vector to a value (though there is a way to do a "scalar" move to set the first).  If the vectors were only 128 bits wide we could bit shift, but with AVX2 you can't shift outside of the 128 bit lanes.  We can obviously do a permute to do this, but, that is also going to be a slower sequence.  The vpinsert set of instructions could conceivably allow us to insert into a vector at a fixed index position, however, for 256 bit wide vectors, the "sequence" that it compiles to also involves creating 2 128 bit vectors and combining them.  

This code compiles cleanly on haswell but should work on any avx2 capable CPU.

nmoinvaz:
@mtl1979 All the intrinsic used start with mm256 how is that avx512?

Edit: oh maybe you were replying to a different comment. Never mind.

KungFuJesus:
> @mtl1979 All the intrinsic used start with mm256 how is that avx512?

I assure you it isn't, I don't even have a capable avx512 CPU in hand yet, hah. I tested this code thoroughly, -mavx2 should be all you need for GCC flags.

mtl1979:
@nmoinvaz Some move/load instructions can map to combination of SSE and AVX instructions on processors that don't support AVX512. This will cause slowdown on processors where there is high penalty moving data between SSE and AVX units. AVX512-capable processors can move data from general-purpose registers directly to AVX unit.

@KungFuJesus I have Haswell-based processor and it's really slow executing bitwise operations on vector registers... It's basically faster to move the data to general-purpose registers and then doing the same there.

--- 
To sum things up, it might be that gcc can fix the slowdowns by mapping the slow intrinsics to faster equivalents, but that's what the original code already tried to do. As such we need to benchmark the pull request with and without the set1+and combination. I didn't notice anything else wrong in the changes, but it was also middle of the night and I was reading the patch on my phone...

KungFuJesus:
> @nmoinvaz Some move/load instructions can map to combination of SSE and AVX instructions on processors that don't support AVX512. This will cause slowdown on processors where there is high penalty moving data between SSE and AVX units. AVX512-capable processors can move data from general-purpose registers directly to AVX unit.
> 
> @KungFuJesus I have Haswell-based processor and it's really slow executing bitwise operations on vector registers... It's basically faster to move the data to general-purpose registers and then doing the same there.
> 
> To sum things up, it might be that gcc can fix the slowdowns by mapping the slow intrinsics to faster equivalents, but that's what the original code already tried to do. As such we need to benchmark the pull request with and without the set1+and combination. I didn't notice anything else wrong in the changes, but it was also middle of the night and I was reading the patch on my phone...

Every compiler worth anything will translate ```_mm256_set1_*``` intrinsics to the vex encoded instructions and will not involve the AVX to SSE transition penalty.  What bitwise operations are you referring to?  The AND'ing of bits?
https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_and&ig_expand=350
That's 1 cycle of latency.  Haswell isn't listed on there but Agner Fog's tables still show it as the same https://www.agner.org/optimize/instruction_tables.pdf (1 cycle of latency, 3 possible execution ports).

Here's what GCC compiles the set1 + AND to:
```
vmovd  %r10d,%xmm7
vmovd  %edx,%xmm8
vpbroadcastd %xmm7,%ymm7
vpbroadcastd %xmm8,%ymm8
vpand  %ymm7,%ymm6,%ymm7
vpand  %ymm8,%ymm6,%ymm8
```
It takes the function call arguments, passed via registers, uses the vex encoded move to move them to XMM registers, then broadcasts them.  The ymm6 mask register is loaded via memory once, at the beginning of the function call.

mtl1979:
@KungFuJesus That assembler output is still wrong... `vpbroadcastd` and vpand`` will eliminate each other, so it's dead code...

KungFuJesus:
> @KungFuJesus That assembler output is still wrong... `vpbroadcastd` and vpand`` will eliminate each other, so it's dead code...

I'm not sure I follow...the "AND" operation that happens here is with the mask:
0, 0, 0, 0, 0, 0, 0, 0xFFFFFFFF
This effectively does the following with %r10d and %rdx (the contents of the adler and sum2 function arguments):
xmm7: { %r10d, 0, 0, 0}.  It then broadcasts the contents of that register to all 8 lanes of the aliased ymm register:
ymm7: {%r10d, %r10d, %r10d, %r10d, %r10d, %r10d, %r10d, %r10d}.  Then finally, it AND's it with the mask configured above such that we only have the value set in the last lane:
ymm7: {0, 0, 0, 0, 0, 0, 0, %r10d}

The same happens for xmm8 (don't recall which is s1 and which is s2).  This is effectively the same thing as your memset(), s[7] = adler sequence, as the compiler would have to generate code that populated the ymm register at that vector position.  This just does it in fewer cycles.  

I assure you broadcasting an xmm register to a ymm register is idiomatic for setting all lanes for ymm register and it is _not_ dead code (and the compiler, at this optimization level, would have certainly eliminated it in an early stage).
The code that's in develop right now is compiling to this sequence:

```
vpxor  %xmm0,%xmm0,%xmm0
movl   $0x0,0x18(%rsp)
movq   $0x0,0x10(%rsp)
mov    %r9d,0x1c(%rsp)
movq   $0x0,0x30(%rsp)
movl   $0x0,0x38(%rsp)
mov    %eax,0x3c(%rsp)
vmovdqa %xmm0,(%rsp)
vmovdqa %xmm0,0x20(%rsp)
...
vmovdqa (%rsp),%ymm4
vmovdqa 0x20(%rsp),%ymm7

```

That is, it's using a zero'ing idiom to fill the xmm register, then it's moving _all_ of the zeros to a location on the stack, as well as the values contained in the registers to a offset at an aligned location.  Then finally, it's moving the zeros generated in xmm0 to a contiguous location on stack just before where the actual values are stored, giving 4 leading zeros before a series of zeros followed by the value.

The code I'm proposing here skips _all_ of that stack churn, and while it amounts to a micro optimization, it's fewer instructions and faster.  The real optimization, though, lies in not having to do a modulus with every vector element that gets extracted.  This saves 8 different 64 bit integer multiplications, shifts, and subtractions.

mtl1979:
@KungFuJesus Your explanation just asserts that it's basically doing vector element reversal... As far as I remember vector units are big-endian even on little-endian CPUs... XMM and YMM registers with equal numbers do overlay on x86_64.

KungFuJesus:
> @KungFuJesus Your explanation just asserts that it's basically doing vector element reversal... As far as I remember vector units are big-endian even on little-endian CPUs... XMM and YMM registers with equal numbers do overlay on x86_64.

Yes...the code generated is exploiting that fact.  I'm laying out positions in memory order here, which of course fights with the convention of the little endian sometimes used to describe the vector registers (e.g. setr vs set intrinsics).  Ignoring all that, you were doing:
memset(s1, 0, 7); s1[7] = adler;
This is doing the equivalent, in fewer instructions, and fewer trips to the stack.  The first element (in memory order) gets broadcasted to all 8 elements.  I AND the result of that with a mask where only the 8th element is nonzero, resulting in the same thing.  The last position element in the partial sums is supposed to be the carry over from previous checksum loops or function calls, as that's how the recurrence relation with the multiplication of (32-i)*vector plays out.  

Edit: Are you maybe confused by the vpbroadcastd xmm7, ymm7?  Yes, they do overlay but this is valid and is idiomatic for setting all values in a vector register to the first value (just as pbroadcastd xmm7, xmm7 would be).  This does not blow away what's already in the register any more than an instruction that accumulates back to the same register.

mtl1979:
@KungFuJesus Like I said earlier, I'm not rejecting the code, I just want to make sure it's the fastest possible way to do things and that future developers know why it is done this way... This is where proper inline comments are useful...

KungFuJesus:
> @KungFuJesus Like I said earlier, I'm not rejecting the code, I just want to make sure it's the fastest possible way to do things and that future developers know why it is done this way... This is where proper inline comments are useful...

In general, if you want all values in a vector register to be the same, ```_mm256_set1_*``` will generate an optimal sequence from the compiler.  Most of the time it invokes the broadcast instruction and in scenarios where there might be a shorter sequence, it will emit that as well.  As far as the AND mask - well I thought that bit was obvious but if not, I can put comments there (though, the original code didn't explain why s[0...6] = 0; s[7] = adler; was being done, either).  Perhaps a link to the Intel whitepaper for this code would be appropriate?  

Is there anything else you think merits explanatory comments?

mtl1979:
@KungFuJesus Part of what I do here requires me to think as dumber person than I actually am... If that makes me unsure about something, then most likely some other contributors are unsure also... A lot of existing code could warrant better explaining, but my job here is not to comment on existing code, only code that is modified or added... I have been around since early days of zlib-ng, but I haven't reviewed every commit or pull request because I have also other projects to work on and sometimes I had to reinstall operating system to fix issues and that reduced the time I have for any third-party projects.

I will do final review after one of the other reviewers have done benchmarking so we can see how much faster this code actually is compared to baseline results. I usually don't do benchmarking myself as I run Linux virtualized, so it doesn't have high-precision timers.

KungFuJesus:
> @KungFuJesus Part of what I do here requires me to think as dumber person than I actually am... If that makes me unsure about something, then most likely some other contributors are unsure also... A lot of existing code could warrant better explaining, but my job here is not to comment on existing code, only code that is modified or added... I have been around since early days of zlib-ng, but I haven't reviewed every commit or pull request because I have also other projects to work on and sometimes I had to reinstall operating system to fix issues and that reduced the time I have for any third-party projects.
> 
> I will do final review after one of the other reviewers have done benchmarking so we can see how much faster this code actually is compared to baseline results. I usually don't do benchmarking myself as I run Linux virtualized, so it doesn't have high-precision timers.

Would that be @Dead2 we're waiting on @nmoinvaz?  I could run the benchmarks on a variety of hardware, as well, just let me know.

nmoinvaz:
@KungFuJesus we usually run benchmarks using https://github.com/zlib-ng/deflatebench. Compile two binaries - one with and one without the changes and then run it for each.

Here is an example: https://github.com/zlib-ng/zlib-ng/pull/934#issuecomment-864765447

KungFuJesus:
> @KungFuJesus we usually run benchmarks using https://github.com/zlib-ng/deflatebench. Compile two binaries - one with and one without the changes and then run it for each.
> 
> Here is an example: [#934 (comment)](https://github.com/zlib-ng/zlib-ng/pull/934#issuecomment-864765447)

I'm pretty sure minideflate's going to be the only thing that exercises this code, right?

nmoinvaz:
We typically use `minigzip`. You might want to test against the larger 200MB corpus mentioned at the bottom of the README. It will probably better show the results for what you are doing.

It takes some time to get the script setup, but it is really useful for benchmarking.

KungFuJesus:
Well sure, but the question is will minigzip actually leverage the adler checksums?  I thought adler32 wasn't called by gzip, but maybe I'm mistaken.

Edit: Looking around it sure _seems_ like gzip eschews the adler32 checksum in the envelop for a CRC32 one instead.  That's mostly from looking around for into about gzip, anyway.

KungFuJesus:
Hmm, minideflate is giving me "deflate error: -3".  So it seems like this tool's broken or never been checked against deflate?  That error happens because it's trying to expand a gzip compressed file rather than a simple DEFLATE stream which has headers that actually _do_ use adler32, if I'm reading things correctly.  Is there some flag I actually need to pass to this to make it use deflate instead?

Scratch that, the file appears to be authored by minideflate, and file is just calling it "data".  Same problem exists on the develop branch of zlib-ng, though, so it's something weird with either how minideflate is built or used.

I of course can't skip the checksum bits as that's where my optimizations lie.  I'll try minigzip, which does appear to be working.  But I'm fairly certain that's not going to touch the code I have in this PR.

KungFuJesus:
So a quick dumb benchmark I wrote to get around minideflate being broken:
```
#!/bin/env python3

import zlib
import sys
import os

fileBytes = None
with open(sys.argv[1], 'rb') as f:
    fileBytes = f.read()

# Let's artificially double the size
fileBytes += fileBytes

pid = os.getpid()
os.system('perf stat -D 0 -e cpu-clock:u -p {} &'.format(pid))
csum = zlib.adler32(fileBytes)
csum = zlib.adler32(fileBytes, csum)
csum = zlib.adler32(fileBytes, csum)
csum = zlib.adler32(fileBytes, csum)
csum = zlib.adler32(fileBytes, csum)
csum = zlib.adler32(fileBytes, csum)
csum = zlib.adler32(fileBytes, csum)
csum = zlib.adler32(fileBytes, csum)
csum = zlib.adler32(fileBytes, csum)
csum = zlib.adler32(fileBytes, csum)
csum = zlib.adler32(fileBytes, csum)
csum = zlib.adler32(fileBytes, csum)
csum = zlib.adler32(fileBytes, csum)
csum = zlib.adler32(fileBytes, csum)
csum = zlib.adler32(fileBytes, csum)
```

It doesn't do anything fancy like take a ton of averages - though someone could make it do this.  This strictly checks the adler32 checksum code in zlib.  I injected zlib-ng from develop into the preload stage, as well as my modifications.  I disabled turbo and fixed the frequency of my CPU, just like deflatebench had been doing.  I also niced the process to -20.

Here's what's on develop's output (using the same example file called out in the deflatebench repo):
```
 Performance counter stats for process id '26505':

          1,273.15 msec cpu-clock:u               #    0.633 CPUs utilized          

       2.009822400 seconds time elapsed
```

Here's with my branch:
```
Performance counter stats for process id '26541':

          1,198.50 msec cpu-clock:u               #    0.596 CPUs utilized          

       2.009962245 seconds time elapsed
``` 

As expected, the improvements are marginal but measurable.  I'm pretty sure the elapsed wall clock is meaningless in this microbenchmark context.

mtl1979:
1275.15 ms -> 1198.50 ms is about 5.86% improvement, which is nicely outside error margin of benchmarks...

KungFuJesus:
Actually, I re-scoreboarded my horizontal sum code and found that it does take fewer cycles if we do a horizontal sum instead of extracting each value to a GPR :).  This is mostly due to the fact that vpextract requires 24 cycles and none of the work on the ALU can overlap the extraction.  I'll push the commit into this PR now.

arch/x86/adler32_avx.c
```diff
@@ -9,11 +9,22 @@
 #include "../../zutil.h"
 
 #include "../../adler32_p.h"
+#include <stdio.h>
```
mtl1979: We shouldn't need this header...

KungFuJesus: Fixed

KungFuJesus:
Anything else holding up this PR?  I've got another branch with AVX512 support on the adler checksums and plan to extend it a bit further.

nmoinvaz:
@Dead2 can you approve the PR to run in workflows?

arch/x86/adler32_avx.c
```diff
@@ -14,6 +14,16 @@
 
 #ifdef X86_AVX2_ADLER32
 
+/* 64 bit horizontal sum, adapted from Agner Fog's
+ * vector library. */
```
nmoinvaz: No need for comment to extend 2 lines.

arch/x86/adler32_avx.c
```diff
@@ -14,6 +14,16 @@
...
+static inline uint64_t hsum(__m256i x)
+{
```
nmoinvaz: Opening bracket should be on same line as function.

arch/x86/adler32_avx.c
```diff
@@ -77,29 +78,71 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
...
+        __m256i s1lo256 = _mm256_cvtepi32_epi64(s1lo);
+        __m256i s1hi256 = _mm256_cvtepi32_epi64(s1hi);
+        __m256i s2lo256 = _mm256_cvtepi32_epi64(s2lo);
+        __m256i s2hi256 = _mm256_cvtepi32_epi64(s2hi);
+
+        /* sum vectors in existing lanes */
+        __m256i s1Sum = _mm256_add_epi64(s1lo256, s1hi256);
+        __m256i s2Sum = _mm256_add_epi64(s2lo256, s2hi256);
+        
+        /* In AVX2-land, this trip through GPRs will probably
```
nmoinvaz: This comment can be made as long as the large comment above it and so condensed into fewer lines.

arch/x86/adler32_avx.c
```diff
@@ -77,29 +78,71 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
...
+        __m128i s1hi = _mm256_extracti128_si256(vs1, 1);
+        __m128i s2hi = _mm256_extracti128_si256(vs2, 1);
+        
+        /* Convert up to 64 bit precision to prevent overflow */
+        __m256i s1lo256 = _mm256_cvtepi32_epi64(s1lo);
+        __m256i s1hi256 = _mm256_cvtepi32_epi64(s1hi);
+        __m256i s2lo256 = _mm256_cvtepi32_epi64(s2lo);
+        __m256i s2hi256 = _mm256_cvtepi32_epi64(s2hi);
+
+        /* sum vectors in existing lanes */
```
nmoinvaz: Capital "s" for consistency here.

arch/x86/adler32_avx.c
```diff
@@ -77,29 +78,71 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
...
+        __m256i s1Sum = _mm256_add_epi64(s1lo256, s1hi256);
```
nmoinvaz: Variable name for s1Sum should be `s1sum` or `s1_sum`. Same for s2Sum.

nmoinvaz (changes requested):
I have made some comments about code formatting only.

arch/x86/adler32_avx.c
```diff
@@ -33,30 +43,21 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
...
-         1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1};
-    __m256i dot1v = _mm256_load_si256((__m256i*)dot1);
-    char ALIGNED_(32) dot2[32] = \
-        {32, 31, 30, 29, 28, 27, 26, 25, 24, 23, 22, 21, 20, 19, 18, 17,
-         16, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1};
-    __m256i dot2v = _mm256_load_si256((__m256i*)dot2);
-    short ALIGNED_(32) dot3[16] = \
-        {1, 1, 1, 1, 1, 1, 1, 1,  1, 1, 1, 1, 1, 1, 1, 1};
-    __m256i dot3v = _mm256_load_si256((__m256i*)dot3);
+    const __m256i vsMask = _mm256_setr_epi32(0, 0, 0, 0, 0, 0, 0, -1);
```
nmoinvaz: Variable name vsMask should be `vsmask` or `vs_mask`.

nmoinvaz:
Might want to rebase to see if that fixes the CI issues.

KungFuJesus:
Is it maybe alright if I merge from develop instead?  This is not rebasing cleanly because I merged with another branch I was working on :-/.  Unless my git fu is somehow poor (likely). 

Actually, a rebase isn't going to fix this, anyway.  I see the issue (the intrinsic I used is defined for all compilers).  I have to add a macro to conditionally define it if it isn't present.  

Edit: it's actually even more interesting.  That intrinsic isn't supported for clang on 32 bit compiles.  Give me a bit, I'll come up with a decent workaround.

nmoinvaz:
I think you would need to rebase to maintain clean git history. We don't use merge commits on this repo.

arch/x86/adler32_avx.c
```diff
@@ -19,7 +19,14 @@ static inline uint64_t hsum(__m256i x) {
     __m256i sum1 = _mm256_shuffle_epi32(x, 0x0E);
     __m256i sum2 = _mm256_add_epi64(x, sum1);
     __m128i sum3 = _mm256_extracti128_si256(sum2, 1);
+#ifdef __X86_64__
     return _mm_cvtsi128_si64(_mm_add_epi64(_mm256_castsi256_si128(sum2), sum3));
+#else
+    __m128i retVec = _mm_add_epi64(_mm256_castsi256_si128(sum2), sum3);
```
nmoinvaz: Don't forget `retVec` and `retVal` are not in the proper casing for the project.

KungFuJesus:
That should compile cleanly now on 32 bit systems

arch/x86/adler32_avx.c
```diff
@@ -14,6 +14,21 @@
 
 #ifdef X86_AVX2_ADLER32
 
+/* 64 bit horizontal sum, adapted from Agner Fog's vector library. */
+static inline uint64_t hsum(__m256i x) {
+    __m256i sum1 = _mm256_shuffle_epi32(x, 0x0E);
+    __m256i sum2 = _mm256_add_epi64(x, sum1);
+    __m128i sum3 = _mm256_extracti128_si256(sum2, 1);
+#ifdef __X86_64__
```
phprus: ``__x86_64__``, not ``__X86_64__``, and ``__x86_64__ `` is undefined on Microsoft Visual Studio.
Use
```c++
#if defined(__x86_64__) || defined(_M_X64)
```

like this:
https://github.com/zlib-ng/zlib-ng/blob/f77af71e77cdb9d41d5e3a606e2a8dd67034e387/cmake/detect-arch.c#L7

nmoinvaz:
@Dead2 this needs approval again?

arch/x86/adler32_avx.c
```diff
@@ -77,29 +83,69 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
...
+        __m256i s1lo256 = _mm256_cvtepi32_epi64(s1lo);
+        __m256i s1hi256 = _mm256_cvtepi32_epi64(s1hi);
+        __m256i s2lo256 = _mm256_cvtepi32_epi64(s2lo);
+        __m256i s2hi256 = _mm256_cvtepi32_epi64(s2hi);
+
+        /* Sum vectors in existing lanes */
+        __m256i s1_sum = _mm256_add_epi64(s1lo256, s1hi256);
+        __m256i s2_sum = _mm256_add_epi64(s2lo256, s2hi256);
+        
+        /* In AVX2-land, this trip through GPRs will probably  be unvoidable, as there's no cheap and easy
```
nmoinvaz: Extra space between _probably_ and _be_

KungFuJesus: Hah, I noticed that too but thought it'd be OCD for me to force push another commit for that one.  Since you noticed it, doing it now.

KungFuJesus:
On a somewhat unrelated note - maybe someone wants to take a look at what's going on with minideflate?  That benchmark repo appears to be invoking minideflate correctly but the program is hanging up prematurely (and it's happening on develop, so it's not related to this PR).

nmoinvaz:
@KungFuJesus can you create a new GitHub issue for that in this repo so it doesn't get lost?

P.S. thanks for fixing those nit-picky formatting issues.

nmoinvaz:
Looks like all CI tests pass nice.

KungFuJesus:
Cool.  What's next for merging?

nmoinvaz:
I think benchmarks using `deflatebench` and then review by @Dead2.

KungFuJesus:
> I think benchmarks using `deflatebench` and then review by @Dead2.

Ah, gotta fix minideflate first, then.

nmoinvaz:
@KungFuJesus I have made some changes to `minideflate` in #1060, let me know if that fixes your problem.

KungFuJesus:
So I managed to run the deflatebenchmark, at least with the verify option disabled.  Though, I'm not quite convinced minideflate is actually producing correct output, still (my md5sums didn't match), the times between develop and my branch should still be commensurate.  As expected, it's not a gargantuan improvement due to the adler checksums already being a relatively small portion of the runtime.

Here's vanilla (with your fixes):
```
 Tool: minideflate Levels: 0-9       
 Runs: 15         Trim worst: 5         

 0    100.015%      0.265/0.310/0.335/0.022        0.270/0.295/0.318/0.018      211,990,111
 1     44.407%      3.517/3.565/3.616/0.030        2.504/2.546/2.573/0.023       94,124,037
 2     35.518%      6.156/6.201/6.236/0.023        2.561/2.588/2.609/0.017       75,282,949
 3     33.882%      8.165/8.211/8.267/0.027        2.425/2.464/2.511/0.026       71,816,466
 4     33.174%      8.990/9.067/9.156/0.061        2.381/2.412/2.451/0.024       70,315,656
 5     32.672%      9.871/9.906/9.958/0.028        2.358/2.393/2.418/0.016       69,249,937
 6     32.517%   11.257/11.360/11.458/0.065        2.351/2.372/2.396/0.016       68,921,480
 7     32.255%   15.287/15.368/15.415/0.041        2.300/2.334/2.360/0.018       68,366,788
 8     32.167%   22.518/22.599/22.746/0.066        2.328/2.351/2.371/0.015       68,180,764
 9     31.887%   24.250/24.352/24.443/0.062        2.369/2.475/2.500/0.040       67,586,430

 avg1  40.849%                       11.094                          2.223
 avg2  45.388%                       12.327                          2.470
 tot                               1109.390                        222.302      865,834,618
```
Here's with my optimizations:
```
 Tool: minideflate Levels: 0-9       
 Runs: 15         Trim worst: 5         

 0    100.015%      0.278/0.301/0.325/0.017        0.232/0.267/0.290/0.021      211,990,111
 1     44.407%      3.567/3.597/3.626/0.021        2.503/2.534/2.569/0.024       94,124,037
 2     35.518%      6.157/6.207/6.250/0.032        2.542/2.587/2.631/0.032       75,282,949
 3     33.882%      7.922/8.007/8.051/0.039        2.453/2.490/2.514/0.018       71,816,466
 4     33.174%      8.834/8.895/8.944/0.040        2.387/2.426/2.460/0.026       70,315,656
 5     32.672%     9.870/9.938/10.038/0.054        2.345/2.372/2.410/0.021       69,249,937
 6     32.517%   11.298/11.358/11.426/0.045        2.332/2.368/2.402/0.020       68,921,480
 7     32.255%   15.261/15.330/15.460/0.075        2.260/2.332/2.360/0.030       68,366,788
 8     32.167%   22.487/22.596/22.706/0.080        2.274/2.322/2.345/0.022       68,180,764
 9     31.887%   24.276/24.376/24.419/0.052        2.417/2.459/2.492/0.024       67,586,430

 avg1  40.849%                       11.061                          2.216
 avg2  45.388%                       12.289                          2.462
 tot                               1106.051                        221.570      865,834,618
```
Of course my little micro benchmark that looked at the adler checksums in isolation is _probably_ a better measuring stick for this, I think.

nmoinvaz:
This code below can be found in `adler32_ssse3`:
```
adler = (s1_unpack[0] % BASE) + (s1_unpack[1] % BASE) + (s1_unpack[2] % BASE) + (s1_unpack[3] % BASE) +
        (s1_unpack[4] % BASE) + (s1_unpack[5] % BASE) + (s1_unpack[6] % BASE) + (s1_unpack[7] % BASE);
adler %= BASE;
s1[7] = adler;

sum2 = (s2_unpack[0] % BASE) + (s2_unpack[1] % BASE) + (s2_unpack[2] % BASE) + (s2_unpack[3] % BASE) +
       (s2_unpack[4] % BASE) + (s2_unpack[5] % BASE) + (s2_unpack[6] % BASE) + (s2_unpack[7] % BASE);
sum2 %= BASE;
s2[7] = sum2;
```
Can a similar optimization be made there too?

mtl1979:
@nmoinvaz I'm pretty sure some of the required instructions either require SSE4.1 or don't exist for any (S)SSE variant.

KungFuJesus:
> This code below can be found in `adler32_ssse3`:
> 
> ```
> adler = (s1_unpack[0] % BASE) + (s1_unpack[1] % BASE) + (s1_unpack[2] % BASE) + (s1_unpack[3] % BASE) +
>         (s1_unpack[4] % BASE) + (s1_unpack[5] % BASE) + (s1_unpack[6] % BASE) + (s1_unpack[7] % BASE);
> adler %= BASE;
> s1[7] = adler;
> 
> sum2 = (s2_unpack[0] % BASE) + (s2_unpack[1] % BASE) + (s2_unpack[2] % BASE) + (s2_unpack[3] % BASE) +
>        (s2_unpack[4] % BASE) + (s2_unpack[5] % BASE) + (s2_unpack[6] % BASE) + (s2_unpack[7] % BASE);
> sum2 %= BASE;
> s2[7] = sum2;
> ```
> 
> Can a similar optimization be made there too?

They could, I could probably put in a separate PR for it.

KungFuJesus:
> @nmoinvaz I'm pretty sure some of the required instructions either require SSE4.1 or don't exist for any (S)SSE variant.

The only one I really see here that is SSE4 only is the cheap conversion from 32 to 64 bit.  This could be emulated with some byte level swizzling (with SSSE3, anyway).  Is there any reason though we couldn't write an SSE3 and an SSE4 version?  The number of CPUs that have SSE3 and not SSSE3 is also likely a pretty small number.

mtl1979:
@KungFuJesus I used to have versions for every variant of SSE and XOP instruction sets... There was some issues with Visual C++ not being able to compile the code correctly or generating bad code. "Multiply high" + "multiply low" was one trick to widen vector elements, but there is no 32-bit version for SSE.

KungFuJesus:
> @KungFuJesus I used to have versions for every variant of SSE and XOP instruction sets... There was some issues with Visual C++ not being able to compile the code correctly or generating bad code. "Multiply high" + "multiply low" was one trick to widen vector elements, but there is no 32-bit version for SSE.

Ugh, that doesn't surprise me, given that Microsoft has removed the ability to shut off compilations <= SSE4.  `#ifdef _MSVC` don't bother?  I don't know, it just seems silly that you have to go back 10 or 15 years to get that instruction granularity.  

We can widen elements easily (and in the same number of cycles) by using pshufb with a permutation vector (using -1 for the upper half of the words).  That's SSSE3.

mtl1979:
@KungFuJesus gcc has dropped support for targeting older than SSE4 too... It generates random code if you try to target SSSE3 or older and use anything more complex... Basically it chokes when it runs out of registers and starts spilling worse than a 1 month  old baby...

KungFuJesus:
Really?  I've not seen that behavior but it's rare I have to target things much older than SSE4 for a while.  The register spilling when compiling with -m32 makes a lot of sense, though, as pre-amd64 the register file was smaller (8 XMM registers instead of 16).  I'd be surprised if -msse3 -mssse3 generated illegal code but if it did, it's worth filing a bug upstream.

I imagine given the mainstream distros dropping i386 altogether, these issues could go unnoticed for a while, but I think it's worth finding these things and reporting them if still true.  It's definitely worth bringing these improvements forward on these older systems, as moves to and from stack are going to be more expensive due to the reduced cache and memory bandwidths.  After this gets merged I can look into what GCC generates for this on SSSE3.

One other optimization that's still on the table that might be a decent improvement - I noticed disassembling IPP's adler32 implementation (though still only 16 bytes wide), they are managing to do this without the 2 x 8 bit wide multiply accumulates.  They are instead using a sum of absolute differences instead (which does have shorter instruction latency), and they are somehow moving a multiplication to the outside of the main loop.  I haven't looked very hard for how they managed to factor that bit out (the Intel whitepaper for both their SIMD optimized adler and fletcher checksums seem to be gone from the internet, now), but it's possible maybe to shorten a critical instruction latency in the data dependency chain.

mtl1979:
@KungFuJesus There was a lot of issues with VC++ 2017... With gcc, I filed bug report, but there's also differences in how Intel and AMD processors implement SSE. Things worked a lot better when I targeted Intel Core 2 instead of generic x86_64, but that did break code for all AMD processors.

KungFuJesus:
Sorry to necrobump, but is there anything currently pending on me for this PR?

mtl1979:
@KungFuJesus @Dead2 is often slow at merging pull requests. If all requested changes have been made, then there is nothing more to improve...

Dead2:
This PR cannot currently be merged due to merge conflicts, so it should be rebased.
Also, I'd like someone else to tag it with their approval first.

arch/x86/adler32_avx.c
```diff
@@ -33,30 +48,21 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
...
-
-    // We will need to multiply by
-    char ALIGNED_(32) shift[16] = {5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
-    __m128i shiftv = _mm_load_si128((__m128i*)shift);
+    const __m256i dot1v = _mm256_set1_epi8(1);
+    const __m256i dot2v = _mm256_setr_epi8(32, 31, 30, 29, 28, 27,
+                                           26, 25, 24, 23, 22, 21,
+                                           20, 19, 18, 17, 16, 15,
+                                           14, 13, 12, 11, 10, 9,
+                                           8, 7, 6, 5, 4, 3, 2, 1);
```
mtl1979: 6 array elements per line will make the lines too short. Even with 16 elements per line, it doesn't go past the 120 character limit per line.

KungFuJesus: I see this project does not subscribe to the 80 col religion like I do.  Fixing now.

KungFuJesus:
> This PR cannot currently be merged due to merge conflicts, so it should be rebased. Also, I'd like someone else to tag it with their approval first.

Hmm, weird, github _says_ it's fine.  Where are the conflicts?

nmoinvaz:
Here are my benchmark tests:
```
adler32_benchmark.exe --benchmark_repetitions=250
2021-12-01T13:22:52-05:00
Running adler32_benchmark.exe
Run on (8 X 2995 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x4)
  L1 Instruction 32 KiB (x4)
  L2 Unified 1280 KiB (x4)
  L3 Unified 12288 KiB (x1)
...
adler32_avx2_old_bench_mean        34079 ns        33961 ns          250
adler32_avx2_old_bench_median      32650 ns        32890 ns          250
adler32_avx2_old_bench_stddev       4855 ns         4624 ns          250
adler32_avx2_old_bench_cv          14.25 %         13.61 %           250
...
adler32_avx2_new_bench_mean        29302 ns        29266 ns          250
adler32_avx2_new_bench_median      28916 ns        29157 ns          250
adler32_avx2_new_bench_stddev       1787 ns         1787 ns          250
adler32_avx2_new_bench_cv           6.10 %          6.11 %           250
```
Here is my code: https://gist.github.com/nmoinvaz/b31b11c35724a549476b37da4ef8cd17. All you have to do is clone https://github.com/google/benchmark into the same directory and then run CMake.

Using median values there is possibly 12% improvement?

nmoinvaz (approved):

nmoinvaz:
It would be better if the style nits commit was squashed - and that particular change if the same number of array elements were on each line, but it is not a condition of approval for me so I have approved this PR.

KungFuJesus:
> Here are my benchmark tests:
> 
> ```
> adler32_benchmark.exe --benchmark_repetitions=250
> 2021-12-01T13:22:52-05:00
> Running adler32_benchmark.exe
> Run on (8 X 2995 MHz CPU s)
> CPU Caches:
>   L1 Data 48 KiB (x4)
>   L1 Instruction 32 KiB (x4)
>   L2 Unified 1280 KiB (x4)
>   L3 Unified 12288 KiB (x1)
> ------------------------------------------------------------------------
> Benchmark                              Time             CPU   Iterations
> ------------------------------------------------------------------------
> ...
> adler32_avx2_old_bench_mean        34079 ns        33961 ns          250
> adler32_avx2_old_bench_median      32650 ns        32890 ns          250
> adler32_avx2_old_bench_stddev       4855 ns         4624 ns          250
> adler32_avx2_old_bench_cv          14.25 %         13.61 %           250
> ...
> adler32_avx2_new_bench_mean        29302 ns        29266 ns          250
> adler32_avx2_new_bench_median      28916 ns        29157 ns          250
> adler32_avx2_new_bench_stddev       1787 ns         1787 ns          250
> adler32_avx2_new_bench_cv           6.10 %          6.11 %           250
> ```
> 
> Here is my code: https://gist.github.com/nmoinvaz/b31b11c35724a549476b37da4ef8cd17. All you have to do is clone https://github.com/google/benchmark into the same directory and then run CMake.
> 
> Using median values there is possibly 12% improvement?

Awesome, that proper benchmark should come in handy.
One possible explanation for the lower coefficient of variation is the reduction in stack trips.  The ymm registers are instead shuffling out directly to GPRs now.

KungFuJesus:
@Dead2 I can rebase with develop if needed but it doesn't look like Github presently has any issues merging this.

nmoinvaz:
Here is also an old adler32 benchmark I made once that is meant to be compiled with zlib-ng. https://gist.github.com/nmoinvaz/8bdc503804e17164b2d33312764feecc

Dead2:
@KungFuJesus
Screenshot 2021-12-01 212719

It does not tell me what the conflict is for some reason.

If you could rebase, lose the merge commit, and squash the last 3 fix-commits, leaving a clean and concise 3 commits, that would be very great. Hopefully that will resolve the conflict problem at the same time. 👍

KungFuJesus:
> @KungFuJesus Screenshot 2021-12-01 212719
> 
> It does not tell me what the conflict is for some reason.
> 
> If you could rebase, lose the merge commit, and squash the last 3 fix-commits, leaving a clean and concise 3 commits, that would be very great. Hopefully that will resolve the conflict problem at the same time. +1

Ugh, I've already branched from this branch to work on an AVX512 variant.  I mean, I should be able to, but it's going to require some annoying amount of git history revision to get there.  The one thing I don't love above rebase is how it royally screws anything that manages to fork from an earlier state.  I'll see about squashing this somehow.

nmoinvaz:
It is a pain but it does make for a cleaner git history in the repository. There are times when I have made PR in zlib-ng and I've had to do lots of rebasing and reworking and waited many months for acceptance. I think you're getting close. I'm looking forward to the AVX512 variant even though I don't have hardware to test it on.

KungFuJesus:
Like every time I attempt rebase - git and I both got royally confused and I ended up having to construct half the state of head of the current branch on the remote via a squashing commit :-/.  Anyway, it should be broken down to the desirable commits now - let me know if I screwed something up (I did make the backup of the pre rebase repo locally).

nmoinvaz:
LGTM

arch/x86/adler32_avx.c
```diff
@@ -33,30 +49,18 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
...
 
-    uint32_t ALIGNED_(32) s1[8], s2[8];
-
-    memset(s1, 0, sizeof(s1)); s1[7] = adler; // TODO: would a masked load be faster?
-    memset(s2, 0, sizeof(s2)); s2[7] = sum2;
+    const __m256i vs_mask = _mm256_setr_epi32(0, 0, 0, 0, 0, 0, 0, -1);
+    __m256i vs1 = _mm256_set1_epi32(adler);
+    __m256i vs2 = _mm256_set1_epi32(sum2);
+    vs1 = _mm256_and_si256(vs1, vs_mask);
+    vs2 = _mm256_and_si256(vs2, vs_mask);
```
mtl1979, line 55: I'm not 100% sure if it would be better or not to group setting up `vs1` first and then setting up `vs2` after that, so it matches how it is done below.

All four lines are short and the readability wouldn't hurt much from reducing the line count from 4 to 2.

arch/x86/adler32_avx.c
```diff
@@ -77,29 +81,70 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
...
+         * conversion from 64 bit integer to 32 bit. This casting to 32 bit is cheap through GPRs 
+         * (just register aliasing), and safe, as our base is significantly smaller than UINT32_MAX */
+        adler = (uint32_t)(hsum(s1_sum) % BASE);
+        sum2 = (uint32_t)(hsum(s2_sum) % BASE);
+
+        vs1 = _mm256_set1_epi32(adler);
+        vs1 = _mm256_and_si256(vs1, vs_mask);
+
+        vs2 = _mm256_set1_epi32(sum2);
+        vs2 = _mm256_and_si256(vs2, vs_mask);
```
mtl1979: See comment above... These lines are short, might be worth reducing them to 2 lines.

arch/x86/adler32_avx.c
nmoinvaz: As far as the line count goes I think the way it is is more readable.

mtl1979: I'm just thinking aloud... Currently it's just inconsistent with the block below in regards to grouping the lines... 

But also, in here we have to declare the types of the variables, which isn't necessary below.

Dead2:
@KungFuJesus You might want to try out git fixup, have a look at https://github.com/zlib-ng/zlib-ng/wiki/Git-workflow-tips if you are not familiar with it. 😄

KungFuJesus:
> @KungFuJesus You might want to try out git fixup, have a look at https://github.com/zlib-ng/zlib-ng/wiki/Git-workflow-tips if you are not familiar with it. smile

I mean that's essentially what I did for the squash.  That wasn't the issue I had, the issue was the rebase causing headaches with a merge commit that was established on this from another branch early on in the development (getting working horizontal sums was done on a different branch).  The conflict resolution for that caused headaches because it was trying to resolve deltas from older commits that had since been reformatted from early git amended commits and force pushes _because_ of all of those fixup changes.  It's working now at least, with maybe some less satisfactory states of commit in between, but I'm pretty sure it'll at least compile and bisect fine.  I just really dislike force push workflows because it breaks _anything_ that has a history established with anything up until that point.

KungFuJesus:
I should mention, if it isn't already obvious, that he failure in the CI checks seems to be a debian packaging related issue with the version of wine being installed.  An apt issue rather than a real one.

Dead2:
@KungFuJesus I understand the problems 😄 
We try to avoid force-pushing the repo main branches stable and develop (especially stable), but force-pushing within PRs is good, because it leads to a much cleaner commit log for the main branches, leading to easier external reviews and bisecting, etc.
In a way it is moving the pain to the PR process instead of having the potential for perpetual pain in the main branches 😉 

And yes, the failing CI is an everlasting problem it seems. The Github Actions build platforms are constantly changing, causing new problems every 1-2 months or so 🙄

arch/x86/adler32_avx.c
KungFuJesus: I tend to like to group declarations together.  I'm willing to reorder the block below it, since that's probably what the instruction scheduling stage of the optimizer should be doing anyway, since both of these operations can be performed on different execution ports in the same clock cycle (though, the OoO nature of the architecture is probably doing this speculatively, anyway).

KungFuJesus: Good now?

mtl1979: Yup... I don't want to be too picky, I just made one final review now that I have no projects that need my attention...

mtl1979 (approved):

KungFuJesus:
Hah, I came to the (now obvious) realization today that NMAX is derived to be the maximum _scalar_ sums that can be performed before overflowing.  This means the trip to 64 bit wide words wasn't even necessary, we can do a plain old 32 bit horizontal sum safely and probably save us a handful of cycles.  

If my math is right it should be 12 cycles for the conversion saved, plus a cycle or two for the additional sums.  Of course this also means the modulus operation probably could have been done on the FPU, but I'd wager it'd be more costly than doing it with GPRs thanks to register aliasing being so inexpensive and many of the instructions simply moving between GPRs and AVX registers instead of needing to go through the stack.

Issue #1059, opened 2021-11-10
Minideflate seems broken with zlib-ng


KungFuJesus:
Trying to run minideflate to decompress returns error -3.

```
adam@eggsbenedict ~/scratch/zlib-ng/build $ ./minideflate -c silesia.tar > test.def
adam@eggsbenedict ~/scratch/zlib-ng/build $ ./minideflate -d test.def > test.out
deflate error: -3
```

I've not tried to bisect to identify a problematic commit.

nmoinvaz:
Commit has been merged in `develop`.

PR #1060, opened 2021-11-12
Fixed minideflate write buffers being overwritten.

nmoinvaz:
See #1059 and #1053.  Certain stream info were not being set at the correct times on the stream, causing some data to be lost.

### DIFF
--- a/test/minideflate.c
+++ b/test/minideflate.c
@@ -70,6 +70,8 @@ void deflate_params(FILE *fin, FILE *fout, int32_t read_buf_size, int32_t write_
     c_stream.opaque = (void *)0;
     c_stream.total_in = 0;
     c_stream.total_out = 0;
+    c_stream.next_out = write_buf;
+    c_stream.avail_out = write_buf_size;
 
     err = PREFIX(deflateInit2)(&c_stream, level, Z_DEFLATED, window_bits, mem_level, strategy);
     CHECK_ERR(err, "deflateInit2");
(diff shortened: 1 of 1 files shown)


KungFuJesus:
So it doesn't bail out anymore with those changes, but something definitely seems wrong:
```
[adam@thinkpad build]$ ./minideflate -c /tmp/silesia.tar > ~/silesia.tar.df
[adam@thinkpad build]$ ./minideflate -d ~/silesia.tar.df >  ~/silesia.tar
[adam@thinkpad build]$ md5sum ~/silesia.tar
d41d8cd98f00b204e9800998ecf8427e  /home/adam/silesia.tar
[adam@thinkpad build]$ md5sum /tmp/silesia.tar 
42e120d96a85a495910b628718770599  /tmp/silesia.tar
```

Did I somehow invoke this wrong?  The deflatebench test also fails, claiming something isn't in  the gzip format:

```
Tool: minideflate
Activated single file mode
Level 0-9: /tmp/silesia.tar  202.1 MiB   211,957,760 B
Starting run 1 of 15
Testing level 0: cdv
gzip: /tmp/zlib-testfil.gz: not in gzip format
Failed, retval(1): gunzip -c /tmp/zlib-testfil.gz
```

KungFuJesus:
ahh, the gzip thing appears to be an oversight by the benchmarking tool.  That's gunzip trying to verify checksums, which don't apply of course, with deflate compression.

nmoinvaz:
I think your second command should be `./minideflate -c -d ~/silesia.tar.df >  ~/silesia.tar`

KungFuJesus:
Ah yes, you're correct. Reading the --help text would have served me there. The md5 checksums match after that.

nmoinvaz:
@Dead2 this is also ready.

Dead2 (approved):

PR #1063, opened 2021-12-04
Workaround for installation failure of wine32.

mtl1979:
Fixes version mismatch between i386 and amd64 packages of dependencies.


nmoinvaz:
The one CI that was failing passes so LGTM.

nmoinvaz (approved):

Issue #1066, opened 2021-12-07
Regression: inflate returns Z_DATA_ERROR on valid input (regressed by #1033)


sfc-gh-sgiesecke:
The following program works with zlib 1.2.7-10_el7_9 from CentOS 7, and with zlib-ng before #1033 was merged as dc3b60841dbfa9cf37be3efb4568f055b4e15580:
```
#include <assert.h>
#include <stdio.h>
#include <zlib.h>

const unsigned char compressedData[] = {
    0x78, 0x9c, 0x0b, 0xc9, 0x48, 0x55, 0x28, 0x2c, 0xcd, 0x4c, 0xce, 0x56, 0x48, 0x2a, 0xca, 0x2f, 0xcf, 0x53, 0x48, 0xcb, 0xaf, 0x50, 0xc8, 0x2a, 0xcd, 0x2d, 0x48, 0x4d, 0x51, 0xc8, 0x2f, 0x4b, 0x2d, 0x52, 0x28, 0xc9, 0x48, 0x55, 0xc8, 0x49, 0xac, 0xaa, 0x54, 0x48, 0xc9, 0x4f, 0x07, 0x00, 0x6b, 0x93, 0x10, 0x30};

int main()
{
    const int windowBits = 32 + MAX_WBITS;
    z_stream_s strm;
    strm.zalloc = Z_NULL;
    strm.zfree = Z_NULL;
    strm.opaque = Z_NULL;

    unsigned char decompressedData[1024];

    int rc = inflateInit2(&strm, windowBits);
    assert(rc == Z_OK);

    strm.next_in = (unsigned char *)compressedData;
    strm.avail_in = sizeof(compressedData);
    strm.next_out = decompressedData;
    strm.avail_out = sizeof(decompressedData);

    rc = inflate(&strm, Z_NO_FLUSH);
    assert(rc == Z_STREAM_END);

    rc = inflateEnd(&strm);
    assert(rc == Z_OK);

    decompressedData[strm.total_out] = 0;
    puts((const char *)decompressedData);

    return 0;
}
```
With #1031, the `inflate` calls retuns `Z_DATA_ERROR`.

Platform: Linux x86_64 (CentOS 7)

Built with: `./configure --zlib-compat --static && make`

nmoinvaz:
I think you forgot to include the program in your comment.

sfc-gh-sgiesecke:
> I think you forgot to include the program in your comment.

[...]

PR #1067, opened 2021-12-07
Don't overwrite adler32 hash with crc32 hash. 

nmoinvaz:
See #1066 for problem description.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1371,6 +1371,7 @@ if(ZLIB_ENABLE_TESTS)
 
     add_simple_test_executable(deflate_quick_bi_valid)
     add_simple_test_executable(deflate_quick_block_open)
+    add_simple_test_executable(inflate_adler32)
     add_simple_test_executable(hash_head_0)
 endif()
 
--- a/inflate.c
+++ b/inflate.c
@@ -502,7 +502,9 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
                 state->head->hcrc = (int)((state->flags >> 9) & 1);
                 state->head->done = 1;
             }
-            strm->adler = state->check = functable.crc32_fold_reset(&state->crc_fold);
+            /* compute crc32 checksum if not in raw mode */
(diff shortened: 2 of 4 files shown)


nmoinvaz:
I'm not sure if it should be `((state->wrap & 4) && (state->flags))`. I don't quite see exactly what the `state->wrap & 4` indicates.

test/inflate_adler_small.c
```diff
@@ -0,0 +1,59 @@
...
+#include <stdlib.h>
+#include <string.h>
+
+z_const unsigned char compressed[] = {
+    0x78, 0x9c, 0x0b, 0xc9, 0x48, 0x55, 0x28, 0x2c, 0xcd, 0x4c, 0xce, 0x56, 0x48, 0x2a, 0xca,
+    0x2f, 0xcf, 0x53, 0x48, 0xcb, 0xaf, 0x50, 0xc8, 0x2a, 0xcd, 0x2d, 0x48, 0x4d, 0x51, 0xc8,
+    0x2f, 0x4b, 0x2d, 0x52, 0x28, 0xc9, 0x48, 0x55, 0xc8, 0x49, 0xac, 0xaa, 0x54, 0x48, 0xc9,
+    0x4f, 0x07, 0x00, 0x6b, 0x93, 0x10, 0x30
+};
+z_const unsigned char* uncompressed_expected = "The quick brown fox jumped over the lazy dog";
```
mtl1979: Text literals are not "unsigned" on all compilers, so needs explicit cast.

mtl1979:
@nmoinvaz `state->wrap & 4` should mean enable/disable check value... Basically in inflate, it is used to enable or disable crc32 calculation in raw mode.

nmoinvaz:
@mtl1979 thanks for the great explanation. It is not documented very well so I have added a commit with some new comments.

mtl1979:
@nmoinvaz Even I didn't know that before today... I had to grep through the sources to see what the purpose of that bit is...

nmoinvaz:
Looks like CI passes now. This is good!

Dead2 (approved):

nmoinvaz:
Thanks for merging so quickly!

PR #1071, opened 2021-12-13
Fix deflateBound and compressBound returning very small size estimates.

Dead2:
Fixes deflateBound and compressBound returning very small size estimates.
Remove workaround in switchlevels.c, so we do actual testing of this.

I also made a few of the magic numbers more understandable by using defines,
I would have liked to do so with the rest too.

##
I reasoned that the `+ 13` comes from compressBound, and thus includes zlib header size (6 bytes).
That seems to be the reason why it was `+13 - 6` in deflateBound, since we do not always want to use the zlib header type there.
The remaining 7 bytes I am unsure about, raw deflate block headers? I was unable to find any documentation saying more than 2-3 bytes for that though, so I left that as a 7.

Also not sure why the `((sourceLen + 13 + 7) >> 3)` is used for deflate_quick, it kind of looks like those additions should have been outside of the bitshifting, but without any comments those are just magic numbers. They also fail to add enough when the buffer size is 1 byte for example, thus the reason for adding the `+1`.

##
With these changes, the buffer size ends up being a lot bigger, but several of the changes we have made have affected the worst case compressed size. Deflate_quick is a big one for example, possibly we could investigate why level 1 becomes bigger than level 0 for random data, but I suspect it has to do with skipping extra checks for raw speed.
Also since hash_bits is now a define and not set depending on wsize, the check on hash_bits no longer checks wsize, and thus we need to be more pessimistic. (unless we add window_bits to the state struct so we can verify that is actually 15)

I also considered using the DFLTCC calculation, but that thing is calculating around 2x the size of the input buffer. Is that a mistake, or does the hardware compression worst case actually become that big? @iii-i 

Resolves the problem reported in #1039

### DIFF
--- a/compress.c
+++ b/compress.c
@@ -91,10 +91,11 @@ z_size_t Z_EXPORT PREFIX(compressBound)(z_size_t sourceLen) {
         return complen + ZLIB_WRAPLEN;
 
 #ifndef NO_QUICK_STRATEGY
-    /* Quick deflate strategy worse case is 9 bits per literal, rounded to nearest byte,
-       plus the size of block & gzip headers and footers */
-    return sourceLen + ((sourceLen + 13 + 7) >> 3) + 18;
+    return sourceLen                       /* The source size itself */
+      + DEFLATE_QUICK_OVERHEAD(sourceLen)  /* Source encoding overhead, padded to next full byte */
+      + DEFLATE_BLOCK_OVERHEAD             /* Deflate block overhead bytes */
+      + ZLIB_WRAPLEN;                      /* zlib wrapper */
 #else
-    return sourceLen + (sourceLen >> 12) + (sourceLen >> 14) + (sourceLen >> 25) + 13;
+    return sourceLen + (sourceLen >> 4) + 7 + ZLIB_WRAPLEN;
 #endif
 }
--- a/zutil.h
+++ b/zutil.h
@@ -78,7 +78,19 @@ extern z_const char * const PREFIX(z_errmsg)[10]; /* indexed by 2-zlib_error */
 #define ADLER32_INITIAL_VALUE 1 /* initial adler-32 hash value */
 #define CRC32_INITIAL_VALUE   0 /* initial crc-32 hash value */
 
-#define ZLIB_WRAPLEN 6 /* zlib format overhead */
+#define ZLIB_WRAPLEN 6      /* zlib format overhead */
+#define GZIP_WRAPLEN 18     /* gzip format overhead */
+
+#define DEFLATE_HEADER_BITS 3
+#define DEFLATE_EOBS_BITS   15
+#define DEFLATE_PAD_BITS    6
+#define DEFLATE_BLOCK_OVERHEAD ((DEFLATE_HEADER_BITS + DEFLATE_EOBS_BITS + DEFLATE_PAD_BITS) >> 3)
+/* deflate block overhead: 3 bits for block start + 15 bits for block end + padding to nearest byte */
+
+#define DEFLATE_QUICK_LIT_MAX_BITS 9
+#define DEFLATE_QUICK_OVERHEAD(x) ((x * (DEFLATE_QUICK_LIT_MAX_BITS - 8) + 7) >> 3)
+/* deflate_quick worst-case overhead: 9 bits per literal, round up to next byte (+7) */
+
 
         /* target dependencies */
 
--- a/deflate.c
+++ b/deflate.c
@@ -611,11 +611,11 @@ unsigned long Z_EXPORT PREFIX(deflateBound)(PREFIX3(stream) *strm, unsigned long
         wraplen = 0;
         break;
     case 1:                                 /* zlib wrapper */
-        wraplen = 6 + (s->strstart ? 4 : 0);
+        wraplen = ZLIB_WRAPLEN + (s->strstart ? 4 : 0);
         break;
 #ifdef GZIP
     case 2:                                 /* gzip wrapper */
-        wraplen = 18;
+        wraplen = GZIP_WRAPLEN;
         if (s->gzhead != NULL) {            /* user-supplied gzip header */
             unsigned char *str;
             if (s->gzhead->extra != NULL) {
@@ -639,7 +639,7 @@ unsigned long Z_EXPORT PREFIX(deflateBound)(PREFIX3(stream) *strm, unsigned long
         break;
 #endif
     default:                                /* for compiler happiness */
-        wraplen = 6;
+        wraplen = ZLIB_WRAPLEN;
     }
 
     /* if not default parameters, return conservative bound */
@@ -647,8 +647,14 @@ unsigned long Z_EXPORT PREFIX(deflateBound)(PREFIX3(stream) *strm, unsigned long
             s->w_bits != 15 || HASH_BITS < 15)
         return complen + wraplen;
 
-    /* default settings: return tight bound for that case */
-    return sourceLen + (sourceLen >> 12) + (sourceLen >> 14) + (sourceLen >> 25) + 13 - 6 + wraplen;
+#ifndef NO_QUICK_STRATEGY
+    return sourceLen                       /* The source size itself */
+      + DEFLATE_QUICK_OVERHEAD(sourceLen)  /* Source encoding overhead, padded to next full byte */
+      + DEFLATE_BLOCK_OVERHEAD             /* Deflate block overhead bytes */
+      + wraplen;                           /* none, zlib or gzip wrapper */
+#else
+    return sourceLen + (sourceLen >> 4) + 7 + wraplen;
+#endif
 }
 
 /* =========================================================================
--- a/test/switchlevels.c
+++ b/test/switchlevels.c
@@ -68,7 +68,7 @@ static int compress_chunk(PREFIX3(stream) *strm, int level, int size, int last) 
         goto done;
     }
 
-    compsize = 100 + 2 * PREFIX(deflateBound)(strm, size);
+    compsize = PREFIX(deflateBound)(strm, size);
     buf = malloc(size + compsize);
     if (buf == NULL) {
         fprintf(stderr, "Out of memory\n");


nmoinvaz:
I think these changes look okay. I don't remember the original reason behind the `13 + 7`. All I could find was this comment of mine: https://github.com/zlib-ng/zlib-ng/issues/284#issuecomment-594303772.

nmoinvaz (approved):

Dead2:
@nmoinvaz Interesting. Do you agree with my second commit?
I was not quite sure about the 7 bits for rounding however. As that rounds it up to 20 bits, and I would have expected 16 or 24 bits..

nmoinvaz (approved):
Looks like you were able to make sense of it. The `>> 3`  does the bit to byte conversion so I had trouble reading it. `+7` is to round to nearest byte I assume.

nmoinvaz:
You may be correct about the number for rounding up. Maybe I had use the number for rounding up to nearest byte. Perhaps if it was 16 or 24 then we wouldn't need that extra 1 byte added at the end.

Dead2:
I tested the theory, and indeed, changing the 7 bits to 11 bits (totaling 24bytes) removes the need for that extra byte.
I amended the second commit with that change.

nmoinvaz (approved):
Looks great!

compress.c
```diff
@@ -91,10 +91,10 @@ z_size_t Z_EXPORT PREFIX(compressBound)(z_size_t sourceLen) {
         return complen + ZLIB_WRAPLEN;
 
 #ifndef NO_QUICK_STRATEGY
-    /* Quick deflate strategy worse case is 9 bits per literal, rounded to nearest byte,
-       plus the size of block & gzip headers and footers */
-    return sourceLen + ((sourceLen + 13 + 7) >> 3) + 18;
+    /* Quick deflate strategy worst case is 9 bits per literal, rounded to nearest byte,
+       plus the size of block headers, padding and wrappers */
```
nmoinvaz: The comment is a bit confusing.

Dead2: Come to think of it, perhaps the 7 bits was not for the deflate blocks but rather to make sure the bits conversion of sourceLen is rounded up to the next byte. I think that is the case actually, and that should stay as a 7, but that the deflate block header bits also needs to be rounded up to the next byte. I'll amend the commit with that change.

Dead2: And that still passes the test 😄

Dead2: I split up the calculation a little, so it is easier to read, this way it also matches the comment a little better.

iii-i:
@Dead2 according to our hardware folks, in degenerate cases this can happen. Fortunately on e.g. squash-benchmark we haven't seen anything that atrocious.

EDIT: Just from the description, shouldn't `DEFLATE_PADBITS` be 7? In case the last bit of End-of-Block Symbol ends up in a new byte.

Dead2:
@iii-i Thanks for confirming the hardware bounds.

I am not sure it should not be 7, but I don't quite understand what you are saying. Could you explain your thinking?

My thinking is that 3 + 10 = 13, and thus needs another 3 bits to fill up two bytes.
The 3 bits for block header is well understood, but I am a little unsure still where the 10 bits come from. RFC1951 does not seem to explain any such static overhead.

I also think the current non-deflate_quick calculation is a bit pessimistic, but better to err on the safe side.

iii-i:
At the end of each deflate block there is an end-of-block symbol. I guess that 10 comes from the assumption that the software implementation can never generate an EOBS longer than 10 bits? The format allows it to be as long as 15, if I'm reading "3.2.7. Compression with dynamic Huffman codes (BTYPE=10)" correctly. 15 is also used in DFLTCC calculations (`DFLTCC_MAX_EOBS_BITS`).

So, when we add padding to the byte boundary at the end, the header size is actually irrelevant. The way I see the worst case then is: EOBS is 15 bits long and starts at byte X bit 2. Then, the last EOBS bit is at byte X + 2 bit 0. In order to get to the byte boundary, we need to skip the remaining 7 bits of byte X + 2.

Dead2:
Current code for reference:
```
    /* Quick deflate strategy worst case is 9 bits per literal, rounded to nearest byte (+7 bits),
       plus the size of deflate block headers and zlib/gzip wrapper */
    return sourceLen + ((sourceLen + 7) >> 3) + ((DEFLATE_WRAPBITS + DEFLATE_PADBITS) >> 3) + wraplen;
```

@iii-i Hmm. Is that worst case valid, considering we already added 7 bits to the compressed data just in case it ends before a byte boundary?

The way I am reading this is we need 3 bits block header, compressed data, 7 bits padding for compressed data, 15bits of EOBS.
So, 3-bit block header is unpadded, and 15-bit EOBS is unpadded. Would that not mean it needs another 6 bits of padding?

If so, that would look something like:
```
DEFLATE_HDR_BITS = 3
DEFLATE_EOBS_BITS = 15
DEFLATE_PAD_BITS = 6  // These can be added together in the header file, making DEFLATE_BLOCK_OVERHEAD or something

    return sourceLen            // The source size itself
    + ((sourceLen + 7) >> 3)    // Source encoding overhad, padded to next full byte
    + ((DEFLATE_HDR_BITS + DEFLATE_EOBS_BITS + DEFLATE_PAD_BITS) >> 3)    // Deflate block overhead
    + wraplen;                  // zlib or gzip wrapper
```

I might still be wrong, what do you think?
I hate magic numbers, so if we can get this somewhat accurate that would be great. (accurate in documentation and concept, not necessarily as in the total calculation being accurate)

Dead2:
Looks like we might have a too small size in some cases after all, I hadn't noticed the failed CI fuzzer.
```
==25==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x603000000e69 at pc 0x00000052472a bp 0x7ffc583a8fa0 sp 0x7ffc583a8768
WRITE of size 1 at 0x603000000e69 thread T0
SCARINESS: 31 (1-byte-write-heap-buffer-overflow)
    #0 0x524729 in __asan_memcpy /src/llvm-project/compiler-rt/lib/asan/asan_interceptors_memintrinsics.cpp:22:3
    #1 0x569613 in flush_pending /src/zlib-ng/deflate.c:677:5
    #2 0x564a90 in zng_deflate /src/zlib-ng/deflate.c:718:9
    #3 0x55ea74 in test_deflate /src/zlib-ng/test/fuzz/example_small_fuzzer.c:54:15
```
Maybe that is related to the 10 vs 15 bits and padding.

The above suggestion would increase the estimate by 8 bits, and I think that would fix this problem at least.

iii-i:
Ah, you are right, we have already added 7. I just tried to come up with a formula on my own, and that's what I ended up with:

```
#define DEFLATE_HDR_BITS 3
#define DEFLATE_QUICK_LIT_MAX_BITS 9
#define DEFLATE_EOBS_MAX_BITS 15

sourceLen + ((DEFLATE_HDR_BITS + sourceLen * (DEFLATE_QUICK_LIT_MAX_BITS - 8) + DEFLATE_EOBS_MAX_BITS + 7) >> 3) + ZLIB_WRAPLEN
```

The initial `sourceLen` appears to be an alternative for having to multiply it by 9 in parentheses - presumably to reduce the chance of an overflow (I think it can still happen with the right inputs, especially on a 32-bit system).

However, the only difference with the original formula now is that the result is 5 bits larger due to `DEFLATE_EOBS_MAX_BITS`.

The difference with your approach is that you shift and pad compressed literals and deflate block overhead separately. Why does that help?

Dead2:
The reason I split it up was to make it easier to read.
It also makes sure the 7 bits added are dedicated to rounding up the sourcelen and is not affected by the deflate-block overhead bits missing any of its required padding bits, by making it clear that the block part needs to have its own padding handled separately.

Therefore HDRBITS (3) + EOBS (15) = 18 bits + padding (6) = 24 bits = 3bytes for the block overhead

Updated pseudo-code to be much more readable:
```
#define DEFLATE_HEADER_BITS 3
#define DEFLATE_EOBS_BITS 15
#define DEFLATE_PAD_BITS 6
#define DEFLATE_BLOCK_OVERHEAD = (DEFLATE_HEADER_BITS + DEFLATE_EOBS_BITS + DEFLATE_PAD_BITS) >> 3)

#define DEFLATE_QUICK_LIT_MAX_BITS = 9
#define DEFLATE_QUICK_OVERHEAD(x) = ((x * (DEFLATE_QUICK_LIT_MAX_BITS - 8) + 7) >> 3)

    return sourceLen                       // The source size itself
      + DEFLATE_QUICK_OVERHEAD(sourceLen)  // Source encoding overhad, padded to next full byte
      + DEFLATE_BLOCK_OVERHEAD             // Deflate block overhead bytes
      + wraplen;                           // zlib or gzip wrapper
```

Would a 32bit overflow really be feasible though? Wouldn't that require allocating a ~3641MiB or thereabouts input buffer? And then you would need an output buffer as well. Hmm, but the calculation is perhaps done on signed ints.

nmoinvaz:
> The format allows it to be as long as 15, if I'm reading "3.2.7. Compression with dynamic Huffman codes (BTYPE=10)" correctly. 15 is also used in DFLTCC calculations (DFLTCC_MAX_EOBS_BITS).

This calculation is mainly used for deflate quick which doesn't use `BTYPE=10`.

The value 7 comes from `zng_tr_emit_end_block` which is used by deflate_quick.

nmoinvaz:
We may not be able to accurately calculate the exact number of bits/bytes. Back in the day there was only one deflate block that deflate quick used. But now due to other issues we had to have more than one deflate block. See `QUICK_START_BLOCK` and `QUICK_END_BLOCK`.

Dead2:
@nmoinvaz Right, there is that too.
The only thing we need to do is make sure we have a value that is good enough to catch the worst-case, preferably without a huge amount of extra bloat that never gets used.

Suggestions for changes to the last pseudo-code I posted?

nmoinvaz:
I think that looks great. Otherwise overhead is misspelled

Dead2:
I made a little test that tests a bunch of different buffer sizes.
For the deflate_quick mthod; Up to about 32 bytes, the new method calculates a tighter buffer than the old method, at 64 bytes the new one is bigger than the old one, in my tests the old one fails at 128 bytes and above. The new one succeeds all tests up to 4MiB and shows no sign of failing higher than that.

zutil.h
```diff
@@ -78,7 +78,19 @@ extern z_const char * const PREFIX(z_errmsg)[10]; /* indexed by 2-zlib_error */
...
+
+#define DEFLATE_HEADER_BITS 3
+#define DEFLATE_EOBS_BITS   15
+#define DEFLATE_PAD_BITS    6
+#define DEFLATE_BLOCK_OVERHEAD ((DEFLATE_HEADER_BITS + DEFLATE_EOBS_BITS + DEFLATE_PAD_BITS) >> 3)
+/* deflate block overhead: 3 bits for block start + 15 bits for block end + padding to nearest byte */
+
+#define DEFLATE_QUICK_LIT_MAX_BITS 9
+#define DEFLATE_QUICK_OVERHEAD(x) ((x * (DEFLATE_QUICK_LIT_MAX_BITS - 8) + 7) >> 3)
+/* deflate_quick worst-case overhead: 9bits per literal, round up to next byte (+7) */
```
mtl1979: Missing space, should be `9 bits`.

PR #1072, opened 2021-12-14
deflate_medium: avoid emitting a suboptimal literal in the restart case

Dead2:
Import external patch to deflate_medium, improving compression very slightly:
##

deflate_medium: avoid emitting a suboptimal literal in the restart case
When we load new data into the window, we invalidate the next match, in
case the match would improve. In this case, the hash has already been
updated with this data, so when we look for a new match it will point
it back at itself. As a result, a literal is generated even when a
better match is available.

This avoids that by catching this case and ensuring we're looking at the
past.

### DIFF
--- a/deflate_medium.c
+++ b/deflate_medium.c
@@ -203,6 +203,9 @@ Z_INTERNAL block_state deflate_medium(deflate_state *s, int flush) {
                 hash_head = functable.quick_insert_string(s, s->strstart);
             }
 
+            if (hash_head && hash_head == s->strstart)
+                hash_head--;
+
             current_match.strstart = (uint16_t)s->strstart;
             current_match.orgstart = current_match.strstart;
 
@@ -237,6 +240,9 @@ Z_INTERNAL block_state deflate_medium(deflate_state *s, int flush) {
         if (LIKELY(!early_exit && s->lookahead > MIN_LOOKAHEAD && (uint32_t)(current_match.strstart + current_match.match_length) < (s->window_size - MIN_LOOKAHEAD))) {
             s->strstart = current_match.strstart + current_match.match_length;
             hash_head = functable.quick_insert_string(s, s->strstart);
+
+            if (hash_head && hash_head == s->strstart)
+                hash_head--;
 
             next_match.strstart = (uint16_t)s->strstart;
             next_match.orgstart = next_match.strstart;


nmoinvaz:
Any benchmarks for compression size comparison?

nmoinvaz:
For some reason I thought we had merged all the changes from Intel's fork.

mtl1979:
Maybe I'm tired, but assuming we need to rewind just one byte seems wrong... Hash sliding is what I have worked on a lot and it seems that instead of invalidating the next match unconditionally we should use "saturated" subtraction for the match index so only match that goes too far will get invalidated. This obviously means we need to know what the match distance was before the window slide and how many bytes the window was slid.

Dead2:
## develop
```
 129486    1376      32  130894   1ff4e libz-ng.so.2

 Tool: minigzip   Levels: 1-9

 Level   Comp   Comptime min/avg/max/stddev   Compressed size
 1     54.164%      1.384/1.394/1.400/0.005        85,234,591
 2     43.876%      1.502/1.518/1.523/0.004        41,427,218
 3     42.383%      1.472/1.481/1.488/0.004        33,347,341
 4     41.669%      1.377/1.386/1.391/0.004        26,228,657
 5     41.211%      1.592/1.601/1.606/0.004        25,940,617
 6     41.038%      1.446/1.453/1.457/0.003        19,373,674
 7     40.781%      1.309/1.314/1.317/0.002        12,834,980
 8     40.706%      1.675/1.680/1.683/0.002        12,811,166
 9     40.423%      1.938/1.944/1.947/0.002        12,722,161

 avg1  42.917%                        1.530
 tot                                413.094       269,920,405
```

## medium fix 5420eb959b365fb34781b987a8187cda76db5ebe
```
 129550    1376      32  130958   1ff8e libz-ng.so.2

 Tool: minigzip   Levels: 1-9

 Level   Comp   Comptime min/avg/max/stddev   Compressed size
 1     54.164%      1.381/1.394/1.398/0.004        85,234,591
 2     43.876%      1.476/1.489/1.494/0.004        41,427,218
 3     42.383%      1.530/1.539/1.545/0.004        33,347,341
 4     41.669%      1.421/1.433/1.438/0.004        26,228,657
 5     41.211%      1.616/1.623/1.628/0.003        25,940,615
 6     41.038%      1.460/1.466/1.470/0.002        19,373,673
 7     40.781%      1.312/1.316/1.318/0.001        12,834,980
 8     40.706%      1.678/1.682/1.685/0.002        12,811,166
 9     40.423%      1.939/1.945/1.947/0.002        12,722,161

 avg1  42.917%                        1.543
 tot                                416.573       269,920,402
```

nmoinvaz:
Compression time is increased on all deflate medium levels. Saves 3 bytes. It would be good to run the same test on a different data set.

Dead2:
I am considering dropping this, I did several more tests and I don't really think the performance drop justifies the compression improvement. In a few cases the compression is even a byte worse instead of better, probably due to missing a better opportunity due to the rewind causing an overlapping opportunity.

Perhaps @mtl1979 has a better implementation idea?

mtl1979:
@Dead2 When I initially read about the logic, my concern was that after the slide, the first two matches are backwards... This is because after the slide, it uses insert_string() and inside that we check that the current and previous values are not identical, but not if those are in wrong order. Essentially instead of going back one byte, we would need to replace the "current" match with the "previous" match in chain as that is the closest usable match after the slide. IF that previous match is 0, that means there wasn't any match before the slide or it went outside the window.

The root cause is that sliding the window is faster than clearing all the hashes and reinserting all matches from start of new window up to the current position we are comparing.

PR #1076, opened 2021-12-17
VPCLMULQDQ implementation of CRC32 folding

nmoinvaz:
Based on https://github.com/jtkukunas/zlib/pull/28.


nmoinvaz:
I don't have one of those CPUs where it has benefit, @KungFuJesus maybe you can run some benchmarks on this PR?

KungFuJesus:
I suspect the xeon silvers I have access to may be a tiny bit too old but I can check come Monday. I could also try adapting this to be 256 bits wide, instead. That may have advantages on tiger lake, even mobile SKUs.

I might try to see if I can get an alder lake setup at work in some capacity, though it may be tougher to get avx512 enabled on a laptop SKU.

nmoinvaz:
11th Gen Intel Core i7 1185G7 @ 3.00Ghz Tiger Lake-U
### DEVELOP https://github.com/zlib-ng/zlib-ng/commit/5b0ffa63f22d4870ae68de38c34f7ec437e7edb1
```

 Tool: ../zlib-ng/build-develop/Release/minigzip.exe Levels: 0-9

 0    100.008%      0.019/0.021/0.023/0.001        0.100/0.121/0.138/0.012       15,737,543
 1     54.166%      0.103/0.118/0.130/0.008        0.052/0.059/0.066/0.004        8,523,732
 2     43.871%      0.163/0.188/0.212/0.015        0.052/0.060/0.068/0.004        6,903,609
 3     42.387%      0.216/0.256/0.293/0.024        0.053/0.059/0.067/0.004        6,670,099
 4     41.647%      0.241/0.286/0.332/0.026        0.051/0.057/0.065/0.004        6,553,723
 5     41.216%      0.256/0.307/0.351/0.025        0.052/0.057/0.064/0.004        6,485,938
 6     41.037%      0.304/0.349/0.392/0.030        0.049/0.057/0.064/0.004        6,457,776
 7     40.778%      0.370/0.440/0.506/0.040        0.049/0.056/0.064/0.004        6,416,919
 8     40.704%      0.470/0.547/0.636/0.051        0.122/0.150/0.171/0.012        6,405,244
 9     40.409%      0.583/0.650/0.718/0.043        0.048/0.055/0.061/0.004        6,358,951

 avg1  48.622%                        0.316                          0.073
 avg2  54.025%                        0.351                          0.081
 tot                                 94.816                         21.939       76,513,534
```
### PR https://github.com/zlib-ng/zlib-ng/pull/1076/commits/7a1d074a93be726ca5280bfb233c79e1f354e5a9
`-DWITH_VPCLMULQDQ=ON`
```
 Tool: ../zlib-ng/build/Release/minigzip.exe Levels: 0-9

 0    100.008%      0.017/0.017/0.017/0.000        0.080/0.088/0.097/0.006       15,737,543
 1     54.166%      0.091/0.092/0.092/0.000        0.046/0.047/0.047/0.000        8,523,732
 2     43.871%      0.145/0.146/0.147/0.001        0.047/0.048/0.049/0.000        6,903,609
 3     42.387%      0.194/0.195/0.197/0.001        0.047/0.047/0.048/0.000        6,670,099
 4     41.647%      0.218/0.219/0.220/0.001        0.044/0.046/0.047/0.000        6,553,723
 5     41.216%      0.230/0.231/0.232/0.001        0.045/0.046/0.046/0.000        6,485,938
 6     41.037%      0.268/0.270/0.271/0.001        0.044/0.046/0.046/0.000        6,457,776
 7     40.778%      0.335/0.337/0.338/0.001        0.045/0.046/0.046/0.000        6,416,919
 8     40.704%      0.430/0.432/0.433/0.001        0.108/0.118/0.124/0.005        6,405,244
 9     40.409%      0.507/0.509/0.514/0.001        0.044/0.045/0.046/0.000        6,358,951

 avg1  48.622%                        0.245                          0.058
 avg2  54.025%                        0.272                          0.064
 tot                                 73.407                         17.327       76,513,534
```
So here I performed another test on my laptop with the power plugged in. I got some good results, ~20% faster compression and decompression.

nmoinvaz:
@KungFuJesus looks like this code uses AVX512 so we will need some of the same AVX512 detection code you already have in your PR. I agree 256 wide loads would be a good improvement.

arch/x86/crc32_fold_vpclmulqdq.c
```diff
@@ -0,0 +1,108 @@
...
+    zmm_crc0 = _mm512_setzero_si512();
+    zmm_t0 = _mm512_loadu_si512((__m512i *)src);
+    zmm_crc1 = _mm512_loadu_si512((__m512i *)src + 1);
+    zmm_crc2 = _mm512_loadu_si512((__m512i *)src + 2);
+    zmm_crc3 = _mm512_loadu_si512((__m512i *)src + 3);
+
+    /* already have intermediate CRC in xmm registers
+        * fold4 with 4 xmm_crc to get zmm_crc0
+    */
+    zmm_crc0 = _mm512_inserti32x4(zmm_crc0, *xmm_crc0, 0);
```
KungFuJesus, line 32: These inserts are likely to be a tad costly.  If you just had a wider load coming into the function, that might actually fare better (though, it won't easily plug into the existing code as well, I don't think).

arch/x86/crc32_fold_pclmulqdq.c
```diff
@@ -274,6 +279,16 @@ Z_INTERNAL void crc32_fold_copy_pclmulqdq(crc32_fold *crc, uint8_t *dst, const u
...
+        size_t n = fold_16_vpclmulqdq(&xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3, dst, src, len);
+
+        len -= n;
+        src += n;
+        dst += n;
+    }
+#endif
+
     while (len >= 64) {
         crc32_fold_load((__m128i *)src, &xmm_t0, &xmm_t1, &xmm_t2, &xmm_t3);
```
KungFuJesus, line 295: The lines below this one (evidently GH won't let me leave a comment far from your modifications) are a missed opportunity for a single 512 bit wide xor.  A lot of this code can have its loads and stores quartered.  On a relative scale, loads and stores are costly.  They have an even higher latency cost on newer versions of the microarchitecture.

nmoinvaz: I don't know enough to be able to make those changes.

nmoinvaz:
I have rebased and also enabled it by default since I was seeing good results.

I'm not sure if we also need AVX512 compiler flags or if the VPCLMULQDQ flag covers it.

arch/x86/crc32_fold_pclmulqdq.c
nmoinvaz: I think I see what you mean with the xors. It might be an interesting idea to have an AVX512 version of this function in the future.

arch/x86/crc32_fold_vpclmulqdq.c
nmoinvaz: Yeah I don't think it would plug into the existing code well either.

nmoinvaz:
Updated commit message.

nmoinvaz:
This PR is waiting on #1065 to be merged for AVX512 flag detection.

nmoinvaz:
Rebased and added the necesssary checks for AVX512.

Dead2 (approved):
LGTM

kangshan0910:
> 11th Gen Intel Core i7 1185G7 @ 3.00Ghz Tiger Lake-U
> ### DEVELOP [5b0ffa6](https://github.com/zlib-ng/zlib-ng/commit/5b0ffa63f22d4870ae68de38c34f7ec437e7edb1)
> 
> ```
> 
>  Tool: ../zlib-ng/build-develop/Release/minigzip.exe Levels: 0-9
>  Runs: 70         Trim worst: 40
> 
>  Level   Comp   Comptime min/avg/max/stddev  Decomptime min/avg/max/stddev  Compressed size
>  0    100.008%      0.019/0.021/0.023/0.001        0.100/0.121/0.138/0.012       15,737,543
>  1     54.166%      0.103/0.118/0.130/0.008        0.052/0.059/0.066/0.004        8,523,732
>  2     43.871%      0.163/0.188/0.212/0.015        0.052/0.060/0.068/0.004        6,903,609
>  3     42.387%      0.216/0.256/0.293/0.024        0.053/0.059/0.067/0.004        6,670,099
>  4     41.647%      0.241/0.286/0.332/0.026        0.051/0.057/0.065/0.004        6,553,723
>  5     41.216%      0.256/0.307/0.351/0.025        0.052/0.057/0.064/0.004        6,485,938
>  6     41.037%      0.304/0.349/0.392/0.030        0.049/0.057/0.064/0.004        6,457,776
>  7     40.778%      0.370/0.440/0.506/0.040        0.049/0.056/0.064/0.004        6,416,919
>  8     40.704%      0.470/0.547/0.636/0.051        0.122/0.150/0.171/0.012        6,405,244
>  9     40.409%      0.583/0.650/0.718/0.043        0.048/0.055/0.061/0.004        6,358,951
> 
>  avg1  48.622%                        0.316                          0.073
>  avg2  54.025%                        0.351                          0.081
>  tot                                 94.816                         21.939       76,513,534
> ```
> 
> ### PR [7a1d074](https://github.com/zlib-ng/zlib-ng/commit/7a1d074a93be726ca5280bfb233c79e1f354e5a9)
> 
> `-DWITH_VPCLMULQDQ=ON`
> 
> ```
>  Tool: ../zlib-ng/build/Release/minigzip.exe Levels: 0-9
>  Runs: 70         Trim worst: 40
> 
>  Level   Comp   Comptime min/avg/max/stddev  Decomptime min/avg/max/stddev  Compressed size
>  0    100.008%      0.017/0.017/0.017/0.000        0.080/0.088/0.097/0.006       15,737,543
>  1     54.166%      0.091/0.092/0.092/0.000        0.046/0.047/0.047/0.000        8,523,732
>  2     43.871%      0.145/0.146/0.147/0.001        0.047/0.048/0.049/0.000        6,903,609
>  3     42.387%      0.194/0.195/0.197/0.001        0.047/0.047/0.048/0.000        6,670,099
>  4     41.647%      0.218/0.219/0.220/0.001        0.044/0.046/0.047/0.000        6,553,723
>  5     41.216%      0.230/0.231/0.232/0.001        0.045/0.046/0.046/0.000        6,485,938
>  6     41.037%      0.268/0.270/0.271/0.001        0.044/0.046/0.046/0.000        6,457,776
>  7     40.778%      0.335/0.337/0.338/0.001        0.045/0.046/0.046/0.000        6,416,919
>  8     40.704%      0.430/0.432/0.433/0.001        0.108/0.118/0.124/0.005        6,405,244
>  9     40.409%      0.507/0.509/0.514/0.001        0.044/0.045/0.046/0.000        6,358,951
> 
>  avg1  48.622%                        0.245                          0.058
>  avg2  54.025%                        0.272                          0.064
>  tot                                 73.407                         17.327       76,513,534
> ```
> 
> So here I performed another test on my laptop with the power plugged in. I got some good results, ~20% faster compression and decompression.

Hi @nmoinvaz,
What dataset did you use? I tried the patch on the Ice Lake (Intel(R) Xeon(R) Platinum 8380 CPU @ 2.30GHz), but I cannot produce the benifit. I used the [deflatebench  ](https://github.com/zlib-ng/deflatebench.git) and tried both [silesia-small](https://mirror.circlestorm.org/silesia-small.tar) and [silesia](https://mirror.circlestorm.org/silesia.tar) datasets through the command "sudo ./deflatebench.py -s". After apply the patch, the performance will be 5% worse than before. I have no idea if I use the proper test or test dataset.
The original data:
Tool: minigzip   Levels: 0-9
 Runs: 15         Trim worst: 5

 0    100.008%      0.000/0.000/0.001/0.001        0.001/0.002/0.003/0.001       15,737,543
 1     54.166%      0.096/0.098/0.099/0.001        0.041/0.042/0.044/0.001        8,523,732
 2     43.871%      0.168/0.170/0.171/0.001        0.039/0.043/0.046/0.002        6,903,609
 3     42.387%      0.194/0.196/0.197/0.001        0.040/0.042/0.045/0.002        6,670,099
 4     41.647%      0.219/0.224/0.226/0.002        0.038/0.041/0.042/0.001        6,553,723
 5     41.216%      0.245/0.248/0.249/0.001        0.038/0.040/0.041/0.001        6,485,938
 6     41.037%      0.303/0.304/0.305/0.001        0.040/0.040/0.042/0.001        6,457,776
 7     40.778%      0.421/0.423/0.425/0.001        0.037/0.041/0.043/0.002        6,416,919
 8     40.704%      0.553/0.554/0.556/0.001        0.038/0.040/0.042/0.001        6,405,244
 9     40.409%      0.646/0.649/0.651/0.001        0.039/0.041/0.041/0.001        6,358,951

 avg1  48.622%                        0.287                          0.037
 avg2  54.025%                        0.318                          0.041
 tot                                 28.656                          3.714       76,513,534

**After apply the patch:**
Tool: minigzip   Levels: 0-9
 Runs: 15         Trim worst: 5

 0    100.008%      0.000/0.000/0.001/0.001        0.000/0.002/0.003/0.001       15,737,543
 1     54.166%      0.098/0.101/0.104/0.002        0.042/0.044/0.045/0.001        8,523,732
 2     43.871%      0.177/0.179/0.181/0.001        0.040/0.045/0.047/0.002        6,903,609
 3     42.387%      0.203/0.205/0.207/0.002        0.041/0.044/0.046/0.001        6,670,099
 4     41.647%      0.235/0.236/0.237/0.001        0.036/0.041/0.043/0.002        6,553,723
 5     41.216%      0.257/0.262/0.264/0.002        0.038/0.042/0.043/0.001        6,485,938
 6     41.037%      0.319/0.321/0.323/0.002        0.040/0.042/0.044/0.001        6,457,776
 7     40.778%      0.444/0.447/0.449/0.002        0.038/0.042/0.043/0.002        6,416,919
 8     40.704%      0.579/0.582/0.584/0.002        0.041/0.043/0.045/0.001        6,405,244
 9     40.409%      0.682/0.686/0.688/0.002        0.036/0.040/0.043/0.002        6,358,951

 avg1  48.622%                        0.302                          0.038
 avg2  54.025%                        0.336                          0.043
 tot                                 30.197                          3.841       76,513,534

KungFuJesus:
That's a 40 core CPU, perhaps the clock penalty is very high? Can you do perf record maybe to get a rough idea for where the CPU cycles are being spent?

Also, can you fix your clock rate to maybe the maximum turbo for avx512 so we can see if what we're seeing is a result of the clock penalty? Supposedly on newer chips it's not supposed to be as drastic, but those multi core monsters are already constrained by clock speed often.

nmoinvaz:
@KungFuJesus I enabled it for all AVX512 capably CPU (not used `well_suited`) variable because I saw a benefit.

@kangshan0910 I used silesia-small.tar.

It could be that your AVX512 CPU is not what we consider well-suited but if it has 40 CPU cores it is hard to believe that to be the case. The information @KungFuJesus asked for will help in determining that.

KungFuJesus:
Yeah if you saw an improvement with a U series ice lake I'd suspect it'd do well on more than just the few CPUs I found an advantage for 512 bit wide Adler checksums.

I do wonder if sub optimal instruction scheduling could be part of it, though. The other thing to try to see if that's the case is messing with the native optimization setting at compile time.

nmoinvaz:
@kangshan0910 when I run `deflatebench` I usually do 120 runs and discard 40. You are only doing 20 runs and discarding 5. I recommend changing your `deflatebench` config settings and doing another test.

KungFuJesus:
I'll see about procuring another ice lake laptop tomorrow to see how gcc is doing with this.

nmoinvaz:
By the way, I am using Microsoft Surface Laptop 4 and compiling with MSVC 2022 if it makes a difference.

nmoinvaz:
Another thing I forgot to mention is when I do my testing with deflatebench I compile minigzip with -DBUILD_WITH_LIBS=OFF so it doesn't try to build with shared library which can be slower.

KungFuJesus:
So here's without vpclmuldq:
```
2022-01-11T09:40:24-05:00
Running ./crc32_benchmark
Run on (16 X 4250.73 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 1280 KiB (x8)
  L3 Unified 24576 KiB (x1)
Load Average: 1.26, 1.38, 1.28
Benchmark                      Time             CPU   Iterations UserCounters...
crc32_generic_bench        70055 ns        70054 ns        19569 CYCLES=316.606k
crc32_little_bench         70407 ns        70356 ns        61659 CYCLES=317.535k
crc32_pclmulqdq_bench       6520 ns         6514 ns       100000 CYCLES=28.2072k
```

Here's with:

```
2022-01-11T09:43:31-05:00
Running ./crc32_benchmark
Run on (16 X 2077.02 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 1280 KiB (x8)
  L3 Unified 24576 KiB (x1)
Load Average: 1.96, 1.36, 1.27
Benchmark                      Time             CPU   Iterations UserCounters...
crc32_generic_bench        71277 ns        71224 ns        19441 CYCLES=315.257k
crc32_little_bench         66980 ns        66889 ns        57817 CYCLES=300.389k
crc32_pclmulqdq_bench       6395 ns         6395 ns       100000 CYCLES=28.5356k
```

So the runtime is better by a smaller margin, though the cycles count is higher :thinking:

KungFuJesus:
Well this is odd, perf's annotate is not showing any 512 bit wide carryless multiplies...I'm going to need to double check the benchmark code to make sure it's calling the fast fold routine (which should _probably_ be inlined?).  @nmoinvaz I just took your commit and applied it as a patch on top of develop, is it missing anything?

mtl1979:
Clock speed drops to half when using vpclmuldq, so the benchmark results are not comparable...

KungFuJesus:
Ahhh, this might be a problem:

```
#ifdef X86_VPCLMULQDQ_CRC
    if (x86_cpu_has_vpclmulqdq && x86_cpu_has_avx512 && (len >= 256)) {
        size_t n = fold_16_vpclmulqdq(&xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3, dst, src, len);

        len -= n;
        src += n;
        dst += n;
    }
#endif
```

Are we only ever compiling this is in if the preprocessor sees support?  That has other issues (generic packages wouldn't be able to call this, it _has_ to be compiled in).  

Err, disregard that, I just realized that you have runtime checks in there and the ifdef is just whether or not support is compiled in.

KungFuJesus:
> Clock speed drops to half when using vpclmuldq, so the benchmark results are not comparable...

The cycles count disregards changes in clock speed and is there to be more reliable for these cases.  It seems to be a much better metric when comparing variants.  Though in this case, the variant didn't actually change it would seem.

KungFuJesus:
I don't think that macro is being set by the time things are being compiled with that source file...I definitely don't see that function being called.
```
has avx512 = 65536, has vpclmuldq = 1024
2022-01-11T09:56:26-05:00
Running ./crc32_benchmark
Run on (16 X 2242.18 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 1280 KiB (x8)
  L3 Unified 24576 KiB (x1)
Load Average: 1.25, 0.97, 1.03
----------------------------------------------------------------
Benchmark                      Time             CPU   Iterations
----------------------------------------------------------------
crc32_generic_bench        70661 ns        70154 ns        19420
crc32_little_bench         71323 ns        71173 ns        60730
crc32_pclmulqdq_bench       6482 ns         6482 ns       100000
```

CPU flags are legit, but a printf in that if block is not firing.

mtl1979:
> > Clock speed drops to half when using vpclmuldq, so the benchmark results are not comparable...
> 
> The cycles count disregards changes in clock speed and is there to be more reliable for these cases. It seems to be a much better metric when comparing variants. Though in this case, the variant didn't actually change it would seem.

Exactly... Some processors can't run on full speed when using all cores or with all vector instruction sets enabled.

KungFuJesus:
The measured clock cycles are supposed to be invariant of clock speed.  It uses PMUs on the CPU to count them:
http://perfmon2.sourceforge.net/manv3/pfm_get_cycle_event.html

KungFuJesus:
```
                if(WITH_VPCLMULQDQ AND WITH_AVX512)
                    check_vpclmulqdq_intrinsics()
                    if(HAVE_VPCLMULQDQ_INTRIN AND HAVE_AVX512_INTRIN)
                        add_definitions(-DX86_VPCLMULQDQ_CRC)
                        set(VPCLMULQDQ_SRCS ${ARCHDIR}/crc32_fold_vpclmulqdq.c)
                        add_feature_info(VPCLMUL_CRC 1 "Support CRC hash generation using VPCLMULQDQ, using \"${VPCLMULFLAG} ${AVX512FLAG}\"")
                        list(APPEND ZLIB_ARCH_SRCS ${VPCLMULQDQ_SRCS})
                        set_property(SOURCE ${VPCLMULQDQ_SRCS} PROPERTY COMPILE_FLAGS "${VPCLMULFLAG} ${AVX512FLAG} ${NOLTOFLAG}")
                    else()
                        set(WITH_VPCLMULQDQ OFF)
                    endif()
                else()
                    set(WITH_VPCLMULQDQ OFF)
                endif()
```

Here's the problem.  You need to set this preprocessor definition on both crc32_fold_* source files, not just that one.  @kangshan0910 any perceived differences you were measuring were likely noise, as this doesn't _seem_ to be running.

KungFuJesus:
@nmoinvaz Ahh, it appears you're still manually adding sources rather than statically linking to the zlib library like I had been for the benchmarks.  That explains a lot.  Here's the result with VPCLMULQDQ:

```
----------------------------------------------------------------
Benchmark                      Time             CPU   Iterations
----------------------------------------------------------------
crc32_generic_bench        70076 ns        70076 ns        19617
crc32_little_bench         70233 ns        70234 ns        61724
crc32_pclmulqdq_bench       7078 ns         7078 ns       180626
```

What's even stranger is if I inject a printf in that block, somehow it only needs to do 100,000 iterations to get consistent timing, and the time drops down to 5000ns.  Google benchmark is perplexing.

KungFuJesus:
I figured out what's going on with Google's benchmark here and the short answer is we're kind of doing it wrong.  @nmoinvaz see my comment on your commit in your personal repo.  For large enough input, anyway, vpclmul is faster when compared in a vacuum.  Here's without, with 8192*8192 input buffers:
```
Running ./crc32_benchmark
Run on (16 X 1629.31 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 1280 KiB (x8)
  L3 Unified 24576 KiB (x1)
Load Average: 0.46, 0.59, 0.80
----------------------------------------------------------------
Benchmark                      Time             CPU   Iterations
----------------------------------------------------------------
crc32_generic_bench    479659134 ns    479660046 ns            2
crc32_little_bench     152819382 ns    152817929 ns            5
crc32_pclmulqdq_bench   18438451 ns     18438469 ns           38
```

Here's with VPCLMULQDQ:
```
Running ./crc32_benchmark
Run on (16 X 3611.27 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 1280 KiB (x8)
  L3 Unified 24576 KiB (x1)
Load Average: 0.41, 0.64, 0.84
----------------------------------------------------------------
Benchmark                      Time             CPU   Iterations
----------------------------------------------------------------
crc32_generic_bench    481337395 ns    480897729 ns            2
crc32_little_bench     153084644 ns    152935167 ns            5
crc32_pclmulqdq_bench   16730053 ns     16702802 ns           42
```

We need to modify the benchmark to feed it varying power of 2 sized inputs instead of the sequential climbs that we're doing now in order to make these comparisons relevant.  That way, we can get a nice histogram for how well the function scales with regard to input size and have a much better comparison metric.  

I only compared this under the microbenchmarking context and not while running deflatebench, which is a much larger battery of tests and is probably more indicative of real world impacts.  I'd be sure that the fold_16 version of the function is actually being called with perf record (it probably is, as this benchmark didn't call the function from the library but installed compiled in the sources separately).  That preprocessor macro definition I'd be sure is being applied to all source files, even ones that were added to the cmake project later (I think it might be, but double check). 

I'm probably going to need to table this for today, I've got some other things to take care of.

nmoinvaz:
The reason I have the benchmarks sending ever increasing buffer lengths is to better test real world conditions (or so I thought). I am okay if we change the benchmarks to be more consistent. 

> We need to modify the benchmark to feed it varying power of 2 sized inputs instead of the sequential climbs that we're doing now in order to make these comparisons relevant. 

I can see might also have a more consistent result. 

One thing we might also want to consider is making a higher treshold for when the VPCLMULQDQ gets called. Right now in order for it to take that branch in `crc32`, `len` has to be greater than 256. But we can make that a higher value like 2048, so that the cost of any context switching is justified.

KungFuJesus:
> One thing we might also want to consider is making a higher treshold for when the VPCLMULQDQ gets called. Right now in order for it to take that branch in crc32, len has to be greater than 256. But we can make that a higher value like 2048, so that the cost of any context switching is justified.

I had considered that as well - there might be a decent cutoff that can be found where the trade-off has larger gains.  The problem with google benchmark's method combined with how we're doing it is that the control flow can be cut off arbitrarily after a time has expired, so you're really comparing apples to oranges.  That's why adding a printf would make things arbitrarily faster - it took longer so it couldn't climb to the higher input sizes before it hit the minimum time before libbenchmark thought things looked consistent enough.  The benchmarking functions themselves need fixed termination criteria as that's what the keep running loop assumes.

KungFuJesus:
@kangshan0910 When the benchmark PR lands could you run that with --benchmark_filter=crc and paste us the output?  It might be good if you could capture a profile as well with perf record, with and without vpclmulqdq support compiled in.  

I suspect the performance loss you're seeing might be outside of the crc routines, perhaps subject to the normal variance that happens with function alignment at the link stage.  Unfortunately we can't benchmark with and without vpclmulqdq in the same binary as the pclmulqdq routine calls the 512 bit wide version inside of the 128 bit version for the vast majority of it.

PR #1084, opened 2021-12-25
Fixed duplicate symbols in infcover when not building shared libs.

nmoinvaz:
CMake commands to reproduce:
```
cmake .. -DBUILD_SHARED_LIBS=OFF -T ClangCl -A Win32
cmake --build . --verbose --config Debug
```
Error output:
```
lld-link : error : duplicate symbol: _zng_inflate_table [C:\Users\Nathan\Source\zlib-ng\build-cl\infcover.vcxproj]
  >>> defined at C:\Users\Nathan\Source\zlib-ng\inftrees.c:33
  >>>            infcover.dir\Debug\inftrees.obj
  >>> defined at zlibstatic-ngd.lib(inftrees.obj)

lld-link : error : duplicate symbol: _zng_inflate_copyright [C:\Users\Nathan\Source\zlib-ng\build-cl\infcover.vcxproj
]
  >>> defined at infcover.dir\Debug\inftrees.obj
  >>> defined at zlibstatic-ngd.lib(inftrees.obj)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1158,7 +1158,9 @@ if(ZLIB_ENABLE_TESTS)
     target_link_libraries(switchlevels zlib)
 
     add_simple_test_executable(infcover)
-    target_sources(infcover PRIVATE inftrees.c)
+    if(NOT BUILD_SHARED_LIBS)
+        target_sources(infcover PRIVATE inftrees.c)
+    endif()
 
     add_executable(makefixed tools/makefixed.c inftrees.c)
     target_include_directories(makefixed PUBLIC


mtl1979:
As far as I know, this is linker bug... 

Duplicate symbols are expected when using static libraries as same library or object file might be linked in more than once due to symbols needing to be in specific order to resolve correctly while symbol tables are being created. However, order of the duplicate symbols is significant as only first of the duplicates is preserved when duplicates are eliminated.

I have nothing against having workarounds that are not likely break compiling with non-buggy compilers and/or linkers.

nmoinvaz:
I have never heard of duplicate symbols being okay. It may be a linker bug as you say. Nice benefit is in some case the same source file doesn't need to be compiled twice. 

infcover links in zng_inflate_tables from inftrees.c when dynamically linked - because it is not exported by library. But when statically linked it ends up linking in twice - once from zlibstatic.lib and once from inftrees.c.

You can run the reproduction steps and see.

mtl1979:
This has been discussed before... We have some internal symbols that should be visible only to other binaries hosted in same repository. Due to `ar` not supporting hiding "internal" symbols from symbol table, we would need to switch to alternate archiving tool.

nmoinvaz:
I remember we talked about using alternative linker due to ` -fno-semantic-interposition` - that is mentioned in the [Wiki](https://github.com/zlib-ng/zlib-ng/wiki/Common-build-problems).

nmoinvaz:
I have split the first commit into its own PR.

mtl1979:
I think it was good idea to split this as two separate pull requests... The patch is now a lot easier to read through...

Dead2 (approved):
LGTM

PR #1087, opened 2021-12-28
Don't build DLL sources if BUILD_SHARED_LIBS=OFF.

nmoinvaz:
In this commit, we only add `ZLIB_DLL_SRCS` if `BUILD_SHARED_LIBS=ON` or not defined.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -920,43 +920,45 @@ set(ZLIB_GZFILE_SRCS
     gzwrite.c
 )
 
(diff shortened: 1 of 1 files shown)


mtl1979:
I see this logic still breaking when `BUILD_SHARED_LIBS` is completely undefined... `ZLIB_DLL_SRCS` should not be added to `ZLIB_ALL_SRCS` when both static and shared versions are built.

nmoinvaz:
@mtl1979 good catch. Should be resolved now.

Dead2 (approved):
LGTM

PR #1090, opened 2021-12-30
Cleanup sse4flag in configure

nmoinvaz:
I used sse4flag instead of sse42flag variable to make it similar to CMake and also because sse4flag was written to log at the end of configure script but not sse42flag which indicates perhaps it was added later.


KungFuJesus:
So I actually half a local change on an old Yorkfield machine that is only sse 4.1 capable.  Unfortunately this particular machine needs -msse4.1 and will not work with -msse4.2 (I mean it might but it gives the compiler free reign to generate 4.2 instructions).

I'll put in a PR for this, but it might make my life easier if 4.1 and 4.2 are separately checked for.  I'm relying on the sum of absolute difference instruction, and it's actually a pretty decent gain (like another 13% or so on the crappy Q9650).

nmoinvaz:
We only use it for detecting crc32 instruction.

KungFuJesus:
> We only use it for detecting crc32 instruction.

At the moment, but I have this...
```
           __m128i v_sad_sum1 = _mm_sad_epu8(vbuf, zero); 
           vs1 = _mm_add_epi32(v_sad_sum1, vs1);
```

in a PR I'll probably put in tomorrow.  _mm_sad_epu8 is an SSE 4.1 feature, which Yorkfield has but notably lacks SSE 4.2.

nmoinvaz:
Closing due to #1094.

PR #1093, opened 2021-12-31
Don't assign C versions of functions if we don't use them.

nmoinvaz:
These two commits save an extra 236 bytes on MSVC. First commit saves 168, second one saves 68.

### DIFF
--- a/functable.c
+++ b/functable.c
@@ -315,31 +315,31 @@ Z_INTERNAL uint32_t adler32_stub(uint32_t adler, const unsigned char *buf, size_
 }
 
 Z_INTERNAL uint32_t crc32_fold_reset_stub(crc32_fold *crc) {
-    functable.crc32_fold_reset = crc32_fold_reset_c;
+    functable.crc32_fold_reset = &crc32_fold_reset_c;
     cpu_check_features();
 #ifdef X86_PCLMULQDQ_CRC
     if (x86_cpu_has_pclmulqdq)
-        functable.crc32_fold_reset = crc32_fold_reset_pclmulqdq;
+        functable.crc32_fold_reset = &crc32_fold_reset_pclmulqdq;
 #endif
     return functable.crc32_fold_reset(crc);
 }
 
 Z_INTERNAL void crc32_fold_copy_stub(crc32_fold *crc, uint8_t *dst, const uint8_t *src, size_t len) {
-    functable.crc32_fold_copy = crc32_fold_copy_c;
+    functable.crc32_fold_copy = &crc32_fold_copy_c;
     cpu_check_features();
 #ifdef X86_PCLMULQDQ_CRC
     if (x86_cpu_has_pclmulqdq)
-        functable.crc32_fold_copy = crc32_fold_copy_pclmulqdq;
+        functable.crc32_fold_copy = &crc32_fold_copy_pclmulqdq;
 #endif
     functable.crc32_fold_copy(crc, dst, src, len);
 }
 
 Z_INTERNAL uint32_t crc32_fold_final_stub(crc32_fold *crc) {
-    functable.crc32_fold_final = crc32_fold_final_c;
+    functable.crc32_fold_final = &crc32_fold_final_c;
     cpu_check_features();
 #ifdef X86_PCLMULQDQ_CRC
     if (x86_cpu_has_pclmulqdq)
-        functable.crc32_fold_final = crc32_fold_final_pclmulqdq;
+        functable.crc32_fold_final = &crc32_fold_final_pclmulqdq;
 #endif
     return functable.crc32_fold_final(crc);
 }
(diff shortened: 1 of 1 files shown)


mtl1979:
This was done initially to avoid running capability tests twice when someone tries to use zlib-ng from multiple threads. Instead of using mutexes which are not available on C code, it returns the default implementation.

nmoinvaz:
I think in these cases the default/fallback implementation is determined at compile time rather than runtime.

mtl1979:
The issue is if `Z_TLS` is defined as empty and something in `cpu_check_features` causes switch from userland to kernel, which is slow. As we don't use thread-local storage, it's possible that the value isn't initialized when some other thread tries to access the stub.

nmoinvaz:
Looks like that only applies to `crc32_stub`.

nmoinvaz:
I made a commit that should fix that issue, and actually if that case were to happen where a thread accesses it before CPU features have been checked, it will now use a better optimized `crc32` function for the endianness.

mtl1979:
Making something better is always more desired choice than introducing regressions in corner cases...

KungFuJesus:
Looking at the preprocessor behavior, we entirely discount optimized versions of string comparisons if unaligned loads aren't legal and we know we have an unaligned access?  

Strictly speaking, there are ways to do unaligned loads with VSX, VMX, and family, though they are a tad bit hairier.  If I remember correctly, Apple's documentation suggested the best way to deal with it is to partially overlap the previous unaligned load carried in the loop and recombine the register contents to a single full 128 bits of content.  Now, for 258 bytes...I dunno if this is that worthwhile, but, maybe?  I know that VMX specifically made unaligned loads a bit simpler, though, by doing something weird that modified endianness for some reason.

mtl1979:
As far as I remember, big-endian PowerPC had instruction to calculate the shift amount required for two separate registers so the contents can be combined. This made unaligned loads a little bit cheaper than on architecture which can't calculate the shift amounts in same instruction. PowerPC also has "extra" offset parameter in most instructions.

The main issue is that compiler can still generate unaligned loads elsewhere if it tries to vectorize code during optimizations. With gcc it usually meant avoiding optimization level 3, which allowed vectorizing code that doesn't use vector intrinsics.

KungFuJesus:
> As far as I remember, big-endian PowerPC had instruction to calculate the shift amount required for two separate registers so the contents can be combined. This made unaligned loads a little bit cheaper than on architecture which can't calculate the shift amounts in same instruction. PowerPC also has "extra" offset parameter in most instructions.
> 
> The main issue is that compiler can still generate unaligned loads elsewhere if it tries to vectorize code during optimizations. With gcc it usually meant avoiding optimization level 3, which allowed vectorizing code that doesn't use vector intrinsics.

Hmm, I'm not sure how the elsewhere case affects this case though, where we are leveraging explicitly the vec_ family of intrinsics.  I suspect GCC would mostly leave that alone apart from rescheduling operations a little bit better for the target architecture.  The -O2 vs -O3 seems like it'd be a separate issue.

From what I recall with altivec, GCC kind of rarely autovectorizes code to begin with.  Though, I haven't tried it in a while (probably not since GCC 8).

mtl1979:
@KungFuJesus I was thinking both about other big-endian architectures -- as those generally don't like unaligned accesses --  and also combining adjacent loads and stores originally using general-purpose registers.

KungFuJesus:
> @KungFuJesus I was thinking both about other big-endian architectures -- as those generally don't like unaligned accesses -- and also combining adjacent loads and stores originally using general-purpose registers.

Ah can't say I've dealt much with MIPS or Z. I do think an unaligned compare258 may be of benefit on PowerPC with altivec though. VSX's special unaligned load may make more sense on modern POWER.

mtl1979:
I haven't had real PowerPC hardware since I changed jobs, so I can't do benchmarks even if I could write the code for compare258 using unaligned loads. Last time I used PPC version of Ubuntu inside qemu to learn AltiVec.

nmoinvaz:
Do we even need the `use_byfour`? Seems like this can be controlled via `UNALIGNED_OK` preprocessor.

functable.c
```diff
@@ -608,6 +600,8 @@ Z_INTERNAL uint32_t longest_match_slow_stub(deflate_state *const s, Pos cur_matc
     if (x86_cpu_has_avx2)
         functable.longest_match_slow = &longest_match_slow_unaligned_avx2;
 #  endif
+#else
```
Dead2, line 599: This, and the above functions too, do not have the _c function as a fallback in the UNALIGNED_OK path.
As far as I can see, this would remain unset if compiled without builtin_ctz and run on an 486 cpu for example.

nmoinvaz: It would use the `longest_match_slow_unaligned_16` function it doesn't need to use `longest_match_slow_c`.

image

nmoinvaz:
It looks like the `use_byfour` was an old detection that was used to determine if there is a 4 byte integer type. See https://github.com/zlib-ng/zlib-ng/commit/6c9bd474aa08312ef2e2e9655a80e18db24a1680. 
We shouldn't have to do this check anymore since we are targeting newer systems. 

`crc32_little` and `crc32_big` can be run without regard to `UNALIGNED_OK` because all the 4 byte reads are aligned.

nmoinvaz:
Squashed two commits.

functable.c
Dead2: Of course, no idea how I missed that. 🤦‍♂️

Dead2 (approved):
LGTM 👍

Issue #1088, opened 2021-12-28
Custom allocators can break alignment that's required for SIMD


valpackett:
I've been using zlib-ng with compat as my system zlib in a customized FreeBSD fork with no issues so far, until right now an innocent `npm install` run just crashed:

```
(lldb) bt
* thread #8, name = 'node', stop reason = signal SIGBUS: hardware error
  * frame #0: 0x000026e7e7271704 libz.so.6`crc32_fold_load(fold=0x000026e80c7450c8, fold0=0x0000000000e08d60, fold1=0x0000000000e08d50, fold2=0x0000000000e08d40, fold3=0x0000000000e08d30) at crc32_fold_pclmulqdq.c:221:14
[...]

PR #1096, opened 2022-01-02
Fixed crc32_fold alignment when deflate and inflate structures are not aligned.

nmoinvaz:
See #1088.

It is possible for an implementation to use their own custom allocators which don't allocate memory on alignment we are expecting. In this case, crc32_fold does not get aligned on 16-byte boundary. This fixes the problem by assigning `fold` to `buffer` on 16-byte alignment.

### DIFF
--- a/zutil.c
+++ b/zutil.c
@@ -109,3 +109,42 @@ void Z_INTERNAL zng_cfree(void *opaque, void *ptr) {
     Z_UNUSED(opaque);
     zng_free(ptr);
 }
+
+/* Since we support custom memory allocators, some which might not align memory as we expect,
+ * we have to ask for extra memory and return an aligned pointer. */
+void Z_INTERNAL *zng_calloc_aligned(zng_calloc_func zalloc, void *opaque, unsigned items, unsigned size, unsigned align) {
+    uintptr_t return_ptr, original_ptr;
+    int32_t alloc_size, align_diff;
+    void *ptr;
+
+    /* Allocate enough memory for proper alignment and to store the original memory pointer */
+    alloc_size = sizeof(void *) + (items * size) + align;
+    ptr = zalloc(opaque, 1, alloc_size);
+    if (!ptr)
+        return NULL;
+
+    /* Calculate return pointer address with space enough to store original pointer */
+    align_diff = align - ((uintptr_t)ptr % align);
+    return_ptr = (uintptr_t)ptr + align_diff;
+    if (align_diff < sizeof(void *))
+        return_ptr += align;
+
+    /* Store the original pointer for free() */
+    original_ptr = return_ptr - sizeof(void *);
+    memcpy((void *)original_ptr, &ptr, sizeof(void *));
+
+    /* Return properly aligned pointer in allocation */
+    return (void *)return_ptr;
+}
+
+void Z_INTERNAL zng_cfree_aligned(zng_cfree_func zfree, void *opaque, void *ptr) {
+    if (!ptr)
+        return;
+
+    /* Calculate offset to original memory allocation pointer */
+    void *original_ptr = (void *)((uintptr_t)ptr - sizeof(void *));
+    void *free_ptr = *(void **)original_ptr;
+
+    /* Free original memory allocation */
+    zfree(opaque, free_ptr);
+}
--- a/zutil.c
+++ b/zutil.c
@@ -117,6 +117,10 @@ void Z_INTERNAL *zng_calloc_aligned(zng_calloc_func zalloc, void *opaque, unsign
     int32_t alloc_size, align_diff;
     void *ptr;
 
+    /* If no custom calloc function used then call zlib-ng's aligned calloc */
+    if (zalloc == zng_calloc)
+        return zng_calloc(opaque, items, size);
+
     /* Allocate enough memory for proper alignment and to store the original memory pointer */
     alloc_size = sizeof(void *) + (items * size) + align;
     ptr = zalloc(opaque, 1, alloc_size);
@@ -138,6 +142,11 @@ void Z_INTERNAL *zng_calloc_aligned(zng_calloc_func zalloc, void *opaque, unsign
 }
 
 void Z_INTERNAL zng_cfree_aligned(zng_cfree_func zfree, void *opaque, void *ptr) {
+    /* If no custom cfree function used then call zlib-ng's aligned cfree */
+    if (zfree == zng_cfree) {
+        zng_cfree(opaque, ptr);
+        return;
+    }
     if (!ptr)
         return;
 
--- a/test/test_aligned_alloc.cc
+++ b/test/test_aligned_alloc.cc
@@ -1,0 +1,44 @@
+/* test_aligned_alloc.cc - Test zng_calloc_aligned and zng_cfree_aligned */
+
+#include <stdlib.h>
+#include <errno.h>
+#include <stdio.h>
+
+extern "C" {
+#  include "zbuild.h"
+#  include "zutil.h"
+}
+
+#include <gtest/gtest.h>
+
+void *zng_calloc_unaligned(void *opaque, unsigned items, unsigned size) {
+    uint8_t *pointer = (uint8_t *)calloc(1, (items * size) + 2);
+    if (pointer == NULL)
+        return pointer;
+    /* Store whether or not our allocation is aligned */
+    *pointer = ((uint64_t)(intptr_t)pointer + 1) % 2 == 0;
+    pointer++;
+    if (*pointer) {
+        /* Return pointer that is off by one */
+        pointer++;
+    }
+    return (void *)pointer;
+}
+
+void zng_cfree_unaligned(void *opaque, void *ptr) {
+    uint8_t *pointer = (uint8_t *)ptr;
+    pointer--;
+    /* Get whether or not our original memory pointer was aligned */
+    if (*pointer) {
+        /* Return original aligned pointer to free() */
+        pointer--;
+    }
+    free(pointer);
+}
+
+TEST(zalloc, aligned_64) {
+    void *return_ptr = zng_calloc_aligned(zng_calloc_unaligned, 0, 1, 100, 64);
+    ASSERT_TRUE(return_ptr != NULL);
+    EXPECT_EQ((intptr_t)return_ptr % 64, 0);
+    zng_cfree_aligned(zng_cfree_unaligned, 0, return_ptr);
+}
--- a/zutil.h
+++ b/zutil.h
@@ -125,8 +125,15 @@ extern z_const char * const PREFIX(z_errmsg)[10]; /* indexed by 2-zlib_error */
 void Z_INTERNAL *zng_calloc(void *opaque, unsigned items, unsigned size);
 void Z_INTERNAL   zng_cfree(void *opaque, void *ptr);
 
-#define ZALLOC(strm, items, size) (*((strm)->zalloc))((strm)->opaque, (items), (size))
-#define ZFREE(strm, addr)         (*((strm)->zfree))((strm)->opaque, (void *)(addr))
+typedef void *zng_calloc_func(void *opaque, unsigned items, unsigned size);
+typedef void  zng_cfree_func(void *opaque, void *ptr);
+
+void Z_INTERNAL *zng_calloc_aligned(zng_calloc_func zalloc, void *opaque, unsigned items, unsigned size, unsigned align);
+void Z_INTERNAL  zng_cfree_aligned(zng_cfree_func zfree, void *opaque, void *ptr);
+
+#define ZALLOC(strm, items, size) zng_calloc_aligned((strm)->zalloc, (strm)->opaque, (items), (size), 64)
+#define ZFREE(strm, addr)         zng_cfree_aligned((strm)->zfree, (strm)->opaque, (void *)(addr))
+
 #define TRY_FREE(s, p)            {if (p) ZFREE(s, p);}
 
 #endif /* ZUTIL_H_ */


crc32_fold.h
```diff
@@ -5,8 +5,12 @@
 #ifndef CRC32_FOLD_H_
 #define CRC32_FOLD_H_
 
+#define CRC32_FOLD_BUFFER_SIZE (16 * 5) + 16
+/* (sizeof(__m128i) * (4 folds & 1 partial fold)) + extra space for 16-byte alignment */
+
 typedef struct crc32_fold_s {
-    uint32_t ALIGNED_(16) fold[4 * 5];
+    uint8_t *fold;
```
mtl1979: This is bad, pointer is only ever 64 bits (4 or 8 bytes), but we are expecting 128-bits (16 bytes) to align properly. As such this assumes we are always misaligned.

KungFuJesus:
> Struct members behave the same way as inside unions, the data is flattened alike when C++ classes are allocated with in-place `new()`. Everything inside struct or union is `by value`, so if you try to assign directly, the destination must be a type that allows assigning by value. Normal pointers can't accept by value as they are basically references. `Const` pointers can be assigned by value.

```
#include <stdlib.h>
#include <stdint.h>
#include <stdio.h>

struct testStruct {
    int array[10];
    int someMember0;
    char someMember1;
};

int main(void)
{
    struct testStruct stackTest;
    struct testStruct *heapTest = (struct testStruct*)malloc(sizeof(struct testStruct) * 1);
    printf("pointer from stack = %p, address of array in stack = %p, address of first element = %p\n",
            stackTest.array, &stackTest.array, &stackTest.array[0]);
    printf("pointer from heap = %p, address of array in heap = %p, address of first element = %p\n",
            heapTest->array, &heapTest->array, &heapTest->array[0]);
    return 0;
}
```

```
pointer from stack = 0x7ffdfaf62400, address of array in stack = 0x7ffdfaf62400, address of first element = 0x7ffdfaf62400
pointer from heap = 0x55a541bab2a0, address of array in heap = 0x55a541bab2a0, address of first element = 0x55a541bab2a0
```

The first and _last_ are equivalent, but not the first and second.  The second is how the array in this struct is being used, which cannot be correct.
Err no wait, those _are_ the same...This does not jive right with the mental model I have for this.  So this must behave just like offsetof() in the context of a struct.

crc32_fold.h
nmoinvaz: Ok I have fixed it by making `crc32_fold_s` in inflate and deflate still have the `ALIGNED_`.

nmoinvaz: And moved buffer to be the first in the `crc32_fold_s` structure.

mtl1979:
@KungFuJesus Some things just don't make sense and contradict with what we're used to.... We just need to know how they work in our own code and forget all the other cases. I get migrate just by trying to read some code that works perfectly.

Dead2:
I think I would rather prefer if we did this workaround for the entire struct, meaning we allocate a little extra and then assign the struct to an aligned position if using a custom allocator.
That would ensure that the alignment of all the structs contents are correct, instead of potentially needing this workaround several places, and maybe at some point being forgotten.

There are also a few other structs we allocate that might benefit from proper alignment.
I think being able to enforce proper alignment of the structs and its contents would be a big win when it comes to resilience against this kind of bugs and the at times confusing debugging it makes for.

This patch already changes the size of the struct, thus requiring changes to nginx patches for example. That is the case either way.

Another thing we have considered at one point, but never implemented, is to do the allocations in the init functions as one big allocation and then assign the struct to different parts of that. That would reduce system calls (and negate the need for nginx to have a custom allocator), and would make it easy to align each struct. There are pros and cons to doing this though, but I am (slightly) leaning towards pro.

nmoinvaz:
> I think I would rather prefer if we did this workaround for the entire struct, meaning we allocate a little extra and then assign the struct to an aligned position if using a custom allocator.

I think we would have to store the original pointer to the memory allocation somewhere so it can be freed. 

Maybe if we just had general purpose alignment functions that stored the original pointer so it could be freed:

image

image

mtl1979:
Most custom allocators only store the original pointer but don't actually align the pointer returned to the calling function... that's why we usually see on 64-bit platforms misalignment by pointer size (8 bytes).

Looking at the GCC documentation, large values for alignment are not always honored, but as I said earlier, alignment by 16 byte boundary should be supported on most 64-bit systems.

For next zlib-ng major release, we might want to deprecate support for custom allocation functions unless compatibility mode is enabled and in that case, we should assume that all pointers can be unaligned.

zutil.c
```diff
@@ -109,3 +109,49 @@ void Z_INTERNAL zng_cfree(void *opaque, void *ptr) {
...
+    }
+
+    if (!ptr)
+        return NULL;
+
+    /* Calculate alignment after compensating for storing original pointer */
+    align_diff = ((uintptr_t)ptr + sizeof(void *)) % align;
+
+    /* Store the original pointer for free() */
+    original_ptr = (uintptr_t)ptr + align_diff;
```
mtl1979: This doesn't look right, `align_diff` is supposed to include the bytes needed for the original pointer.

KungFuJesus:
For things that actually support aligned_alloc (from C11), the returned pointer is actually what you feed to free.  For a long time now on Linux _mm_malloc() has been just a wrapper of aligned_alloc and prior to that posix_memalign.  Most POSIX operating systems support some form of this in the allocator, though Mac OS _may_ be the exception to this rule (I can't remember if they actually implemented it or not).  For windows, we're stuck with returning an offset from an original pointer and relying on the library to track these things through a custom free.  It's a shame, aligned allocators are incredibly convenient.

zutil.c
nmoinvaz: Calculating the alignment after compensating for the size of the pointer is easier because then we only have to skip ahead those number of bytes and store the original pointer.

zutil.c
```diff
@@ -109,3 +109,49 @@ void Z_INTERNAL zng_cfree(void *opaque, void *ptr) {
...
+    memcpy((void *)original_ptr, &ptr, sizeof(void *));
+
+    /* Return properly aligned pointer in allocation */
+    return (void *)(original_ptr + sizeof(void *));
```
mtl1979: This can be `ptr + align_diff` like above.

zutil.c
mtl1979: Yes, but currently you are off by 4/8 bytes in next 4 lines...

mtl1979: Should be `original_ptr - sizeof(void *)` to get the location to store the actual "original" pointer returned by allocation callback.

nmoinvaz: I'm not understanding. `original_ptr` is where the original pointer is stored. We want to store it and return the pointer plus the size of the pointer.

I'm trying to compensating for when `align_diff = 0`. In other words I want to align it *after* I store the pointer.

mtl1979: If `align_diff` is 0, then you must add `align` and subtract size of `void *`.

mtl1979: `align` is guaranteed to always be more than size of `void *`.

nmoinvaz: I've changed it a bit should be resolved now.

nmoinvaz:
One thing I could do is, if it is not a custom memory allocator then just call `zng_calloc`/`zng_cfree` and return, however what if somebody changes the stream custom memory allocators mid-stream, then it would try and free the allocations incorrectly.

nmoinvaz:
@Dead2 I haven't removed the original commit, because I am waiting on your thoughts on the new approach in the second commit.

Dead2:
@nmoinvaz That looks pretty much exactly like the concept I was thinking about, nice.
The problem with custom allocators changing mid-stream is a potential problem, however I imagine that would be a BAD thing even in stock zlib (perhaps it is also prohibited in docs already).

So you have my vote for the second commit 👍

nmoinvaz:
I have changed the first commit so it is just a bit of cleanup to make it easier to read.

nmoinvaz:
The header docs don't specifically say you *can't* change the zalloc and zfree pointers mid-stream.

nmoinvaz:
I split the first commit into its own PR and rebased.

nmoinvaz:
Rebased.

zutil.c
```diff
@@ -109,3 +109,52 @@ void Z_INTERNAL zng_cfree(void *opaque, void *ptr) {
...
+ * we have to ask for extra memory and return an aligned pointer. */
+void Z_INTERNAL *zng_calloc_aligned(zng_calloc_cb calloc, void *opaque, unsigned items, unsigned size, unsigned align) {
+    uintptr_t return_ptr, original_ptr;
+    int32_t alloc_size, align_diff;
+    void *ptr;
+
+    /* Allocate enough memory for proper alignment and to store the original memory pointer */
+    alloc_size = sizeof(void *) + (items * size) + align;
+
+    if (calloc) {
```
KungFuJesus: _might_ want to make sure this is resolving the pointer and not the calloc function.  Assuming you don't include stdlib.h anywhere on the include path this _shouldn't_ be the case, but it's possible.  Or maybe precedence makes the parameter name shadow the actual libc function.  In any case, it might behoove us to rename that parameter.

nmoinvaz: I have renamed it, but I have also removed the if statement, since even before zalloc is called, it is defaulted to zng_alloc, etc.

KungFuJesus: This then presents another issue though, no?  If we called zng_alloc, we definitely don't need to over allocate, as posix_memalign is handling this for us.  I mean technically this will work, of course, but we're adding more work for us to find the original pointer written 8 bytes behind the start of the allocation.  

Couldn't we just keep the check for NULL assignment at that pointer and invoke the custom allocators if so and if not, use the overallocate + write real pointer 8 bytes behind trick?  I know it gets tricky to handle things conditionally like this, but it seems reasonable.  Is it unreasonable to expect the user to not change the allocation function for the lifetime of the zstream object?

nmoinvaz: `zalloc` and `zfree` are never NULL because we set it to default value of `zng_calloc` and `zng_cfree` in `deflateInit` and `inflateInit`.

I have added a check that will just use `zng_calloc/cfree` if not using custom allocators.

KungFuJesus:
Excellent on that last commit :).  Now, we'll see what Mac OS's CI test does on my alignment assertion this time.

nmoinvaz:
@unrelentingtech can you test against these recent changes?

valpackett:
Works great!

Dead2:
One unfortunate thing is this (except the bypass in commit #2) does not actually ever get tested by the CI (at least not as far as codecov sees).
I guess we don't have any tests where we set a "custom" alloc/free, do we? (It doesn't really have to be custom at all, just needs to be set to the OS ones for example)

Other than that, this looks great. I would feel better about it if it did add some testing as well though.

KungFuJesus:
Hah, for kicks and giggles we could make the test use the highly dangerous and much frowned upon alloca() stack allocating functions (with a nop free function).  I imagine for a lot of our unit tests, including allocating a window, this would probably put us dangerously close to smashing the stack, though.

nmoinvaz:
@Dead2 I think I would prefer to add unit tests after Google Test is incorporated. I am waiting for some PRs to get merged so I can get back to Google Test PR.

valpackett:
> It doesn't really have to be custom at all, just needs to be set to the OS ones for example

In order to catch any other issues like this one, it should be something like `malloc(size + 1) + 1` that intentionally returns unaligned pointers all the time.

mtl1979:
> > It doesn't really have to be custom at all, just needs to be set to the OS ones for example
> 
> In order to catch any other issues like this one, it should be something like `malloc(size + 1) + 1` that intentionally returns unaligned pointers all the time.

Obviously it would need to add more than 1 to `size`, as it would need to try to realign the bad pointer again and test if it can write to end of allocation. It would basically test start of allocation + 1, start of allocation + 8, and for example end of allocation - 15 with and without lowest 4 bits zeroed.

Adding just 1 to `size` would also restrict minimum usable value of `size` as noticed earlier when our code crashed when passing for example `7` as size.

valpackett:
ping. Is this going to land?

Dead2:
@unrelentingtech This is really only waiting for a test to be written, currently this added code is never actually run during our CI tests, and that means this code only gets compile-tested. Therefore, #1126 might end up going in before this.

nmoinvaz:
I have added unit tests now that Google test framework is available. 

I have also made a fix by changing:
```
align_diff = (uintptr_t)ptr % align;
```
to: 
```
align_diff = align - ((uintptr_t)ptr % align);
```

nmoinvaz:
@iii-i it seems the S390x CI keeps not running on a regular basis..

iii-i:
@nmoinvaz thanks for the heads up - apparently a leftover container has blocked further runs. I've deleted it and the service is running again. I'll work on adjusting the scripts to be more resilient to this.

Dead2:
https://github.com/zlib-ng/zlib-ng/runs/5544422151?check_suite_focus=true
```
Run sudo apt-get update
Hit:1 http://azure.archive.ubuntu.com/ubuntu focal InRelease
Hit:2 http://azure.archive.ubuntu.com/ubuntu focal-updates InRelease
Hit:3 http://azure.archive.ubuntu.com/ubuntu focal-backports InRelease
Hit:4 https://packages.microsoft.com/ubuntu/20.04/prod focal InRelease
Hit:[5](https://github.com/zlib-ng/zlib-ng/runs/5544422151?check_suite_focus=true#step:5:5) http://security.ubuntu.com/ubuntu focal-security InRelease
Hit:[6](https://github.com/zlib-ng/zlib-ng/runs/5544422151?check_suite_focus=true#step:5:6) http://ppa.launchpad.net/ondrej/php/ubuntu focal InRelease
Hit:[7](https://github.com/zlib-ng/zlib-ng/runs/5544422151?check_suite_focus=true#step:5:7) http://ppa.launchpad.net/ubuntu-toolchain-r/test/ubuntu focal InRelease
Reading package lists...
Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 wine32:i3[8](https://github.com/zlib-ng/zlib-ng/runs/5544422151?check_suite_focus=true#step:5:8)6 : Depends: libwine:i386 (= 5.0-3ubuntu1) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```
Perhaps we should ignore errors on `apt-get update`?
We can still error on `apt-get install` of course.

Dead2 (approved):

PR #1097, opened 2022-01-03
Split up example into several test binaries.

nmoinvaz:
This PR splits `example` into multiple test binaries. It has some benefits:

* No longer have a huge C file with a bunch of tests that are interdependent upon one another.
* Promotes development of single test binaries by developers. 
* Easier to include new tests into the eco-system.

Here are what the projects look like in Visual Studio:

image


test/test_shared.h
```diff
@@ -0,0 +1,31 @@
...
+
+    if (zVersion()[0] != my_version[0]) {
+        error("incompatible zlib version\n");
+    } else if (strcmp(zVersion(), PREFIX2(VERSION)) != 0) {
+        fprintf(stderr, "warning: different zlib version\n");
+    }
+
+    printf("zlib-ng version %s = 0x%08lx, compile flags = 0x%lx\n",
+            ZLIBNG_VERSION, ZLIBNG_VERNUM, PREFIX(zlibCompileFlags)());
+}
```
mtl1979: Missing empty line at end of file.

nmoinvaz:
@Dead2 initial thoughts? Any idea how you want the commits split up?

test/test_shared.h
nmoinvaz: Fixed.

nmoinvaz:
@Dead2 I could change this PR into these 4 commits:

- Add all new source code for new tests.
- Connect all new tests in build system.
- Delete old example.
- Rename fuzzer and possibly coverage tests.

What do you think? Also I'm looking for general feedback as well.

Dead2:
As per my other comment, I am a bit uncertain what I think about this one.
I love the cmake and naming cleanup for sure. 👍 

Whether splitting tests up into dozens of separate executables, that I am not sure about.
I do feel that it adds a lot of clutter to the build process, output files and build+test logs.
I would prefer one/few executables that run a bunch of tests but also handles the test failures better than today (don't just quit on the first error, also don't print a bunch of stuff if nothing fails unless --verbose).
Should we perhaps start using an actual test framework? Should we also have one for unit testing? (I hate making tests btw, but it is also so important)

In any case, I am delaying this until other PRs are merged.

mtl1979:
Test frameworks are generally bad idea when using a lot of different compiler versions as the range of compilers they support can be limited.

nmoinvaz:
I've used Google Test framework in the past. It is pretty decent. It uses C++. We could include it using FetchContent but that would only work for cmake. We'd have to clone the repo in configure. Nmake would have to figure out a way there. Then test it works on all our platforms.

nmoinvaz:
I agree there are some cons to splitting it up. You can no longer test all functionality using a single binary. You can run CTest though. And it does make the build Makefiles more complex. CMake changes are simple though. Configure script might be improved if we can loop through all test_* files.

mtl1979:
I still use Google Test with my other projects, but the projects use cmake only... I use older version of Google Test framework, so it works with clang 6.0 and gcc 6.5 and all more recent versions of both, but I've patched it and it still throws quite a lot of warnings.

nmoinvaz:
Another cavet to using Google Test (C++) is that CMake would need to set all the *CXX_FLAGS* in additional to the *C_FLAGS* we already use. This would be in order to test with sanitizers.

nmoinvaz:
Rebased.

nmoinvaz:
Closing in favor of #1126.

PR #1099, opened 2022-01-04
Remove old win32 readme.

nmoinvaz:
This file doesn't contain any valuable information and in some cases the information is out of date.

I would also think we should remove DLL_FAQ.txt for the same reasons.


mtl1979:
Unmaintained documentation is always bad...

PR #1101, opened 2022-01-06
Fix building shared tests

mtl1979:
* Don't add non-PIC gz sources to shared version of test executables if they are already included in shared library as PIC sources


mtl1979:
Seems like this doesn't need to be fixed for cmake, it already handles it correctly... @Dead2

Dead2 (approved):
LGTM

PR #1086, opened 2021-12-28
Use memcpy/memcmp for unaligned reads.

nmoinvaz:
Redid #934 using macros suggested in that PR.

PR #1102, opened 2022-01-07
Use memcpy for unaligned reads.

nmoinvaz:
Split from #1086.

### DIFF
--- a/chunkset.c
+++ b/chunkset.c
@@ -5,7 +5,7 @@
 #include "zbuild.h"
 #include "zutil.h"
 
-// We need sizeof(chunk_t) to be 8, no matter what.
+/* Define 8 byte chunks differently depending on unaligned support */
 #if defined(UNALIGNED64_OK)
 typedef uint64_t chunk_t;
 #elif defined(UNALIGNED_OK)
@@ -34,27 +34,24 @@ static inline void chunkmemset_1(uint8_t *from, chunk_t *chunk) {
 static inline void chunkmemset_4(uint8_t *from, chunk_t *chunk) {
 #if defined(UNALIGNED64_OK)
     uint32_t half_chunk;
-    half_chunk = *(uint32_t *)from;
+    memcpy(&half_chunk, from, sizeof(half_chunk));
     *chunk = 0x0000000100000001 * (uint64_t)half_chunk;
 #elif defined(UNALIGNED_OK)
-    chunk->u32[0] = *(uint32_t *)from;
+    memcpy(&chunk->u32[0], from, sizeof(chunk->u32[0]));
     chunk->u32[1] = chunk->u32[0];
 #else
     uint8_t *chunkptr = (uint8_t *)chunk;
-    memcpy(chunkptr, from, 4);
-    memcpy(chunkptr+4, from, 4);
+    memcpy(chunkptr, from, sizeof(uint32_t));
+    memcpy(chunkptr+4, from, sizeof(uint32_t));
 #endif
 }
 
 static inline void chunkmemset_8(uint8_t *from, chunk_t *chunk) {
-#if defined(UNALIGNED64_OK)
-    *chunk = *(uint64_t *)from;
-#elif defined(UNALIGNED_OK)
-    uint32_t* p = (uint32_t *)from;
-    chunk->u32[0] = p[0];
-    chunk->u32[1] = p[1];
+#if defined(UNALIGNED_OK) && !defined(UNALIGNED64_OK)
+    memcpy(&chunk->u32[0], from, sizeof(chunk->u32[0]));
+    memcpy(&chunk->u32[1], from+4, sizeof(chunk->u32[1]));
 #else
-    memcpy(chunk, from, sizeof(chunk_t));
+    memcpy(chunk, from, sizeof(uint64_t));
 #endif
 }
 
@@ -63,14 +60,7 @@ static inline void loadchunk(uint8_t const *s, chunk_t *chunk) {
 }
 
 static inline void storechunk(uint8_t *out, chunk_t *chunk) {
-#if defined(UNALIGNED64_OK)
-    *(uint64_t *)out = *chunk;
-#elif defined(UNALIGNED_OK)
-    ((uint32_t *)out)[0] = chunk->u32[0];
-    ((uint32_t *)out)[1] = chunk->u32[1];
-#else
-    memcpy(out, chunk, sizeof(chunk_t));
-#endif
+    memcpy(out, chunk, sizeof(uint64_t));
 }
 
 #define CHUNKSIZE        chunksize_c
--- a/insert_string_tpl.h
+++ b/insert_string_tpl.h
@@ -31,7 +31,7 @@
 #ifndef HASH_CALC_READ
 #  ifdef UNALIGNED_OK
 #    define HASH_CALC_READ \
-        val = *(uint32_t *)(strstart);
+        memcpy(&val, strstart, sizeof(val));
 #  else
 #    define HASH_CALC_READ \
         val  = ((uint32_t)(strstart[0])); \
--- a/compare258.c
+++ b/compare258.c
@@ -123,10 +123,12 @@ static inline uint32_t compare256_unaligned_32_static(const unsigned char *src0,
     uint32_t len = 0;
 
     do {
-        uint32_t sv = *(uint32_t *)src0;
-        uint32_t mv = *(uint32_t *)src1;
-        uint32_t diff = sv ^ mv;
+        uint32_t sv, mv, diff;
 
+        memcpy(&sv, src0, sizeof(sv));
+        memcpy(&mv, src1, sizeof(mv));
+
+        diff = sv ^ mv;
         if (diff) {
             uint32_t match_byte = __builtin_ctz(diff) / 8;
             return len + match_byte;
@@ -170,10 +172,12 @@ static inline uint32_t compare256_unaligned_64_static(const unsigned char *src0,
     uint32_t len = 0;
 
     do {
-        uint64_t sv = *(uint64_t *)src0;
-        uint64_t mv = *(uint64_t *)src1;
-        uint64_t diff = sv ^ mv;
+        uint64_t sv, mv, diff;
 
+        memcpy(&sv, src0, sizeof(sv));
+        memcpy(&mv, src1, sizeof(mv));
+
+        diff = sv ^ mv;
         if (diff) {
             uint64_t match_byte = __builtin_ctzll(diff) / 8;
             return len + (uint32_t)match_byte;
--- a/deflate.h
+++ b/deflate.h
@@ -302,13 +302,11 @@ typedef enum {
  * IN assertion: there is enough room in pending_buf.
  */
 static inline void put_short(deflate_state *s, uint16_t w) {
-#if defined(UNALIGNED_OK)
-    *(uint16_t *)(&s->pending_buf[s->pending]) = w;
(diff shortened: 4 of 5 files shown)


chunkset.c
```diff
@@ -34,27 +34,24 @@ static inline void chunkmemset_1(uint8_t *from, chunk_t *chunk) {
...
 }
 
 static inline void chunkmemset_8(uint8_t *from, chunk_t *chunk) {
-#if defined(UNALIGNED64_OK)
-    *chunk = *(uint64_t *)from;
-#elif defined(UNALIGNED_OK)
-    uint32_t* p = (uint32_t *)from;
-    chunk->u32[0] = p[0];
-    chunk->u32[1] = p[1];
+#if defined(UNALIGNED_OK) && !defined(UNALIGNED64_OK)
```
Dead2, line 50: I have been looking at this one too, is there any point to splitting this up into two memcpy calls or could we just use the single one below?
My limited testing does not show any change, but I have not dug into whether other `UNALIGNED_OK`-enabled code or endianness requires this somehow.

I do get the feeling that chunkset.c could use a refactoring after this though.

Dead2:
Ran some tests with GCC4.8 on X5660

## Develop 3adbc36c8aa1ddaeab1f8d2fd7122106d3221889
```
 Tool: minigzip   Levels: 1-9
 Runs: 50         Trim worst: 28

 1     54.165%      2.678/2.713/2.738/0.018        1.178/1.245/1.264/0.020      153,422,996
 2     43.877%      2.435/2.472/2.490/0.013        0.604/0.631/0.648/0.013       62,140,932
 3     42.381%      2.160/2.222/2.241/0.021        0.470/0.482/0.490/0.006       46,684,570
 4     41.665%      2.118/2.135/2.148/0.010        0.378/0.394/0.407/0.008       39,338,830
 5     41.211%      2.362/2.417/2.440/0.021        0.372/0.394/0.403/0.008       38,910,121
 6     41.036%      2.251/2.288/2.305/0.012        0.310/0.322/0.330/0.006       32,287,513
 7     40.784%      2.286/2.317/2.340/0.015        0.229/0.254/0.265/0.011       25,671,718
 8     40.707%      2.108/2.137/2.150/0.010        0.177/0.191/0.197/0.006       19,217,360
 9     40.419%      2.555/2.590/2.603/0.010        0.187/0.195/0.200/0.004       19,081,493

 avg1  42.916%                        2.366                          0.457
 tot                                468.433                         90.423      436,755,533
 
 130407    1456      32  131895   20337 libz-ng.so.2
```

## PR #1102 08bda6391a4e293282fc51f68c879b7c866c7a89
```
 Tool: minigzip   Levels: 1-9
 Runs: 50         Trim worst: 28

 1     54.165%      2.684/2.720/2.743/0.015        1.194/1.240/1.267/0.022      153,422,996
 2     43.877%      2.474/2.496/2.513/0.013        0.597/0.625/0.647/0.013       62,140,932
 3     42.381%      2.205/2.236/2.251/0.014        0.461/0.480/0.493/0.010       46,684,570
 4     41.665%      2.105/2.143/2.161/0.015        0.373/0.391/0.402/0.008       39,338,830
 5     41.211%      2.377/2.439/2.460/0.019        0.379/0.393/0.403/0.007       38,910,121
 6     41.036%      2.282/2.307/2.321/0.011        0.304/0.324/0.330/0.007       32,287,513
 7     40.784%      2.304/2.330/2.343/0.011        0.247/0.258/0.271/0.007       25,671,718
 8     40.707%      2.119/2.144/2.153/0.009        0.175/0.193/0.200/0.007       19,217,360
 9     40.419%      2.578/2.591/2.603/0.009        0.176/0.193/0.201/0.006       19,081,493

 avg1  42.916%                        2.378                          0.455
 tot                                470.912                         90.132      436,755,533
 
 130199    1456      32  131687   20267 libz-ng.so.2
```

Up to around 0.8% slower than before. Not great, but acceptable for this amount of cleanup I think.

chunkset.c
nmoinvaz: I suppose that depends on if the compiler automatically converts 1 x 8 copy to 2 x 4 copies if the architecture doesn't support 64-bit unaligned copies.

Dead2: Hmm, I thought that was guaranteed to be handled by memcpy automatically, now I wonder

nmoinvaz: I'm not exactly sure what hardware configuration this would apply to. I don't mind if `chunk_t` was only `uint64_t` and we got rid of the `ifndef UNALIGNED64_OK` paths. Perhaps that can be done as a separate PR..

nmoinvaz: https://godbolt.org/z/sxq61bYPK

nmoinvaz: image

For ARM (32-bit) it requires at least GCC 9 in order to take advantage of it, otherwise it does do a call to `memcpy`. But it does convert it to two 4 byte unaligned operations.

Perhaps I should add back `zmemcpy` and have a switch for less than GCC ARM 9. What do you think?

Dead2: memcpy is required to handle unaligned buffers btw.

Seems like GCC8 on ARM just calls the generic memcpy, but GCC9 uses its builtin handling.

From what I gather, just using memcpy should be perfectly safe (and probably more performant too).
I re-ran the tests with changing the below to always use the 8-byte copy, and that regains *some* of the performance loss too. Possibly GCC 4.8 issues two calls to the generic memcpy function.

I don't really think we need to special-case memcpy from what I am seeing so far.

Dead2:
Bench with this PR, but with chunkmemset_8 always using the single 8-byte memcpy
```
 1     54.165%      2.669/2.720/2.749/0.021        1.198/1.233/1.252/0.014      153,422,996
 2     43.877%      2.452/2.489/2.514/0.017        0.600/0.627/0.642/0.011       62,140,932
 3     42.381%      2.192/2.232/2.253/0.015        0.454/0.484/0.495/0.011       46,684,570
 4     41.665%      2.122/2.148/2.169/0.015        0.373/0.392/0.403/0.009       39,338,830
 5     41.211%      2.407/2.435/2.448/0.012        0.377/0.392/0.401/0.008       38,910,121
 6     41.036%      2.265/2.307/2.329/0.017        0.309/0.322/0.330/0.005       32,287,513
 7     40.784%      2.288/2.322/2.337/0.014        0.246/0.257/0.264/0.005       25,671,718
 8     40.707%      2.125/2.146/2.156/0.008        0.179/0.190/0.196/0.005       19,217,360
 9     40.419%      2.544/2.589/2.599/0.011        0.175/0.193/0.199/0.006       19,081,493

 avg1  42.916%                        2.376                          0.454
 tot                                470.529                         89.977      436,755,533
 130199    1456      32  131687   20267 libz-ng.so.2
```

chunkset.c
```diff
@@ -34,27 +34,24 @@ static inline void chunkmemset_1(uint8_t *from, chunk_t *chunk) {
...
 #elif defined(UNALIGNED_OK)
-    chunk->u32[0] = *(uint32_t *)from;
+    memcpy(&chunk->u32[0], from, sizeof(chunk->u32[0]));
     chunk->u32[1] = chunk->u32[0];
 #else
     uint8_t *chunkptr = (uint8_t *)chunk;
-    memcpy(chunkptr, from, 4);
-    memcpy(chunkptr+4, from, 4);
+    memcpy(chunkptr, from, sizeof(uint32_t));
+    memcpy(chunkptr+4, from, sizeof(uint32_t));
```
Dead2, line 45: I think this second memcpy is erronous btw, I thought them using the same from pointer strange.
Removing the second one completely seems to make no difference.

chunkset.c
nmoinvaz: Here is another view:
image
You can see in the last compiler ARM GCC 4 that the first one calls `memcpy` and the second one doesn't.

chunkset.c
nmoinvaz: We are doing essentially a `memset` with size of 4. So our chunk is 8 bytes. We have to copy the same 4 byte value from `from` twice.

I'm not sure why it didn't make a difference. Perhaps that code is not being hit.

Dead2: No, I think this function is supposed to copy only 4 bytes.

nmoinvaz: Yes, but the chunk size is 8 bytes. So it copies four bytes twice. If our chunk is `uint64_t` it needs to be able to use that `uint64_t` to perform multiple copies of the 4 bytes in a row over and over again.

nmoinvaz: image
That is the original code e40d88adc9e8180969c6c56a0deb4ec69c3ec92b. You can see it only does 1 x 4 byte copy. But then it uses the instruction `_mm_set1_epi32` to duplicate that.

In our code here we are essentially doing what `_mm_set1_epi32` does but with 64-bit integer instead of 128-bit integer as the chunk that we will eventually copy over and over.

chunkset.c
Dead2: Yes, but I am not sure the performance loss is enough to warrant the workaround, in addition to it only happening on the older compilers AFAIK. If the alternative is unaligned type-punning I mean.

chunkset.c
Dead2: I see your point now. I still don't grok the chunkset stuff it seems, I never quite did.

chunkset.c
nmoinvaz: Ok that makes sense. I can rework it in another PR after this one is merged. I will just remove the `typedef struct chunk_t { uint32_t u32[2]; } chunk_t;` and all its associated code as most modern compilers are able to do 8 byte `memcpy` optimizations themselves.

chunkset.c
nmoinvaz: It is easier to understand and look at one of the chunkset sources for instruction sets like `chunkset_sse.c`. There are still some parts of chunkset overall that are fuzzy to me as well.

chunkset.c
nmoinvaz: I wonder if we should just also not do `zmemcmp` and only replace with `memcmp`. And we can just note these compiler oddies in the Wiki page somewhere. We do list recommended build minimums to be GCC 5.5. Then in CMake and Configure we could print a warning if using a compiler that outputs suboptimally. Thoughts?

Dead2: @nmoinvaz I am keeping that possibility open for sure, but I want to do the benchmarks and such first so we have better knowledge about the downsides before deciding.

Dead2 (approved):
LGTM

insert_string_tpl.h
```diff
@@ -31,7 +31,7 @@
 #ifndef HASH_CALC_READ
 #  ifdef UNALIGNED_OK
 #    define HASH_CALC_READ \
-        val = *(uint32_t *)(strstart);
+        memcpy(&val, strstart, sizeof(val));
```
Hello71, line 34: why is this guarded by UNALIGNED_OK? it seems equivalent to #else case below now

nmoinvaz: The only thing I can think of is maybe it would be better if it checked for big endian rather than unaligned ok. I have noticed it in one of my other PRs but not sure about it. Feel free to open a PR.

mtl1979: Obviously on big-endian architectures the reads need to be done byte-by-byte and combined one-by-one as 32-bit variable, otherwise the CPU would trap because of misalignment. Doing byte-by-byte also makes sure little- and big-endian architectures give exactly same result when the bytes are combined as 32-bit unsigned integer.

PR #1103, opened 2022-01-08
Memcopy cleanup

nmoinvaz:


mtl1979:
Requiring gcc 9 or newer is quite bad in my opinion, as a lot of operating systems still require gcc 6.5 or older, especially when running under qemu. Limiting the usable gcc versions essentially limits what projects can switch from zlib to zlib-ng.

When Android NDK did switch to clang 9 from older gcc, it basically made compiling binaries impossible as newer versions are missing some essential functions for co-operative threading.

nmoinvaz:
Older versions of GCC can compile just fine they just don't optimize well on 32-bit ARM. It's no secret that newer versions of GCC can probably optimize better than older versions.

mtl1979:
It's still a regression... We're basically reverting essential optimizations...

nmoinvaz:
Actually GCC < 9.3 on 32-ARM doesn't even produce unaligned read for 32-bit integer with `memcpy(dst, src, 4)`. You would have to specify `*(uin32_t*)dst = *(uint32_t *)src`. Sounds like a compiler issue to me. I suggested to @Dead2 we just put in warnings in configure and CMake when encountering older versions of compilers that we know won't optimize well. I can always bring back the `zmemcpy_4` stuff.. but that may produce type alias warnings..?

nmoinvaz:
I thought the whole reason to use memcpy and memcmp was to allow the compiler to choose what to do.

mtl1979:
@nmoinvaz Requiring gcc 9.0 or later is same as requiring AVX512 or better... Basically it would exclude most of the systems we are supposed to still support.

As done before, if compiler clearly can't handle something correctly, we use workarounds instead of letting compiler generate suboptimal code or return build error or even crash or corrupt data.

nmoinvaz:
We are not requiring them to use GCC 9.3 or better.  They can use whatever version they want. It is the compiler choosing to output the way it is - for whatever reason. If people want better performance they should use a newer version. We should try and force old compilers to work like new compilers. And truly if people wanted better performance they wouldn't be using 32-bit ARM, they would use 64-bit.

mtl1979:
64-bit applications don't give better performance, they just allow accessing more memory, but also use more memory due to increased length of instructions.

nmoinvaz:
I'm not talking about 64-bit applications. I'm talking about 64-bit CPUs where all the improvements are happening these days.

nmoinvaz:
If your point is that there is a regression, I already know and I have already mentioned it in the commit and also to @Dead2 in another PR. I'm the one who took the time and discovered the behavior of this ARM GCC variant. If you have a solution I'm open to hear it.

mtl1979:
> I'm not talking about 64-bit applications. I'm talking about 64-bit CPUs where all the improvements are happening these days.

Obviously you can run 32-bit applications on 64-bit processor... That gives performance hit due to being able to fetch more instructions at given clock cycle.

mtl1979:
> If your point is that there is a regression, I already know and I have already mentioned it in the commit and also to @Dead2 in another PR. I'm the one who took the time and discovered the behavior of this ARM GCC variant. If you have a solution I'm open to hear it.

I wasn't really fond of the idea when we merged source codes for targeting 32-bit ARM and AArch64... Originally there was "slight" differences, but it seems those differences are not "slight" anymore...

Obviously the best solution is to use memcpy only for every other platform except 32-bit ARM when using older GCC version. That's what for we originally had architecture specific functions and macros.

nmoinvaz:
ARM GCC 8.5 requires `-march` greater than `armv5` to do unaligned access optimizations.
ARM GCC 9.3 does not require `-march` to do unaligned access optimizations.

https://godbolt.org/z/1rT6voqvM

nmoinvaz:
So perhaps we need an `UNALIGNEDFLAG` in both CMake and Configure that is used for armv6+ when using ARM GCC < 9.3.

Or perhaps it would be enough to compile with `NATIVEFLAG` when compiling with ARM GCC < 9.3.

nmoinvaz:
Even with that, many old compilers don't use unaligned access instructions to replace `memcmp`. Perhaps we should just have a `FORCE_UNALIGNED` option that can be used by old compilers that would replace `memcpy/memcmp` with what we had before.

mtl1979:
I don't know what is the default processor for every single compiler... I've seen so far some compilers defaulting to either armv2 or armv7... At least cmake currently does require specifying processor explicitly to detect correct supported features.

PR #1107, opened 2022-01-09
Merge crc32_little and crc32_big

nmoinvaz:
Just an idea I had about cleaning up the `crc32` endian variants. Instead of doing the endian check in `functable`, we do it in `crc32` function. And the preprocessor macros are defined different per endian. The remove `crc32_generic` because it is not used anywhere.


Dead2 (approved):
LGTM

PR #1108, opened 2022-01-09
Improved AVX2 adler32 performance

KungFuJesus:
Did this by simply doing 32 bit horizontal sums and using the same sum
of absolute difference instructions as done in the SSE4 and AVX512_VNNI
versions.

### DIFF
--- a/arch/x86/adler32_avx2.c
+++ b/arch/x86/adler32_avx2.c
@@ -14,19 +14,24 @@
 
 #ifdef X86_AVX2_ADLER32
 
-/* 64 bit horizontal sum, adapted from Agner Fog's vector library. */
-static inline uint64_t hsum(__m256i x) {
-    __m256i sum1 = _mm256_shuffle_epi32(x, 0x0E);
-    __m256i sum2 = _mm256_add_epi64(x, sum1);
-    __m128i sum3 = _mm256_extracti128_si256(sum2, 1);
-#if defined(__x86_64__) || defined(_M_X64)
-    return _mm_cvtsi128_si64(_mm_add_epi64(_mm256_castsi256_si128(sum2), sum3));
-#else
-    __m128i ret_vec = _mm_add_epi64(_mm256_castsi256_si128(sum2), sum3);
-    uint64_t ret_val;
-    _mm_storel_epi64((__m128i*)&ret_val, ret_vec);
-    return ret_val;
-#endif
+/* 32 bit horizontal sum, adapted from Agner Fog's vector library. */
+static inline uint32_t hsum(__m256i x) {
+    __m128i sum1  = _mm_add_epi32(_mm256_extracti128_si256(x, 1),
+                                  _mm256_castsi256_si128(x));
+    __m128i sum2  = _mm_add_epi32(sum1, _mm_unpackhi_epi64(sum1, sum1));
+    __m128i sum3  = _mm_add_epi32(sum2, _mm_shuffle_epi32(sum2, 1));
+    return (uint32_t)_mm_cvtsi128_si32(sum3);
+}
+
+static inline uint32_t partial_hsum(__m256i x) {
+    /* We need a permutation vector to extract every other integer. The
+     * rest are going to be zeros */
+    const __m256i perm_vec = _mm256_setr_epi32(0, 2, 4, 6, 1, 1, 1, 1);
+    __m256i non_zero = _mm256_permutevar8x32_epi32(x, perm_vec);
+    __m128i non_zero_sse = _mm256_castsi256_si128(non_zero);
+    __m128i sum2  = _mm_add_epi32(non_zero_sse,_mm_unpackhi_epi64(non_zero_sse, non_zero_sse));
+    __m128i sum3  = _mm_add_epi32(sum2, _mm_shuffle_epi32(sum2, 1));
+    return (uint32_t)_mm_cvtsi128_si32(sum3);
 }
 
 Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_t len) {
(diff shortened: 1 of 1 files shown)


KungFuJesus:
The before:
```
2022-01-09T11:56:10-05:00
Running ./adler32_benchmark
Run on (28 X 4100.04 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x14)
  L1 Instruction 32 KiB (x14)
  L2 Unified 1024 KiB (x14)
  L3 Unified 19712 KiB (x1)
Load Average: 0.20, 0.25, 0.17
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
adler32_c_bench               42731 ns        42730 ns        72151
adler32_ssse3_bench           12177 ns        12177 ns       100000
adler32_avx2_bench             9635 ns         9634 ns       100000
adler32_avx512_bench           6598 ns         6597 ns       100000
adler32_avx512vnni_bench       5415 ns         5415 ns       100000
```

The after:

```
2022-01-09T11:56:56-05:00
Running ./adler32_benchmark
Run on (28 X 4099.54 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x14)
  L1 Instruction 32 KiB (x14)
  L2 Unified 1024 KiB (x14)
  L3 Unified 19712 KiB (x1)
Load Average: 0.13, 0.22, 0.16
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
adler32_c_bench               43102 ns        43101 ns        73120
adler32_ssse3_bench           12219 ns        12218 ns       100000
adler32_avx2_bench             7295 ns         7295 ns       100000
adler32_avx512_bench           6419 ns         6419 ns       100000
adler32_avx512vnni_bench       5391 ns         5391 ns       100000
```

arch/x86/adler32_avx.c
```diff
@@ -48,19 +54,17 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
     if (UNLIKELY(len < 16))
         return adler32_len_16(adler, buf, len, sum2);
 
-    const __m256i vs_mask = _mm256_setr_epi32(0, 0, 0, 0, 0, 0, 0, -1);
-    __m256i vs1 = _mm256_set1_epi32(adler);
-    __m256i vs2 = _mm256_set1_epi32(sum2);
-    vs1 = _mm256_and_si256(vs1, vs_mask);
-    vs2 = _mm256_and_si256(vs2, vs_mask);
+    __m256i vs1 = _mm256_castsi128_si256(_mm_cvtsi32_si128(adler));;
```
nmoinvaz: Double semi-colons at the end of this line and the next.

arch/x86/adler32_avx.c
```diff
@@ -14,19 +14,25 @@
...
+static inline uint32_t hsum(__m256i x) {
+    __m128i sum1  = _mm_add_epi32(_mm256_extracti128_si256(x, 1),
+                                  _mm256_castsi256_si128(x));
+    __m128i sum2  = _mm_add_epi32(sum1, _mm_unpackhi_epi64(sum1, sum1));
+    __m128i sum3  = _mm_add_epi32(sum2, _mm_shuffle_epi32(sum2, 1));
+    return (uint32_t)_mm_cvtsi128_si32(sum3);
+}
+
+static inline uint32_t partial_hsum(__m256i x)
+{
```
nmoinvaz: Opening bracket should be on previous line.

nmoinvaz (approved):

KungFuJesus:
Using @nmoinvaz's improved benchmarks from his PR, here's the improvements with better measurements, as taken from a low clock speed haswell -U series laptop (Thinkpad x240):

Before: 
```
adler32/avx2/2048            185 ns          203 ns      3869309
adler32/avx2/4096            346 ns          371 ns      2123199
adler32/avx2/32768          2393 ns         2576 ns       272217
adler32/avx2/262144        19845 ns        21539 ns        32886
adler32/avx2/2097152      164407 ns       179753 ns         3852
adler32/avx2/4194304      382851 ns       410988 ns         1703
```

After:
```
adler32/avx2/2048            111 ns          122 ns      5733796
adler32/avx2/4096            209 ns          231 ns      2918230
adler32/avx2/32768          1666 ns         1795 ns       397546
adler32/avx2/262144        17293 ns        18668 ns        29235
adler32/avx2/2097152      141763 ns       152206 ns         4095
adler32/avx2/4194304      342989 ns       368295 ns         1865
```

arch/x86/adler32_avx.c
```diff
@@ -109,41 +114,22 @@ Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const unsigned char *buf, size_
...
-        __m256i s2lo256 = _mm256_cvtepi32_epi64(s2lo);
-        __m256i s2hi256 = _mm256_cvtepi32_epi64(s2hi);
-
-        /* Sum vectors in existing lanes */
-        __m256i s1_sum = _mm256_add_epi64(s1lo256, s1hi256);
-        __m256i s2_sum = _mm256_add_epi64(s2lo256, s2hi256);
-        
         /* In AVX2-land, this trip through GPRs will probably be unvoidable, as there's no cheap and easy
          * conversion from 64 bit integer to 32 bit. This casting to 32 bit is cheap through GPRs 
          * (just register aliasing), and safe, as our base is significantly smaller than UINT32_MAX */
```
Dead2: Looks like this comment might need to get an update. The cast to 32 bit is removed for example.

KungFuJesus: Yes and no.  The optimization the compiler is performing for that fixed modulus involves casting up to 64 bit precision and shifting the carry bits back into 32 bit.  For that reason we don't do the modulus within the vector register - it's still a chore to do so.  We also don't happen to need to do it to all of them, and since summing into the first vector position and moving it into eax is relatively cheap to begin with, there's definitely no reason to prefer to not make the trip to GPRs.

Arguably, if we could inexpensively do the modulus in vector registers, we could avoid the scalar horizontal sum, I guess.  But yes, the sequence needed to go to 64 bit precision and back to 32 bit is quite high, likely higher than the horizontal sum + move to GPR + move back when scoreboarded.  Basically what I'm saying is the ALU is doing a cheap and necessary cast for us here that we can't as easily do in both directions on the FPU.

KungFuJesus: Also even if we could cheaply move between 32 bit to 64 bit back to 32 bit precision again in the vector register - ultimately the thing that would murder all of our cycles would be the multiplication.  Full 64 bit multiplication is very expensive, and even though I think the "low" 64 bits are all that we'd need, that instruction has quite a bit of latency associated with it, and we'd need to do it twice since we'd be split over two 256 bit registers for that precision, with no real hope of ILP.

KungFuJesus: I modified the comment to clarify exactly what I'm referring to when I talk about switching precision.  Let me know if it's still too vague.

KungFuJesus:
Rebased.

Dead2 (approved):
LGTM

PR #1106, opened 2022-01-09
Change compare258 to compare256

nmoinvaz:
Requires PR #1095. Should reduce code size by around 1700 bytes (MSVC). We move the 2 byte check from `compare258` to `deflate_quick` where it is used then only need `compare256` everywhere. 

### DEVELOP https://github.com/zlib-ng/zlib-ng/commit/49d1704a1800c0b01241f13dee1dafe2961d1541
```
 Tool: ../zlib-ng/build/Release/minigzip-develop.exe Levels: 1-9
 Runs: 70         Trim worst: 40

 Level   Comp   Comptime min/avg/max/stddev  Decomptime min/avg/max/stddev  Compressed size
 1     54.166%      0.099/0.102/0.107/0.002        0.050/0.051/0.053/0.001        8,523,732
 2     43.871%      0.163/0.168/0.173/0.003        0.051/0.054/0.055/0.001        6,903,609
 3     42.387%      0.223/0.229/0.237/0.004        0.050/0.053/0.055/0.001        6,670,099
 4     41.647%      0.243/0.251/0.259/0.005        0.048/0.051/0.053/0.001        6,553,723
 5     41.216%      0.252/0.264/0.272/0.005        0.048/0.051/0.054/0.002        6,485,938
 6     41.037%      0.290/0.302/0.310/0.006        0.047/0.050/0.052/0.001        6,457,776
 7     40.778%      0.367/0.385/0.393/0.006        0.046/0.051/0.053/0.002        6,416,919
 8     40.704%      0.474/0.486/0.493/0.005        0.047/0.050/0.053/0.001        6,405,244
 9     40.409%      0.551/0.575/0.587/0.009        0.048/0.051/0.054/0.002        6,358,951

 avg1  42.913%                        0.307                          0.051
 tot                                 82.885                         13.851       60,775,991
```
### PR https://github.com/zlib-ng/zlib-ng/commit/e275c3aea793a602ab5b38891cc15488158fb4cb
```
 Tool: ../zlib-ng/build/Release/minigzip-compare256.exe Levels: 1-9
 Runs: 70         Trim worst: 40

 Level   Comp   Comptime min/avg/max/stddev  Decomptime min/avg/max/stddev  Compressed size
 1     54.166%      0.096/0.098/0.101/0.001        0.048/0.052/0.053/0.001        8,523,732
 2     43.871%      0.160/0.165/0.167/0.003        0.051/0.053/0.055/0.001        6,903,609
[...]

PR #1111, opened 2022-01-12
Adler32, compare256, crc32, slide_hash benchmarks.

nmoinvaz:
These benchmarks pull down Google Benchmark suite in CMake.

* Had to separate out all the CPU variant functions from functable.c so they can more easily be used by the benchmarks. 
* Had to extern "C" some cpu variables in order to detect CPU features in the benchmarks.

image


KungFuJesus:
Awesome.  One quality of life change might be to, once the interfaces are established for the benchmark, write a macro or constexpr so that the user only has to add a function name, cpu capability name for it to declare an extern function, and subs in the function pointer into the shared routine so that the user can add new methods with just a single line.

arch/x86/x86.h
```diff
@@ -1,11 +1,15 @@
-/* cpu.h -- check for CPU features
+/* x86.h -- check for CPU features
 * Copyright (C) 2013 Intel Corporation Jim Kukunas
 * For conditions of distribution and use, see copyright notice in zlib.h
 */
 
 #ifndef CPU_H_
 #define CPU_H_
 
+#ifdef __cplusplus
```
KungFuJesus: If you wrap in the benchmark code `extern "C" { }` around the include you don't need to put `#ifdef __cplusplus`

nmoinvaz: I thought about that too but wasn't sure which way to go.

test/benchmarks/benchmark_compare258.cc
```diff
@@ -0,0 +1,73 @@
...
+    unsigned char str1[274];
+    unsigned char str2[274];
+    uint32_t len;
+    int32_t x = 0;
+
+    memset(str1, 'a', sizeof(str1));
+    memset(str2, 'a', sizeof(str2));
+
+    while (state.KeepRunning()) {
+        str2[x] = 0;
```
KungFuJesus: This shares the previous issue of faster routines rolling over the the len counter more times, landing at unpredictable states waiting for the previous iteration to complete.  We'll need to come up with a way to instead parameterize the length of this byte string comparison so that it's fixed within the infinite timing loop.

nmoinvaz: So we should use the same method of increasing by power of 2 up to 256..?

KungFuJesus: That's sensible for input lengths.  Since the timing will be data dependent we'll need a way to modify where the first unmatched character occurs prior to entering the benchmark loop.

KungFuJesus: oh duh, you're basically doing that with memsets.  So it's just a matter of modifying where the memset stops are based on range.

KungFuJesus:
> * Had to separate out all the CPU variant functions from functable.c so they can more easily be used by the benchmarks.

This may not have been entirely necessary, but it does save you the trouble of having to explicitly declare an extern function in the benchmark's source code.  It also kind of cleans up the function table implementation in that we can separate declarations versus assignment - though we still need preprocessor checks to make sure support was compiled for the feature, anyway.

KungFuJesus:
Might as well add slidehash to the mix:
https://gist.github.com/KungFuJesus/da987bfd0a70b09207168bf52d175ce2

I'm not sure if that input size range is quite realistic as to what gets used in practice - I just know that it must always have been a multiple of 16 uint16's or the AVX version would never work.  I believe it also happens to be a multiple of 32, as the AVX512 version I had written worked as well.

mtl1979:
@KungFuJesus For slidehash, the expected range is 1, 2, 4, 8, 16 or 32 kB. I don't think it makes much sense to use window smaller than 1 kB.

KungFuJesus:
> @KungFuJesus For slidehash, the expected range is 1, 2, 4, 8, 16 or 32 kB. I don't think it makes much sense to use window smaller than 1 kB.

Cool, the example can then be adjusted to instead use a fixed number of bins.  There's an easy way to do that, I think.  Yeah, looks like a RangeMultiplier of 2 and a range starting at 1024.

Edit: Gist updated to include sane ranges.  Still probably could use some refactor, for sure those extern function prototypes are no longer needed.  Also can be made more compact with the function pointer substitution like what's happening in the other benchmarks.

nmoinvaz:
@KungFuJesus I have incorporated your slide_hash gist. I have changed it to use Fixtures. Thoughts? I could change the rest of the tests the same way.

arch/x86/x86.h
nmoinvaz: Fixed.

test/benchmarks/benchmark_slidehash.cc
```diff
@@ -0,0 +1,78 @@
...
+#ifdef X86_SSSE3_ADLER32
+BENCHMARK_DEFINE_F(slide_hash, sse2)(benchmark::State& state) {
+    s_g->w_size = (uint32_t)state.range(0);
+
+    for (auto _ : state) {
+        slide_hash_sse2(s_g);
+        benchmark::DoNotOptimize(s_g);
+    }
+}
+BENCHMARK_REGISTER_F(slide_hash, sse2)->DenseRange(2048, MAX_RANDOM_INTS, 2048);
```
KungFuJesus: You'll want to modify this to use RangeMultiplier(2)->Range() like in my latest gist - that way we only test the possible sizes for slide_hash.

test/benchmarks/benchmark_slidehash.cc
```diff
@@ -0,0 +1,78 @@
...
+        s_g = s;
+    }
+
+    void TearDown(const ::benchmark::State& state) {
+        free(l0);
+        free(l1);
+    }
+};
+
+BENCHMARK_DEFINE_F(slide_hash, c)(benchmark::State& state) {
```
KungFuJesus: Hmm, the fixtures don't seem to be buying us a whole lot.  Don't suppose there's a good way to condense this into a preprocessor or constexpr statement where the function pointer gets subbed in and a cpu feature support flag is checked prior to adding it to the benchmark list?  It'd be great if we could do say:
`SLIDEHASH_BENCHMARK("slide_hash_avx2", slide_hash_avx2, x86_has_avx2);`
Instead of having to copy and paste that entire function body + function declaration per implementation.

Actually I feel like the most obvious thing to do is to write one test and parameterize out the function pointer.  Then we just declare variants by switching out the function pointer.  That should be simple, no?

nmoinvaz: Using the fixtures isolates the source code for each test which I like. But there might be some way to make a single function as you say. But in recent commit, I also have a check that checks that the CPU supports the instruction set, which might be harder to parameterize.

nmoinvaz: I'm not familiar enough with constexpr to know about that.

test/benchmarks/benchmark_slidehash.cc
nmoinvaz: Done.

test/benchmarks/benchmark_slidehash.cc
KungFuJesus: I mean it's better, that's for sure.  It's still about 6 lines for something I'd have hoped would be 1 or 2.  But, 2 of those lines are compile time checks for the support.

KungFuJesus: Actually what you have there now I think probably could be wrapped into a macro since they all have the same exact pattern, but you'd still need the compile time checks surrounding the macro.

test/benchmarks/benchmark_adler32.cc
```diff
@@ -0,0 +1,125 @@
...
+    void TearDown(const ::benchmark::State& state) {
+        free(random_ints);
+    }
+};
+
+
+BENCHMARK_DEFINE_F(adler32, c)(benchmark::State& state) {
+    Bench(state, adler32_c);
+}
+BENCHMARK_REGISTER_F(adler32, c)->Range(1, MAX_RANDOM_INTS);
```
KungFuJesus: Looking good so far, but we should probably change this to reflect number of bytes rather than integer.  As in, remove the *sizeof(uint32_t) within the Bench function and modify range to be 1 to sizeof(uint32_t) * MAX_RANDOM_INTS.

test/benchmarks/benchmark_slidehash.cc
nmoinvaz: Google benchmark supports Setup/Teardown without having to resort to fixtures....
https://github.com/google/benchmark/blob/main/docs/user_guide.md#setupteardown
So I wouldn't need to use them..

test/benchmarks/benchmark_compare258.cc
```diff
@@ -0,0 +1,90 @@
...
+
+#define MAX_COMPARE_SIZE (258)
+
+typedef uint32_t (*compare258_func)(const unsigned char *src0, const unsigned char *src1);
+
+class compare258: public benchmark::Fixture {
+private:
+    uint8_t str1[274];
+    uint8_t str2[274];
+public:
```
KungFuJesus: Minor suggestion:
```
diff --git a/test/benchmarks/benchmark_compare258.cc b/test/benchmarks/benchmark_compare258.cc
index bd75120..6b11088 100644
--- a/test/benchmarks/benchmark_compare258.cc
+++ b/test/benchmarks/benchmark_compare258.cc
@@ -21,8 +21,8 @@ typedef uint32_t (*compare258_func)(const unsigned char *src0, const unsigned ch
 
 class compare258: public benchmark::Fixture {
 private:
-    uint8_t str1[274];
-    uint8_t str2[274];
+    uint8_t str1[MAX_COMPARE_SIZE];
+    uint8_t str2[MAX_COMPARE_SIZE];
 public:
     void SetUp(const ::benchmark::State& state) {
         memset(str1, 'a', sizeof(str1));
@@ -33,7 +33,7 @@ public:
         int32_t x = (int32_t)state.range(0);
         uint32_t len;
 
-        while (state.KeepRunning()) {
+        for (auto _ : state) {
             str2[x] = 0;
             len = compare258((const uint8_t *)str1, (const uint8_t *)str2);
             str2[x] = 'a';
```

Unless you were oversizing those strings on purpose for alignment purposes?

KungFuJesus: Err, the diff is missing a brace, but you get the gist.

test/benchmarks/benchmark_adler32.cc
nmoinvaz: Good call. Done.

test/benchmarks/benchmark_compare258.cc
nmoinvaz: I think there was a reason I was oversizing them but I can't remember now. Anyways I have fixed it.

test/benchmarks/benchmark_slidehash.cc
```diff
@@ -0,0 +1,85 @@
...
+BENCHMARK_DEFINE_F(slide_hash, sse2)(benchmark::State& state) {
+    if (!x86_cpu_has_sse2) {
+        state.SkipWithError("CPU does not support SSE2");
+        return;
+    }
+    Bench(state, slide_hash_sse2);
+}
+BENCHMARK_REGISTER_F(slide_hash, sse2)->RangeMultiplier(2)->Range(1024, MAX_RANDOM_INTS);
+#endif
+#ifdef X86_AVX2_SLIDEHASH
```
KungFuJesus: Probably _should_ be fixed in CmakeLists.txt and configure but, this needs to be:
`X86_AVX2`

nmoinvaz: Yes I just noticed that as well.

nmoinvaz: Fixed.

KungFuJesus:
Heh these benchmarks sort of confirmed my suspicion about the SSE4 string instructions.  Once a few more of these PRs go through I'm definitely going to write an SSE2 version of that since the SSE4 is about twice as slow as unaligned_64 on my Haswell laptop.  Assuming the SSE2 method will be roughly half the performance of the AVX version, we should be looking at about a 2x gain over the unaligned_64 version (makes sense, it's a double wide comparison).

nmoinvaz:
How will you select between the two? Or are you saying SSE2 will be faster than SSE4?

KungFuJesus:
> How will you select between the two? Or are you saying SSE2 will be faster than SSE4?

I estimate the simpler sse2 version should be twice as fast as unaligned_64 version which is already twice as fast as the sse4. The problem with those string compare instructions is that they are made more for handling string specific things like not knowing where the null terminator sits or performing case insensitive compares. They also have a pretty high decode latency. It makes them pretty much worse for anything but a select few cases with string comparisons. Where it actually helps is probably however glibc is using them. 

I think the qt devs at one point put out a blog post about it. I seem to remember them using these instructions for multi byte character encodings, but the details were a little fuzzy.

KungFuJesus:
@nmoinvaz Slightly more compact representation :):
```
diff --git a/test/benchmarks/benchmark_adler32.cc b/test/benchmarks/benchmark_adler32.cc
index 491b327..ad8328e 100644
--- a/test/benchmarks/benchmark_adler32.cc
+++ b/test/benchmarks/benchmark_adler32.cc
@@ -18,6 +18,15 @@ extern "C" {
 
 #define MAX_RANDOM_INTS (1024 * 1024)
 #define MAX_RANDOM_INTS_SIZE (MAX_RANDOM_INTS * sizeof(uint32_t))
+#define BENCHMARK_ADLER(fixture, name, fptr, support_flag) \
+BENCHMARK_DEFINE_F(fixture, name)(benchmark::State& state) { \
+    if (!support_flag) { \
+        state.SkipWithError("CPU does not support " #name); \
+    } \
+    Bench(state, fptr); \
+}\
+BENCHMARK_REGISTER_F(adler32, name)->Range(2048, MAX_RANDOM_INTS_SIZE);
+
 
 typedef uint32_t (*adler32_func)(uint32_t adler, const unsigned char *buf, size_t len);
 
@@ -51,75 +60,29 @@ public:
 };
 
 
-BENCHMARK_DEFINE_F(adler32, c)(benchmark::State& state) {
-    Bench(state, adler32_c);
-}
-BENCHMARK_REGISTER_F(adler32, c)->Range(1, MAX_RANDOM_INTS_SIZE);
+BENCHMARK_ADLER(adler32, c, adler32_c, 1);
+
 #ifdef ARM_NEON_ADLER32
-BENCHMARK_DEFINE_F(adler32, neon)(benchmark::State& state) {
-    Bench(state, adler32_neon);
-}
-BENCHMARK_REGISTER_F(adler32, neon)->Range(1, MAX_RANDOM_INTS_SIZE);
+BENCHMARK_ADLER(adler32, neon, adler32_neon, arm_has_neon);
 #endif
 #ifdef PPC_VMX_ADLER32
-BENCHMARK_DEFINE_F(adler32, vmx)(benchmark::State& state) {
-    Bench(state, adler32_vmx);
-}
-BENCHMARK_REGISTER_F(adler32, vmx)->Range(1, MAX_RANDOM_INTS_SIZE);
+BENCHMARK_ADLER(adler32, vmx, adler32_vmx, power_cpu_has_altivec);
 #endif
 #ifdef X86_SSE41_ADLER32
-BENCHMARK_DEFINE_F(adler32, sse41)(benchmark::State& state) {
-    if (!x86_cpu_has_sse41) {
-        state.SkipWithError("CPU does not support SSE4.1");
-        return;
-    }
-    Bench(state, adler32_sse41);
-}
-BENCHMARK_REGISTER_F(adler32, sse41)->Range(1, MAX_RANDOM_INTS_SIZE);
+BENCHMARK_ADLER(adler32, sse41, adler32_sse41, x86_cpu_has_sse41);
 #endif
 #ifdef X86_SSSE3_ADLER32
-BENCHMARK_DEFINE_F(adler32, ssse3)(benchmark::State& state) {
-    if (!x86_cpu_has_ssse3) {
-        state.SkipWithError("CPU does not support SSSE3");
-        return;
-    }
-    Bench(state, adler32_ssse3);
-}
-BENCHMARK_REGISTER_F(adler32, ssse3)->Range(1, MAX_RANDOM_INTS_SIZE);
+BENCHMARK_ADLER(adler32, ssse3, adler32_ssse3, x86_cpu_has_ssse3);
 #endif
 #ifdef X86_AVX2_ADLER32
-BENCHMARK_DEFINE_F(adler32, avx2)(benchmark::State& state) {
-    if (!x86_cpu_has_avx2) {
-        state.SkipWithError("CPU does not support AVX2");
-        return;
-    }
-    Bench(state, adler32_avx2);
-}
-BENCHMARK_REGISTER_F(adler32, avx2)->Range(1, MAX_RANDOM_INTS_SIZE);
+BENCHMARK_ADLER(adler32, avx2, adler32_avx2, x86_cpu_has_avx2);
 #endif
 #if defined(X86_AVX512_ADLER32)
-BENCHMARK_DEFINE_F(adler32, avx512)(benchmark::State& state) {
-    if (!x86_cpu_has_avx512) {
-        state.SkipWithError("CPU does not support AVX512");
-        return;
-    }
-    Bench(state, adler32_avx512);
-}
-BENCHMARK_REGISTER_F(adler32, avx512)->Range(1, MAX_RANDOM_INTS_SIZE);
+BENCHMARK_ADLER(adler32, avx512, adler32_avx512, x86_cpu_has_avx512);
 #endif
 #if defined(X86_AVX512VNNI_ADLER32)
-BENCHMARK_DEFINE_F(adler32, avx512_vnni)(benchmark::State& state) {
-    if (!x86_cpu_has_avx512vnni) {
-        state.SkipWithError("CPU does not support AVX512_VNNI");
-        return;
-    }
-    Bench(state, adler32_avx512_vnni);
-}
-BENCHMARK_REGISTER_F(adler32, avx512_vnni)->Range(1, MAX_RANDOM_INTS_SIZE);
+BENCHMARK_ADLER(adler32, avx512_vnni, adler32_avx512_vnni, x86_cpu_has_avx512vnni);
 #endif
 #ifdef POWER8_VSX_ADLER32
-BENCHMARK_DEFINE_F(adler32, power8)(benchmark::State& state) {
-    Bench(state, adler32_power8);
-}
-BENCHMARK_REGISTER_F(adler32, power8)->Range(1, MAX_RANDOM_INTS_SIZE);
+BENCHMARK_ADLER(adler32, power8, adler32_power8, power_cpu_has_arch_2_07);
 #endif
```

KungFuJesus:
Unfortunately, for the compile time checks to make sure the function exists in the binary, there's no way to nest a macro to use a preprocessor definition inside of a macro statement.  Maybe there's a way to do it with constexpr since technically they are a little bit more powerful than macros, but I'm not sure.  In any case, it should make things way easier for future implementations to be slipped into the benchmark.

It's also a little fortunate that the adler32 name isn't clobbering.  I'm guessing something in the C++ grammar allows it to resolve the ambiguity between function argument name and class definition, so perhaps it's not something we need to worry about so much.

KungFuJesus:
Heh, I was curious:
https://github.com/lattera/glibc/blob/master/sysdeps/x86_64/multiarch/strcmp-sse42.S#L335

It comes as no surprise that they only use the version of this with the implicit string length, not explicit.  That way it saves them a search to find the null byte in both strings.  This also means that strncmp is actually _kind_ of suboptimal if you know you're in a range where you won't encounter the null byte terminator.

KungFuJesus:
One last change to propose:
https://gist.github.com/KungFuJesus/429cf1172bfc6d1e92715b97dd228f8f

I've included an explanatory comment for why.

KungFuJesus:
Excellent news - these benchmarks helped me find an artificial data pipeline stall in my vnni version that when I alleviated it, basically got twice as fast :).

nmoinvaz:
@KungFuJesus I've updated all the benchmarks with the new macro. Take a look.

nmoinvaz:
@KungFuJesus you are right on my machine SSE4 compare258 is also slower than unaligned-64 version. 

Perhaps we should get rid of SSE4 compare258 variant?... @mtl1979. @Dead2 

image

I'd also like to get rid of the unaligned_ prefix for sse4 and avx2 compare258 function names. Seems unnecessary. But maybe I can clean that up after #1106 is merged in.

nmoinvaz:
I have squashed the commits and taking this PR out of draft status. @KungFuJesus thank you for your invaluable feedback and help with this PR. I have added your name as co-author.

KungFuJesus:
> Perhaps we should get rid of SSE4 compare258 variant?... @mtl1979. @Dead2

Yes, likely.  I'll write an SSE2 version soon that replaces it.  The tzcnt instruction the AVX2 variant relies on doesn't exist but an equivalent one did on older CPUs (bit set first).  It has undefined behavior if the input is 0, but in the circumstance we're using it that is guaranteed to not be true.  The __builtin_ctz will call this with GCC, which is yet another case where having -march=native would help, as for BMI2 the BSF instruction is slower than tzcnt because of BSF's UB corner case, so Intel never really focused on making it faster.  But, if you have BMI2, you have AVX2 so you're probably not going to use that method, anyway.

test/benchmarks/benchmark_slidehash.cc
```diff
@@ -0,0 +1,80 @@
...
+        Bench(state, fptr); \
+    } \
+    BENCHMARK_REGISTER_F(slide_hash, name)->RangeMultiplier(2)->Range(1024, MAX_RANDOM_INTS);
+
+BENCHMARK_SLIDEHASH(c, slide_hash_c, 1);
+#ifdef X86_SSE2
+BENCHMARK_SLIDEHASH(sse2, slide_hash_sse2, x86_cpu_has_sse2);
+#endif
+#ifdef X86_AVX2
+BENCHMARK_SLIDEHASH(avx2, slide_hash_avx2, x86_cpu_has_avx2);
```
KungFuJesus, line 90: This is missing the neon and power implementations, right?

nmoinvaz: Good catch. Fixed.

mtl1979:
> @KungFuJesus you are right on my machine SSE4 compare258 is also slower than unaligned-64 version.
> 
> Perhaps we should get rid of SSE4 compare258 variant?... @mtl1979. @Dead2

If it can't be fixed, and it's slower than other compatible versions, then it should be removed. Like I've said before, using more recent instruction set doesn't always result in better performance. For me, it doesn't make sense to have zillion variants, if there is no performance gain.

KungFuJesus:
> > @KungFuJesus you are right on my machine SSE4 compare258 is also slower than unaligned-64 version.
> > Perhaps we should get rid of SSE4 compare258 variant?... @mtl1979. @Dead2
> 
> If it can't be fixed, and it's slower than other compatible versions, then it should be removed. Like I've said before, using more recent instruction set doesn't always result in better performance. For me, it doesn't make sense to have zillion variants, if there is no performance gain.

It's not so much that it's bad at what it does but this direct comparison isn't really what it's for.  It has some limited uses, like allowing a string comparison to skip a search for a null terminator or for finding a small substring like strstr does.  
https://www.strchr.com/strcmp_and_strlen_using_sse_4.2

Now, it's probably no coincidence they didn't extend these instructions to AVX - they are probably a bear to decode.  These definitely weren't the best designed instructions in the ISA, but they do have some limited utility.  This just isn't one of them.

mtl1979:
> It's not so much that it's bad at what it does but this direct comparison isn't really what it's for. It has some limited uses, like allowing a string comparison to skip a search for a null terminator or for finding a small substring like strstr does. https://www.strchr.com/strcmp_and_strlen_using_sse_4.2
> 
> Now, it's probably no coincidence they didn't extend these instructions to AVX - they are probably a bear to decode. These definitely weren't the best designed instructions in the ISA, but they do have some limited utility. This just isn't one of them.

I don't even pretend to be expert in x86 instruction set... But I know enough about that some instructions can be abused for other purposes by using for example 0 or 1 as immediate. As such, there is multiple ways to implement something and the most logical way is not always the best performing...

KungFuJesus:
One strange perhaps unintended side effect of where this sits is make test now runs benchmarks silently.  In the long term that may not be horrible if we wanted performance regression tests but for now they are likely to just slow down CI.  Unless they are writing the results somewhere on the backend?

Google benchmark does have compare.py if we did want to do some comparative analysis but that's probably outside the scope of this PR.  It's also kind of tricky because even if you do get good averages it's going to vary in performance depending on what node the CI gets assigned to.

nmoinvaz:
Yeah I didn't really put the benchmark into the test suite to measure performance. It was just run it to make sure it runs sort of thing.

KungFuJesus:
I guess that's sensible - if something breaks the ability to run a benchmark, we'd want to catch it before merging it in.  It's not as though the CI was lightning fast to begin with, hah.

KungFuJesus (approved):
Looks great to me.

nmoinvaz:
Rebased.

Dead2:
Just a couple of thoughts:
Could this also be used to test the accuracy of the results? Or should we make a separate version that does the accuracy tests?
Being able to test accuracy of each arch version would be useful to prevent and pinpoint regressions.

Also, could we perhaps have a parameter that tells it to return an error at the end if an arch version benchmarks as slower than the C version?

KungFuJesus:
So as-is, the fixtures newly initialize the data per test, so the results likely wouldn't be comparable.  There are ways to get return arguments from the test, but I don't think Google Benchmark was intended to be used as a unit testing framework, so it may be a bit kludgey.  

They do have a bundled script called compare.py that can run and compare the results of two separate benchmark binaries or even the same benchmark binary's results to itself with different methods for time.  But of course that much can be sort of obvious by just glancing at the results in a table.  If you're looking for something that's more "pass/fail" with regard to a certain percentage of the vanilla implementation's runtime, I don't know a great way to go about doing that.  

The way that this is built now (linking to the static lib) means that we can change the implementation, move the old benchmark binary to a different name, and recompile and run the new one.  That's sort of valuable because those two benchmark binaries can also be profiled separately.

KungFuJesus:
Thus far, this unit test has been pretty decent at finding all my issues:
https://github.com/zlib-ng/zlib-ng/blob/develop/test/adler32_test.c

The only issue is that is just runs what the library dispatches to so you have to build a limited binary if you want to test different versions.  Perhaps it could be augmented to instead run all of the internal methods?  Similarly for the CRC32, slide_hash, compare258, and other vectorized variants.

KungFuJesus:
Another cool feature of Google Benchmark, when compiled in conjunction with libpfm, you can count the cycles for the shorter run benchmarks:
```
-----------------------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations UserCounters...
-----------------------------------------------------------------------------------
adler32/c/2048                  536 ns          536 ns      1305902 CYCLES=2.19425k
adler32/sse41/2048              104 ns          104 ns      6743352 CYCLES=425.29
adler32/ssse3/2048              126 ns          126 ns      5571341 CYCLES=514.625
adler32/avx2/2048              64.5 ns         64.5 ns     10854508 CYCLES=231.788
adler32/avx512/2048            52.9 ns         52.9 ns     13384412 CYCLES=168.984
adler32/avx512_vnni/2048       33.9 ns         33.9 ns     20627247 CYCLES=91.6705
```

For larger run benchmarks this doesn't work because those PMU counters overflow, but it's a great way to get a clock agnostic view of how efficient something is.

nmoinvaz:
I agree with @KungFuJesus, Google Benchmark isn't really meant for testing the validity of the function only its speed. Because in some cases when you test the function for validity it can slow the benchmark. It is better for other things such as what we have for CI to test validity. As far as testing that each benchmark is faster than one or another, Google Benchmark doesn't appear to have anything with regard to that so I can only assume they didn't intend for it to be used like that. 

If we moved toward using Google Test, then testing for validity and things like that are something that would be nice to add to it.

nmoinvaz:
Please merge #1106 first, I will need to rebase this after that.

Dead2 (approved):
LGTM, after rebase 😉

nmoinvaz:
Rebased.

KungFuJesus (approved):

PR #1118, opened 2022-01-15
Change fuzzer binaries and solutions to prefix with fuzzer_.

nmoinvaz:
This commit is from #1097. I figured it would be easier to get it in if it were its own PR. It complements #1111 because all benchmark projects will begin with "benchmark_".


Dead2 (approved):
LGTM

PR #1109, opened 2022-01-10
Memcpy/memcmp macros

nmoinvaz:
* Define preprocessor macros `zmemcpy/zmemcmp` that force unaligned access on supported architectures
* Use `zmemcpy/zmemcmp` in the code where necessary.

PR #1119, opened 2022-01-15
Clean up and remove bestcmp_t type for longest_match. 

nmoinvaz:
Use 8 byte buffer (max int comparison size) and cast to integer type only when making comparison.

This PR is a commit from #1109. I figured it would be easier to get it in if it were its own PR. It makes the code easier to read in my opinion and I believe it removes the type-alias warnings we may have.


Dead2 (approved):
LGTM

PR #1120, opened 2022-01-15
Clean up crc32_fold structure and clearly define the size.

nmoinvaz:
This PR is a commit from #1096. I figured it would be easier to get it in if it were its own PR.


Dead2 (approved):
LGTM

PR #1124, opened 2022-01-18
Move UNALIGNED_OK detection to compile time instead of configure time.

nmoinvaz:
Pros: 
* It moves detection to one place instead of duplicating similar logic in both build scripts.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -51,10 +51,11 @@ jobs:
             codecov: ubuntu_gcc_osb
             cflags: -O1 -g3
 
-          - name: Ubuntu GCC -O3
+          - name: Ubuntu GCC -O3 No Unaligned
             os: ubuntu-latest
             compiler: gcc
             cxx-compiler: g++
+            cmake-args: -DWITH_UNALIGNED=OFF
             codecov: ubuntu_gcc_o3
             cflags: -O3
 
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -59,6 +59,14 @@ jobs:
             codecov: ubuntu_gcc_o3
             cflags: -O3
 
+          - name: Ubuntu GCC 32-bit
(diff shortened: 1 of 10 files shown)


nmoinvaz:
Rebased.

nmoinvaz:
@Dead2 thoughts on this one if I were to make it configurable again? Like have some kind of `NO_UNALIGNED` that is set in the build scripts? And is checked in the header before the `UNALIGNED_OK` determination..

nmoinvaz:
Rebased and added back configuration.

.github/workflows/cmake.yml
```diff
@@ -41,11 +41,11 @@ jobs:
             cmake-args: -DZLIB_COMPAT=ON -DZLIB_SYMBOL_PREFIX=zTest_
             codecov: ubuntu_gcc_compat_sprefix
 
-          - name: Ubuntu GCC OSB -O1 No Unaligned64 UBSAN
+          - name: Ubuntu GCC OSB -O1 UBSAN
             os: ubuntu-latest
             compiler: gcc
             cxx-compiler: g++
-            cmake-args: -DWITH_UNALIGNED=ON -DUNALIGNED64_OK=OFF -DWITH_SANITIZER=Undefined
+            cmake-args: -DWITH_SANITIZER=Undefined
```
Dead2, line 48: We should still run tests for this mixed case, as that is what an i386-i686 would run, so I guess that would mean adding a new test that compiles and runs i686 for example.

nmoinvaz: I thought we had such a case with mingw

nmoinvaz: I have added a new test for `-m32`, hopefully that should be sufficent.

Dead2 (approved):
LGTM, needs rebase though

nmoinvaz:
Rebased.

nmoinvaz:
Rebased again.

PR #1128, opened 2022-01-22
CPU features cleanup

nmoinvaz:
Clean up and moving arch feature includes out of zutil.h. This is good because zutil.h is included by many files and not all of them use cpu features.

### DIFF
--- a/cpu_features.h
+++ b/cpu_features.h
@@ -172,18 +172,18 @@ extern Pos quick_insert_string_acle(deflate_state *const s, const uint32_t str);
 
 /* slide_hash */
 #ifdef X86_SSE2
-void slide_hash_sse2(deflate_state *s);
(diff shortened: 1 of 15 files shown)


cpu_features.h
```diff
@@ -23,6 +23,7 @@
 extern void cpu_check_features();
 
 /* adler32 */
+typedef uint32_t (*adler32_func)(uint32_t adler, const unsigned char *buf, size_t len);
```
Dead2, line 26: Nitpicking here, but wouldn't it be better to have an empty line between the typedef and the big block of externs, just to set those a little apart to make it easier to read?

nmoinvaz: Fixed.

Dead2 (approved):

PR #1130, opened 2022-01-23
Fixed inadvertent breakage of CPUID usage

KungFuJesus:
In removing the AVX512 "well-suited" flag, a second cpuid call that's
needed for feature detection was accidently removed.  This brings that
back.


KungFuJesus:
@Dead2 I'm not sure I can create labels but this one should be "bug".

mtl1979 (approved):
LGTM

Dead2 (approved):

PR #1132, opened 2022-01-23
Move macros from zutil.h to zbuild.h

nmoinvaz:
This removes the dependency of zutil.h in several files. Why does it matter? Because zutil.h includes zlib.h. We should only include zlib.h or zlib-ng.h where it is absolutely necessary.

### DIFF
--- a/uncompr.c
+++ b/uncompr.c
@@ -3,7 +3,6 @@
  * For conditions of distribution and use, see copyright notice in zlib.h
  */
 
-#define Z_INTERNAL
 #include "zbuild.h"
 #ifdef ZLIB_COMPAT
 #  include "zlib.h"
--- a/zutil_p.h
+++ b/zutil_p.h
@@ -1,13 +1,9 @@
 /* zutil_p.h -- Private inline functions used internally in zlib-ng
- *
+ * For conditions of distribution and use, see copyright notice in zlib.h
  */
 
 #ifndef ZUTIL_P_H
 #define ZUTIL_P_H
-
-#if defined(HAVE_POSIX_MEMALIGN) && !defined(_POSIX_C_SOURCE)
-#  define _POSIX_C_SOURCE 200112L  /* For posix_memalign(). */
-#endif
 
 #if defined(__APPLE__) || defined(HAVE_POSIX_MEMALIGN)
 #  include <stdlib.h>
--- a/adler32.c
+++ b/adler32.c
@@ -4,7 +4,6 @@
  */
 
 #include "zbuild.h"
-#include "zutil.h"
 #include "functable.h"
 #include "adler32_p.h"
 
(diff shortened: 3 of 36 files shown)


nmoinvaz:
@dzenanz why am I receiving an error with the symbol prefix CI instance? I believe this has to do with the name mangling..

zutil_p.h
```diff
@@ -11,6 +11,7 @@
 
 #if defined(__APPLE__) || defined(HAVE_POSIX_MEMALIGN)
 #  include <stdlib.h>
+#  include <malloc.h>
```
dzenanz: I guess this was not needed, as it does not help. [Manual](https://man7.org/linux/man-pages/man3/posix_memalign.3.html) says we need `#include <stdlib.h>`, which is already included.

uncompr.c
```diff
@@ -3,7 +3,6 @@
  * For conditions of distribution and use, see copyright notice in zlib.h
  */
 
-#define Z_INTERNAL
```
dzenanz, line 6: Could removal of this definition be causing the problem?

dzenanz (commented):
By looking at the code changes, I don't know why CI is failing. I will try building locally.

Did something update in the CI infrastructure? Uh, no. The other PRs are green.

mtl1979:
We need `zlib.h` or `zlib-ng.h` for name mangling to work correctly. The name mangling header is included from either `zconf.h` or `zconf-ng.h` and doesn't get included from any other headers.

nmoinvaz:
That is partially why I did this PR. Because not every file needs the name mangling and all these macros don't require it.

mtl1979:
@nmoinvaz It seems some files like `inftrees.c` require name mangling, but the header doesn't get included....

uncompr.c
KungFuJesus: Z_INTERNAL decorates a bunch of internal functions.  Yes, removing that would cause a whole host of issues, including breaking compilation everywhere it's used.  I'm guessing that define was needed for some includes under it.

nmoinvaz:
Rebased and attempted to fix inftrees.c issue.

dzenanz:
> attempted to fix inftrees.c issue

You beat me to this. But I don't see `inftrees` mentioned anywhere in the diff?

mtl1979:
> > attempted to fix inftrees.c issue
> 
> You beat me to this. But I don't see `inftrees` mentioned anywhere in the diff?

The issue was that changing `inftrees.c` caused the name mangling header to not always get included and that resulted in different symbol names when declaring and using the functions the file contains.

dzenanz:
Oh, the [first force-push](https://github.com/zlib-ng/zlib-ng/compare/ad18ebafdb211569264011c495ae8ecfc626d283..328b297112f6051405d16fd39ae8ab71e0810045) did that. The second one was just a rebase.

Dead2 (approved):
LGTM

PR #1134, opened 2022-01-23
Use add_compile_options in cmake which works with both C & C++ builds.

nmoinvaz:
Some strings had to be removed because it treats a single string with multiple options as a single compiler option which confuses the compiler.


nmoinvaz:
This one may be dependent upon #1135.

nmoinvaz:
Now #1139.

PR #1135, opened 2022-01-23
Use add_link_options in CMake projects.

nmoinvaz:
CMake 3.14 is required for `add_link_options`, so throw error for code coverage if using CMake less than 3.14. This also makes the CMake code less branchy. Using `add_link_options` should help with getting C++ linking working better. I encountered many issues when adding Google Test (C++) trying to compile sanitizers and it not working unless using `add_link_options`.


nmoinvaz:
Now #1139.

PR #1126, opened 2022-01-21
Implement Google Test framework

nmoinvaz:
* CMake now uses Google Test instead of `example`.
* Configure/nmake no longer test `example`, but just use simple `minigzip` tests. This is because fetching Google Test repository might add more complexity than it is worth.
* This resolves #1017 by testing each Adler32 and Crc32 variant available on the host platform.

image

PR #1136, opened 2022-01-23
Update CMake toolchains for alternative gcc/g++ names.

nmoinvaz:
Remove compiler attribute in GHA when using CMake toolchain files.

Sometimes g++ compiler has g++ prefix or other times it has g++ suffix.


Dead2:
Looks like a good change to me, but I have not tested any of it myself. I see that CI passes though. 👍

nmoinvaz:
This PR is needed for #1126.

PR #1139, opened 2022-01-24
Use add_compile_options and add_link_options in CMake

nmoinvaz:
* We use `add_compile_options` so we don't have to individually set C and CXX flags. (Necessary for Google Test)
* Also added `add_link_options` with a fallback in < CMake 3.14.
* Now we send `-fsanitize` argument using `add_link_options` because `add_compile_options` does not do it automatically.


nmoinvaz:
@dankegel any clue as to why the _CI Pkgcheck - macOS Clang Native_  build is failing? I can't quite figure out what is going wrong.

nmoinvaz:
@dankegel nevermind I figured it out. It seems we were not setting native flags when doing intrinsics checks in cmake.

nmoinvaz:
This PR is needed for #1126.

nmoinvaz:
Rebased. I'm not sure why Windows GCC is failing.

KungFuJesus:
> Rebased. I'm not sure why Windows GCC is failing.

It appears to be segfaulting on the adler test?

```
6: Test command: D:\a\zlib-ng\zlib-ng\test_inflate_adler32.exe
6: Test timeout computed to be: 10000000
  1/686 Test   #1: test_adler32 ...............................................................................................................***Exception: SegFault  0.06 sec
```

Just curious, whole program optimization isn't turned on, is it?  I think that's the equivalent in windows to link time optimization.  If so, it's possible that something like AVX512 instructions could end up leaking into the bits of the binary that are supposed to be agnostic to it and you can get an illegal instruction.  Though, a signal from that segfault would certainly be helpful in making that determination.  Errr, well, I guess the -flto flag would be what I'm thinking of since this is GCC.

I noticed you have flags for PGO on and you've enabled -march=native.  It would seem strange that this would create issues but...maybe?

KungFuJesus:
Oh! So some older versions of mingw had bugs with regard to avx512.  They had some weird issue with exception handling and saving the zmm register states.  See: https://mails.dpdk.org/archives/dev/2021-February/199057.html
It may make sense to either 1.) do a compile time check on the mingw version or 2.) disable it on mingw for windows entirely?

KungFuJesus:
Although...the version of gcc it claims to be is _fairly_ new.  I find it interesting the "no optim" version is fine.  A stack trace with that segfault would be very helpful.

@nmoinvaz You're a Windows user, can you reproduce this with GCC?

So I'm not _convinced_ it's specific to AVX512, though, it is also segfaulting on the CRC32 check:

```
test 12
        Start  12: makecrct-crc32-combine

12: Test command: "C:\Program Files\CMake\bin\cmake.exe" "-DCOMMAND=D:/a/zlib-ng/zlib-ng/makecrct.exe;-c" "-DOUTPUT=D:/a/zlib-ng/zlib-ng/crc32_comb_tbl._h" "-DCOMPARE=D:/a/zlib-ng/zlib-ng/crc32_comb_tbl.h" "-DIGNORE_LINE_ENDINGS=ON" "-P" "D:/a/zlib-ng/zlib-ng/cmake/run-and-compare.cmake"
12: Test timeout computed to be: 10000000
  7/686 Test   #7: example ....................................................................................................................***Exception: SegFault  0.03 sec
```
Not sure that uses VPCLMULQDQ or not (if it's bare crc32, it probably doesn't).  That kind of points to something a little less specific.

nmoinvaz:
I'm seeing a lot of these errors in the log:
```
Data file mismatch - some data files may have been concurrently updated without locking support
```
Also turning off code coverage appears to make that CI work, so it indicates something is wrong with coverage support.

KungFuJesus:
> I'm seeing a lot of these errors in the log:
> 
> ```
> Data file mismatch - some data files may have been concurrently updated without locking support
> ```
> 
> Also turning off code coverage appears to make that CI work, so it indicates something is wrong with coverage support.

Hmm, that's...odd.  At least the code is off the hook, I guess.  One thing I noticed with cmake a few days ago - it doesn't effectively append -O3 for some reason.  Are we explicitly not allowing that is that CMake being CMake?

nmoinvaz:
There is some code in CMake that prevents -O3 in MSVC yes and converts it to -O2. I forget the reason why.

KungFuJesus:
> There is some code in CMake that prevents -O3 in MSVC yes and converts it to -O2. I forget the reason why.

Probably to disable auto vectorization, I imagine. But there are a lot of places in inflate alone that can auto vectorize. And BMI2 for all the shift without effecting flags helps a lot there. I did notice a lot of loops were unrolling quite a bit more when I forced -O3, hah. Speaking of which...I did get a faster avx512 compare with a fully unrolled loop for longer strings.

nmoinvaz:
Here is the commit:
https://github.com/zlib-ng/zlib-ng/commit/57b5617f14ce63e70bd8101f2a64d711a74183ae
It was probably done so that `configure` and `CMake` produce the same compiler flags, but CMake tries to do its own thing.

dankegel:
Landed in my spam folder :-(

Congrats on the move to gtest!

PR #1142, opened 2022-01-24
[ARM] Use armv8-crc+simd when compiling ACLE code on toolchains that don't enable FPU by default.

### LINKED ISSUE #1138: arm-linux-musleabihf build failed: cc1: error: ‘-mfloat-abi=hard’: selected architecture lacks an FPU
When building with `arm-linux-musleabihf`, it failed:

```sh
$ CHOST=arm-linux-musleabi ./configure --prefix="/cross_root/arm-linux-musleabi" --static --zlib-compat
$ make
....

arm-linux-musleabihf-gcc -O2 -s -std=c99 -Wall -D_LARGEFILE64_SOURCE=1 -DHAVE_POSIX_MEMALIGN -DZLIB_COMPAT -DWITH_GZFILEOP -DHAVE_VISIBILITY_HIDDEN -DHAVE_VISIBILITY_INTERNAL -DHAVE_BUILTIN_CTZ -DHAVE_BUILTIN_CTZLL -DARM_FEATURES -mfloat-abi=hard -DUNALIGNED_OK -DUNALIGNED64_OK -DARM_ACLE_CRC_HASH -DARM_NEON_ADLER32 -DARM_NEON_CHUNKSET -DARM_NEON_SLIDEHASH -I/usr/src/zlib-ng-2.0.6/arch/arm -I/usr/src/zlib-ng-2.0.6 -c -o armfeature.o /usr/src/zlib-ng-2.0.6/arch/arm/armfeature.c
make -C arch/arm crc32_acle.o
make[1]: Entering directory '/usr/src/zlib-ng-2.0.6/arch/arm'
arm-linux-musleabihf-gcc -O2 -s -std=c99 -Wall -D_LARGEFILE64_SOURCE=1 -DHAVE_POSIX_MEMALIGN -DZLIB_COMPAT -DWITH_GZFILEOP -DHAVE_VISIBILITY_HIDDEN -DHAVE_VISIBILITY_INTERNAL -DHAVE_BUILTIN_CTZ -DHAVE_BUILTIN_CTZLL -DARM_FEATURES -mfloat-abi=hard -DUNALIGNED_OK -DUNALIGNED64_OK -DARM_ACLE_CRC_HASH -DARM_NEON_ADLER32 -DARM_NEON_CHUNKSET -DARM_NEON_SLIDEHASH -march=armv8-a+crc -fno-lto -I/usr/src/zlib-ng-2.0.6/arch/arm -I/usr/src/zlib-ng-2.0.6 -c -o crc32_acle.o /usr/src/zlib-ng-2.0.6/arch/arm/crc32_acle.c
cc1: error: ‘-mfloat-abi=hard’: selected architecture lacks an FPU
```

What's wrong ?

mtl1979:
Fixes #1138.


KungFuJesus:
Does the CMakeList not need a corresponding change?

mtl1979:
@KungFuJesus When cross-compiling, target CPU is specified in cmake toolchain file. It would not be possible to test MUSL in non-cross configuration using cmake.

configure
```diff
@@ -1129,6 +1129,30 @@ EOF
     fi
 }
 
+
+check_acle_intrinsics() {
```
nmoinvaz: This function doesn't really check for ACLE intrinsics. It checks for availability of ACLE architecture compiler flags. So maybe rename `check_acle_compiler_flag`?

nmoinvaz:
> When cross-compiling, target CPU is specified in cmake toolchain file. It would not be possible to test MUSL in non-cross configuration using cmake.

But it is possible to use CMake without a toolchain file. By default it will use CC and CXX environment variables, etc.

mtl1979:
@nmoinvaz Yes, it's possible to use cmake without toolchain file, but the result is unpredictable, and often the build fails due to either mixing files/executables from two different toolchains or targeting wrong CPU.

configure
mtl1979: Neither does `check_neon_intrinsics()` in configure... We would need to rename both and that would require another PR.

configure
```diff
@@ -1129,6 +1129,30 @@ EOF
...
+    # Check whether -march=armv8-a+crc works correctly
+    cat > $test.c << EOF
+int main() { return 0; }
+EOF
+    if try $CC -c $CFLAGS -march=armv8-a+crc $test.c; then
+        ACLE_AVAILABLE=1
+        echo "Check whether -march=armv8-a+crc works ... Yes." | tee -a configure.log
```
nmoinvaz, line 1139: Consider using same language as `check_neon_intrinsics`..
```
Check whether -march=armv8-a+crc is available ... Yes."
```

configure
nmoinvaz: Why not fix your commit and then the other PR can fix the other one?

nmoinvaz: You could even add another commit to this PR that fixed it. I think it would get accepted.

configure
mtl1979: It is available, but unusable... It's matter of two compiler flags being mutually exclusive...

mtl1979:
@nmoinvaz I tried cmake without toolchain file, it targets "armv5" and can't be overridden on command-line.

configure
mtl1979: I'd rather keep "bug fixes" and cleanup commits separate... Makes it a lot easier to review...

mtl1979:
Added cmake version if someone wants to try compiling natively on system that use MUSL by default... Tried cross-compile with toolchain file but qemu-arm version I have can't load the binaries.

cmake/detect-intrinsics.cmake
```diff
@@ -2,14 +2,24 @@
 # Licensed under the Zlib license, see LICENSE.md for details
 
 macro(check_acle_intrinsics)
-    if(NOT NATIVEFLAG)
-        set(ACLEFLAG "-march=armv8-a+crc")
+    if(NOT NATIVEFLAG AND NOT HAVE_ACLE_INTRIN)
+        set(ACLEFLAG "-march=armv8-a+crc" CACHE INTERNAL "Compiler option to enable ACLE support")
     endif()
```
mtl1979: We do not want to overwrite the cached variable, if the previous test run succeeded.

cmake/detect-intrinsics.cmake
```diff
@@ -2,14 +2,24 @@
...
         HAVE_ACLE_INTRIN FAIL_REGEX "not supported")
+    if(NOT NATIVEFLAG AND NOT HAVE_ACLE_INTRIN)
+        set(ACLEFLAG "-march=armv8-a+crc+simd" CACHE INTERNAL "Compiler option to enable ACLE support" FORCE)
+        # Check whether compiler supports ACLE flag
+        set(CMAKE_REQUIRED_FLAGS "${ACLEFLAG}")
+        check_c_source_compiles(
+            "int main() { return 0; }"
+            HAVE_ACLE_INTRIN2 FAIL_REGEX "not supported")
+        set(HAVE_ACLE_INTRIN ${HAVE_ACLE_INTRIN2} CACHE INTERNAL "Have ACLE intrinsics" FORCE)
+        unset(HAVE_ACLE_INTRIN2 CACHE) # Don't cache this internal variable
```
mtl1979: We need to use separate test variable, otherwise cmake use the cached result from test above. We short-circuit the next reconfigure step to use the cached `ACLEFLAG`, to avoid testing both possibilities again. Clearing `CMakeCache.txt` will force cmake to re-evaluate both tests.

cmake/detect-intrinsics.cmake
```diff
@@ -1,15 +1,25 @@
...
         "int main() { return 0; }"
         HAVE_ACLE_INTRIN FAIL_REGEX "not supported")
+    if(NOT NATIVEFLAG AND NOT HAVE_ACLE_INTRIN)
+        set(ACLEFLAG "-march=armv8-a+crc+simd" CACHE INTERNAL "Compiler option to enable ACLE support" FORCE)
+        # Check whether compiler supports ACLE flag
+        set(CMAKE_REQUIRED_FLAGS "${ACLEFLAG}")
+        check_c_source_compiles(
+            "int main() { return 0; }"
+            HAVE_ACLE_INTRIN2 FAIL_REGEX "not supported")
+        set(HAVE_ACLE_INTRIN ${HAVE_ACLE_INTRIN2} CACHE INTERNAL "Have ACLE intrinsics" FORCE)
```
nmoinvaz: Why storing value in `HAVE_ACLE_INTRIN2`?

cmake/detect-intrinsics.cmake
```diff
@@ -1,15 +1,25 @@
...
+    if(NOT NATIVEFLAG AND NOT HAVE_ACLE_INTRIN)
+        set(ACLEFLAG "-march=armv8-a+crc" CACHE INTERNAL "Compiler option to enable ACLE support")
     endif()
     # Check whether compiler supports ACLE flag
     set(CMAKE_REQUIRED_FLAGS "${ACLEFLAG}")
     check_c_source_compiles(
         "int main() { return 0; }"
         HAVE_ACLE_INTRIN FAIL_REGEX "not supported")
+    if(NOT NATIVEFLAG AND NOT HAVE_ACLE_INTRIN)
+        set(ACLEFLAG "-march=armv8-a+crc+simd" CACHE INTERNAL "Compiler option to enable ACLE support" FORCE)
```
nmoinvaz, line 14: Is this so `ACLEFLAG` can be overwritten in the command line or for something else?

cmake/detect-intrinsics.cmake
```diff
@@ -1,15 +1,25 @@
 # detect-intrinsics.cmake -- Detect compiler intrinsics support
 # Licensed under the Zlib license, see LICENSE.md for details
 
-macro(check_acle_intrinsics)
-    if(NOT NATIVEFLAG)
-        set(ACLEFLAG "-march=armv8-a+crc")
+macro(check_acle_compiler_flag)
+    if(NOT NATIVEFLAG AND NOT HAVE_ACLE_INTRIN)
```
nmoinvaz: Why would `HAVE_ACLE_INTRIN` already be set?

cmake/detect-intrinsics.cmake
mtl1979: Because it's cached variable... Without using unique name, the second test would never get executed.

cmake/detect-intrinsics.cmake
mtl1979: It's to allow cmake reconfigure itself if `CMakeLists.txt` gets updated between builds.

cmake/detect-intrinsics.cmake
mtl1979: Because it's cached value... We need to allow cmake to reconfigure itself if `CMakeLists.txt` is updated between builds.

configure
```diff
@@ -1129,7 +1129,31 @@ EOF
     fi
 }
 
-check_neon_intrinsics() {
+
+check_acle_compiler_flag() {
```
nmoinvaz, line 1132: I think there is an extra line above.

cmake/detect-intrinsics.cmake
```diff
@@ -1,15 +1,25 @@
...
     check_c_source_compiles(
         "int main() { return 0; }"
         HAVE_ACLE_INTRIN FAIL_REGEX "not supported")
+    if(NOT NATIVEFLAG AND NOT HAVE_ACLE_INTRIN)
+        set(ACLEFLAG "-march=armv8-a+crc+simd" CACHE INTERNAL "Compiler option to enable ACLE support" FORCE)
+        # Check whether compiler supports ACLE flag
+        set(CMAKE_REQUIRED_FLAGS "${ACLEFLAG}")
+        check_c_source_compiles(
+            "int main() { return 0; }"
+            HAVE_ACLE_INTRIN2 FAIL_REGEX "not supported")
```
nmoinvaz: Maybe it is better named variable would be `HAVE_ACLE_FLAG` since like before it is not checking intrinsics?

configure
nmoinvaz: Ok.

configure
mtl1979: Fixed.

cmake/detect-intrinsics.cmake
mtl1979: Changed in the cleanup commit...

Dead2 (approved):
LGTM

PR #1146, opened 2022-01-28
More than double adler32 performance with altivec

KungFuJesus:
Bits of low hanging and high hanging fruit in this round of
optimization.  Altivec has a sum characters into 4 lanes of integers
instructions (intrinsic vec_sum4s) that seems basically made for this
algorithm.  Additionally, there's a similar multiply-accumulate routine
that takes two character vectors for input and outputs a vector of 4
ints for their respective adjacent sums.  This alone was a good amount
of the performance gains.

Additionally, the shifting by 4 was still done in the loop when it was
easy to roll outside of the loop and do only once.  This removed some
latency for a dependent operand to be ready.  We also unrolled the loop
with independent sums, though, this only seems to help for much larger
input sizes.

Additionally, we reduced feeding the two 16 bit halves of the sum simply
by packing them into an aligned allocation in the stack next to each
other.  Then, when loaded, we permute and shift the values to two
separate vector registers from the same input registers.  The separation
of these scalars probably could have been done in vector registers
through some tricks but we need them in scalar GPRs anyhow every time
they leave the loop so it was naturally better to keep those separate
before hitting the vectorized code.

For the horizontal addition, the code was modified to use a sequence of
shifts and adds to produce a vector sum in the first lane.  Then, the
much cheaper vec_ste was used to store the value into a general purpose
register rather than vec_extract.

Lastly, instead of doing the relatively expensive modulus in GPRs after
we perform the scalar operations to align all of the loads in the loop,
we can instead reduce "n" here for the first round to be n minus the
alignment offset.

### DIFF
--- a/arch/power/adler32_vmx.c
+++ b/arch/power/adler32_vmx.c
@@ -1,6 +1,7 @@
 /* adler32_vmx.c -- compute the Adler-32 checksum of a data stream
  * Copyright (C) 1995-2011 Mark Adler
  * Copyright (C) 2017-2021 Mika T. Lindqvist <postmaster@raasu.org>
+ * Copyright (C) 2021 Adam Stylinski <kungfujesus06@gmail.com>
  * For conditions of distribution and use, see copyright notice in zlib.h
  */
 
@@ -11,16 +12,6 @@
 
 #define vmx_zero()  (vec_splat_u32(0))
 
-vector unsigned short vec_hadduh(vector unsigned char a) {
-    vector unsigned char vmx_one = vec_splat_u8(1);
-    return vec_add(vec_mulo(a, vmx_one), vec_mule(a, vmx_one));
-}
-
-vector unsigned int vec_hadduw(vector unsigned short a) {
-    vector unsigned short vmx_one = vec_splat_u16(1);
-    return vec_add(vec_mulo(a, vmx_one), vec_mule(a, vmx_one));
-}
-
 static inline void vmx_handle_head_or_tail(uint32_t *pair, const unsigned char *buf, size_t len) {
     unsigned int i;
     for (i = 0; i < len; ++i) {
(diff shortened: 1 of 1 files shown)


mtl1979:
Seems like qemu-ppc is not supporting VMX at all, so the coverage is 0%... I might need to dig my old Ubuntu/PPC and check if I can get some adler32 benchmark comparisons done...

KungFuJesus:
Please do.  I have half a mind to try to unroll this a bit further as everything I've read seems to indicate the G5 has a pretty deeply staged pipeline and to maximize the throughput I'm going to need to pop some more data dependency bubbles.  But it could ultimately be that the dependent loads from memory are going to be the limiting factor.  At the very least I can unroll to the weird cache line size of the G5 (128 bytes).

KungFuJesus:
```
Comparing adler32/vmx (from ./benchmark_zlib_old) to adler32/vmx (from ./benchmark_zlib)
Benchmark                                                      Time             CPU      Time Old      Time New       CPU Old       CPU New
-------------------------------------------------------------------------------------------------------------------------------------------
[adler32/vmx vs. adler32/vmx]/2048_mean                     -0.6817         -0.6817          1159           369          1159           369
[adler32/vmx vs. adler32/vmx]/2048_median                   -0.6816         -0.6816          1159           369          1159           369
[adler32/vmx vs. adler32/vmx]/2048_stddev                   -0.1940         +0.7143             0             0             0             0
[adler32/vmx vs. adler32/vmx]/4096_mean                     -0.6927         -0.6926          2286           703          2286           703
[adler32/vmx vs. adler32/vmx]/4096_median                   -0.6926         -0.6926          2286           703          2286           703
[adler32/vmx vs. adler32/vmx]/4096_stddev                   -0.7952         -0.7431             0             0             0             0
[adler32/vmx vs. adler32/vmx]/32768_mean                    -0.6953         -0.6953         18157          5532         18156          5532
[adler32/vmx vs. adler32/vmx]/32768_median                  -0.6953         -0.6953         18156          5532         18156          5532
[adler32/vmx vs. adler32/vmx]/32768_stddev                  -0.7204         -0.4717             2             1             1             0
[adler32/vmx vs. adler32/vmx]/262144_mean                   -0.6961         -0.6961        145631         44257        145622         44256
[adler32/vmx vs. adler32/vmx]/262144_median                 -0.6961         -0.6961        145629         44257        145620         44256
[adler32/vmx vs. adler32/vmx]/262144_stddev                 -0.8199         -0.7243            13             2             8             2
[adler32/vmx vs. adler32/vmx]/2097152_mean                  -0.5188         -0.5188       1290244        620873       1290164        620858
[adler32/vmx vs. adler32/vmx]/2097152_median                -0.5187         -0.5187       1290254        620964       1290157        620952
[adler32/vmx vs. adler32/vmx]/2097152_stddev                +0.2937         +0.6945           209           271           158           268
[adler32/vmx vs. adler32/vmx]/4194304_mean                  -0.5028         -0.5028       2588475       1286875       2588345       1286849
[adler32/vmx vs. adler32/vmx]/4194304_median                -0.5028         -0.5028       2588463       1286932       2588365       1286905
[adler32/vmx vs. adler32/vmx]/4194304_stddev                -0.3738         -0.1664           538           337           405           337
```

KungFuJesus:
Why is this derping on the buildkite CI test for multiple days?  "Pending"?

nmoinvaz:
That has happened before, usually @Dead2 has to restart something on the box..

Dead2:
Ohh, I had partly forgotten about that. Lets just ignore that test for now.

We moved the last of the servers out of my jobs server room on Friday.
That specific server (Snipp) is not going to come back up (Dual Xeon Westmere is more power hungry than I want to be hosting from home on my own money, especially after power prices went up 5-10x the past 4 months)..
I do have a haswell machine that could be used perhaps, but that will probably take some time.

Btw, I just spent ~6 hours setting up an armv6l machine for our testing purposes, that and the aarch64 one might possibly be added to CI at some point. They are pretty slow though, so probably not good as complete replacement for the ones currently running CI via qemu. They are based on RPI 1B and RPI 3, would have been nice to have two RPI 4 (or better) to use instead, but we will have to make do.

mtl1979:
Coverage data is still inconsistent... Might need rebase on top of current `develop` tip to update...

mtl1979 (approved):
LGTM.

Dead2 (approved):
LGTM, but I have no way to test this

KungFuJesus:
In theory, anyway, CI should be testing it.  For benchmarking the gains - it sounds like @mtl1979 has a set of PPC hardware that he can test with if you need a second data point.

mtl1979:
When I wrote the original vectorized version, I was still learning the instruction set and the other intricacies of CPU internals. My highest priority was keeping the code as readable as possible without deviating too much from the original code. That was to attract more experienced developers to improve the code.

KungFuJesus:
> When I wrote the original vectorized version, I was still learning the instruction set and the other intricacies of CPU internals. My highest priority was keeping the code as readable as possible without deviating too much from the original code. That was to attract more experienced developers to improve the code.

I get it - my first cut of Altivec code does some suboptimal things around unaligned loads when I first learned.  I'm suggesting you can benchmark this PR with your hardware so that we can get a second data point.

mtl1979:
> I get it - my first cut of Altivec code does some suboptimal things around unaligned loads when I first learned. I'm suggesting you can benchmark this PR with your hardware so that we can get a second data point.

I just need to make some room so I can connect everything again... Doesn't really help that I have two cats here jumping around, not forgetting all the other projects I'm working on currently.

mtl1979:
Building benchmark_zlib requires more recent cmake version than my 32-bit Ubuntu (16.10) has... There is no later release for 32-bit PowerPC.

mtl1979:
```
before:

--------------------------------------------------------------
Benchmark                    Time             CPU   Iterations
--------------------------------------------------------------
adler32/vmx/2048          6720 ns         6717 ns       105787
adler32/vmx/4096         13358 ns        13345 ns        52555
adler32/vmx/32768       105754 ns       105705 ns         6590
adler32/vmx/262144      848151 ns       846999 ns          821
adler32/vmx/2097152    6995895 ns      6990538 ns          100
adler32/vmx/4194304   14016579 ns     13949114 ns           50

after:

--------------------------------------------------------------
Benchmark                    Time             CPU   Iterations
--------------------------------------------------------------
adler32/vmx/2048          2963 ns         2959 ns       236459
adler32/vmx/4096          5853 ns         5839 ns       119032
adler32/vmx/32768        45709 ns        45612 ns        15529
adler32/vmx/262144      363820 ns       363571 ns         1930
adler32/vmx/2097152    3110848 ns      3108457 ns          227
adler32/vmx/4194304    6199476 ns      6194799 ns          112
```

mscastanho:
> There is no later release for 32-bit PowerPC.

@mtl1979 Debian is one of the only distros that still support 32-bit PowerPC, but only in the unstable branch (Debian sid) [Link](http://deb.debian.org/debian-ports/dists/sid/).

KungFuJesus:
My bet is Gentoo is one of the others :).  I think there's Adelie as well.  PPC64 big endian support is probably just as hard to come by, I imagine.

mtl1979:
> > There is no later release for 32-bit PowerPC.
> 
> @mtl1979 Debian is one of the only distros that still support 32-bit PowerPC, but only in the unstable branch (Debian sid) [Link](http://deb.debian.org/debian-ports/dists/sid/).

I'm using Ubuntu 16.10 and Debian 10 (I had Debian 8 originally, but it was too old)... Debian installer doesn't work on my setup, so I have to manually install all the core packages one by one using dpkg and that takes almost a day... After I have booted Debian once, apt works. Seems like they forgot to add gnupg to the installer image and that caused apt to fail downloading anything that isn't on the network installation CD.

KungFuJesus:
It is curious that @mtl1979's CPU only received a 2.2ishx boost when fitting into cache better whereas the G5 did better (something like 3.2x in some cases).  I wonder if it happens to be pipelining better.

mtl1979:
@KungFuJesus I was using pretty new gcc version, which is kinda counterproductive with old processor. Older versions of gcc did choke on compiling VSX code and I didn't initially use `WITH_POWER8=OFF` with cmake.

KungFuJesus:
> @KungFuJesus I was using pretty new gcc version, which is kinda counterproductive with old processor. Older versions of gcc did choke on compiling VSX code and I didn't initially use `WITH_POWER8=OFF` with cmake.

I was using gcc 10.3, I think the differences were strictly in the pipeline differences between the ppc970 and the 7400 or 7450.

mtl1979:
> I was using gcc 10.3, I think the differences were strictly in the pipeline differences between the ppc970 and the 7400 or 7450.

Not a competition -- lol --, but I was using gcc 11.2.0 without changing any tuning parameters.

KungFuJesus:
Hmm, well I've been updating the Gentoo installation for about a week, I have a newer gcc I can try to see if we have a regression. GCC has in the past regressed the instruction scheduling to the point where packages like ATLAS specifically remove optimization stages of instruction scheduling, so it's not impossible. Hope it's not true, though.

mtl1979:
@KungFuJesus Comparing to the benchmarks I did when I initially wrote the VMX version did show bigger difference between C and VMX implementations of adler32... With gcc 11, the difference is quite small... There is so many factors to consider, that usually I don't do benchmarks at all... I just test that the code works and that is good enough for me...

PR #1151, opened 2022-02-01
[Benchmarks] Fix adler32/vmx benchmark not found under 32-bit PowerPC

mtl1979:
If `WITH_POWER8` is set, VMX adler32 benchmark isn't enabled even if current CPU doesn't support VSX.

### DIFF
--- a/test/benchmarks/benchmark_adler32.cc
+++ b/test/benchmarks/benchmark_adler32.cc
@@ -69,10 +69,13 @@ BENCHMARK_ADLER32(c, adler32_c, 1);
 
 #ifdef ARM_NEON_ADLER32
 BENCHMARK_ADLER32(neon, adler32_neon, arm_cpu_has_neon);
-#elif defined(POWER8_VSX_ADLER32)
-BENCHMARK_ADLER32(power8, adler32_power8, power_cpu_has_arch_2_07);
-#elif defined(PPC_VMX_ADLER32)
+#endif
+
+#ifdef PPC_VMX_ADLER32
 BENCHMARK_ADLER32(vmx, adler32_vmx, power_cpu_has_altivec);
+#endif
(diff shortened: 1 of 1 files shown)


KungFuJesus:
I'm not sure what happened with the Windows GCC No Opt CI

mtl1979:
Those Windows CI checks have been flaky since early days...

test/benchmarks/benchmark_adler32.cc
```diff
@@ -69,11 +69,14 @@ BENCHMARK_ADLER32(c, adler32_c, 1);
 
 #ifdef ARM_NEON_ADLER32
 BENCHMARK_ADLER32(neon, adler32_neon, arm_cpu_has_neon);
-#elif defined(POWER8_VSX_ADLER32)
-BENCHMARK_ADLER32(power8, adler32_power8, power_cpu_has_arch_2_07);
-#elif defined(PPC_VMX_ADLER32)
+#endif
+
+#if defined(PPC_VMX_ADLER32)
```
nmoinvaz: Perhaps use `#ifdef` instead of `#if defined(..)` in both places.. ?

nmoinvaz (approved):

test/benchmarks/benchmark_adler32.cc
mtl1979: Done.

nmoinvaz:
LGTM

nmoinvaz:
Btw, it sounds like you were able to get cmake working on your PPC?

mtl1979:
> Btw, it sounds like you were able to get cmake working on your PPC?

Yeah... Had to upgrade from Debian 8 to Debian 10... The installer had some issues, but after few hours I managed to manually install enough packages that it would boot...

mtl1979:
@Dead2

PR #1154, opened 2022-02-04
Added power8 cpu flag to powerpc64 CMake toolchain files.

nmoinvaz:


mtl1979 (approved):
LGTM... PowerPC/Power versions of gcc don't have sane default for `-cpu`. Explicitly setting oldest compatible CPU allows us to catch any errors caused by mixing instructions only supported by newer CPUs and still allow consistently using instructions we need.

Dead2 (approved):

PR #1155, opened 2022-02-04
Clean up header includes

nmoinvaz:
zbuild.h already includes stdint.h so no need to reinclude it.

### DIFF
--- a/cpu_features.h
+++ b/cpu_features.h
@@ -6,7 +6,6 @@
 #ifndef CPU_FEATURES
 #define CPU_FEATURES
 
-#include "deflate.h"
 #include "crc32_fold.h"
 
 #if defined(X86_FEATURES)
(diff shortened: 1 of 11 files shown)


cpu_features.h
```diff
@@ -128,6 +127,7 @@ extern uint32_t compare256_unaligned_avx2(const uint8_t *src0, const uint8_t *sr
 #endif
 #endif
 
+#ifdef DEFLATE_H_
```
mtl1979, line 130: Instead of conditionally including these lines, we might want to move them to another header. This makes sure everything works even if people include the headers in wrong order.

nmoinvaz: That kind of refactoring is probably best left for a future PR. Right now however, I want to be able to include cpu_features.h without having to include deflate.h which includes zlib.h and zlib name mangling...

mtl1979: I was just mentioning for future reference... I really don't want us to introduce a compiler warning if the headers are included in backwards order. That would cause a lot of unnecessary noise in compiler output even if it would make obvious that the order of includes must be correct for consistent output.

Dead2 (approved):
LGTM

PR #1156, opened 2022-02-05
Rename CPU feature header and source files for consistency.

nmoinvaz:
ARM has a weird `armfeature.c` so to avoid all conflicts just rename everything with `features` suffix..


Dead2 (approved):
LGTM

PR #1160, opened 2022-02-07
Update NMake GitHub Actions to use Visual Studio 2022 Enterprise.

mtl1979:
It makes more sense to explicitly specify Visual Studio version, so we don't need to constantly check if the runner has correct version installed.


nmoinvaz (approved):

mtl1979:
@Dead2

PR #1161, opened 2022-02-08
Use pclmulqdq accelerated CRC for exported function

KungFuJesus:
We were already using this internally for our CRC calculations, however
the exported function to CRC checksum any arbitrary stream of bytes was
still using a generic C based version that leveraged tables. This
function is now called when len is at least 64 bytes.

### DIFF
--- a/arch/x86/crc32_fold_pclmulqdq.c
+++ b/arch/x86/crc32_fold_pclmulqdq.c
@@ -6,6 +6,7 @@
  * https://www.intel.com/content/dam/www/public/us/en/documents/white-papers/fast-crc-computation-generic-polynomials-pclmulqdq-paper.pdf
  *
  * Copyright (C) 2013 Intel Corporation. All rights reserved.
+ * Copyright (C) 2016 Marian Beermann (support for initial value)
  * Authors:
  *     Wajdi Feghali   <wajdi.k.feghali@intel.com>
  *     Jim Guilford    <james.guilford@intel.com>
@@ -24,12 +25,18 @@
 #include <smmintrin.h> // _mm_extract_epi32
 
 #include "x86_features.h"
+#include "cpu_features.h"
 
 #include "../../crc32_fold.h"
+#include "../../crc32_p.h"
+#include <assert.h>
 
 #ifdef X86_VPCLMULQDQ_CRC
 extern size_t fold_16_vpclmulqdq(__m128i *xmm_crc0, __m128i *xmm_crc1,
     __m128i *xmm_crc2, __m128i *xmm_crc3, uint8_t *dst, const uint8_t *src, size_t len);
+extern size_t fold_16_vpclmulqdq_nocp(__m128i *xmm_crc0, __m128i *xmm_crc1,
+    __m128i *xmm_crc2, __m128i *xmm_crc3, const uint8_t *src, size_t len, __m128i init_crc,
+    int32_t first);
 #endif
 
 static void fold_1(__m128i *xmm_crc0, __m128i *xmm_crc1, __m128i *xmm_crc2, __m128i *xmm_crc3) {
@@ -380,6 +387,124 @@ done:
     crc32_fold_save_partial((__m128i *)crc->fold, xmm_crc_part);
 }
 
+#define ONCE(op) if (first) { \
+    first = 0; \
+    (op); \
+}
+#define XOR_INITIAL(where) ONCE(where = _mm_xor_si128(where, xmm_initial))
+
+Z_INTERNAL void crc32_fold_pclmulqdq(crc32_fold *crc, const uint8_t *src, size_t len, uint32_t init_crc) {
+    unsigned long algn_diff;
+    __m128i xmm_t0, xmm_t1, xmm_t2, xmm_t3;
+    __m128i xmm_crc0, xmm_crc1, xmm_crc2, xmm_crc3, xmm_crc_part;
+    __m128i xmm_initial = _mm_cvtsi32_si128(init_crc);
+    int32_t first = 1;
+
+    /* Technically the CRC functions don't even call this for input < 64, but a bare minimum of 31
+     * bytes of input is needed for the aligning load that occurs.  If there's an initial CRC, to 
+     * carry it forward through the folded CRC there must be 16 - src % 16 + 16 bytes available, which
+     * by definition can be up to 15 bytes + one full vector load. */
+    assert(len >= 31);
+    crc32_fold_load((__m128i *)crc->fold, &xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3);
+
+    algn_diff = ((uintptr_t)16 - ((uintptr_t)src & 0xF)) & 0xF;
+    if (algn_diff) {
+        if (algn_diff >= 4) {
+            xmm_crc_part = _mm_loadu_si128((__m128i *)src);
+
+            src += algn_diff;
+            len -= algn_diff;
+
+            XOR_INITIAL(xmm_crc_part);
+            partial_fold(algn_diff, &xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3, &xmm_crc_part);
+        } else {
+            xmm_t0 = _mm_loadu_si128((__m128i*)src);
+            xmm_crc_part = _mm_loadu_si128((__m128i*)src + 1);
+            XOR_INITIAL(xmm_t0);
+            fold_1(&xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3);
+            xmm_crc3 = _mm_xor_si128(xmm_crc3, xmm_t0);
+            partial_fold(algn_diff, &xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3, &xmm_crc_part);
+
+            src += (algn_diff + 16);
+            len -= (algn_diff + 16);
+        }
+    }
+
+    xmm_crc_part = _mm_setzero_si128();
+
+#ifdef X86_VPCLMULQDQ_CRC
+    if (x86_cpu_has_vpclmulqdq && x86_cpu_has_avx512 && (len >= 256)) {
+        size_t n = fold_16_vpclmulqdq_nocp(&xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3, src, len,
+                xmm_initial, first);
+        first = 0;
+
+        len -= n;
+        src += n;
+    }
+#endif
+
+    while (len >= 64) {
+        crc32_fold_load((__m128i *)src, &xmm_t0, &xmm_t1, &xmm_t2, &xmm_t3);
+        XOR_INITIAL(xmm_t0);
+        fold_4(&xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3);
+
+        xmm_crc0 = _mm_xor_si128(xmm_crc0, xmm_t0);
+        xmm_crc1 = _mm_xor_si128(xmm_crc1, xmm_t1);
+        xmm_crc2 = _mm_xor_si128(xmm_crc2, xmm_t2);
+        xmm_crc3 = _mm_xor_si128(xmm_crc3, xmm_t3);
+
+        src += 64;
+        len -= 64;
+    }
+
+    /*
+     * len = num bytes left - 64
+     */
+    if (len >= 48) {
+        xmm_t0 = _mm_load_si128((__m128i *)src);
+        xmm_t1 = _mm_load_si128((__m128i *)src + 1);
+        xmm_t2 = _mm_load_si128((__m128i *)src + 2);
+        XOR_INITIAL(xmm_t0);
+
+        fold_3(&xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3);
+
+        xmm_crc1 = _mm_xor_si128(xmm_crc1, xmm_t0);
+        xmm_crc2 = _mm_xor_si128(xmm_crc2, xmm_t1);
+        xmm_crc3 = _mm_xor_si128(xmm_crc3, xmm_t2);
+        len -= 48;
+        src += 48; 
+    } else if (len >= 32) {
+        xmm_t0 = _mm_load_si128((__m128i *)src);
+        xmm_t1 = _mm_load_si128((__m128i *)src + 1);
+        XOR_INITIAL(xmm_t0);
+
+        fold_2(&xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3);
+
+        xmm_crc2 = _mm_xor_si128(xmm_crc2, xmm_t0);
+        xmm_crc3 = _mm_xor_si128(xmm_crc3, xmm_t1);
+
+        len -= 32;
+        src += 32;
+    } else if (len >= 16) {
+        xmm_t0 = _mm_load_si128((__m128i *)src);
+        XOR_INITIAL(xmm_t0);
+
+        fold_1(&xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3);
+
+        xmm_crc3 = _mm_xor_si128(xmm_crc3, xmm_t0);
+
+        len -= 16;
+        src += 16;
+    }
+
+    if (len) {
+        memcpy(&xmm_crc_part, src, len);
+        partial_fold((size_t)len, &xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3, &xmm_crc_part);
+    }
+
+    crc32_fold_save((__m128i *)crc->fold, xmm_crc0, xmm_crc1, xmm_crc2, xmm_crc3);
+}
+
 static const unsigned ALIGNED_(16) crc_k[] = {
     0xccaa009e, 0x00000000, /* rk1 */
     0x751997d0, 0x00000001, /* rk2 */
@@ -460,6 +585,18 @@ Z_INTERNAL uint32_t crc32_fold_final_pclmulqdq(crc32_fold *crc) {
     crc->value = ~((uint32_t)_mm_extract_epi32(xmm_crc3, 2));
 
     return crc->value;
+}
+
+uint32_t crc32_pclmulqdq(uint32_t crc32, const unsigned char* buf, uint64_t len) {
+    /* For lens < 64, crc32_byfour method is faster. The CRC32 instruction for
+     * these short lengths might also prove to be effective */
+    if (len < 64) 
+        return crc32_byfour(crc32, buf, len);
+
+    crc32_fold ALIGNED_(16) crc_state;
+    crc32_fold_reset_pclmulqdq(&crc_state);
+    crc32_fold_pclmulqdq(&crc_state, buf, len, crc32);
+    return crc32_fold_final_pclmulqdq(&crc_state);
 }
 
 #endif
--- a/arch/x86/crc32_fold_vpclmulqdq.c
+++ b/arch/x86/crc32_fold_vpclmulqdq.c
@@ -5,8 +5,15 @@
 
 #ifdef X86_VPCLMULQDQ_CRC
 #include "../../zbuild.h"
+#include "../../fallback_builtins.h"
 
 #include <immintrin.h>
+
+#define ONCE(op) if (first) { \
+    first = 0; \
+    (op); \
+}
+#define XOR_INITIAL(where) ONCE(where = _mm512_xor_si512(where, zmm_initial))
 
 size_t fold_16_vpclmulqdq(__m128i *xmm_crc0, __m128i *xmm_crc1,
     __m128i *xmm_crc2, __m128i *xmm_crc3, uint8_t *dst, const uint8_t *src, size_t len) {
@@ -80,6 +87,97 @@ size_t fold_16_vpclmulqdq(__m128i *xmm_crc0, __m128i *xmm_crc1,
         len -= 256;
         src += 256;
         dst += 256;
+    }
+    // zmm_crc[0,1,2,3] -> zmm_crc0
+    z0 = _mm512_clmulepi64_epi128(zmm_crc0, zmm_fold4, 0x01);
+    zmm_crc0 = _mm512_clmulepi64_epi128(zmm_crc0, zmm_fold4, 0x10);
+    zmm_crc0 = _mm512_xor_si512(z0, zmm_crc0);
+    zmm_crc0 = _mm512_xor_si512(zmm_crc0, zmm_crc1);
+
+    z0 = _mm512_clmulepi64_epi128(zmm_crc0, zmm_fold4, 0x01);
+    zmm_crc0 = _mm512_clmulepi64_epi128(zmm_crc0, zmm_fold4, 0x10);
+    zmm_crc0 = _mm512_xor_si512(z0, zmm_crc0);
+    zmm_crc0 = _mm512_xor_si512(zmm_crc0, zmm_crc2);
+
+    z0 = _mm512_clmulepi64_epi128(zmm_crc0, zmm_fold4, 0x01);
+    zmm_crc0 = _mm512_clmulepi64_epi128(zmm_crc0, zmm_fold4, 0x10);
+    zmm_crc0 = _mm512_xor_si512(z0, zmm_crc0);
+    zmm_crc0 = _mm512_xor_si512(zmm_crc0, zmm_crc3);
+
+    // zmm_crc0 -> xmm_crc[0, 1, 2, 3]
+    *xmm_crc0 = _mm512_extracti32x4_epi32(zmm_crc0, 0);
+    *xmm_crc1 = _mm512_extracti32x4_epi32(zmm_crc0, 1);
+    *xmm_crc2 = _mm512_extracti32x4_epi32(zmm_crc0, 2);
+    *xmm_crc3 = _mm512_extracti32x4_epi32(zmm_crc0, 3);
+
+    return (len_tmp - len);  // return n bytes processed
+}
+
+size_t fold_16_vpclmulqdq_nocp(__m128i *xmm_crc0, __m128i *xmm_crc1,
+    __m128i *xmm_crc2, __m128i *xmm_crc3, const uint8_t *src, size_t len,
+    __m128i init_crc, int32_t first) {
+    size_t len_tmp = len;
+    __m512i zmm_t0, zmm_t1, zmm_t2, zmm_t3;
+    __m512i zmm_crc0, zmm_crc1, zmm_crc2, zmm_crc3;
+    __m512i z0, z1, z2, z3;
+    __m512i zmm_initial = _mm512_zextsi128_si512(init_crc);
+    const __m512i zmm_fold4 = _mm512_set4_epi32(
+        0x00000001, 0x54442bd4, 0x00000001, 0xc6e41596);
+    const __m512i zmm_fold16 = _mm512_set4_epi32(
+        0x00000001, 0x1542778a, 0x00000001, 0x322d1430);
+
+    // zmm register init
+    zmm_crc0 = _mm512_setzero_si512();
+    zmm_t0 = _mm512_loadu_si512((__m512i *)src);
+    XOR_INITIAL(zmm_t0);
+    zmm_crc1 = _mm512_loadu_si512((__m512i *)src + 1);
+    zmm_crc2 = _mm512_loadu_si512((__m512i *)src + 2);
+    zmm_crc3 = _mm512_loadu_si512((__m512i *)src + 3);
+
+    /* already have intermediate CRC in xmm registers
+        * fold4 with 4 xmm_crc to get zmm_crc0
+    */
+    zmm_crc0 = _mm512_inserti32x4(zmm_crc0, *xmm_crc0, 0);
+    zmm_crc0 = _mm512_inserti32x4(zmm_crc0, *xmm_crc1, 1);
+    zmm_crc0 = _mm512_inserti32x4(zmm_crc0, *xmm_crc2, 2);
+    zmm_crc0 = _mm512_inserti32x4(zmm_crc0, *xmm_crc3, 3);
+    z0 = _mm512_clmulepi64_epi128(zmm_crc0, zmm_fold4, 0x01);
+    zmm_crc0 = _mm512_clmulepi64_epi128(zmm_crc0, zmm_fold4, 0x10);
+    zmm_crc0 = _mm512_xor_si512(z0, zmm_crc0);
+    zmm_crc0 = _mm512_xor_si512(zmm_crc0, zmm_t0);
+
+    len -= 256;
+    src += 256;
+
+    // fold-16 loops
+    while (len >= 256) {
+        zmm_t0 = _mm512_loadu_si512((__m512i *)src);
+        zmm_t1 = _mm512_loadu_si512((__m512i *)src + 1);
+        zmm_t2 = _mm512_loadu_si512((__m512i *)src + 2);
+        zmm_t3 = _mm512_loadu_si512((__m512i *)src + 3);
+
+        z0 = _mm512_clmulepi64_epi128(zmm_crc0, zmm_fold16, 0x01);
+        z1 = _mm512_clmulepi64_epi128(zmm_crc1, zmm_fold16, 0x01);
+        z2 = _mm512_clmulepi64_epi128(zmm_crc2, zmm_fold16, 0x01);
+        z3 = _mm512_clmulepi64_epi128(zmm_crc3, zmm_fold16, 0x01);
+
+        zmm_crc0 = _mm512_clmulepi64_epi128(zmm_crc0, zmm_fold16, 0x10);
+        zmm_crc1 = _mm512_clmulepi64_epi128(zmm_crc1, zmm_fold16, 0x10);
+        zmm_crc2 = _mm512_clmulepi64_epi128(zmm_crc2, zmm_fold16, 0x10);
+        zmm_crc3 = _mm512_clmulepi64_epi128(zmm_crc3, zmm_fold16, 0x10);
+
+        zmm_crc0 = _mm512_xor_si512(z0, zmm_crc0);
+        zmm_crc1 = _mm512_xor_si512(z1, zmm_crc1);
+        zmm_crc2 = _mm512_xor_si512(z2, zmm_crc2);
+        zmm_crc3 = _mm512_xor_si512(z3, zmm_crc3);
+
+        zmm_crc0 = _mm512_xor_si512(zmm_crc0, zmm_t0);
+        zmm_crc1 = _mm512_xor_si512(zmm_crc1, zmm_t1);
+        zmm_crc2 = _mm512_xor_si512(zmm_crc2, zmm_t2);
+        zmm_crc3 = _mm512_xor_si512(zmm_crc3, zmm_t3);
+
+        len -= 256;
+        src += 256;
     }
     // zmm_crc[0,1,2,3] -> zmm_crc0
     z0 = _mm512_clmulepi64_epi128(zmm_crc0, zmm_fold4, 0x01);
--- a/cpu_features.h
+++ b/cpu_features.h
@@ -54,7 +54,9 @@ extern uint32_t adler32_power8(uint32_t adler, const unsigned char* buf, size_t 
 #ifdef X86_PCLMULQDQ_CRC
 extern uint32_t crc32_fold_reset_pclmulqdq(crc32_fold *crc);
 extern void     crc32_fold_copy_pclmulqdq(crc32_fold *crc, uint8_t *dst, const uint8_t *src, size_t len);
+extern void     crc32_fold_pclmulqdq(crc32_fold *crc, const uint8_t *src, size_t len, uint32_t init_crc);
 extern uint32_t crc32_fold_final_pclmulqdq(crc32_fold *crc);
+extern uint32_t crc32_pclmulqdq(uint32_t crc32, const unsigned char* buf, uint64_t len);
 #endif
 
 /* memory chunking */
--- a/functable.c
+++ b/functable.c
@@ -393,6 +393,9 @@ Z_INTERNAL uint32_t crc32_stub(uint32_t crc, const unsigned char *buf, uint64_t 
 #elif defined(S390_CRC32_VX)
     if (s390_cpu_has_vx)
         functable.crc32 = &s390_crc32_vx;
+#elif defined(X86_PCLMULQDQ_CRC)
+    if (x86_cpu_has_pclmulqdq)
+        functable.crc32 = &crc32_pclmulqdq;
 #endif
 
     return functable.crc32(crc, buf, len);
--- a/test/benchmarks/benchmark_crc32.cc
+++ b/test/benchmarks/benchmark_crc32.cc
@@ -66,11 +66,5 @@ BENCHMARK_CRC32(power8, crc32_power8, power_cpu_has_arch_2_07);
 BENCHMARK_CRC32(vx, s390_crc32_vx, s390_cpu_has_vx);
 #elif defined(X86_PCLMULQDQ_CRC)
 /* CRC32 fold does a memory copy while hashing */
-uint32_t crc32_pclmulqdq(uint32_t crc32, const unsigned char* buf, uint64_t len) {
-    crc32_fold ALIGNED_(16) crc_state;
-    crc32_fold_reset_pclmulqdq(&crc_state);
-    crc32_fold_copy_pclmulqdq(&crc_state, (uint8_t *)buf, buf, len);
-    return crc32_fold_final_pclmulqdq(&crc_state);
-}
 BENCHMARK_CRC32(pclmulqdq, crc32_pclmulqdq, x86_cpu_has_pclmulqdq);
 #endif


KungFuJesus:
@nmoinvaz I wrapped your benchmark code for this.  We could also _probably_ use the CRC (SSE4.2) instructions for when the length is less than 64, but I wasn't quite sure how to make that work with polynomials that are expected.  It's probably simple, I didn't look _too_ hard at it.

KungFuJesus:
We also probably _could_ forego the write back to buf since it doesn't actually change the results.  Though in a circumstance where the function was fully inlined, the compiler is likely able to optimize away this write back.

arch/x86/crc32_fold_pclmulqdq.c
```diff
@@ -462,4 +462,11 @@ Z_INTERNAL uint32_t crc32_fold_final_pclmulqdq(crc32_fold *crc) {
     return crc->value;
 }
 
+uint32_t crc32_pclmulqdq(uint32_t crc32, const unsigned char* buf, uint64_t len) {
```
nmoinvaz, line 590: I don't think this will work right because it doesn't do anything with `crc32` value.

KungFuJesus: Pretty sure it does?  The return statement finalizes the CRC value by folding up the 4 xmm registers to a single sum, no?

nmoinvaz: I wonder if `crc32_combine` can be used to combine the initial value with the next value. OR if there is a way to load the initial value into our `crc32_fold..` stuff.

nmoinvaz: The `crc32` variable is unused in the function..

KungFuJesus:
> We also probably _could_ forego the write back to buf since it doesn't actually change the results. Though in a circumstance where the function was fully inlined, the compiler is likely able to optimize away this write back.

Interesting - jumping over all the stores (by designating dst as NULL) barely moves the mean (3% at most) but it does move the standard deviation significantly in some places (for whatever that's worth).   Given that these values are written back in place, I have to wonder if the writes are mainly being absorbed by fill buffers.

Whether it's worth the branching and extra complexity...I dunno.  We could completely duplicate the functions without the copy as something else we call, with a small bit of code inflation (and a bit of a nuisance to maintain), but it could fare a _little_ better maybe.

Here's the "jumping over the store" comparison:

```
Comparing crc32/pclmulqdq (from ./benchmark_zlib_before) to crc32/pclmulqdq (from ./benchmark_zlib)
Benchmark                                                              Time             CPU      Time Old      Time New       CPU Old       CPU New
---------------------------------------------------------------------------------------------------------------------------------------------------
[crc32/pclmulqdq vs. crc32/pclmulqdq]/1                             -0.0667         -0.0667            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/1                             -0.0675         -0.0675            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/1                             -0.0670         -0.0670            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/1                             -0.0672         -0.0672            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/1                             -0.0669         -0.0669            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/1                             -0.0668         -0.0668            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/1                             -0.0674         -0.0674            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/1                             -0.0668         -0.0668            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/1                             -0.0668         -0.0668            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/1                             -0.0669         -0.0669            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/1_pvalue                       0.0002          0.0002      U Test, Repetitions: 10 vs 10
[crc32/pclmulqdq vs. crc32/pclmulqdq]/1_mean                        -0.0670         -0.0670            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/1_median                      -0.0669         -0.0669            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/1_stddev                      -0.8392         -0.8415             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/1_cv                          -0.8277         -0.8301             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/8                             -0.0655         -0.0655            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/8                             -0.0655         -0.0655            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/8                             -0.0656         -0.0656            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/8                             -0.0655         -0.0655            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/8                             -0.0656         -0.0656            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/8                             -0.0656         -0.0656            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/8                             -0.0658         -0.0658            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/8                             -0.0658         -0.0658            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/8                             -0.0655         -0.0655            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/8                             -0.0656         -0.0656            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/8_pvalue                       0.0002          0.0002      U Test, Repetitions: 10 vs 10
[crc32/pclmulqdq vs. crc32/pclmulqdq]/8_mean                        -0.0656         -0.0656            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/8_median                      -0.0656         -0.0656            22            21            22            21
[crc32/pclmulqdq vs. crc32/pclmulqdq]/8_stddev                      -0.6136         -0.6259             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/8_cv                          -0.5865         -0.5996             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64                            -0.0365         -0.0365            12            12            12            12
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64                            -0.0362         -0.0362            12            12            12            12
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64                            -0.0364         -0.0364            12            12            12            12
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64                            -0.0361         -0.0361            12            12            12            12
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64                            -0.0365         -0.0364            12            12            12            12
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64                            -0.0364         -0.0364            12            12            12            12
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64                            -0.0366         -0.0366            12            12            12            12
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64                            -0.0363         -0.0363            12            12            12            12
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64                            -0.0363         -0.0363            12            12            12            12
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64                            -0.0370         -0.0370            12            12            12            12
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64_pvalue                      0.0002          0.0002      U Test, Repetitions: 10 vs 10
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64_mean                       -0.0364         -0.0364            12            12            12            12
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64_median                     -0.0365         -0.0365            12            12            12            12
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64_stddev                     -0.3683         -0.3712             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64_cv                         -0.3444         -0.3474             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512                           -0.0058         -0.0058            34            34            34            34
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512                           -0.0065         -0.0065            34            34            34            34
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512                           -0.0072         -0.0072            34            34            34            34
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512                           -0.0075         -0.0075            34            34            34            34
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512                           -0.0061         -0.0061            34            34            34            34
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512                           -0.0068         -0.0068            34            34            34            34
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512                           -0.0070         -0.0069            34            34            34            34
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512                           -0.0067         -0.0067            34            34            34            34
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512                           -0.0063         -0.0063            34            34            34            34
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512                           -0.0077         -0.0077            34            34            34            34
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512_pvalue                     0.0002          0.0002      U Test, Repetitions: 10 vs 10
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512_mean                      -0.0067         -0.0068            34            34            34            34
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512_median                    -0.0066         -0.0066            34            34            34            34
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512_stddev                    -0.3900         -0.3996             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512_cv                        -0.3859         -0.3955             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096                          -0.0114         -0.0114           175           173           175           173
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096                          -0.0114         -0.0114           175           173           175           173
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096                          -0.0101         -0.0102           175           173           175           173
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096                          -0.0060         -0.0059           174           173           174           173
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096                          -0.0110         -0.0110           175           173           175           173
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096                          -0.0117         -0.0117           175           173           175           173
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096                          -0.0125         -0.0125           175           173           175           173
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096                          -0.0080         -0.0080           174           173           174           173
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096                          -0.0076         -0.0076           174           173           174           173
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096                          -0.0080         -0.0080           174           173           174           173
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096_pvalue                    0.0002          0.0002      U Test, Repetitions: 10 vs 10
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096_mean                     -0.0098         -0.0098           175           173           175           173
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096_median                   -0.0105         -0.0105           175           173           175           173
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096_stddev                   -0.9085         -0.9099             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096_cv                       -0.9076         -0.9090             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768                         -0.0065         -0.0066          1277          1269          1277          1269
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768                         -0.0066         -0.0066          1277          1269          1277          1269
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768                         -0.0070         -0.0070          1277          1269          1277          1269
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768                         -0.0065         -0.0065          1277          1269          1277          1269
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768                         -0.0068         -0.0068          1278          1269          1278          1269
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768                         -0.0067         -0.0067          1277          1269          1277          1269
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768                         -0.0070         -0.0071          1278          1269          1278          1269
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768                         -0.0066         -0.0065          1277          1269          1277          1269
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768                         -0.0069         -0.0069          1277          1269          1277          1269
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768                         -0.0066         -0.0066          1277          1269          1277          1269
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768_pvalue                   0.0002          0.0002      U Test, Repetitions: 10 vs 10
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768_mean                    -0.0067         -0.0067          1277          1269          1277          1269
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768_median                  -0.0067         -0.0067          1277          1269          1277          1269
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768_stddev                  -0.7308         -0.7141             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768_cv                      -0.7289         -0.7122             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144                        -0.0003         -0.0003         10034         10031         10033         10031
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144                        -0.0004         -0.0004         10033         10030         10033         10030
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144                        -0.0004         -0.0004         10033         10030         10033         10029
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144                        -0.0006         -0.0006         10035         10030         10035         10029
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144                        -0.0003         -0.0003         10034         10031         10034         10030
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144                        -0.0005         -0.0005         10034         10029         10034         10029
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144                        -0.0004         -0.0004         10034         10030         10034         10030
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144                        -0.0003         -0.0003         10034         10031         10034         10031
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144                        -0.0004         -0.0004         10034         10030         10034         10030
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144                        -0.0004         -0.0004         10034         10030         10034         10029
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144_pvalue                  0.0002          0.0002      U Test, Repetitions: 10 vs 10
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144_mean                   -0.0004         -0.0004         10034         10030         10034         10030
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144_median                 -0.0004         -0.0004         10034         10030         10034         10030
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144_stddev                 +0.0665         +0.0242             1             1             1             1
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144_cv                     +0.0669         +0.0246             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152                       +0.0011         +0.0011        119362        119492        119359        119488
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152                       -0.0366         -0.0366        120229        115830        120227        115827
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152                       +0.0022         +0.0022        120100        120362        120096        120360
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152                       -0.0036         -0.0036        118815        118385        118812        118382
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152                       -0.0244         -0.0244        120298        117358        120294        117355
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152                       -0.0052         -0.0052        118944        118329        118943        118326
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152                       -0.0087         -0.0087        119260        118223        119257        118222
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152                       -0.0103         -0.0103        118202        116986        118199        116982
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152                       -0.0024         -0.0024        118622        118341        118621        118337
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152                       -0.0061         -0.0061        118840        118114        118837        118113
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152_pvalue                 0.0312          0.0312      U Test, Repetitions: 10 vs 10
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152_mean                  -0.0094         -0.0094        119267        118142        119265        118139
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152_median                -0.0069         -0.0069        119102        118276        119100        118274
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152_stddev                +0.7314         +0.7326           725          1255           725          1256
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152_cv                    +0.7479         +0.7491             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304                       -0.0014         -0.0014        239948        239617        239939        239612
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304                       -0.0193         -0.0193        241274        236614        241266        236607
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304                       -0.0399         -0.0399        244911        235131        244908        235124
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304                       -0.0344         -0.0344        242342        234005        242339        233997
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304                       -0.0188         -0.0188        241378        236830        241370        236828
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304                       +0.0015         +0.0015        236748        237113        236742        237107
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304                       -0.0182         -0.0183        241284        236883        241281        236875
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304                       -0.0079         -0.0079        238580        236696        238572        236694
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304                       -0.0288         -0.0288        240504        233576        240494        233568
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304                       +0.0198         +0.0197        238274        242984        238272        242976
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304_pvalue                 0.0091          0.0091      U Test, Repetitions: 10 vs 10
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304_mean                  -0.0149         -0.0149        240524        236945        240518        236939
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304_median                -0.0171         -0.0171        240889        236763        240880        236761
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304_stddev                +0.1843         +0.1839          2306          2731          2307          2731
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304_cv                    +0.2022         +0.2018             0             0             0             0
OVERALL_GEOMEAN                                                     -0.0343         -0.0342             0             0             0             0
```

Note, the values below 64 bytes aren't really valid since we only use it for >= 64 bytes.  There's still likely some performance to be had by using crc32 instruction.

arch/x86/crc32_fold_pclmulqdq.c
nmoinvaz: In the benchmarks I had to add some kind of `supports_initial_value` in order to benchmark it... so it doesn't support having an initial `crc32` value..

KungFuJesus: ahh, you are correct there, if it's non-zero that will be a problem.  Yes we should be able to feed it after the "reset" function.

functable.c
```diff
@@ -393,6 +393,12 @@ Z_INTERNAL uint32_t crc32_stub(uint32_t crc, const unsigned char *buf, uint64_t
 #elif defined(S390_CRC32_VX)
     if (s390_cpu_has_vx)
         functable.crc32 = &s390_crc32_vx;
+#elif defined(X86_PCLMULQDQ_CRC)
+    /* For lens < 64, crc32_byfour method is faster. The CRC32 instruction for
+     * these short lengths might also prove to be effective */
+    if (x86_cpu_has_pclmulqdq && len >= 64) {
+        functable.crc32 = &crc32_pclmulqdq;
```
mtl1979: Checking `len` here doesn't work as expected as the stub is called only once. The check should be inside the callback function and the callback function should call the fallback function if `len` there is too small...

arch/x86/crc32_fold_pclmulqdq.c
KungFuJesus: Ok, that last push should fix that.

functable.c
KungFuJesus: Yep, fixed it.

arch/x86/crc32_fold_pclmulqdq.c
nmoinvaz: I'm doing some tests but I don't think it is possible to just xor the two values..

mtl1979: Combining two CRC32 values need at least length of one of the original byte streams... It might be better to use the existing combine function instead...

nmoinvaz: crc32_combine functions look like they also need the length..

mtl1979: `crc32_combine` needs the length as it adds number of zeroes between the two byte streams, so they are of equal length. But here we already know how long the second byte stream is...

KungFuJesus: Ahhh, incidentally I'm getting segfaults on the unit tests.  Weirdly, on an unaligned move.  For some reason it does not seem well behaved when unaligned, which is very strange.  

I'll call the combine function instead (though, it did seem like in my testing the simple XOR'ing was correct).

nmoinvaz: @mtl1979 weird that the docs/comment for `crc32_combine` mentions a variable `len1` but only exists `len2`.

nmoinvaz: @KungFuJesus I had mentioned previously benchmarks, I was wrong. It was in the google unit tests PR that I checked for `works_with_initial_value`. Also I have pushed a commit which should fix some cpu feature checking.

mtl1979: @nmoinvaz I haven't read the documentation in a while... I just know how combing two crc32 values is supposed to work from reading the source code...

nmoinvaz: @mtl1979 you are right, it works like this.. probably needs to be cleaned up..
image

mtl1979: @nmoinvaz AFAIK it doesn't need length of first byte stream as crc32 of initial zeroes is same as resetting the state, so basically has no effect.

nmoinvaz: @KungFuJesus in my crc32 shim I had to have an extra buffer to write to because I think it is passing in constant strings and such as `buf`.

nmoinvaz:
@KungFuJesus it might be worth it to have the extra checks for safety. In case (as in the unit tests) we pass in constant strings and such as `buf`? I mean it only shaves a few ns off the total... but that is possibly not the only reason to have it. What do you think?

arch/x86/crc32_fold_pclmulqdq.c
KungFuJesus: Ah well it only matters for writing. That makes a good deal of sense, since constant strings are mapped read only in the binary.

Perhaps it makes sense then to write separate non-copying versions? I suppose we _could_ templatize things such that a macro compiled the copy in or not, I'm not sure if that's much better as far as maintainability.

Off topic but, I had half a mind to write out a similar sort of thing for the Adler checksums where it's used in deflate (roll the values through the checksum while they're in cache or registers) but doing that effectively without making ridiculously long function bodies would require us templating it out like we do with some of the functions. Since we're blistering fast when fitting into cache, avoiding the trip to memory entirely would probably have substantial gains at the cost of writing out inlined versions of sections of the function so that we can wrap deflate into the checksum effectively (maximizing ILP and all of that).

KungFuJesus:
> @KungFuJesus it might be worth it to have the extra checks for safety. In case (as in the unit tests) we pass in constant strings and such as `buf`? I mean it only shaves a few ns off the total... but that is possibly not the only reason to have it. What do you think?

Copying out the entire buffer being checksummed seems less than ideal. I'd just as soon remove the copy in the way I've described previously (we could also jump past it like I had tried above). I'm pretty sure the reads will never be an issue.

nmoinvaz:
Yes I was saying also we should do it the way you described. Making copying optional.

KungFuJesus:
I'll take a stab at this tomorrow. The read only version would probably more sensibly have a different function signature rather than just ignoring when dst is null, but I'll have to see. It would not surprise me if not having to jump over the copies proved to be a modest win.

arch/x86/crc32_fold_pclmulqdq.c
KungFuJesus: So, the crc32_combine function works but when compiling in zlib compatibility mode it complains about the signature being different.  I obviously can cast it down to a 32 bit integer but, is there a cleaner way?

KungFuJesus: Ooof, that crc32_combine function is also not an insignificant amount of the computation time.  It doesn't correlate to a ton of extra time on the benchmark for the larger sizes but it's like 7x worse for the smaller sizes.  I don't think it's any wonder the "byfour" isn't using it to incorporate the running crc into its checksum.  

Perf reports:

```
  57.25%  benchmark_zlib  benchmark_zlib        [.] crc32_pclmulqdq
  17.62%  benchmark_zlib  benchmark_zlib        [.] crc32_combine
  12.38%  benchmark_zlib  benchmark_zlib        [.] crc32_byfour
   7.84%  benchmark_zlib  libc.so.6             [.] random
   2.51%  benchmark_zlib  benchmark_zlib        [.] crc32_pclmulqdq_Benchmark::BenchmarkCase
```

nmoinvaz: Here is an [implementation](https://github.com/borgbackup/borg/blob/2197e94a316d0ad92149c47226d4206e3fe00c37/src/borg/algorithms/crc32_clmul.c) that uses initial value. Although that code is structured differently. I wonder if it is better to just do something like that (where everything is in one function) - no separate reset/update/finish functions. I don't know that we really gain much by splitting it up. Maybe save on a couple of saves/loads. But at the cost of maintainability? Chromium also combines it all into one function.

KungFuJesus: That's actually pretty straightforward. I should be able to do this without too much rework.

The stores could be isolated to a single block/label that gets skipped over maybe, it just depends on whether or not register pressure imposes that things end up spilling onto stack or not as to whether we lose the efficiency of the ancillary copy.

As far as splitting up the disparate shared stages, some of that is ideally inlined as the function level abstraction otherwise calls for pushing quite a few things on and off the stack where we'd be better off shuffling through registers.  Apart from maybe the reset function I believe that is what's happening. So as far as efficiency I don't believe it's costing us much to have them broken out like that, but I'll need to take a closer look at a profile.

KungFuJesus: So XOR is commutative, no?  Why are they constantly trying to XOR the very first term read from the buffer (with a goofy branchy do-once macro) when they could just as easily do it one time in the initial first term of the CRC?

KungFuJesus: Hmm, it would appear that the carryless multiplication operations that occur inbetween the XORs won't commute.  I really do need to do this on the very first term and put it everywhere there's a branch for a possible first load.  I mean, I guess that's doable...

nmoinvaz: Well chromium version only does a single xor at the beginning of the function.. 
https://chromium.googlesource.com/chromium/src/third_party/zlib/+/refs/heads/main/crc32_simd.c#48

nmoinvaz: I think it only does that at the beginning, but due to different values of `len`, we don't know where the first read will happen at..?

KungFuJesus: So, not _exactly_ a direct comparison as the first version didn't incorporate the initial CRC value, but here's a comparison with a version that doesn't have to branch at all and bypasses the stores by duplicating the whole function (sans store):
```
Comparing crc32/pclmulqdq (from ./benchmark_zlib_before) to crc32/pclmulqdq (from ./benchmark_zlib)
Benchmark                                                              Time             CPU      Time Old      Time New       CPU Old       CPU New
---------------------------------------------------------------------------------------------------------------------------------------------------
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64_mean                       +0.4669         +0.4669            12            18            12            18
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64_median                     +0.4671         +0.4671            12            18            12            18
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64_stddev                     -0.5437         -0.5969             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/64_cv                         -0.6890         -0.7252             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512_mean                      +0.0307         +0.0307            34            35            34            35
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512_median                    +0.0307         +0.0307            34            35            34            35
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512_stddev                    -0.9257         -0.9234             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/512_cv                        -0.9279         -0.9256             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096_mean                     -0.0041         -0.0041           175           174           175           174
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096_median                   -0.0036         -0.0036           175           174           175           174
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096_stddev                   -0.9845         -0.9860             1             0             1             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4096_cv                       -0.9844         -0.9859             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768_mean                    -0.0045         -0.0045          1277          1271          1277          1271
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768_median                  -0.0047         -0.0048          1277          1271          1277          1271
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768_stddev                  -0.9223         -0.9268             1             0             1             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/32768_cv                      -0.9219         -0.9264             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144_mean                   -0.0004         -0.0004         10034         10030         10034         10030
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144_median                 -0.0004         -0.0004         10034         10030         10034         10030
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144_stddev                 +1.3674         +0.9789             0             1             0             1
[crc32/pclmulqdq vs. crc32/pclmulqdq]/262144_cv                     +1.3684         +0.9797             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152_mean                  -0.0989         -0.0989        119229        107441        119226        107438
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152_median                -0.0968         -0.0968        119278        107734        119275        107731
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152_stddev                +1.2576         +1.2594           511          1153           511          1154
[crc32/pclmulqdq vs. crc32/pclmulqdq]/2097152_cv                    +1.5053         +1.5073             0             0             0             0
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304_mean                  -0.0984         -0.0984        238786        215286        238780        215281
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304_median                -0.0982         -0.0982        238520        215106        238515        215098
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304_stddev                -0.0913         -0.0921          2375          2158          2377          2158
[crc32/pclmulqdq vs. crc32/pclmulqdq]/4194304_cv                    +0.0079         +0.0070             0             0             0             0
```
The 9% deltas for the larger sizes are nice, at least.  The extra 6 ns to the mean for the 64 byte size makes some sense, as the "if not loaded load, load" logic may cause an initial branch mispredict that maybe wasn't there before.

KungFuJesus: > Well chromium version only does a single xor at the beginning of the function.. https://chromium.googlesource.com/chromium/src/third_party/zlib/+/refs/heads/main/crc32_simd.c#48

Right, I no doubt only need to do it once but apparently I can't do it to values inside the "fold" buffer as that won't properly commute.  The carryless multiplication compounded with the shuffle/lookup table that gets used must be an operation that can't actually commute the value as though it were a simple multiplication or addition on the first set of bytes.  It's just unfortunate that doing that is rather ugly and requires at minimum a check to see "did I do this yet?" at every possible branch of a first load.

functable.c
```diff
@@ -393,6 +393,12 @@ Z_INTERNAL uint32_t crc32_stub(uint32_t crc, const unsigned char *buf, uint64_t
 #elif defined(S390_CRC32_VX)
     if (s390_cpu_has_vx)
         functable.crc32 = &s390_crc32_vx;
+#elif defined(X86_PCLMULQDQ_CRC)
+    /* For lens < 64, crc32_byfour method is faster. The CRC32 instruction for
+     * these short lengths might also prove to be effective */
```
mtl1979: This comment is misplaced now that the `len` check was moved inside the function.

arch/x86/crc32_fold_pclmulqdq.c
nmoinvaz: Can you add:
```
Copyright (C) 2016 Marian Beermann (support for initial value)
```
From the implementation I mentioned, to the top?

KungFuJesus: done

functable.c
```diff
@@ -393,6 +393,10 @@ Z_INTERNAL uint32_t crc32_stub(uint32_t crc, const unsigned char *buf, uint64_t
 #elif defined(S390_CRC32_VX)
     if (s390_cpu_has_vx)
         functable.crc32 = &s390_crc32_vx;
+#elif defined(X86_PCLMULQDQ_CRC)
+    if (x86_cpu_has_pclmulqdq) {
```
nmoinvaz: No need for brackets..

arch/x86/crc32_fold_pclmulqdq.c
```diff
@@ -24,12 +25,17 @@
...
 #include "x86_features.h"
+#include "cpu_features.h"
 
 #include "../../crc32_fold.h"
+#include "../../crc32_p.h"
 
 #ifdef X86_VPCLMULQDQ_CRC
 extern size_t fold_16_vpclmulqdq(__m128i *xmm_crc0, __m128i *xmm_crc1,
     __m128i *xmm_crc2, __m128i *xmm_crc3, uint8_t *dst, const uint8_t *src, size_t len);
+extern size_t fold_16_vpclmulqdq_nocp(__m128i *xmm_crc0, __m128i *xmm_crc1,
```
nmoinvaz, line 37: Why not just use `fold_16_vpclmulqdq` and check for `dst == NULL`?

arch/x86/crc32_fold_pclmulqdq.c
```diff
@@ -380,6 +386,133 @@ Z_INTERNAL void crc32_fold_copy_pclmulqdq(crc32_fold *crc, uint8_t *dst, const u
     crc32_fold_save_partial((__m128i *)crc->fold, xmm_crc_part);
 }
 
+#define ONCE(op) if(first) { \
```
nmoinvaz: Add a space after `if` keyword.

arch/x86/crc32_fold_pclmulqdq.c
```diff
@@ -24,12 +25,17 @@
...
 
 #include "../../crc32_fold.h"
+#include "../../crc32_p.h"
 
 #ifdef X86_VPCLMULQDQ_CRC
 extern size_t fold_16_vpclmulqdq(__m128i *xmm_crc0, __m128i *xmm_crc1,
     __m128i *xmm_crc2, __m128i *xmm_crc3, uint8_t *dst, const uint8_t *src, size_t len);
+extern size_t fold_16_vpclmulqdq_nocp(__m128i *xmm_crc0, __m128i *xmm_crc1,
+    __m128i *xmm_crc2, __m128i *xmm_crc3, const uint8_t *src, size_t len, __m128i init_crc,
+    int first);
```
nmoinvaz: It's best if we can use fixed-width integer types if we can.

nmoinvaz:
I almost think we are better off doing templating `_tpl.h` like you suggested. What do you think?

KungFuJesus:
> I almost think we are better off doing templating `_tpl.h` like you suggested. What do you think?

I mean I can do it, it's just rather ugly to mangle function names with macros and define/undefine them.  The main thing is that they probably would end up with different function signatures since "dst" is unused, and as far as I know we only have used the template approach for functions which share an identical signature.  From a maintenance standpoint, it sure makes sense.  I'll admit, it took me a bit to unwind what "longest_match" was doing when I first saw it, though.  

It's a shame that C can't overload due to its simple and unchanging calling convention.

nmoinvaz:
Hmm. Maybe it is better for a separate commit or PR then.

arch/x86/crc32_fold_pclmulqdq.c
KungFuJesus: That's a good point, given that we unconditionally get to do _1_ load here (guaranteed by the initial size check), we only need a single branch at the beginning.  Have to run out for a bit but this will be a quick change.

KungFuJesus: Ah, at second glance I also removed the simultaneous copy from that version.  So, in a loop I'm avoiding a bunch of if dst != NULL jumps.

KungFuJesus:
Hmm I'm a little confused why the libpng crc is failing. I'll have see if I can reproduce that one in a debugger. Maybe a missed INITIAL_XOR?

arch/x86/crc32_fold_vpclmulqdq.c
```diff
@@ -105,4 +112,95 @@ size_t fold_16_vpclmulqdq(__m128i *xmm_crc0, __m128i *xmm_crc1,
 
     return (len_tmp - len);  // return n bytes processed
 }
+
+size_t fold_16_vpclmulqdq_nocp(__m128i *xmm_crc0, __m128i *xmm_crc1,
+    __m128i *xmm_crc2, __m128i *xmm_crc3, const uint8_t *src, size_t len,
+    __m128i init_crc, int32_t first) {
```
nmoinvaz, line 118: Shouldn't this be `int32_t *first` so that the value of `first` is retained?

arch/x86/crc32_fold_pclmulqdq.c
```diff
@@ -380,6 +386,133 @@ Z_INTERNAL void crc32_fold_copy_pclmulqdq(crc32_fold *crc, uint8_t *dst, const u
...
+        partial_fold(algn_diff, &xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3, &xmm_crc_part);
+    } else {
+        xmm_crc_part = _mm_setzero_si128();
+    }
+
+#ifdef X86_VPCLMULQDQ_CRC
+    if (x86_cpu_has_vpclmulqdq && x86_cpu_has_avx512 && (len >= 256)) {
+        size_t n = fold_16_vpclmulqdq_nocp(&xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3, src, len,
+                xmm_initial, first);
+        first = 0;
```
nmoinvaz, line 439: Ah I see. You set `first = 0` here.

nmoinvaz:
Nothing obvious stands out.

KungFuJesus:
By a somewhat painful process I was able to find an offending sequence in libpng's failing tests.  I modified the CRC callback to print the input crc and lengths as well as the resulting CRCs for pre and post pclmulqdq modifications.  I compared their outputs in vimdiff and it's this:

png-crc-fail

So to debug this further I'm going to need to excise that 1890 byte string and see where they differ.  I probably won't have time for this until the evening.

KungFuJesus:
Ooo, interesting thing found - the only thing that's different about that address out of _all_ of them is that it's not aligned to 16 bytes.  I think the bug is hiding in there somewhere (which may mean it exists in the current unmodified code, too).

arch/x86/crc32_fold_pclmulqdq.c
```diff
@@ -380,6 +386,133 @@ Z_INTERNAL void crc32_fold_copy_pclmulqdq(crc32_fold *crc, uint8_t *dst, const u
...
+    if (algn_diff) {
+        xmm_crc_part = _mm_loadu_si128((__m128i *)src);
+
+        src += algn_diff;
+        len -= algn_diff;
+
+        XOR_INITIAL(xmm_crc_part);
+        partial_fold(algn_diff, &xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3, &xmm_crc_part);
+    } else {
+        xmm_crc_part = _mm_setzero_si128();
```
nmoinvaz: What if we loaded the first 16 bytes to do the initial XOR and then we don't have to check `first` variable in the loops?

```
else {
    xmm_crc_part = _mm_setzero_si128();
    // XOR initial crc value since len is guaranteed to be >= 16
    xmm_t0 = _mm_load_si128((__m128i *)src);
    XOR_INITIAL(xmm_t0);

    fold_1(&xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3);

    xmm_crc3 = _mm_xor_si128(xmm_crc3, xmm_t0);

    src += 16;
    len -= 16;
    if (len == 0)
        goto done_nocopy;
}
```
Also we wouldn't need to check it again, because it was either done in `align_diff` branch or done in this branch.

KungFuJesus: Kinda might work but we'd have to still handle edge cases of < 16.  Though in our case I'm not even calling this unless we have 64 bytes due to it being slower otherwise.

However, I believe I just found the source of the bug.  The issue in this circumstance is that algn_diff is 2, not a full 4 bytes.  So when we fold it in, we don't accommodate for the entirety of the initial crc32 checksum.  Hah, that was...tricky.  I'm not sure what the best way to solve it is.  Given that we we call this function, it's at a minimum of 64 bytes, maybe I can account for that by doing unaligned loads out to the next remainder?  Or I could do something a _little_ stranger, like shifting in zeros for the initial value for the XOR.

nmoinvaz: Actually perhaps the `align_diff` is the issue you are seeing? It is not CRCing a full 16 bytes? Not sure if that makes a difference.

nmoinvaz: Looks like you found it and it was similar to what I was thinking. What if we do unaligned loads for 16 bytes to load the initial CRC value, before `align_diff` branch?

nmoinvaz: `fold_1` is aligned so we would need an unaligned version possibly...

mtl1979:
It could be easier to use the C version to align the buffer and then load the new initial value... Nobody knows if the align difference is 1, 2, or 3 bytes, and taking account all the possibilities using vector instructions might make the code messy.

arch/x86/crc32_fold_pclmulqdq.c
KungFuJesus: I think I got it, check out the latest push

KungFuJesus:
> It could be easier to use the C version to align the buffer and then load the new initial value... Nobody knows if the align difference is 1, 2, or 3 bytes, and taking account all the possibilities using vector instructions might make the code messy.

It wasn't too awful, there's already code to do partial folds with full vectors + some residual.  Though, you might be right, sub integer alignment adjustments could be slower to roll into the carryless multiplication chains if it's only a handful of bytes.  I might try it that way later tonight and benchmark the two to see if one is significantly faster than the other.

arch/x86/crc32_fold_pclmulqdq.c
```diff
@@ -380,6 +387,131 @@ Z_INTERNAL void crc32_fold_copy_pclmulqdq(crc32_fold *crc, uint8_t *dst, const u
...
+            partial_fold(algn_diff, &xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3, &xmm_crc_part);
+        } else {
+            xmm_t0 = _mm_loadu_si128((__m128i*)src);
+            xmm_crc_part = _mm_loadu_si128((__m128i*)src + 1);
+            XOR_INITIAL(xmm_t0);
+            fold_1(&xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3);
+            xmm_crc3 = _mm_xor_si128(xmm_crc3, xmm_t0);
+            partial_fold(algn_diff, &xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3, &xmm_crc_part);
+
+            src += (algn_diff + 16);
```
nmoinvaz, line 428: Wouldn't you also want to process initial value when `algn_diff == 0`, so you don't have to do it in the loops?

arch/x86/crc32_fold_pclmulqdq.c
nmoinvaz: Nice job with the unaligned load. That is simpler than I thought it would be.

mtl1979:
> It wasn't too awful, there's already code to do partial folds with full vectors + some residual. Though, you might be right, sub integer alignment adjustments could be slower to roll into the carryless multiplication chains if it's only a handful of bytes. I might try it that way later tonight and benchmark the two to see if one is significantly faster than the other.

Benchmarking different alternative implementations is always a good thing... Even if something seems fast in paper, it might not be the fastest possible implementation. We have to weigh both correctness and readability, so other people can review the changes properly without adding excessive inline commenting.

arch/x86/crc32_fold_pclmulqdq.c
KungFuJesus: Yes but in doing that I'd still need to account for the load of xmm_t0.  I could conceivably jump past the load in the loop I guess, but that seems a little bit more complicated, especially given that we have another function we hit first if vpclmulqdq is supported...

The xmm_t0 register still needs to be loaded every time in the loop, I can't get rid of that, but if I XOR the initial CRC value with it, I need special first run behavior for it.

KungFuJesus:
> > It wasn't too awful, there's already code to do partial folds with full vectors + some residual. Though, you might be right, sub integer alignment adjustments could be slower to roll into the carryless multiplication chains if it's only a handful of bytes. I might try it that way later tonight and benchmark the two to see if one is significantly faster than the other.
> 
> Benchmarking different alternative implementations is always a good thing... Even if something seems fast in paper, it might not be the fastest possible implementation. We have to weigh both correctness and readability, so other people can review the changes properly without adding excessive inline commenting.

I couldn't help myself, it was a simple thing to test so I just did it.  It's not significantly better by any margin and is in fact worse in some cases.  I wouldn't mind being able to use the CRC instruction up to the aligning number of bytes, though, I just need to find proper examples for how to use it.  The way insert_string is using it isn't super intuitive.

arch/x86/crc32_fold_pclmulqdq.c
KungFuJesus: This also means borg backup has a bug, hah.  Maybe someone should tell them?

KungFuJesus:
Doing a tiny bit of research revealed that the SSE4.2 instructions are actually for CRC32-C, not the standard CRC polynomial.  There's no great way to go from one to the other, so, what's there is probably best for the shorter strings.  

Also _probably_ worth noting that this bit here in the readme is _technically_ wrong:
> Hash table implementation using CRC32-C intrinsics on x86 and ARM

The insert_string function with ACLE is just using plain ol' vanilla CRC32-B.  Though, supposedly the Castagnoli is a better one and would perhaps produce fewer collisions?  I'm not sure if that table really _needs_ strong protection against collisions or not, but that's the working theory for why it's superior.  ARM's ACLE extensions have instructions for both, and obviously probably complete in identical times.

nmoinvaz:
Yeah I noticed that the README was out of date last night too. Although I wonder now that you say CRC32-C is available on both platforms, if we should just use that for consistency.

cpu_features.h
```diff
@@ -54,7 +54,9 @@ extern uint32_t adler32_power8(uint32_t adler, const unsigned char* buf, size_t
 #ifdef X86_PCLMULQDQ_CRC
 extern uint32_t crc32_fold_reset_pclmulqdq(crc32_fold *crc);
 extern void     crc32_fold_copy_pclmulqdq(crc32_fold *crc, uint8_t *dst, const uint8_t *src, size_t len);
+extern void     crc32_fold_pclmulqdq(crc32_fold *crc, const uint8_t *src, size_t len, uint32_t init_crc);
```
nmoinvaz, line 57: Do we need this `crc32_fold_pclmulqdq` function to be extern?

arch/x86/crc32_fold_pclmulqdq.c
```diff
@@ -24,12 +25,18 @@
...
 
 #include "../../crc32_fold.h"
+#include "../../crc32_p.h"
+#include <assert.h>
 
 #ifdef X86_VPCLMULQDQ_CRC
 extern size_t fold_16_vpclmulqdq(__m128i *xmm_crc0, __m128i *xmm_crc1,
     __m128i *xmm_crc2, __m128i *xmm_crc3, uint8_t *dst, const uint8_t *src, size_t len);
+extern size_t fold_16_vpclmulqdq_nocp(__m128i *xmm_crc0, __m128i *xmm_crc1,
+    __m128i *xmm_crc2, __m128i *xmm_crc3, const uint8_t *src, size_t len, __m128i init_crc,
```
nmoinvaz, line 38: I thought you were going to just have one `fold_16` function with `if (dst == NULL)` checks. What happened to that?

arch/x86/crc32_fold_pclmulqdq.c
```diff
@@ -380,6 +387,131 @@ Z_INTERNAL void crc32_fold_copy_pclmulqdq(crc32_fold *crc, uint8_t *dst, const u
...
+    (op); \
+}
+#define XOR_INITIAL(where) ONCE(where = _mm_xor_si128(where, xmm_initial))
+
+Z_INTERNAL void crc32_fold_pclmulqdq(crc32_fold *crc, const uint8_t *src, size_t len, uint32_t init_crc) {
+    unsigned long algn_diff;
+    __m128i xmm_t0, xmm_t1, xmm_t2, xmm_t3;
+    __m128i xmm_crc0, xmm_crc1, xmm_crc2, xmm_crc3, xmm_crc_part;
+    __m128i xmm_initial = _mm_cvtsi32_si128(init_crc);
+    int first = 1;
```
nmoinvaz: Let's start using fixed-width integer types.

cpu_features.h
KungFuJesus: I"m pretty sure we do, it's assigned into functable and we need some assurance that that function is declared somewhere.  We mostly want to prevent the issue of it being multiply declared, as that's a warning that newer GCCs actually treat as an error.  Though, this mostly applies to variable declarations and not so much function prototypes.

arch/x86/crc32_fold_pclmulqdq.c
KungFuJesus: I wanted to avoid having yet another set of branches in the code to jump over the stores.

KungFuJesus: So it's been weighing on my mind a bit - the templated form of this probably wouldn't be too ugly, though the signatures would probably be different, slightly.  I'll see about opening up another PR once this is merged, I should be able to unify this a bit.

And, if this method works out, I'd like to try a few other things:

- There's an obvious place to do exactly the same thing for the adler checksums in inflate_p.h
- aarch64/neon has a vectorized carryless multiplication instruction that can be used as a wider CRC calculation as well.  We could very easily write a version that copies and computes the checksum at the same time for aarch64/neon, where the copy is also similarly nearly free

arch/x86/crc32_fold_pclmulqdq.c
```diff
@@ -380,6 +387,131 @@ Z_INTERNAL void crc32_fold_copy_pclmulqdq(crc32_fold *crc, uint8_t *dst, const u
...
+#define XOR_INITIAL(where) ONCE(where = _mm_xor_si128(where, xmm_initial))
+
+Z_INTERNAL void crc32_fold_pclmulqdq(crc32_fold *crc, const uint8_t *src, size_t len, uint32_t init_crc) {
+    unsigned long algn_diff;
+    __m128i xmm_t0, xmm_t1, xmm_t2, xmm_t3;
+    __m128i xmm_crc0, xmm_crc1, xmm_crc2, xmm_crc3, xmm_crc_part;
+    __m128i xmm_initial = _mm_cvtsi32_si128(init_crc);
+    int32_t first = 1;
+
+    assert(len >= 64);
```
nmoinvaz: Is this `assert` correct? I think this function could handle `len < 64`.

arch/x86/crc32_fold_pclmulqdq.c
```diff
@@ -5,6 +5,7 @@
  * A white paper describing this algorithm can be found at:
  * https://www.intel.com/content/dam/www/public/us/en/documents/white-papers/fast-crc-computation-generic-polynomials-pclmulqdq-paper.pdf
  *
+ * Copyright (c) 2016 Marian Beermann (support for initial value)
```
nmoinvaz: Uppercase `(C)` to be similar to line below it? Also maybe include Intel Copyright first..

arch/x86/crc32_fold_pclmulqdq.c
KungFuJesus: I think it'd need to be at least 31, though technically we never call it for input sizes under 64 bytes.  The reason being is for how we handle the aligning partial load.  Instead of xor'ing the input bytes on the partial fold (where we don't yet have enough input), we do it on the the next set of bytes that will have at least 4 to accommodate the 4 byte initial CRC.

KungFuJesus: Adjusted assertion statement and added explanatory comment in case someone gets too bold with it.

arch/x86/crc32_fold_pclmulqdq.c
```diff
@@ -380,6 +387,135 @@ Z_INTERNAL void crc32_fold_copy_pclmulqdq(crc32_fold *crc, uint8_t *dst, const u
...
+#define XOR_INITIAL(where) ONCE(where = _mm_xor_si128(where, xmm_initial))
+
+Z_INTERNAL void crc32_fold_pclmulqdq(crc32_fold *crc, const uint8_t *src, size_t len, uint32_t init_crc) {
+    unsigned long algn_diff;
+    __m128i xmm_t0, xmm_t1, xmm_t2, xmm_t3;
+    __m128i xmm_crc0, xmm_crc1, xmm_crc2, xmm_crc3, xmm_crc_part;
+    __m128i xmm_initial = _mm_cvtsi32_si128(init_crc);
+    int32_t first = 1;
+
+    /* Technically the CRC functions don't even call this for input < 64, but a bare minimum of 31
```
nmoinvaz, line 403: You are correct that `functable.crc32` will not call this for input < 64. However, `functable.crc32_fold_copy` can be called from inflate/deflate directly with less than 64 bytes. Or maybe I am missing something?

KungFuJesus: crc32_fold_copy doesn't have to deal with an initial CRC

Dead2:
## Baseline 43dbfd6709fb3a8028430ea30f3da88fbeb3ced9
```
 Tool: minigzip   Levels: 0-9

 0    100.008%      0.007/0.014/0.017/0.002        0.014/0.018/0.021/0.002       15,737,543
 1     54.166%      0.207/0.220/0.225/0.004        0.100/0.114/0.119/0.004        8,523,732
 2     43.871%      0.421/0.430/0.434/0.003        0.111/0.119/0.124/0.004        6,903,609
 3     42.387%      0.539/0.547/0.552/0.004        0.098/0.114/0.118/0.005        6,670,099
 4     41.647%      0.592/0.601/0.604/0.004        0.104/0.113/0.117/0.004        6,553,723
 5     41.216%      0.652/0.659/0.666/0.004        0.099/0.111/0.116/0.003        6,485,938
 6     41.037%      0.735/0.745/0.751/0.003        0.094/0.108/0.115/0.006        6,457,776
 7     40.778%      0.920/0.935/0.942/0.006        0.097/0.109/0.114/0.004        6,416,919
 8     40.704%      1.138/1.146/1.152/0.004        0.100/0.109/0.113/0.004        6,405,244
 9     40.409%      1.350/1.362/1.367/0.004        0.103/0.112/0.117/0.003        6,358,951

 avg1  48.622%                        0.666                          0.103
 avg2  54.025%                        0.740                          0.114
 tot                                199.801                         30.771       76,513,534

 130831    1456      48  132335   204ef libz-ng.so.2
```
## PR #1161 88505b7bb4caddc35329b04698aa9e95f239a91e
```
 0    100.008%      0.010/0.014/0.017/0.002        0.014/0.018/0.020/0.002       15,737,543
 1     54.166%      0.212/0.220/0.226/0.004        0.103/0.114/0.119/0.004        8,523,732
 2     43.871%      0.416/0.431/0.436/0.005        0.107/0.117/0.121/0.004        6,903,609
 3     42.387%      0.528/0.536/0.541/0.004        0.098/0.114/0.122/0.007        6,670,099
 4     41.647%      0.574/0.589/0.595/0.006        0.104/0.113/0.117/0.003        6,553,723
 5     41.216%      0.647/0.661/0.667/0.006        0.103/0.109/0.113/0.003        6,485,938
 6     41.037%      0.729/0.748/0.753/0.005        0.094/0.110/0.118/0.005        6,457,776
 7     40.778%      0.929/0.940/0.947/0.005        0.093/0.107/0.110/0.005        6,416,919
 8     40.704%      1.130/1.149/1.156/0.006        0.103/0.109/0.113/0.003        6,405,244
 9     40.409%      1.365/1.372/1.377/0.003        0.104/0.112/0.117/0.004        6,358,951

 avg1  48.622%                        0.666                          0.102
 avg2  54.025%                        0.740                          0.114
 tot                                199.808                         30.662       76,513,534

 132309    1464      48  133821   20abd libz-ng.so.2
```

A bit bigger code size. Also 8 bytes of extra data, not sure where that came from.
As expected, I see no ill effects in any of my tests.

Dead2 (approved):
LGTM

PR #1164, opened 2022-02-11
Benchmarks compilation fix on older compilers

Dead2:
Fix compilation of benchmark when compiler supports, but does not default to enable C++11 or higher.

### DIFF
--- a/test/benchmarks/CMakeLists.txt
+++ b/test/benchmarks/CMakeLists.txt
@@ -2,6 +2,9 @@ cmake_minimum_required(VERSION 3.12)
 
 include(FetchContent)
 
+set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS ON)
 enable_language(CXX)
 
 # Search for Google benchmark package


nmoinvaz:
Looks like codecov is working again.

test/benchmarks/CMakeLists.txt
```diff
@@ -4,6 +4,21 @@ include(FetchContent)
 
 enable_language(CXX)
 
+# If not specified on the command line, enable C++11 as the default
+if(NOT CMAKE_CXX_STANDARD)
```
nmoinvaz: I would just ignore any incoming values. Just set all the values that are expected. Just no need to `if(NOT ..)` for everything imo.

test/benchmarks/CMakeLists.txt
```diff
@@ -3,6 +3,9 @@ cmake_minimum_required(VERSION 3.12)
 include(FetchContent)
 
 enable_language(CXX)
+set(CMAKE_CXX_STANDARD 11)          # The CXX standard required to build this target
```
nmoinvaz: I am not so sure the comments are necessary.

I am wondering if these should be set/enabled before the `enable_language` call.

Dead2: It fixed my compilation problems as it was, but I changed the order anyways to avoid any possible confusion.

nmoinvaz (approved):

PR #1166, opened 2022-02-12
Improved adler32 NEON performance by 30-47%

KungFuJesus:
We unlocked some ILP by allowing for independent sums in the loop and
reducing these sums outside of the loop. Additionally, the multiplication
by 32 (now 64) is moved outside of this loop.  We have enough register
headroom that we could additionally track some more of the sums in the loop
separately but it's suspected that the widening pairwise additions are the
actual bottlenecks here.  It might be possible with additional branching to do
half of the pairwise additions by summing to 16 bit intermediates but this requires
a fair bit of bookkeeping to get correct. The 16 bit intermediates every 128 loop
iterations need to be summed into 32 bit values to not overflow.

On the Odroid-N2, the Cortex-A73 cores are ~20-25% faster on the adler32 benchmark,
and the Cortex-A53 cores are anywhere from 25-30% faster.

### DIFF
--- a/arch/arm/adler32_neon.c
+++ b/arch/arm/adler32_neon.c
@@ -1,7 +1,8 @@
 /* Copyright (C) 1995-2011, 2016 Mark Adler
  * Copyright (C) 2017 ARM Holdings Inc.
- * Author: Adenilson Cavalcanti <adenilson.cavalcanti@arm.com>
- *
+ * Authors: 
+ *   Adenilson Cavalcanti <adenilson.cavalcanti@arm.com>
+ *   Adam Stylinski <kungfujesus06@gmail.com>
  * For conditions of distribution and use, see copyright notice in zlib.h
  */
 #ifdef ARM_NEON_ADLER32
@@ -12,50 +13,124 @@
 #endif
 #include "../../zbuild.h"
 #include "../../adler32_p.h"
+#include "../../fallback_builtins.h"
 
 static void NEON_accum32(uint32_t *s, const unsigned char *buf, size_t len) {
-    static const uint8_t taps[32] = {
+    static const uint16_t ALIGNED_(16) taps[64] = {
+        64, 63, 62, 61, 60, 59, 58, 57,
+        56, 55, 54, 53, 52, 51, 50, 49,
+        48, 47, 46, 45, 44, 43, 42, 41,
+        40, 39, 38, 37, 36, 35, 34, 33,
         32, 31, 30, 29, 28, 27, 26, 25,
         24, 23, 22, 21, 20, 19, 18, 17,
-        16, 15, 14, 13, 12, 11, 10, 9,
+        16, 15, 14, 13, 12, 11, 10, 9, 
         8, 7, 6, 5, 4, 3, 2, 1 };
 
-    uint32x2_t adacc2, s2acc2, as;
-    uint8x16_t t0 = vld1q_u8(taps), t1 = vld1q_u8(taps + 16);
+    uint32x4_t adacc = vdupq_n_u32(0);
+    uint32x4_t s2acc = vdupq_n_u32(0);
+    uint32x4_t s2acc_0 = vdupq_n_u32(0);
+    uint32x4_t s2acc_1 = vdupq_n_u32(0);
+    uint32x4_t s2acc_2 = vdupq_n_u32(0);
 
-    uint32x4_t adacc = vdupq_n_u32(0), s2acc = vdupq_n_u32(0);
     adacc = vsetq_lane_u32(s[0], adacc, 0);
     s2acc = vsetq_lane_u32(s[1], s2acc, 0);
 
-    while (len >= 2) {
-        uint8x16_t d0 = vld1q_u8(buf), d1 = vld1q_u8(buf + 16);
-        uint16x8_t adler, sum2;
-        s2acc = vaddq_u32(s2acc, vshlq_n_u32(adacc, 5));
-        adler = vpaddlq_u8(       d0);
-        adler = vpadalq_u8(adler, d1);
-        sum2 = vmull_u8(      vget_low_u8(t0), vget_low_u8(d0));
-        sum2 = vmlal_u8(sum2, vget_high_u8(t0), vget_high_u8(d0));
-        sum2 = vmlal_u8(sum2, vget_low_u8(t1), vget_low_u8(d1));
-        sum2 = vmlal_u8(sum2, vget_high_u8(t1), vget_high_u8(d1));
-        adacc = vpadalq_u16(adacc, adler);
-        s2acc = vpadalq_u16(s2acc, sum2);
-        len -= 2;
-        buf += 32;
+    uint32x4_t s3acc = vdupq_n_u32(0);
+    uint32x4_t adacc_prev = adacc;
+
+    uint16x8_t s2_0, s2_1, s2_2, s2_3;
+    s2_0 = s2_1 = s2_2 = s2_3 = vdupq_n_u16(0);
+
+    uint16x8_t s2_4, s2_5, s2_6, s2_7;
+    s2_4 = s2_5 = s2_6 = s2_7 = vdupq_n_u16(0);
+
+    int num_iter = len >> 2;
+    int rem = len & 3;
+
+    for (int i = 0; i < num_iter; ++i) {
+        uint8x16x4_t d0_d3 = vld1q_u8_x4(buf);
+
+        /* Unfortunately it doesn't look like there's a direct sum 8 bit to 32
+         * bit instruction, we'll have to make due summing to 16 bits first */
+        uint16x8x2_t hsum, hsum_fold;
+        hsum.val[0] = vpaddlq_u8(d0_d3.val[0]);
+        hsum.val[1] = vpaddlq_u8(d0_d3.val[1]);
+
+        hsum_fold.val[0] = vpadalq_u8(hsum.val[0], d0_d3.val[2]);
+        hsum_fold.val[1] = vpadalq_u8(hsum.val[1], d0_d3.val[3]);
+
+        adacc = vpadalq_u16(adacc, hsum_fold.val[0]); 
+        s3acc = vaddq_u32(s3acc, adacc_prev);
+        adacc = vpadalq_u16(adacc, hsum_fold.val[1]); 
+        
+        /* If we do straight widening additions to the 16 bit values, we don't incur
+         * the usual penalties of a pairwise add. We can defer the multiplications
+         * until the very end. These will not overflow because we are incurring at
+         * most 408 loop iterations (NMAX / 64), and a given lane is only going to be
+         * summed into once. This means for the maximum input size, the largest value
+         * we will see is 255 * 102 = 26010, safely under uint16 max */
+        s2_0 = vaddw_u8(s2_0, vget_low_u8(d0_d3.val[0])); 
+        s2_1 = vaddw_high_u8(s2_1, d0_d3.val[0]);
+        s2_2 = vaddw_u8(s2_2, vget_low_u8(d0_d3.val[1]));
+        s2_3 = vaddw_high_u8(s2_3, d0_d3.val[1]);
+        s2_4 = vaddw_u8(s2_4, vget_low_u8(d0_d3.val[2])); 
+        s2_5 = vaddw_high_u8(s2_5, d0_d3.val[2]);
+        s2_6 = vaddw_u8(s2_6, vget_low_u8(d0_d3.val[3]));
+        s2_7 = vaddw_high_u8(s2_7, d0_d3.val[3]);
+
+        adacc_prev = adacc;
+        buf += 64;
     }
 
-    while (len > 0) {
-        uint8x16_t d0 = vld1q_u8(buf);
-        uint16x8_t adler, sum2;
-        s2acc = vaddq_u32(s2acc, vshlq_n_u32(adacc, 4));
-        adler = vpaddlq_u8(d0);
-        sum2 = vmull_u8(      vget_low_u8(t1), vget_low_u8(d0));
-        sum2 = vmlal_u8(sum2, vget_high_u8(t1), vget_high_u8(d0));
-        adacc = vpadalq_u16(adacc, adler);
-        s2acc = vpadalq_u16(s2acc, sum2);
-        buf += 16;
-        len--;
+    s3acc = vshlq_n_u32(s3acc, 6);
+
+    if (rem) {
+        uint32x4_t s3acc_0 = vdupq_n_u32(0);
+        while (rem--) {
+            uint8x16_t d0 = vld1q_u8(buf);
+            uint16x8_t adler;
+            adler = vpaddlq_u8(d0);
+            s2_6 = vaddw_u8(s2_6, vget_low_u8(d0));
+            s2_7 = vaddw_high_u8(s2_7, d0);
+            adacc = vpadalq_u16(adacc, adler);
+            s3acc_0 = vaddq_u32(s3acc_0, adacc_prev);
+            adacc_prev = adacc;
+            buf += 16;
+        }
+
+        s3acc_0 = vshlq_n_u32(s3acc_0, 4);
+        s3acc = vaddq_u32(s3acc_0, s3acc);
     }
 
+    uint16x8x4_t t0_t3 = vld1q_u16_x4(taps);
+    uint16x8x4_t t4_t7 = vld1q_u16_x4(taps + 32);
+
+    s2acc = vmlal_high_u16(s2acc, t0_t3.val[0], s2_0);
+    s2acc_0 = vmlal_u16(s2acc_0, vget_low_u16(t0_t3.val[0]), vget_low_u16(s2_0));
+    s2acc_1 = vmlal_high_u16(s2acc_1, t0_t3.val[1], s2_1);
+    s2acc_2 = vmlal_u16(s2acc_2, vget_low_u16(t0_t3.val[1]), vget_low_u16(s2_1));
+
+    s2acc = vmlal_high_u16(s2acc, t0_t3.val[2], s2_2);
+    s2acc_0 = vmlal_u16(s2acc_0, vget_low_u16(t0_t3.val[2]), vget_low_u16(s2_2));
+    s2acc_1 = vmlal_high_u16(s2acc_1, t0_t3.val[3], s2_3);
+    s2acc_2 = vmlal_u16(s2acc_2, vget_low_u16(t0_t3.val[3]), vget_low_u16(s2_3));
+
+    s2acc = vmlal_high_u16(s2acc, t4_t7.val[0], s2_4);
+    s2acc_0 = vmlal_u16(s2acc_0, vget_low_u16(t4_t7.val[0]), vget_low_u16(s2_4));
+    s2acc_1 = vmlal_high_u16(s2acc_1, t4_t7.val[1], s2_5);
+    s2acc_2 = vmlal_u16(s2acc_2, vget_low_u16(t4_t7.val[1]), vget_low_u16(s2_5));
+
+    s2acc = vmlal_high_u16(s2acc, t4_t7.val[2], s2_6);
+    s2acc_0 = vmlal_u16(s2acc_0, vget_low_u16(t4_t7.val[2]), vget_low_u16(s2_6));
+    s2acc_1 = vmlal_high_u16(s2acc_1, t4_t7.val[3], s2_7);
+    s2acc_2 = vmlal_u16(s2acc_2, vget_low_u16(t4_t7.val[3]), vget_low_u16(s2_7));
+
+    s2acc = vaddq_u32(s2acc_0, s2acc);
+    s2acc_2 = vaddq_u32(s2acc_1, s2acc_2);
+    s2acc = vaddq_u32(s2acc, s2acc_2);
+
+    uint32x2_t adacc2, s2acc2, as;
+    s2acc = vaddq_u32(s2acc, s3acc);
     adacc2 = vpadd_u32(vget_low_u32(adacc), vget_high_u32(adacc));
     s2acc2 = vpadd_u32(vget_low_u32(s2acc), vget_high_u32(s2acc));
     as = vpadd_u32(adacc2, s2acc2);
@@ -91,7 +166,6 @@ uint32_t adler32_neon(uint32_t adler, const unsigned char *buf, size_t len) {
     uint32_t pair[2];
     int n = NMAX;
     unsigned int done = 0;
-    unsigned int i;
 
     /* Split Adler-32 into component sums, it can be supplied by
      * the caller sites (e.g. in a PNG file).
@@ -99,18 +173,37 @@ uint32_t adler32_neon(uint32_t adler, const unsigned char *buf, size_t len) {
     pair[0] = adler;
     pair[1] = sum2;
 
-    for (i = 0; i < len; i += n) {
-        if ((i + n) > len)
-            n = (int)(len - i);
+    /* If memory is not SIMD aligned, do scalar sums to an aligned
+     * offset, provided that doing so doesn't completely eliminate
+     * SIMD operation. Aligned loads are still faster on ARM, even
+     * though there's no explicit aligned load instruction */
+    unsigned int align_offset = ((uintptr_t)buf & 15);
+    unsigned int align_adj = (align_offset) ? 16 - align_offset : 0;
+
+    if (align_offset && len >= (16 + align_adj)) {
+        NEON_handle_tail(pair, buf, align_adj);
+        n -= align_adj;
+        done += align_adj;
+
+    } else {
+        /* If here, we failed the len criteria test, it wouldn't be
+         * worthwhile to do scalar aligning sums */
+        align_adj = 0;
+    }
+
+    while (done < len) {
+        int remaining = (int)(len - done);
+        n = MIN(remaining, (done == align_adj) ? n : NMAX);
 
         if (n < 16)
             break;
 
-        NEON_accum32(pair, buf + i, n / 16);
+        NEON_accum32(pair, buf + done, n >> 4);
         pair[0] %= BASE;
         pair[1] %= BASE;
 
-        done += (n / 16) * 16;
+        int actual_nsums = (n >> 4) << 4;
+        done += actual_nsums;
     }
 
     /* Handle the tail elements. */
@@ -123,4 +216,5 @@ uint32_t adler32_neon(uint32_t adler, const unsigned char *buf, size_t len) {
     /* D = B * 65536 + A, see: https://en.wikipedia.org/wiki/Adler-32. */
     return (pair[1] << 16) | pair[0];
 }
+
 #endif
--- a/fallback_builtins.h
+++ b/fallback_builtins.h
@@ -64,6 +64,14 @@ static inline __m512i _mm512_zextsi128_si512(__m128i a) {
 
 #endif // __AVX2__
 
+#if defined(ARM_NEON_ADLER32) && !defined(__aarch64__)
+/* Compatibility shim for the _high family of functions */
+#define vmull_high_u8(a, b) vmull_u8(vget_high_u8(a), vget_high_u8(b))
+#define vmlal_high_u8(a, b, c) vmlal_u8(a, vget_high_u8(b), vget_high_u8(c))
+#define vmlal_high_u16(a, b, c) vmlal_u16(a, vget_high_u16(b), vget_high_u16(c))
+#define vaddw_high_u8(a, b) vaddw_u8(a, vget_high_u8(b))
+#endif
+
 #ifdef ARM_NEON_SLIDEHASH
 
 #define vqsubq_u16_x4_x1(out, a, b) do { \
@@ -90,6 +98,15 @@ static inline uint16x8x4_t vld1q_u16_x4(uint16_t *a) {
                           vld1q_u16(a+8),
                           vld1q_u16(a+16),
                           vld1q_u16(a+24)}};
+    return ret;
+}
+
+static inline uint8x16x4_t vld1q_u8_x4(uint8_t *a) {
+    uint8x16x4_t ret = (uint8x16x4_t) {{
+                          vld1q_u8(a),
+                          vld1q_u8(a+16),
+                          vld1q_u8(a+32),
+                          vld1q_u8(a+48)}};
     return ret;
 }
 


KungFuJesus:
Comparison for A73:
a73cmp

Comparison for A53:
a53cmp

nmoinvaz:
Wow, nice work again. You must have access to a lot of different CPUs!

KungFuJesus:
> Wow, nice work again. You must have access to a lot of different CPUs!

Well the Odroid is big.LITTLE, so it has 2 different ARM SIPs on.  But yeah, I have a pretty decent boutique collection, hah.  I don't think what I have is too terribly exotic - the Powermac G5 and the SPARC-T4 may be the more unique of them (though far from _rare_).

KungFuJesus:
The 16 bit running sums I mentioned earlier in the commit message...I tried that it got kind of messy and slower.  The widening sum to a larger precision appears to be a rather costly instruction (at least for the A72, according to this: https://documentation-service.arm.com/static/5ed75eeeca06a95ce53f93c7?token=)

However, their little footnote confusingly seems to imply that this latency is somehow pipelined away such that the total throughput ends up being 1 per cycle for the accumulation part.  So, what we end possibly losing on the pairwise sums we end up possibly making up for by aggregating the adds in parallel to the 32 bit sums.  Knowing this, we might be able to do at least marginally better by not leveraging the single operand pairwise adds at all.

For the "s2" sums, we come dangerously close to overflowing pretty early with those "tap" constants, but I'm pretty sure I did the math there such that it works out given where adjacent elements end up landing for the folding part of it.

KungFuJesus:
Hmm, I took a stab at it and couldn't get anything but regressions.  There might be a more efficient sequence here that I'm missing but I think I'm at about as efficient as I'm going to get for today.

KungFuJesus:
Ughh...GCC was annoyingly late to implementing all of these intrinsics.  Now I have to figure out when they were introduced and add a fallback function, again...

Oh fun, it's just that it's aarch64 only.  Well, I can add in a compatibility shim, since it's certainly advantageous to be able to use an instruction that operates over only the high lanes of a given vector instead of requiring a separate extraction.

KungFuJesus:
Well that's really weird - adjusting the n loop iterations that can run during the first round the loop seems to break the fuzzer test against the CMakeLists.txt file and alter the checksums.  I'm doing this to avoid a rebase back to the prime number base of the adler checksum, since it's the maximum number of scalar sums that can be performed with max input bytes before we roll over 32 bits.  

@mtl1979 You mainly did the code for the loop on powerpc, is there something I'm doing that's really wrong?  To make things work I just rebase again, but the commented out bits are still there that conditionally adjust n for when at the beginning of the loop.  The checksums do end up being different but I can't for the life of me figure out why.  Even weirder, I did this same optimization with the VMX code and that doesn't seem to be failing the fuzzer at all (despite it having identical vector widths).

KungFuJesus:
Hmm, I think the logic surrounding n v nmax is correct which leads me to believe the fuzzer is catching some corner case in the remainder peeling or something.  I tried to make a similar length buffer to see if I could trigger a non-matching checksum with the C variant and couldn't initially.

I may try writing a more thorough fuzzing that infinitely tries random buffers, buffer lengths, and alignment offsets and evaluates them against the scalar C to see if I can do create the problem.

I also might try excising the buffer that's parsed in this fuzzer test right now that's causing the issue to see if I step through them line by line and find an error.

mtl1979:
@KungFuJesus Main difference between ARM and PowerPC is that ARM is little-endian and PowerPC is big-endian, which can result data getting loaded in different order. My bet is still overflow due to a corner case. With this wide loads few more bytes due to aligning the buffer can be enough to cause overflow even if you try to adjust the first iteration. It all becomes to the question what was the adler32 value when entering the function, if one or both of the components was close to maximum, then it will definitely overflow.

When adjusting the first iteration, we need to be careful not to round towards zero, as it will result in adjusting being cancelled. Wider the loads, more careful we need to be, as we're basically counting number of remaining iterations, not number of bytes.

KungFuJesus:
> @KungFuJesus Main difference between ARM and PowerPC is that ARM is little-endian and PowerPC is big-endian, which can result data getting loaded in different order. My bet is still overflow due to a corner case. With this wide loads few more bytes due to aligning the buffer can be enough to cause overflow even if you try to adjust the first iteration. It all becomes to the question what was the adler32 value when entering the function, if one or both of the components was close to maximum, then it will definitely overflow.
> 
> When adjusting the first iteration, we need to be careful not to round towards zero, as it will result in adjusting being cancelled. Wider the loads, more careful we need to be, as we're basically counting number of remaining iterations, not number of bytes.

I suspect looking for overflow might be the right track. I'll bet I'm overflowing on the 16 bit multiply-accumulates since they can't accumulate back to 32 bit values (which is the main difference between this vectorization and all the others). The multiplication is safe but I think that one of the accumulations are biting us. I can probably get around it by modifying where the larger products accumulate back to, though.

mtl1979:
@KungFuJesus First hunch is often the correct one... Have to eliminate the other possibilities first and the remaining one is the culprit... Like I said before, every time something doesn't work as we planned, we learn more.

KungFuJesus:
Interesting...it might actually be in the remainder peeling.  I modified the unrolled loop so that it's skipped over and basically does the remainder peeling (with no unroll factor) and I still see the issue.  This is very odd.

KungFuJesus:
It really doesn't _seem_ like there's any potential overflow in here:
```
    if (rem) {                                                                                   
        adacc_prev = vshlq_n_u32(adacc_prev, 4);                                                 
        while (rem--) {                                                                          
            uint8x16_t d0 = vld1q_u8(buf);                                                       
            uint16x8_t adler, sum2, sum2_lo;                                                     
            adler = vpaddlq_u8(d0);                                                              
            sum2 = vmull_high_u8(t0_t3.val[3], d0);                                              
            sum2_lo = vmull_u8(vget_low_u8(t0_t3.val[3]), vget_low_u8(d0));                      
            adacc = vpadalq_u16(adacc, adler);                                                   
            s2acc = vpadalq_u16(s2acc, sum2);                                                    
            s3acc = vaddq_u32(s3acc, adacc_prev);                                                
            s2acc = vpadalq_u16(s2acc, sum2_lo);                                                 
            adacc_prev = vshlq_n_u32(adacc, 4);                                                  
            buf += 16;                                                                           
        }                                                                                        
    } 
```

That loop is even free of multiply accumulates - yet if I trim off the first 13 iterations in scalar sums and then subtract it from the max loop iterations I get erroneous checksums.

KungFuJesus:
Hah, bringing in the old version of the inlined function, even, results in incorrect behavior.  Now it may be presumptuous to assume the original implementation was perfectly correct but it does seem like something is up with this sum if I try to trim unaligned starting bytes.

This is really weird.

mtl1979:
> That loop is even free of multiply accumulates - yet if I trim off the first 13 iterations in scalar sums and then subtract it from the max loop iterations I get erroneous checksums.

That would still imply that the modulo becomes no-op... We basically assume that the modulo is never no-op and that makes sure we never overflow even if we do wide loads.

KungFuJesus:
But it _is_ a no-op, effectively.  NMAX is derived based on the maximum number of scalar sums.  I'm not sure what assumptions it makes about the starting sums but even so, using NMAX-1 or NMAX-50 still results in the bug.  Hell, even doing the modulus anyway and subtracting out the initial scalar runs of the loop still results in it being wrong.

mtl1979:
Starting sum should be in range of 0...4294961743... That means before the modulo, the 32-bit sum should be between 4294961743 and 4294967295.

KungFuJesus:
> Starting sum should be in range of 0...4294961743... That means before the modulo, the 32-bit sum should be between 4294961743 and 4294967295.

I see what you mean but even if I force it to do the first modulo operation anyway and still reduce the max first sums it ends up being wrong, which is really strange.

mtl1979:
> > Starting sum should be in range of 0...4294961743... That means before the modulo, the 32-bit sum should be between 4294961743 and 4294967295.
> 
> I see what you mean but even if I force it to do the first modulo operation anyway and still reduce the max first sums it ends up being wrong, which is really strange.

That would mean it needs to process "tail" bytes before hitting exactly NMAX... Basically the logic assumes the alignment adjust + tail bytes equal 16 bytes.

KungFuJesus:
> > > Starting sum should be in range of 0...4294961743... That means before the modulo, the 32-bit sum should be between 4294961743 and 4294967295.
> > 
> > 
> > I see what you mean but even if I force it to do the first modulo operation anyway and still reduce the max first sums it ends up being wrong, which is really strange.
> 
> That would mean it needs to process "tail" bytes before hitting exactly NMAX... Basically the logic assumes the alignment adjust + tail bytes equal 16 bytes.

I _thought_ that was being handled via the done variable, no?

KungFuJesus:
It's not the end of the world performance wise having to do one extra pair of modulo operations in GPRs here but I sure would like to understand why this doesn't work. If nothing else to ensure I'm more confident in the solution.

mtl1979:
> I _thought_ that was being handled via the done variable, no?

Not quite... that handling is too late... Basically handling it correctly would result in next iteration being unaligned again...

KungFuJesus:
Can you show me what you mean exactly? I'm not quite sure what I'm missing here

mtl1979:
@KungFuJesus Where it handles `done`, the issue can be that `len - done` is more than what is remaining before hitting `NMAX`. Or to put it in another way, the difference is more than `16 - align_adj`.

KungFuJesus:
> @KungFuJesus Where it handles `done`, the issue can be that `len - done` is more than what is remaining before hitting `NMAX`. Or to put it in another way, the difference is more than `16 - align_adj`.

But it isn't, though.  It breaks out of the loop only if what is remaining is less than 16.  N is conditionally modified inside the loop such that only on the first iteration it's adjusted to be NMAX-align_adj.  The done amount is accounted for before going into the loop, and the component sums are rebased at the bottom of the loop.  Even allowing it to rebase after 16-align_offset sums doesn't seem to result in correct behavior.  The loop counter i is tracking exactly the number of bytes that have been summed, and we start the loop from wherever the alignment offset began.  It terminates when i finally reaches len or such that i + 16 > len, in which case the tail processing should be handling it below.

KungFuJesus:
ahhhh, I think I see the point you were making about n.  We need to increment i by the number of sums we actually accomplished rather than actually n.

Hmm, I think this is a potential bug in the powerpc code as well, now.

mtl1979:
> ahhhh, I think I see the point you were making about n. We need to increment i by the number of sums we actually accomplished rather than actually n.
> 
> Hmm, I think this is a potential bug in the powerpc code as well, now.

Sometimes it's really hard to explain things when I do hundreds of iterations in my head instead of using paper like most people... At my first job back in early 1998 I had to do 1000 pages of calculations in my head.

KungFuJesus:
@Dead2 The issues with this PR are fixed, ready for your review.

KungFuJesus:
Ooof, doing the math here we're very easily overflowing uint16 for the top 2 taps - I'll need to come back to this.  I did find a way to get even better performance but in doing so I found the very obvious overflow hazard.  Obviously none of the tests are hitting it.  

Arm needs a sum to from 8 bit to 32 bit precision instruction badly :-/.

KungFuJesus:
Err no, per lane, we're ok.  In fact, my optimization still fits in here.  I only actually need to account for the top first lane that falls on the highest value of the those multiplication constants in the "taps".  Everything below that actually can fit, too, if we assume maximum byte values of 255.

I had forgotten that adjacent values aren't folded together with this during the widening process.

mtl1979:
@KungFuJesus I was expecting things to overflow if we try 256 bytes per iteration... Even 128 bytes per iteration shouldn't overflow, even though it gets pretty close...

KungFuJesus:
> @KungFuJesus I was expecting things to overflow if we try 256 bytes per iteration... Even 128 bytes per iteration shouldn't overflow, even though it gets pretty close...

Yep, see my comment where the accumulates are happening.

mtl1979:
> > @KungFuJesus I was expecting things to overflow if we try 256 bytes per iteration... Even 128 bytes per iteration shouldn't overflow, even though it gets pretty close...
> 
> Yep, see my comment where the accumulates are happening.

I noticed you did the addition of adjacent 4 lanes with 16 bits precision, even though with only 2 lanes at a time and the second addition with 32 bits precision there would be no risk of overflow...

KungFuJesus:
> > > @KungFuJesus I was expecting things to overflow if we try 256 bytes per iteration... Even 128 bytes per iteration shouldn't overflow, even though it gets pretty close...
> > 
> > 
> > Yep, see my comment where the accumulates are happening.
> 
> I noticed you did the addition of adjacent 4 lanes with 16 bits precision, even though with only 2 lanes at a time and the second addition with 32 bits precision there would be no risk of overflow...

I'm not sure I follow what you're getting at, here.  There's no widening FMA instruction on aarch64 and that's the opportunity being capitalized on with the 16 bit additions.  Immediately adding back to the 32 bit operands with the widening add would result in slower code and less ILP.

mtl1979:
> I'm not sure I follow what you're getting at, here.  There's no widening FMA instruction on aarch64 and that's the opportunity being capitalized on with the 16 bit additions.  Immediately adding back to the 32 bit operands with the widening add would result in slower code and less ILP.  

You're basically alternating between `sum2_a` and `sum2b` which create dependency. You could use 4 variables instead and then sum them pairwise before summing to final variable.

Dead2:
Remember to keep around any corner-case testcases so they can be added as regression tests if possible..

KungFuJesus:
> > I'm not sure I follow what you're getting at, here.  There's no widening FMA instruction on aarch64 and that's the opportunity being capitalized on with the 16 bit additions.  Immediately adding back to the 32 bit operands with the widening add would result in slower code and less ILP.
> 
> You're basically alternating between `sum2_a` and `sum2b` which create dependency. You could use 4 variables instead and then sum them pairwise before summing to final variable.

I was initially using 4 variables for this very reason however it turns out that the multiply-accumulate that occurs afterward (with the fused accumulate) saves more cycles than doing additional vpadalq_u16's.  I suspect that the uadalp instruction is rather expensive and the savings were greater by minimizing that in the loop.

I could maybe try tracking 16 bit sums and adding a branch in here to sum back just before overflow.  I'll see if that's a benefit.

KungFuJesus:
> Remember to keep around any corner-case testcases so they can be added as regression tests if possible..

In this case the corner case was caught by the tests.  I'll need to write a PR for the powerpc code, as I think there's potential for this issue to happen there, as well.  Perhaps a good test case to add to the batch is an array of 64 bytes of '255'.

KungFuJesus:
> I could maybe try tracking 16 bit sums and adding a branch in here to sum back just before overflow. I'll see if that's a benefit.

I tried a version that does this and while it doesn't pass _every_ test for correctness, the branching results in a very clear loss in performance in the loop already.  I don't think that route is worth pursuing, we might even be getting close to the maximum overlap of most modern ARM cores (I think the cortex-a72 can do at most something like 2 integer operations per clock).

KungFuJesus:
It's just unfortunate that ARM doesn't have a widening pairwise multiply-accumulate like power and x86 do (vmsummbm and pmaddwd/pmaddubs).  These seemed like obvious choices in an ISA for operating over a bunch of 8 bits values.  The uadalp instruction is the only widening version it seems and it's strictly an addition.  It gets us something to work with but the benefits are clear if you can widen and fuse the multiply+add (even in x86 they do an implicit pairwise add populate the solution into lower half of the lanes as though they were 32 bits).  My core 2 quad manages about 2x performance with a 30% higher clock and the same vector widths.  The Powermac G5 is in a similar boat at 2.5GHz (even marginally faster than the slightly newer core 2 quad).  Granted both of these machines have a ridiculously higher power envelope, but they are significantly older.  

We're within spitting distance of those two machines, so I guess we can't complain too much.  A friend of mine tried this benchmark on an M1 mac the performance came dangerously close to AVX2's times on my HEDT chip (60ish ns for the 2k input).  Perhaps this speaks volumes about a simplified decode stage when put on a higher power budget.

KungFuJesus:
Hah, I just found a better way that's similar to the way chromium does this (but still is faster).  I'll push it shortly after some more testing.

KungFuJesus:
Cortex-A53:
```
Comparing adler32/neon (from ./benchmark_zlib_old) to adler32/neon (from ./benchmark_zlib)
Benchmark                                                        Time             CPU      Time Old      Time New       CPU Old       CPU New
---------------------------------------------------------------------------------------------------------------------------------------------
[adler32/neon vs. adler32/neon]/2048_pvalue                    0.0000          0.0000      U Test, Repetitions: 50 vs 50
[adler32/neon vs. adler32/neon]/2048_mean                     -0.4456         -0.4456           918           509           905           502
[adler32/neon vs. adler32/neon]/2048_median                   -0.4458         -0.4456           918           509           905           502
[adler32/neon vs. adler32/neon]/2048_stddev                   -0.4564         -0.4809             2             1             1             0
[adler32/neon vs. adler32/neon]/2048_cv                       -0.0195         -0.0636             0             0             0             0
[adler32/neon vs. adler32/neon]/4096_pvalue                    0.0000          0.0000      U Test, Repetitions: 50 vs 50
[adler32/neon vs. adler32/neon]/4096_mean                     -0.4724         -0.4722          1801           950          1776           937
[adler32/neon vs. adler32/neon]/4096_median                   -0.4720         -0.4722          1801           951          1776           937
[adler32/neon vs. adler32/neon]/4096_stddev                   -0.4137         -0.3935             4             3             1             1
[adler32/neon vs. adler32/neon]/4096_cv                       +0.1112         +0.1492             0             0             0             0
[adler32/neon vs. adler32/neon]/32768_pvalue                   0.0000          0.0000      U Test, Repetitions: 50 vs 50
[adler32/neon vs. adler32/neon]/32768_mean                    -0.4608         -0.4608         14746          7951         14529          7834
[adler32/neon vs. adler32/neon]/32768_median                  -0.4607         -0.4608         14752          7956         14532          7836
[adler32/neon vs. adler32/neon]/32768_stddev                  -0.4229         -0.3982            40            23            10             6
[adler32/neon vs. adler32/neon]/32768_cv                      +0.0704         +0.1161             0             0             0             0
[adler32/neon vs. adler32/neon]/262144_pvalue                  0.0000          0.0000      U Test, Repetitions: 50 vs 50
[adler32/neon vs. adler32/neon]/262144_mean                   -0.4356         -0.4370        122223         68979        119436         67238
[adler32/neon vs. adler32/neon]/262144_median                 -0.4343         -0.4357        122195         69130        119318         67328
[adler32/neon vs. adler32/neon]/262144_stddev                 -0.3380         -0.3504          1230           814           971           631
[adler32/neon vs. adler32/neon]/262144_cv                     +0.1729         +0.1540             0             0             0             0
[adler32/neon vs. adler32/neon]/2097152_pvalue                 0.0000          0.0000      U Test, Repetitions: 50 vs 50
[adler32/neon vs. adler32/neon]/2097152_mean                  -0.3178         -0.3187       1375573        938470       1345848        916876
[adler32/neon vs. adler32/neon]/2097152_median                -0.3177         -0.3184       1374351        937657       1344902        916710
[adler32/neon vs. adler32/neon]/2097152_stddev                -0.3634         -0.3778         19925         12684         19173         11929
[adler32/neon vs. adler32/neon]/2097152_cv                    -0.0670         -0.0867             0             0             0             0
[adler32/neon vs. adler32/neon]/4194304_pvalue                 0.0000          0.0000      U Test, Repetitions: 50 vs 50
[adler32/neon vs. adler32/neon]/4194304_mean                  -0.3359         -0.3360       2687522       1784701       2629296       1745939
[adler32/neon vs. adler32/neon]/4194304_median                -0.3368         -0.3375       2684437       1780241       2631286       1743253
[adler32/neon vs. adler32/neon]/4194304_stddev                -0.4465         -0.4806         51368         28432         47981         24923
[adler32/neon vs. adler32/neon]/4194304_cv                    -0.1665         -0.2178             0             0             0             0
OVERALL_GEOMEAN                                               -0.4145         -0.4149             0             0             0             0
```

Cortex-A73:

```
Comparing adler32/neon (from ./benchmark_zlib_old) to adler32/neon (from ./benchmark_zlib)
Benchmark                                                        Time             CPU      Time Old      Time New       CPU Old       CPU New
---------------------------------------------------------------------------------------------------------------------------------------------
[adler32/neon vs. adler32/neon]/2048_pvalue                    0.0000          0.0000      U Test, Repetitions: 50 vs 50
[adler32/neon vs. adler32/neon]/2048_mean                     -0.4024         -0.4023           467           279           463           277
[adler32/neon vs. adler32/neon]/2048_median                   -0.4022         -0.4024           467           279           463           276
[adler32/neon vs. adler32/neon]/2048_stddev                   -0.2239         +0.7229             1             1             0             0
[adler32/neon vs. adler32/neon]/2048_cv                       +0.2987         +1.8825             0             0             0             0
[adler32/neon vs. adler32/neon]/4096_pvalue                    0.0000          0.0000      U Test, Repetitions: 50 vs 50
[adler32/neon vs. adler32/neon]/4096_mean                     -0.4323         -0.4326           904           513           896           508
[adler32/neon vs. adler32/neon]/4096_median                   -0.4325         -0.4326           904           513           896           508
[adler32/neon vs. adler32/neon]/4096_stddev                   -0.4830         -0.5147             1             1             0             0
[adler32/neon vs. adler32/neon]/4096_cv                       -0.0893         -0.1447             0             0             0             0
[adler32/neon vs. adler32/neon]/32768_pvalue                   0.0000          0.0000      U Test, Repetitions: 50 vs 50
[adler32/neon vs. adler32/neon]/32768_mean                    -0.4279         -0.4278          7406          4237          7327          4192
[adler32/neon vs. adler32/neon]/32768_median                  -0.4277         -0.4277          7407          4239          7326          4193
[adler32/neon vs. adler32/neon]/32768_stddev                  -0.4057         -0.3968            16            10             5             3
[adler32/neon vs. adler32/neon]/32768_cv                      +0.0388         +0.0542             0             0             0             0
[adler32/neon vs. adler32/neon]/262144_pvalue                  0.0000          0.0000      U Test, Repetitions: 50 vs 50
[adler32/neon vs. adler32/neon]/262144_mean                   -0.4410         -0.4410         59082         33024         58418         32655
[adler32/neon vs. adler32/neon]/262144_median                 -0.4410         -0.4410         59091         33033         58417         32655
[adler32/neon vs. adler32/neon]/262144_stddev                 -0.4547         -0.4566           112            61            26            14
[adler32/neon vs. adler32/neon]/262144_cv                     -0.0244         -0.0280             0             0             0             0
[adler32/neon vs. adler32/neon]/2097152_pvalue                 0.0000          0.0000      U Test, Repetitions: 50 vs 50
[adler32/neon vs. adler32/neon]/2097152_mean                  -0.3059         -0.3076        540242        374969        526417        364482
[adler32/neon vs. adler32/neon]/2097152_median                -0.3052         -0.3076        540270        375380        527136        364989
[adler32/neon vs. adler32/neon]/2097152_stddev                -0.7277         -0.6536         10204          2778          7621          2640
[adler32/neon vs. adler32/neon]/2097152_cv                    -0.6077         -0.4998             0             0             0             0
[adler32/neon vs. adler32/neon]/4194304_pvalue                 0.0000          0.0000      U Test, Repetitions: 50 vs 50
[adler32/neon vs. adler32/neon]/4194304_mean                  -0.3081         -0.3095       1086807        751946       1056576        729612
[adler32/neon vs. adler32/neon]/4194304_median                -0.3078         -0.3098       1086821        752302       1057253        729710
[adler32/neon vs. adler32/neon]/4194304_stddev                -0.7816         -0.7102         20179          4407         14959          4335
[adler32/neon vs. adler32/neon]/4194304_cv                    -0.6844         -0.5803             0             0             0             0
OVERALL_GEOMEAN                                               -0.3890         -0.3894             0             0             0             0
```

Looks like the M1 may have gotten marginally slower?  I don't have direct access to the system to really figure out what might be going on there, but it's a clear big win for the Cortex family of SIP.

KungFuJesus:
### Cortex-A53
```Comparing adler32/neon (from ./benchmark_zlib_old) to adler32/neon (from ./benchmark_zlib)
Benchmark                                                        Time             CPU      Time Old      Time New       CPU Old       CPU New
---------------------------------------------------------------------------------------------------------------------------------------------
[adler32/neon vs. adler32/neon]/2048_pvalue                    0.0002          0.0002      U Test, Repetitions: 10 vs 10
[adler32/neon vs. adler32/neon]/2048_mean                     -0.4477         -0.4472           918           507           905           500
[adler32/neon vs. adler32/neon]/2048_median                   -0.4484         -0.4472           918           506           905           500
[adler32/neon vs. adler32/neon]/2048_stddev                   -0.2403         +0.4199             2             2             0             1
[adler32/neon vs. adler32/neon]/2048_cv                       +0.3756         +1.5688             0             0             0             0
[adler32/neon vs. adler32/neon]/4096_pvalue                    0.0002          0.0002      U Test, Repetitions: 10 vs 10
[adler32/neon vs. adler32/neon]/4096_mean                     -0.4729         -0.4731          1800           949          1776           935
[adler32/neon vs. adler32/neon]/4096_median                   -0.4730         -0.4732          1801           949          1776           935
[adler32/neon vs. adler32/neon]/4096_stddev                   -0.6842         -0.5982             5             2             1             0
[adler32/neon vs. adler32/neon]/4096_cv                       -0.4009         -0.2373             0             0             0             0
[adler32/neon vs. adler32/neon]/32768_pvalue                   0.0002          0.0002      U Test, Repetitions: 10 vs 10
[adler32/neon vs. adler32/neon]/32768_mean                    -0.4752         -0.4753         14424          7569         14227          7465
[adler32/neon vs. adler32/neon]/32768_median                  -0.4748         -0.4752         14422          7574         14228          7467
[adler32/neon vs. adler32/neon]/32768_stddev                  -0.6055         -0.5020            53            21            13             6
[adler32/neon vs. adler32/neon]/32768_cv                      -0.2482         -0.0508             0             0             0             0
[adler32/neon vs. adler32/neon]/262144_pvalue                  0.0002          0.0002      U Test, Repetitions: 10 vs 10
[adler32/neon vs. adler32/neon]/262144_mean                   -0.4472         -0.4487        121588         67209        119061         65638
[adler32/neon vs. adler32/neon]/262144_median                 -0.4459         -0.4468        121485         67310        118856         65750
[adler32/neon vs. adler32/neon]/262144_stddev                 -0.2633         -0.3717          1197           882           975           613
[adler32/neon vs. adler32/neon]/262144_cv                     +0.3328         +0.1397             0             0             0             0
[adler32/neon vs. adler32/neon]/2097152_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
[adler32/neon vs. adler32/neon]/2097152_mean                  -0.3443         -0.3447       1377615        903244       1349073        884086
[adler32/neon vs. adler32/neon]/2097152_median                -0.3419         -0.3432       1373784        904054       1347699        885172
[adler32/neon vs. adler32/neon]/2097152_stddev                -0.0535         -0.0780         19128         18105         17301         15953
[adler32/neon vs. adler32/neon]/2097152_cv                    +0.4436         +0.4070             0             0             0             0
[adler32/neon vs. adler32/neon]/4194304_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
[adler32/neon vs. adler32/neon]/4194304_mean                  -0.3455         -0.3459       2739827       1793285       2684339       1755776
[adler32/neon vs. adler32/neon]/4194304_median                -0.3508         -0.3524       2757114       1789947       2701683       1749663
[adler32/neon vs. adler32/neon]/4194304_stddev                -0.7487         -0.7746         71300         17921         66435         14973
[adler32/neon vs. adler32/neon]/4194304_cv                    -0.6160         -0.6554             0             0             0             0
OVERALL_GEOMEAN                                               -0.4248         -0.4251             0             0             0             0

```

### Cortex-A73:
```
Comparing adler32/neon (from ./benchmark_zlib_old) to adler32/neon (from ./benchmark_zlib)
Benchmark                                                        Time             CPU      Time Old      Time New       CPU Old       CPU New
---------------------------------------------------------------------------------------------------------------------------------------------
[adler32/neon vs. adler32/neon]/2048_pvalue                    0.0002          0.0002      U Test, Repetitions: 10 vs 10
[adler32/neon vs. adler32/neon]/2048_mean                     -0.4010         -0.4004           467           280           463           277
[adler32/neon vs. adler32/neon]/2048_median                   -0.4011         -0.4005           467           280           463           277
[adler32/neon vs. adler32/neon]/2048_stddev                   -0.4249         -0.2189             1             0             0             0
[adler32/neon vs. adler32/neon]/2048_cv                       -0.0399         +0.3028             0             0             0             0
[adler32/neon vs. adler32/neon]/4096_pvalue                    0.0002          0.0002      U Test, Repetitions: 10 vs 10
[adler32/neon vs. adler32/neon]/4096_mean                     -0.4336         -0.4330           904           512           895           508
[adler32/neon vs. adler32/neon]/4096_median                   -0.4334         -0.4331           904           512           895           508
[adler32/neon vs. adler32/neon]/4096_stddev                   -0.6206         -0.5217             2             1             0             0
[adler32/neon vs. adler32/neon]/4096_cv                       -0.3302         -0.1564             0             0             0             0
[adler32/neon vs. adler32/neon]/32768_pvalue                   0.0002          0.0002      U Test, Repetitions: 10 vs 10
[adler32/neon vs. adler32/neon]/32768_mean                    -0.4330         -0.4330          7412          4203          7343          4163
[adler32/neon vs. adler32/neon]/32768_median                  -0.4330         -0.4331          7412          4203          7343          4163
[adler32/neon vs. adler32/neon]/32768_stddev                  -0.2399         -0.3063             7             5             2             2
[adler32/neon vs. adler32/neon]/32768_cv                      +0.3406         +0.2235             0             0             0             0
[adler32/neon vs. adler32/neon]/262144_pvalue                  0.0002          0.0002      U Test, Repetitions: 10 vs 10
[adler32/neon vs. adler32/neon]/262144_mean                   -0.4455         -0.4450         59145         32796         58473         32452
[adler32/neon vs. adler32/neon]/262144_median                 -0.4453         -0.4450         59145         32810         58478         32453
[adler32/neon vs. adler32/neon]/262144_stddev                 -0.4678         -0.6407            98            52            34            12
[adler32/neon vs. adler32/neon]/262144_cv                     -0.0402         -0.3527             0             0             0             0
[adler32/neon vs. adler32/neon]/2097152_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
[adler32/neon vs. adler32/neon]/2097152_mean                  -0.3023         -0.3106        534381        372849        522874        360494
[adler32/neon vs. adler32/neon]/2097152_median                -0.2929         -0.3031        529538        374457        518423        361286
[adler32/neon vs. adler32/neon]/2097152_stddev                -0.5137         -0.4276         13778          6700         10813          6190
[adler32/neon vs. adler32/neon]/2097152_cv                    -0.3030         -0.1697             0             0             0             0
[adler32/neon vs. adler32/neon]/4194304_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
[adler32/neon vs. adler32/neon]/4194304_mean                  -0.3031         -0.3103       1070766        746172       1044398        720343
[adler32/neon vs. adler32/neon]/4194304_median                -0.3031         -0.3106       1078719        751732       1049211        723301
[adler32/neon vs. adler32/neon]/4194304_stddev                -0.3899         -0.1544         25776         15727         17120         14478
[adler32/neon vs. adler32/neon]/4194304_cv                    -0.1244         +0.2260             0             0             0             0
OVERALL_GEOMEAN                                               -0.3894         -0.3914             0             0             0             0
```

mtl1979:
> Looks like the M1 may have gotten marginally slower? I don't have direct access to the system to really figure out what might be going on there, but it's a clear big win for the Cortex family of SIP.

When trying to optimize for specific processor or group of processors with similar internals, it's always expected that some processors might perform differently and the code gets slower... That's why we have so many alternate versions for x86 architecture. In the end it's balance between one getting slightly slower and the others getting much faster. As long as the slowdown is within the error margin, we really don't worry much. Some processors clearly perform better if the code is tuned for that specific processor, processor vendor or processor family (even AMD and Intel processors have incompatible differences).

mtl1979:
Seems to fail several CI runs due to incompatible types in the fallback functions.

KungFuJesus:
> Seems to fail several CI runs due to incompatible types in the fallback functions.

That last push should fix it, I think.

KungFuJesus:
Looks like it's about to be all green.  Hah, this definitely wouldn't have been feasible without the huge number of architectural registers available, that's for sure.  Actually, I don't have access to a NEON capable aarch32 system - it'd be handy if somebody could measure if the spills from having register starvation end up being a problem here.  

I don't _think_ it would, I think GCC does a decent job about spilling to GPRs first and those moves are _relatively_ painless compared to stack, but it'd be nice to confirm that we're still at least faster than the previous implementation for aarch32.  Though, you _shouldn't_ be using an aarch32 distribution on an aarch64 system, you're basically handicapping yourself already *shakes fist at raspberry pi foundation*.  There were compelling performance reasons for them to switch to aarch64 for a long long time and they refused to do an aarch64 raspbian port until like a week ago when they made it official.

Dead2:
RPI 3, aarch64, GCC 10

## Baseline GCC10 321028cde93dac9f47b9b92ad186c4cbffb77cf2
```
 Tool: minideflate Levels: 0-9
 Runs: 32         Trim worst: 18

 0    100.015%      0.020/0.033/0.040/0.005        0.036/0.055/0.064/0.009       15,738,731
 1     54.167%      0.550/0.566/0.576/0.008        0.225/0.248/0.261/0.011        8,523,873
 2     43.877%      0.955/0.966/0.974/0.006        0.231/0.248/0.263/0.011        6,904,637
 3     42.389%      1.291/1.301/1.311/0.006        0.224/0.246/0.257/0.010        6,670,464
 4     41.647%      1.393/1.433/1.444/0.014        0.215/0.239/0.255/0.012        6,553,751
 5     41.226%      1.520/1.540/1.555/0.012        0.220/0.234/0.244/0.009        6,487,426
 6     41.044%      1.772/1.802/1.815/0.012        0.200/0.222/0.238/0.012        6,458,801
 7     40.778%      2.404/2.419/2.430/0.008        0.191/0.227/0.241/0.012        6,416,912
 8     40.703%      2.929/2.949/2.963/0.009        0.212/0.227/0.238/0.007        6,405,232
 9     40.409%      3.309/3.349/3.365/0.014        0.208/0.231/0.249/0.014        6,358,939

 avg1  48.626%                        1.636                          0.218
 avg2  54.028%                        1.817                          0.242
 tot                                228.999                         30.476       76,518,766

 116222    1664      16  117902   1cc8e libz-ng.so.2
 
adler32/c/2048                    3216 ns         3216 ns       217665
adler32/c/4096                    6404 ns         6404 ns       109217
adler32/c/32768                  48871 ns        48871 ns        14325
adler32/c/262144                392175 ns       392176 ns         1784
adler32/c/2097152              3234376 ns      3233998 ns          216
adler32/c/4194304              6494097 ns      6494141 ns          108
adler32/neon/2048                 1398 ns         1398 ns       500747
adler32/neon/4096                 2744 ns         2743 ns       255071
adler32/neon/32768               22103 ns        22101 ns        31677
adler32/neon/262144             176862 ns       176862 ns         3952
adler32/neon/2097152           1595711 ns      1595351 ns          445
adler32/neon/4194304           3204555 ns      3204200 ns          222
```

## PR #1166 GCC10 50d94bc960980b6dbf54f746dabbc19535a0ff70
```
 0    100.015%      0.016/0.028/0.032/0.005        0.035/0.049/0.060/0.008       15,738,731
 1     54.167%      0.538/0.560/0.573/0.011        0.222/0.244/0.259/0.012        8,523,873
 2     43.877%      0.949/0.969/0.980/0.009        0.212/0.243/0.253/0.010        6,904,637
 3     42.389%      1.253/1.290/1.305/0.016        0.207/0.234/0.248/0.012        6,670,464
 4     41.647%      1.401/1.427/1.443/0.012        0.193/0.228/0.244/0.015        6,553,751
 5     41.226%      1.528/1.550/1.561/0.011        0.211/0.232/0.244/0.011        6,487,426
 6     41.044%      1.781/1.795/1.808/0.009        0.208/0.226/0.236/0.009        6,458,801
 7     40.778%      2.391/2.419/2.432/0.013        0.184/0.216/0.232/0.014        6,416,912
 8     40.703%      2.939/2.960/2.983/0.015        0.198/0.222/0.231/0.010        6,405,232
 9     40.409%      3.305/3.342/3.354/0.014        0.192/0.232/0.248/0.015        6,358,939

 avg1  48.626%                        1.634                          0.213
 avg2  54.028%                        1.816                          0.236
 tot                                228.776                         29.762       76,518,766

 116734    1664      16  118414   1ce8e libz-ng.so.2

adler32/c/2048                    3216 ns         3216 ns       217723
adler32/c/4096                    6405 ns         6405 ns       109302
adler32/c/32768                  48838 ns        48836 ns        14328
adler32/c/262144                392281 ns       392232 ns         1785
adler32/c/2097152              3244762 ns      3244374 ns          215
adler32/c/4194304              6512863 ns      6512889 ns          108
adler32/neon/2048                  795 ns          795 ns       879910
adler32/neon/4096                 1493 ns         1493 ns       468779
adler32/neon/32768               11984 ns        11984 ns        58389
adler32/neon/262144              99748 ns        99736 ns         7006
adler32/neon/2097152            925089 ns       925041 ns          763
adler32/neon/4194304           1833326 ns      1833336 ns          381
```

Compiled code size increases a bit.
Compression sees no significant change.
Decompression is around 2.3% faster, so that is nice.
Google benchmarks shows nearly double the speed of adler32 neon, very nice.

Dead2:
~Not related to this PR, but incidentally I found that our arm builds have gotten quite a bit slower between 88f331e6a64653b533721a645f8253b78e7f391a and 321028cde93dac9f47b9b92ad186c4cbffb77cf2.~
~I am not sure what commit could have caused that, thought I already benchmarked all the likely suspects.~

Oops, I think I just confused myself by comparing old minigzip benchmarks to newer minideflate ones. So probably not a problem after all, but running a quick bench just in case.

Verified: Very slightly slower, but nothing worth looking into after all.

KungFuJesus:
A friend of mine saw improvements on his M1 mac that were in the same ballpark, though the version of this that did vmla in the loop body for some reason did _slightly_ better.  Deferring the vmla operation for me on the Cortex-A73.Cortex-A53 was substantially better.

PR #1169, opened 2022-02-18
Fix failing LD4 test...

mtl1979:
Cross-compilers don't support `-march=native`, so don't try to use it.

### DIFF
--- a/configure
+++ b/configure
@@ -1170,6 +1170,13 @@ EOF
 }
 
 check_neon_ld4_intrinsics() {
+    if test $buildneon -eq 1 && test $native -eq 0; then
+        if test "$CC_ARCH" = "aarch64" || test "$CC_ARCH" = "aarch64_be"; then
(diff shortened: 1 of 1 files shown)


KungFuJesus:
So long as aarch64 assumes asimd this should be fine. For aarch32 this maybe an issue. The neon flags are needed for the compilation test.

mtl1979:
@KungFuJesus `neonflag` is not available at the moment when the test is run... I might need to reorder setting `neonflag`.

I changed that the compile test uses `-mfpu=neon` if supported and NEON support is not disabled.

KungFuJesus:
Does -mfpu=neon work with aarch64? I thought they changed that flag when compiling for 64 bit.

KungFuJesus:
Oh I see, you're handling that in the patch.

KungFuJesus (approved):

nmoinvaz (approved):

Issue #1127, opened 2022-01-21
We should include more holistic benchmarks 


KungFuJesus:
Microbenchmarks are great and a good way to get a microscopic view and we should continue to add those as we see fit.  We also should add benchmarks for larger chains of operation such as inflate and deflate so that we can see macro effects of seemingly small optimizations.  This is also helpful for scenarios where microbenchmarks can't offer a clear or consistent view of performance implications.  This is especially helpful for the much less conclusive memcpy surrogate functions,  which are going be called and varying frequencies and size distributions depending of compression level and compressibility.

nmoinvaz:
Do we want to still keep this issue open?

KungFuJesus:
I guess the one thing left I might want to implement / want implemented is for someone to reproduce what minigzip and minideflate are doing as callable functions with compression levels so that they can have google benchmark harnesses.

nmoinvaz:
I've thought about doing this too for google test. I wonder how it should be done. Because minigzip, minideflate, fuzzer_minigzip all share very similar code..

nmoinvaz:
Currently CMake enumerates all files in test/data and runs a test on each file for each compression level/strategy. So I'm not sure if it makes sense to also do that for google test and we could possibly get rid of those cmake tests..

KungFuJesus:
I was thinking for benchmark we could just have a canonical path string for the benchmark data and allow the user to drop in whatever file they want to benchmark against

nmoinvaz:
That makes sense especially considering that the user will likely want to manually specify a file that can fit in memory, where as google tests doesn't necessary need to worry about speed and doesn't need to load the entire files  into memory.

nmoinvaz:
I think we have accomplished a lot of this with all the benchmarks we have now. 

> I guess the one thing left I might want to implement / want implemented is for someone to reproduce what minigzip and minideflate are doing as callable functions with compression levels so that they can have google benchmark harnesses.

`benchmark_deflate` is now separated by compression level and with crc on and off. Benchmark for `minigzip` and `minideflate` can be done with deflatebench repo.

> I was thinking for benchmark we could just have a canonical path string for the benchmark data and allow the user to drop in whatever file they want to benchmark against

This is now possible with `benchmark_corpora`.

PR #1170, opened 2022-02-21
Adding some application-specific benchmarks

KungFuJesus:
So far there's only added png encode and decode with predictably
compressible bytes. This gives us a rough idea of more holistic
impacts of performance improvements (and regressions).

An interesting thing found with this, when compared with stock zlib,
we're slower for png decoding at levels 8 & 9. When we are slower, we
are spending a fair amount of time in the chunk copy function. This
probably merits a closer look.

This code creates optionally an alternative benchmark binary that links
with an alternative static zlib implementation. This can be used to
quickly compare between different forks.


KungFuJesus:
This partially addresses https://github.com/zlib-ng/zlib-ng/issues/1127

mtl1979:
I would rather see all the files in `test/benchmarks` as it would avoid duplicating the code to fetch Google Tests repository. It would also allow making `WITH_BENCHMARK_APPS` depend on enabling of `WITH_BENCHMARKS`.

KungFuJesus:
> I would rather see all the files in `test/benchmarks` as it would avoid duplicating the code to fetch Google Tests repository. It would also allow making `WITH_BENCHMARK_APPS` depend on enabling of `WITH_BENCHMARKS`.

Resolved

KungFuJesus:
Another interesting observation:

For whatever reason deflate_fast's functable entry points to longest_match_unaligned_sse2 instead of longest_match_unaligned_avx2 like it is at encode.  Is there something that's somehow resetting this function table entry on deflate?  Interestingly forcing it on in this scenario maybe seems like it might be slower?  It's hard to say for sure exactly.

mtl1979:
@KungFuJesus If you try to read functable from different thread than was originally used, you get different output. This is because on 64-bit AMD and Intel processors, SSE2 check is bypassed.

KungFuJesus:
> @KungFuJesus If you try to read functable from different thread than was originally used, you get different output. This is because on 64-bit AMD and Intel processors, SSE2 check is bypassed.

Nothing in this code as far as I'm aware is explicitly multithreaded, but it's not impossible for google benchmark to maybe spawn up a new thread for new tests.  Still, you'd think the functable would then have thread local copies on the LWP creation of the empty entries and it'd rerun the code to detect things again.

:thinking: this is odd.  The body of main for the tests pretty much ensures that when the process first launches, the check_cpu_features() function gets called.

mtl1979:
functable should use thread-local storage, but the CPU feature check variables should be global... It's possible the compiler has stored the CPU feature variables in thread-local storage, which is incorrect...

KungFuJesus:
What's interesting here is that specifically chunkcopy_safe* seems to be where most of the time is spent.  If I sub in a version of that function that calls straight memcpy and returns out+len, we get faster than we were but still not as fast as stock zlib for png decoding at levels 8 & 9.

### Unaltered test (us)
```
2022-02-21T11:48:03-05:00
Running ./benchmark_zlib_apps
Run on (28 X 2187.41 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x14)
  L1 Instruction 32 KiB (x14)
  L2 Unified 1024 KiB (x14)
  L3 Unified 19712 KiB (x1)
Load Average: 0.20, 0.21, 0.14
------------------------------------------------------------------
Benchmark                        Time             CPU   Iterations
------------------------------------------------------------------
png_decode/png_decode/0       3930 us         3930 us          178
png_decode/png_decode/1       1473 us         1473 us          475
png_decode/png_decode/2       1468 us         1468 us          476
png_decode/png_decode/3       1610 us         1610 us          435
png_decode/png_decode/4       1442 us         1442 us          486
png_decode/png_decode/5       1776 us         1776 us          394
png_decode/png_decode/6       1754 us         1754 us          399
png_decode/png_decode/7       5145 us         5145 us          136
png_decode/png_decode/8       4959 us         4959 us          141
png_decode/png_decode/9       4959 us         4959 us          141
```

### Stock zlib test
```
2022-02-21T11:48:50-05:00
Running ./benchmark_zlib_apps_alt
Run on (28 X 3774.91 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x14)
  L1 Instruction 32 KiB (x14)
  L2 Unified 1024 KiB (x14)
  L3 Unified 19712 KiB (x1)
Load Average: 0.25, 0.23, 0.14
------------------------------------------------------------------
Benchmark                        Time             CPU   Iterations
------------------------------------------------------------------
png_decode/png_decode/0       3277 us         3277 us          213
png_decode/png_decode/1       1872 us         1872 us          374
png_decode/png_decode/2       1864 us         1864 us          376
png_decode/png_decode/3       1855 us         1855 us          378
png_decode/png_decode/4       2394 us         2394 us          293
png_decode/png_decode/5       2370 us         2370 us          295
png_decode/png_decode/6       2334 us         2334 us          300
png_decode/png_decode/7       2334 us         2334 us          300
png_decode/png_decode/8       2287 us         2287 us          306
png_decode/png_decode/9       2285 us         2285 us          306
```

### Using memcpy for "safe" routine
```
2022-02-21T11:45:25-05:00
Running ./benchmark_zlib_apps
Run on (28 X 3622.76 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x14)
  L1 Instruction 32 KiB (x14)
  L2 Unified 1024 KiB (x14)
  L3 Unified 19712 KiB (x1)
Load Average: 0.22, 0.18, 0.11
------------------------------------------------------------------
Benchmark                        Time             CPU   Iterations
------------------------------------------------------------------
png_decode/png_decode/0       3938 us         3938 us          178
png_decode/png_decode/1       1471 us         1471 us          476
png_decode/png_decode/2       1467 us         1467 us          476
png_decode/png_decode/3       1541 us         1541 us          454
png_decode/png_decode/4       1445 us         1445 us          484
png_decode/png_decode/5       1631 us         1631 us          429
png_decode/png_decode/6       1620 us         1620 us          432
png_decode/png_decode/7       3479 us         3479 us          201
png_decode/png_decode/8       3373 us         3373 us          207
png_decode/png_decode/9       3376 us         3376 us          207
```

This is interesting but I guess not totally surprising since we noticed chunkcopy_safe variants were unusually slow in a lot of circumstances.  I do wonder if the extra function call as opposed to the simple copy loop inline that's likely happening with inflate in vanilla zlib is _also_ part of the problem.

KungFuJesus:
So this is interesting.  I made things way faster in general on the decode side of things in a couple of ways.
- For whatever reason, chunkcopy_safe_* are slow.  It seems that chunkcopy_avx is fine or at least rarely called so it's not an issue.  When it is called, I couldn't add a surrogate memcpy routine in there for some reason, so it must be inducing some side effects with the tailed extraneous copying that I'm missing somehow.  Oh well, I can leave that alone.
- Next up is chunk_memset.  This is also slow.  I noticed the "dist" being copied for this is 3, something that is perhaps quite smaller than the code expects.  I was able to write a generally faster version of this using standard libc functions for these oddball cases.  The way I did it was created a stack allocated buff of size sizeof(chunk_t).  The is pretty much a guaranteed hint to the compiler that it can use the builtin_memcpy routine and just do simple vector load stores.  If dist was 1, just do simple memcpy, we probably won't beat that.  if dist was <= chunk_t size, then fill the stack buffer with with the floor division number of times the dist bytes will fit in there.  Then just increment dist * floor_div times per loop iteration.  With the remainder there's a simple loop to handle it.  Because our template functions are compiling with the best possible vectorizations per compilation unit, the auto vectorizer does the right thing.

Here's the end result of that:

```
------------------------------------------------------------------
Benchmark                        Time             CPU   Iterations
------------------------------------------------------------------
png_decode/png_decode/0       3943 us         3943 us          178
png_decode/png_decode/1       1472 us         1472 us          476
png_decode/png_decode/2       1471 us         1471 us          476
png_decode/png_decode/3       1474 us         1474 us          475
png_decode/png_decode/4       1444 us         1444 us          485
png_decode/png_decode/5       1489 us         1489 us          470
png_decode/png_decode/6       1481 us         1481 us          472
png_decode/png_decode/7       1889 us         1889 us          371
png_decode/png_decode/8       1831 us         1831 us          382
png_decode/png_decode/9       1829 us         1829 us          383
```
Level 0 is mostly slow because the CRC is slow and my PR hasn't been merged for that yet.  Zero compression just means way more bytes to checksum in the PNG file that libpng is externally calling crc32 from per IDAT entry.

test/benchmarks/CMakeLists.txt
```diff
@@ -41,3 +42,43 @@ if(ZLIB_ENABLE_TESTS)
     add_test(NAME benchmark_zlib
         COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:benchmark_zlib>)
 endif()
+
+if (WITH_BENCHMARK_APPS)
+    option(BUILD_ALT_BENCH "Link against alternative zlib implementation" OFF)
+
+    # Search for libpng package
+    find_package(PNG QUIET REQUIRED)
```
nmoinvaz: Have you thought about using `FetchContent` to download and compile `libpng`?

test/benchmarks/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ include(FetchContent)
 
 enable_language(CXX)
 
+
```
nmoinvaz: Extra line?

nmoinvaz:
I think the benchmark app should dual link zlib and zlib-ng -- instead of having one benchmark app for zlib and one benchmark app for zlib-ng. Then we could get rid of `ZLIB_DUAL_LINK` option because this would test it better than that. What do you think?

test/benchmarks/CMakeLists.txt
KungFuJesus: I guess I could in the even that libpng is not found.  I suspect it's basically what google benchmark is doing?

KungFuJesus:
> I think the benchmark app should dual link zlib and zlib-ng -- instead of having one benchmark app for zlib and one benchmark app for zlib-ng. Then we could get rid of `ZLIB_DUAL_LINK` option because this would test it better than that. What do you think?

So how exactly does that work?  Is that compatible with the zlib_compat flag that's currently required for this to work?  Would we need to modify libPNG's calls to zlib or recompile it?  I was hoping to modify as little as possible on libPNG.  

Having two separate binaries makes things a little easier for running compare.py, but if we can just export some environment variable before launching, perhaps that can accommodate things similarly?

nmoinvaz:
Actually, that is a good point: libpng only works with zlib-compat.

test/benchmarks/png_funcs.h
```diff
@@ -0,0 +1,150 @@
...
+
+    for (; i < green_stop; ++i) {
+       buf[3*i] = 0; 
+       buf[3*i+1] = 0; 
+       buf[3*i+2] = 255; 
+    }
+}
+
+static inline void encode_png(png_bytep buf, png_dat *outpng, int comp_level) {
+        png_structp png = png_create_write_struct(PNG_LIBPNG_VER_STRING, NULL, NULL, NULL);
```
nmoinvaz: Indentation here is wrong I think. Should only be 4 bytes..

test/benchmarks/benchmark_png_encode.cc
```diff
@@ -0,0 +1,55 @@
...
+public:
+    /* Let's make the vanilla version have something extremely
+     * compressible */
+    virtual void init_img(png_bytep img_bytes, size_t width, size_t height) {
+        init_compressible(img_bytes, width*height);
+    }
+
+    void SetUp(const ::benchmark::State& state) {
+        input_img_buf = (uint8_t*)malloc(IMWIDTH * IMHEIGHT * 3);
+        outpng.buf = (uint8_t*)malloc(IMWIDTH*IMHEIGHT*3);
```
nmoinvaz: Inconsistent malloc size formatting. Use spaces between operators on second line like on the first. Also on the `outpng.buf_rem` assignment.

test/benchmarks/benchmark_png_encode.cc
```diff
@@ -0,0 +1,55 @@
+#include <stdio.h>
+#include <benchmark/benchmark.h>
+#include "png_funcs.h"
+#include <assert.h>
```
nmoinvaz, line 2: Consider moving <assert.h> above <benchmark/benchmark.h>

test/benchmarks/png_funcs.h
```diff
@@ -0,0 +1,150 @@
...
+
+        dat->buf_rem += 2*len;
+    }
+
+    memcpy(dat->buf + dat->len, data, len);
+    dat->len = curSize;
+    dat->buf_rem -= len;
+}
+
+#if 0
```
nmoinvaz: What is this code, is it needed?

test/benchmarks/png_funcs.h
```diff
@@ -0,0 +1,150 @@
...
+    if (!info) abort();
+
+    png_set_read_fn(png, dat, read_from_pngdat);
+    png_read_info(png, info);
+
+    png_uint_32 width = 0;
+    png_uint_32 height = 0;
+    int bit_depth = 0;
+    int color_type = -1;
+    png_get_IHDR(png, info,
```
nmoinvaz: This can probably be condensed to one line.

test/benchmarks/png_funcs.h
```diff
@@ -0,0 +1,150 @@
...
+static void read_from_pngdat(png_structp png, png_bytep out, png_size_t bytes_to_read)
+{
+    png_parse_dat *io = (png_parse_dat*)png_get_io_ptr(png);
+    memcpy(out, io->cur_pos, bytes_to_read);
+    io->cur_pos += bytes_to_read;
+}
+
+static inline void decode_png(png_parse_dat *dat, png_bytep out_bytes) {
+    png_bytep out_rows[IMHEIGHT];
+    for (size_t i = 0; i < IMHEIGHT; ++i) out_rows[i] = &out_bytes[IMWIDTH*i];
```
nmoinvaz: Break this `for` line into two lines. It is kind of harder to read it being on one line.

nmoinvaz (changes requested):
Is there anywhere where you are checking that the encoded/decoded bytes are as expected? Or is it assumed and you are just checking for speed only.

test/benchmarks/CMakeLists.txt
```diff
@@ -41,3 +42,43 @@ if(ZLIB_ENABLE_TESTS)
...
+        target_link_libraries(benchmark_zlib_apps_alt libpng.a ${ZLIB_ALT_LIB} benchmark::benchmark)
+        target_compile_definitions(benchmark_zlib_apps_alt PRIVATE BUILD_ALT=1)
+        target_include_directories(benchmark_zlib_apps_alt PRIVATE
+            ${CMAKE_SOURCE_DIR}
+            ${CMAKE_BINARY_DIR}
+            ${PNG_INCLUDE_DIR}
+            ${benchmark_SOURCE_DIR}/benchmark/include)
+    endif()
+
+
```
nmoinvaz: Extra line

test/benchmarks/CMakeLists.txt
```diff
@@ -41,3 +42,43 @@ if(ZLIB_ENABLE_TESTS)
...
+        ${CMAKE_SOURCE_DIR}
+        ${CMAKE_BINARY_DIR}
+        ${PNG_INCLUDE_DIR}
+        ${benchmark_SOURCE_DIR}/benchmark/include)
+
+    # We need the static png library if we're statically linking to zlib,
+    # otherwise it will resolve these things in the system provided dynamic
+    # libraries (likely linked to stock zlib)
+    target_link_libraries(benchmark_zlib_apps libpng.a zlibstatic benchmark::benchmark)
+
```
nmoinvaz: Extra line

test/benchmarks/CMakeLists.txt
nmoinvaz: Yeah. And at least then it can be compiled on Windows.

test/benchmarks/CMakeLists.txt
```diff
@@ -41,3 +42,43 @@ if(ZLIB_ENABLE_TESTS)
     add_test(NAME benchmark_zlib
         COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:benchmark_zlib>)
 endif()
+
+if (WITH_BENCHMARK_APPS)
```
nmoinvaz: No space after `if` in cmake files.

test/benchmarks/CMakeLists.txt
```diff
@@ -41,3 +42,43 @@ if(ZLIB_ENABLE_TESTS)
...
+
+    set(BENCH_APP_SRCS 
+        benchmark_png_encode.cc
+        benchmark_png_decode.cc
+        benchmark_main.cc
+    )
+
+    add_executable(benchmark_zlib_apps ${BENCH_APP_SRCS})
+
+    if (DEFINED BUILD_ALT_BENCH)
```
nmoinvaz: No space after `if` in cmake files.

test/benchmarks/CMakeLists.txt
```diff
@@ -41,3 +42,43 @@ if(ZLIB_ENABLE_TESTS)
...
+        set(ZLIB_ALT_LIB "libz.a" CACHE FILEPATH "Optional alternative zlib implementation (defaults to stock zlib)")
```
nmoinvaz, line 71: I almost wonder if we should use `FetchContent` for madler/zlib too. This would mean that it would not build the alternative app if `ZLIB_COMPAT=ON`, but that would be totally fine imo. If zlib-compat is OFF, then it would only build the one app...

nmoinvaz: But I wonder if that would work or not. Because the project names are so similar.

KungFuJesus:
> Is there anywhere where you are checking that the encoded/decoded bytes are as expected? Or is it assumed and you are just checking for speed only.

Speed only - though it's caught some issues due to the CRC's being checked in the code.  If something is amiss the png decode process seems likely to notice.

test/benchmarks/png_funcs.h
KungFuJesus: I had an alternative version meant to compress poorly but it seemed to have really weird variance run to run, even when fixing the random seed.  It might be worth re-enabling in the future but as you can see, it's not a complicated function.

KungFuJesus:
@nmoinvaz are you now able to build and test this on windows?

test/benchmarks/CMakeLists.txt
```diff
@@ -41,3 +41,52 @@ if(ZLIB_ENABLE_TESTS)
     add_test(NAME benchmark_zlib
         COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:benchmark_zlib>)
 endif()
+
+if(WITH_BENCHMARK_APPS)
+    option(BUILD_ALT_BENCH "Link against alternative zlib implementation" OFF)
+
+    # Search for libpng package
+    find_package(PNG QUIET REQUIRED)
```
nmoinvaz: This shouldn't be `REQUIRED`.

nmoinvaz:
No I get this error:
```
CMake Error at C:/Program Files/CMake/share/cmake-3.22/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find PNG (missing: PNG_LIBRARY PNG_PNG_INCLUDE_DIR)
```

test/benchmarks/CMakeLists.txt
KungFuJesus: Ah, I see.  Fixing now.

KungFuJesus:
With all of my recent optimizations merged in one branch, it looks like we're handily beating stock zlib but not beating chromium's changes with png encode (we definitely still are with decode).

A perf profile reveals that their "inflate fast" seems to be be doing way less copying - to the point that most of the time is actually occupied by their adler checksum.  I'll have to compare these a little bit more closely.  I know the chunkmemset is ~7% of the time and there's an obvious optimization there with building up a permutation table.  Nearly half of the time is measured in memcpy, though (and no, forcing back on our copies doesn't appreciably move the needle, as I suspected it wouldn't).  

It's quite possible that their "encode" is writing an easier to inflate stream, I suppose, but I'll have to verify if that's the case.

So I do see at least one optimization that they are employing that we aren't.  It looks as though they are identifying the overlapping chunkcopy's figuring out the overlap, and then repeating the elements over and over for a specific period (basically exploiting the fact that the self modifying copy can repeat itself.

Still, at least in some of the initial testing I've done these overlaps aren't _super_ common place.  Oh and given until recently we weren't handling that behavior correctly in all cases leads me to believe the test imagery I'm generating in that buffer doesn't reproduce that overlapping copy, anyway.

nmoinvaz:
That is great that you've tested against Chromium. I haven't done any comparisons against that fork myself. I know when Chromium developers worked on their fork it was specifically for png optimization.

I wouldn't spend too much time on a permutation table if the table is going to be large because then it will bloat the code size.

KungFuJesus:
I don't expect it'd be all that large in the grand scheme of things. No worse than the shuffle table that exists for the crc folding, anyway.

I suspect inffast.c is where the larger difference is hiding though. I'll have to thoroughly compare them even though they are extremely similar implementations.

test/benchmarks/CMakeLists.txt
KungFuJesus: Marking as resolved for now.  The suggestion about a parallel fetch for zlib is likely to cause problems given that they are exported "zlib" as the project name in both places.  We only actually need the library itself and I generally expect the user to specify a path to this if not the distro installed zlib on a Linux distribution.

Dead2 (approved):
LGTM, but pending re-review by @nmoinvaz since he was listed as "requested changes".

test/benchmarks/README.md
```diff
@@ -17,3 +16,32 @@ To filter out which benchmarks are performed use:
...
+
+### Benchmark benchmark_zlib_apps
+These benchmarks measure applications of zlib as a whole.  Currently the only examples
+are PNG encoding and decoding. The PNG encode and decode tests leveraging procedurally
+generated and highly compressible image data.
+
+Additionally, a test called `png_decode_realistic` that will decode any RGB 8 BPP encoded
+set of PNGs in the working directory under a directory named "test_pngs" with files named
+{0..1}.png. If these images do not exist, they will error out and the benchmark will move
+on to the next set of benchmarks.
```
nmoinvaz, line 42: If these images do not exist, shouldn't it skip the tests rather than error out?

test/benchmarks/benchmark_main.cc
```diff
@@ -7,13 +7,17 @@
 
 #include <benchmark/benchmark.h>
 
+#ifndef BUILD_ALT
```
nmoinvaz, line 10: Instead of `BUILD_ALT` couldn't we use `ZLIB_COMPAT` instead?

test/benchmarks/png_funcs.h
```diff
@@ -0,0 +1,146 @@
+#pragma once
```
nmoinvaz: Maybe we can rename this file to `benchmark_png_shared.h`? Thoughts?

test/benchmarks/png_funcs.h
```diff
@@ -0,0 +1,146 @@
...
+
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+
+#define IMWIDTH 1024
+#define IMHEIGHT 1024
+
+extern "C" {
+    #include <png.h>
```
nmoinvaz: It should be `#  include <png.h>` like in the other files with two spaces between # and include and no tab at the beginning?

test/benchmarks/png_funcs.h
```diff
@@ -0,0 +1,146 @@
...
+    if (dat->buf_rem < len) {
+        dat->buf = (uint8_t*)realloc(dat->buf, dat->len + dat->buf_rem + 2*len);
+
+        if (!dat->buf) {
+            /* Pretty unlikely but we'll put it here just in case */
+            fprintf(stderr, "realloc failed, exiting\n");
+            exit(1);
+        }
+
+        dat->buf_rem += 2*len;
```
nmoinvaz: Space between 2 * len.

test/benchmarks/png_funcs.h
```diff
@@ -0,0 +1,146 @@
...
+
+/* Write a customized write callback so that we write back to an in-memory buffer.
+ * This allows the testing to not involve disk IO */
+static void png_write_cb(png_structp pngp, png_bytep data, png_size_t len) {
+    png_dat *dat = (png_dat*)png_get_io_ptr(pngp);
+    size_t curSize = dat->len + len;
+
+    /* realloc double the requested buffer size to prevent excessive reallocs */
+    if (dat->buf_rem < len) {
+        dat->buf = (uint8_t*)realloc(dat->buf, dat->len + dat->buf_rem + 2*len);
```
nmoinvaz: Space between 2 * len consider parens too..

test/benchmarks/png_funcs.h
```diff
@@ -0,0 +1,146 @@
...
+    info = png_create_info_struct(png);
+    if (!info) abort();
+
+    png_set_read_fn(png, dat, read_from_pngdat);
+    png_read_info(png, info);
+
+    int bit_depth = 0, color_type = -1;
+    png_get_IHDR(png, info, &width, &height, &bit_depth, &color_type, NULL, NULL, NULL);
+
+    size_t im_size = width * height * bit_depth/8;
```
nmoinvaz: Space between / operator.

test/benchmarks/png_funcs.h
```diff
@@ -0,0 +1,146 @@
...
+    /* It doesn't actually matter what we make this, but for
+     * the sake of a reasonable test image, let's make this
+     * be a stripe of R, G, & B, with no alpha channel */
+    int i = 0;
+    int red_stop = num_pix/3;
+    int blue_stop = 2*num_pix/3;
+    int green_stop = num_pix;
+
+    for (; i < red_stop; ++i) {
+       buf[3*i] = 255;
```
nmoinvaz: It might be possible to use `int32_t x = 3 * i;` and use x in the array index..

test/benchmarks/png_funcs.h
```diff
@@ -0,0 +1,146 @@
...
+    dat->buf_rem -= len;
+}
+
+static void init_compressible(png_bytep buf, size_t num_pix) {
+    /* It doesn't actually matter what we make this, but for
+     * the sake of a reasonable test image, let's make this
+     * be a stripe of R, G, & B, with no alpha channel */
+    int i = 0;
+    int red_stop = num_pix/3;
+    int blue_stop = 2*num_pix/3;
```
nmoinvaz: Space between operators.

test/benchmarks/benchmark_png_encode.cc
```diff
@@ -0,0 +1,55 @@
...
+class png_encode: public benchmark::Fixture {
+private:
+    png_dat outpng;
+
+    /* Backing this on the heap is a more realistic benchmark */
+    uint8_t *input_img_buf = NULL;
+
+public:
+    /* Let's make the vanilla version have something extremely
+     * compressible */
```
nmoinvaz: This can be one line.

test/benchmarks/benchmark_png_encode.cc
```diff
@@ -0,0 +1,55 @@
...
+    virtual void init_img(png_bytep img_bytes, size_t width, size_t height) {
+        init_compressible(img_bytes, width*height);
```
nmoinvaz: Space between operator.

test/benchmarks/README.md
KungFuJesus: Probably. Does benchmark have a skip without error? If so I can use that.

test/benchmarks/benchmark_png_decode.cc
```diff
@@ -0,0 +1,123 @@
...
+        /* Let's take all the images at different compression levels and jam their bytes into buffers */
+        char test_fname[25];
+        FILE *files[10];
+
+        /* Set all to NULL */
+        memset(files, 0, sizeof(FILE*));
+        for (size_t i = 0; i < 10; ++i)
+            inpng[i] = { NULL, 0, 0 };
+
+        for (size_t i = 0; i < 10; ++i) {
```
nmoinvaz, line 64: This for loop and the one where you are reading the bytes in the file, I think they can just be merged into one for loop.  

Actually, all these for loops can be merged into one. I think it would be cleaner.

test/benchmarks/benchmark_png_decode.cc
```diff
@@ -0,0 +1,123 @@
...
+class png_decode: public benchmark::Fixture {
+protected:
+    png_dat inpng[10];
+
+    /* Backing this on the heap is a more realistic benchmark */
+    uint8_t *output_img_buf = NULL;
+
+public:
+    /* Let's make the vanilla version have something extremely
+     * compressible */
```
nmoinvaz: Can be a single line.

nmoinvaz (commented):
I have made another review as suggested by @Dead2. Some formatting changes, but also some questions and possible improvements..

test/benchmarks/README.md
nmoinvaz: There is a `GTEST_SKIP()`.. but that is for google test..

test/benchmarks/benchmark_main.cc
KungFuJesus: Could we?  I think I'm letting cmake use our headers in both circumstances, I'm only modifying what's being linked to.  I think the headers should be API compatible so that shouldn't be an issue.

test/benchmarks/png_funcs.h
KungFuJesus: Sure

test/benchmarks/png_funcs.h
KungFuJesus: This is honestly one of the more oddball style choices I've seen but I can make it consistent.

test/benchmarks/benchmark_png_decode.cc
KungFuJesus: The reason I did that is because the destructor of the class frees everything.  Because the error handling can happen mid loop, everything needs to be initialized as NULL or a wild pointer will be freed.

test/benchmarks/README.md
KungFuJesus: Looks as though there's only a "SkipWithError":
https://github.com/google/benchmark/blob/d08e7b60564549aeaaa23d7bb075b6b9abf01d49/src/benchmark.cc#L212

The error doesn't halt the rest of the benchmarks so I think it's probably the cleanest thing that we can do, since the bootstrapping of these tests is basically done at compile time.

test/benchmarks/benchmark_main.cc
nmoinvaz: Yes it should work so long as you always ensure `ZLIB_COMPAT` is set..

test/benchmarks/CMakeLists.txt
```diff
@@ -41,3 +41,52 @@ if(ZLIB_ENABLE_TESTS)
...
+        benchmark_main.cc
+    )
+
+    add_executable(benchmark_zlib_apps ${BENCH_APP_SRCS})
+
+    if(DEFINED BUILD_ALT_BENCH)
+        set(ZLIB_ALT_LIB "libz.a" CACHE FILEPATH "Optional alternative zlib implementation (defaults to stock zlib)")
+        add_executable(benchmark_zlib_apps_alt ${BENCH_APP_SRCS})
+        target_link_libraries(benchmark_zlib_apps_alt libpng.a ${ZLIB_ALT_LIB} benchmark::benchmark)
+        target_compile_definitions(benchmark_zlib_apps_alt PRIVATE BUILD_ALT=1)
```
nmoinvaz, line 74: This here should be `ZLIB_COMPAT` instead of `BUILD_ALT=1`. That way the alt benchmark app tool is always built against zlib.h..

test/benchmarks/benchmark_png_shared.h
```diff
@@ -0,0 +1,146 @@
...
+    /* It doesn't actually matter what we make this, but for
+     * the sake of a reasonable test image, let's make this
+     * be a stripe of R, G, & B, with no alpha channel */
+    int i = 0;
+    int red_stop = num_pix / 3;
+    int blue_stop = 2 * num_pix / 3;
+    int green_stop = num_pix;
+
+    for (; i < red_stop; ++i) {
+       buf[3 * i] = 255;
```
nmoinvaz: Actually this can be improved using:
```
for (int32_t x = 0; i < red_stop; ++i, x += 3) {
    buf[x] = 255;
    buf[x + 1] = 0;
    buf[x + 2] = 0;
}
```
Same for the loops below it.

test/benchmarks/benchmark_png_shared.h
```diff
@@ -0,0 +1,146 @@
...
+    memcpy(dat->buf + dat->len, data, len);
+    dat->len = curSize;
+    dat->buf_rem -= len;
+}
+
+static void init_compressible(png_bytep buf, size_t num_pix) {
+    /* It doesn't actually matter what we make this, but for
+     * the sake of a reasonable test image, let's make this
+     * be a stripe of R, G, & B, with no alpha channel */
+    int i = 0;
```
nmoinvaz: Let's use fixed width integer types in these lines.

test/benchmarks/benchmark_png_shared.h
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    for (; i < green_stop; ++i) {
+       buf[3 * i] = 0; 
+       buf[3 * i + 1] = 0; 
+       buf[3 * i + 2] = 255; 
+    }
+}
+
+static inline void encode_png(png_bytep buf, png_dat *outpng, int comp_level, int width, int height) {
```
nmoinvaz: Fixed width integer types.

test/benchmarks/benchmark_main.cc
KungFuJesus: That ifdef is made to prevent a call to a cpu_feature_check() function that doesn't exist in vanilla zlib or other various forks.  If we're building against the same headers at compile time and only switching out what's being linked, then we end up with an undefined reference when linking to the alternative static library.

test/benchmarks/benchmark_png_decode.cc
nmoinvaz: Can the two for loops below be merged then? So that `fopen` and `fread` occur in the same loop? What do you think?

nmoinvaz:
This is getting closer. I just had some final things.

test/benchmarks/benchmark_png_decode.cc
KungFuJesus: It can happen but then the cleanup becomes an extra for loop for setting the not yet traversed indices to NULL so that TearDown() doesn't crash

KungFuJesus: Added what was necessary to do this, have a look.

KungFuJesus: Oh for the fopen and fread - I'd rather there be all or nothing on that so that we don't have to test for the existence of a valid buffer to decode a PNG per test.  There's seem little point to parse the first 5 files only to find out the level 6 file doesn't exist, then end up freeing all of the buffers.

nmoinvaz (approved):

PR #1172, opened 2022-02-21
Call feature detection at the beginning of every stub function

KungFuJesus:
This was being done for a few but not all.  This covers ours bases just
in case the cpu features aren't detected at initialization.  This
inconsistency was found all looking for another issue but is unknown if
it was actually a problem (it certainly wasn't the culprit).

### DIFF
--- a/functable.c
+++ b/functable.c
@@ -8,6 +8,7 @@
 #include "crc32_p.h"
 #include "deflate.h"
 #include "deflate_p.h"
+#include <stdio.h>
 
 #include "functable.h"
 
@@ -97,7 +98,7 @@ Z_INTERNAL void slide_hash_stub(deflate_state *s) {
 }
 
 Z_INTERNAL uint32_t longest_match_stub(deflate_state *const s, Pos cur_match) {
-
+    cpu_check_features();
 #ifdef UNALIGNED_OK
 #  if defined(UNALIGNED64_OK) && defined(HAVE_BUILTIN_CTZLL)
     functable.longest_match = &longest_match_unaligned_64;
@@ -122,6 +123,7 @@ Z_INTERNAL uint32_t longest_match_stub(deflate_state *const s, Pos cur_match) {
 }
 
 Z_INTERNAL uint32_t longest_match_slow_stub(deflate_state *const s, Pos cur_match) {
+    cpu_check_features();
 
 #ifdef UNALIGNED_OK
 #  if defined(UNALIGNED64_OK) && defined(HAVE_BUILTIN_CTZLL)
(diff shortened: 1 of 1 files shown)


functable.c
```diff
@@ -284,8 +288,9 @@ Z_INTERNAL uint8_t* chunkcopy_safe_stub(uint8_t *out, uint8_t const *from, unsig
         functable.chunkcopy_safe = &chunkcopy_safe_sse2;
 #endif
 #ifdef X86_AVX_CHUNKSET
-    if (x86_cpu_has_avx2)
+    if (x86_cpu_has_avx2) {
```
nmoinvaz: Don't need extra braces here.

nmoinvaz:
I'm not sure if it is needed but I don't mind either.

KungFuJesus:
> I'm not sure if it is needed but I don't mind either.

Yeah the overhead should be fairly minimal since it's a fixed one time cost, anyway.  I would much rather check a boolean a couple more times than have the stub assign a random function because it's uninitialized memory for the cpu feature flag.

mtl1979:
I recall trying to make this kind of pull request a while ago and it was deemed an overkill back then... But with all the benchmarks and stuff, it's not so much an overkill anymore... Nobody can guarantee anymore that the stubs are going to be called in predetermined order, especially after adding support for thread-local variables.

functable.c
```diff
@@ -97,7 +98,7 @@ Z_INTERNAL void slide_hash_stub(deflate_state *s) {
 }
 
 Z_INTERNAL uint32_t longest_match_stub(deflate_state *const s, Pos cur_match) {
-
+    cpu_check_features();
```
mtl1979, line 101: One empty line after this, to be consistent.

functable.c
```diff
@@ -286,6 +290,7 @@ Z_INTERNAL uint8_t* chunkcopy_safe_stub(uint8_t *out, uint8_t const *from, unsig
 #ifdef X86_AVX_CHUNKSET
     if (x86_cpu_has_avx2)
         functable.chunkcopy_safe = &chunkcopy_safe_avx;
+
```
mtl1979, line 293: Extra empty line.

functable.c
```diff
@@ -250,6 +252,7 @@ Z_INTERNAL uint32_t chunksize_stub(void) {
 Z_INTERNAL uint8_t* chunkcopy_stub(uint8_t *out, uint8_t const *from, unsigned len) {
     // Initialize default
     functable.chunkcopy = &chunkcopy_c;
+    cpu_check_features();
```
mtl1979, line 255: Until we have benchmark for `chunkcopy`, this might still be an overkill.

functable.c
```diff
@@ -302,6 +307,7 @@ Z_INTERNAL uint8_t* chunkcopy_safe_stub(uint8_t *out, uint8_t const *from, unsig
 Z_INTERNAL uint8_t* chunkunroll_stub(uint8_t *out, unsigned *dist, unsigned *len) {
     // Initialize default
     functable.chunkunroll = &chunkunroll_c;
+    cpu_check_features();
```
mtl1979, line 310: `chunkunroll` should be never called alone, so this is most likely an overkill.

functable.c
KungFuJesus: Possibly.  I was failing tests when I implemented one from memcpy so I think there's some strange side effects I don't seem to be reproducing that it relies on.

mtl1979: Codecov is usually pretty accurate at finding dead code... Sometimes, it's a configuration issue; sometimes, it's just that the compiler knows it has no effect because the code block was already executed from another function or code path and it has no side-effects that change between runs...

PR #1175, opened 2022-02-23
Minor fixes and improvements minideflate

nmoinvaz:


Dead2:
Still not quite right for some reason.
Now minideflate hangs forever during inflate.
Probably stuck in the loop instead of erroring out like before.

nmoinvaz:
I wonder if I have to do `while (d_stream.avail_in > 0);`

KungFuJesus:
Heh, good thing CI doesn't invoke that tool.

Dead2:
Uhm, it seems this is wrong actually.

According to documentation, windowbits should be:
-1 to -15 is raw deflate without header or trailer
0 to 15 is deflate wrapper (adler32)
16 to 31 is gzip wrapper (crc32c)
32 to 47 is wrapper auto-detect (but not raw deflate)

So, doesn't that mean that we were already doing deflate with adler32?
Also, perhaps minideflate should do decompress with 47 by default? (would need to be tested first of course)

nmoinvaz:
@Dead2 it looks like you are correct about the `window_bits`. I've updated the help usage to reflect it. And `minideflate` does use `adler32` by default all your benchmarks before were valid. And for `inflateInit` by default it will now auto-detect the wrapper. That may have been the issue with the `Z_BUF_ERROR` that you were experiencing. 

If you still get the `Z_BUF_ERROR` can you tell me which file you are testing against? I tried against silesia-small.tar and didn't get that error.

Dead2:
Ok, I still get the error, and I guess I should have looked a bit closer at what is happening, because it is obvious it will error out when told to decompress an empty file.. (Maybe we should handle that somehow?)
The problem was that I had `-c` after the filename, so it silently ignored it and compressed into `testfil.gz` instead of my redirect into `testfil.z`, and then that file ended up empty. So this was a clear PEBKAC, sorry about that.

However, this also made me discover a little problem with file names.
When compressing, minideflate will create a file that ends with `.gz`. However, this is not always the correct filetype, especially not if you use the defaults.
So, I suggest setting the file extension to `.z` for deflate streams, and `.zraw` for the raw deflate stream.

Another problem is that it also adds the extension when decompressing. So `testfil.gz` gets decompressed to `testfil.gz.gz`, and that easily becomes a mess 😄 

An alternative is to also require an output filename as the last parameter, then the user (or our tests) will have to select the name instead. This would probably save a lot of logic to handle the output filename, but probably still smart to just check that input and output are not the same file.

I would like to see a `-k` keep parameter btw, so it does not delete the source file, as that would save us having to use `-c` stdout redirection to file in the first place 😅

mtl1979:
I see no reason why we shouldn't have `-k` parameter...

nmoinvaz:
Yes I noticed there was no `-k` argument also, but that is because it doesn't automatically delete the input file.

Dead2:
I guess that should perhaps be made clear somehow then, because that differs from how minigzip and many other compression utils works :)

Dead2:
@nmoinvaz Did you do testing of the new features? If so, then LGTM.

nmoinvaz:
I did test it while developing. I also tested that each extension was appended properly during compression and removed during decompression. And also tested the auto-wrapper stuff for inflate.

Dead2 (approved):
LGTM

PR #1173, opened 2022-02-21
Speed up chunkcopy and memset

KungFuJesus:
This was found to have a significant impact on a highly compressible PNG
for both the encode and decode.  Some deltas show performance improving
as much as 60%+.

For the scenarios where the "dist" is not an even modulus of our chunk
size, we simply repeat the bytes as many times as possible into our
vector registers.  We then copy the entire vector and then advance the
quotient of our chunksize divided by our dist value.

If dist happens to be 1, there's no reason to not just call memset from
libc (this is likely to be just as fast if not faster).

PR #1177, opened 2022-02-28
Use state->pending and state->pending_buf pointers to pending output.

nmoinvaz:
Code simplification.

PR #1178, opened 2022-03-07
Revert inflate window reorganization

nmoinvaz:
This PR removes the changes in PR #1033. Most of the performance benefits of those changes were due to the use of pclmulqdq when calculating the crc32. Now that we have PR #1161 that incorporates pclmulqdq into the `functable.crc32` function, there is no need for reorganizing the inflate window. I think the inflate window was originally reorganized simply to handle the extra memory copy that the pclmulqdq version of `crc32` required. According to the discussion on #1173, reverting these changes should also benefit png decoding performance.

## DEVELOP https://github.com/zlib-ng/zlib-ng/commit/43dbfd6709fb3a8028430ea30f3da88fbeb3ced9
```
OS: Windows 10 10.0.22000 AMD64
CPU: Intel64 Family 6 Model 158 Stepping 10, GenuineIntel
Tool: ../zlib-ng/build-develop/Release/minigzip.exe
Levels: 0-9

 0    100.008%      0.018/0.019/0.020/0.000        0.018/0.019/0.020/0.001       15,737,543
 1     54.166%      0.094/0.095/0.096/0.001        0.046/0.047/0.049/0.001        8,523,732
 2     43.871%      0.153/0.155/0.156/0.001        0.047/0.049/0.051/0.001        6,903,609
 3     42.387%      0.210/0.212/0.214/0.001        0.047/0.048/0.050/0.001        6,670,099
 4     41.647%      0.232/0.234/0.236/0.001        0.044/0.047/0.048/0.001        6,553,723
 5     41.216%      0.243/0.246/0.247/0.001        0.045/0.046/0.047/0.001        6,485,938
 6     41.037%      0.277/0.281/0.283/0.002        0.044/0.046/0.047/0.001        6,457,776
 7     40.778%      0.345/0.350/0.353/0.002        0.045/0.046/0.047/0.001        6,416,919
 8     40.704%      0.433/0.438/0.440/0.002        0.044/0.046/0.047/0.001        6,405,244
 9     40.409%      0.514/0.518/0.521/0.002        0.045/0.046/0.048/0.001        6,358,951

 avg1  48.622%                        0.255                          0.044
 avg2  54.025%                        0.283                          0.049
 tot                                 76.445                         13.234       76,513,534

OS: Windows 10 10.0.22000 AMD64
CPU: Intel64 Family 6 Model 158 Stepping 10, GenuineIntel
Tool: ../zlib-ng/build-develop/Release/minigzip.exe
Levels: 0-9

 0    100.008%      0.152/0.161/0.171/0.006        0.154/0.164/0.180/0.008      211,973,953
 1     44.409%      0.981/1.027/1.092/0.036        0.486/0.510/0.541/0.015       94,127,290
 2     35.518%      1.600/1.679/1.794/0.060        0.506/0.533/0.558/0.015       75,282,961
 3     33.882%      2.274/2.385/2.525/0.062        0.488/0.517/0.544/0.014       71,816,478
 4     33.174%      2.576/2.665/2.787/0.071        0.481/0.509/0.532/0.017       70,315,668
 5     32.660%      2.692/2.776/2.924/0.072        0.478/0.501/0.526/0.014       69,225,542
 6     32.508%      3.225/3.335/3.569/0.099        0.469/0.493/0.521/0.013       68,902,222
 7     32.255%      4.390/4.511/4.754/0.113        0.477/0.495/0.519/0.013       68,366,800
 8     32.167%      7.122/7.269/7.570/0.139        0.473/0.497/0.523/0.015       68,180,776
 9     31.887%      7.307/7.448/7.896/0.177        0.473/0.494/0.521/0.015       67,586,442

 avg1  40.847%                        3.326                          0.471
 avg2  45.385%                        3.695                          0.524
 tot                                997.655                        141.397      865,778,132
```
## PR https://github.com/zlib-ng/zlib-ng/pull/1178/commits/fa118e26e83ea62227b80f25248313a1cf42dcfe
```
OS: Windows 10 10.0.22000 AMD64
CPU: Intel64 Family 6 Model 158 Stepping 10, GenuineIntel
Tool: ../zlib-ng/build-revert/Release/minigzip.exe
Levels: 0-9

 0    100.008%      0.018/0.020/0.021/0.001        0.017/0.018/0.019/0.001       15,737,543
 1     54.166%      0.092/0.094/0.096/0.001        0.047/0.048/0.049/0.000        8,523,732
 2     43.871%      0.154/0.156/0.157/0.001        0.049/0.050/0.051/0.001        6,903,609
 3     42.387%      0.208/0.212/0.215/0.001        0.047/0.048/0.049/0.001        6,670,099
 4     41.647%      0.231/0.233/0.235/0.001        0.046/0.047/0.048/0.001        6,553,723
 5     41.216%      0.235/0.239/0.240/0.001        0.046/0.047/0.049/0.001        6,485,938
 6     41.037%      0.271/0.276/0.278/0.001        0.045/0.047/0.048/0.001        6,457,776
 7     40.778%      0.344/0.349/0.353/0.002        0.046/0.047/0.048/0.001        6,416,919
 8     40.704%      0.434/0.438/0.441/0.002        0.045/0.047/0.047/0.001        6,405,244
 9     40.409%      0.511/0.516/0.519/0.002        0.045/0.047/0.048/0.001        6,358,951

 avg1  48.622%                        0.253                          0.045
 avg2  54.025%                        0.281                          0.050
 tot                                 75.976                         13.391       76,513,534

OS: Windows 10 10.0.22000 AMD64
CPU: Intel64 Family 6 Model 158 Stepping 10, GenuineIntel
Tool: ../zlib-ng/build-revert/Release/minigzip.exe
Levels: 0-9

 0    100.008%      0.148/0.154/0.159/0.003        0.136/0.145/0.150/0.004      211,973,953
 1     44.409%      0.970/0.989/1.002/0.008        0.485/0.499/0.507/0.006       94,127,290
 2     35.518%      1.607/1.631/1.645/0.010        0.506/0.517/0.522/0.005       75,282,961
 3     33.882%      2.277/2.311/2.340/0.015        0.485/0.497/0.504/0.006       71,816,478
 4     33.174%      2.553/2.589/2.606/0.013        0.477/0.487/0.495/0.005       70,315,668
 5     32.660%      2.605/2.642/2.659/0.014        0.477/0.489/0.498/0.006       69,225,542
 6     32.508%      3.154/3.183/3.205/0.017        0.472/0.481/0.489/0.005       68,902,222
 7     32.255%      4.379/4.415/4.430/0.012        0.470/0.482/0.490/0.005       68,366,800
 8     32.167%      7.048/7.143/7.194/0.040        0.472/0.480/0.488/0.005       68,180,776
 9     31.887%      7.240/7.292/7.322/0.023        0.468/0.481/0.487/0.004       67,586,442

 avg1  40.847%                        3.235                          0.456
 avg2  45.385%                        3.594                          0.507
 tot                                970.505                        136.757      865,778,132
```


Dead2:
You would want this to go in before #1177, right?

KungFuJesus:
Hmm, before pulling the trigger on this, there is a potential experiment I think we could do that can still reap the benefits of the reorganized inflate window, but I need to understand more exactly how it was reorganized.  If there's any benefit from keeping things sequential, we can still do so without sliding the bytes around.  I alluded to earlier the potential for instead using a circular buffer for the window, instead.  We can easily slide a pointer forward rather than sliding the bytes back.

The more expensive bit of all of this was the memmove whenever the window was sliding out.  You can still enjoy the same benefit, with the copy + checksum being in the same sequence, by simply not moving bytes, just pointers.  The copy comes for free with the checksum, and whenever the window wraps around, you handle it by either combining both halves to a temporary buffer, or, in the case that you're copying the inflate window out, you span the checksum + copy over both halves, then re-advancing the pointer.

KungFuJesus:
At the very least we should have a parallel branch + PR implemented so that we can test the performance of both :).  Combining the checksum and copy is in general a good idea, It'd be good to not lose that.

mtl1979:
I think it's possible to slide the pointer once before it is required to move remaining data to start of allocation. To slide more than once, it is required to increase the size of the allocation, or otherwise we can't look up distant matches...

KungFuJesus:
> I think it's possible to slide the pointer once before it is required to move remaining data to start of allocation. To slide more than once, it is required to increase the size of the allocation, or otherwise we can't look up distant matches...

Right, you just wrap it around to the beginning of the buffer.  There will be room, we're just moving a write cursor and the write cursor will always point to the position of the start of the expired bytes we slide out.  For something that does wrap around that requires a contiguous view, we simply present the inflate code with a temporary buffer that has both halves combined.  The wrap around I'm expecting to be relatively rare so we won't be spending a ton of time going in and out of this temporary buffer.

mtl1979:
@KungFuJesus We just need to track when the whole hash table needs to be recalculated to make sure we don't have stale data. In the old code we always had stale data after the sliding of the window.

KungFuJesus:
> @KungFuJesus We just need to track when the whole hash table needs to be recalculated to make sure we don't have stale data. In the old code we always had stale data after the sliding of the window.

I'm not even sure anything actually needs to change on that front.  You can treat the positions the same, you're just sliding around the start position in the window.  It's only when you read bytes that end up needing to wrap around do things actually change.  The indexing is just relative to position 0 and position 0 moves whenever the window slides.

mtl1979:
@KungFuJesus Without any magic, it will overflow the available 16 bits of index...

KungFuJesus:
> @KungFuJesus Without any magic, it will overflow the available 16 bits of index...

The indexes themselves wrap around with modular indexing.  There can't be any overflowing because the index math is always done modulo window size.  Since window size is always a power of two, the modulo is simply an AND operation with (1 << wbits) - 1.

mtl1979:
@KungFuJesus I mean the index stored in the "previous" table... It always tracks the previous index associated with the hash value... That is always relative to start of allocation, not start of window (which would slide within the allocation boundaries).

KungFuJesus:
> @KungFuJesus I mean the index stored in the "previous" table... It always tracks the previous index associated with the hash value... That is always relative to start of allocation, not start of window (which would slide within the allocation boundaries).

Right that can still be the same.  At the moment, the start of the allocation is the start of the window and bytes are slid out backward such that the beginning of the allocation is always the start of the buffer.  That indexing doesn't have to change, we just change what the "start" of the buffer is.  The only thing we have to do is detect when the indexing will wrap around, and we simply do that by tracking the relative position within the allocation.  If something is going to wrap, anything accessing the wrapped bytes gets a temporarily combined version of the two halves.  Anything that writes those back ends up writing back the two halves.  

We can do all of this with a very simple abstraction layer so that the code performing this can't tell the difference.  Basically the harder part to doing all of this is already taken care of in the code.  We just invert the problem by moving to the proper bytes in the buffer rather than moving the bytes to the starting position.

nmoinvaz:
I'm just not so sure the cache locality really adds any benefits. Can't both pending buffer and window buffer just fit in the cache together?

KungFuJesus:
> I'm just not so sure the cache locality really adds any benefits. Can't both pending buffer and window buffer just fit in the cache together?

The copy elision with the checksum saves you a trip to cache entirely for the reads, since you basically only read the bytes once.  At that point, you're not just preventing cache pollution from an extra read to something that's about to expire, anyway, but you're always saving a trip to cache and potentially memory.  Basically, the bytes are in your registers by that point and you have execution ports dedicated for writing that are being almost entirely unused.  

Basically Jim's idea was really good but his execution prevented us from reaping the benefits, since they were lost almost entirely in the memmove's.

KungFuJesus:
I'll see about giving my idea a spin tonight after work, I'll need to identify every time the inflation window gets written to or read from.

KungFuJesus:
So reading some of PR #1033 - I'm not entirely sure I follow the significance of the history window versus pending.  However, we could probably get most of the benefits here while still stitching things together from two buffers.  Can't we just modify the "update_window" routine to call the CRC + checksum code?  The existing checksum would be carried forward via the "crc_fold" buffer we carry around.  For the adler checksum we can just emulate this for now but writing a copy that happens with the checksum should not be difficult.  

That way we can keep pretty much everything where it is while still getting free copies.  We may not the perfectly contiguous advantage (so, we're probably calling the "partial" CRC fold bits a little more than we should), but other than I think we should still see a benefit.

I'm still looking at possibly adapting the reorganized inflate window to do this, though.  Part of the issue is that the inflate loop likes to continuously update a pointer position rather than an index.  I also don't know that I fully appreciate what the history buffer is doing here.

KungFuJesus:
Ugh, adapting either solution seems more complicated than I'd want.  In the pre-PR #1033 variant, it looks as though it'd be _relatively_ simple, we just modify anything that writes to out/put to also run the bytes through a checksum.  There seem to also be dedicated op code bits that are for doing checksums (case CHECK), I'm not entirely sure what that's there for but it writes to state->check and state->adler.

What's the distinction between state->adler and state->check?  Also, there appear to be several output buffers?  Is this the case (stream->next_out)?  It's a bit hard to intuit exactly the flow of this code by just trying to read comments and follow the presumed control flow of the code.

nmoinvaz:
- `state->check` appears to be the crc32 checksum for gzip header
- `state->adler32` is either the checksum for the uncompressed data (crc32 for gzip or adler32 for deflate)

nmoinvaz:
My understanding is that it uses `stream->next_out` as the pending buffer. The `stream->next_out` is a user supplied memory buffer. When `updatewindow` is called in `inf_leave` it will copy the inflated bytes in the pending buffer to the history window `state->window`. This does seem like it might be a place to try and do a `crc32_fold`.

One thing to be mindful of is that `updatewindow` might not be called due to the `if` statement above - but for the case of using it as `crc32_fold` it will need to be done regardless of some of those conditions.

KungFuJesus:
Ah interestingly it also seems to copy things back in a somewhat different order from the actual steam of bytes if I'm reading this correctly. Conditionally doing the checksum copy or just a plain copy isn't hard to account for, we're doing that now.

So, updatewindow is where we can put this?  Oh I see what you mean by the conditional: https://github.com/zlib-ng/zlib-ng/blob/fa118e26e83ea62227b80f25248313a1cf42dcfe/inflate.c#L1005

I don't think incurring an extra copy for the very last window will be worse than the gains from getting free copies from the checksums.

*edit*: Actually looking at that macro, it looks like it was there to account for s/390 not needing it.  I think we also modify the code to call the standalone checksum without a folded copy if on s/390 if that is in fact more efficient.  I guess we'd need to ask the z/OS guy in here.  @iii-i ?

nmoinvaz:
> ...it also seems to copy things back in a somewhat different order from the actual steam of bytes if I'm reading this correctly.

How so?

KungFuJesus:
> > ...it also seems to copy things back in a somewhat different order from the actual steam of bytes if I'm reading this correctly.
> 
> How so?

https://github.com/zlib-ng/zlib-ng/blob/fa118e26e83ea62227b80f25248313a1cf42dcfe/inflate.c#L226

I might be misreading exactly how it works, but it looks like window + wsize is copied to before window.  Perhaps only one of those copies are what I need to take note of?

nmoinvaz:
Actually, I think you are right, in the first copy:
```
memcpy(state->window, end - state->wsize, state->wsize);
```
It is copying only the last 32k of pending bytes. So if pending bytes is using 64k buffer, it will only copy the last 32k. 

The other two `memcpy`s are pretty straightforward I think because they just copy to the next window pointer, which may circle back around to the start of the window. So you would need to take note of both those copies.

KungFuJesus:
> Actually, I think you are right, in the first copy:
> 
> ```
> memcpy(state->window, end - state->wsize, state->wsize);
> ```
> 
> It is copying only the last 32k of pending bytes. So if pending bytes is using 64k buffer, it will only copy the last 32k.
> 
> The other two `memcpy`s are pretty straightforward I think because they just copy to the next window pointer, which may circle back around to the start of the window. So you would need to take note of both those copies.

Does that mean that there's some overlap happening there?  Is this piecewise memcpy meant to prevent that issue?  It's not a show stopper if so, but it does mean we need be careful to do things outside of the overlap first in a way that may or not need to resort to temporary buffers, since the order the bytes matters when running them through the checksum.

Do both of these branches get taken from the inf_leave label or does realistically only one?  I think realistically only one of the two copies in the second branch needs to also include the checksum bytes (I think the first copy).  I also believe that since we copy out from "next_out" that there's no overlap to worry about when called from inf_leave, so there should be a clean mechanism for this.

KungFuJesus:
In any case, I think we have an avenue for a copy elision here, and I think staring at this long enough this seems adaptable and likely just as or even more efficient than what's on HEAD right now when combined with a copy + checksum.

I say we merge this and make the copy elision a separate PR, this is an obvious win for the fact that it doesn't require sliding bytes around.  What we have now _might_ require the memmove that's happening.  It's frustratingly hard to tell.

KungFuJesus:
Oh I finally unwound what that thing's doing.  This should still be relatively straightforward.  The input is expressed as an offset from end, so the copy actually is in the order we want, albeit in a somewhat convoluted manner.  I should be able to rework this to a checksum + copy once it's merged in.

nmoinvaz:
> Does that mean that there's some overlap happening there?

With the `s->window` circular buffer it is filling up the window, and after it is full it goes to write at the start of the buffer again. So the overlap is expected and I don't think you should have to worry about it because by the time it is overlapping you should have already hashed those bytes.

The first `memcpy` is the one you need to worry about. You have to do two operations:
```
// need to copy more than window size..
if (copy >= state->wsize) {
    // hash first 0-32k pending bytes
    if (copy > state->wsize) {
        crc32_no_copy(end, copy - state->wsize);
    }
    // hash and copy 32k-+ of pending bytes
    // instead of `memcpy(state->window, end - state->wsize, state->wsize);` ...
    crc32_with_copy(state->window, end - state->wsize, state->wsize);
}
```
The other two `memcpy` they can just be `crc32_copy`.

KungFuJesus:
> > Does that mean that there's some overlap happening there?
> 
> With the `s->window` circular buffer it is filling up the window, and after it is full it goes to write at the start of the buffer again. So the overlap is expected and I don't think you should have to worry about it because by the time it is overlapping you should have already hashed those bytes.
> 
> The first `memcpy` is the one you need to worry about. You have to do two operations:
> 
> ```
> // need to copy more than window size..
> if (copy >= state->wsize) {
>     // hash first 0-32k pending bytes
>     if (copy > state->wsize) {
>         crc32_no_copy(end, copy - state->wsize);
>     }
>     // hash and copy 32k-+ of pending bytes
>     // instead of `memcpy(state->window, end - state->wsize, state->wsize);` ...
>     crc32_with_copy(state->window, end - state->wsize, state->wsize);
> }
> ```
> 
> The other two `memcpy` they can just be `crc32_copy`.

I think the second crc32 function there would need to happen first, since the order matters.

KungFuJesus:
I do envision a potential snag where the non-copying folded CRC requires 31 bytes but I think I have a way around that.  We can initialize "first" to be init_crc != 0 in that, then we can carry on per usual.  We also need to modify the functable to expose the copyless CRC, which we're really already faking quite a bit for the likes of aarch64.

A side note, I did attempt to get a folded CRC via carryless multiplications to work on aarch64, but it seemed to be basically within margin of error for runtime _and_ it wasn't correct.  I managed to write an inlined function that behaved similarly to the x86 vpclmulqdq but for some reason the high versus low halves argument couldn't behave consistently (sometimes I had to modify that constant to be something else in order for the results to match).

In any case, correct or not the relative number of instructions and order of instructions was more or less correct and it provided no discernible improvement.  That was somewhat disappointing, I'm guessing the "polynomial multiplication" was something ARM envisioned for new cryptographic ciphers rather than to be used to implement CRC32.

nmoinvaz:
> I think the second crc32 function there would need to happen first, since the order matters.
 
I don't think so, but you can test it and see if that is the case. The pending bytes buffer is sequential so we would be hashing the first part of it and then copy and hashing the second part of it to the history window.

nmoinvaz:
Rebased.

KungFuJesus:
> > I think the second crc32 function there would need to happen first, since the order matters.
> 
> I don't think so, but you can test it and see if that is the case. The pending bytes buffer is sequential so we would be hashing the first part of it and then copy and hashing the second part of it to the history window.

end - state->wsize comes before end.  In order to be fed to the checksum in the correct order the copy would need to happen before the non-copying checksum bits.

nmoinvaz:
I see what you are saying.
```
crc32_no_copy(end, copy - state->wsize);
```
Should be
```
crc32_no_copy(end - copy, copy - state->wsize);
```
So:
```
// need to copy more than window size..
if (copy >= state->wsize) {
    // hash first 0-32k pending bytes
    if (copy > state->wsize) {
        crc32_no_copy(end - copy, copy - state->wsize);
    }
    // hash and copy 32k-+ of pending bytes
    // instead of `memcpy(state->window, end - state->wsize, state->wsize);` ...
    crc32_with_copy(state->window, end - state->wsize, state->wsize);
}
```

iii-i:
@KungFuJesus For the s390 case, if we are using the hardware acceleration (DFLTCC), we actually need neither checksumming nor copying - DFLTCC does both on its own.

Btw, that code is for Linux on z, not z/OS :-)

KungFuJesus:
Ah, well, with a lot of those mainframes the OS sits in some hardware hypervisor managed by something like z/OS, at least that's the way LPARs worked when I dealt with that in the POWER world.  IBM supports Linux on Z on the bare metal?

iii-i:
Ah, I see what you mean. Linux on bare metal is not officially supported. The zlib-ng builder runs in z/VM. I've also tested this code in an LPAR and in KVM.

z/OS folks have [their own DFLTCC zlib integration](https://www.ibm.com/docs/en/zos/2.2.0?topic=SSLTBW_2.2.0/com.ibm.zos.v2r2.ieac100/UsingZlibMixed%20Environment.htm).

KungFuJesus:
Oh cool, good to have a working Linux userland stack there. Short-circuiting both the checksum and copy here should be trivial, then.

KungFuJesus:
Fascinating, @nmoinvaz I merged your change into a branch that contains my chunkset/copy changes and we saw drastic decode improvements on everything _except_ for levels 7, 8, & 9, where we saw a huge regression.

I may also see if we see the same behavior without that patch.  I do wonder if this might have been one of the cases where we were helped by the contiguous copy regions.

Oh wait, that branch didn't have my faster chunkcopy's integrated.  Hmm, time to do some git surgery.

A perf profile reveals chunkcopy_safe is the vast majority of time.  Since it's not invoking memcpy proper, I have to assume there's a _fair_ bit of overlap in that scenario.

nmoinvaz:
Ok let me know if I didn't revert the changes correctly and I can fix it.

KungFuJesus:
Ok, yeah we're looking great now.  Here we are compared to chromium's zlib fork:

```
Comparing ./benchmark_zlib_apps_alt to ./benchmark_zlib_apps
Benchmark                                                            Time             CPU      Time Old      Time New       CPU Old       CPU New
-------------------------------------------------------------------------------------------------------------------------------------------------
png_decode/png_decode/0_pvalue                                     0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode/png_decode/0_mean                                      -0.2059         -0.2059           774           615           774           615
png_decode/png_decode/0_median                                    -0.2061         -0.2061           774           614           774           614
png_decode/png_decode/0_stddev                                    -0.2614         -0.2613             2             2             2             2
png_decode/png_decode/0_cv                                        -0.0699         -0.0697             0             0             0             0
png_decode/png_decode/1_pvalue                                     0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode/png_decode/1_mean                                      -0.2731         -0.2731           617           448           617           448
png_decode/png_decode/1_median                                    -0.2756         -0.2756           619           449           619           449
png_decode/png_decode/1_stddev                                    +0.3879         +0.3880             5             7             5             7
png_decode/png_decode/1_cv                                        +0.9092         +0.9094             0             0             0             0
png_decode/png_decode/2_pvalue                                     0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode/png_decode/2_mean                                      -0.2460         -0.2460           609           459           609           459
png_decode/png_decode/2_median                                    -0.2428         -0.2428           611           463           611           463
png_decode/png_decode/2_stddev                                    +0.3939         +0.3942             5             7             5             7
png_decode/png_decode/2_cv                                        +0.8488         +0.8492             0             0             0             0
png_decode/png_decode/3_pvalue                                     0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode/png_decode/3_mean                                      -0.2229         -0.2229           592           460           592           460
png_decode/png_decode/3_median                                    -0.2245         -0.2245           594           461           594           461
png_decode/png_decode/3_stddev                                    +0.0586         +0.0586             5             5             5             5
png_decode/png_decode/3_cv                                        +0.3622         +0.3622             0             0             0             0
png_decode/png_decode/4_pvalue                                     0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode/png_decode/4_mean                                      -0.6691         -0.6691          1300           430          1300           430
png_decode/png_decode/4_median                                    -0.6655         -0.6655          1303           436          1303           436
png_decode/png_decode/4_stddev                                    +0.3452         +0.3449             8            11             8            11
png_decode/png_decode/4_cv                                        +3.0648         +3.0639             0             0             0             0
png_decode/png_decode/5_pvalue                                     0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode/png_decode/5_mean                                      -0.6454         -0.6454          1287           456          1287           456
png_decode/png_decode/5_median                                    -0.6442         -0.6442          1288           458          1288           458
png_decode/png_decode/5_stddev                                    +0.1546         +0.1545             5             5             5             5
png_decode/png_decode/5_cv                                        +2.2558         +2.2556             0             0             0             0
png_decode/png_decode/6_pvalue                                     0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode/png_decode/6_mean                                      -0.6324         -0.6324          1281           471          1281           471
png_decode/png_decode/6_median                                    -0.6317         -0.6317          1279           471          1279           471
png_decode/png_decode/6_stddev                                    -0.3027         -0.3026            10             7            10             7
png_decode/png_decode/6_cv                                        +0.8969         +0.8972             0             0             0             0
png_decode/png_decode/7_pvalue                                     0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode/png_decode/7_mean                                      -0.3837         -0.3837          1269           782          1269           782
png_decode/png_decode/7_median                                    -0.3851         -0.3851          1268           780          1268           780
png_decode/png_decode/7_stddev                                    +1.5789         +1.5778             3             8             3             8
png_decode/png_decode/7_cv                                        +3.1845         +3.1826             0             0             0             0
png_decode/png_decode/8_pvalue                                     0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode/png_decode/8_mean                                      -0.4063         -0.4063          1247           740          1247           740
png_decode/png_decode/8_median                                    -0.4080         -0.4080          1248           739          1248           739
png_decode/png_decode/8_stddev                                    +3.2422         +3.2405             3            13             3            13
png_decode/png_decode/8_cv                                        +6.1456         +6.1427             0             0             0             0
png_decode/png_decode/9_pvalue                                     0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode/png_decode/9_mean                                      -0.4081         -0.4081          1247           738          1247           738
png_decode/png_decode/9_median                                    -0.4078         -0.4078          1248           739          1248           739
png_decode/png_decode/9_stddev                                    +0.1949         +0.1953             4             5             4             5
png_decode/png_decode/9_cv                                        +1.0186         +1.0194             0             0             0             0
png_decode_realistic/png_decode_realistic/0_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode_realistic/png_decode_realistic/0_mean                  +0.0825         +0.0825          9980         10803          9979         10803
png_decode_realistic/png_decode_realistic/0_median                +0.0851         +0.0851          9971         10819          9970         10819
png_decode_realistic/png_decode_realistic/0_stddev                +1.7872         +1.7880            26            71            26            71
png_decode_realistic/png_decode_realistic/0_cv                    +1.5746         +1.5754             0             0             0             0
png_decode_realistic/png_decode_realistic/1_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode_realistic/png_decode_realistic/1_mean                  +0.1366         +0.1366         23440         26641         23439         26640
png_decode_realistic/png_decode_realistic/1_median                +0.1366         +0.1366         23438         26639         23438         26638
png_decode_realistic/png_decode_realistic/1_stddev                +5.6739         +5.6589             7            45             7            45
png_decode_realistic/png_decode_realistic/1_cv                    +4.8720         +4.8588             0             0             0             0
png_decode_realistic/png_decode_realistic/2_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode_realistic/png_decode_realistic/2_mean                  +0.1034         +0.1034         22816         25175         22816         25175
png_decode_realistic/png_decode_realistic/2_median                +0.1036         +0.1036         22798         25161         22798         25160
png_decode_realistic/png_decode_realistic/2_stddev                -0.0197         -0.0190            44            43            44            43
png_decode_realistic/png_decode_realistic/2_cv                    -0.1115         -0.1109             0             0             0             0
png_decode_realistic/png_decode_realistic/3_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode_realistic/png_decode_realistic/3_mean                  +0.0875         +0.0875         22256         24203         22255         24202
png_decode_realistic/png_decode_realistic/3_median                +0.0879         +0.0879         22246         24202         22245         24202
png_decode_realistic/png_decode_realistic/3_stddev                -0.6251         -0.6258            27            10            27            10
png_decode_realistic/png_decode_realistic/3_cv                    -0.6553         -0.6559             0             0             0             0
png_decode_realistic/png_decode_realistic/4_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode_realistic/png_decode_realistic/4_mean                  +0.0514         +0.0514         23851         25077         23851         25076
png_decode_realistic/png_decode_realistic/4_median                +0.0513         +0.0513         23849         25074         23849         25073
png_decode_realistic/png_decode_realistic/4_stddev                +0.4693         +0.4701             8            12             8            12
png_decode_realistic/png_decode_realistic/4_cv                    +0.3975         +0.3983             0             0             0             0
png_decode_realistic/png_decode_realistic/5_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode_realistic/png_decode_realistic/5_mean                  +0.0447         +0.0447         23364         24407         23363         24406
png_decode_realistic/png_decode_realistic/5_median                +0.0447         +0.0447         23363         24407         23362         24406
png_decode_realistic/png_decode_realistic/5_stddev                -0.7919         -0.7916            33             7            33             7
png_decode_realistic/png_decode_realistic/5_cv                    -0.8008         -0.8005             0             0             0             0
png_decode_realistic/png_decode_realistic/6_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode_realistic/png_decode_realistic/6_mean                  +0.0446         +0.0446         22726         23739         22725         23738
png_decode_realistic/png_decode_realistic/6_median                +0.0451         +0.0451         22714         23738         22713         23738
png_decode_realistic/png_decode_realistic/6_stddev                -0.4374         -0.4360            26            15            26            15
png_decode_realistic/png_decode_realistic/6_cv                    -0.4614         -0.4600             0             0             0             0
png_decode_realistic/png_decode_realistic/7_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode_realistic/png_decode_realistic/7_mean                  +0.0420         +0.0420         22628         23578         22627         23577
png_decode_realistic/png_decode_realistic/7_median                +0.0421         +0.0421         22629         23580         22628         23580
png_decode_realistic/png_decode_realistic/7_stddev                -0.5962         -0.5969            23             9            23             9
png_decode_realistic/png_decode_realistic/7_cv                    -0.6125         -0.6131             0             0             0             0
png_decode_realistic/png_decode_realistic/8_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode_realistic/png_decode_realistic/8_mean                  +0.0390         +0.0390         22445         23320         22444         23320
png_decode_realistic/png_decode_realistic/8_median                +0.0389         +0.0389         22446         23319         22446         23318
png_decode_realistic/png_decode_realistic/8_stddev                +0.0729         +0.0811            12            12            12            13
png_decode_realistic/png_decode_realistic/8_cv                    +0.0326         +0.0405             0             0             0             0
png_decode_realistic/png_decode_realistic/9_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_decode_realistic/png_decode_realistic/9_mean                  +0.0381         +0.0381         22133         22976         22132         22975
png_decode_realistic/png_decode_realistic/9_median                +0.0381         +0.0381         22132         22975         22131         22975
png_decode_realistic/png_decode_realistic/9_stddev                +0.9148         +0.8779             7            13             7            13
png_decode_realistic/png_decode_realistic/9_cv                    +0.8446         +0.8090             0             0             0             0
png_encode/encode_compressible/0_pvalue                            0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_encode/encode_compressible/0_mean                             -0.0692         -0.0692           772           719           772           719
png_encode/encode_compressible/0_median                           -0.0684         -0.0684           772           719           772           719
png_encode/encode_compressible/0_stddev                           +0.3257         +0.3255             1             2             1             2
png_encode/encode_compressible/0_cv                               +0.4242         +0.4241             0             0             0             0
png_encode/encode_compressible/1_pvalue                            0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_encode/encode_compressible/1_mean                             -0.4240         -0.4240          1490           858          1490           858
png_encode/encode_compressible/1_median                           -0.4264         -0.4264          1492           856          1492           856
png_encode/encode_compressible/1_stddev                           +4.9990         +5.0014             3            21             3            21
png_encode/encode_compressible/1_cv                               +9.4140         +9.4182             0             0             0             0
png_encode/encode_compressible/2_pvalue                            0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_encode/encode_compressible/2_mean                             -0.2272         -0.2272          1487          1149          1487          1149
png_encode/encode_compressible/2_median                           -0.2282         -0.2282          1489          1149          1489          1149
png_encode/encode_compressible/2_stddev                           +4.5809         +4.5826             3            17             3            17
png_encode/encode_compressible/2_cv                               +6.2217         +6.2238             0             0             0             0
png_encode/encode_compressible/3_pvalue                            0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_encode/encode_compressible/3_mean                             -0.1640         -0.1640          1475          1233          1475          1233
png_encode/encode_compressible/3_median                           -0.1650         -0.1650          1477          1233          1477          1233
png_encode/encode_compressible/3_stddev                           +4.6263         +4.6271             3            16             3            16
png_encode/encode_compressible/3_cv                               +5.7298         +5.7309             0             0             0             0
png_encode/encode_compressible/4_pvalue                            0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_encode/encode_compressible/4_mean                             -0.8595         -0.8595          8179          1149          8179          1149
png_encode/encode_compressible/4_median                           -0.8410         -0.8410          7232          1150          7232          1149
png_encode/encode_compressible/4_stddev                           -0.9946         -0.9946          3018            16          3018            16
png_encode/encode_compressible/4_cv                               -0.9617         -0.9617             0             0             0             0
png_encode/encode_compressible/5_pvalue                            0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_encode/encode_compressible/5_mean                             -0.7584         -0.7584          7289          1761          7289          1761
png_encode/encode_compressible/5_median                           -0.7582         -0.7582          7283          1761          7283          1761
png_encode/encode_compressible/5_stddev                           -0.4883         -0.4884            34            17            34            17
png_encode/encode_compressible/5_cv                               +1.1180         +1.1176             0             0             0             0
png_encode/encode_compressible/6_pvalue                            0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_encode/encode_compressible/6_mean                             -0.7318         -0.7318          8213          2203          8213          2203
png_encode/encode_compressible/6_median                           -0.7319         -0.7319          8215          2202          8215          2202
png_encode/encode_compressible/6_stddev                           -0.4111         -0.4111            29            17            29            17
png_encode/encode_compressible/6_cv                               +1.1961         +1.1959             0             0             0             0
png_encode/encode_compressible/7_pvalue                            0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_encode/encode_compressible/7_mean                             -0.3348         -0.3348          8415          5598          8415          5598
png_encode/encode_compressible/7_median                           -0.3339         -0.3339          8417          5606          8417          5606
png_encode/encode_compressible/7_stddev                           +0.0034         +0.0027            29            29            29            29
png_encode/encode_compressible/7_cv                               +0.5085         +0.5074             0             0             0             0
png_encode/encode_compressible/8_pvalue                            0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_encode/encode_compressible/8_mean                             -0.3228         -0.3228          9853          6672          9853          6672
png_encode/encode_compressible/8_median                           -0.3227         -0.3227          9848          6670          9848          6670
png_encode/encode_compressible/8_stddev                           -0.2156         -0.2173            32            25            32            25
png_encode/encode_compressible/8_cv                               +0.1583         +0.1558             0             0             0             0
png_encode/encode_compressible/9_pvalue                            0.0002          0.0002      U Test, Repetitions: 10 vs 10
png_encode/encode_compressible/9_mean                             -0.2228         -0.2228          9855          7659          9855          7659
png_encode/encode_compressible/9_median                           -0.2225         -0.2225          9852          7660          9852          7660
png_encode/encode_compressible/9_stddev                           -0.5145         -0.5157            34            16            34            16
png_encode/encode_compressible/9_cv                               -0.3752         -0.3769             0             0             0             0
OVERALL_GEOMEAN                                                   -0.3271         -0.3271             0             0             0             0
```

We're not always winning on decode, but I can't wait to elide the copy + checksum, I suspect that'll give us some decent gains.

KungFuJesus:
Interestingly, on the real imagery (that test imagery from wikipedia) this is a net loss in decode performance:

### Develop + PR #1173 
```
--------------------------------------------------------------------------------------
Benchmark                                            Time             CPU   Iterations
--------------------------------------------------------------------------------------
png_decode_realistic/png_decode_realistic/0      10559 us        10559 us           65
png_decode_realistic/png_decode_realistic/1      24454 us        24453 us           28
png_decode_realistic/png_decode_realistic/2      20643 us        20642 us           34
png_decode_realistic/png_decode_realistic/3      18528 us        18528 us           37
png_decode_realistic/png_decode_realistic/4      22842 us        22841 us           30
png_decode_realistic/png_decode_realistic/5      21579 us        21578 us           32
png_decode_realistic/png_decode_realistic/6      20504 us        20504 us           34
png_decode_realistic/png_decode_realistic/7      20301 us        20300 us           34
png_decode_realistic/png_decode_realistic/8      19915 us        19915 us           35
png_decode_realistic/png_decode_realistic/9      19502 us        19501 us           36
```

### Develop + PR #1178 + PR #1173
```
--------------------------------------------------------------------------------------
Benchmark                                            Time             CPU   Iterations
--------------------------------------------------------------------------------------
png_decode_realistic/png_decode_realistic/0      11513 us        11513 us           60
png_decode_realistic/png_decode_realistic/1      27164 us        27163 us           26
png_decode_realistic/png_decode_realistic/2      25527 us        25527 us           27
png_decode_realistic/png_decode_realistic/3      24569 us        24569 us           28
png_decode_realistic/png_decode_realistic/4      25474 us        25473 us           27
png_decode_realistic/png_decode_realistic/5      24777 us        24776 us           28
png_decode_realistic/png_decode_realistic/6      24045 us        24045 us           29
png_decode_realistic/png_decode_realistic/7      23893 us        23892 us           29
png_decode_realistic/png_decode_realistic/8      23659 us        23658 us           29
png_decode_realistic/png_decode_realistic/9      23297 us        23296 us           30
```

I believe these to be the very real cache benefits for having contiguous runs and it might have been partially responsible for why we were beating google for deflating these real images (that, plus a faster adler checksum).  I think we can regain some of this by making the copy + checksum change, but it is interesting.  On heavily compressible data it seems like the code spent a lot of time sliding those bytes backward in the window.  On less compressible things, not so much.  In a world where code size and maintainability _didn't_ matter, keeping both around and deriving a heuristic to use one inflate over the other might make some sense.

Unfortunately that's quite a bit of upkeep and undertaking to deal with.  What's interesting is that for your larger test dataset, it was a net win across the board for minigzip (and that data was only marginally compressible).  It's curious what other differences I may be able to unearth here.

KungFuJesus:
We are losing to google on the real imagery by a smaller margin than we are to our prior inflate technique (makes sense, we're pretty much back to google's (and stock zlibs?) behavior for inflate.  The difference I did notice in the profiles is that they seem to be calling their "inflate_fast" under more circumstances.  The headers revealed a smaller min (theirs is 6 as opposed to our 8), but even with me adjusting that our "inflate_fast" doesn't seem to be called, so something must be setting up the "have" bytes inside the inflate state struct to be something smaller for whatever reason.

nmoinvaz:
@KungFuJesus in current develop it uses 32k pending buffer size which is the second half of the window. But in libpng, I suspect the pending buffer size to be much smaller. Looking at the libpng source it might only be allocating enough space for a single row. See `->next_out` memory allocation or `->avail_out` at the start of first `inflate` call. I suspect the different buffer sizes might make up for the difference in speed.

KungFuJesus:
> @KungFuJesus in current develop it uses 32k pending buffer size which is the second half of the window. But in libpng, I suspect the pending buffer size to be much smaller. Looking at the libpng source it might only be allocating enough space for a single row. See `->next_out` memory allocation or `->avail_out` at the start of first `inflate` call. I suspect the different buffer sizes might make up for the difference in speed.

Hmm, the difference in speed with zlib-chromium and us + these PRs or the difference in speed with develop + these PRs?  Because we're linking to the same libpng here.  Should I try bumping up a buffer size?

nmoinvaz:
It would account for the difference in speed between develop and this PR.

nmoinvaz:
I think you might be right about the `INFLATE_FAST_MIN_HAVE` being different between our code and chromium's fork. It might make more sense to spend time using 6 instead of 8 and benchmarking to see if we can make some gains.

KungFuJesus:
Weirdly bumping that down to 6 did not make inflate_fast get invoked. So I dunno, is have or left really small in our version versus chromium's for some reason?

Also, if we can regain the speed that's on develop by simply oversizing a buffer on our end that's well worth it in my opinion. Where do I specify this allocation size? It can't be at the caller because I made no changes from within libpng.

nmoinvaz:
@KungFuJesus the buffer I was referring to was the `next_out` buffer specified by libpng. https://github.com/glennrp/libpng/blob/a37d4836519517bdce6cb9d956092321eca3e73b/pngpread.c#L657

I'm not sure if that is the correct instance, but you can see at least in that one that it is only using the number of bytes required for an entire row. If the row is small then the pending bytes buffer is small and performance is poor.

nmoinvaz:
On another thought -- I suppose having a large `next_out` buffer, say 64k, has speed benefits, because it only has to copy the last 32k of the pending bytes buffer for the window each time `updatewindow` is called. Where as, if the pending bytes buffer size is <= 32k then it will end up copying every single byte that is inflated/decompressed into the pending bytes buffer in `updatewindow`.

nmoinvaz:
Because `minigzip` uses a large pending bytes buffer size of 128k it should benefit from this - having to copy bytes into the history window less frequently. 

With `develop` code that contains the refactored inflate window, the inflate window was split in half - one half was the history window and the other half was pending bytes buffer - each 32k (2 * window size). The `develop` code updated the window when the `window_flush_output` called, but that would happen every 32k, so all inflated/decompressed bytes would end up being copied to the history window no matter what. So it could not benefit the same way that `updatewindow` code does.

KungFuJesus:
Hmm, this could explain why minigzip was across the board faster at inflate than minideflate, maybe, for even the same input files.  Can we do a similar change where the pending bytes gets its own separate buffer instead of sharing the same space in the window?  Or can we just do something like allocate double the window size during inflation?

nmoinvaz:
I have added a commit https://github.com/zlib-ng/zlib-ng/pull/1175/commits/2f15152e2d9e64870974c3a268346bac8d196934 that should bring `minideflate` default buffer sizes up to match `minigzip`. Note that with `minideflate` you can also specify custom buffer sizes if you wanted via the command line using `-r/-t`.

Also if you are curious you can find the benchmarks I did with regard to buffer size here https://github.com/zlib-ng/zlib-ng/pull/868. That PR mostly dealt with deflate performance and here in this PR we discovered the inflate performance aspects.

KungFuJesus:
Hmm, still not quite as fast as minigzip, strangely:

```
Tool: minigzip
Activated single file mode
Level 0-9: /tmp/silesia.tar  202.1 MiB   211,957,760 B
Starting run 1 of 15
Testing level 0: cd   0.017   0.023     211,973,953 100.008%
Testing level 1: cd   0.827   0.416      94,127,290  44.409%
Testing level 2: cd   1.512   0.435      75,282,961  35.518%
Testing level 3: cd   1.872   0.417      71,816,478  33.882%
Testing level 4: cd   2.156   0.414      70,315,668  33.174%
Testing level 5: cd   2.382   0.410      69,225,542  32.660%
Testing level 6: cd   2.888   0.405      68,902,222  32.508%
Testing level 7: cd   4.272   0.387      68,366,800  32.255%
Testing level 8: cd   6.650   0.409      68,180,776  32.167%
Testing level 9: cd   7.106   0.394      67,586,442  31.887%
```

```
Tool: minideflate
Activated single file mode
Level 0-9: /tmp/silesia.tar  202.1 MiB   211,957,760 B
Starting run 1 of 15
Testing level 0: cd   0.010   0.016     211,973,941 100.008%
Testing level 1: cd   0.996   0.466      94,127,278  44.409%
Testing level 2: cd   1.778   0.496      75,282,949  35.518%
Testing level 3: cd   2.241   0.476      71,816,466  33.882%
Testing level 4: cd   2.576   0.461      70,315,656  33.174%
Testing level 5: cd   2.838   0.473      69,225,529  32.660%
Testing level 6: cd   3.421   0.464      68,902,210  32.508%
Testing level 7: cd   4.900   0.442      68,366,788  32.255%
Testing level 8: cd   7.433   0.450      68,180,764  32.167%
Testing level 9: cd   7.758   0.463      67,586,430  31.887%
```

It is a marked improvement though.  Pre that patch:
```
Tool: minideflate
Activated single file mode
Level 0-9: /tmp/silesia.tar  202.1 MiB   211,957,760 B
Starting run 1 of 15
Testing level 0: cd   0.046   0.049     211,990,111 100.015%
Testing level 1: cd   1.030   0.644      94,124,037  44.407%
Testing level 2: cd   1.815   0.659      75,282,949  35.518%
Testing level 3: cd   2.241   0.636      71,816,466  33.882%
Testing level 4: cd   2.615   0.615      70,315,656  33.174%
Testing level 5: cd   2.857   0.611      69,249,937  32.672%
Testing level 6: cd   3.438   0.599      68,921,480  32.517%
Testing level 7: cd   4.995   0.590      68,366,788  32.255%
Testing level 8: cd   7.814   0.598      68,180,764  32.167%
Testing level 9: cd   8.225   0.625      67,586,430  31.887%
```

It'd be great if we could somehow reap these benefits regardless of user specified buffers.  If I understand what you're saying - the reorganization kind of accomplished that for us just because it needed double the window width in order cleanly sort pending output versus current working window.  Perhaps a similar sort of overallocation could be done internally to get much of the performance back, unbeknownst to the calling library.  Obviously we _could_ author libpng-ng or propose patches upstream if they are simple enough but there are a lot of zlib consumers that could benefit from this if we did it without altering the callers.

nmoinvaz:
If we did that it would have to be the circular buffer idea - not the split buffer thing it was doing before. The thing I don't like is that it uses 2 times the memory which may not be necessary for all users of zlib. I think it would be better to do crc32 in `updatewindow` first.

KungFuJesus:
Well sure but the trade off seems pretty worthwhile. I'm tempted to patch my libpng locally, even, hah.

Yes the obvious next patch after this one is incorporating the crc back in there and after that implementing an Adler one as well.

KungFuJesus:
Although...https://github.com/glennrp/libpng/blob/a37d4836519517bdce6cb9d956092321eca3e73b/pngpread.c#L657

I'm pretty sure libpng is written in a way that each row is separately compressed with deflate compression (I think there's a single IDAT entry per row).  So in theory, that incorporates the expected total size of the decompressed stream.  So I'm not entirely sure the undersize buffer theory holds water here for the performance differences between pre and post this PR, at least for PNG images.

Clearly for things that are immensely compressible, this is a huge improvement (as much as 4x faster).  For things which aren't - maybe it really was just fitting into / sharing cache lines better.

For gzip compression, anyway, the CRC is performing the "final" fold more often than is necessary.  That alone should benefit from adding the inlined checksum back, let alone the copies that get mostly get eliminated.

KungFuJesus:
As far as why _Google's_ implementation is beating ours for the bigger test images - I did notice that a lot of their time is measured in inflate_fast, but the instructions for where most of that time lives are vectorized move instructions.  Their chunk copies are basically explicitly inlined into the function, whereas ours only chunkcopy_safe are, and even then it calls memcpy when possible.  This has very obvious advantages for larger widths, as we can tell with the very fluffy RGB stripe, but maybe we're getting taxed a bit on the function call overhead a tad bit more.  

Granted, we were getting this before, just with our own functions (dispatched via functable).

KungFuJesus:
Seems like perf places majority of our time in inflate in this part of the loop:
```
       │      if (bits < 15) {                                                                                                                                                                                                                                                                                                                                              ▒
  0.01 │ 769:   cmp     $0xe,%r12d                                                                                                                                                                                                                                                                                                                                          ▒
  1.54 │      ↓ ja      77f                                                                                                                                                                                                                                                                                                                                                 ▒
       │      load_64_bits():                                                                                                                                                                                                                                                                                                                                               ▒
       │      return chunk << bits;                                                                                                                                                                                                                                                                                                                                         ▒
  4.80 │        shlx    %r12,(%r15),%rax                                                                                                                                                                                                                                                                                                                                    ▒
       │      zng_inflate_fast():                                                                                                                                                                                                                                                                                                                                           ▒
       │      in += 6;                                                                                                                                                                                                                                                                                                                                                      ▒
  0.10 │        add     $0x6,%r15                                                                                                                                                                                                                                                                                                                                           ▒
       │      hold |= load_64_bits(in, bits);                                                                                                                                                                                                                                                                                                                               ▒
  0.66 │        or      %rax,%rbx                                                                                                                                                                                                                                                                                                                                           ▒
       │      bits += 48;                                                                                                                                                                                                                                                                                                                                                   ▒
  0.02 │        add     $0x30,%r12d                                                                                                                                                                                                                                                                                                                                         ▒
       │      here = lcode + (hold & lmask);                                                                                                                                                                                                                                                                                                                                ▒
  1.25 │ 77f:   mov     -0xb0(%rbp),%rax                                                                                                                                                                                                                                                                                                                                    ▒
  0.16 │        and     %rbx,%rax                                                                                                                                                                                                                                                                                                                                           ▒
  1.82 │      ↓ jmp     7b1                                                                                                                                                                                                                                                                                                                                                 ▒
       │        nop                                                                                                                                                                                                                                                                                                                                                         ▒
       │      } else if (op & 16) {                     /* length base */                                                                                                                                                                                                                                                                                                   ▒
  1.01 │ 790:┌─→test    $0x10,%al                                                                                                                                                                                                                                                                                                                                           ▒
  3.37 │     │↓ jne     18a0                                                                                                                                                                                                                                                                                                                                                ▒
       │     │} else if ((op & 64) == 0) {              /* 2nd level length code */                                                                                                                                                                                                                                                                                         ▒
  1.78 │     │  test    $0x40,%al                                                                                                                                                                                                                                                                                                                                           ▒
  0.01 │     │↓ jne     1ae0                                                                                                                                                                                                                                                                                                                                                ▒
       │     │here = lcode + here->val + BITS(op);                                                                                                                                                                                                                                                                                                                          ▒
  0.37 │     │  movzwl  0x2(%rdx),%edx                                                                                                                                                                                                                                                                                                                                      ▒
  0.01 │     │  shlx    %eax,%r13d,%eax                                                                                                                                                                                                                                                                                                                                     ▒
  0.02 │     │  dec     %eax                                                                                                                                                                                                                                                                                                                                                ▒
  0.06 │     │  and     %rbx,%rax                                                                                                                                                                                                                                                                                                                                           ▒
  0.11 │     │  add     %rdx,%rax                                                                                                                                                                                                                                                                                                                                           ▒
       │     │here = lcode + (hold & lmask);                                                                                                                                                                                                                                                                                                                                ▒
  0.33 │ 7b1:│  lea     (%r14,%rax,4),%rdx                                                                                                                                                                                                                                                                                                                                  ▒
       │     │DROPBITS(here->bits);                                                                                                                                                                                                                                                                                                                                         ▒
  6.84 │     │  movzbl  0x1(%rdx),%ecx                                                                                                                                                                                                                                                                                                                                      ▒
  0.04 │     │  mov     %ecx,%eax                                                                                                                                                                                                                                                                                                                                           ▒
  1.90 │     │  shrx    %rax,%rbx,%rbx                                                                                                                                                                                                                                                                                                                                      ▒
       │     │op = here->op;                                                                                                                                                                                                                                                                                                                                                ▒
  0.23 │     │  movzbl  (%rdx),%eax                                                                                                                                                                                                                                                                                                                                         ▒
       │     │DROPBITS(here->bits);                                                                                                                                                                                                                                                                                                                                         ▒
  0.16 │     │  sub     %ecx,%r12d                                                                                                                                                                                                                                                                                                                                          ▒
       │     │op = here->op;                                                                                                                                                                                                                                                                                                                                                ▒
  0.04 │     │  mov     %eax,%edi                                                                                                                                                                                                                                                                                                                                           ▒
       │     │if (op == 0) {                          /* literal */                                                                                                                                                                                                                                                                                                         ▒
  0.60 │     ├──test    %eax,%eax                                                                                                                                                                                                                                                                                                                                           ▒
  1.27 │     └──jne     790                                                                                                                                                                                                                                                                                                                                                 ▒
       │      *out++ = (unsigned char)(here->val);                                                                                                                                                                                                                                                                                                                          ▒
  5.35 │        movzwl  0x2(%rdx),%eax                                                                                                                                                                                                                                                                                                                                      ▒
  0.01 │        inc     %r8                                                                                                                                                                                                                                                                                                                                                 ▒
  1.55 │        mov     %al,-0x1(%r8)                                                                                                                                                                                                                                                                                                                                       ▒
       │      break;                                                                                                                                                                                                                                                                                                                                                        ▒
       │      } else {                                                                                                                                                                                                                                                                                                                                                      ▒
       │      SET_BAD("invalid literal/length code");                                                                                                                                                                                                                                                                                                                       ▒
       │      break;                                                                                                                                                                                                                                                                                                                                                        ▒
       │      }   
```

Hah, actually that is inflate_fast, the body has just been inlined into inflate.  That actually explains a fair bit.

KungFuJesus:
Well this is interesting...
microstats

I expected to see a significant difference but not only is the diff minor there, but the newer version should in theory be technically superior.  It has fewer migrations or context switches, fewer cache access _and_ fewer loads, for both the last level and the L1, fewer branches, fewer branches, although more of them missed in prediction.  I mean...these things are all kind of limited to the sampling resolution of the counters with perf, but I'm very surprised that it does that much worse given those very minute differences.  The IPC is technically higher on the older version, but not in any way that I'd consider above noise with the IPC counting mechanics via the PMCs.  I do suspect a fair number of those counters are lower because google benchmark needed to run the loop fewer times to get a consistent timing.

nmoinvaz:
It is hard to tell on the benchmark results which one is compiled with which library. I think perhaps it is not doing enough iterations. Have you tried `--benchmark_repetitions=10`?

KungFuJesus:
Sorry new is post this pr + #1173, the one that's not new is just develop + #1173

KungFuJesus:
> It is hard to tell on the benchmark results which one is compiled with which library. I think perhaps it is not doing enough iterations. Have you tried `--benchmark_repetitions=10`?

I did, it's not meaningfully different.  I would like to see if there's any discernibly costlier branch mispredictions with this PR, though the entire inflate decode loop is one giant finite state machine of branches.

KungFuJesus:
So it looks like there was a simplification in inffast that was removed by this revert.  I tried adding it back but of course it doesn't work.  Did this simplification rely on the reorganized window?

```
-                }
-
-                if (len > dist || dist < state->chunksize) {
-                    out = functable.chunkmemset(out, dist, len);
+                    from = window;
+                    if (wnext == 0) {           /* very common case */
+                        from += wsize - op;
+                    } else if (wnext >= op) {   /* contiguous in window */
+                        from += wnext - op;
+                    } else {                    /* wrap around window */
+                        op -= wnext;
+                        from += wsize - op;
+                        if (op < len) {         /* some from end of window */
+                            len -= op;
+                            out = functable.chunkcopy_safe(out, from, op, safe);
+                            from = window;      /* more from start of window */
+                            op = wnext;
+                            /* This (rare) case can create a situation where
+                               the first chunkcopy below must be checked.
+                             */
+                        }
+                    }
+                    if (op < len) {             /* still need some from output */
+                        len -= op;
+                        out = functable.chunkcopy_safe(out, from, op, safe);
+                        out = functable.chunkunroll(out, &dist, &len);
+                        out = functable.chunkcopy_safe(out, out - dist, len, safe);
+                    } else {
+                        out = functable.chunkcopy_safe(out, from, len, safe);
+                    }
+                } else if (extra_safe) {
+                    /* Whole reference is in range of current output. */
+                    if (dist >= len || dist >= state->chunksize)
+                        out = functable.chunkcopy_safe(out, out - dist, len, safe);
+                    else
+                        out = functable.chunkmemset_safe(out, dist, len, (unsigned)((safe - out) + 1));
                 } else {
-                    out = functable.chunkcopy(out, out - dist, len);
+                    /* Whole reference is in range of current output.  No range checks are
+                       necessary because we start with room for at least 258 bytes of output,
+                       so unroll and roundoff operations can write beyond `out+len` so long
+                       as they stay within 258 bytes of `out`.
+                    */
+                    if (dist >= len || dist >= state->chunksize)
+                        out = functable.chunkcopy(out, out - dist, len);
+                    else
+                        out = functable.chunkmemset(out, dist, len);
                 }
```

This is the diff from the revert.

KungFuJesus:
Looking at that commit a little bit more it looks like what was there before was handling wrap around when on the boundary of the window, so I guess it makes a pretty great deal of sense we can't really simplify that much without the memmove that pushed the bytes backward.

Though, we probably can handle the wrap around indexing better with modulo arithmetic, at least.  The branchiness there is probably not helping things.

KungFuJesus:
#1161 has been in for a few days now, we can probably remove the PR dependency tag

Dead2:
So, with #1185 now working and all, this PR is now ready for merge? (after rebase ofc)
IMO it makes sense to do so.

nmoinvaz:
Rebased and cleaned it up a bit to remove an extra commit.

nmoinvaz:
I'm not sure why the checks are not running. I try to force push again but they aren't being picked up.

phprus:
@nmoinvaz 
"Incident with GitHub Actions, Codespaces, Issues, and Webhooks":
https://www.githubstatus.com/incidents/fpk08rxnqjz2

KungFuJesus:
@nmoinvaz It looks like on your rebase you screwed up the inffast.  The conflict resolution should have been to drop the functable. in addition to _also_ calling things the old way.  Basically your rebase on commit: 59a2e17dfc4dee96fd557e8c422ef7ce6cc4b0bf needs to be revised.

nmoinvaz:
Ok, I think I got it fixed. I did a comparison against what I had from your PR and this.
>  in addition to also calling things the old way.

I'm not sure what you mean exactly. 

Sadly looks like the CI is still down.

KungFuJesus:
> > in addition to also calling things the old way.
> 
> I'm not sure what you mean exactly.

So your code reverted to the old method.  Simply doing that is not enough, you also have to change references _in_ the old method from `functable.chunkcopy_safe` to `chunkcopy_safe`.

KungFuJesus:
```
/home/adam/scratch/zlib-ng-inline_crc_copy/inflate.c: In function ‘inflate’:
/home/adam/scratch/zlib-ng-inline_crc_copy/inflate.c:917:80: error: ‘buf_left’ undeclared (first use in this function)
  917 |                 chunkcopy_safe(next_out, next_out - state->offset, copy, put + buf_left);
      |                                                                                ^~~~~~~~
/home/adam/scratch/zlib-ng-inline_crc_copy/inflate.c:917:80: note: each undeclared identifier is reported only once for each function it appears in
/home/adam/scratch/zlib-ng-inline_crc_copy/inflate.c:348:20: warning: unused variable ‘from’ [-Wunused-variable]
  348 |     unsigned char *from;        /* where to copy match bytes from */
      |                    ^~~~
/home/adam/scratch/zlib-ng-inline_crc_copy/inflate.c: In function ‘inflate’:
/home/adam/scratch/zlib-ng-inline_crc_copy/inflate.c:917:80: error: ‘buf_left’ undeclared (first use in this function)
  917 |                 chunkcopy_safe(next_out, next_out - state->offset, copy, put + buf_left);
      |                                                                                ^~~~~~~~
/home/adam/scratch/zlib-ng-inline_crc_copy/inflate.c:917:80: note: each undeclared identifier is reported only once for each function it appears in
/home/adam/scratch/zlib-ng-inline_crc_copy/inflate.c:348:20: warning: unused variable ‘from’ [-Wunused-variable]
  348 |     unsigned char *from;        /* where to copy match bytes from */
      |                    ^~~~
make[2]: *** [CMakeFiles/zlib.dir/build.make:342: CMakeFiles/zlib.dir/inflate.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
[ 17%] Building C object CMakeFiles/zlibstatic.dir/arch/x86/adler32_avx512_vnni.c.o
[ 17%] Building C object CMakeFiles/zlibstatic.dir/arch/x86/adler32_avx512.c.o
make[2]: *** [CMakeFiles/zlibstatic.dir/build.make:342: CMakeFiles/zlibstatic.dir/inflate.c.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:232: CMakeFiles/zlibstatic.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:206: CMakeFiles/zlib.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```

Looks like that should be left?

KungFuJesus:
This is what my merged branch looks like right around there (that I never intended to see the light of day since this project doesn't do merge commits):
```
#ifdef INFLATE_ALLOW_INVALID_DISTANCE_TOOFAR_ARRR
                    Trace((stderr, "inflate.c too far\n"));
                    copy -= state->whave;
                    copy = MIN(copy, state->length);
                    copy = MIN(copy, left);
                    left -= copy;
                    state->length -= copy;
                    do {
                        *put++ = 0;
                    } while (--copy);
                    if (state->length == 0)
                        state->mode = LEN;
                    break;
#endif
                }
                if (copy > state->wnext) {
                    copy -= state->wnext;
                    from = state->window + (state->wsize - copy);
                } else {
                    from = state->window + (state->wnext - copy);
                }
                copy = MIN(copy, state->length);
                copy = MIN(copy, left);

                put = chunkcopy_safe(put, from, copy, put + left);
            } else {                             /* copy from output */
                copy = MIN(state->length, left);

                put = functable.chunkmemset_safe(put, state->offset, copy, left);
            }
            left -= copy;
            state->length -= copy;
```

KungFuJesus:
Basically, to resolve the conflict on the rebase, accept what your changes were doing in their entirety, except modify `functable.chunkcopy_safe` to be `chunkcopy_safe`.

nmoinvaz:
@KungFuJesus I think I got it now. Yea that particular area TortoiseGit was having problems merging and I didn't quite understand it.

KungFuJesus:
Assuming this comes back error free I can rebase on your rebase and then when this gets merged it should be trivial.

KungFuJesus:
It does appear a "ZALLOC" leveraging commit was somehow dropped.  Was that intentional?

nmoinvaz:
Yes, I just didn't allow it to revert those changes during the revert commits, so there is no need for that extra commit.

nmoinvaz:
I just rechecked all the files in each commit and they match up with what was before the original comments. Not sure why all those CI instances are failing.

Essentially all the big-endian instances..

Rebased again anyways..

KungFuJesus:
Yeah, that's definitely qemu segfaulting.  Seemingly entirely on minigzip.  The commit in question that disappeared was for using ZALLOC in minigzip.  This guy: https://github.com/zlib-ng/zlib-ng/commit/831419bbc9a0f61d7fd1f9b5521ffabc7ed313d5

KungFuJesus:
Also the lack of segfaults on your other PRs suggest it's not qemu acting up or anything on the builders.  It must be that missing commit.

KungFuJesus:
Hmm, looks as though that managed to get rolled into possibly other commits, I think I see those changes in the final version.  This is strange.

KungFuJesus:
I'm seeing minor changes from a previous version I had merged up in my tree but not meaningful ones at the moment.  The one thing I do see that's perhaps meaningful is the zng_calloc_aligned stuff that was merged into develop. Perhaps that's causing issues?

Yep, I'll bet something's going on with those function pointers.  Unfortunately I can't boot up the G5 right now to test it but it's gotta be related. Well, maybe.  Come to think of it, I never did push my merged branch into a PR.  I wonder if chunkcopy_safe _could_ have something to do with it, at least how it's used here.

KungFuJesus:
So there do seem to be a fair number of issues with the unaligned test running through UBSAN for chunkmemset.  It seems to be unhappy with our usage of chunkmemset() even though we were doing the same exact thing before (dereferencing an arbitrary unaligned pointer).  Which to be fair, is being caught correctly in ubsan but...why wasn't it an issue before?

KungFuJesus:
I'm somewhat curious if you pushed a version through CI free of the better chunkcopy_safe if it would not segfault.

KungFuJesus:
The ubsan I have a PR for, that should be an easy painless fix.  If this does fix things then it's likely a bug where chunkcopy_safe is getting used and perhaps we were getting very lucky?  I might try to reproduce it on the G5 tomorrow night if possible.

nmoinvaz:
I'm going to try and revert the chunk code first to see if that is it.

KungFuJesus:
That would be interesting given that I found that calling it as-is but not compiled as strictly vectorized code, I could make it do increments that made it break the self-modifying copy behavior that was expected (e.g. increments of strictly 8 bytes at a time until it hit the 4, 2, and 1 byte increments).  Then _everything_ would break.  I went back and looked at the Google code it came from and it actually tries to detect the presence of overlap and when doing so, it falls back to a byte by byte copy (see https://chromium.googlesource.com/chromium/src/third_party/zlib/+/refs/heads/main/contrib/optimizations/chunkcopy.h#405)

KungFuJesus:
It is interesting that the failures seem to be on strictly big endian architectures.  SPARC is notoriously strict about alignment but PPC64 is not.  At the very least a reproducible segfault is a lot for me to go on, I think I can probably deduce what happened if this happens to be the culprit.  I could remotely turn on the sparc server right now but it'd might be painfully just to get this thing compiling on Solaris with standard modern build toolchains.

KungFuJesus:
The hunt begins tomorrow.

KungFuJesus:
Well good news - I just got home and am able to reproduce segfaults on native hardware:
```
 6/65 Test  #6: makecrct-crc32-combine ...........   Passed    0.06 sec
      Start  7: fuzzer_checksum
 7/65 Test  #7: fuzzer_checksum ..................   Passed    0.03 sec
      Start  8: fuzzer_compress
 8/65 Test  #8: fuzzer_compress ..................***Exception: SegFault  0.01 sec
```

Question, what does make test launch as input into fuzzer_compress?

nmoinvaz:
It feeds all the files in the repository into the fuzzers.

KungFuJesus:
Ok, I think I got it.  GDB shows a segfault in the chunk_memset_c variant, I suspect this won't be too hard to narrow down. I'm guessing that in our examples we just never ending up having to invoke chunk_memset until that patch.  Which definitely makes this a bug we'll want fixed on develop asap.  I should come back with a patch shortly.

KungFuJesus:
Hah, yeah, it was a dumb issue, I'm pretty sure.  Chunkmemset for whatever reason is sometimes being called with len 0. Now _why_ this is...well, maybe it's an endianness issue somehow?  Aliasing the front half a uint64 as a uint32?  Though, that's definitely UB that I would think the sanitizers would have found.  Or perhaps it was a case that was previously handled cleanly before but isn't now.  Compiling on x86 with debug should have triggered the assertion that complains about chunkcopy (and all versions in the functable generated from this template) that len 0 isn't allowed.

In any case, it seems to be passing all tests.  I'll try turning on more sanitizers to see if I can find out why this is is happening and maybe see if I can induce it on the x86, but I think I have something that can be a PR soon.

nmoinvaz:
Rebased and removed my top commit that was used to determine the cause of the big-endian issues.

KungFuJesus:
Ahhh, I see why, I think.  The "align" loop in chunkmemset_safe is feeding through all of the bytes of len.  Evidently UNALIGNED64 is not ok for PPC64, PPC, s390 and SPARC?  For SPARC and s/390 I can see this being true but for the PPC variants?  And for some reason PPC64 little endian is UNALIGNED64_OK as well?  In any case, that loop terminates after bringing len down to 0, which seems like a bug.  It counts len down to 0 and just assumes the if statement in the second block willl be true (left <  3*sizeof(chunk_t)).  Because this is not true, the statement that's supposed to return out never does.  Anyway, the immediate fix for this is simple, lf (!len), return out.  That way we bypass a function call that is 100% guaranteed to fail.  We could _also_ make chunkmemset and chunkcopy safe with 0 sized lens, but...it makes sense to do it here.

Also, the issue _was_ happening for x86, I just failed to see it.  This build configuration, in particular: https://github.com/zlib-ng/zlib-ng/runs/5594051447?check_suite_focus=true

KungFuJesus:
@nmoinvaz a rebase should make all CI pass now

nmoinvaz:
Rebased.

KungFuJesus:
Interesting, it looks as though ubsan runtime errors aren't flat out causing CI's to fail like they ought to, they are just printing an error but continuing:

```
436: /home/runner/work/zlib-ng/zlib-ng/arch/x86/chunkset_avx.c:24:5: runtime error: load of misaligned address 0x7f53d2ced069 for type 'uint16_t', which requires 2 byte alignment
436: 0x7f53d2ced069: note: pointer points here
436:  00 72 00  04 00 40 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  00 00 00 00 00
436:               ^ 
436: /home/runner/work/zlib-ng/zlib-ng/inffast.c:17:5: runtime error: load of misaligned address 0x7f53d2d2f783 for type 'uint64_t', which requires 8 byte alignment
436: 0x7f53d2d2f783: note: pointer points here
436:  cd  34 ff df b1 1d f9 a6 d2  e4 78 4d 27 ce 5f 9c a6  fc 6e 26 cd f7 ba 1a 2d  aa 26 89 f6 71 d2 92
436:               ^ 
436: /home/runner/work/zlib-ng/zlib-ng/inffast.c:17:5: runtime error: load of misaligned address 0x7f53d2d30f11 for type 'uint64_t', which requires 8 byte alignment
436: 0x7f53d2d30f11: note: pointer points here
436:  fe 23 f7  e0 85 dd 99 aa ea 49 ad  7a fe dc 91 e6 3b d3 3c  2d 2b 22 86 09 1c 17 bb  8b de f1 95 45
436:               ^ 
436: /home/runner/work/zlib-ng/zlib-ng/arch/x86/chunkset_avx.c:30:5: runtime error: load of misaligned address 0x7f53d2cf7b1b for type 'uint32_t', which requires 4 byte alignment
436: 0x7f53d2cf7b1b: note: pointer points here
436:  38  05 35 8c 38 ee 3c 82 43  8a 35 b2 43 9b a6 e4 43  35 10 c1 43 70 6e bd 43  db 18 ae 43 1d 8d b1
436:               ^ 
```

This is probably the nature of how the ubsan works, I guess.  I'm betting they want to find all the different instances of UB at runtime instead of just exiting at the first one.  In any case, some of this shows unaligned scalar access in places that we should be protecting with the zmemcpy macro.  I'm a little surprised the return value then ends up being 0, though.  If any error printed it should have returned non-zero.

KungFuJesus:
It looks like maybe the way to actually behave so that we catch UB in CI is to add this option to the UBSAN_OPTIONS:
`halt_on_error=1`
Of course in doing that we'll also want to fix all the cases where we still have UB.

nmoinvaz:
I have fixed some of those issues in #1195. I thought about using `halt_on_error=1` also, but wasn't sure, because it seems like `abort_on_error=1` isn't causing the CI to fail..

Dead2:
Is this ready to go in now?

nmoinvaz:
Yes

PR #1183, opened 2022-03-09
Fixed inftrees.c should be compiled with infcover 

nmoinvaz:
I think my previous commit https://github.com/zlib-ng/zlib-ng/commit/99913c9cd9d55c1211669406aeac2b5f61e79333 had this wrong. Inftrees.c should be compiled in whenever zlib is a shared library, because it doesn't have access to `zng_inflate_table` or `inflate_copyright`.

This error happens when compiling with ClangCl.


Dead2 (approved):
LGTM

PR #1186, opened 2022-03-13
Allow bypassing runtime feature check of TZCNT instructions.

mtl1979:
* This avoids conditional branch when it's known at build time that TZCNT instructions are always supported

See discussion in comments of 9bd28d93813db93aa2fb1b14309cced1efe7d721


CMakeLists.txt
```diff
@@ -792,6 +793,10 @@ if(WITH_OPTIM)
                 set(WITH_SSE4 OFF)
             endif()
         endif()
+        if(FORCE_TZCNT)
```
nmoinvaz, line 796: It might be better if this block was near FORCE_SSE2 in CMake..

nmoinvaz (commented):
Need to add this flag to README.md.

nmoinvaz:
We are also missing `FORCE_SSE2` in the README.md (but the configure flag is there). If you add `FORCE_SSE2` to readme I think it would be okay as separate commit.

CMakeLists.txt
nmoinvaz: Actually, it looks like in the CMakeLists there is a `cmake_dependent_option`. 
image

So `FORCE_TZCNT` might be different and perhaps it is okay where it is.

mtl1979:
> Need to add this flag to README.md.

Added... I was delaying non-code additions due to long backlog on GitHub actions...

nmoinvaz (approved):

Dead2 (approved):

PR #1188, opened 2022-03-15
IBM Z: Delete stale self-hosted builder containers

iii-i:
Inspired by https://github.com/zlib-ng/zlib-ng/pull/1096#issuecomment-1067275910.
This change is already deployed on the IBM Z builder.

Due to things like power outage ExecStop may not run, resulting in a
stale actions-runner container. This would prevent ExecStart from
succeeding, so try deleting such stale containers in ExecStartPre.


Dead2 (approved):
LGTM

nmoinvaz (approved):

PR #1189, opened 2022-03-15
Fix build with ClangCl

mtl1979:
* ClangCl doesn't support -mtune, so don't use it


nmoinvaz (commented):
- Turning off `-mtune` for ClangCl is a good change.
- If the ClangCl segfault is due to unaligned memory, then it should be fixed with #1096 (see [this CI run](https://github.com/zlib-ng/zlib-ng/runs/5544422985?check_suite_focus=true)), so that commit isn't necessary. 

Nice work tracking it down. I had no idea what the heck was wrong.

mtl1979:
@nmoinvaz I just added fprintf() to print out the pointer and offset... offset was multiple of 16, but the pointer did end in `8`... I could drop the "redundant" commit, but I think the 64 byte alignment is overkill for this case...

nmoinvaz:
It will get aligned to 64 bytes anyways because we make the adjustments internally and these are custom memory allocators.

mtl1979:
@nmoinvaz Rebased.

nmoinvaz:
Lot of CI appears to be failing. Maybe it just needs a bump to rerun them all..

Dead2 (approved):
LGTM

cmake/detect-intrinsics.cmake
```diff
@@ -34,7 +34,10 @@ macro(check_avx512_intrinsics)
         if(NOT NATIVEFLAG)
             # For CPUs that can benefit from AVX512, it seems GCC generates suboptimal
             # instruction scheduling unless you specify a reasonable -mtune= target
-            set(AVX512FLAG "-mavx512f -mavx512dq -mavx512bw -mavx512vl -mtune=cascadelake")
+            set(AVX512FLAG "-mavx512f -mavx512dq -mavx512bw -mavx512vl")
```
nmoinvaz, line 37: If you wanted to clean it up you could remove the double quotes.
Then use `list(APPEND AVX512FLAG -mtune=cascadelake)` below.

mtl1979: Then it would be inconsistent how concatenation for flag variables is handled for other functions in the same file... It might be worth making separate PR to cleanup the whole file and test it doesn't cause any issues...

PR #1191, opened 2022-03-16
Explicitly install dependencies for wine32.

mtl1979:
* Allow downgrading packages to resolve conflicts


mtl1979:
Adding `ppa:ondrej/php` was a temporary fix until the official repositories are updated... However, it now causes more issues than it originally fixed. As such we need now to manually downgrade two packages so amd64 and i386 packages have same version.

This will result in the offending packages to be removed and some packages to be upgraded to satisfy dependencies of wine32 package.

mtl1979:
@nmoinvaz @dead2

Dead2 (approved):
LGTM

nmoinvaz (approved):

PR #1192, opened 2022-03-16
Fixed various compiler warnings

nmoinvaz:


Dead2:
CI failure:
https://github.com/zlib-ng/zlib-ng/runs/5577068946?check_suite_focus=true

Dead2:
Still fails with
```bash
In file included from /usr/include/string.h:49[5](https://github.com/zlib-ng/zlib-ng/runs/5614959606?check_suite_focus=true#step:5:5),
                 from /home/runner/work/zlib-ng/zlib-ng/zbuild.h:10,
                 from /home/runner/work/zlib-ng/zlib-ng/inflate.c:[6](https://github.com/zlib-ng/zlib-ng/runs/5614959606?check_suite_focus=true#step:5:6):
In function ‘memcpy’,
    inlined from ‘chunkcopy_safe’ at /home/runner/work/zlib-ng/zlib-ng/inflate_p.h:236:5,
    inlined from ‘zng_inflate’ at /home/runner/work/zlib-ng/zlib-ng/inflate.c:890:1[7](https://github.com/zlib-ng/zlib-ng/runs/5614959606?check_suite_focus=true#step:5:7):
/usr/include/x[8](https://github.com/zlib-ng/zlib-ng/runs/5614959606?check_suite_focus=true#step:5:8)6_64-linux-gnu/bits/string_fortified.h:34:10: error: ‘__builtin_memcpy’ specified bound between 1844674406[9](https://github.com/zlib-ng/zlib-ng/runs/5614959606?check_suite_focus=true#step:5:9)414584321 and 18446744073709551615 exceeds maximum object size 9223372036854775807 [-Werror=stringop-overflow=]
   34 |   return __builtin___memcpy_chk (__dest, __src, __len, __bos0 (__dest));
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

mtl1979:
@Dead2 Those are negative numbers converted to unsigned integer...

First one is -4294967295, second one is -1

nmoinvaz:
Any thoughts on how to fix this? I don't really understand why it is complaining. I changed it to use `ptrdiff_t` which I believe is a smaller type but that didn't fix it.. @KungFuJesus?

KungFuJesus:
I happen to see this problem as well which is why I made that type an int.  The fortify string bits of GCC are just trying to do due diligence for all possible bounds of values that could be fed here are within range of the argument.  Making that type int happened to appease this check for arguments into memcpy.  

Basically, string fortify doesn't realize that the ternary statement for doing the actual pointer math prevents the value from ever underflowing below 0.

KungFuJesus:
The maximum value, according to that fortify message, is the max value of int64 (not uint64, surprisingly).  Perhaps just doing abs() on that subtraction is enough (probably more efficient, too), but you may still need to interpret it as signed.

nmoinvaz:
I ended up just changing it back to `int32_t` and then casting the value to `int32_t` as well.

Dead2:
New error:
`/usr/lib/gcc/x86_64-linux-gnu/9/include/avx512fintrin.h:15496:3: runtime error: signed integer overflow: 1400734789 + 1069144178 cannot be represented in type 'int'`
https://github.com/zlib-ng/zlib-ng/runs/5668908576?check_suite_focus=true

Didn't check the others.

KungFuJesus:
@Dead2 That's been fixed here #1199. A rebase should make that a non-issue.

nmoinvaz:
Rebased.

KungFuJesus:
And now CI's not going to start, lol.

nmoinvaz:
Just my luck..

Dead2:
Doesn't look like there is anything I can do to re-trigger it. Github looks like it would if we didn't have any CI, no info, no warnings, no errors, just like it never existed.  🙄 💔

nmoinvaz:
I have no clue why those errors are showing in: https://github.com/zlib-ng/zlib-ng/runs/5681144884?check_suite_focus=true.
I tried to reproduce locally with no luck.

KungFuJesus:
Looks like maybe a missing prototype?

mtl1979:
Obviously unaligned functions should not exist if "No Unaligned" is mentioned on the CI run description... The prototypes are guarded by `UNALIGNED_OK`.

KungFuJesus:
Is that function guarded against that macro?  I don't think it should be if it is, the only loads it does are officially sanctioned unaligned loads via vmovdqu.  The loadu intrinsic itself is intended to have arbitrary alignment for the load and I believe that makes it the opposite of undefined behavior.

mtl1979:
> Is that function guarded against that macro? I don't think it should be if it is, the only loads it does are officially sanctioned unaligned loads via vmovdqu. The loadu intrinsic itself is intended to have arbitrary alignment for the load and I believe that makes it the opposite of undefined behavior.

I'm just stating what is likely the issue... Mixing `UNALIGNED_OK` and `UNALIGNED64_OK` is destined to fail and cause one or other CI runs to also fail... That kinda is the whole purpose of having CI runs, to catch hard-to-notice typos.

KungFuJesus:
Right, I'm just saying if the intent of that macro is to hide from ubsan, it's not needed here. Ubsan probably won't detect this because the access is not dereferencing an unaligned pointer, it's loading from an arbitrarily aligned address.

mtl1979:
> Right, I'm just saying if the intent of that macro is to hide from ubsan, it's not needed here. Ubsan probably won't detect this because the access is not dereferencing an unaligned pointer, it's loading from an arbitrarily aligned address.

I'm pretty sure @nmoinvaz finds a solution that actually works as intended...

nmoinvaz:
It looks like there was no bug, just codecov fails to upload for that same instance.

nmoinvaz:
Everything looks ok. Failing tests are just codecov failing to upload.

Dead2 (approved):

nmoinvaz:
> The loadu intrinsic itself is intended to have arbitrary alignment for the load and I believe that makes it the opposite of undefined behavior.

@KungFuJesus this is a great point and perhaps those intrinsics versions should not be lumped into the unaligned code. I will try and make a PR to fix it up.

PR #1194, opened 2022-03-17
Fix UBSAN's cry afoul

KungFuJesus:
Technically, we weren't actually doing this the way C wants us to,
legally.  The zmemcpy's turn into NOPs for pretty much all > 0
optimization levels and this gets us defined behavior with the
sanitizer, putting the optimized load by arbitrary alignment into the
compiler's hands instead of ours.

### DIFF
--- a/arch/x86/chunkset_avx.c
+++ b/arch/x86/chunkset_avx.c
@@ -20,15 +20,21 @@ static inline void chunkmemset_1(uint8_t *from, chunk_t *chunk) {
 }
 
 static inline void chunkmemset_2(uint8_t *from, chunk_t *chunk) {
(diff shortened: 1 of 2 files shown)


nmoinvaz (approved):
Only suggestion is to use a different variable name than `tmp`. At work we try to not name any variables `temp`. 

Maybe call it `set1` or `broadcast` or `repeat`..

Other than that it looks good.

KungFuJesus:
I was just going on the precedent set by the power8 and aarch64 ones that did the same thing.

nmoinvaz:
Ah if they do the same then it can wait for another commit if we want to clean it all up later on. Anyways, I have already reviewed it and marked it approved. Good find.

Dead2 (approved):
LGTM

PR #1196, opened 2022-03-18
Fix a latent issue with chunkmemset

KungFuJesus:
It would seem that on some platforms, namely those which are
!UNALIGNED64_OK, there was a likelihood of chunkmemset_safe_c copying all
the bytes before passing control flow to chunkcopy, a function which is
explicitly unsafe to be called with a zero length copy.

This fixes that bug for those platforms.

### DIFF
--- a/chunkset_tpl.h
+++ b/chunkset_tpl.h
@@ -171,5 +171,8 @@ Z_INTERNAL uint8_t* CHUNKMEMSET_SAFE(uint8_t *out, unsigned dist, unsigned len, 
         }
         return out;
     }
-    return CHUNKMEMSET(out, dist, len);
+    if (len)
+        return CHUNKMEMSET(out, dist, len);
+
+    return out;
 }


Dead2 (approved):

chunkset_tpl.h
```diff
@@ -171,5 +171,8 @@ Z_INTERNAL uint8_t* CHUNKMEMSET_SAFE(uint8_t *out, unsigned dist, unsigned len,
         }
         return out;
     }
-    return CHUNKMEMSET(out, dist, len);
+    if (len)
+        return CHUNKMEMSET(out, dist, len);
+
+    return out;
```
KungFuJesus, line 177: Hah, interesting that the coverage check shows this as never being hit until we revert the window layout.  It definitely explains why we don't hit this issue until then.  In any case, this coverage drop will disappear after #1178 so it's definitely not something that bothers me.

mtl1979: That's why some bugs can take years to manifest and are hard to reproduce... Sometimes it's matter of hitting cascade event before all bugs are found.

PR #1197, opened 2022-03-18
Add unit tests for compare256 variants.

nmoinvaz:
These tests confirm that the proper match length is returned when supplied two buffers with that number of matching characters.

### DIFF
--- a/test/test_compare256.cc
+++ b/test/test_compare256.cc
@@ -1,0 +1,75 @@
+/* test_compare256.cc -- compare256 unit tests
+ * Copyright (C) 2022 Nathan Moinvaziri
+ * For conditions of distribution and use, see copyright notice in zlib.h
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+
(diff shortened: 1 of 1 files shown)


nmoinvaz:
@Dead2 we need a new label for "testing framework". Maybe it can be used for fuzzers/benchmarks/gtest.. or perhaps Continuous Integration works? But it is not really CI changes..

test/test_compare256.cc
```diff
@@ -0,0 +1,75 @@
...
+            return; \
+        } \
+        compare256_match_check(func); \
+    }
+
+TEST_COMPARE256(c, compare256_c, 1)
+
+#ifdef UNALIGNED_OK
+TEST_COMPARE256(unaligned_16, compare256_unaligned_16, 1)
+#ifdef HAVE_BUILTIN_CTZ
```
Dead2, line 62: Perhaps indent these ifdef to show their hierarchy.

Dead2 (approved):

PR #1198, opened 2022-03-18
macOs M1 build fix on arm cpu checks.

devnexen:


arch/arm/arm_features.c
```diff
@@ -11,6 +11,7 @@
 #    define ID_AA64ISAR0_CRC32_VAL ID_AA64ISAR0_CRC32
 #  endif
 #elif defined(__APPLE__)
+#  define _DARWIN_C_SOURCE /* enable types aliases (eg u_int) */
```
mtl1979: This might break if zlib-ng is included by another project... Like other similar preprocessor defines, it should always be guarded.

nmoinvaz:
What error does this fix?

arch/arm/arm_features.c
devnexen: sorry I missed your comment.

devnexen:
> What error does this fix?

It enables type aliases such as u_int .. etc

mtl1979 (approved):
Preferred way would be to add _DARWIN_C_SOURCE to whole project, but we don't use regular expressions, so mixing source files with and without the define is considered safe. We also do the same with other similar defines. The relevant manual page would be compat.5 on affected systems.

PR #1199, opened 2022-03-18
Fix an issue with the ubsan for overflow

KungFuJesus:
While this didn't _actually_ cause any issues for us, technically the
_mm512_reduce_add_epi32() intrinsics returns a signed integer and it
does the very last summation in scalar GPRs as signed integers. While
the ALU still did the math properly (the negative representation is the
same addition in hardware, just interpreted differently), the sanitizer
caught window of inputs here definitely outside the range of a signed
integer for this immediate operation.

The solution, as silly as it may seem, would be to implement our own 32
bit horizontal sum function that does all of the work in vector
registers. This allows us to implicitly keep things in vector register
domain and convert at the very end after we've summed the summation.

The compiler's sanitizer doesn't know the wiser and the solution still
results in being correct.


arch/x86/adler32_avx512_vnni.c
```diff
@@ -142,7 +143,7 @@ Z_INTERNAL uint32_t adler32_avx512_vnni(uint32_t adler, const unsigned char *buf
 
         adler = partial_hsum(vs1) % BASE;
         vs1 = _mm512_zextsi128_si512(_mm_cvtsi32_si128(adler));
-        sum2 = _mm512_reduce_add_epi32(vs2) % BASE;
+        sum2 = ((uint32_t)_mm512_reduce_add_epu32(vs2)) % BASE;
```
mtl1979: Cast here seems redundant...

KungFuJesus: True, fixed.

KungFuJesus:
Hmm, not entirely sure why that statically defined defined function in that header is undefined for windows.  From the looks of this,  __AVX512F__ should be defined when compiling with AVX512:
https://docs.microsoft.com/en-us/cpp/build/reference/arch-x64?view=msvc-170

nmoinvaz:
I added some extra debugging flags to CMake to get the line numbers:
```
    #0 0x5638940b5c52 in _mm512_reduce_add_epi32 /usr/lib/gcc/x86_64-linux-gnu/9/include/avx512fintrin.h:15496
    #1 0x5638940b5c52 in adler32_avx512 /home/runner/work/zlib-ng/zlib-ng/arch/x86/adler32_avx512.c:115
    #2 0x563894071dd4 in read_buf /home/runner/work/zlib-ng/zlib-ng/deflate.c:1086
    #3 0x563894083783 in deflate_stored /home/runner/work/zlib-ng/zlib-ng/deflate_stored.c:160
    #4 0x56389406d806 in zng_deflate /home/runner/work/zlib-ng/zlib-ng/deflate.c:916
    #5 0x5638940033fb in deflate_params_Test::TestBody() /home/runner/work/zlib-ng/zlib-ng/test/test_deflate_params.cc:92
    #6 0x563894266f0a in void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) /home/runner/work/zlib-ng/zlib-ng/_deps/googletest-src/googletest/src/gtest.cc:2595
    #7 0x563894266f0a in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) /home/runner/work/zlib-ng/zlib-ng/_deps/googletest-src/googletest/src/gtest.cc:2631
    #8 0x5638941815d5 in testing::Test::Run() /home/runner/work/zlib-ng/zlib-ng/_deps/googletest-src/googletest/src/gtest.cc:2670
    #9 0x5638941825b4 in testing::TestInfo::Run() /home/runner/work/zlib-ng/zlib-ng/_deps/googletest-src/googletest/src/gtest.cc:2849
    #10 0x563894183c35 in testing::TestSuite::Run() /home/runner/work/zlib-ng/zlib-ng/_deps/googletest-src/googletest/src/gtest.cc:3008
    #11 0x5638941fc0da in testing::internal::UnitTestImpl::RunAllTests() /home/runner/work/zlib-ng/zlib-ng/_deps/googletest-src/googletest/src/gtest.cc:5866
    #12 0x56389426426a in bool testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) /home/runner/work/zlib-ng/zlib-ng/_deps/googletest-src/googletest/src/gtest.cc:2595
    #13 0x56389426426a in bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) /home/runner/work/zlib-ng/zlib-ng/_deps/googletest-src/googletest/src/gtest.cc:2631
    #14 0x56389417b17d in testing::UnitTest::Run() /home/runner/work/zlib-ng/zlib-ng/_deps/googletest-src/googletest/src/gtest.cc:5440
    #15 0x56389405e08a in RUN_ALL_TESTS() /home/runner/work/zlib-ng/zlib-ng/_deps/googletest-src/googletest/include/gtest/gtest.h:2284
    #16 0x563893f795c3 in main /home/runner/work/zlib-ng/zlib-ng/test/test_main.cc:16
    #17 0x7fd064eeb0b2 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x270b2)
    #18 0x563893f7b3dd in _start (/home/runner/work/zlib-ng/zlib-ng/test/gtest_zlib+0x25f3dd)
```

nmoinvaz:
Looks like you guys were right on the offending line number.

KungFuJesus:
Yeah I definitely narrowed it down to that intrinsic.  My surrogate function should do it, I don't know why the cmake MSVC CI was failing on that guy, though.

KungFuJesus:
*sigh*: this looks to be why it's not working for MSVC:
ignoring unknown option '/ARCH:AVX512' 

What version of MSVC is CI using?  It shouldn't be trying to compile AVX512 code if the intrinsics aren't supported.

Oh I have a rough idea of what might be going on.  I might try to include it with a different preprocessor macro.

KungFuJesus:
Ugh, fixed it in one, broke it in the other.

KungFuJesus:
I think some of the issue is that cpu_features.h pulls in fallback_builtins.h already.  That and for some reason /arch:avx512 doesn't work on MSVC?  Maybe cmake is feeding the argument wrong to cl.exe or something.  Maybe I can just throw this in another header?

KungFuJesus:
Finally.  Now, to start it all over so that cmakelists populates the file into "header" structure properly in visual studio for windows users.

nmoinvaz:
@Dead2 maybe it makes sense to have a label for sanitizing issues too.

Dead2:
The other header files containing private functions are have the postfix `_p.h`, perhaps that should be used here as well to stay consistent.

KungFuJesus:
> The other header files containing private functions are have the postfix `_p.h`, perhaps that should be used here as well to stay consistent.

We could.  What should I name it?  Technically they are utility functions that aren't necessarily specific to adler.  `avx512_funcs_p.h`?

KungFuJesus:
Looks like there's a typo in the tag name.

Dead2:
@nmoinvaz @mtl1979 Thoughts about file naming?
`avx512_funcs_p.h` does seem to be a logical alternative.

mtl1979:
> @nmoinvaz @mtl1979 Thoughts about file naming? `avx512_funcs_p.h` does seem to be a logical alternative.

I do agree that private headers should have `_p` at the end of base filename. What comes before it should not be misleading, but basically can be chosen by the person creating the file as long as it follows the pattern of similar possibly already existing files.

KungFuJesus:
Should be good to go to go now.

nmoinvaz:
Since these functions appear to only be for adler32 calculations maybe it makes sense to name the file `adler32_avx512_p.h`. If in the future they are used for multiple things then we can rename it..

Dead2 (approved):

PR #1203, opened 2022-03-20
Rename adler32_sse41 to adler32_ssse3

KungFuJesus:
As it turns out, the sum of absolute differences instruction _did_ exist
in SSSE3 all along. SSE41 introduced a stranger, less commonly used
variation of the sum of absolute difference instruction.  Knowing this,
the old SSSE3 method can be axed entirely and the SSE41 method can now
be used on CPUs only having SSSE3.

Removing this extra functable entry shrinks the code and allows for a
simpler planned refactor later for the adler checksum and copy elision.

I've kept the feature detection in the build system in case we do need to rely on
SSE41 features at any point.


CMakeLists.txt
```diff
@@ -734,11 +734,6 @@ if(WITH_OPTIM)
         if(WITH_SSE41)
             check_sse41_intrinsics()
             if(HAVE_SSE41_INTRIN)
-                add_definitions(-DX86_SSE41 -DX86_SSE41_ADLER32)
```
mtl1979, line 737: We could leave `-DX86_SSE41` as placeholder, so the block isn't completely empty. Same for configure, even though it didn't have it before... Other option would be to negate the test so only the line `set(WITH_SSE41 OFF)` would stay...

configure
```diff
@@ -1535,13 +1535,6 @@ case "${ARCH}" in
 
             check_sse41_intrinsics
 
-            if test ${HAVE_SSE41_INTRIN} -eq 1; then
-                CFLAGS="${CFLAGS} -DX86_SSE41_ADLER32"
-                SFLAGS="${SFLAGS} -DX86_SSE41_ADLER32"
-                ARCH_STATIC_OBJS="${ARCH_STATIC_OBJS} adler32_sse41.o"
-                ARCH_SHARED_OBJS="${ARCH_SHARED_OBJS} adler32_sse41.lo"
```
mtl1979, line 1542: Just replace so that only `-DX86_SSE41` is added, so we don't need to remove the whole if block and re-add it later... This makes it more consistent with how we do checks with cmake, and line `check_sse41_intrinsics` doesn't become stray...

KungFuJesus:
It's a little surprising the code coverage went down at all here given the reduction in code size. I mean it's in the noise levels but that is strange.

mtl1979:
> It's a little surprising the code coverage went down at all here given the reduction in code size. I mean it's in the noise levels but that is strange.

There is two changes in coverage, the change in functable and the adler32 tests... Apparently it does ignore changes in deleted files...

nmoinvaz:
Sometimes the code coverage is a little wonky like that.

Dead2 (approved):
LGTM

Dead2:
Reduction in code coverage is logical if the deleted file had a higher code coverage than the rest of the projects average.
Reducing the amount of code to maintain is a big win either way 👍

PR #1207, opened 2022-03-25
Correct typo in functable

rincebrain:
I could be wrong about this being an error, but I don't see any discussion suggesting this was intended, and can't immediately come up with why it would be the correct choice given the context.


Dead2:
That sure looks like a bug to me, good catch @rincebrain 👍

Dead2 (approved):

mtl1979 (approved):

nmoinvaz (approved):

PR #1208, opened 2022-03-27
Added unit test for CVE-2018-25032

nmoinvaz:
See https://github.com/madler/zlib/issues/605 for background.

Using minideflate compiled against zlib `master` fails the test.
Using minideflate compiled against the zlib-ng passes the test.


nmoinvaz:
I want to add the new test case they came up with `Z_DEFAULT_STRATEGY`.

nmoinvaz:
Rebased.

nmoinvaz:
Added the new test case against `Z_DEFAULT_STRATEGY`.

Dead2:
I think we should test this on both level 1 and level 2 as well, since as far as I have understood this problem, it is basically a buffer overrun situation when using low compression. Both deflate_quick (level 1) and deflate_fast (level 2) should thus probably get tested separately. Especially since our level 2 is essentially zlibs level 1 (with modifications of course, such as 4-byte matches instead of 3-byte, thus possibly making this problem more likely).

nmoinvaz:
@Dead2 I have added those test levels.

Dead2 (approved):

PR #1211, opened 2022-03-27
Added GitHub workflow to scan and upload to coverity.

nmoinvaz:
This workflow uses this GH action:
https://github.com/vapier/coverity-scan-action

You must add two secrets to this repository:
```
COVERITY_SCAN_EMAIL
COVERITY_SCAN_TOKEN
```
@Dead2 I figure this might make it easier for you so you don't have to manually upload the scans.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -1,0 +1,27 @@
+name: CI Coverity Scan
+# We only want to test official release code, not every pull request.
+on:
+  push:
+    branches:
+      - stable
+      - develop
+      - pre-release
+      - '2.*'
+    tags:
+      - '*'
+
+jobs:
+  coverity:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


Dead2:
Two problems with this;
One is that Coverity Scan doesn't allow very many uploads per month, so we would get blocked.
The other is that in order to get better code coverage, I do multiple (5?) scripted compiles with different settings in order to get more of the code analyzed (Though I only do x86-32/x86-64 so far).

We tried to implement a Travis task once that would do the coverity uploads only when pushed to a certain branch, unfortunately we never managed to get it actually working, so it became easier to just make a script and do it locally.

One thing that always annoyed me with Coverity Scan is that it has no support for branches at all, so testing stable/develop branches will mark problems as new/fixed/and then they come back again whatever they call it then.

Kind of wish I had access to a commercial installation of Coverity again.

PR #1212, opened 2022-03-27
Allow SSE2 and AVX2 functions with -DWITH_UNALIGNED=OFF. 

nmoinvaz:
Even though they use unaligned loads, they don't result in undefined behavior.

This is from an idea from @KungFuJesus mentioned in #1192.


nmoinvaz:
Rebased.

Dead2 (approved):

PR #1213, opened 2022-03-27
Use size_t types for len arithmetic, matching signature

KungFuJesus:
This suppresses a warning and keeps everything safely the same type.
While it's unlikely that the input for any of this will exceed the size
of an unsigned 32 bit integer, this approach is cleaner than casting and
should not result in a performance degradation.

This fixes a bug to suppress a warning which was only partially fixed.

### DIFF
--- a/arch/x86/adler32_ssse3.c
+++ b/arch/x86/adler32_ssse3.c
@@ -59,10 +59,10 @@ Z_INTERNAL uint32_t adler32_ssse3(uint32_t adler, const unsigned char *buf, size
      * additions worthwhile or if it's worth it to just eat the cost of an unaligned
      * load. This is a pretty simple test, just test if 16 - the remainder + len is
      * < 16 */
-    uint32_t max_iters = NMAX;
-    uint32_t rem = (uintptr_t)buf & 15;
-    uint32_t align_offset = 16 - rem;
-    uint32_t k = 0;
+    size_t max_iters = NMAX;
+    size_t rem = (uintptr_t)buf & 15;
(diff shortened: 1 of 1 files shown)


KungFuJesus:
Rebased

Dead2:
I am uncertain whether I like this approach. I am unaccustomed to widespread use of size_t, and we typically don't do that to such a degree anywhere else. Typically we have tried to keep datatypes smaller unless it impacts performance.

Dead2:
I am uncertain whether I like this approach. I am unaccustomed to widespread use of size_t, and we typically don't do that to such a degree anywhere else. Typically we have tried to keep datatypes smaller unless it impacts performance.

KungFuJesus:
I mean the size_t signature for length of the buffer is used everywhere (explicit in the zlib-ng interfaces, anyway).  Obviously for zlib_compat this ultimately gets passed in as an integer but, it's definitely already there.

As far as size, well if the function signature has few enough arguments, on most ISAs these end up being passed down the call stack in registers, anyway.  A wider register in this case is going to be the same as a narrower register, and for the most part on x86, a 64 bit integer operation in scalar GPRs is going to be effectively the same runtime as a 32 bit one.  All of this is to just suppress a warning anyway, the damage of being 64 bit wide down the call stack is already there.  We can cast things away so that everything is unsigned 32 bit or something, but ultimately it's coming in on a 64 bit register.

Dead2 (approved):
I had another look, and yes I see your point.

PR #1214, opened 2022-03-27
Fixed clang signed/unsigned warning in chunkcopy_safe.

nmoinvaz:


nmoinvaz:
@Dead2 maybe we should have another label for compiler warning fixes? Or just "code quality" or something along those lines..

Dead2 (approved):

KungFuJesus:
Do you get the string fortify problem with this?  I think that was a problem before, no?

nmoinvaz:
@KungFuJesus string fortify problem? I'm not sure what you are referring to. I remember there was a static analysis issue that doesn't appear with this. Perhaps it was only because of `ptrdiff_t`..

nmoinvaz:
Rebased.

KungFuJesus:
> @KungFuJesus string fortify problem? I'm not sure what you are referring to. I remember there was a static analysis issue that doesn't appear with this. Perhaps it was only because of `ptrdiff_t`..

Yes, that's the one.  The issue stemmed from the fact that builtin_memcpy has some checks to ensure the possible input can't be outside the range of what memcpy can accept.  Apparently the acceptable range is the size of int64_t, at least from what I remember from that static analyzer message.  Of course, if this somehow gets around that, maybe the issue was in the casting somehow?  Mixing signed with unsigned type maybe?  I think the switch to size_t ensures that the subtraction never ends up being a negative result because it'll always be interpreted as unsigned.

PR #1215, opened 2022-03-28
cmake: Properly probe for unaligned access

### LINKED ISSUE #1100: build option defaults are inappropriate for a compression library
> UNALIGNED_OK 		Allow unaligned reads 	ON (x86, arm):
> WITH_UNALIGNED 		Allow optimizations that use unaligned reads if safe on current arch 	ON

Any unaligned access is always undefined behaviour in C, the architecture does not matter. C code that uses unaligned access is broken by definition. Unaligned access, and also type punning through pointers, is not safe for any C code.

It is very obvious that nobody has ever fuzzed zlib-ng with ubsan, or even just run the tests with ubsan:
```
./configure --with-sanitizer=undefined
make
make test
```
throws a ton of errors.

Any system's most basic compression library is probably one of the last places where I would tolerate any undefined behavior code. This can easily result in data loss and exploitable security vulnerabilites.

The build option defaults give the impression that zlib-ng's goal is "performance at all cost", sacrificing stability, reliability, and hardened code at every possible point. With this attitude, zlib-ng has sadly no chance to ever replace original zlib.

Please change your build defaults to be **correct** instead of **performance-at-all-cost**.

rurban:
just checking the architecture is not enough.
it can be arm with -maligned-access, or even some secret bit set on intel.
or ubsan.
Fixes #1100

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -249,10 +249,55 @@ else()
     endif()
 endif()
 
+# probe for NO_UNALIGNED
+if(NOT CMAKE_CROSSCOMPILING AND WITH_UNALIGNED)
+  include(CMakePushCheckState)
+  include(CheckCSourceRuns)
+  set(OLD_CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
+  cmake_push_check_state()
+  if(CMAKE_COMPILER_IS_GNUCC
+      OR (CMAKE_C_COMPILER_ID STREQUAL AppleClang)
+      OR (CMAKE_C_COMPILER_ID STREQUAL Clang))
+    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS_DEBUG} -fno-sanitize-recover=undefined")
+    else()
+      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-sanitize-recover=undefined")
+    endif()
+  endif()
+  check_c_source_runs(
+    "
+#include <stdlib.h>
+int main() {
+    int i;
+    int *p;
+    int *q;
+    char *str;
+    str = (char *) malloc(40);
+    for (i = 0; i < 40; i++) {
+      *(str + i) = i;
+    }
+    p = (int *) (str + 1);
+    q = (int *) (str + 2);
+    return (*p == *q);
+}"
+    ALLOW_ALIGNED_ACCESS)
+  cmake_pop_check_state()
+  set(CMAKE_C_FLAGS "${OLD_CMAKE_C_FLAGS}")
+  if(NOT ALLOW_ALIGNED_ACCESS)
+    add_definitions(-DNO_UNALIGNED)
+    message(STATUS "Set NO_UNALIGNED (ubsan or strict hardware, like arm, sparc)")
+  else()
+    message(STATUS "NO_UNALIGNED not required")
+  endif()
+else()
+  add_definitions(-DNO_UNALIGNED)
+  message(STATUS "Assuming aligned access is required, as on arm with -maligned-access, ubsan or sparc")
+endif()
+
 # Set architecture alignment requirements
 if(NOT WITH_UNALIGNED)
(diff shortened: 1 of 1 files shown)


mtl1979:
This has no effect on most CI runs as they use cross-compiler.

CMakeLists.txt
```diff
@@ -249,10 +249,55 @@ else()
...
+    str = (char *) malloc(40);
+    for (i = 0; i < 40; i++) {
+      *(str + i) = i;
+    }
+    p = (int *) (str + 1);
+    q = (int *) (str + 2);
+    return (*p == *q);
+}"
+    ALLOW_ALIGNED_ACCESS)
+  cmake_pop_check_state()
```
nmoinvaz, line 284: What version of cmake requires `cmake_pop_check_state`?

CMakeLists.txt
```diff
@@ -249,10 +249,55 @@ else()
...
+  set(CMAKE_C_FLAGS "${OLD_CMAKE_C_FLAGS}")
+  if(NOT ALLOW_ALIGNED_ACCESS)
```
nmoinvaz, line 286: Tab size is not correct for these changes.

CMakeLists.txt
```diff
@@ -249,10 +249,55 @@ else()
     endif()
 endif()
 
+# probe for NO_UNALIGNED
+if(NOT CMAKE_CROSSCOMPILING AND WITH_UNALIGNED)
```
nmoinvaz, line 253: Why not `CMAKE_CROSSCOMPILING`? Shouldn't a good method be able to detect even with cross compiler?

CMakeLists.txt
```diff
@@ -249,10 +249,55 @@ else()
...
+  include(CMakePushCheckState)
+  include(CheckCSourceRuns)
+  set(OLD_CMAKE_C_FLAGS "${CMAKE_C_FLAGS}")
+  cmake_push_check_state()
+  if(CMAKE_COMPILER_IS_GNUCC
```
nmoinvaz, line 258: Formatting is inconsistent with what we have.

nmoinvaz (commented):
A lot of projects use architecture detection (at least from cursory search on GH). It seems easier to me to just maintain architecture list versus maintaining probe detection.

This PR will need configure script support too.

thesamesam:
As a distribution person, my 2c is: it's preferable if you can to stick to actual feature detection given that it can vary based on e.g. ARM type, but it's not a big deal as long as it's overridable.

CMakeLists.txt
rurban: Because you cannot run it then, unless you specify how to run it cross-compiled

nmoinvaz:
@Dead2 since we now use `memcpy` everywhere even for `memcmp`, I think this PR is no longer needed..

Dead2:
I agree, this should no longer be a problem.

There might be a need for going through the UNALIGNED[64]_OK related checks to see whether those are still relevant restrictions any more, I am not aware of the current state of those. IIRC we also use those checks for selecting optimizations using unaligned intrinsics as well in places, so maybe they are still useful or maybe it needs a refactor with more appropriate checks.

nmoinvaz:
I think `UNALIGNED_OK` should probably stay to turn disable the unaligned code for when we know that the `memcpy` calls will not convert to unaligned reads/writes. Like perhaps on big endian architectures..

Dead2:
> I think `UNALIGNED_OK` should probably stay to turn disable the unaligned code for when we know that the `memcpy` calls will not convert to unaligned reads/writes. Like perhaps on big endian architectures..

I think so too, but is it the best naming for our current usage? It might be that it is, I am just making sure to air this before we forget.

mtl1979:
I think the naming is still the best possible... Unaligned access is still undefined behaviour in C standard and relying on compilers and processors to behave the same way is just asking for trouble... Besides big-endian processors, some older ARM processors also don't like unaligned access and we do support all old and new ARM processors that can handle full width of vector instructions in hardware instead of breaking instructions to two half-width instructions and combining the results.

PR #1218, opened 2022-03-31
Remove support for building fuzzers from configure.

nmoinvaz:
Requires https://github.com/google/oss-fuzz/pull/7467 to be merged in first.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -154,7 +154,7 @@ PIC_OBJS = $(PIC_OBJC)
 
 all: static shared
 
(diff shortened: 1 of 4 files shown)


nmoinvaz:
Ready for review.

Dead2 (approved):

nmoinvaz:
@Dead2 should I do another PR to remove sanitizers from configure script as well? It is not well supported and it doesn't do all the compiler flag checks that cmake does. Either way, let me know what you think.

Dead2:
@nmoinvaz I would be good with that, yes. 👍

Issue #1219, opened 2022-04-01
Regression: abicheck.sh fails with strange compilation error when building zlib-ng


dankegel:
Symptom:
```
$ sh test/abicheck.sh --refresh-if
...
/usr/bin/ld: zutil.lo: warning: relocation against `zng_z_errmsg' in read-only section `.text'
/usr/bin/ld: adler32.lo: relocation R_X86_64_PC32 against symbol `functable' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: bad value
collect2: error: ld returned 1 exit status
make: *** [Makefile:268: libz-ng.so.2.0.0] Error 1
```

This appears to be because exporting CHOST causes zlib-ng's configure script to set CROSS_PREFIX, and then the test
```
if test "$gcc" -eq 1 && ($cc $CFLAGS -c $test.c) >> configure.log 2>&1; then
```
fails because the host compiler isn't available with a cross name, leading to a whole big block of code getting skipped, including the code that adds -fPIC.

Looks like this is a regression introduced by https://github.com/zlib-ng/zlib-ng/commit/a1a807b814e9b1990dda4f80ebc73896c5410acd

nmoinvaz:
I think this might be convered in the wiki. Maybe use different linker..

mtl1979:
This is a bug in zlib's configure; I've already reported it.

dankegel:
Well.... maybe not quite.  I am describing a zlib-ng-only problem, not a zlib problem, so this is not related to https://github.com/madler/zlib/issues/617, though the symptom and causes are very similar to that.

See the fix I propose in #1220.

It seems that running zlib-ng's configure with CHOST set is a user error if the self-cross is not installed, so abicheck.sh shouldn't do it.

mtl1979:
In my Linux, both /usr/bin/cc and /usr/bin/gcc are symbolic links to non-canonicalized name, for example /usr/bin/x86_64-linux-gnu-gcc-11, so using CHOST works for both zlib-ng and zlib. However zlib 1.2.12 sets "uname" to "gnu" due to bug in PCRE used to parse CHOST, so CC needs to be explicitly set.

dankegel:
I was using a freshly installed ubuntu 20.04 beta, with compiler installed via "apt install build-essential".

Which Linux are you using?

mtl1979:
[...]

Issue #1217, opened 2022-03-30
Zlib 1.2.12 released


abcfy2:
Zlib 1.2.12 released. Would you plan to be compatible with this version ?

Thanks.

nmoinvaz:
We already have most of the commits that are in zlib 1.2.12. We are missing exporting some crc32 combine functions AFAKI.

mtl1979:
We already have CI action to test if in compatibility mode we have same exported symbols as stock zlib... That test just needs to be updated to use zlib 1.2.12 as baseline and then fix any differences it reports... Even though we diverged from zlib years ago, we still try to include as many commits as possible and if necessary, adapt them to accommodate any overlapping changes specific to zlib-ng.

We didn't want to change the baseline commit before the final version of zlib 1.2.12 was released as that would have required keeping the test up-to-date with each and every commit that adds or changes any exported symbols.

UPDATE: zlib 1.2.12 fails to build shared library, so it's useless for comparing against zlib-ng

mtl1979:
With zlib 1.2.12.1, I get:

```
+ abidiff --no-added-syms --suppressions test/abi/ignore test/abi/zlib-ec3df00224d4b396e2ac6586ab5d25f673caa4c2-x86_64-linux-gnu.abi btmp2/zlib-built.abi
ELF SONAME changed
Functions changes summary: 0 Removed, 0 Changed (24 filtered out), 0 Added (2 filtered out) functions
Variables changes summary: 3 Removed, 0 Changed, 0 Added (1 filtered out) variables

SONAME changed from '' to 'libz.so.1'

3 Removed variables:

  'const char[71] const deflate_copyright'    {deflate_copyright}
  'const char[50] const inflate_copyright'    {inflate_copyright}
  'char* const[10] const z_errmsg'    {z_errmsg}
```

nmoinvaz:
Hidden in this commit are the changes to crc32 functions exports: https://github.com/madler/zlib/commit/21767c654d31d2dccdde4330529775c6c5fd5389

mtl1979:
I don't think we can support both the old and new versions of the changed crc32 related functions as there is no function overloading in standard C.

mtl1979:
This is what I get locally:

```
+ abidiff --no-added-syms --suppressions test/abi/ignore test/abi/zlib-ec3df00224d4b396e2ac6586ab5d25f673caa4c2-x86_64-linux-gnu.abi btmp2/zlib-built.abi
[...]

PR #1220, opened 2022-04-01
Refresh abicheck

dankegel:
Updates reference tags for both zlib and zlib-ng.

Also:
- fix missing --zlib-compat option
- don't assume configure is a sh script
- fix funky CHOST problem described at https://github.com/zlib-ng/zlib-ng/issues/1219; this deserves careful review.

Note that zlib changed the signatures of two functions (cf. discussion in https://github.com/zlib-ng/zlib-ng/issues/1217 )
so this does not quite pass yet.

Best reviewed as individual commits.


dankegel:
Closing in favor of individual pull requests for each fix.

PR #1222, opened 2022-04-02
abicheck.sh: implement --refresh-if option as documented.

dankegel:
Fixes a sad little typo I didn't catch when submitting abicheck.sh initially.
Keeps the misspelled option for compatibility.

### DIFF
--- a/.github/workflows/pkgcheck.yml
+++ b/.github/workflows/pkgcheck.yml
@@ -128,7 +128,7 @@ jobs:
       # macOS runner does not contain abigail
       if: runner.os != 'macOS'
       run: |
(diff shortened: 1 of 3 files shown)


nmoinvaz:
I don't think we need to keep backwards compatibility for this script so if you want to just rename it that would be good too.

dankegel:
kudos on having a good shell checker in CI, it found a no-op assignment!

dankegel:
Should be ready for re-review.

mtl1979 (approved):
LGTM

Dead2 (approved):

PR #1226, opened 2022-04-02
Fixed building tests when -DWITH_GZFILEOP=OFF

nmoinvaz:
Also cmake docs frown upon using `file(GLOB..)` for building srcs.


Dead2 (approved):

mtl1979 (approved):
LGTM... Source file containing main() should be always first.

PR #1227, opened 2022-04-02
abicheck.sh: zlib-ng configure is a bash script, not a sh script

dankegel:
So don't hardcode shell when running configure, it might fail if sh is not bash.

This assumes 'configure' gets checked out with executable permissions, but that
seems to be the case as far back as I checked in git.

### DIFF
--- a/test/abicheck.sh
+++ b/test/abicheck.sh
@@ -121,7 +121,7 @@ then
   git reset --hard FETCH_HEAD
   cd ..
   # Build unstripped, uninstalled, very debug shared library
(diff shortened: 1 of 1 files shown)


Dead2 (approved):

mtl1979 (approved):
LGTM... CodeFactor notice can be fixed in another pull request.

dkegel-fastly:
Thanks!  See https://github.com/zlib-ng/zlib-ng/pull/1222 for the fix for the codefactor notice.

PR #1228, opened 2022-04-02
abicheck.sh: update reference versions

dankegel:
Can't land this until the differences against zlib's new version's function signatures are dealt with somehow.

### DIFF
--- a/test/abi/zlib-ng-2.0.0-x86_64-pc-linux-gnu.abi
+++ b/test/abi/zlib-ng-2.0.0-x86_64-pc-linux-gnu.abi
@@ -1,0 +1,1852 @@
+<abi-corpus version='2.0' path='btmp1/libz-ng.so.2.0.0' architecture='elf-amd-x86_64' soname='libz-ng.so.2'>
+  <elf-needed>
+    <dependency name='libc.so.6'/>
+  </elf-needed>
+  <elf-function-symbols>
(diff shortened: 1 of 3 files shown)


mtl1979:
As Windows (maybe also other non-ELF operating systems) doesn't support symbol versions, in my opinion we should just drop the old versions but in the end it's @Dead2 who makes the final call.

dankegel:
I think this didn't find the new functions mentioned in https://github.com/zlib-ng/zlib-ng/pull/1229, so I should look at that...

nmoinvaz:
Well those functions were already being exported in the .map file.

It was just in the win32 .def files mostly where they weren't being exported.

nmoinvaz:
I'm not sure if the check scripts that you have take into account win32 or not.

Dead2:
I am sorry, I have lost track a bit, but afaik this should pass tests after a rebase, right?

nmoinvaz:
I think we are going to want this for #1268.

mtl1979:
@nmoinvaz Kind of chicken and egg problem... Would need to be sure that merging both separately results in no CI failures...

nmoinvaz:
@dankegel are you able to rebase this? Thanks.

nmoinvaz:
Requires #1268.

Dead2:
I guess #1163 should also go in before this.

nmoinvaz:
Merged in as #1276.

PR #1185, opened 2022-03-12
Reinline crc

KungFuJesus:
Bring back some of the performance benefits of the reorganized inflate window.

PR #1233, opened 2022-04-04
Disable LTO in CMakeLists.txt

phprus:
Forced disable of LTO by CMake.
Required if the parent project uses ``set(CMAKE_INTERPROCEDURAL_OPTIMIZATION ON)``

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -249,6 +249,14 @@ else()
     endif()
 endif()
 
+# Disable LTO
+if(NOT WITH_NATIVE_INSTRUCTIONS)
+    set(CMAKE_INTERPROCEDURAL_OPTIMIZATION OFF)
+    foreach(_cfg_name IN LISTS CMAKE_CONFIGURATION_TYPES)
+        set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_${_cfg_name} OFF)
+    endforeach()
+endif()
+
 # Set architecture alignment requirements
 if(NOT WITH_UNALIGNED)
     add_definitions(-DNO_UNALIGNED)


KungFuJesus:
So, LTO does work with this project and there are shims in place to specifically disable it for things that compile with architecture specific vectorizations (to prevent IPO from say, using a block from one function in another that has possibly incompatible instructions).  Does this PR suggest that's not enough?  

Though in my more recent experiments LTO has hurt more than helped performance on zlib-ng.

phprus:
I don't really know.
The PR was maked based on the information from the comment https://github.com/zlib-ng/zlib-ng/pull/1185#issuecomment-1068973228

> You mention -flto, I hope you are aware that it is incompatible with zlib-ng unless compiling with Native enabled
> (Because it requires all compilation units to be compiled with the same compiler flags, and that breaks -march, -mavx and similar).

cc @Dead2

mtl1979:
@KungFuJesus As far as I know, this pull request only makes sense for cmake 3.9 or later... Currently we support older cmake versions than that...

phprus:
@mtl1979, Yes, but for cmake < 3.9 this code is safe.
For cmake >= 3.9 the code protects against CMAKE_INTERPROCEDURAL_OPTIMIZATION enabled

mtl1979:
> @mtl1979, Yes, but for cmake < 3.9 this code is safe. For cmake >= 3.9 the code protects against CMAKE_INTERPROCEDURAL_OPTIMIZATION enabled

I'm not sure which order "-fno-lto" and "-flto" would appear in the command line if conflicting definitions would happen, but I do agree it's better to be safe than sorry...

nmoinvaz:
Looks okay to me. It makes sense that we wouldn't want to allow LTO in zlib-ng if included in parent project that does allow LTO.

Dead2 (approved):

sfc-gh-sgiesecke:
> So, LTO does work with this project and there are shims in place to specifically disable it for things that compile with architecture specific vectorizations (to prevent IPO from say, using a block from one function in another that has possibly incompatible instructions). Does this PR suggest that's not enough?

Yes, I am also confused why this would be necessary, and what the consequences were if this were removed.

> Because it requires all compilation units to be compiled with the same compiler flags, and that breaks -march, -mavx and similar

Is that so? For which compilers, is there a reference for that?

> Though in my more recent experiments LTO has hurt more than helped performance on zlib-ng.

That's a different question, and probably depends on the specific toolchain / compiler version.

PR #1237, opened 2022-04-06
Fix the custom PNG image based benchmark

KungFuJesus:
The height parameter was using a fixed macro, written at a time when the
test imagery was fully synthetic.  Because of this, images smaller than
than our in-memory generated imagery will artificially throw a CRC
error.


nmoinvaz (approved):
LGTM

Issue #1232, opened 2022-04-04
Tests failed if zlib-ng source directory is readonly


phprus:
zlib-ng 2.0.6

```
phprus@mbp zlib-ng % ctest --rerun-failed --output-on-failure
Test project /.../build/target_osx_arm64/zlib-ng
    Start 5: makefixed
1/4 Test #5: makefixed ........................***Failed    0.02 sec
CMake Error: Could not open file for write in copy operation /.../3rdparty/zlib-ng/src/inffixed_tbl.h.cmp.tmp
CMake Error: : System Error: Inappropriate ioctl for device
CMake Error at /.../3rdparty/zlib-ng/src/cmake/run-and-compare.cmake:44 (configure_file):
  configure_file Problem configuring file



    Start 6: maketrees
2/4 Test #6: maketrees ........................***Failed    0.01 sec
CMake Error: Could not open file for write in copy operation /.../3rdparty/zlib-ng/src/trees_tbl.h.cmp.tmp
CMake Error: : System Error: Inappropriate ioctl for device
CMake Error at /.../3rdparty/zlib-ng/src/cmake/run-and-compare.cmake:44 (configure_file):
  configure_file Problem configuring file



    Start 7: makecrct-crc32
3/4 Test #7: makecrct-crc32 ...................***Failed    0.01 sec
CMake Error: Could not open file for write in copy operation /.../3rdparty/zlib-ng/src/crc32_tbl.h.cmp.tmp
CMake Error: : System Error: Inappropriate ioctl for device
CMake Error at /.../3rdparty/zlib-ng/src/cmake/run-and-compare.cmake:44 (configure_file):
  configure_file Problem configuring file



    Start 8: makecrct-crc32-combine
4/4 Test #8: makecrct-crc32-combine ...........***Failed    0.01 sec
CMake Error: Could not open file for write in copy operation /.../3rdparty/zlib-ng/src/crc32_comb_tbl.h.cmp.tmp
CMake Error: : System Error: Inappropriate ioctl for device
CMake Error at /.../3rdparty/zlib-ng/src/cmake/run-and-compare.cmake:44 (configure_file):
  configure_file Problem configuring file




0% tests passed, 4 tests failed out of 4

Total Test time (real) =   0.05 sec

The following tests FAILED:
      5 - makefixed (Failed)
      6 - maketrees (Failed)
      7 - makecrct-crc32 (Failed)
      8 - makecrct-crc32-combine (Failed)
Errors while running CTest
[...]

PR #1241, opened 2022-04-09
Fixed failed tools tests when source directory is read-only.

nmoinvaz:
See #1232.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1209,7 +1209,7 @@ if(ZLIB_ENABLE_TESTS)
     add_test(NAME makefixed
         COMMAND ${CMAKE_COMMAND}
         "-DCOMMAND=${MAKEFIXED_COMMAND}"
-        -DOUTPUT=${CMAKE_CURRENT_BINARY_DIR}/inffixed_tbl._h
+        -DOUTPUT=${CMAKE_CURRENT_BINARY_DIR}/Testing/Temporary/inffixed_tbl._h
         -DCOMPARE=${CMAKE_CURRENT_SOURCE_DIR}/inffixed_tbl.h
         -DIGNORE_LINE_ENDINGS=ON
         -P ${CMAKE_CURRENT_SOURCE_DIR}/cmake/run-and-compare.cmake)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -1209,7 +1209,7 @@ if(ZLIB_ENABLE_TESTS)
     add_test(NAME makefixed
         COMMAND ${CMAKE_COMMAND}
         "-DCOMMAND=${MAKEFIXED_COMMAND}"
-        -DOUTPUT=${CMAKE_CURRENT_BINARY_DIR}/inffixed_tbl._h
+        -DOUTPUT=${CMAKE_CURRENT_BINARY_DIR}/Testing/Temporary/inffixed_tbl._h
```
mtl1979, line 1212: This doesn't make sense at all... `CMAKE_CURRENT_BINARY_DIR` is never supposed to be read-only.

nmoinvaz: Good point, perhaps if `CMAKE_CURRENT_BINARY_DIR==CMAKE_CURRENT_SOURCE_DIR`?

mtl1979:
See https://github.com/zlib-ng/zlib-ng/blob/develop/cmake/run-and-compare.cmake#L44

When `COMPARE` points to `CMAKE_CURRENT_SOURCE_DIR` and we're doing out-of-tree build, it needs to be replaced with `CMAKE_CURRENT_BINARY_DIR`.

CMakeLists.txt
mtl1979: Obviously we need to check if we are doing out-of-tree build... See my previous comment...

nmoinvaz:
@mtl1979 should be fixed now. Good catch!

mtl1979:
> @mtl1979 should be fixed now. Good catch!

Usually I can see what's going wrong, but I'm not really big fan of learning all the cmake functions because they keep changing them in every version...

phprus:
This PR fixes issue #1232. Thanks!

Dead2 (approved):

PR #1243, opened 2022-04-09
Remove CVE testing from configure script.

nmoinvaz:
These tests already exist in CMake and they are better supported there.


mtl1979:
I'm not sure how much tests we should remove from configure as cmake is still limited to fairly recent operating systems. Even my 32-bit PPC is barely new enough for cmake and I had to install latest available Debian as last Ubuntu was too old.

nmoinvaz:
I don't think we should try and support anything that madler/zlib doesn't support (such as these tests). Configure script support is a maintance burden in my opinion, especially trying to support every feature in both configure and cmake. We don't even support all features in nmake, so there is already precedence.

mtl1979:
@nmoinvaz nmake is nothing like configure or cmake... It's supported because configure scripts can't be run without Cygwin, MinGW or MSYS and cmake support was added in VS2017, but we still support VS2015...

Dead2 (approved):
Since this only removes CVE tests I'll approve. Those are not likely to only fail on older platforms etc.

PR #1236, opened 2022-04-05
Add one extra byte to return value of deflateBound for small lengths.

mtl1979:
Due to right shift by 3 in DEFLATE_QUICK_OVERHEAD() macro, the return value of deflateBound() is 1 too small for lengths smaller than 9.

See #1235.

PR #1244, opened 2022-04-10
deflateBound and compressBound tests for Google Test

nmoinvaz:
Based on #1236 and already existing deflateBound test code we had in there.

### DIFF
--- a/test/test_compress_bound.cc
+++ b/test/test_compress_bound.cc
@@ -1,0 +1,59 @@
+/* test_compress_bound.cc - Test compressBound() with small buffers */
+
+#include "zbuild.h"
+#ifdef ZLIB_COMPAT
+#  include "zlib.h"
+#else
+#  include "zlib-ng.h"
+#endif
+
+#include <stdio.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "test_shared.h"
+
+#include <gtest/gtest.h>
+
+#define MAX_LENGTH (32)
+
+class compress_bound_variant : public testing::TestWithParam<z_size_t> {
+public:
+    void estimate(z_size_t level) {
+        z_size_t estimate_len = 0;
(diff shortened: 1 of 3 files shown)


mtl1979:
Some tests are expected to fail until the pull request for fixing the calculation is merged in. As I already said in the original issue, I'm all in for improving testing for both cmake and configure...

mtl1979:
@Dead2 Should be merged along with #1236... I have no personal preference on which one gets merged first. Due to requirement for cmake 3.12 or newer for Google Test framework (and benchmarks), the tests are not mutually exclusive, as quite many Linux systems are still limited to cmake 3.10 or older (For example, my own Ubuntu only has 3.10.2).

test/test_compress_bound.cc
```diff
@@ -0,0 +1,60 @@
...
+        for (int j = 0; j < MAX_LENGTH; j++) {
+            uncompressed[j] = (uint8_t)j;
+        }
+
+        for (int32_t i = 0; i < MAX_LENGTH; i++) {
+            /* calculate actual output length */
+            estimate_len = PREFIX(compressBound)(i);
+
+            if (dest != NULL) {
+                z_size_t dest_len;
```
mtl1979: `dest_len` should be initialized to output buffer length, otherwise compress2() will not know how many bytes it can write to output buffer before it overflows. Returned value is how many bytes was actually used, so the parameter is read/write.

nmoinvaz:
Rebased.

Dead2:
This seems to be failing all over the place. Problem with the test or the fix?

mtl1979:
@Dead2 I found one issue in the test itself that seems still to be unfixed...

test/test_compress_bound.cc
nmoinvaz: Fixed.

nmoinvaz:
Rebased and issue @mtl1979 mentioned is fixed.

PR #1245, opened 2022-04-10
Improved chunkset substantially where it's heavily used

KungFuJesus:
For most realistic use cases, this doesn't make a ton of difference.
However, for things which are highly compressible and enjoy very large
run length encodes in the window, this is a huge win.

We leverage a permutation table to swizzle the contents of the memory
chunk into a vector register and then splat that over memory with a fast
copy loop.

In essence, where this helps, it helps a lot.  Where it doesn't, it does
no measurable damage to the runtime.

This commit also simplifies a chunkcopy_safe call for determining a
distance.  Using labs is enough to give the same behavior as before,
with the added benefit that no predication is required _and_, most
importantly, static analysis by GCC's string fortification can't throw a
fit because it conveys better to the compiler that the input into
builtin_memcpy will always be in range.


KungFuJesus:
```
Comparing png_decode/ (from ./benchmark_zlib_apps_oldchunk) to png_decode/ (from ./benchmark_zlib_apps)
Benchmark                                                                            Time             CPU      Time Old      Time New       CPU Old       CPU New
-----------------------------------------------------------------------------------------------------------------------------------------------------------------
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]0_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]0_mean                  -0.0363         -0.0363           719           693           719           693
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]0_median                -0.0381         -0.0381           719           692           719           692
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]0_stddev                +0.6699         +0.6695             4             6             4             6
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]0_cv                    +0.7327         +0.7323             0             0             0             0
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]1_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]1_mean                  +0.0090         +0.0089           551           556           551           556
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]1_median                +0.0077         +0.0077           551           555           551           555
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]1_stddev                -0.2292         -0.2302             2             1             2             1
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]1_cv                    -0.2360         -0.2370             0             0             0             0
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]2_pvalue                 0.0008          0.0008      U Test, Repetitions: 10 vs 10
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]2_mean                  +0.0094         +0.0094           571           577           571           577
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]2_median                +0.0096         +0.0096           571           577           571           577
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]2_stddev                -0.3157         -0.3158             3             2             3             2
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]2_cv                    -0.3221         -0.3221             0             0             0             0
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]3_pvalue                 0.0008          0.0008      U Test, Repetitions: 10 vs 10
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]3_mean                  -0.0101         -0.0101           575           569           575           569
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]3_median                -0.0113         -0.0113           576           570           576           570
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]3_stddev                -0.3997         -0.3994             3             2             3             2
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]3_cv                    -0.3936         -0.3933             0             0             0             0
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]4_pvalue                 0.1212          0.1212      U Test, Repetitions: 10 vs 10
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]4_mean                  +0.0011         +0.0011           542           543           542           543
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]4_median                +0.0017         +0.0017           542           543           542           543
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]4_stddev                -0.2119         -0.2120             2             2             2             2
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]4_cv                    -0.2128         -0.2129             0             0             0             0
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]5_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]5_mean                  -0.0269         -0.0269           582           567           582           567
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]5_median                -0.0274         -0.0274           582           566           582           566
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]5_stddev                +0.0730         +0.0728             2             2             2             2
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]5_cv                    +0.1027         +0.1025             0             0             0             0
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]6_pvalue                 0.7337          0.7337      U Test, Repetitions: 10 vs 10
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]6_mean                  +0.0017         +0.0017           555           556           555           556
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]6_median                +0.0001         +0.0001           555           555           555           555
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]6_stddev                -0.3152         -0.3150             3             2             3             2
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]6_cv                    -0.3164         -0.3162             0             0             0             0
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]7_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]7_mean                  -0.2925         -0.2925           948           671           948           671
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]7_median                -0.2937         -0.2937           949           670           949           670
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]7_stddev                -0.6430         -0.6426             4             1             4             1
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]7_cv                    -0.4955         -0.4949             0             0             0             0
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]8_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]8_mean                  -0.2832         -0.2832           902           647           902           647
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]8_median                -0.2841         -0.2841           903           646           903           646
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]8_stddev                -0.3166         -0.3172             2             1             2             1
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]8_cv                    -0.0466         -0.0474             0             0             0             0
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]9_pvalue                 0.0002          0.0002      U Test, Repetitions: 10 vs 10
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]9_mean                  -0.2830         -0.2830           902           647           902           647
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]9_median                -0.2841         -0.2841           903           646           903           646
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]9_stddev                -0.3868         -0.3869             2             1             2             1
[png_decode/ vs. png_decode/][png_decode/ vs. png_decode/]9_cv                    -0.1448         -0.1448             0             0             0             0
OVERALL_GEOMEAN                                                                   -0.1010         -0.1010             0             0             0             0
```

nmoinvaz:
Very interesting code! 

Benchmarks show better performance on higher compression levels which confirms what you said about the benefits with regards to compressibility. If I'm reading it correctly it is 30% improvement?

Total table size is about 1k. It's not too bad. And platforms that don't use it, won't include it so that's good.

arch/x86/chunk_permute_table.h
```diff
@@ -0,0 +1,53 @@
...
+    16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 0, 1, 2, 3, 4, 5, /* dist 26 */
+    16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 0, 1, 2, 3, 4, /* dist 27 */
+    16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 0, 1, 2, 3, /* dist 28 */
+    16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 0, 1, 2, /* dist 29 */
+    16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 0, 1, /* dist 30 */
+    16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 0, /* dist 31 */
+};
+
+typedef struct lut_rem_pair_s {
+    uint32_t idx;
```
nmoinvaz: Does this need to be uint_32? Seems like the first parameter can be 16 bit size and the second 8 bit size.

inflate_p.h
```diff
@@ -5,6 +5,8 @@
 #ifndef INFLATE_P_H
 #define INFLATE_P_H
 
+#include "functable.h"
```
nmoinvaz: What's the include here needed for?

inflate_p.h
```diff
@@ -146,7 +148,7 @@ static inline uint8_t* chunkcopy_safe(uint8_t *out, uint8_t *from, size_t len, u
      * initial bulk memcpy of the nonoverlapping region. Then, we can leverage the size of this to determine the safest
      * atomic memcpy size we can pick such that we have non-overlapping regions. This effectively becomes a safe look
      * behind or lookahead distance */
-    size_t non_olap_size = ((from > out) ? from - out : out - from);
+    size_t non_olap_size = labs(from - out);
```
nmoinvaz: I'm not seeing where labs is defined.

arch/x86/chunkset_sse41.c
```diff
@@ -0,0 +1,89 @@
...
+
+#define HAVE_CHUNKMEMSET_1
+#define HAVE_CHUNKMEMSET_2
+#define HAVE_CHUNKMEMSET_4
+#define HAVE_CHUNKMEMSET_8
+#define HAVE_CHUNK_MAG
+
+static const lut_rem_pair perm_idx_lut[29] = {
+    {0, 1},      /* 3 */
+    {0, 0}, /* don't care */
```
nmoinvaz: Maybe comments should be aligned.

arch/x86/chunkset_sse41.c
```diff
@@ -0,0 +1,89 @@
+/* chunkset_sse2.c -- SSE2 inline functions to copy small data chunks.
```
nmoinvaz: Need to fix filename?

arch/x86/chunkset_sse2.c
```diff
@@ -16,6 +16,22 @@ typedef __m128i chunk_t;
 #define HAVE_CHUNKMEMSET_4
 #define HAVE_CHUNKMEMSET_8
 
+#if 0
```
nmoinvaz: What's this table for? Maybe a comment is needed if we want to keep it in.

inflate_p.h
KungFuJesus: It's in math.h I think, it's just the long version of abs

inflate_p.h
KungFuJesus: It's isn't, I had been messing with dispatching to chunkmemset_safe et al but didn't see a measurable improvement.

inflate_p.h
mtl1979: In my Linux, it's in stdlib.h.

arch/x86/chunk_permute_table.h
KungFuJesus: It could I guess, I'm wondering what that really does for efficiency though. The compression it buys you is small but also gcc likes to perform a keyhole optimization where adjacent 32 bit loads are performed with a single 64 bit load and shifted and and'd out. I didn't specifically check if this was happening but I suspect it might be. Also on some architectures this might end up being padded out, anyway.

nmoinvaz: I was thinking for code size not necessarily for speed. Even if both were uint16. Then it would be padded already.

KungFuJesus:
The msan error might be something, I'll have to look at it maybe later tonight. It could also just be complaining about loading a full vector register's width of data and then swizzling over it with the shuffle instruction. The initial load touches memory that could have been uninitialized.

It's funny because I had an if tree of minimal dist sizes where I'd do a cvtsi(x)_si128 but the compiler was just translating that to full vector loads, anyway.

I also at one point was checking for minimum lengths before loading up a magazine there but realized a lot of the code is written in such a way that it's allowed to assume at least one vector register past the bounds.

I'll have to see if I can use msan + the debugger to find the exact case where it's happening and find a sensible way around it.

inflate_p.h
nmoinvaz: > it's just the long version of abs

Ah, it the name just looks a little weird that is why I didn't recognize it. 😁

arch/x86/chunk_permute_table.h
KungFuJesus:
```   15f50:       83 ee 03                sub    $0x3,%esi
   15f53:       48 8d 05 86 ca 00 00    lea    0xca86(%rip),%rax        # 229e0 <perm_idx_lut>
   15f5a:       8b 4c f0 04             mov    0x4(%rax,%rsi,8),%ecx
   15f5e:       8b 04 f0                mov    (%rax,%rsi,8),%eax
   15f61:       48 8d 35 78 cb 00 00    lea    0xcb78(%rip),%rsi        # 22ae0 <permute_table>
```
Doesn't look like GCC is applying that optimization, anyway.  Yeah if they're both uint16's, it shouldn't apply any padding to the list of structs.  I'm not seeing any real compelling reasons not to make it half the size.

arch/x86/chunkset_sse41.c
nmoinvaz: Looks like maybe you may have forgotten this one. Or aligned a different table..

arch/x86/chunkset_avx.c
```diff
@@ -5,6 +5,14 @@
 
 #ifdef X86_AVX_CHUNKSET
 #include <immintrin.h>
+#include "chunk_permute_table.h"
+
+#if defined(__has_feature)
+#  if __has_feature(memory_sanitizer)
+#    define MEMORY_SANITIZER 1
+#include<sanitizer/msan_interface.h>
```
nmoinvaz: Missing a space after `include`.

arch/x86/chunkset_avx.c
```diff
@@ -45,6 +90,42 @@ static inline void storechunk(uint8_t *out, chunk_t *chunk) {
...
+    __m256i perm_vec, ret_vec;
+    /* While technically we only need to read 4 or 8 bytes into this vector register
+     * for a lot of cases, GCC is compiling this to a shared load for all branches,
+     * preferring the simpler code.  Given that the buf value isn't in GPRs to begin with,
+     * the 256 bit load is _probably_ just as inexpensive */
+    *chunk_rem = lut_rem.remval;
+
+#ifdef MEMORY_SANITIZER
+    /* See note in chunkset_sse4.c for why this is ok */
+    __msan_unpoison(buf+dist, 32 - dist);
```
nmoinvaz: This is interesting. There is another place where this could be used.
- https://github.com/zlib-ng/zlib-ng/blob/e43e0b21893b55d6b44d3f8b7105fcb655e87f5a/inflate.c#L208

nmoinvaz: Actually, I thought it was zeroing the whole window. Maybe not worth it.

nmoinvaz: It does though, seem like the exact same situation.

KungFuJesus: Maybe.  Is that being initialized to 0 just to prevent msan from flagging it?  In this case we're unpoisoning the memory because we need to do the full load but we know we're only using the first n bytes of it after the shuffle instruction.

nmoinvaz: Yeah I think it is the same. I have put together a commit https://github.com/nmoinvaz/zlib-ng/commit/80e12c06b3e204b335241455d85159d8330295e9.

KungFuJesus: Ah yeah, definitely basically the same exact use case.  And this way, you don't have write to that memory when not being built with sanitizers :).  Though, it'd be better to only unpoison it just before the read you know is doing this rather than just doing it preemptively.  That way you catch if you unintentionally read that part of hte window when you don't expect it.

nmoinvaz: > Though, it'd be better to only unpoison it just before the read you know is doing this rather than just doing it preemptively

That would be better however it would be a lot more tricky.

nmoinvaz:
Now that #1247 has been merged, you can update the code for zbuild.h that now has `Z_MEMORY_SANITIZER`. It can be used and you don't have to include sanitizers/msan.h cause it already included it for you.

I previously posted this in the wrong PR. This is the right PR.

arch/x86/chunkset_avx.c
```diff
@@ -45,6 +83,51 @@ static inline void storechunk(uint8_t *out, chunk_t *chunk) {
...
+     * the 256 bit load is _probably_ just as inexpensive */
+    *chunk_rem = lut_rem.remval;
+
+#ifdef Z_MEMORY_SANITIZER
+    /* See note in chunkset_sse4.c for why this is ok */
+    __msan_unpoison(buf+dist, 32 - dist);
+#endif
+
+    if (dist < 16) {
+        /* This simpler case still requires us to shuffle in a 128 bit lanes, so we must
```
mtl1979: Mixing singular and plural in `in a` and `lanes`.

arch/x86/chunkset_avx.c
```diff
@@ -45,6 +83,51 @@ static inline void storechunk(uint8_t *out, chunk_t *chunk) {
...
+         * apply a static offset after broadcasting the first vector register to both
```
mtl1979: Short line with sentence spanning to next line... We use 120 characters per line.

arch/x86/chunkset_avx.c
```diff
@@ -14,6 +15,43 @@ typedef __m256i chunk_t;
...
+    {3 * 32, 4},            /* 7 */
+    {0 * 32, 0},            /* don't care */
+    {4 * 32, 5},            /* 9 */
+    {5 * 32, 22},           /* 10 */
+    {6 * 32, 21},           /* 11 */
+    {7 * 32, 20},           /* 12 */
+    {8 * 32, 16},           /* 13 */
+    {9 * 32, 4},            /* 14 */
+    {10 * 32, 2},           /* 15 */
+    {0 * 32 , 0},           /* don't care */
```
mtl1979: Space before comma looks weird... Maybe should use space before `0` instead and use ` 0 * 32` consistently to make second item of array element align better... Also add space before `1` ... `9` when the sequence goes higher than `9`.

KungFuJesus:
I actually made this marginally faster (~4-5%) in our benchmarks.  I had been conditionally switching on the distance length as to whether or not to call the lesser variant when in reality I should have switched based on len.  When len <= 16, it does us no favors to use the avx2 version.  When it's >, we can still benefit from the wider registers, we just need to be more careful on handling the shuffles.  I was confusingly missing the case for when dist == 16, which of course, there's no proper table entry for and we simply want to load and broadcast the 128bit load to a 256 bit load.  That's what those last 4 or so pushes were about.

Also weirder, in regular Intel fashion, the vbroadcasti128 instruction is slower than combining the same register with itself with a vinserti128 instruction.  Intel being Intel, I guess, but it seems like a waste of the encoding space if you ask me.  Similar to how the SSE41 dot product instruction is 100% worse all of the time.

KungFuJesus:
So, I'm trying to reproduce those errors but I'm not seeing them.  I don't have clang 11 but clang 12 is giving me nothing with debug on.  Where is that test data coming from?

KungFuJesus:
Nevermind, I figured it out.  One of the remainders in the table had been wrong (a 1 slipped in front of the 6).  Though, that should have been obvious to me looking at the table.

chunkset_tpl.h
```diff
@@ -5,6 +5,10 @@
 #include "zbuild.h"
 #include <stdlib.h>
 
+#if CHUNK_SIZE == 32 && defined(X86_SSE41) && defined(X86_SSE2)
```
nmoinvaz, line 8: Instead of doing this, maybe just set a preprocessor definition in `chunkset_sse41.c`? 
```c
#if CHUNK_SIZE == 32 && defined(X86_SSE41) && defined(X86_SSE2)
#  define CHUNKSET_QUICKSET \
    if (len <= 16) { \
        return chunkmemset_sse41(out, dist, len); \
    }
#else
#  define CHUNKSET_QUICKSET
#endif
```
Then just call
`CHUNKSET_QUICKSET` in `chunkset_tpl.h`. Or something similar?

nmoinvaz: Then you don't really need that extern I would think.

KungFuJesus: I would still need the extern in the chunkset_avx.c file, though.  And every version of chunkset_*.c would then need to have CHUNKSET_QUICKSET defined.

KungFuJesus:
Rebased

arch/x86/chunkset_avx.c
```diff
@@ -14,6 +15,43 @@ typedef __m256i chunk_t;
...
+    {11 * 32, 15},          /* 17 */
+    {11 * 32 + 16, 14},     /* 18 */
+    {11 * 32 + 16 * 2, 13}, /* 19 */
+    {11 * 32 + 16 * 3, 12}, /* 20 */
+    {11 * 32 + 16 * 4, 11}, /* 21 */
+    {11 * 32 + 16 * 5, 10}, /* 22 */
+    {11 * 32 + 16 * 6, 9},  /* 23 */
+    {11 * 32 + 16 * 7, 8},  /* 24 */
+    {11 * 32 + 16 * 8, 7},  /* 25 */
+    {11 * 32 + 16 * 9, 6},  /* 26 */
```
mtl1979: Pad 6-9 in second array element, so it aligns with lines above and below.

arch/x86/chunkset_avx.c
```diff
@@ -45,6 +83,51 @@ static inline void storechunk(uint8_t *out, chunk_t *chunk) {
...
+static inline chunk_t GET_CHUNK_MAG(uint8_t *buf, uint32_t *chunk_rem, uint32_t dist) {
+    lut_rem_pair lut_rem = perm_idx_lut[dist - 3];
+    __m256i ret_vec;
+    /* While technically we only need to read 4 or 8 bytes into this vector register for a lot of cases, GCC is compiling this
+     * to a shared load for all branches, preferring the simpler code.  Given that the buf value isn't in GPRs to begin with,
+     * the 256 bit load is _probably_ just as inexpensive */
+    *chunk_rem = lut_rem.remval;
+
+#ifdef Z_MEMORY_SANITIZER
+    /* See note in chunkset_sse4.c for why this is ok */
```
mtl1979, line 88: I'm not sure this comment is useful, people usually don't view two files in parallel.

arch/x86/chunkset_avx.c
```diff
@@ -45,6 +83,51 @@ static inline void storechunk(uint8_t *out, chunk_t *chunk) {
...
+        perm_vec = _mm256_add_epi8(perm_vec, permute_xform);
+        ret_vec = _mm256_inserti128_si256(_mm256_castsi128_si256(ret_vec0), ret_vec0, 1);
+        ret_vec = _mm256_shuffle_epi8(ret_vec, perm_vec);
+    } else if (dist == 16) {
+        __m128i ret_vec0 = _mm_loadu_si128((__m128i*)buf);
+        return _mm256_inserti128_si256(_mm256_castsi128_si256(ret_vec0), ret_vec0, 1);
+    } else {
+        __m128i ret_vec0 = _mm_loadu_si128((__m128i*)buf);
+        __m128i ret_vec1 = _mm_loadu_si128((__m128i*)(buf + 16));
+        /* Take advantage of the fact that only the latter half of the 256 bit vector
```
mtl1979: Short line.

arch/x86/chunkset_avx.c
```diff
@@ -45,6 +83,51 @@ static inline void storechunk(uint8_t *out, chunk_t *chunk) {
...
+         * will actually differ */
+        __m128i perm_vec1 = _mm_load_si128((__m128i*)(permute_table + lut_rem.idx));
+        __m128i xlane_permutes = _mm_cmpgt_epi8(_mm_set1_epi8(16), perm_vec1);
+        __m128i xlane_res  = _mm_shuffle_epi8(ret_vec0, perm_vec1);
+        /* Since we can't wrap twice, we can simply keep the later half exactly how
```
mtl1979: Short line.

arch/x86/chunkset_sse41.c
```diff
@@ -0,0 +1,103 @@
+/* chunkset_sse41.c -- SSE4 inline functions to copy small data chunks.
+ * For conditions of distribution and use, see copyright notice in zlib.h
+ */
+
+#include "zbuild.h"
+
+/* This requires SSE2 support. While it's implicit with SSE4, we can minimize
+ * code size by sharing the chunkcopy functions, which will certainly compile
+ * to identical machine code */
+#if defined(X86_SSE41) && defined(X86_SSE2)
```
mtl1979, line 10: This doesn't really make sense... We can share the functions, but we don't need to test for `X86_SSE2`here... If user wants to disable SSE2 support, but not SSE41 support, then this will break horribly.

functable.c
```diff
@@ -315,6 +315,10 @@ Z_INTERNAL uint8_t* chunkmemset_stub(uint8_t *out, unsigned dist, unsigned len)
 # endif
         functable.chunkmemset = &chunkmemset_sse2;
 #endif
+#if defined(X86_SSE41) && defined(X86_SSE2)
```
mtl1979, line 318: This is not good, see above for why...

functable.c
```diff
@@ -342,6 +346,10 @@ Z_INTERNAL uint8_t* chunkmemset_safe_stub(uint8_t *out, unsigned dist, unsigned
 # endif
         functable.chunkmemset_safe = &chunkmemset_safe_sse2;
 #endif
+#if defined(X86_SSE41) && defined(X86_SSE2)
```
mtl1979, line 349: Not sure this is the best way to handle these...

arch/x86/chunkset_sse41.c
KungFuJesus: I don't know about horribly, it just won't compile the functions, which is exactly what we want.  We're relying on SSE2 support to call the chunkcopy_sse variants of the functions instead of declaring the same functions as SSE4 ones, which will undoubtedly compile to the same exact code and just make the binary larger.

The worst effect one would experience from this is only the C based chunkcopy, set, and other various chunk routines would be called.  That is, if somebody really wants to go out of their way to turn off SSE2 support, anyway.

arch/x86/chunkset_avx.c
KungFuJesus: What are we calling the maximum width here?  This comment is well over 128 characters so it has to wrap at some point.

arch/x86/chunkset_avx.c
KungFuJesus: I'd prefer not to verbatim copy and paste that lengthy comment.  If the user really has questions as to whether that unpoison call is right, I don't think having them open one more file is the end of the world.

inflate_p.h
phprus: This is a wrong change.

The ``from`` and ``out`` are pointers, ``typeof(from - out)`` is ``ptrdiff_t``.

```c
long labs( long n );
```

But, on Win64:
``long`` - 32 bit
pointers, ``size_t`` and ``ptrdiff_t`` - 64 bit

See: https://pvs-studio.com/en/blog/terms/0012/

KungFuJesus: What is the 64 bit abs on Windows if not labs?  Bear in mind we're doing this to get around some string fortification that GCC is unwisely flagging as potentially hazardous when we were doing this math as a ternary statement (it detected an unrealistic bounds from the pointer arithmetic that in practice never happened because the condition of comparing the greater pointer prior to the subtraction meant that the size never went negative).

KungFuJesus: Ah, a little bit of google sleuthing seems to reveal llabs is what we want?  I never understood Microsoft's hack that was ILP32LL - it just seems like they could have written most if not all of their code to used fixed width types if they needed this compatibility this badly.

phprus: ``long long`` is always 64 bit (on all 32 and 64 bit platforms).
```
long long llabs( long long n ); (since C99)
Defined in header <inttypes.h>
```
(https://en.cppreference.com/w/c/numeric/math/abs)

arch/x86/chunkset_sse41.c
KungFuJesus: As far as I can tell, there really isn't a cleaner way to dispatch to the SSE2 functions, here.  You could go the long, roundabout way of dispatching through the function table but that has its own set of issues.  For one, you specifically want to call the shorter vector versions when you're on an architecture that has 256 bit wide versions, otherwise you're jumping twice (which maybe isn't the end of the world, but it'd be nice to avoid).  For two, you're losing a convenience afforded by having the macro named versions for each respective architecture when you want to call one from the other.  There's also the fact that calling into the function table does in fact require a trip to thread local storage, which does mean you have to visit segment registers to find the current address of your function (a small bit of overhead, but these are already routines dealing with relatively small chunks of data).

I think this is the lesser of two evils and it has no side effects as far as I've tested, even toggling all kinds of combinations of supported vectorizations into the build options.

mtl1979: Recent GCC versions can't generate anything older than SSE4, so trying to compile code using just `-msse2` will most likely result in "bad" binary. I've worked with gcc people for over 6 years and some of my bug reports regarding older SSE versions are still open and unsolved. At work, we still use gcc 4.9 and gcc 6.5 as anything newer has issues that cause random crashes or poor performance depending on the CPU the code is run.

If someone wants to disable SSE2 code paths, it should revert to compiling the relevant functions using SSE4.1.

arch/x86/chunkset_avx.c
mtl1979: As close to 120 characters as possible, but not exceeding it.... Sometimes it makes sense to split before related two words, but this line is clearly shorter than other comments in this pull request.

arch/x86/chunkset_sse41.c
KungFuJesus: If someone wants to disable SSE2 code paths, then we have no chunkcopy function, as the SSE4 version uses SSE2 chunkcopy directly, as does the SSE2 version.  So right now, disabling SSE2 code has the same effect.  If you don't have SSE2 support compiled in, you don't have accelerated chunkcopy, chunkmemset, etc (unless you happen to have AVX2).  

The SSE4 enabled paths for these same templates are going to compile to the exact same code, we are only increasing the binary size by doing this.  Right now (as in on develop), if you have an SSE4 enabled CPU and disable SSE2 support, you won't have vectorized chunk routines (exactly the same as in this PR).

KungFuJesus: Here's chunkcopy_sse2 with only sse2 flags at build time:
```asm
0000000000013c40 <chunkcopy_sse2>:
   13c40:       8d 4a ff                lea    -0x1(%rdx),%ecx
   13c43:       f3 0f 6f 0e             movdqu (%rsi),%xmm1
   13c47:       83 e1 0f                and    $0xf,%ecx
   13c4a:       44 8d 41 01             lea    0x1(%rcx),%r8d
   13c4e:       0f 11 0f                movups %xmm1,(%rdi)
   13c51:       4a 8d 04 07             lea    (%rdi,%r8,1),%rax
   13c55:       4c 01 c6                add    %r8,%rsi
   13c58:       44 29 c2                sub    %r8d,%edx
   13c5b:       74 17                   je     13c74 <chunkcopy_sse2+0x34>
   13c5d:       01 c2                   add    %eax,%edx
   13c5f:       90                      nop
   13c60:       f3 0f 6f 06             movdqu (%rsi),%xmm0
   13c64:       48 83 c0 10             add    $0x10,%rax
   13c68:       48 83 c6 10             add    $0x10,%rsi
   13c6c:       0f 11 40 f0             movups %xmm0,-0x10(%rax)
   13c70:       39 c2                   cmp    %eax,%edx
   13c72:       75 ec                   jne    13c60 <chunkcopy_sse2+0x20>
   13c74:       c3                      ret    
   13c75:       66 66 2e 0f 1f 84 00    data16 cs nopw 0x0(%rax,%rax,1)
   13c7c:       00 00 00 00 

```

Here's with -msse4 applied to the global CFLAGS:

```asm
0000000000013c60 <chunkcopy_sse2>:
   13c60:       8d 4a ff                lea    -0x1(%rdx),%ecx
   13c63:       f3 0f 6f 0e             movdqu (%rsi),%xmm1
   13c67:       83 e1 0f                and    $0xf,%ecx
   13c6a:       44 8d 41 01             lea    0x1(%rcx),%r8d
   13c6e:       0f 11 0f                movups %xmm1,(%rdi)
   13c71:       4a 8d 04 07             lea    (%rdi,%r8,1),%rax
   13c75:       4c 01 c6                add    %r8,%rsi
   13c78:       44 29 c2                sub    %r8d,%edx
   13c7b:       74 17                   je     13c94 <chunkcopy_sse2+0x34>
   13c7d:       01 c2                   add    %eax,%edx
   13c7f:       90                      nop
   13c80:       f3 0f 6f 06             movdqu (%rsi),%xmm0
   13c84:       48 83 c0 10             add    $0x10,%rax
   13c88:       48 83 c6 10             add    $0x10,%rsi
   13c8c:       0f 11 40 f0             movups %xmm0,-0x10(%rax)
   13c90:       39 c2                   cmp    %eax,%edx
   13c92:       75 ec                   jne    13c80 <chunkcopy_sse2+0x20>
   13c94:       c3                      ret    
   13c95:       66 66 2e 0f 1f 84 00    data16 cs nopw 0x0(%rax,%rax,1)
   13c9c:       00 00 00 00 
```

The resulting machine code is basically identical.

mtl1979: > If someone wants to disable SSE2 code paths, then we have no chunkcopy function, as the SSE4 version uses SSE2 chunkcopy directly, as does the SSE2 version. 

Obviously that is bad, but this pull request does nothing to make it more acceptable. We should in no case endorse use of incompatible or conflicting build options (except if building with compatibility enabled and zlib gracefully handles the build options). As such testing for `X86_SSE2` too early is unacceptable compromise.

Dead2: Honestly, I don't really see a big problem here.

Having duplicate functions would not be a good solution.

But an alternative that might be cleaner and more easily understood is to compile the required sse2 functions if `X86_SSE2` OR  `X86_SSE41` is defined (instead of only for sse2). That way it will always be available for either flag.
That would pull a subset of SSE2 functions into the SSE41 preference, but I think that only makes sense since there is no benefit in making a new SSE41 version of the specific functions. (Wholesale enabling all SSE2 functions would not make sense though).

Also these defines are not to be confused with the `-msse2` etc compiler flags, since those are very separate from this config-preference define.

mtl1979: That `OR` is basically same as what I suggested. The real function gets compiled from `chunkset_sse2.c` if SSE2 is enabled, and from `chunkset_sse41.c` if SSE2 is disabled (for example to reduce binary size). This makes sure we always have optimized version if at least one vector instruction set is both enabled and supported.

KungFuJesus: The way the templates are written that is impossible without a whole lot of nasty preprocessor churn to conditionally compile an SSE4 version if there is no sse2 version. It'd be ifdef spaghetti all for the same net effect.

SSE2 is a subset of SSE4 and it's actually a requirement now for 64 bit x86 processors. The odds of somebody trying to disable this on x86 are slim to none and misguided at best if they are optimization efforts. We've also concluded that there's no benefit to compiling the chunkcopy_* family of functions with sse4 specific optimizations, so the argument for writing this duplicate function is nil, since they compile to the same thing. Writing a preprocessor web to ensure they don't coexist is probably more trouble than it's worth and essentially by doing so, they would be bringing the function back they set out to disable.

KungFuJesus: Rereading this, @Dead2's suggestion for OR actually isn't as hard a pill to swallow as I was thinking had been suggested. I still contend it's a silly thing for the user to do, though. And, it's somewhat counter intuitive to what they might mean when disabling the WITH_SSE2 option.  

@mtl1979 your suggestion is more or less the ifdef spaghetti I was envisioning, though.  Making sure that both functions don't exist requires quite a bit more boolean logic.  Basically, you either:
 1. Build the SSE2 version if SSE4 isn't enabled, and otherwise build the "sse4" version if and only if the SSE2 flags aren't defined
 2. Or, build the SSE4 version and only build the SSE2 version if WITH_SSE4 isn't enabled

This same set of preprocessor logic has to bubble out to the function table as well.  And around around stub that dispatches to any function named in these templates (chunkcopy_unroll, chunkcopy, chunkcopy_safe, etc).

Dead2: There is some merit to disabling the SSE2 functions if they can all be replaced by better alternatives AND the user knows that this will not run on any older cpus (or the C versions are ok for any stragglers).

This is kind of a gray area in our configuration options though, since it is currently hard to know whether you have all the "best" non-C functions enabled if you want to minimize the code size of the library. I guess it could be done in configure and cmake by having them print out each group of functions (adler, crc32, etc) along with the enabled instruction sets for each of them.

KungFuJesus: The merit for that kind of dwindles when you realize that SSE4 doesn't add anything additional to the mix and doesn't justify creating SSE4 variants of them, though.  The memset function does, obviously, because we're leveraging shuffles and blends there, but all the other functions basically end up compiling to the same thing.  

On the flip side, renaming the copy functions to SSE4 and requiring SSE4 for it to work doesn't feel like a great solution, either.  Then older CPUs end up losing this capability.  Trying to make them mutually exclusive is messy.  We could split up the files as different compilation units but then you lose the benefit where the functions directly call each other (happens in memset and a couple of other routines) rather going through the indirection of the function table.  Not to mention the compiler also is more likely to inline these routines when they are called directly, so that has an added benefit for keeping them together.

mtl1979: All I see is that if you move the test for `X86_SSE2` as late as possible, all the "ugly" magic ends up in single file, not "all over the sources"... I know it means the test has to be done multiple times (with fallback path) in this file for it to work as expected, but it's still better than having the test done in multiple files possibly unnecessarily.

arch/x86/chunkset_sse41.c
```diff
@@ -0,0 +1,103 @@
...
+    {5 * 32, 6}, /* 10 */
+    {6 * 32, 5}, /* 11 */
+    {7 * 32, 4}, /* 12 */
+    {8 * 32, 3}, /* 13 */
+    {9 * 32, 2}, /* 14 */
+    {10 * 32, 1},/* 15 */
+};
+
+
+static inline void chunkmemset_1(uint8_t *from, chunk_t *chunk) {
```
nmoinvaz: We don't need `chunkmemset_1` anymore.

arch/x86/chunkset_sse41.c
```diff
@@ -0,0 +1,103 @@
...
+ * to identical machine code */
+#if defined(X86_SSE41) && defined(X86_SSE2)
+#include <immintrin.h>
+#include "chunk_permute_table.h"
+
+typedef __m128i chunk_t;
+
+#define CHUNK_SIZE 16
+
+#define HAVE_CHUNKMEMSET_1
```
nmoinvaz: We don't need `HAVE_CHUNKMEMSET_1` anymore.

arch/x86/chunkset_sse41.c
KungFuJesus: I say we modify the && to || in chunkset_tpl.h for this macro as suggested by @Dead2.  It doesn't make sense to name the other functions _sse4 as a fallback when sse2 support is not compiled in because the instructions they will generate will merely be comprised of the sse2 extensions with no additional instructions as far as I'm aware.  

The user turning off WITH_SSE2 also has other ill-effects that are plenty reason to motivate not doing it.  The slide hash functions, unless you have a CPU with AVX support, would also be disabled, as well as a myriad of other functions such as the fast compare routines which do not have faster counterparts until AVX support.  As of now, there are also a number of functions that are made faster by having both SSE support and AVX support compiled in (e.g. if you have fewer than 32 elements for an adler checksum and have the SSE functions compiled in, you'll avoid a lot of scalar calculations of that hash).  In benchmarks this was actually found to be quite a bit faster and I imagine it's commonplace to get small-ish windows like that for decompression.  I modified these things such that if they don't exist there's a safe fallback, but, if the user is looking to do this for speed optimization purposes, they're probably approaching it the wrong way.

arch/x86/chunkset_avx.c
KungFuJesus: :+1:

arch/x86/chunkset_sse41.c
```diff
@@ -0,0 +1,98 @@
...
+static inline chunk_t GET_CHUNK_MAG(uint8_t *buf, uint32_t *chunk_rem, uint32_t dist) {
+    lut_rem_pair lut_rem = perm_idx_lut[dist - 3];
+    __m128i perm_vec, ret_vec;
+#ifdef Z_MEMORY_SANITIZER
+    /* Important to note: 
+     * This is _not_ to subvert the memory sanitizer but to instead unpoison some
+     * bytes we willingly and purposefully load unitialized that we swizzle over
+     * in a vector register, anyway.  If what we assume is wrong about what is used,
+     * the memory sanitizer will still usefully flag it */
+    __msan_unpoison(buf+dist, 16 - dist);
```
nmoinvaz: Whitespace between `buf` and `dist`.

arch/x86/chunkset_sse41.c
```diff
@@ -0,0 +1,98 @@
...
+#endif
+    ret_vec = _mm_loadu_si128((__m128i*)buf);
+    *chunk_rem = lut_rem.remval;
+
+    perm_vec = _mm_load_si128((__m128i*)(permute_table+lut_rem.idx));
```
nmoinvaz: Whitespace between `permute_table` and `lut_rem.idx`

arch/x86/chunkset_avx.c
```diff
@@ -14,10 +15,43 @@ typedef __m256i chunk_t;
 #define HAVE_CHUNKMEMSET_2
 #define HAVE_CHUNKMEMSET_4
 #define HAVE_CHUNKMEMSET_8
+#define HAVE_CHUNK_MAG
 
-static inline void chunkmemset_1(uint8_t *from, chunk_t *chunk) {
-    *chunk = _mm256_set1_epi8(*(int8_t *)from);
-}
+/* Populate don't cares so that this is a direct lookup
```
nmoinvaz: This comment should be formatted so it takes fewer lines.

nmoinvaz (approved):
Only have some formatting changes, otherwise it looks good to me.

arch/x86/chunkset_sse41.c
Dead2: What I meant is that it would make sense in theory if a user knows that any machine they deploy to is for example a Haswell or newer cpu. In that case there is no upside to keeping around any functions that will never be used, such as SSE2 functions that have better suited versions.

We don't currently support properly selecting this though, but if we want to we could have a config parameters like `MIN_CPU=SSE4`  and`MAX_CPU=AVX2`  for example. That could then make sure we don't compile in AVX512 functions at all, as well as any SSE2/SSSE3 functions that have better SSSE3/SSE4 variants, unless needed by those replacements. It would still select between any SSE4/AVX1/AVX2 where available. If a function only has SSE2 and AVX512 implementations, then SSE2 would still be used since that is the best one available under those constraints. C fallbacks could potentially be kept or removed depending on a `C_FALLBACK` parameter for example.

This would of course mean a relatively major restructuring of CMakeLists.txt and probably some changes to defines used to select functions, but I think it does have some merit. The only major gain from this of course is that the library size is reduced and that can potentially improve cache locality.
A further step would be to bypass functable for functions where we know there is only the one option left, but that would probably be a lot more messy.

Dead2:
This needs a rebase again, and a few requested changes above, after that we'll hopefully be able to land this soon ;)

arch/x86/chunkset_sse41.c
nmoinvaz: I wouldn't suggest implementing `MIN_CPU` and `MAX_CPU` until we have is a use case and demand the justifies it, instead of just theory. It reminds me of all the work I did on the `attribute(target)` PR that never made it in.

KungFuJesus: I get why the user might want to I just mean there are reasons that they shouldn't, mostly for the places where we do use both functions in tandem.

On my other PR I've been doing some interesting experiments.  I created a branch where this changed was merged into the inline adler changes and it became universally slower, but I think for reasons that are unrelated to either improvements.  The test case of the trivially compressible image for levels 7, 8, & 9 are still better off with this.  

I can't say for sure what's happening but I did notice that a lot of constants that are populated into vector registers did seem to be measuring some small amount of CPU time where they weren't before (and they were being loaded from a different base variable and offset from said base variable).  However, disabling SSE2 (and not chunkmemset_avx2), managed to restore a fair bit of performance.  This may be the effect that @Dead2 had been referring to with regard to function alignment, since while compiling fewer functions there, the binary wasn't appreciably smaller since we still had the permutation table.  

@Dead2 you mentioned something about moving hotter functions somewhere in the binary somehow?  Is this something PGO could maybe do?  Or maybe a custom linker script? You said you had a compile flag to at least make this consistent, what was it again?

KungFuJesus: Hah, yeah, -falign-functions=32 seems to have brought back pretty much all of that performance.  *sigh*, compilers.  It is interesting how little function alignment is discussed in the realm of optimization, though.  Even PGO didn't seem to pick up on this.

Interesting reads:
https://easyperf.net/blog/2018/01/25/Code_alignment_options_in_llvm
https://easyperf.net/blog/2018/01/18/Code_alignment_issues

Dead2: Yes, function alignment is part of the "problem" since it is aligned to 16 bytes by default (at least on x86-64) while cachelines are 64 bytes. So when I do comparison benchmarks I add this CMake parameter `-DCMAKE_C_FLAGS="-falign-functions=64"`. This means that each function is always placed the same in the cacheline across multiple builds, instead of starting at a random 16-byte aligned spot (sometimes this makes a big difference when a hot function switches between crossing cachelines and not).

The other problem is cache aliasing, and that is more likely to get worse with the above, since the change above means we request more cachelines total (max 1 function per cacheline instead of max 4 functions), so there is a bigger chance of cachelines getting demoted to L2 for example. But I don't think our code size is big enough that this usually matters for most L1I caches.

In release builds you can get some performance improvements by having multiple related-to-each-other functions in the same cacheline.

PGO utilizes this effect by grouping hot functions so that no cold functions end up residing in the same cacheline as hot functions. This relies on good profiling however, and since we have so many exported functions it might be hit/miss depending on actual use. A more manual approach (guided by profiling) would be to add gcc/clang attributes that group certain functions into separate blocks, such as "inflate", "deflate", "common" and the rest of the more cold functions into the default block. This might be less of an improvement than PGO (since PGO also does optimization changes for hot/cold functions), but unlike PGO it would not mean any extra steps for the packager/end-user.

Regarding MIN_CPU/MAX_CPU, that would be a pretty neat thing to support, and probably not very hard either, but it would of course require that situations where for example SSE4 functions requiring certain SSE2 functions actually pulls those in automatically as well instead of being a strict lower exclusion limit.

😄

mtl1979: When you specify minimum CPU high enough, compiler can optimize the code better as the related functions are all in same compilation unit. This is kind of half-way between having LTO disabled and enabled, as some function variants still need to make extern calls to "shorter" variants. This is similar to bypassing functable when there is only one or two versions (basically unoptimized C variant and one vectorized variant).

KungFuJesus: From my limited testing in just trying to use PGO on the benchmark build artifacts (which is of course, probably stupid but a fun experiment), PGO wasn't able to pick function alignments nearly as well as the static alignment specified to GCC.  It certainly seems like there are a lot of functions we know will be cold (all of the stub functions, the cpu detection functions, etc), so I do wonder if a custom linker script could group all of those together to get us a tighter binary while allowing us to put the hotter functions with loops where alignment really matters to the instruction cache at 32 or 64 byte aligned boundaries.  Then again, a `#pragma GCC optimize()` macro could probably give us a similar effect.

There also might be some fruit in reordering some of the fields in these structs such as inflate_state.  I know things accessed frequently together should be close together but there might be some amount of padding that is occurring between struct members as well for alignment purposes (not egregious padding, since everything is mostly 32 or 64 bit, but some).  For sure the crc_fold struct should probably be first, since its first members require 64 byte alignment.  The "work" and "len" structs, while a modulus of 4 bytes, might still be better at the tail end of the struct.  The "dmax" field probably also can be guarded by a macro conditionally if INFLATE_STRICT is enabled or not.

Dead2: @KungFuJesus The attribute I mentioned earlier for grouping inflate/deflate/common etc is called `section` :
https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html

Separating inflate and deflate functions should provide better grouping than just bunching them all into one big "warm" section. Of course that also requires a common section for any shared hot functions, but I don't imagine having a couple hot sections switch between incurs as big a penalty as just having random hot/cold functions bunched together.
All the cold code should then just go to the generic `text` section like it normally would.
Using these should also be easier than using and maintaining linker-scripts.
There is a risk that these could get stripped if a packager is too eager to strip any "extra" data from the library/executable though, but then it would also quickly fail any actual test.

While the attributes `hot` and `cold` also put code into different sections, these affect optimization as well and also cause branch prediction changes. `cold` should really only be used for functions that nearly never get used, such as error handlers. These are the ones used by PGO, but carry a higher risk for slowing down the code when done manually, also the above illustrates the need for good and accurate profiling for PGO (something more easily done for a specific applications use-case rather than library-wide).

KungFuJesus: Interesting, I've never seen these gcc attributes before. The PMCs listed in that article are available on my CPU, it might be an interesting experiment to find the hotter icache miss stalls from that.

Dead2 (approved):

PR #1248, opened 2022-04-11
Use PREFIX() for some of the Z_INTERNAL symbols

iii-i:
https://github.com/powturbo/TurboBench links zlib and zlib-ng into the
same binary, causing non-static symbol conflicts. Fix by using PREFIX()
for flush_pending(), bi_reverse(), inflate_ensure_window() and all of
the IBM Z symbols.

Note: do not use an explicit zng_, since one of the long-term goals is
to be able to link two versions of zlib-ng into the same binary for
benchmarking [1].

[1] https://github.com/zlib-ng/zlib-ng/pull/1248#issuecomment-1096648932

### DIFF
--- a/arch/s390/crc32-vx.c
+++ b/arch/s390/crc32-vx.c
@@ -198,7 +198,7 @@ static uint32_t crc32_le_vgfm_16(uint32_t crc, const unsigned char *buf, size_t 
 #define VX_ALIGNMENT 16L
 #define VX_ALIGN_MASK (VX_ALIGNMENT - 1)
 
-uint32_t Z_INTERNAL s390_crc32_vx(uint32_t crc, const unsigned char *buf, uint64_t len) {
+uint32_t Z_INTERNAL PREFIX(s390_crc32_vx)(uint32_t crc, const unsigned char *buf, uint64_t len) {
     uint64_t prealign, aligned, remaining;
 
     if (len < VX_MIN_LEN + VX_ALIGN_MASK)
--- a/arch/s390/dfltcc_common.c
+++ b/arch/s390/dfltcc_common.c
@@ -35,7 +35,7 @@ static inline int is_dfltcc_enabled(void) {
     return is_bit_set((const char *)facilities, DFLTCC_FACILITY);
 }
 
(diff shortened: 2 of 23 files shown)


mtl1979:
Should use `PREFIX` macro from zbuild.h instead of hardcoding the prefix. Most of documentation/comments shouldn't use any prefix as it can be changed at configure time.

Dead2:
IIRC, we already don't use `PREFIX` for a few of the "internal" symbols, just a prefix of `zng_`.
Not sure whether we actually need those to be configurable.

iii-i:
Yeah, I was following what seemed to be an existing convention. E.g.: `zng_tr_flush_bits()`, `zng_tr_align()`, `zng_tr_stored_block()`.

mtl1979:
I'm not so sure hard-coding them in architecture-specific functions is well-established convention. For the generic internal functions following existing style is acceptable.

Dead2:
But what do we gain by using `PREFIX` for intenal functions, even in architure-speficic code?
Prefixing them means we also need to use prefixing when we call them, so that is another layer of complexity that we don't need unless it brings us something tangible.

mtl1979:
@Dead2 The whole purpose of using the macro instead of hardcoding is to discourage linking against both zlib and zlib-ng in compatibility mode as static library... When it's absolutely needed, symbol prefix should be explicitly set instead.

Dead2:
@mtl1979 But that is still "discouraged against" since all the exported symbols would already be in conflict between stock zlib and zlib-ng in compat mode. We don't need the internal symbols to conflict as well.
The only real use-case I can think of is if someone wants to statically link two instances of zlib-ng for whatever reason, is that something we need to support?

mtl1979:
@Dead2 For benchmarking purposes, it makes perfectly sense to link against two zlib-ng copies with major internal changes.

When just using compatibility mode, it makes sense to use prefix `zng_` as little as possible.

iii-i:
* Undo an accidental comment change.

iii-i:
* Rebased.

Dead2 (approved):

PR #1249, opened 2022-04-12
arch/arm/arm_features.c: fix uclibc build

ffontaine:
Fix the following build failure on uclibc:

```
/home/autobuild/autobuild/instance-1/output-1/build/zlib-ng-2.0.6/arch/arm/armfeature.c:4:12: fatal error: sys/auxv.h: No such file or directory
    4 | #  include <sys/auxv.h>
      |            ^~~~~~~~~~~~
```

Fixes:
 - http://autobuild.buildroot.org/results/6e04affd52528d7c8398f303f4af720f3ad67bfd

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/arch/arm/arm_features.c
+++ b/arch/arm/arm_features.c
@@ -1,6 +1,6 @@
 #include "../../zbuild.h"
 
-#if defined(__linux__)
+#if defined(__linux__) && (defined(ARM_AUXV_HAS_CRC32) || defined(ARM_AUXV_HAS_NEON))
 #  include <sys/auxv.h>
 #  ifdef ARM_ASM_HWCAP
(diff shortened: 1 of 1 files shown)


mtl1979:
We could also add test for header `sys/auxv.h` using dummy `getauxval` call in both cmake and configure, so we don't need to test if any of the feature flags are defined.

ffontaine:
Indeed, this check could be useful especially as it seems that `getauxval `is also unconditionally used for powerpc ans s390

ffontaine:
I'm closing this one in favor of #1250

PR #1250, opened 2022-04-12
Check that sys/auxv.h exists at configure time and add preprocessor define for it.

mtl1979:
Makes it easier to check in source files if we have header `sys/auxv.h`, so we don't try to include it if it doesn't exist.

See #1249.


nmoinvaz (approved):
LGTM

ffontaine:
I would advise to also use `HAVE_SYS_AUXV_H` in 

https://github.com/zlib-ng/zlib-ng/blob/5eefff644e382305a7b64f061344aecb73fad88b/arch/s390/s390_features.c#L4

and

https://github.com/zlib-ng/zlib-ng/blob/5eefff644e382305a7b64f061344aecb73fad88b/arch/power/power_features.c#L7

mtl1979:
> arch/s390/s390_features.c

Done... However `arch/s390/s390_features.c` might require additional checks to also protect call to `getauxval` as it is not checked at configure time when using cmake.

ffontaine:
Sure, we don't have any failures on buildroot autobuilders for powerpc and s390x (yet) but indeed `getauxval` should also be protected.

mtl1979:
> Sure, we don't have any failures on buildroot autobuilders for powerpc and s390x (yet) but indeed `getauxval` should also be protected.

On s390, the whole file only gets included if the specific feature is enabled during configuring... For other architectures we include the file if optimizations are enabled in general.

CMakeLists.txt
```diff
@@ -653,6 +657,9 @@ if(WITH_OPTIM)
         endif()
     elseif(BASEARCH_S360_FOUND)
         if(WITH_CRC32_VX)
```
nmoinvaz, line 659: I wouldn't wrap `check_s390_intrinsics` in the if statement if we end up using that function for other things.

nmoinvaz: I wouldn't wrap check_s390_intrinsics in the if statement if we end up using that function for other things.

arch/power/power_features.c
```diff
@@ -1,10 +1,12 @@
 /* power_features.c - POWER feature check
  * Copyright (C) 2020 Matheus Castanho <msc@linux.ibm.com>, IBM
- * Copyright (C) 2021 Mika T. Lindqvist <postmaster@raasu.org>
+ * Copyright (C) 2021-2022 Mika T. Lindqvist <postmaster@raasu.org>
  * For conditions of distribution and use, see copyright notice in zlib.h
  */
 
-#include <sys/auxv.h>
+#if defined(HAVE_SYS_AUXV_H)
```
nmoinvaz: Should just use `ifdef`?

arch/s390/s390_features.c
```diff
@@ -1,10 +1,14 @@
 #include "../../zbuild.h"
 #include "s390_features.h"
 
-#include <sys/auxv.h>
+#if defined(HAVE_SYS_AUXV_H)
```
nmoinvaz: Should just use `ifdef`?

arch/s390/s390_features.c
```diff
@@ -1,10 +1,14 @@
...
+#  include <sys/auxv.h>
+#endif
 
-Z_INTERNAL int s390_cpu_has_vx;
+Z_INTERNAL int s390_cpu_has_vx = 0;
 
 void Z_INTERNAL s390_check_features(void) {
+#if defined(S390_FEATURES)
```
nmoinvaz: Should just use `ifdef`?

arch/s390/s390_features.c
```diff
@@ -1,10 +1,14 @@
 #include "../../zbuild.h"
 #include "s390_features.h"
 
-#include <sys/auxv.h>
+#if defined(HAVE_SYS_AUXV_H)
+#  include <sys/auxv.h>
+#endif
 
-Z_INTERNAL int s390_cpu_has_vx;
+Z_INTERNAL int s390_cpu_has_vx = 0;
```
nmoinvaz, line 8: Aren't all static variables automatically initialized to zero?

configure
```diff
@@ -733,6 +733,19 @@ else
   echo "Checking for strerror... No." | tee -a configure.log
 fi
 
+# check for getauxval() for architecture feature detection at run-time
+cat > $test.c <<EOF
+#include <sys/auxv.h>
+int main() { return getauxval(0); }
```
nmoinvaz, line 739: This one is passing in 0 to `getauxval` but the cmake version is passing in other values..

nmoinvaz (commented):
Looks good except a few minor things.

CMakeLists.txt
mtl1979: I wasn't sure about that... I can unwrap it...

arch/s390/s390_features.c
mtl1979: We explicitly declare them on other files too...

arch/s390/s390_features.c
mtl1979: Both works if there is only one test on a line... I can change that...

configure
mtl1979: That's because here we only test that the function exists... Later we test if the feature macros are defined, as those are not always defined in `sys/auxv.h`.

arch/s390/s390_features.c
nmoinvaz: Ok

Dead2 (approved):

PR #1251, opened 2022-04-13
Don't try to build tests if cmake is too old for them.

mtl1979:
This allows configuration step of cmake to finish even if cmake is too old. Some recent machines still run operating systems that are limited to cmake 3.10 or older.


mtl1979:
@nmoinvaz

CMakeLists.txt
```diff
@@ -1134,7 +1134,9 @@ option(ZLIB_ENABLE_TESTS "Build test binaries" ON)
 if(ZLIB_ENABLE_TESTS)
     enable_testing()
 
-    add_subdirectory(test)
+    if("${CMAKE_VERSION}" VERSION_GREATER_EQUAL "3.12")
```
nmoinvaz: No need for double quotes in either side of the equation because at the top of CMakeLists.txt we use something like:
```
```

CMakeLists.txt
```diff
@@ -1134,7 +1134,9 @@ option(ZLIB_ENABLE_TESTS "Build test binaries" ON)
...
+        add_subdirectory(test)
+    endif()
     if(WITH_BENCHMARKS)
```
nmoinvaz: The `WITH_BENCHMARKS` if statement should also be wrapped in the version check because it also uses `FetchContent` to retrieve `googlebenchmark` same as test does with `googletest`.

nmoinvaz:
I wonder if we should also print a message to the user:
```
message(STATUS "Minimum cmake version 3.12 not met for Google test or benchmark")
```
Can use STATUS or WARNING.

mtl1979:
> I wonder if we should also print a message to the user:
> 
> ```
> message(STATUS "Minimum cmake version 3.12 not met for Google test or benchmark")
> ```
> 
> Can use STATUS or WARNING.

If I wrap `WITH_BENCHMARKS` also, then a warning for it would be a good idea...

CMakeLists.txt
mtl1979: Changed.

CMakeLists.txt
mtl1979: Done.

nmoinvaz (approved):
LGTM

mtl1979:
@Dead2

Dead2 (approved):

PR #1252, opened 2022-04-13
Create crc32 fold templates

nmoinvaz:
This PR does something similar to #1240 in that it moves the crc32 and crc32 with copying functions into templates in order to reduce overall code needing to be maintained.


arch/x86/crc32_fold_pclmulqdq_tpl.h
```diff
@@ -0,0 +1,201 @@
...
+        memcpy(&xmm_crc_part, src, len);
+#ifdef COPY
+        _mm_storeu_si128((__m128i *)partial_buf, xmm_crc_part);
+        memcpy(dst, partial_buf, len);
+#endif
+        partial_fold((size_t)len, &xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3, &xmm_crc_part);
+    }
+    crc32_fold_save((__m128i *)crc->fold, xmm_crc0, xmm_crc1, xmm_crc2, xmm_crc3);
+#ifndef COPY
+    crc32_fold_save_partial((__m128i *)crc->fold, xmm_crc_part);
```
KungFuJesus: One thing I had specifically removed in the not copy version that I could have removed in both I believe is this "save_partial" call.  I don't believe this bit is important.

arch/x86/crc32_fold_pclmulqdq_tpl.h
```diff
@@ -0,0 +1,201 @@
...
+
+        memcpy(partial_buf, src, len);
+        xmm_crc_part = _mm_load_si128((const __m128i *)partial_buf);
+        memcpy(dst, partial_buf, len);
+#endif
+        goto partial;
+    }
+
+    algn_diff = ((uintptr_t)16 - ((uintptr_t)src & 0xF)) & 0xF;
+#ifdef COPY
```
nmoinvaz: @KungFuJesus I'm wondering if there is any better way to `#ifdef` this here. I just did the whole `algn_diff` if statement.. because the no copy version has the extra if statement..

arch/x86/crc32_fold_pclmulqdq_tpl.h
```diff
@@ -0,0 +1,201 @@
...
+            XOR_INITIAL(xmm_t0);
+            fold_1(&xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3);
+            xmm_crc3 = _mm_xor_si128(xmm_crc3, xmm_t0);
+            partial_fold(algn_diff, &xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3, &xmm_crc_part);
+
+            src += (algn_diff + 16);
+            len -= (algn_diff + 16);
+        }
+
+        xmm_crc_part = _mm_setzero_si128();
```
nmoinvaz: @KungFuJesus this part is interesting in it sets `xmm_crc_part` to zero after having previously loading it and then calling `partial_fold` with it..

arch/x86/crc32_fold_pclmulqdq_tpl.h
```diff
@@ -0,0 +1,186 @@
...
+    algn_diff = ((uintptr_t)16 - ((uintptr_t)src & 0xF)) & 0xF;
+    if (algn_diff) {
+        xmm_crc_part = _mm_loadu_si128((__m128i *)src);
+#ifdef COPY
+        _mm_storeu_si128((__m128i *)dst, xmm_crc_part);
+        dst += algn_diff;
+#else
+        XOR_INITIAL(xmm_crc_part);
+
+        if (algn_diff < 4 && init_crc != 0) {
```
nmoinvaz, line 66: @KungFuJesus this is what I was able to reduce this too. Hopefully that looks right. I'm not sure why the `algn_diff < 4` condition because it loads at least 16  bytes extra.

KungFuJesus: So the reason for that is if you have an input such that you have fewer than 4 bytes to fold in to make things align correctly, you end up with not enough bytes to XOR the initial checksum with (since 0 XOR with anything is 0). This is to catch that corner case. Instead the next 4 bytes are XOR'd with for the initial checksum.

nmoinvaz:
I think there was a bug in the current implementation where it was not loading the partial value in crc32_fold_load. It could have been something I introduced while cleaning up. The original crc32_folding code did so I've updated it.

arch/x86/crc32_fold_pclmulqdq_tpl.h
nmoinvaz: I've removed this part since original crc32_folding code did not have it. This may have been some kind of weird workaround for the fact that the partial crc was not being loaded.

arch/x86/crc32_fold_pclmulqdq_tpl.h
nmoinvaz: My second commit should have a cleaner version.

arch/x86/crc32_fold_pclmulqdq_tpl.h
nmoinvaz: It looks like that might have been a bug - in that it was never using the xmm_crc_part during crc32_fold_load.

KungFuJesus:
Hmm I'm not sure that the saved partials are being used after they are stored into crc_fold, are they? The partial fold just instead immediately folds it into the checksum. To actually make use of that you'd need to do a fairly complicated blend step at the start of each call.

Now I had a similar strategy with the adler checksums but found that it was slower.  Then again, we were only stashing a single value so the sunk cost of 3 x vector register stores hadn't been there before (2x component sums + leftovers).  In this case the partial fold store is only one more vector register store to the heap :shrug: .  I do suspect the 4 folded running sums _are_ worth doing as they're saving you from a slightly more expensive set of XOR multiplications between calls.  But then again, what I was doing with the adler checksums was saving us from from horizontal sums, shifts, subtractions, and integer multiplications in scalar GPRs and it still was faster to do those things anyway, hah.

nmoinvaz:
> Hmm I'm not sure that the saved partials are being used after they are stored into crc_fold, are they? The partial fold just instead immediately folds it into the checksum. To actually make use of that you'd need to do a fairly complicated blend step at the start of each call.

Bingo! Good observation. I think you are right. `xmm_crc_part` is definitely used. But each time it is folded into the checksum.

KungFuJesus (approved):

nmoinvaz:
Rebased.

Dead2:
This failed a lot of tests with MSAN enabled https://github.com/zlib-ng/zlib-ng/runs/6232192334?check_suite_focus=true

KungFuJesus:
> This failed a lot of tests with MSAN enabled https://github.com/zlib-ng/zlib-ng/runs/6232192334?check_suite_focus=true

It appears that a good number of these are for the partial store code.  I'm guessing the partial store previously was zero initialized before storing the contents of those 4 registers onto the heap.  It _might_ be harmless but it's worth a second look to see why we aren't seeing that right now.

nmoinvaz:
It is very odd. I'm not understanding why there is a memory sanitizer issue in `crc_fold_save` because `xmm_crc_part` is never stored there.

nmoinvaz:
I just initialized it to zero at the top at least that is where we had it before in one of the functions. Only god knows.

nmoinvaz:
This is ready.

arch/x86/crc32_fold_pclmulqdq.c
```diff
@@ -3,7 +3,7 @@
  * instruction.
  *
  * A white paper describing this algorithm can be found at:
- * https://www.intel.com/content/dam/www/public/us/en/documents/white-papers/fast-crc-computation-generic-polynomials-pclmulqdq-paper.pdf
+ * http://intel.ly/2ySEwL0
```
Dead2: I actually think having the original filename of the white paper can be quite useful.
Not sure whether the direct link or the shortened url will be more reliable in the long term, but with the filename it is easier to search for if the link it broken.

Dead2:
Looks to me like these commits can be squashed, as they don't actually seem to add anything valuable when separate and rather make it more difficult to review since minor things are seemingly changed back and forth.

Dead2:
# x86-64 i9-9900K
### baseline 856884022cc7202b159ba76c8e89bb1d9f7300d5
```
 Tool: minigzip   Levels: 1-9
 Runs: 15         Trim worst: 8

 1     44.409%      1.063/1.073/1.083/0.007        0.503/0.508/0.512/0.003       94,127,290
 2     35.518%      1.986/2.004/2.011/0.009        0.535/0.541/0.547/0.004       75,282,961
 3     33.882%      2.480/2.516/2.543/0.022        0.513/0.517/0.521/0.003       71,816,478
 4     33.174%      2.835/2.868/2.892/0.021        0.487/0.499/0.504/0.006       70,315,668
 5     32.660%      3.191/3.215/3.238/0.016        0.487/0.494/0.499/0.005       69,225,542
 6     32.508%      3.697/3.729/3.755/0.026        0.486/0.490/0.495/0.003       68,902,222
 7     32.255%      5.177/5.231/5.276/0.040        0.476/0.482/0.489/0.006       68,366,800
 8     32.167%      7.825/7.873/7.920/0.038        0.475/0.489/0.495/0.007       68,180,776
 9     31.887%      8.490/8.537/8.572/0.032        0.477/0.483/0.486/0.003       67,586,442

 avg1  34.273%                        4.116                          0.500
 tot                                259.326                         31.517      653,804,179

 133180    1392      48  134620   20ddc libz-ng.so.2
```


### PR #1252 a9f9d678a1fa318884eb67a51b17c1dd98552745
```
 1     44.409%      1.066/1.080/1.088/0.010        0.502/0.508/0.517/0.006       94,127,290
 2     35.518%      1.982/2.000/2.009/0.009        0.532/0.539/0.543/0.004       75,282,961
 3     33.882%      2.515/2.533/2.551/0.014        0.514/0.517/0.521/0.002       71,816,478
 4     33.174%      2.840/2.850/2.861/0.008        0.496/0.502/0.507/0.004       70,315,668
 5     32.660%      3.155/3.179/3.212/0.021        0.478/0.493/0.499/0.008       69,225,542
 6     32.508%      3.682/3.721/3.749/0.028        0.477/0.488/0.492/0.006       68,902,222
 7     32.255%      5.109/5.156/5.183/0.028        0.481/0.490/0.495/0.006       68,366,800
 8     32.167%      7.779/7.848/7.898/0.042        0.480/0.485/0.492/0.005       68,180,776
 9     31.887%      8.485/8.541/8.597/0.045        0.478/0.483/0.486/0.003       67,586,442

 avg1  34.273%                        4.101                          0.501
 tot                                258.347                         31.541      653,804,179

 132820    1392      48  134260   20c74 libz-ng.so.2
```

Nice text reduction, the compression speed is overall faster but I would call it inconclusive due to low amount of samples and high variation. It does not seem to have hurt anything at least 👍

KungFuJesus:
Hmm it ought to have generated the same exact code. That's a little surprising. Hopefully the compiler isn't deciding to not inline those "load" and "save" routines.

nmoinvaz:
The code is not exactly the same because we were able to simplify an if statement or two. When you try to merge the two code paths it becomes apparent.

KungFuJesus:
> The code is not exactly the same because we were able to simplify an if statement or two. When you try to merge the two code paths it becomes apparent.

Ahh that's vaguely familiar. From what I recall I simplified a bunch of branching that was happening with a ladder of goto's.

nmoinvaz:
Yea I think that and some changes to the `align_diff` if statement.

Dead2:
# rpi3 Aarch64
### baseline 856884022cc7202b159ba76c8e89bb1d9f7300d5
```
 0    100.008%      0.004/0.017/0.024/0.006        0.024/0.029/0.032/0.003       15,737,543
 1     54.165%      0.529/0.541/0.550/0.008        0.173/0.190/0.197/0.007        8,523,602
 2     43.877%      0.904/0.922/0.936/0.009        0.160/0.189/0.200/0.011        6,904,649
 3     42.389%      1.213/1.226/1.238/0.008        0.164/0.183/0.197/0.012        6,670,476
 4     41.647%      1.362/1.375/1.384/0.008        0.174/0.182/0.189/0.005        6,553,763
 5     41.217%      1.485/1.504/1.521/0.012        0.172/0.177/0.184/0.004        6,485,981
 6     41.037%      1.740/1.760/1.773/0.010        0.160/0.178/0.184/0.006        6,457,780
 7     40.778%      2.406/2.423/2.431/0.007        0.140/0.171/0.183/0.012        6,416,924
 8     40.704%      2.920/2.948/2.952/0.008        0.156/0.167/0.175/0.006        6,405,244
 9     40.409%      3.355/3.368/3.378/0.007        0.171/0.181/0.188/0.005        6,358,951

 avg1  48.623%                        1.608                          0.165
 avg2  54.026%                        1.787                          0.183
 tot                                225.177                         23.052       76,514,913

 115286    1672      16  116974   1c8ee libz-ng.so.2
```

### PR #1252 a9f9d678a1fa318884eb67a51b17c1dd98552745
```
 0    100.008%      0.008/0.017/0.020/0.004        0.016/0.028/0.036/0.006       15,737,543
 1     54.165%      0.530/0.547/0.553/0.007        0.171/0.191/0.200/0.008        8,523,602
 2     43.877%      0.911/0.924/0.937/0.008        0.172/0.193/0.204/0.009        6,904,649
 3     42.389%      1.194/1.222/1.238/0.013        0.159/0.184/0.195/0.010        6,670,476
 4     41.647%      1.362/1.378/1.382/0.005        0.168/0.179/0.188/0.007        6,553,763
 5     41.217%      1.487/1.502/1.514/0.009        0.166/0.178/0.184/0.006        6,485,981
 6     41.037%      1.744/1.760/1.768/0.006        0.156/0.178/0.188/0.009        6,457,780
 7     40.778%      2.393/2.427/2.438/0.012        0.155/0.169/0.182/0.009        6,416,924
 8     40.704%      2.940/2.950/2.963/0.007        0.156/0.165/0.180/0.007        6,405,244
 9     40.409%      3.360/3.370/3.379/0.006        0.161/0.176/0.184/0.006        6,358,951

 avg1  48.623%                        1.610                          0.164
 avg2  54.026%                        1.789                          0.182
 tot                                225.355                         22.962       76,514,913

 115286    1672      16  116974   1c8ee libz-ng.so.2
```

Not much happening here.

arch/x86/crc32_fold_pclmulqdq.c
nmoinvaz: Removed. I think it is best if we include the pdf in our docs.

Dead2:
@nmoinvaz What about my question about merging this into one commit? As far as I can see the 3 follow-up commits only contain churn that should have been resolved in the initial commit, or is there some reason they are split up?

Dead2:
@nmoinvaz What about my question about merging this into one commit? As far as I can see the 3 follow-up commits only contain churn that should have been resolved in the initial commit, or is there some reason they are split up?

nmoinvaz:
I don't know why I didn't see your comments earlier. I have squashed the commits.

PR #1253, opened 2022-04-13
IBM Z DFLTCC: Buffer deflate() input data 

iii-i:
Hi,

I'd like to gather some opinions on how to proceed with this.
I tried to initiate a discussion on the zlib mailing list, but it never got past the moderator, so only people on Cc: received my message:

```
Hello zlib developers,

I've been investigating a problem in the QEMU test suite on IBM Z [1]
[2] in connection with the IBM Z compression accelerator patch [3].

The problem is that a QEMU thread compresses data that is being
modified by another QEMU thread. zlib manual [4] does not state that
this is safe, however, the current stable zlib in fact tolerates it.

The accelerator, however, does not: not only what it compresses ends up
being unpredictable - QEMU actually resolves this just fine -
but the accelerator's internal state also ends up being corrupted.

I have a design question in connection to this: does zlib guarantee
that modifying deflate() input concurrently with deflate() is safe?
Or does it reserve the right to change this in the future versions?

Cc:ing zlib-ng folks for their opinion as well.

[1] https://lists.gnu.org/archive/html/qemu-devel/2022-03/msg06841.html
[2] https://lists.gnu.org/archive/html/qemu-devel/2022-04/msg00329.html
[3] https://github.com/madler/zlib/pull/410
[4] https://zlib.net/manual.html

Best regards,
Ilya
```

So far I got only the following responses from Dr. David Alan Gilbert:

```
Specifically what qemu does is :

    a) write block of RAM
    b) <notices block changed)
    c) write block of RAM again

we don't care whether (a)'s data is coherent on the destination
since we know we're going to keep rewriting it (c) until it stabilises.
So we accept undefined decoding of (a) under the assumption that the
stream will still be a valid stream and that a change in one instance
won't affect a future encoding to be invalid.  Is that assumption
correct?
```

and from @mtl1979:

```
Then the only issue is that the internal state used by hardware deflate
acceleration gets out of sync when the data is changed and compression is
restarted from known "good" position... As I said, in that case, the internal
state used by the accelerator needs to be either reset or resynced with the
state structure. As the code flow is unique to this specific "accelerator", it
is responsibility of the accelerator to detect it, not generic compression
code (in either zlib or zlib-ng).
```

I put some more details regarding the fix into the commit message: https://github.com/zlib-ng/zlib-ng/commit/28429507d69a5138a2f83ff3dba38e34cb56480e

Best regards,
Ilya

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -252,6 +252,12 @@ jobs:
             ldflags: -static
             codecov: ubuntu_gcc_s390x_dfltcc_compat
 
+          - name: Ubuntu Clang S390X DFLTCC MSAN
+            os: z15
+            compiler: clang-11
+            cxx-compiler: clang++-11
+            cmake-args: -GNinja -DWITH_DFLTCC_DEFLATE=ON -DWITH_DFLTCC_INFLATE=ON -DWITH_SANITIZER=Memory
+
           - name: Ubuntu MinGW i686
             os: ubuntu-latest
             cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-mingw-i686.cmake
--- a/arch/s390/dfltcc_deflate.c
+++ b/arch/s390/dfltcc_deflate.c
@@ -25,6 +25,9 @@ struct dfltcc_deflate_state {
     uint32_t block_size;               /* New block each X bytes */
     size_t block_threshold;            /* New block after total_in > X */
     uint32_t dht_threshold;            /* New block only if avail_in >= X */
+    uint8_t buf[DFLTCC_BUF_SIZE];      /* Internal buffer for input data */
+    uint32_t buf_offset;               /* Offset of input data in buf */
+    uint32_t buf_size;                 /* Size of input data in buf */
 };
 
 #define GET_DFLTCC_DEFLATE_STATE(state) ((struct dfltcc_deflate_state *)GET_DFLTCC_STATE(state))
@@ -44,6 +47,10 @@ void Z_INTERNAL PREFIX(dfltcc_reset_deflate_state)(PREFIX3(streamp) strm) {
     dfltcc_state->block_size = DFLTCC_BLOCK_SIZE;
     dfltcc_state->block_threshold = DFLTCC_FIRST_FHT_BLOCK_SIZE;
     dfltcc_state->dht_threshold = DFLTCC_DHT_MIN_SAMPLE_SIZE;
+
+    /* Initialize the internal buffer */
+    dfltcc_state->buf_offset = 0;
+    dfltcc_state->buf_size = 0;
 }
 
 void Z_INTERNAL PREFIX(dfltcc_copy_deflate_state)(void *dst, const void *src) {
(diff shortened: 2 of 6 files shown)


mtl1979:
There was some talk outside the mailing list, but no word from Mark Adler what is his "official" opinion about if zlib should tolerate it or not.

Dead2:
I can not speak for what zlib intends to support or not, in our compat-mode we attempt to be as compatible with zlib as we can, so this would depend on the zlib implementation (I realize this might be something madler would have to make a new policy about, and that he might in fact not clarify this).

But I think we could discuss whether zlib-ng in native-mode should clarify whether we want to restrict such usage.
My personal opinion is that while zlib-ng has the pointer to the data and is executing on it, nobody else should be touching that data and that it is the responsibility of the application to make sure that is the case.

On a related note; A pro if we were to always copy input to an internal buffer would be that we can be certain of alignment and extra trailing buffer space (in relation to vector operations).
The big con is that copying to a buffer is always slower than zero-copy, and zero-copy is IMO a very good thing to strive towards. Also buffers occupy more ram and could push other things out of cache.

dagrh:
> I can not speak for what zlib intends to support or not, in our compat-mode we attempt to be as compatible with zlib as we can, so this would depend on the zlib implementation (I realize this might be something madler would have to make a new policy about, and that he might in fact not clarify this).
> 
> But I think we could discuss whether zlib-ng in native-mode should clarify whether we want to restrict such usage. My personal opinion is that while zlib-ng has the pointer to the data and is executing on it, nobody else should be touching that data and that it is the responsibility of the application to make sure that is the case.
> 
> On a related note; A pro if we were to always copy input to an internal buffer would be that we can be certain of alignment and extra trailing buffer space (in relation to vector operations). The big con is that copying to a buffer is always slower than zero-copy, and zero-copy is IMO a very good thing to strive towards. Also buffers occupy more ram and could push other things out of cache.

Yep, our migration code has the same challenge of course; if we know zlib is defined to need the copy then that's fine and we can do it, but otherwise as you say it eats ram and cpu.  The code was originally written around just a write()/sendmsg and has had the compression layer inserted into it;  the 'write' case is fine with the data changing underneath us because of the structure of what's written (anything that changes always sets a flag to say we need to rewrite again later); so we're OK with undefined ness and don't care about the data that ends up in the block when received, but we don't expect the undefinedness to spil over into future blocks.

nmoinvaz:
[Use PREFIX() for some of the Z_INTERNAL symbols](https://github.com/zlib-ng/zlib-ng/pull/1253/commits/a75f5b3e3deabc8c05d0022f49ced3913ef90c6d) should be its own PR imo. It's been mentioned in issue #1279 also.

iii-i:
I need to rebase this, this part has already been committed as https://github.com/zlib-ng/zlib-ng/commit/9be98893aaf69892811d617122e28ea2dd8e17ba (apparently more is needed to solve the problem in a generic way, I was just concerned with the s390 part here).

Regarding this issue: I'm currently looking into using the inline history mode instead of the circular buffer history. This may reduce the overall amount of copying and somewhat reduce the 25% overhead.

iii-i:
Benchmarking has shown that inline history doesn't help much, so we'll most likely have to go with this solution. I'll let it sit in draft status until MSan and concurrency test PRs are merged and then rebase again.

One idea that I haven't tried yet would be to use `UFFDIO_WRITEPROTECT` and clear the history when writes are detected. However, userfaultfd is not available on all kernels, so a fallback to an internal buffer will be required anyway.

iii-i:
This PR is not necessary: concurrent modification of deflate() buffer is not allowed: https://lists.gnu.org/archive/html/qemu-devel/2022-07/msg00889.html

PR #1254, opened 2022-04-18
NEON version of compare256 

nmoinvaz:
@KungFuJesus's benchmarks on `compare256_neon`:
```shell
compare256/c/1                    4.46 ns         4.46 ns    157125252
compare256/c/8                    11.3 ns         11.3 ns     62119037
compare256/c/64                   79.2 ns         79.1 ns      8844665
compare256/c/256                   320 ns          320 ns      2187658
compare256/unaligned_16/1         4.46 ns         4.46 ns    157125577
compare256/unaligned_16/8         7.60 ns         7.60 ns     92106726
compare256/unaligned_16/64        46.1 ns         46.1 ns     15176921
compare256/unaligned_16/256        186 ns          186 ns      3756849
compare256/unaligned_32/1         4.98 ns         4.98 ns    140585352
compare256/unaligned_32/8         8.32 ns         8.31 ns     82366307
compare256/unaligned_32/64        37.0 ns         37.0 ns     18944178
compare256/unaligned_32/256        144 ns          144 ns      4847769
compare256/unaligned_64/1         4.98 ns         4.98 ns    140586140
compare256/unaligned_64/8         4.98 ns         4.98 ns    140585374
compare256/unaligned_64/64        20.2 ns         20.2 ns     34689923
compare256/unaligned_64/256       75.0 ns         75.0 ns      9218886
compare256/neon/1                 4.72 ns         4.72 ns    148395330
compare256/neon/8                 4.74 ns         4.74 ns    127194676
compare256/neon/64                13.6 ns         13.6 ns     51367775
compare256/neon/256               43.0 ns         43.0 ns     16287387
```


arch/arm/compare256_neon.c
```diff
@@ -0,0 +1,61 @@
...
+#  include <arm_neon.h>
+#endif
+#include "../../zbuild.h"
+#include "../../fallback_builtins.h"
+
+static inline uint32_t compare256_neon_static(const uint8_t *src0, const uint8_t *src1) {
+    uint32_t len = 0;
+
+    do {
+        uint8x16_t a, b, cmp;
```
KungFuJesus, line 18: We want to try to align one of the two loads here.  In real world use cases, it was definitely a benefit to x86 hardware that cared about alignment.  See: https://github.com/zlib-ng/zlib-ng/blob/develop/arch/x86/compare256_sse2.c#L36

For the most part, the added complexity usually doesn't hurt the cases where you'd be matching in the first vector and where you'd match in the last vector (where you'd have to do 2+ extra compares) seems highly unlikely for most real world use cases (even so, the extra XORs are likely to be inexpensive).

nmoinvaz: Agreed. I think it might be better as a separate PR. That way we can benchmark those changes separately. I doubt I'd able to tell the performance difference in qemu.

nmoinvaz:
I have noticed a difference in configure versus CMake, in that MFPU_NEON_AVAILABLE is required for all NEON files, but in configure it only adds the -mfpu=neon flag. I'm not sure which is the correct behavior.

nmoinvaz:
Another interesting difference is that the COMPARE256 versions don't have an associated `-DARM_NEON_COMPARE256`, well not for any platform..

arch/arm/compare256_neon.c
mtl1979: With aligning one of the inputs, it might not unroll the whole loop anymore as length is no longer guaranteed to be multiple of 16... I tried changing to half-width Neon and it didn't unroll anymore using latest gcc trunk.

There is slight penalty when moving 64 bit elements from vector registers as on ARM it is translated as move from floating-point register to integer register.

mtl1979:
> Another interesting difference is that the COMPARE256 versions don't have an associated `-DARM_NEON_COMPARE256`, well not for any platform..

Do we really need the option to disable vectorized COMPARE256?

nmoinvaz:
I mean we probably don't need to disable any of these either right? `-DARM_NEON_ADLER32 -DARM_NEON_CHUNKSET -DARM_NEON_SLIDEHASH`..

nmoinvaz:
Those could simply be hidden behind an `-DARM_NEON` instead.

nmoinvaz:
Those could simply be hidden behind an `-DARM_NEON` instead.

mtl1979:
> I mean we probably don't need to disable any of these either right? `-DARM_NEON_ADLER32 -DARM_NEON_CHUNKSET -DARM_NEON_SLIDEHASH`..

If I remember correctly, the idea was that we could disable them one by one, so we could test easily if there is a bug in the individual functions... I've seen processors that can't handle specific instructions, even though claim supporting the instruction set.

mtl1979:
> Those could simply be hidden behind an `-DARM_NEON` instead.

There is so many variants of NEON that it makes sense to restrict `ARM_NEON` meaning the 128-bit variant... We currently don't support the 64-bit or half-width variant, but do have partial support for double-width and quad-width loads/stores. Support for triple-width loads/stores was removed as it caused more issues than it improved performance.

arch/arm/compare256_neon.c
KungFuJesus: Is it any faster to move 4 32 bit values, though? It seems like having more jumps and compares would probably eat into those gains.

nmoinvaz:
> I have noticed a difference in configure versus CMake, in that MFPU_NEON_AVAILABLE is required for all NEON files, but in configure it only adds the -mfpu=neon flag. I'm not sure which is the correct behavior.

@mtl1979 any comment on this? Do you know which is the proper way? If so I can make the update to the one or the other file.

mtl1979:
@nmoinvaz Some compiler versions don't support `-mfpu=neon` even if the processor has NEON, others do. configure is more optimistic as it's more likely that `-mfpu=neon` is supported as long as the CPU family has NEON.

With cmake, it's possible to use compiler that targets alternate C library, so we fail if we encounter mismatch between `-march` and `-mfpu`.

Dead2 (approved):

Dead2:
@nmoinvaz Is this one ready to go? I see a review thread above that is not resolved and while it looks unimportant, I am not sure whether you want to do some changes or not.

nmoinvaz:
I had not planned any further changes. It is good to go.

KungFuJesus:
Now that this is in, interestingly I'm finding no discernible influence on alignment with the A-73 but the A-53 of course does make an appreciable difference.  Hmm, might still be worth doing if the aligning doesn't impose too large a penalty, though.

KungFuJesus:
Interesting, aligning one of the inputs doesn't appreciable help the situation on Cortex-A53.  The SSE2 version also managed to get some gains there by eliding the loads with the compare for when it knew the loads were aligned, so perhaps that had a larger effect than the gained alignment.  Or just x86 can pipeline things better when at least one half of the loads are aligned, :shrug: .

PR #1255, opened 2022-04-19
Implement power9 version of compare256.

nmoinvaz:
I am creating a PR to continue the discussion. This is based on @mscastanho's work on zlib. I am using qemu so my benchmarks are not reliable. Need somebody to benchmark on real hardware. Anybody who has more experience feel free to take over this PR or just submit changes here and I'll update it.


arch/power/compare256_power9.c
```diff
@@ -0,0 +1,58 @@
...
+
+#include "match_tpl.h"
+
+#define LONGEST_MATCH_SLOW
+#define LONGEST_MATCH       longest_match_slow_power9
+#define COMPARE256          compare256_power9_static
+
+#include "match_tpl.h"
+
+#endif
```
mtl1979: Missing empty line at end of file.

arch/power/compare256_power9.c
```diff
@@ -0,0 +1,58 @@
...
+#include <altivec.h>
+#include "../../zbuild.h"
+#include "../../zendian.h"
+
+static inline uint32_t compare256_power9_static(const uint8_t *src0, const uint8_t *src1) {
+    vector unsigned char vsrc0, vsrc1, vc;
+    uint32_t len = 0, cmplen;
+
+    do {
+        vsrc0  = *((vector unsigned char *)src0);
```
mtl1979: Extra whitespace before `=`.

arch/power/compare256_power9.c
```diff
@@ -0,0 +1,58 @@
...
+        /* Compare 16 bytes at a time. Each byte of vc will be either
+        * all ones or all zeroes, depending on the result of the comparison. */
+        vc = (vector unsigned char)vec_cmpne(vsrc0, vsrc1);
+
+        /* Since the index of matching bytes will contain only zeroes
+         * on vc (since we used cmpne), counting the number of consecutive
+         * bytes where LSB == 0 is the same as counting the length of the match. */
+#if BYTE_ORDER == LITTLE_ENDIAN
+        cmplen = __builtin_vec_vctzlsbb(vc);
+        //asm volatile("vctzlsbb %0, %1\n\t" : "=r" (len) : "v" (vc));
```
mtl1979: We don't need the asm version as comment, we prefer intrinsics when available.

arch/power/compare256_power9.c
```diff
@@ -0,0 +1,58 @@
...
+#else
+        cmplen = __builtin_vec_vclzlsbb(vc);
+        //asm volatile("vclzlsbb %0, %1\n\t" : "=r" (len) : "v" (vc));
```
mtl1979: We don't need the asm version as comment, we prefer intrinsics when available.

arch/power/compare256_power9.c
nmoinvaz: @mscastanho do you know why the assembly is the same in both branches? I saw this on your zlib PR too. https://github.com/madler/zlib/pull/459/files#r852564423

KungFuJesus: I don't think it is, one is trailing zero the other is leading zero.  I also would wager the Power 9 big endian path is untested since the common configuration for newer power seems to be little endian.

KungFuJesus: Good lord, can somebody from IBM tell them they need to publish more documentation on their ISA?  It's gotten to the point that Google has no idea what I'm talking about when I a search for the exact instruction name, let alone the intrinsic.  From what I can tell, when I finally found what the acronym was in a patent document of all things, it counts the number of bytes in the vector for which the least significant bit is zero, trailing or leading depending on whatever direction you pick.  Which is basically an instruction that shortcuts the whole "movemask + tzcnt" sequence in x86 land (probably because up until more modern versions of POWER, moving between GPRs and vectors wasn't an option).

I'd like to know what the minimum version of POWER is for this instruction to be supported since I'd like to implement this on altivec hardware.  A GCC include _kind of_ indicates that it might just work, but possibly only on PPC64?  I do have a 64 bit OS and a 64 bit PPC CPU, but a good number of PPC systems of this era are 32 bit only.  A compile time #ifdef I guess would take care of that...

KungFuJesus: It looks like where it is in this header indicates that it may not be supported until POWER9:
https://gist.github.com/Cyan4973/1f87f02b2c96a37470205baaf12e1407

And only if compiling with PPC64 support, as well.  So, the < POWER9 variants will need to use a sequence that takes a trip through stack memory, barring some other clever mechanism to find the first set bit in a vector.

mtl1979: We really don't need to worry about trying to support every iteration of POWER architecture with vector optimized variants... Even POWER9 doesn't have all the useful features as far as I know, as POWER10 added some useful instructions.

Until someone or some company donates us hardware that we can use to test and benchmark every possible architecture and CPU variant, we can only merge in pull requests that we actually can verify works as intended.

mtl1979:
This seems to be missing changes required in `arch/power/Makefile.in`. Build using configure complains about missing files during linking.

arch/power/compare256_power9.c
KungFuJesus: Right, but arguably POWER4 and its derivatives from Apple are going to be a bit more commonplace due to their ubiquity and wide scale marketing in the consumer market.  Though, workstation hardware from IBM and Talos may eventually exceed that stockpile.  It'd sure help if they didn't cost as much as cars.  Talos is a step in the right direction, but it's still a bit steep.

mtl1979: My first encounter with PowerPC/Power was when BeOS was released... After that I haven't bought newer hardware. I'm not counting how many x86 or ARM devices I currently have at home as that number is quite high due to my previous work.

KungFuJesus: @nmoinvaz  actually, strictly speaking I don't think it actually matters if comparing the lead or the trail given that for equality either bit would be zero of the byte, hah.

KungFuJesus: Well...perhaps not.  It should either be one way or the other but the fact that this code switches based on endianness tells me that POWER implicitly swaps bytes somehow when loading a vector.  That more or less jives with their unaligned load semantics from what I recall a while back when reading about some of the newer POWER variants.  This is, of course, kind of confusing, you would think that loading a vector of characters would not have a implicit endianness switch with it.

mtl1979: Even on little-endian architectures, the vector processing unit is usually big-endian, so element order is reversed. This is separate from byte order inside larger elements.

Some architectures can load and reverse the element order (so it matches memory order) using single instruction.

nmoinvaz: Fixed.

arch/power/compare256_power9.c
nmoinvaz: Fixed.

arch/power/compare256_power9.c
nmoinvaz: Fixed.

nmoinvaz:
Should be fixed now.

arch/power/compare256_power9.c
nmoinvaz: Fixed.

nmoinvaz:
Rebased.

arch/power/compare256_power9.c
mscastanho: @KungFuJesus The go-to place for the Power ISA docs is the OpenPower Foundation website: https://openpower.foundation/specifications/isa/

And you're right, `vctzlsbb`/`vclzlsbb` were added in ISA version 3.0 (easier to verify in the tables on Appendices E-H of the ISA), which was first implemented by Power 9, so they won't be available for earlier processors.

> POWER implicitly swaps bytes somehow when loading a vector.

Yep. The way the loaded bytes are placed in the vector changes if it's BE or LE, that's why the code counts either trailing zeroes or leading zeroes, depending on the endianness.

arch/power/compare256_power9.c
mscastanho: @mtl1979 @nmoinvaz Unfortunately intrinsics may not be the best option in this specific case =(
Their semantics were misimplemented by GCC, and are bound to change (to use the correct one) starting with the upcoming GCC 12 (see [this GCC commit](https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=3f30f2d1dbb3228b8468b26239fe60c2974ce2ac)). So to make sure it works on both newer and older GCC versions, we must use the inline assembly to generate the instruction we want.

arch/power/compare256_power9.c
KungFuJesus: I'll keep that in mind for the future.  It's a bit strange to me that google isn't indexing content from there very well - you'd think that'd be the most relevant search result.  

On a sadder note it seems that the informatimago mirror that had a gold mine of the old Apple optimization guides for the PPC970 and the 7400/7450 is now gone.  Archive.org seems to have captured a good amount of it but I wish I had stashed that before it disappeared.

mscastanho:
@nmoinvaz I gave it a try on a P9 (shared system, not completely tunned for running benchmarks, but the results were consistent across a few runs):
```
$ ./test/benchmarks/benchmark_zlib --benchmark_filter=compare256*
[...]
compare256/c/1                    4.33 ns         4.33 ns    161723123
compare256/c/8                    6.09 ns         6.09 ns    115011891
compare256/c/64                   29.5 ns         29.5 ns     23756728
compare256/c/256                   118 ns          118 ns      5934979
compare256/unaligned_16/1         4.40 ns         4.39 ns    159373570
compare256/unaligned_16/8         4.62 ns         4.62 ns    151676456
compare256/unaligned_16/64        15.4 ns         15.4 ns     45362280
compare256/unaligned_16/256       52.6 ns         52.6 ns     13302486
compare256/unaligned_32/1         4.29 ns         4.29 ns    163085134
compare256/unaligned_32/8         4.30 ns         4.30 ns    162712326
compare256/unaligned_32/64        11.4 ns         11.4 ns     61214092
compare256/unaligned_32/256       38.5 ns         38.5 ns     18189140
compare256/unaligned_64/1         4.25 ns         4.25 ns    164767206
compare256/unaligned_64/8         4.25 ns         4.25 ns    164765168
compare256/unaligned_64/64        8.14 ns         8.13 ns     87002943
compare256/unaligned_64/256       20.0 ns         20.0 ns     34968546
compare256/power9/1               4.63 ns         4.63 ns    151304726
compare256/power9/8               4.63 ns         4.63 ns    151311836
compare256/power9/64              5.36 ns         5.36 ns    130658917
compare256/power9/256             14.6 ns         14.6 ns     48006414
```
So looks like this is indeed speeding up the comparisons, specially for larger lengths.

KungFuJesus:
> So looks like this is indeed speeding up the comparisons, specially for larger lengths.

That's more or less in line with what we see on other architectures, so that's about what I'd expect (marginal gains for early terminating string matches, larger gains for longer matches).  It's interesting IBM decided to make the ctz and clz instructions operate directly on the vector registers - I take it you still can't move from vector to scalar registers without going through memory?

mscastanho:
> It's interesting IBM decided to make the ctz and clz instructions operate directly on the vector registers 

Note there are also `cntlz*` and `cnttz*` instructions, which operate on general purpose registers. The ones used here are different, and will count trailing/leading zeroes considering only the first bit of each byte in the vector.

> I take it you still can't move from vector to scalar registers without going through memory?

@KungFuJesus That's been possible for a while now, but later ISA versions added more instructions that can also handle bitmasks and different operand sizes. There are `mtvsr*` and `mfvsr*` family of instructions for that now.

KungFuJesus:
> The ones used here are different, and will count trailing/leading zeroes considering only the first bit of each byte in the vector

Right, this is the x86 analog of PMOVMSKB followed by a count trailing/leading zero on the GPR operand of that instruction.  It just skips the trip through scalar registers, which is why I figured it still wasn't possible.  It's a neat shortcut, maybe that's why IBM filed a patent for it.

KungFuJesus:
@mscastanho Just curious, can you run the adler benchmarks?  I made some improvements to the "vmx" variants and looking at the power8 versions, they are fairly similar except that the power 8 versions can do unaligned loads.  The PPC970 enjoyed some pretty nice gains with some extra unrolls that aren't present in the power8 version and I wonder if that one could benefit from a similar treatment.

Dead2:
@nmoinvaz This looks like it is good to go, is it?

@mscastanho Would you approve this PR?

nmoinvaz:
Yes there isn't anything more to do AFAIK. @mscastanho approval would be good.

arch/power/compare256_power9.c
```diff
@@ -0,0 +1,56 @@
...
+
+        /* Compare 16 bytes at a time. Each byte of vc will be either
+         * all ones or all zeroes, depending on the result of the comparison. */
+        vc = (vector unsigned char)vec_cmpne(vsrc0, vsrc1);
+
+        /* Since the index of matching bytes will contain only zeroes
+         * on vc (since we used cmpne), counting the number of consecutive
+         * bytes where LSB == 0 is the same as counting the length of the match. */
+#if BYTE_ORDER == LITTLE_ENDIAN
+        cmplen = __builtin_vec_vctzlsbb(vc);
```
mscastanho: Unfortunately intrinsics may not be the best option in this specific case =(
Their semantics were misimplemented by GCC, and are bound to change (to use the correct one) starting with the upcoming GCC 12 (see [this GCC commit](https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=3f30f2d1dbb3228b8468b26239fe60c2974ce2ac)). So to make sure it works on both newer and older GCC versions, we must use the inline assembly to generate the instruction we want.

mscastanho (changes requested):
@Dead2 @nmoinvaz I had a [comment](https://github.com/zlib-ng/zlib-ng/pull/1255#discussion_r857652460) that probably got lost in the resolved conversation. I copied it below to the review.

I think that's the only thing missing. But I'll do a final review just in case.

arch/power/compare256_power9.c
Dead2: We don't like to use asm if it can be avoided, since it typically makes the code less readable and maintainable and more likely to develop errors when changed.

So perhaps an alternative would be to make our own "intrinsics" as a workaround for this. When compilers fix theirs, we can switch to only use them as fallbacks for older compilers. Probably best to prefix them so they are not confused with actual compiler intrinsics, though we could possibly rename them to match future compilers when those are released.

Does that sound like a good solution?

mscastanho: Sounds good to me.

nmoinvaz: Fixed.

nmoinvaz:
Rebased and fix asm issue. Also updated readme.

mscastanho (approved):

PR #1256, opened 2022-04-22
Don't force CMAKE_BUILD_TYPE on multi config generators

nmoinvaz:
This issue is currently causing problems for us when building with multi-config generator such as Xcode or MSVC. If we choose `RelWithDebInfo` through `cmake .. -A Win32 && cmake --build . --config RelWithDebInfo` then it ends up building in `Release` build type which conflicts with the project as a whole. 

Ideally it would be better if we don't try and set `CMAKE_BUILD_TYPE` ourselves, but in leu of that..

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,13 +59,15 @@ endif()
 
 # Make sure we use an appropriate BUILD_TYPE by default, "Release" to be exact
 # this should select the maximum generic optimisation on the current platform (i.e. -O3 for gcc/clang)
-if(NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Ideally it would be better if we don't try and set `CMAKE_BUILD_TYPE` ourselves, but in leu of that..

mtl1979:
@nmoinvaz When building from command-line, a sane default is useful... Without explicit default, we get huge binaries as for some compilers, the default is `RelWithDebugInfo`.

mtl1979 (approved):

Dead2 (approved):

PR #1258, opened 2022-04-27
Regression: stream adler checksum no longer matches after inflateSetDictionary (bisected)

joshtriplett:
The latest version of zlib-ng produced a regression, detected by the flate2-rs testsuite: after setting a custom dictionary with `inflateSetDictionary`, the stream's adler checksum doesn't match that dictionary's adler checksum.

I reduced this to the new test included in this PR, and bisected with this test. This regression was introduced by commit 8550a90de4dcb8589a7d48fe308c4c45bba5a466 ("Leverage inline CRC + copy"); the test passes before that commit, and fails on or after that commit.

### DIFF
--- a/test/test_dict.cc
+++ b/test/test_dict.cc
@@ -72,6 +72,7 @@ TEST(dictionary, basic) {
             EXPECT_EQ(d_stream.adler, dict_adler);
             err = PREFIX(inflateSetDictionary)(&d_stream, (const unsigned char*)dictionary,
                 (uint32_t)sizeof(dictionary));
+            EXPECT_EQ(d_stream.adler, dict_adler);
         }
         EXPECT_EQ(err, Z_OK);
     }


KungFuJesus:
Good catch.  I think the issue here is the call to update window.  We want the copy behavior but not the checksum behavior in this case.

KungFuJesus:
So I'm not sure if it's a proper fix, but...does this test now pass if you say, do this before and after the updatewindow call in inflateSetDictionary?

```
int old_wrap = state->wrap;
state->wrap = 0;
ret = <updatewindow>;
state->wrap = old_wrap;
```

KungFuJesus:
Ahh, confusingly your branch follows a much earlier revision so that test passes.  I'm going to need to cherry pick your commit onto develop.

nmoinvaz (approved):
LGTM

KungFuJesus (approved):

Dead2:
Nice catch @joshtriplett, and awesome that you contribute a test for it 👍
I'll postpone merging the test until the fix is in develop, so other PRs can still go in, shouldn't be too long.

Dead2 (approved):

Dead2:
@joshtriplett Could you please rebase on develop, so we can see whether the fix and test light up all green?

joshtriplett:
Done.

PR #1259, opened 2022-04-27
Fixed regression introduced by inlining CRC + copy

KungFuJesus:
Pretty much every time updatewindow has been called, implicitly a
checksum was performed unless on s/390 or state->wrap & 4 == 0. The
inflateSetDictionary function instead separately calls this checksum
before invoking update window and checks the checksum to see if it
matches the initial checksum (a property that happens from parsing the
DICTID section of the headers).  Because of this, we need to temporarily
disable the checksum in updatewindow by setting the second wrap bit to
0 and then switching it back on when done.

(see: https://github.com/zlib-ng/zlib-ng/pull/1258)

### DIFF
--- a/inflate.c
+++ b/inflate.c
@@ -15,7 +15,7 @@
 
 /* function prototypes */
 static int inflateStateCheck(PREFIX3(stream) *strm);
-static int updatewindow(PREFIX3(stream) *strm, const uint8_t *end, uint32_t copy);
+static int updatewindow(PREFIX3(stream) *strm, const uint8_t *end, uint32_t len, int32_t cksum);
 static uint32_t syncsearch(uint32_t *have, const unsigned char *buf, uint32_t len);
 
 static inline void inf_chksum_cpy(PREFIX3(stream) *strm, uint8_t *dst,
@@ -237,7 +237,7 @@ int Z_INTERNAL inflate_ensure_window(struct inflate_state *state) {
    output will fall in the output data, making match copies simpler and faster.
    The advantage may be dependent on the size of the processor's data caches.
  */
-static int32_t updatewindow(PREFIX3(stream) *strm, const uint8_t *end, uint32_t copy) {
+static int32_t updatewindow(PREFIX3(stream) *strm, const uint8_t *end, uint32_t len, int32_t cksum) {
     struct inflate_state *state;
     uint32_t dist;
 
@@ -245,16 +245,16 @@ static int32_t updatewindow(PREFIX3(stream) *strm, const uint8_t *end, uint32_t 
 
     if (inflate_ensure_window(state)) return 1;
 
-    /* copy state->wsize or less output bytes into the circular window */
-    if (copy >= state->wsize) {
+    /* len state->wsize or less output bytes into the circular window */
+    if (len >= state->wsize) {
         /* Only do this if the caller specifies to checksum bytes AND the platform requires
          * it (s/390 being the primary exception to this. Also, for now, do the adler checksums
          * if not a gzip based header. The inline adler checksums will come in the near future,
          * possibly the next commit */
-        if (INFLATE_NEED_CHECKSUM(strm) && (state->wrap & 4)) {
+        if (INFLATE_NEED_CHECKSUM(strm) && cksum) {
             /* We have to split the checksum over non-copied and copied bytes */
-            if (copy > state->wsize)
-                inf_chksum(strm, end - copy, copy - state->wsize);
+            if (len > state->wsize)
+                inf_chksum(strm, end - len, len - state->wsize);
             inf_chksum_cpy(strm, state->window, end - state->wsize, state->wsize);
         } else {
             memcpy(state->window, end - state->wsize, state->wsize);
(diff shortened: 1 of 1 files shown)


inflate.c
```diff
@@ -1140,7 +1140,10 @@ int32_t Z_EXPORT PREFIX(inflateSetDictionary)(PREFIX3(stream) *strm, const uint8
 
     /* copy dictionary to window using updatewindow(), which will amend the
        existing dictionary if appropriate */
+    int wrap_val = state->wrap; // Temporarily save to disable checksum in this function
+    state->wrap = 0;
```
mtl1979: This contradicts the commit message... Essentially you should AND `state->wrap` with bit-inverse of `4` to just clear the relevant bit.

nmoinvaz:
Why not just have a checksum override variable for updatewindow. It is an internal function anyways.

inflate.c
KungFuJesus: We could but we know the control flow here doesn't check state->wrap for anything other than this and it's simpler, less error prone, and faster to wholesale copy the whole value and write it back.

KungFuJesus:
> Why not just have a checksum override variable for updatewindow. It is an internal function anyways.

I suppose we could, but that does add yet another variable to check every time.  Honestly if everything didn't set both state->check and state->adler at the same time all the time we probably could check state->adler with state->check after calling update window, instead, but I don't know if that would work in other parts of the code.

inflate.c
mtl1979: Maybe then update the commit message to rephrase the conflicting part. Saves from adding yet another comment here explaining why clearing whole variable is better than clearing just one bit.

`by setting the second wrap bit to 0 and then switching it back on when done.`

KungFuJesus:
Maybe a possibly not crazy way to do it:
Copy state->check into dictid.  Set state->adler to INIT_ADLER_VALUE, run update window, check dictId against state->adler?  Or, does that not work because the custom dictionary can appear at random portions of the inflation stream?

inflate.c
KungFuJesus: Sure, will fix.

nmoinvaz:
You could just add another variable called `copy` and then just call `updatewindow(var1, var2, var3, state->check & 4)`. That way it is only evaluated once. The existing `copy` variable should be renamed `len`.

KungFuJesus:
> You could just add another variable called `copy` and then just call `updatewindow(var1, var2, var3, state->check & 4)`. That way it is only evaluated once. The existing `copy` variable should be renamed `len`.

Yeah, that's a little better for efficiency.  I'm still a bit bothered that the normal updatewindow workflow doesn't quite work here.  It probably could be made to work that way but it seems like we always set state->check to state->adler simultaneously.  Does anyone know the reason for this?

mtl1979:
@KungFuJesus My copy of the source tree doesn't even have `state->adler`, so I guess there is no reason whatsoever ;)

KungFuJesus:
> @KungFuJesus My copy of the source tree doesn't even have `state->adler`, so I guess there is no reason whatsoever ;)

err, strm->adler

mtl1979:
> > @KungFuJesus My copy of the source tree doesn't even have `state->adler`, so I guess there is no reason whatsoever ;)
> 
> err, strm->adler

The dictionary check is the only one to use only `state->check`, otherwise the two variables are used at the same time, synchronizing the values... Obviously `state->check` is not supposed to be touched by any external code.

KungFuJesus:
I mean at the moment there aren't any really compelling efficiency reasons for this (except maybe in the case that there's a custom dictionary with a gzip file, perhaps?), but if we didn't always set state->check to be strm->adler and only set it when parsing the headers for the "case DICTID" bit, we could instead do:
```C
      ret = updatewindow(strm, dictionary + dictLength, dictLength);
      if (ret) {
          state->mode = MEM;
          return Z_MEM_ERROR;
      }

      /* check for correct dictionary identifier */
      if (state->mode == DICT) {
          if (strm->adler != state->check)
              return Z_DATA_ERROR;
      }

      strm->adler = INIT_ADLER_VALUE;
      state->havedict = 1;
      Tracev((stderr, "inflate:   dictionary set\n"));
      return Z_OK;
```
Of course this assumes that the dictionary comes before anything is supposed to be checksummed.  I'm not sure if that's always the case or not.

mtl1979:
@KungFuJesus I don't think efficiency has been the most important factor here... It's more about that `strm` is externally visible, so it can be corrupted by third-party code.

KungFuJesus:
Ahh, so it's also used as an interface for the caller to see the running checksum for a stream that's partially been decompressed.  I guess that makes some sense.  What's in here now I think will do it, then.

mtl1979:
@KungFuJesus Sometimes the "container" has own checksum, so the zlib stream can be chunked... Usually the zlib checksum is at the end of the stream, not between all blocks.

nmoinvaz (changes requested):
I still think we should add extra parameter to `updatewindow` rather than storing the wrap value in a temporary variable and then restoring it. Seems kind of hacky.

KungFuJesus:
> I still think we should add extra parameter to `updatewindow` rather than storing the wrap value in a temporary variable and then restoring it. Seems kind of hacky.

Hmm I made that change last night, I must not have done another force push.

KungFuJesus:
@Dead2 this fix is ready to go.

Dead2 (approved):

mtl1979 (approved):
Makes updatewindow() slightly cleaner by avoiding repeated `state->wrap & 4`.

inflate.c
```diff
@@ -237,24 +237,24 @@ int Z_INTERNAL inflate_ensure_window(struct inflate_state *state) {
    output will fall in the output data, making match copies simpler and faster.
    The advantage may be dependent on the size of the processor's data caches.
  */
-static int32_t updatewindow(PREFIX3(stream) *strm, const uint8_t *end, uint32_t copy) {
+static int32_t updatewindow(PREFIX3(stream) *strm, const uint8_t *end, uint32_t len, int copy) {
```
nmoinvaz: I think perhaps the `copy` variable should be renamed `checksum`. I think that's right.. it is turning on and off checksum with copy.. the else statements still copy..

inflate.c
```diff
@@ -1070,7 +1070,7 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
     if (INFLATE_NEED_UPDATEWINDOW(strm) &&
             (state->wsize || (out != strm->avail_out && state->mode < BAD &&
                  (state->mode < CHECK || flush != Z_FINISH)))) {
-        if (updatewindow(strm, strm->next_out, out - strm->avail_out)) {
+        if (updatewindow(strm, strm->next_out, out - strm->avail_out, state->wrap & 4)) {
```
nmoinvaz, line 1074: A comment above this might be like `update sliding window with checksum only if using gzip or adler32`

inflate.c
```diff
@@ -15,7 +15,7 @@
 
 /* function prototypes */
 static int inflateStateCheck(PREFIX3(stream) *strm);
-static int updatewindow(PREFIX3(stream) *strm, const uint8_t *end, uint32_t copy);
+static int updatewindow(PREFIX3(stream) *strm, const uint8_t *end, uint32_t len, int copy);
```
nmoinvaz: It would be good to use fixed width type `int32_t`.

nmoinvaz (changes requested):

inflate.c
KungFuJesus: Sure.  I had just been matching the type defined in inflate_state.

inflate.c
KungFuJesus: Isn't that all cases?  With the except of maybe raw mode?

inflate.c
KungFuJesus: Ah yes, you are correct.

PR #1261, opened 2022-04-27
Fixed MSVC warning about unknown option for AVX512 flag.

nmoinvaz:
```
cl : command line warning D9002: ignoring unknown option '/ARCH:AVX512'
```


KungFuJesus:
Man, Microsoft is case insensitive in all places until you don't expect them to be.  If that silences the warning then that's definitely working and what we want.  This is one of the few times I've seen Microsoft actually care about case in arguments.  Good catch.

KungFuJesus:
Is there an equivalent bit in the nmake build that needs this?

nmoinvaz:
@KungFuJesus it looks like we don't do AVX512 source files in nmake.

KungFuJesus:
Hmm, should we?  It's probably another PR, if so.

KungFuJesus (approved):

nmoinvaz:
I have always been in favor of getting rid of _nmake_ altogether. In part because it is outdated often. It has gotten better with the CI checks that I added. But I haven't been able to get it removed completely because it has been opposed several times. So it's up to you.

KungFuJesus:
I think there's a lot of extra churn trying to maintain both a configure script and cmake, but I could see those that don't want to install cmake just to build something. It's understandable for systems that can't easily get new versions of it installed I guess.

mtl1979:
Visual C++ 2019 was the first version fully supported by cmake... So there is quite good reason to support nmake as long as we support 32-bit architectures. After we only support 64-bit architectures, we can set Visual C++ 2022 as minimum.

My 2 years old development PC doesn't have new enough cmake, so that pretty much tells how bleeding edge everything has to be for cmake to be useful.

Funny thing is that my 20 years old PowerPC, which is only 32-bit, has more recent cmake. But it runs Debian, while my development PC now has Ubuntu, FreeBSD, NetBSD and Windows 10.

nmoinvaz:
> Visual C++ 2019 was the first version fully supported by cmake...

CMake was available on Windows prior to it being included in Visual Studio 2019.

nmoinvaz:
> So there is quite good reason to support nmake as long as we support 32-bit architectures. After we only support 64-bit architectures, we can set Visual C++ 2022 as minimum.

Only the Visual Studio 2022 IDE will be 64-bit, but it will still be able to compile 32-bit apps.
https://docs.microsoft.com/en-us/visualstudio/productinfo/vs-roadmap

mtl1979:
> > Visual C++ 2019 was the first version fully supported by cmake...
> 
> CMake was available on Windows prior to it being included in Visual Studio 2019.

Yes it was, but the support was limited. 2017 was first to include cmake. Microsoft also use custom version of cmake for Visual Studio.

mtl1979:
> > So there is quite good reason to support nmake as long as we support 32-bit architectures. After we only support 64-bit architectures, we can set Visual C++ 2022 as minimum.
> 
> Only the Visual Studio 2022 IDE will be 64-bit, but it will still be able to compile 32-bit apps.
> https://docs.microsoft.com/en-us/visualstudio/productinfo/vs-roadmap

My laptop has 32-bit Windows ;)

Dead2 (approved):

Issue #1262, opened 2022-05-02
crc32_combine failure in ZLIB_COMPAT mode


pmqs:
This is a follow-up to https://github.com/zlib-ng/zlib-ng/issues/1234 which is concerned with getting zlib-ng working with my Perl bindings for zlib, namely [Compress::Raw::Zlib](https://github.com/pmqs/Compress-Raw-Zlib).

I've been playing with getting zlib-ng built in  `ZLIB_COMPAT`  working with a development copy of  my code (downstream issue for this work is https://github.com/pmqs/Compress-Raw-Zlib/issues/9).

I started by building zlib-ng from source  with `ZLIB_COMPAT` enabled. I think that is all I need to do to get a zlib-compatible library. Please shout if I need something else.

```
cmake -GNinja -DZLIB_COMPAT=ON  .. 
```

That built fine. 

Next step is to get  [Compress::Raw::Zlib](https://github.com/pmqs/Compress-Raw-Zlib) to use the zlib-ng library. That can be done as follows

```
$ ZLIB_LIB=/media/paul/Linux-Shared/base/git/zlib-ng/build ZLIB_INCLUDE=/media/paul/Linux-Shared/base/git/zlib-ng/build BUILD_ZLIB=0  perl5.32.1 Makefile.PL
Parsing config.in...
Auto Detect Gzip OS Code..
Setting Gzip OS Code to 3 [Unix/Default]
Looks Good.
Up/Downgrade not needed.
Generating a Unix-style Makefile
Writing Makefile for Compress::Raw::Zlib
Writing MYMETA.yml and MYMETA.json
```

First issue are some compiler warnings when I build the Perl bindings to zlib
```
$ make
Skip blib/lib/Compress/Raw/Zlib.pm (unchanged)
Running Mkbootstrap for Zlib ()
chmod 644 "Zlib.bs"
"/usr/bin/perl5.32.1" -MExtUtils::Command::MM -e 'cp_nonempty' -- Zlib.bs blib/arch/auto/Compress/Raw/Zlib/Zlib.bs 644
[...]

PR #1263, opened 2022-05-03
Disable redirection to 64-bit function variants when Z_SOLO is defined

mtl1979:
See #1262.


mtl1979:
@nmoinvaz @Dead2 

Only failing test is due to DNS error...

pmqs:
@mtl1979 this change fixed the issues I was seeing in https://github.com/zlib-ng/zlib-ng/issues/1262

PR #1265, opened 2022-05-03
Fixed warning about strict prototypes for cpu_check_features.

nmoinvaz:
This warning happens for every time `cpu_features.h` is included:
```c
cpu_features.h:22:31: warning: this function declaration is not a prototype [-Wstrict-prototypes]
extern void cpu_check_features();
                              ^
                               void
```
In cpu_features.c it is defined as:
```c
Z_INTERNAL void cpu_check_features(void)
```

### DIFF
--- a/cpu_features.h
+++ b/cpu_features.h
@@ -19,7 +19,7 @@
 #  include "arch/s390/s390_features.h"
 #endif
 
-extern void cpu_check_features();
+extern void cpu_check_features(void);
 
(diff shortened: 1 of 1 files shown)


Dead2 (approved):

mtl1979 (approved):
I don't like "void" as sole parameter for function, but this is about strict compliance and C is a lot stricter than C++ when it comes to prototypes of functions.

PR #1267, opened 2022-05-05
Fix incorrect character case of the CMAKE_INTERPROCEDURAL_OPTIMIZATION_* names

phprus:
The names ``CMAKE_INTERPROCEDURAL_OPTIMIZATION_*`` must be uppercase.
Variable names in CMake are case-sensitive.

Bug in PR #1233.

LTO will not be disabled in zlib-ng only if in the parent project any of the ``CMAKE_INTERPROCEDURAL_OPTIMIZATION_{RELEASE,RELWITHDEBINFO, * }`` variables are defined to ``ON``.

If the parent project defined only ``CMAKE_INTERPROCEDURAL_OPTIMIZATION`` everything works fine.


Dead2 (approved):
I don't fully understand why this fix is needed, but it looks good to me.

phprus:
@Dead2

For example, parent project:
```
# Enable LTO only for Release build.
set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELEASE ON)

add_subdirectory(zlib-ng)
```

In zlib-ng, the CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELEASE variable will not be overwritten (PR #1233) and CMake controlled LTO will not be disabled.

When developing PR #1233, I missed the point that the variable names in CMake are case-sensitive, and in the CMAKE_CONFIGURATION_TYPES list, configuration names are not in upper case :(

Dead2:
@phprus Ah, now it makes sense, thanks for explaining it. 👍

Issue #1264, opened 2022-05-03
AppleClang 8: error: initializing '__m256i' (vector of 4 'long long' values) with an expression of incompatible type 'int'


seanm:
I library I use ([ITK](https://github.com/InsightSoftwareConsortium/ITK)) recently updated to [zlib-ng 2022-04-27 (d41f8ead)](https://github.com/InsightSoftwareConsortium/ITK/pull/3275#) (*from* what I'm not sure) and it now fails to build on macOS 10.11 with AppleClang 8, with:

```
/Users/builder/external/ITK/Modules/ThirdParty/ZLIB/src/itkzlib-ng/arch/x86/adler32_avx2.c:55:13: error: initializing '__m256i' (vector of 4 'long long' values) with an expression
      of incompatible type 'int'
    __m256i vs1 = _mm256_zextsi128_si256(_mm_cvtsi32_si128(adler));
            ^     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/builder/external/ITK/Modules/ThirdParty/ZLIB/src/itkzlib-ng/arch/x86/adler32_avx2.c:56:13: error: initializing '__m256i' (vector of 4 'long long' values) with an expression
      of incompatible type 'int'
    __m256i vs2 = _mm256_zextsi128_si256(_mm_cvtsi32_si128(sum2));
            ^     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

See: https://open.cdash.org/viewBuildError.php?buildid=7889929

Perhaps `X86_AVX2_ADLER32` should not be set for this compiler and/or macOS deployment target?

mtl1979:
This seems to be missing the initial error, which is that those intrinsic functions have no prototype defined. You should try to find out if it is because the correct header is not included, or if the compiler doesn't support those intrinsics, or that those intrinsics have different names.

seanm:
Yup, there's just no declaration of `_mm256_zextsi128_si256` in Xcode 8:

```
$ cat test.c
#include <x86intrin.h> 

__m256i zero_upper_part(__m256i a) {
   return _mm256_zextsi128_si256(_mm256_castsi256_si128(a));
}

$ xcrun clang -mavx2 -fsyntax-only test.c
[...]

PR #1271, opened 2022-05-09
Fixed issue #1264: Use fallback for _mm256_zextsi128_si256 on Xcode <…

seanm:
… 9.3

### DIFF
--- a/fallback_builtins.h
+++ b/fallback_builtins.h
@@ -46,11 +46,14 @@ static __forceinline unsigned long long __builtin_ctzll(uint64_t value) {
 #endif // Microsoft AMD64/IA64/x86/ARM/ARM64 test
 #endif // _MSC_VER & !clang
 
-/* Unfortunately GCC didn't support these things until version 10 */
+/* Unfortunately GCC didn't support these things until version 10.
+ * Similarly, AppleClang didn't support them in Xcode 9.2 but did in 9.3.
+ */
 #ifdef __AVX2__
 #include <immintrin.h>
 
-#if (!defined(__clang__) && defined(__GNUC__) && __GNUC__ < 10)
+#if (!defined(__clang__) && defined(__GNUC__) && __GNUC__ < 10) \
+    || (defined(__apple_build_version__) && __apple_build_version__ < 9020039)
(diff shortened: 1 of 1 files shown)


fallback_builtins.h
```diff
@@ -46,11 +46,13 @@ static __forceinline unsigned long long __builtin_ctzll(uint64_t value) {
...
 
-/* Unfortunately GCC didn't support these things until version 10 */
+/* Unfortunately GCC didn't support these things until version 10.
+ * Similarly, AppleClang didn't support them in Xcode 9.2 but did in 9.3.
+ */
 #ifdef __AVX2__
 #include <immintrin.h>
 
-#if (!defined(__clang__) && defined(__GNUC__) && __GNUC__ < 10)
+#if (!defined(__clang__) && defined(__GNUC__) && __GNUC__ < 10) || (defined(__apple_build_version__) && __apple_build_version__ < 9020039)
```
nmoinvaz: I am wondering if after the `||` there should be a new line break and the next line align the `(`.

Dead2: Personally I like to move the `||` and have that start the next line, since I think that makes it more readable.
Like this:
```
#if (!defined(__clang__) && defined(__GNUC__) && __GNUC__ < 10)
    || (defined(__apple_build_version__) && __apple_build_version__ < 9020039)
```

phprus: Why are you checking the XCode version instead of the AppleClang version?

mtl1979: The line is clearly longer than 120 characters, so should be split.

@phprus Because checking just AppleClang version would match stock Clang too...

seanm: I've never participated in this project, and I don't know your coding standards, and I don't know which preference here is authoritative.  You don't have some kind of clang-tidy-auto-format-on-merge script?  I can modify the commit formatting, but I'd rather have to do it only once. :)  Which way must it be?

mtl1979: If everyone says different things, it's safe bet to do what @Dead2 says as he's the maintainer...

nmoinvaz (approved):
Looks ok to me.

Dead2 (approved):

seanm:
Thanks all!

PR #1273, opened 2022-05-12
Make directory for output files

phprus:
If zlib-ng is added to the CMake project via ``add_subdirectory`` then the ``${CMAKE_CURRENT_BINARY_DIR}/Testing/Temporary`` directory may not exist when the tests are run.

CMake creates a ``Testing/Temporary`` directory only for the top project (``${CMAKE_BINARY_DIR}/Testing/Temporary``).


nmoinvaz (approved):
Looks ok to me.

Dead2 (approved):

PR #1274, opened 2022-05-13
fix off-by-one bug in crc32_acle.c

landfillbaby:
would be doing unaligned memory reads on very small buffers

### DIFF
--- a/arch/arm/crc32_acle.c
+++ b/arch/arm/crc32_acle.c
@@ -22,7 +22,7 @@ uint32_t crc32_acle(uint32_t crc, const unsigned char *buf, uint64_t len) {
         len--;
     }
 
-    if ((len > sizeof(uint16_t)) && ((ptrdiff_t)buf & sizeof(uint16_t))) {
+    if ((len >= sizeof(uint16_t)) && ((ptrdiff_t)buf & sizeof(uint16_t))) {
         buf2 = (const uint16_t *) buf;
         c = __crc32h(c, *buf2++);
         len -= sizeof(uint16_t);
@@ -32,7 +32,7 @@ uint32_t crc32_acle(uint32_t crc, const unsigned char *buf, uint64_t len) {
     }
 
 #if defined(__aarch64__)
-    if ((len > sizeof(uint32_t)) && ((ptrdiff_t)buf & sizeof(uint32_t))) {
+    if ((len >= sizeof(uint32_t)) && ((ptrdiff_t)buf & sizeof(uint32_t))) {
         c = __crc32w(c, *buf4++);
         len -= sizeof(uint32_t);
     }
--- a/arch/arm/crc32_acle.c
+++ b/arch/arm/crc32_acle.c
@@ -37,6 +37,11 @@ uint32_t crc32_acle(uint32_t crc, const unsigned char *buf, uint64_t len) {
         len -= sizeof(uint32_t);
     }
 
+    if (len == 0) {
+        c = ~c;
+        return c;
+    }
+
     const uint64_t *buf8 = (const uint64_t *) buf4;
 
     while (len >= sizeof(uint64_t)) {
@@ -60,6 +65,11 @@ uint32_t crc32_acle(uint32_t crc, const unsigned char *buf, uint64_t len) {
 
     buf = (const unsigned char *) buf2;
 #else /* __aarch64__ */
+
+    if (len == 0) {
+        c = ~c;
+        return c;
+    }
 
     while (len >= sizeof(uint32_t)) {
         c = __crc32w(c, *buf4++);
--- a/arch/arm/crc32_acle.c
+++ b/arch/arm/crc32_acle.c
@@ -39,16 +39,6 @@ uint32_t crc32_acle(uint32_t crc, const unsigned char *buf, uint64_t len) {
 
     const uint64_t *buf8 = (const uint64_t *) buf4;
 
-#ifdef UNROLL_MORE
-    while (len >= 4 * sizeof(uint64_t)) {
-        c = __crc32d(c, *buf8++);
-        c = __crc32d(c, *buf8++);
-        c = __crc32d(c, *buf8++);
-        c = __crc32d(c, *buf8++);
-        len -= 4 * sizeof(uint64_t);
-    }
-#endif
-
     while (len >= sizeof(uint64_t)) {
         c = __crc32d(c, *buf8++);
         len -= sizeof(uint64_t);
@@ -70,20 +60,6 @@ uint32_t crc32_acle(uint32_t crc, const unsigned char *buf, uint64_t len) {
 
     buf = (const unsigned char *) buf2;
 #else /* __aarch64__ */
-
-#  ifdef UNROLL_MORE
-    while (len >= 8 * sizeof(uint32_t)) {
-        c = __crc32w(c, *buf4++);
-        c = __crc32w(c, *buf4++);
-        c = __crc32w(c, *buf4++);
-        c = __crc32w(c, *buf4++);
-        c = __crc32w(c, *buf4++);
-        c = __crc32w(c, *buf4++);
-        c = __crc32w(c, *buf4++);
-        c = __crc32w(c, *buf4++);
-        len -= 8 * sizeof(uint32_t);
-    }
-#  endif
 
     while (len >= sizeof(uint32_t)) {
         c = __crc32w(c, *buf4++);


mtl1979:
I'm not sure the fix is necessary... First part of the code just tries to align the pointer if buffer is large enough to need or benefit from aligning... Last part of the code handles small buffers.

landfillbaby:
if you think about it though, the last part operates in reverse, so in the case of a small buffer misaligned by 2 bytes, the 4 byte read will be unaligned

landfillbaby:
decided to simplify it so the compiler understands it better. if you really want to you can just cherry-pick the 1st commit

mtl1979:
> if you think about it though, the last part operates in reverse, so in the case of a small buffer misaligned by 2 bytes, the 4 byte read will be unaligned

There wouldn't be 4 byte read when the remaining buffer length is only 2 (16 bits).

landfillbaby:
alright, make it consistent in the opposite way then, replace `if (len && ((ptrdiff_t)buf & 1))` with `if ((len > 1) && ((ptrdiff_t)buf & 1))`

KungFuJesus:
I'd argue if you're going to modify this function, you may as well remove the "#ifdef UNROLL_MORE" section, as this really didn't have any hope of being faster.  The only thing it avoided was potential some loop condition checking overhead more often, but there's no opportunity for superscalar parallelism here.

mtl1979:
> alright, make it consistent in the opposite way then, replace `if (len && ((ptrdiff_t)buf & 1))` with `if ((len > 1) && ((ptrdiff_t)buf & 1))`

That is supposed to test that `len` is non-zero, not that it's more than 1. It's valid to call crc functions with zero length.

We can have a lot of cases when we first need to crc one byte and then crc more bytes and then 1 byte again... It makes sense to make 1 byte crc as fast as possible without slowing down the other cases with extra checks.

This is all about optimizing which branch is more likely to be taken. With one byte crc, it is equally likely that the last bit of the pointer is 0 or 1. But in this case testing last bit of 1 first benefits the other lengths more.

Dead2:
This PR contains two very different changes, one is the described fix from the PR title, and another is a bigger rewrite. Please split these into two different PRs so they can be discussed, tested and reviewed separately.

Regarding the fix, I am still unclear as to whether this is a bug or not, and the lack of any comment stating the original intentions does not help ascertain what is in fact correct. Since you did a rewrite commit, please include some comments that help establish the intentions of the code.

Dead2:
This PR contains two very different changes, one is the described fix from the PR title, and another is a bigger rewrite. Please split these into two different PRs so they can be discussed, tested and reviewed separately.

Regarding the fix, I am still unclear as to whether this is a bug or not, and the lack of any comment stating the original intentions does not help ascertain what is in fact correct. Since you did a rewrite commit, please include some comments that help establish the intentions of the code.

landfillbaby:
pushed it back

nmoinvaz (approved):
Looks ok to me.

mtl1979:
I still say this is not a bug... There can't be unaligned access as I explained above.

KungFuJesus:
The diff that's there now (>= len(type)) _is_ fixing a bug (albeit a minor one).  It allows the CRC instructions to operate on the largest size possible for the first handful to reach 64 bit alignment rather than terminating early and moving down to the next size before it has to.  In some cases I think this could lead to unaligned 64 bit strides, as you could remove modulo 1 byte but not the modulo 2 bytes or 4 bytes that were needed prior to reach alignment.

mtl1979:
@KungFuJesus It will never reach larger sizes as the len is too small by then... There is minimal speed gain possible as we don't test if `len == 0` after each step...

KungFuJesus:
> @KungFuJesus It will never reach larger sizes as the len is too small by then... There is minimal speed gain possible as we don't test if `len == 0` after each step...

https://github.com/zlib-ng/zlib-ng/blob/develop/arch/arm/crc32_acle.c#L26

I don't think so, this is at the beginning, when subtracting off the modulus so that things can be aligned to the largest possible crc32 operand.  So, say that you have an address that happens to be a multiple of 3 bytes (and also happens to be that length):

This:
```c
if (len && ((ptrdiff_t)buf & 1)) {
        c = __crc32b(c, *buf++);
        len--;
    }
```

Subtracts off the first byte.  Then this:

```C
if ((len > sizeof(uint16_t)) && ((ptrdiff_t)buf & sizeof(uint16_t))) {
        buf2 = (const uint16_t *) buf;
        c = __crc32h(c, *buf2++);
        len -= sizeof(uint16_t);
        buf4 = (const uint32_t *) buf2;
    } else {
        buf4 = (const uint32_t *) buf;
    }
```

Ends up not running even though it could, and it could do optimally at 2 byte alignment and be finished.  Instead, it falls through all the way to the bottom and does the last 2 bytes.  

Edit: eh well, it ends up being done in the loop body, instead.  But still, the intention of the code seems to read that it was supposed to be done earlier.  Granted, this is a small corner case if len is just barely the size of what's being peeled in the unaligned access and unlikely to be a _huge_ issue.  

I don't think >= vs > is a meaningful difference (there's a scalar instruction for each and I think they complete in the same number of cycles).

https://www.cs.princeton.edu/courses/archive/spr19/cos217/reading/ArmInstructionSetOverview.pdf (page 67)

mtl1979:
@KungFuJesus It doesn't need to do further aligning if `len == 3` at the top of function... If it is anything larger, then the test already returns `true` and everything still works...

KungFuJesus:
> @KungFuJesus It doesn't need to do further aligning if `len == 3` at the top of function... If it is anything larger, then the test already returns `true` and everything still works...

It still works, yes, but it does mean it has to be handled later in the loop that assumes access is aligned.  Granted, you weren't going to get aligned access with a 3 byte aligned 3 byte width buffer, anyway, but the intention of the beginning of the code reads like it's trying to rip off the unaligned access early and terminate there if possible.  Whether it's handled at the top or later at the bottom probably doesn't matter much other than the fact that control flow has to jump further than possibly necessary.

mtl1979:
@KungFuJesus That `terminating part` isn't really implemented and I'm not sure if I will implement it... There will always be unnecessary branches, but that's what by definition makes unaligned versions slower than aligned versions...

arch/arm/crc32_acle.c
```diff
@@ -32,23 +32,13 @@ uint32_t crc32_acle(uint32_t crc, const unsigned char *buf, uint64_t len) {
     }
 
 #if defined(__aarch64__)
-    if ((len > sizeof(uint32_t)) && ((ptrdiff_t)buf & sizeof(uint32_t))) {
+    if ((len >= sizeof(uint32_t)) && ((ptrdiff_t)buf & sizeof(uint32_t))) {
         c = __crc32w(c, *buf4++);
         len -= sizeof(uint32_t);
     }
 
     const uint64_t *buf8 = (const uint64_t *) buf4;
```
mtl1979, line 45: If `len == 0` before this line, we might want to do early exit.

KungFuJesus:
> @KungFuJesus That `terminating part` isn't really implemented and I'm not sure if I will implement it... There will always be unnecessary branches, but that's what by definition makes unaligned versions slower than aligned versions...

Sure it will, by the very fact that you subtracted from len.  It means that by the time you get to the loop portions of the code len is already zero and it will fall through to the bottom.  You could also explicitly return with an extra branch (I'm guessing that's what you meant), I suppose, preventing further evaluations for len at each loop sequence.  Whether or not that makes an observable difference is up the air (likely not, unless you feed a lot of unaligned small buffers into the CRC function).  

The fact that this works regardless of the changes looks incidental rather than intentional, but you did write it so you'd probably know (assuming you remember, anyway).

arch/arm/crc32_acle.c
```diff
@@ -71,20 +61,6 @@ uint32_t crc32_acle(uint32_t crc, const unsigned char *buf, uint64_t len) {
     buf = (const unsigned char *) buf2;
 #else /* __aarch64__ */
```
mtl1979, line 68: If `len == 0` here, we might want to do early exit.

arch/arm/crc32_acle.c
```diff
@@ -32,22 +32,17 @@ uint32_t crc32_acle(uint32_t crc, const unsigned char *buf, uint64_t len) {
...
-    const uint64_t *buf8 = (const uint64_t *) buf4;
-
-#ifdef UNROLL_MORE
-    while (len >= 4 * sizeof(uint64_t)) {
-        c = __crc32d(c, *buf8++);
-        c = __crc32d(c, *buf8++);
-        c = __crc32d(c, *buf8++);
-        c = __crc32d(c, *buf8++);
-        len -= 4 * sizeof(uint64_t);
+    if(len == 0) {
```
nmoinvaz: Missing space after `if`

arch/arm/crc32_acle.c
```diff
@@ -71,19 +66,10 @@ uint32_t crc32_acle(uint32_t crc, const unsigned char *buf, uint64_t len) {
...
-        c = __crc32w(c, *buf4++);
-        c = __crc32w(c, *buf4++);
-        c = __crc32w(c, *buf4++);
-        c = __crc32w(c, *buf4++);
-        c = __crc32w(c, *buf4++);
-        c = __crc32w(c, *buf4++);
-        c = __crc32w(c, *buf4++);
-        c = __crc32w(c, *buf4++);
-        len -= 8 * sizeof(uint32_t);
+    if(len == 0) {
```
nmoinvaz: Missing space after `if`

Dead2 (approved):
LGTM

mtl1979:
I don't think there is anything else to do than to reword the commit message of the first commit and possibly squash the commits... This is more a logic change than a bug fix by definition... It should slightly improve speed for short lengths, but the gain might be marginal... However it does improve readability of the code.

KungFuJesus:
One would hope the compiler is making the early returns a single label and having them jump to a common termination but one _could_ accomplish such a thing with gotos.  Perhaps for the compliment + return though that's a bit overkill.

Dead2:
> I don't think there is anything else to do than to reword the commit message of the first commit and possibly squash the commits... This is more a logic change than a bug fix by definition... It should slightly improve speed for short lengths, but the gain might be marginal... However it does improve readability of the code.

@landfillbaby Would you be willing to reword the commit message for the first commit?

landfillbaby:
how's that as a commit message

PR #1275, opened 2022-05-18
Remove references to TLS code

pps83:
This TLS code was pure nonsense,  added by 16b6cda67be0038327b416e5fa5c7243ed138e21 with zero explanation why at all this was needed in the first place.
There is no library out there that uses similar function pointers and also uses TLS for these pointer (example: ffmpeg/libav/libavcodec).
My wild guess the code was added was to "handle" possibility to have different cores (like hi-pef vs low-power cores). In case if there is any doubt that the code wasn't needed, then obviously there is a problem with the TLS code if execution moving between cores, which certainly does happen all the time.

### DIFF
--- a/functable.c
+++ b/functable.c
@@ -13,7 +13,7 @@
 
 #include "cpu_features.h"
 
-Z_INTERNAL Z_TLS struct functable_s functable;
+Z_INTERNAL struct functable_s functable;
 
 /* stub functions */
 Z_INTERNAL uint32_t update_hash_stub(deflate_state *const s, uint32_t h, uint32_t val) {
@@ -427,7 +427,7 @@ Z_INTERNAL uint32_t compare256_stub(const uint8_t *src0, const uint8_t *src1) {
 }
 
 /* functable init */
-Z_INTERNAL Z_TLS struct functable_s functable = {
+Z_INTERNAL struct functable_s functable = {
     adler32_stub,
     crc32_stub,
     crc32_fold_reset_stub,
--- a/functable.h
+++ b/functable.h
@@ -30,6 +30,6 @@ struct functable_s {
     uint32_t (* update_hash)        (deflate_state *const s, uint32_t h, uint32_t val);
 };
 
-Z_INTERNAL extern Z_TLS struct functable_s functable;
+Z_INTERNAL extern struct functable_s functable;
 
 #endif
--- a/zbuild.h
+++ b/zbuild.h
@@ -25,20 +25,6 @@
 #  endif
 #endif
 
-/* Determine compiler support for TLS */
-#ifndef Z_TLS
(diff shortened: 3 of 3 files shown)


mtl1979:
Wild guess is obviously no valid reason to suggest removing something... There is no guarantee that another thread doesn't try to call same hook before the stub function finishes and replaces the hook function in functable with correct optimized version.

It has obviously nothing to do with a thread switching from one CPU to another with different instruction set support... In most cases it would crash anyways if the new CPU has less supported instructions...

pps83:
> Wild guess is obviously no valid reason to suggest removing something...

TLS code is bogus. That's valid enough to me. Maybe I misunderstand the reason it's there in the first place? Proper commit message absolutely **had** to explain why it was added as it's probably not something trivial.

> There is no guarantee that another thread doesn't try to call same hook before the stub function finishes and replaces the hook function in functable with correct optimized version.

So, can you please explain what was actual reason for adding TLS there? My second guess is that without TLS there is a potential to crash if a thread tries to run any of the function while another thread is half way done updating the pointer (if updating of the func ptr isn't atomic operation for whatever reason). Once again, proper commit message absolutely **had** to explain why it was added as it's probably not something trivial.

> It has obviously nothing to do with a thread switching from one CPU to another with different instruction set support... In most cases it would crash anyways if the new CPU has less supported instructions...

Yes, that would be incredibly buggy CPU. For general purpose, such cores would have to have identical instruction sets.

If my second guess is correct (again, if commit was properly explaining reason for the change, others wouldn't need make such wild guesses), then, perhaps a better/proper approach is to have some sort of run-once init function that gets executed on first zlib api use? I recall that's the way libavcodec inits its function pointers. The function could as well use locking to init table once than resorting to TLS, which at best adds perf overhead on all calls though such TLS func table.

pps83:
Perhaps, a better way is to add `functable` as a variable to `struct z_stream_s` and init it inside of `inflateInit`/`deflateInit`?

nmoinvaz:
What is the error that you are getting?

mtl1979:
Using locking would require switching to C++, as C as a language doesn't have any locking functions in standard library...
I do agree that proper commit message would avoid the need to go through guessing what was discussed in our private discussion groups.

pps83:
> What is the error that you are getting?

I used an older version of zlib-ng where I had to make some fixes back then to make it work. I tried to rebase and had merge conflicts one of them was related to TLS. Since there was zero info in the commit message and no reviews of the commit on github, I guessed it was perhaps related to different cores on mobiles. This assumption wasn't correct, and the TLS code cannot be simply removed.
I believe that TLS shouldn't be used for the purpose here. Instead, init should be moved to inflateInit/deflateInit. In any case, the PR isn't valid, and should be closed.

PR #1276, opened 2022-05-25
abicheck.sh: update reference versions.

nmoinvaz:
This is a more up-to-date version of #1228.


Dead2:
Would you be able to update the zlib-ng native commit reference as well? Could use the current develop commit for that.

nmoinvaz:
Done

PR #1278, opened 2022-05-27
CMakeLists.txt: fix version in zlib.pc when building statically

ffontaine:
When building statically (i.e. with `BUILD_SHARED_LIBS=OFF`), `ZLIB_FULL_VERSION` is not set resulting in an empty `version` in `zlib.pc` and the following build failure with transmission:

```
checking for ZLIB... configure: error: Package requirements (zlib >= 1.2.3) were not met:

Package dependency requirement 'zlib >= 1.2.3' could not be satisfied.
Package 'zlib' has version '', required version is '>= 1.2.3'
```

Fixes:
 - http://autobuild.buildroot.org/results/b3b882482f517726e5c780ba4c37818bd379df82

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


Dead2 (approved):
LGTM

PR #1281, opened 2022-05-31
[ARM] We need to include NEON headers when testing for -mfpu=neon.

mtl1979:
* If -mfpu is already specified in C_FLAGS, it can disable NEON support.


nmoinvaz (changes requested):
We should have the same check in `configure`. There exists header include for `check_neon_ld4_intrinsics` but not `check_neon_compiler_flag`.

https://github.com/zlib-ng/zlib-ng/blob/856884022cc7202b159ba76c8e89bb1d9f7300d5/configure#L1106-L1114

https://github.com/zlib-ng/zlib-ng/blob/856884022cc7202b159ba76c8e89bb1d9f7300d5/configure#L1120-L1136

nmoinvaz (approved):
LGTM

PR #1286, opened 2022-06-06
Fixed inflate size conversion warning in chunkcopy_safe.

nmoinvaz:
Using `-A Win32 -D WITH_MAINTAINER_WARNINGS=ON`.
```
  inflate_p.h(142,27): warning C4244: 'function': conversion from 'uint64_t' to 'size_t', possible loss of data
```
Here I just move the cast from `uint64_t` to `size_t` to the call to `memcpy`.

### DIFF
--- a/inflate_p.h
+++ b/inflate_p.h
@@ -130,16 +130,16 @@
     } while (0)
 
 /* Behave like chunkcopy, but avoid writing beyond of legal output. */
-static inline uint8_t* chunkcopy_safe(uint8_t *out, uint8_t *from, size_t len, uint8_t *safe) {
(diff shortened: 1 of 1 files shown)


KungFuJesus (approved):
Also looks fine.  Technically the argument to memcpy being casted isn't semantically, however, as established previously, it's impossible to actually copy that much memory on a 32 bit platform, anyway.  You'd overflow the pointer.

nmoinvaz:
Yeah the actual warning was not casting to `size_t` from `llabs`. However, rather than cast it there, I thought it might be best to cast it at `memcpy`. But you are right, it would overflow either way.

Dead2 (approved):

PR #1290, opened 2022-06-08
Add a test for concurrently modifying deflate() input

iii-i:
Hi,

I'm still working on a fix for #1253, but in the meantime I'd like to contribute a test, which it makes sense to have anyway.

Best regards,
Ilya

---

The test simulates what one of the QEMU live migration tests is doing:
increments each buffer byte by 1 while deflate()ing it.

The test tries to produce a race condition and therefore is
probabilistic. The longer it runs, the better are the chances to catch
an issue. The scenario in question is known to be broken on IBM Z
with DFLTCC, and there it is caught in 100ms most of the time. The
run time is therefore set to 1 second in order to balance usability and
reliability.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -22,9 +22,6 @@ if(NOT TARGET GTest::GTest)
     # Prevent overriding the parent project's compiler/linker settings for Windows
     set(gtest_force_shared_crt ON CACHE BOOL
         "Use shared (DLL) run-time lib even when Google Test is built as static lib." FORCE)
-    # Disable pthreads for simplicity
-    set(gtest_disable_pthreads ON CACHE BOOL
-        "Disable uses of pthreads in gtest." FORCE)
 
     # Allow specifying alternative Google test repository
     if(NOT DEFINED GTEST_REPOSITORY)
@@ -99,6 +96,16 @@ if(WITH_SANITIZER STREQUAL "Memory")
 endif()
 
 target_link_libraries(gtest_zlib zlibstatic GTest::GTest)
+
+find_package(Threads)
+if(Threads_FOUND)
+    target_sources(gtest_zlib PRIVATE test_deflate_concurrency.cc)
+    if(UNIX AND NOT APPLE)
+        # On Linux, use a workaround for https://gcc.gnu.org/bugzilla/show_bug.cgi?id=52590
+        target_link_libraries(gtest_zlib -Wl,--whole-archive -lpthread -Wl,--no-whole-archive)
+    endif()
+    target_link_libraries(gtest_zlib Threads::Threads)
+endif()
 
 if(ZLIB_ENABLE_TESTS)
     add_test(NAME gtest_zlib
--- a/test/test_deflate_concurrency.cc
+++ b/test/test_deflate_concurrency.cc
@@ -1,0 +1,170 @@
+/* Test deflate() on concurrently modified next_in.
+ *
+ * Plain zlib does not document that this is supported, but in practice it tolerates this, and QEMU live migration is
+ * known to rely on this. Make sure zlib-ng tolerates this as well.
+ */
+
+#include "zbuild.h"
+#ifdef ZLIB_COMPAT
+#include "zlib.h"
+#else
+#include "zlib-ng.h"
+#endif
+
+#include <gtest/gtest.h>
+
+#include <algorithm>
+#include <atomic>
+#include <cstring>
+#include <thread>
+
+static uint8_t buf[8 * 1024];
+static uint8_t zbuf[4 * 1024];
+static uint8_t tmp[8 * 1024];
+
+/* Thread that increments all bytes in buf by 1. */
+class Mutator {
+    enum class State {
+        PAUSED,
+        RUNNING,
+        STOPPED,
+    };
+
+public:
+    Mutator()
+        : m_state(State::PAUSED), m_target_state(State::PAUSED),
+          m_thread(&Mutator::run, this) {}
+    ~Mutator() {
+        transition(State::STOPPED);
+        m_thread.join();
+    }
+
+    void pause() {
+        transition(State::PAUSED);
+    }
+
+    void resume() {
+        transition(State::RUNNING);
+    }
+
+private:
+    void run() {
+        while (true) {
+            m_state.store(m_target_state);
+            if (m_state == State::PAUSED)
+                continue;
+            if (m_state == State::STOPPED)
+                break;
+            for (uint8_t & i: buf)
+                i++;
+        }
+    }
+
+    void transition(State target_state) {
+        m_target_state = target_state;
+        while (m_state != target_state) {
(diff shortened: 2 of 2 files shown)


test/CMakeLists.txt
```diff
@@ -89,6 +90,9 @@ target_include_directories(gtest_zlib PRIVATE
     ${CMAKE_SOURCE_DIR}
     ${CMAKE_BINARY_DIR})
 
+find_package(Threads REQUIRED)
```
nmoinvaz: It would be good to include `test_deflate_concurrency.cc` only if `Threads` is found.

nmoinvaz (commented):
It is great to see new tests being added to Google Test framework.

test/CMakeLists.txt
iii-i: Done.

iii-i:
The new test fails on s390x, ppc64 and sparc in QEMU. However, I cannot reproduce this neither on real hardware, nor in QEMU. So I've added a temporary commit to this PR, that shows what syscalls QEMU is handling. Maybe this will shed some light on the issue.

phprus:
``pthread`` support is disabled in ``gtest``:
https://github.com/zlib-ng/zlib-ng/blob/201188691a990468ae1c6ebba4e12bd1a62166b0/test/CMakeLists.txt#L25-L27

iii-i:
@phprus thanks, I missed this! I think we should remove it now: gtest will then do `find_package(Threads)`.

I haven't noticed anything suspicious in the strace output, so I'll push the amended commit now.

iii-i:
The failures are caused by https://gcc.gnu.org/bugzilla/show_bug.cgi?id=52590. I've added the workaround proposed there (they actually suggest not to use static linking at all, but I've decided not to touch that part of the test infrastructure).

nmoinvaz:
I wonder if there are any problems compiling on Windows where pthreads library doesn't exist.
https://stackoverflow.com/questions/13218646/cmake-pthread-h-not-found-in-windows

iii-i:
The test code doesn't use pthreads directly (it uses the C++ standard library); and on Windows CMake's FindThreads assumes that threading just works:

```
if(CMAKE_SYSTEM_NAME MATCHES "Windows")
  set(CMAKE_USE_WIN32_THREADS_INIT 1)
  set(Threads_FOUND TRUE)
endif()
```

Dead2 (approved):
LGTM

test/test_deflate_concurrency.cc
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+
+    std::atomic<State> m_state, m_target_state;
+    std::thread m_thread;
+};
+
+TEST(deflate, concurrency) {
+#ifdef S390_DFLTCC_DEFLATE
+    GTEST_SKIP() << "XFAIL S390_DFLTCC_DEFLATE";
```
nmoinvaz: Is `XFAIL` intended?

iii-i: Yes, since it's eXpected to fail. However, come to think of it, this word is normally used when you really let the test fail, which we are not doing here. Let me replace this with the more straightforward "known to be broken with S390_DFLTCC_DEFLATE".

test/CMakeLists.txt
```diff
@@ -83,6 +80,11 @@ if(WITH_GZFILEOP)
     list(APPEND TEST_SRCS test_gzio.cc)
 endif()
 
+find_package(Threads)
```
nmoinvaz: In #1299, I use `target_sources` instead of trying to add to `TEST_SRCS`. It might be worth simplifying the two blocks of changes in this file into one like I have done.

iii-i: Done.

iii-i:
CMake / Ubuntu GCC S390X DFLTCC Compat UBSAN: failed to upload coverage report.
CMake / Windows GCC Compat No Opt: doesn't seem to be related to this change, but looks scary:
```
2022-06-22T09:26:58.0837250Z 557: D:/a/zlib-ng/zlib-ng/minigzip.exe: <fd:0>: incorrect data check
2022-06-22T09:26:58.0873491Z 557: CMake Error at cmake/run-and-redirect.cmake:53 (message):
2022-06-22T09:26:58.0914751Z 557:   D:/a/zlib-ng/zlib-ng/minigzip.exe;-d;-c failed: 1
2022-06-22T09:26:58.2082861Z 557: 
2022-06-22T09:26:58.2111575Z 557: 
2022-06-22T09:26:58.2469329Z 557: CMake Error at cmake/test-compress.cmake:246 (message):
2022-06-22T09:26:58.2622435Z 557:   Decompress gzip failed: 1
2022-06-22T09:26:58.2642465Z 557: 
2022-06-22T09:26:58.2647497Z 557: 
2022-06-22T09:26:58.2692595Z 552/686 Test #557: minigzip-xml-h .............................................................................................................***Failed    3.06 sec
```

test/CMakeLists.txt
```diff
@@ -100,6 +97,16 @@ endif()
 
 target_link_libraries(gtest_zlib zlibstatic GTest::GTest)
 
+find_package(Threads)
+if(Threads_FOUND)
+	target_sources(gtest_zlib PRIVATE test_deflate_concurrency.cc)
```
nmoinvaz: I think there is a tab character here. Sorry 😕

iii-i: Ouch. Thanks for noticing this.

nmoinvaz (approved):

nmoinvaz:
It looks like this has broken the Windows GCC builds. Perhaps we should disable pthreads support on Windows that is not MSVC.

iii-i:
Sorry about that, I will take a look.

iii-i:
I've added

```
    - name: Setup tmate session
      uses: mxschmitt/action-tmate@v3
```

to ssh into the Windows runner, installed windbg using

```
env -u TEMP -u TMP choco install --force windows-sdk-10-version-2004-windbg
```

enabled and dumped loader logs with

```
'/c/Program Files (x86)/Windows Kits/10/Debuggers/x64/gflags.exe' -i gtest_zlib.exe +sls
'/c/Program Files (x86)/Windows Kits/10/Debuggers/x64/cdb.exe' gtest_zlib.exe
```

and finally saw the following:

```
1a88:1200 @ 03123968 - LdrpReportError - ERROR: Locating export "_ZSt28__throw_bad_array_new_lengthv" for DLL "D:\a\zlib-ng\zlib-ng\test\gtest_zlib.exe" failed with status: 0xc0000139.
(1a88.1200): Unknown exception - code c0000139 (first chance)
```

This is an indication that we are running with a different c++ standard library than we were building with. And indeed:

```
PATH=/c/ProgramData/chocolatey/lib/mingw/tools/install/mingw64/bin ./gtest_zlib.exe
```

runs successfully. I'll try to figure out what to do with this now.

PR #1268, opened 2022-05-08
Speed up software CRC-32 computation by a factor of 1.5 to 3.

nmoinvaz:
Use the interleaved method of Kadatch and Jenkins in order to make
use of pipelined instructions through multiple ALUs in a single
core. This also speeds up and simplifies the combination of CRCs,
and updates the functions to pre-calculate and use an operator for
CRC combination.

PR #1294, opened 2022-06-13
Extend GZIP conditional

stoeckmann:
If gzip support has been disabled during compilation then also
consider gzip relevant states as invalid in deflateStateCheck.

Also the gzip state definitions can be removed.

This change leads to failure in test/example, and I am not sure
what the GZIP conditional is trying to achieve. All gzip related
functions are still defined in zlib.h

Alternative approach is to remove the GZIP define.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -349,11 +349,11 @@ static int deflateStateCheck (PREFIX3(stream) *strm) {
     if (s == NULL || s->strm != strm || (s->status != INIT_STATE &&
 #ifdef GZIP
                                            s->status != GZIP_STATE &&
-#endif
                                            s->status != EXTRA_STATE &&
                                            s->status != NAME_STATE &&
                                            s->status != COMMENT_STATE &&
                                            s->status != HCRC_STATE &&
+#endif
                                            s->status != BUSY_STATE &&
                                            s->status != FINISH_STATE))
         return 1;
--- a/deflate.h
+++ b/deflate.h
@@ -54,16 +54,16 @@
 #define END_BLOCK 256
 /* end of block literal code */
 
-#define INIT_STATE    42    /* zlib header -> BUSY_STATE */
+#define INIT_STATE      42    /* zlib header -> BUSY_STATE */
 #ifdef GZIP
-#  define GZIP_STATE  57    /* gzip header -> BUSY_STATE | EXTRA_STATE */
+#  define GZIP_STATE    57    /* gzip header -> BUSY_STATE | EXTRA_STATE */
+#  define EXTRA_STATE   69    /* gzip extra block -> NAME_STATE */
+#  define NAME_STATE    73    /* gzip file name -> COMMENT_STATE */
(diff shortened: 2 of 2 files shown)


nmoinvaz (commented):
It may be possible to also wrap these struct fields in `#ifdef GZIP`:
```
    uint32_t             gzindex;          /* where in extra, name, or comment */
    PREFIX(gz_headerp)   gzhead;           /* gzip header information to write */
```

Dead2:
> It may be possible to also wrap these struct fields in `#ifdef GZIP`:
> 
> ```
>     uint32_t             gzindex;          /* where in extra, name, or comment */
>     PREFIX(gz_headerp)   gzhead;           /* gzip header information to write */
> ```

The problem with doing that is that it affects cacheline alignment and grouping for the whole rest of the struct, potentially making adjacent hot variables no longer reside in the same cacheline etc, so probably not worth it unless we do a much bigger job of it and ensure these get placed at the end of the struct where they will do no harm (or replace them with dummy values of the same size, not that that would really make any difference other than potentially for readability).

stoeckmann:
Neither the `configure` script nor the `CMakeLists.txt` offer a reasonable way to disable GZIP. Since zlib has only the configure script, the statement is true for that project as well. So maybe it wouldn't hurt to remove these defines. It reduces complexity after all.

Or is it desirable to extend these build environments?

Dead2 (approved):

Dead2:
This define is only really useful when building zlib statically into an application that does not need to support gzip at all.

deflate.h has this comment:
```
/* define NO_GZIP when compiling if you want to disable gzip header and
   trailer creation by deflate().  NO_GZIP would be used to avoid linking in
   the crc code when it is not needed.  For shared libraries, gzip encoding
   should be left enabled. */
#ifndef NO_GZIP
#  define GZIP
#endif
```

I am not sure how much code space is saved by disabling gzip and crc support, so that would have to be investigated before any real discussion on this could take place. I feel pretty sure that it does not affect performance measurably (except by random cache alignment chance). I also think gzip support can safely be enabled even when an application expects it to be absent.
So removing this support could potentially be done while still remaining compatible with zlib, the question is whether doing so is worth losing the (likely meagre) gains when using it and whether this actually simplifies the code enough to warrant a removal.

mtl1979:
Especially the new CRC tables from 1.2.12 are quite large, so the savings can be few kilobytes.

nmoinvaz:
@mtl1979 the code size after the new tables were incorporated decreased. 

See https://github.com/zlib-ng/zlib-ng/pull/1268#issuecomment-1136398485

nmoinvaz:
@mtl1979 the tables look like they might be larger in the header file but actually only a portion of those generated tables are used.

mtl1979:
@nmoinvaz Code size might have decreased but const variables aren't considered code. Old CRC code had no support for 64-bit processors, so it had simpler tables. When we separated little-endian and big-endian tables, we practically reduced the size of the tables by 50%.

PR #1295, opened 2022-06-15
IBM Z DFLTCC: Test with MSan

iii-i:
* Add a CI job.
* Do not collect coverage: LLVM's gcov support (part of compiler-rt)
  cannot be built with the MSan instrumentation, which means that
  whenever it's called (in particular, in order to write the results to
  a file at the end), there is a risk of false positives.
* Add __msan_unpoison() calls to DFLTCC inline assembly.
* Make parameter block sizes symbolic constants.
* Move dfltcc() definition after struct dfltcc_param_v0 definition.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -252,6 +252,12 @@ jobs:
             ldflags: -static
             codecov: ubuntu_gcc_s390x_dfltcc_compat
 
+          - name: Ubuntu Clang S390X DFLTCC MSAN
+            os: z15
+            compiler: clang-11
+            cxx-compiler: clang++-11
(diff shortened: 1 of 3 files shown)


nmoinvaz:
Side-note: You can replace the top 4 includes in `dfltcc_detail.h` with `#include "../../zbuild.h"`

iii-i:
> Side-note: You can replace the top 4 includes in `dfltcc_detail.h` with `#include "../../zbuild.h"`

I think `stdio.h` needs to be kept, because `zbuild.h` includes it conditionally, and I need it here for `sprintf()`.

Dead2 (approved):
LGTM

Issue #1279, opened 2022-05-30
`libz-ng.a` and `libz.a` from zlib-compat mode have symbol conflicts


joshtriplett:
zlib-ng in native mode avoids symbol conflicts with stock zlib, but still has symbol conflicts with zlib supplied from zlib-ng built in zlib-compat mode.

This can cause an issue in a few different scenarios:
- If a distribution uses zlib-ng in zlib-compat mode as its system zlib, and an application statically links native zlib-ng but also links a library (such as a system library) that links to zlib, expecting them not to conflict.
- If an application uses native zlib-ng, but has many dependencies, one of which uses zlib-ng in zlib-compat mode (e.g. because it needs to link a C library that needs zlib symbols but wants the performance of zlib-ng).

One instance of this manifested as these link errors:

```
  = note: liblibz_ng_sys-fb58de474afbb4a4.rlib(inflate.obj) : error LNK2005: fixedtables already defined in liblibz_sys-38cf6ff5b07be7a7.rlib(inflate.obj)
          liblibz_ng_sys-fb58de474afbb4a4.rlib(crc32_braid.obj) : error LNK2005: crc32_braid already defined in liblibz_sys-38cf6ff5b07be7a7.rlib(crc32_braid.obj)
          liblibz_ng_sys-fb58de474afbb4a4.rlib(zutil.obj) : error LNK2005: z_error already defined in liblibz_sys-38cf6ff5b07be7a7.rlib(zutil.obj)
          liblibz_ng_sys-fb58de474afbb4a4.rlib(zutil.obj) : error LNK2005: zlibng_version already defined in liblibz_sys-38cf6ff5b07be7a7.rlib(zutil.obj)
          liblibz_ng_sys-fb58de474afbb4a4.rlib(zutil.obj) : error LNK2005: zng_alloc_aligned already defined in liblibz_sys-38cf6ff5b07be7a7.rlib(zutil.obj)
          liblibz_ng_sys-fb58de474afbb4a4.rlib(zutil.obj) : error LNK2005: zng_calloc already defined in liblibz_sys-38cf6ff5b07be7a7.rlib(zutil.obj)
          liblibz_ng_sys-fb58de474afbb4a4.rlib(zutil.obj) : error LNK2005: zng_cfree already defined in liblibz_sys-38cf6ff5b07be7a7.rlib(zutil.obj)
[...]

PR #1299, opened 2022-06-19
Improve dual link testing

nmoinvaz:
- Added a `gtest` test that compiles against the zlib that we are using for dual link testing.
- Moved dual linking test into its own GHA workflow
- In new CI, it tests compiling and linking `madler/zlib:develop` against zlib-ng native api
- In new CI, it tests compiling and linking zlib-ng compat api against zlib-ng native api 
  - #1279


Dead2:
@nmoinvaz Test supposed to fail linking currently?

nmoinvaz:
Yes it is suppose to fail until we actually fix those issues. I could try and do it in this PR or another.. I would just go the macro route..

Dead2 (approved):

mtl1979:
I'd rather see the fix in either separate commit in this PR or merged before this PR... This way we can verify that the fix is complete and doesn't cause any unacceptable regressions. The reasoning is pretty much same as was for the fix and tests for deflateBound/compressBound.

nmoinvaz:
I am not so sure that the ordering really matters. There is a way to verify that the tests pass either way.

PR #1305, opened 2022-06-25
Prefix shared functions to prevent symbol conflict when linking

nmoinvaz:
Prevents symbol conflict when linking native API against compatibility API. See #1279.
```
crc32_braid.c:(.text+0x60): multiple definition of `crc32_braid'; ../../compat/libz.a(crc32_braid.c.o):crc32_braid.c:(.text+0x70): first defined here
deflate.c:(.text+0x1a50): multiple definition of `read_buf'; ../../compat/libz.a(deflate.c.o):deflate.c:(.text+0x1a40): first defined here
deflate.c:(.text+0x1ae0): multiple definition of `fill_window'; ../../compat/libz.a(deflate.c.o):deflate.c:(.text+0x1ae0): first defined here
inflate.c:(.text+0x460): multiple definition of `fixedtables'; ../../compat/libz.a(inflate.c.o):inflate.c:(.text+0x450): first defined here
zutil.c:(.text+0x0): multiple definition of `zng_calloc'; ../../compat/libz.a(zutil.c.o):zutil.c:(.text+0x0): first defined here
zutil.c:(.text+0x60): multiple definition of `zng_cfree'; ../../compat/libz.a(zutil.c.o):zutil.c:(.text+0x60): first defined here
zutil.c:(.text+0x70): multiple definition of `zlibng_version'; ../../compat/libz.a(zutil.c.o):zutil.c:(.text+0x80): first defined here
zutil.c:(.text+0xb0): multiple definition of `zng_alloc_aligned'; ../../compat/libz.a(zutil.c.o):zutil.c:(.text+0xc0): first defined here
zutil.c:(.text+0x160): multiple definition of `zng_free_aligned'; ../../compat/libz.a(zutil.c.o):zutil.c:(.text+0x170): first defined here
/usr/bin/ld: ../libz-ng.a(zutil.c.o):(.rodata+0x0): multiple definition of `zlibng_string'; ../../compat/libz.a(zutil.c.o):(.rodata+0x0): first defined here
```

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -202,11 +202,11 @@ int32_t ZNG_CONDEXPORT PREFIX(deflateInit2)(PREFIX3(stream) *strm, int32_t level
 
     strm->msg = NULL;
     if (strm->zalloc == NULL) {
-        strm->zalloc = zng_calloc;
+        strm->zalloc = PREFIX3(calloc);
         strm->opaque = NULL;
     }
     if (strm->zfree == NULL)
-        strm->zfree = zng_cfree;
+        strm->zfree = PREFIX3(cfree);
 
     if (level == Z_DEFAULT_COMPRESSION)
         level = 6;
@@ -400,14 +400,14 @@ int32_t Z_EXPORT PREFIX(deflateSetDictionary)(PREFIX3(stream) *strm, const uint8
     next = strm->next_in;
     strm->avail_in = dictLength;
     strm->next_in = (z_const unsigned char *)dictionary;
-    fill_window(s);
+    PREFIX(fill_window)(s);
     while (s->lookahead >= STD_MIN_MATCH) {
         str = s->strstart;
         n = s->lookahead - (STD_MIN_MATCH - 1);
         s->insert_string(s, str, n);
         s->strstart = str + n;
         s->lookahead = STD_MIN_MATCH - 1;
-        fill_window(s);
+        PREFIX(fill_window)(s);
     }
     s->strstart += s->lookahead;
     s->block_start = (int)s->strstart;
@@ -1090,7 +1090,7 @@ int32_t Z_EXPORT PREFIX(deflateCopy)(PREFIX3(stream) *dest, PREFIX3(stream) *sou
  * allocating a large strm->next_in buffer and copying from it.
  * (See also flush_pending()).
  */
-Z_INTERNAL unsigned read_buf(PREFIX3(stream) *strm, unsigned char *buf, unsigned size) {
+Z_INTERNAL unsigned PREFIX(read_buf)(PREFIX3(stream) *strm, unsigned char *buf, unsigned size) {
     uint32_t len = strm->avail_in;
 
     len = MIN(len, size);
@@ -1175,7 +1175,7 @@ static void lm_init(deflate_state *s) {
  *    option -- not supported here).
  */
 
-void Z_INTERNAL fill_window(deflate_state *s) {
+void Z_INTERNAL PREFIX(fill_window)(deflate_state *s) {
     unsigned n;
     unsigned int more;    /* Amount of free space at the end of the window. */
     unsigned int wsize = s->w_size;
@@ -1219,7 +1219,7 @@ void Z_INTERNAL fill_window(deflate_state *s) {
          */
         Assert(more >= 2, "more < 2");
 
-        n = read_buf(s->strm, s->window + s->strstart + s->lookahead, more);
+        n = PREFIX(read_buf)(s->strm, s->window + s->strstart + s->lookahead, more);
         s->lookahead += n;
 
         /* Initialize the hash value now that we have some input: */
--- a/zutil.c
+++ b/zutil.c
@@ -20,18 +20,18 @@ z_const char * const PREFIX(z_errmsg)[10] = {
     (z_const char *)""
 };
 
-const char zlibng_string[] =
-    " zlib-ng 2.1.0.devel forked from zlib";
-
 #ifdef ZLIB_COMPAT
 const char * Z_EXPORT zlibVersion(void) {
     return ZLIB_VERSION;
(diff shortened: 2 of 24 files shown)


zutil.c
```diff
@@ -20,18 +20,18 @@ z_const char * const PREFIX(z_errmsg)[10] = {
...
-const char zlibng_string[] =
-    " zlib-ng 2.1.0.devel forked from zlib";
-
 #ifdef ZLIB_COMPAT
 const char * Z_EXPORT zlibVersion(void) {
     return ZLIB_VERSION;
 }
-#endif
+#else
+const char zlibng_string[] =
```
nmoinvaz, line 28: This string is not used anywhere - only exported. Not sure if we really need it imo.

deflate.c
```diff
@@ -202,11 +202,11 @@ int32_t ZNG_CONDEXPORT PREFIX(deflateInit2)(PREFIX3(stream) *strm, int32_t level
 
     strm->msg = NULL;
     if (strm->zalloc == NULL) {
-        strm->zalloc = zng_calloc;
+        strm->zalloc = PREFIX3(calloc);
```
mtl1979, line 205: This should be exported symbol, so needs updating the list of exported symbols. `PREFIX3` adds `z_` prefix when building in compatibility mode.

deflate.c
```diff
@@ -202,11 +202,11 @@ int32_t ZNG_CONDEXPORT PREFIX(deflateInit2)(PREFIX3(stream) *strm, int32_t level
...
         strm->opaque = NULL;
     }
     if (strm->zfree == NULL)
-        strm->zfree = zng_cfree;
+        strm->zfree = PREFIX3(cfree);
```
mtl1979, line 209: This should be exported symbol, so needs updating the list of exported symbols. `PREFIX3` adds `z_` prefix when building in compatibility mode.

zutil.c
Dead2: It makes sure we have that plaintext string in the resulting binary, thus on linux we can run `strings libz-ng.so` and it will show that string. It can therefore be used for simple identification of version and potentially detection of zlib-ng being included in third-party applications along with what version they compiled in.

nmoinvaz: I have changed it to be only in native api. Should it be prefixed some how too?

mtl1979: In compatibility mode, we could use something like `zlibng_compat_string`, this would avoid conflict when mixing native zlib-ng and compatibility library, and also mixing zlib-ng with other zlib forks.

deflate.c
nmoinvaz: Why should it be added to export symbols?

nmoinvaz: Nevermind I see zlib.map.

zutil.c
nmoinvaz: Actually it looks to only be exported from zlib-ng.map anyways and not zlib.map.

mtl1979: `zlib-ng.map` has `zng_zcalloc` and `zlib.map` has `zcalloc`... I haven't checked when that extra `z` after `_` was dropped and why...

nmoinvaz: Yeah but only zlib-ng.map exports `zlibng_string`. zlib.map doesn't export such a string.

mtl1979: I was just saying the map files are not quite up-to-date... Even if we build with compatibility mode, it would make sense to have easy way to check if some binary is compiled against zlib-ng, especially if it is compiled against both native version and compatibility version.

Dead2: We need to include it in both, it does not actually need to be exported, it only needs to end up in the binary, and the name it has is not important at all.
Adding "Native API" or "Compat API" to that string depending on compiled version could be an interesting addition.

nmoinvaz: Ok I have updated the strings. Let me know if you like it or not.

nmoinvaz:
Rebased.

zutil.c
```diff
@@ -20,18 +20,21 @@ z_const char * const PREFIX(z_errmsg)[10] = {
     (z_const char *)""
 };
 
-const char zlibng_string[] =
-    " zlib-ng 2.1.0.devel forked from zlib";
-
 #ifdef ZLIB_COMPAT
+const char zlibng_string_compat[] =
+    " zlib-ng 2.1.0.devel compat api forked from zlib";
```
mtl1979: `API` should be all caps, as it is initialism.

zutil.c
```diff
@@ -20,18 +20,21 @@ z_const char * const PREFIX(z_errmsg)[10] = {
...
+
 const char * Z_EXPORT zlibVersion(void) {
     return ZLIB_VERSION;
 }
-#endif
+#else
+const char zlibng_string[] =
+    " zlib-ng 2.1.0.devel native api forked from zlib";
```
mtl1979: `API` should be all caps, as it is initialism.

joshtriplett:
This looks great, thank you. I'd love to have this available, and I'd be happy to adapt the Rust package to use a version with this change.

dzenanz:
What are the remaining blockers for integrating this?

nmoinvaz:
@Dead2 this one is ready too.

nmoinvaz:
@Dead2 have you seen this?

Dead2:
@nmoinvaz I did, but I have a different possible solution that I want to present. Just have to get the time to gather and present all the relevant details 😉

dzenanz:
Has there been any movement on this?

Dead2:
@nmoinvaz Will this need any changes to accomodate #1322 or other recent changes?
I don't see any conflicts, but I'd like you to chime in too before merging this. Perhaps a rebase would be best in any case.

nmoinvaz:
From what I can see, I think it should be okay to merge both. I have rebased this PR as requested.

PR #1307, opened 2022-06-27
Uninstall strawberryperl

iii-i:
strawberryperl installs /c/Strawberry/c/bin/libstdc++-6.dll, which is
incompatible with the mingw64 one. zlib-ng does not need perl, so
simply remove it.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -425,6 +425,9 @@ jobs:
     - name: Install packages (Windows)
       if: runner.os == 'Windows'
       run: |
+        # strawberryperl installs /c/Strawberry/c/bin/libstdc++-6.dll, which is incompatible with the mingw64 one.
+        # zlib-ng does not need perl, so simply remove it.
+        choco uninstall strawberryperl --no-progress
         choco install ninja ${{ matrix.packages }} --no-progress
 
     - name: Install packages (macOS)


iii-i:
This fixes CI CMake / Windows GCC, but not CI CMake / Windows GCC Compat No Opt.
The second one produces errors like:
```
2022-06-27T19:03:05.1423512Z 128: D:/a/zlib-ng/zlib-ng/minigzip.exe: <fd:0>: incorrect data check
```
I'll have a look what is wrong there.

iii-i:
Bisect points to

```
commit 363a95fb9b8c4bef7225164fa1deb5481a35c3c0
Author: Nathan Moinvaziri <nathan@nathanm.com>
Date:   Sun Jan 9 15:01:23 2022 -0800

    Introduce zmemcpy to use unaligned access for architectures we know support unaligned access, otherwise use memcpy.
```

iii-i:
There is a warning, might be related:

```
In file included from D:/a/zlib-ng/zlib-ng/chunkset.c:5:
D:/a/zlib-ng/zlib-ng/chunkset_tpl.h: In function 'chunkmemset_c':
D:/a/zlib-ng/zlib-ng/zbuild.h:199:52: warning: array subscript 4 is outside array bounds of 'chunk_t[1]' {aka 'long long unsigned int[1]'} [-Warray-bounds]
  199 | #  define zmemcpy_4(dest, src) *((uint32_t *)dest) = *((uint32_t *)src)
      |                                ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~
D:/a/zlib-ng/zlib-ng/chunkset.c:22:5: note: in expansion of macro 'zmemcpy_4'
   22 |     zmemcpy_4(dest+4, from);
      |     ^~~~~~~~~
In file included from D:/a/zlib-ng/zlib-ng/chunkset.c:44:
D:/a/zlib-ng/zlib-ng/chunkset_tpl.h:116:13: note: while referencing 'chunk'
  116 |     chunk_t chunk;
      |             ^~~~~
```

iii-i:
The warning is fixed in ab6665b.

I've experimented with chunkset, and the following helps with some (but not all) failing tests:

```
--- a/chunkset_tpl.h
+++ b/chunkset_tpl.h
@@ -36,7 +36,7 @@ Z_INTERNAL uint8_t* CHUNKCOPY(uint8_t *out, uint8_t const *from, unsigned len) {
     len -= align;
     while (len > 0) {
         loadchunk(from, &chunk);
-        storechunk(out, &chunk);
+        memcpy(out, &chunk, sizeof(uint64_t));
         out += sizeof(chunk_t);
         from += sizeof(chunk_t);
         len -= sizeof(chunk_t);
```

Can this be related to #1100? In any case, Windows GCC Compat No Opt failure looks like a separate issue to me.

nmoinvaz:
Looks like perhaps dest in zmemcpy_4 needs parentheses around it. But I agree it is a separate issue.

nmoinvaz (approved):

nmoinvaz:
Oddly my zlib-ng defines `zmemcpy_4` like so:
```
#  define zmemcpy_4(dest, src)    (*((uint32_t *)(dest)) = *((uint32_t *)(src)))
```
Perhaps you need a rebase?

iii-i:
> Oddly my zlib-ng defines `zmemcpy_4` like so:
> 
> ```
> #  define zmemcpy_4(dest, src)    (*((uint32_t *)(dest)) = *((uint32_t *)(src)))
> ```
> 
> Perhaps you need a rebase?

That's what I meant by "The warning is fixed in https://github.com/zlib-ng/zlib-ng/commit/ab6665b1bee170c27d229899ed23e6053ebd900b" - I've been testing the bisected version. However, the diff that I posted was tested against develop.

iii-i:
I tried to compare `chunkcopy_c` asm with and without https://github.com/zlib-ng/zlib-ng/pull/1307#issuecomment-1167922606 - that failure I can fortunately reproduce locally. I can't claim I understood everything, but the general structure there is:

```
if (aligned)
  vectorized_loop;
else
  simple_loop;
```

The version without the patch has fewer alignment checks than the one with the patch. This may indeed have to do with compiler assuming that `out`/`from` are properly aligned (since they are accessed via `*(uint64_t*)`), while in reality they are not.

phprus:
@iii-i, @nmoinvaz 

Compiler has the right to assume that the memory accessed through the pointer ``(uint64_t*)ptr`` will be aligned to the ``sizeof(uint64_t)``.

``memcpy`` doesn't add extra instructions:
https://godbolt.org/z/jMzjz9x84

https://github.com/zlib-ng/zlib-ng/blob/41d67396924ccc7ab1ff9a7e7d434bfb0887b136/zbuild.h#L221-L225

The comment talks about old compilers, but even gcc 4.1.2 inlines memcpy:
https://godbolt.org/z/77K96aMoa

https://github.com/zlib-ng/zlib-ng/blob/1bcd15eaf32dbfcd834f1b21f5894edd6b466095/CMakeLists.txt#L23-L26

zlib-ng requires C99. I'm assuming all C99-compilers support ``memcpy`` inlining.

But inlining ``memcmp`` is only gcc-7+. msvc does not support it:
https://godbolt.org/z/rzG1vG86f

msvc does not support Strict aliasing rules.
For msvc we can use code with type cast, and for new GCCs (4+ for ``memcpy`` and 7+ for ``memcmp``) or Clang 3+ with ``mem*`` functions.

Other solution - using gcc and clang magick: ``__attribute__((__may_alias__))``
https://godbolt.org/z/P9GenPdW9
https://gcc.gnu.org/onlinedocs/gcc-4.1.2/gcc/Type-Attributes.html

nmoinvaz:
> The comment talks about old compilers, but even gcc 4.1.2 inlines memcpy: https://godbolt.org/z/77K96aMoa

ARM GCC < 9.3 does not.

@phprus perhaps it makes sense to just remove `z_memcpy` defines or point them all to memcpy always.

phprus:
@nmoinvaz 
Or ``__attribute__((aligned(1), __may_alias__))``:
https://stackoverflow.com/questions/46790550/c-undefined-behavior-strict-aliasing-rule-or-incorrect-alignment#comment118832347_52654425

> If you want an aliasing-safe unaligned u16, you can typedef it with __attribute__((aligned(1), may_alias)). GNU C gives you the tools.

nmoinvaz:
@Dead2 may have an opinion which way to go.

Dead2:
I think we should go for straight memcpy (I suggested that when we did a previous rework of this as well), then we only have memcmp left requiring special consideration, reducing the maintenance burden a little.

Dead2 (approved):
LGTM

PR #1309, opened 2022-06-29
Don't use unaligned access for memcpy instructions

nmoinvaz:


phprus:
Compare without ``memcmp`` (see: ``eq_uint64_t_tmp``):
https://godbolt.org/z/TavGsbTh9

GCC, Clang and MSVC does not call ``memcpy``.
But I don't know if ``eq_uint64_t_tmp`` function will inlining.

On ARM and GCC < 9 there will be calls to ``memcmp``.

nmoinvaz:
@phprus the `__attribute__((__may_alias__))` would be a good idea for a separate PR, I can see it definitely helps on old GCC.

phprus:
@nmoinvaz 
On GCC ``__may_alias__ `` and ``aligned(1)`` (to decrease alignment).

See link to stackoverflow in my comment:
https://github.com/zlib-ng/zlib-ng/pull/1307#issuecomment-1169569193
and
https://gcc.gnu.org/onlinedocs/gcc/Common-Type-Attributes.html

But Clang does not support the ``aligned`` attribute. I don't know how to force Clang to guarantee unaligned reads.

nmoinvaz:
Rebased.

shuffle2:
> Compare without `memcmp` (see: `eq_uint64_t_tmp`): https://godbolt.org/z/TavGsbTh9
> 
> GCC, Clang and MSVC does not call `memcpy`. But I don't know if `eq_uint64_t_tmp` function will inlining.
> 
> On ARM and GCC < 9 there will be calls to `memcmp`.

Hi, I just want to point out that you've typo'd the msvc arguments there. You meant `/O2` and put `/02`. Also, [`/O2` implies `/Og /Oi /Ot /Oy /Ob2 /GF /Gy`](https://docs.microsoft.com/en-us/cpp/build/reference/o1-o2-minimize-size-maximize-speed?view=msvc-170). So `/Oi` is not needed. `/Ob3` would add more aggressive inlining but isn't needed in this case.

So in the end, msvc is generating code nearly same as gcc/clang here: https://godbolt.org/z/9MaYYsTbK
msvc/arm64 also emits the memcmp

Dead2 (approved):
LGTM

PR #1310, opened 2022-06-29
Fixed conversion warning with level in compress_bound unit tests.

nmoinvaz:
```
  test_compress_bound.cc(43,1): warning C4267: 'argument': conversion from 'size_t' to 'int32_t', possible loss of data
```

### DIFF
--- a/test/test_compress_bound.cc
+++ b/test/test_compress_bound.cc
@@ -18,9 +18,9 @@
 
 #define MAX_LENGTH (32)
 
-class compress_bound_variant : public testing::TestWithParam<z_size_t> {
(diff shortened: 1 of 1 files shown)


Dead2 (approved):
LGTM

test/test_compress_bound.cc
```diff
@@ -20,7 +20,7 @@
 
 class compress_bound_variant : public testing::TestWithParam<z_size_t> {
```
mtl1979: Shouldn't this use `int32_t`too?

nmoinvaz: Yes you are right.

nmoinvaz: Fixed.

PR #1311, opened 2022-06-30
Fix inflateBack to detect invalid input with distances too far.

nmoinvaz:
Original commit: https://github.com/madler/zlib/commit/2333419cd76cb9ae5f15c9b240b16a2052b27691


Dead2 (approved):
LGTM

PR #1317, opened 2022-07-12
Set `CMAKE_*_COMPILER_TARGET` to full triplet

longnguyen2004:
Leaving it as-is breaks clang faking as gcc, so set it to the full triplet.


nmoinvaz (approved):

nmoinvaz:
Can you append " for mingw CMake toolchains" to the commit message?

longnguyen2004:
Renamed!

cmake/toolchain-mingw-i686.cmake
```diff
@@ -14,8 +14,8 @@ set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
 
 # Prefer posix gcc variant for gtest pthread support
 find_program(C_COMPILER_FULL_PATH NAMES
-    ${CMAKE_C_COMPILER_TARGET}-w64-mingw32-gcc-posix
-    ${CMAKE_C_COMPILER_TARGET}-w64-mingw32-gcc
+    ${CMAKE_C_COMPILER_TARGET}-gcc-posix
+    ${CMAKE_C_COMPILER_TARGET}-gcc
     gcc-mingw32-w64-${CMAKE_C_COMPILER_TARGET})
```
mtl1979: This like needs also be corrected...

cmake/toolchain-mingw-i686.cmake
```diff
@@ -24,14 +24,14 @@ set(CMAKE_C_COMPILER ${C_COMPILER_FULL_PATH})
 
 find_program(CXX_COMPILER_FULL_PATH NAMES
     g++-mingw-w64-${CMAKE_CXX_COMPILER_TARGET}
```
mtl1979: This like needs also be corrected...

cmake/toolchain-mingw-i686.cmake
longnguyen2004: I'm not aware of any distro that uses such a naming scheme, so I'll remove it.

cmake/toolchain-mingw-i686.cmake
longnguyen2004: I'm not aware of any distro that uses such a naming scheme, so I'll remove it.

Dead2 (approved):
LGTM

PR #1318, opened 2022-07-17
cmake: respect custom `RC` flags and delete `GCC_WINDRES`

vszakats:
Before this patch, `zlib.rc` was compiled using a manual command [1] when
using the MinGW (and MSYS/Cygwin) toolchains. This method ignores
`CMAKE_RC_FLAGS` and offers no other way to pass a custom flag, breaking
the build in cases where a custom `windres` option is required. E.g.
`--target=` or `-I` on some platforms and configuration, in particular
with `llvm-windres`.

This patch deletes the special case for these toolchains and lets CMake
compile the `.rc` file the default way used for all Windows targets.

I'm not entirely sure why this special case was added back in 2011. The
need to pass `-DGCC_WINDRES` is my suspect. We can resolve this much
simpler by adding this line for the targets that require it:
```cmake
   set(CMAKE_RC_FLAGS "${CMAKE_RC_FLAGS} -DGCC_WINDRES")
```

But, the `.rc` line protected by `GCC_WINDRES`, these days work just fine
with `windres`. Moreover, that protected line are oboslete flags from the
16-bit era, which for a long time have no effect, as documented here:
<https://docs.microsoft.com/windows/win32/menurc/common-resource-attributes>

So, this patch deletes `GCC_WINDRES` from the project entirely.

[1] dc5a43e


Dead2 (approved):
Looks good to me, but I would prefer to see someone more familiar with windows dlls verify this before pulling.

vszakats:
I'm happy to answer questions, if there are any.

Dead2:
The important part is that (for compat-mode) this does not change the ABI in any way that breaks compatibility.
I am personally not familiar enough with this and the relevant tools used to verify this on windows.

vszakats:
Got it. This patch doesn't affect the ABI in any way. This particular Windows resource is static version information only. Also the actual version information isn't changed by the patch.

The four deleted flags (`MOVEABLE`, `IMPURE`, `LOADONCALL`, `DISCARDABLE`) used to affect the way Windows manages the chunk of memory the resource occupies, way back when. These days they are ignored and only retained for compatibility. This is documented here: <https://docs.microsoft.com/windows/win32/menurc/common-resource-attributes>

UPDATE: Refreshed commit message with this information.

nmoinvaz:
When I use CMake to compile dynamic libraries, all I do is add the .rc to the source files list and CMake does the rest.

vszakats:
@nmoinvaz: Indeed, that's what I'm aiming for with this patch.

PR #1325, opened 2022-08-12
cmake: Set the mingw CMAKE_<LANG>_COMPILER_TARGET to the full triple,…

mstorsjo:
… like for other targets

When compiling with GCC; the CMAKE_<LANG>_COMPILER_TARGET variables
don't have any direct effect, but are only used as convenience
variables in the cmake toolchain files.

When compiling with Clang (which also can target Mingw), those
variables suddenly have an effect. If passing just the architecture
name, it ends up with CMake passing plain e.g. `--target=x86_64` to
Clang - which is interpreted as "architecture x86_64, on the current
OS".

Even if the toolchain file specifically is looking for a compiler
named e.g. x86_64-w64-mingw32-gcc, it can be picked up by a matching
tool frontend backed by Clang (which is provided by e.g. the llvm-mingw
toolchain), but the incorrect CMAKE_<LANG>_COMPILER_TARGET variable
breaks things.

This fixes issues reported in
https://github.com/mstorsjo/llvm-mingw/issues/293.


mtl1979:
Duplicate of #1317.

mstorsjo:
> Duplicate of #1317.

Oops, sorry about that!

PR #1326, opened 2022-08-17
Fix build as subproject.

phprus:
Bug added in commit 1943c9721b61ebceaeeba4c299c461479101b773

``CMAKE_SOURCE_DIR`` and ``CMAKE_BINARY_DIR`` is the full path to the top level project of the current CMake project tree.
``PROJECT_SOURCE_DIR`` and ``PROJECT_BINARY_DIR`` - is the full path to the current project.

@nmoinvaz


nmoinvaz:
Build as subproject? What does that mean? You mean using add_subdirectory? 

Is it possible to add a unit test some how to test that?

gdevenyi:
> Build as subproject? What does that mean? You mean using add_subdirectory?

This would be for ExternalProject_Add, something superbuild projects use:
https://cmake.org/cmake/help/latest/module/ExternalProject.html

Example using the mainline ZLIB here, https://github.com/BIC-MNI/minc-toolkit-v2/blob/master/cmake-modules/BuildZLIB.cmake

phprus:
Yes, add_subdirectory.
Test added, but it is slow because it recompiles zlib-ng.

nmoinvaz:
Instead of modifying test/CMakeLists.txt, it might be better as its own GitHub Actions yaml. Or perhaps it can be incorporated in to the `cmake.yaml` matrix using `build-dir` and `build-src-dir`.

https://github.com/zlib-ng/zlib-ng/blob/89763032d57e3da5301f4cd6e1f363e7a1f85f02/.github/workflows/cmake.yml#L52-L53

Ubuntu 18.04 complains about:
```
/home/runner/work/zlib-ng/zlib-ng/test/add-subdirectory-test/main.c: In function ‘main’:
/home/runner/work/zlib-ng/zlib-ng/test/add-subdirectory-test/main.c:4:1: error: parameter name omitted
 int main(int, char **) {
 ^~~
```

phprus:
@nmoinvaz
Done.
New GitHub Actions check: "Ubuntu GCC -O3 OSB add_subdirectory"

.github/workflows/cmake.yml
```diff
@@ -54,6 +54,14 @@ jobs:
             codecov: ubuntu_gcc_osb
             cflags: -O3
 
+          - name: Ubuntu GCC -O3 OSB add_subdirectory
+            os: ubuntu-latest
+            compiler: gcc
+            cxx-compiler: g++
+            build-dir: ../build
+            build-src-dir: ../zlib-ng/test/add-subdirectory-project
+            cflags: -O3
```
nmoinvaz: I don't think we need `cflag` or `-O3` in the name.

phprus: Done

nmoinvaz (approved):

Dead2 (approved):
LGTM

PR #1329, opened 2022-08-20
Add support for compiling to WebAssembly using Emscripten.

nmoinvaz:
This is a remake of #794. We skip all the tests that use stdin/stdout and only run certain compatible Google Tests due to limitations with Emscripten.


.github/workflows/cmake-configure.yml
```diff
@@ -0,0 +1,37 @@
+name: Configure (wasm32)
```
mtl1979: Name of this file doesn't make sense... Maybe `configure-wasm32.yml` instead.

nmoinvaz: Looks like I goofed when renaming. I'm going to combine the two workflows into one.

nmoinvaz: I have cleaned it up and moved Emscripten into `cmake.yml` and `configure.yml`, it is much better than originally because I am using a GitHub Action to simplify installing Emscripten.

nmoinvaz:
I want to see if I can add back `check_cross_dep`.

nmoinvaz:
I added back `check_cross_dep` and rebased. It is ready now.

RReverser:
Can this be made work with vcpkg (`vcpkg install zlib-ng:wasm32-emscripten`)? Currently, even if I point it to your fork, it still fails due to trying to include the CPUID assembly from x86 folder. I suspect some small changes are necessary to auto-identify CHOST from the target triplet, but I'm not a CMake expert to figure out what they'd look like.

RReverser:
This is what I'm seeing in the logs when vcpkg is configuring it:

```
-- Arch not recognized, falling back to cmake arch: 'x86'
-- Basearch 'x86' not recognized, defaulting to 'x86'.
-- Basearch of 'x86' has been detected as: 'x86'
```

RReverser:
Ok found it and made a PR against your branch/fork: https://github.com/nmoinvaz/zlib-ng/pull/1

Not sure why the C-based detection doesn't work in vcpkg environment, but the CMake-one does work with vcpkg, and should be a bit faster as well.

RReverser:
Note: another large part Wasm support in zlib-ng is still missing is checking for SSE2 / SSSE3 / etc. intrinsics when `ARCH=wasm32`.

Currently both `configure` and `CMakeLists.txt` assume that those intrinsics only exist on x86 and x86-64, but Emscripten has pretty good support for SSE emulation via WebAssembly SIMD instructions: https://emscripten.org/docs/porting/simd.html

Enabling them would make zlib-ng on Wasm much faster.

mtl1979:
> Note: another large part Wasm support in zlib-ng is still missing is checking for SSE2 / SSSE3 / etc. intrinsics when `ARCH=wasm32`.
> 
> Currently both `configure` and `CMakeLists.txt` assume that those intrinsics only exist on x86 and x86-64, but Emscripten has pretty good support for SSE emulation via WebAssembly SIMD instructions: https://emscripten.org/docs/porting/simd.html
> 
> Enabling them would make zlib-ng on Wasm much faster.

Even though WASM has partial support for SSE intrinsics, I'm not completely sure if we should rely on the emulation or we should limit ourselves to WASM's own intrinsics. It's pretty much design decision that could make huge difference in performance. A lot of WASM is still experimental, so following the development process could/will cause huge amount of pull requests that can affect unrelated code due to needed workarounds if we don't isolate WASM support in early stage.

RReverser:
> Even though WASM has partial support for SSE intrinsics, I'm not completely sure if we should rely on the emulation or we should limit ourselves to WASM's own intrinsics.

At least for SSE2, the emulation is pretty good, and it makes a big difference performance-wise. We cross-compiled several image codecs - JPEG-XL, WebP and WebP v2 - with complex SIMD usages in [squoosh.app](https://squoosh.app) and used them there for ~1.5 years with no issues, and I know several larger projects that use them too ever since it was stabilised.

> A lot of WASM is still experimental

The core Wasm has been stable for >5 years, and newer features go through clear standardisation process with clear stages for those features. SIMD in particular is [finished](https://github.com/WebAssembly/proposals/blob/main/finished-proposals.md). It still should be opt-in at `configure` stage so that users could build for older browsers too, but the opcodes or the API won't change anymore, so it shouldn't add to the maintenance churn. 

In fact, some projects that make use of zlib-ng in Wasm already build it with SIMD support, but have to do some really hacky stuff to achieve that: https://github.com/kleisauke/wasm-vips/blob/4584bd9fe872b068f6fc8f4a8cfda918d79be68c/build.sh#L182-L189. That's where making it possible to `./configure` zlib-ng with SIMD on Wasm targets, at least optionally, would already help as users won't have to reach in and manually patch the build files.

mtl1979:
@RReverser I agree that making it optional feature is the best choice. If something fails at build time, we can always add test for configure stage to detect it and disable the optimizations that don't build correctly, or add workarounds if just one or two intrinsics are not fully supported or implemented.

RReverser:
> I agree that making it optional feature is the best choice.

Just to be clear, you're referring to SIMD not Wasm entirely? If so, yeah, I agree it's the best course of action.

Having a built-in, but optional, support is especially convenient since it gives zlib-ng the ability to choose which SIMD subset it wants to support on Wasm (e.g. SSE2 is definitely worthwhile, but it's best to skip SSSE3 as the instructions zlib-ng uses are currently unavailable).

Making it optional is also useful because Safari, as usual, is dragging their feet with implementation of new Web standards, and still haven't shipped Wasm SIMD support, so users who want the best perf will need to compile two versions of Wasm anyway (one w/ SIMD and another, a slower fallback, without).

mtl1979:
@RReverser I meant emulated SIMD support in WASM... It's technically same as using for example qemu to run binaries designed for other processors. Even if most of the API is stable, even small change in the compiler or bytecode translator can make big difference in performance, both positive and negative. 

Even though a lot of the processors currently in use are x86_64 (a.k.a AMD64) compatible, increasingly more are ARM-based, which means trying to emulate SSE doesn't make a lot of sense in long term.

RReverser:
> It's technically same as using for example qemu to run binaries designed for other processors

I'd argue it's an unfair comparison. Emscripten simply provides API-compatible implementation of intrinsics that forward operations to Wasm intrinsics where possible. That's about as fast as direct implementation except for where something is not yet implemented, as everything happens at compile time. QEMU comparison would require Emscripten to be able to emulate x86 instructions themselves, which it cannot do and which would be, yes, terrible for performance :)

I agree NEON could be an even more interesting path, but it's not yet well implemented in simd-everywhere / Emscripten for Wasm.

nmoinvaz:
Rebased.

Dead2 (approved):
LGTM

PR #1330, opened 2022-08-21
Remove ZLIB_DUAL_LINK option in favor of detecting ZLIB_LIBRARIES.

nmoinvaz:


mtl1979:
It would make more sense to check for presence of `ZLIB_ROOT` as it's the primary define for zlib detection. All other defines are optional due to using `find_package()`.

nmoinvaz:
Hmm.. We aren't passing in ZLIB_ROOT in cmake for the Link workflow..

nmoinvaz:
ZLIB_ROOT is probably only used for ZLIB installations, where in our case in the GHA workflows, we are building them from scratch.

nmoinvaz:
I improved the commit message.

mtl1979:
I still think the new version causes even more confusion than the original. Why use `find_package()` when the paths are supposed to always be hard-coded... This is pretty much counter-intuitive with depending on `ZLIB_LIBRARIES` being correctly defined.

The original version might not have been the best choice, but replacing one bad choice with another bad choice isn't really a good idea....

The safest choice would be to check if either `ZLIB_ROOT` or `ZLIB_LIBRARIES` is defined, but that would increase complexity and pretty much duplicate the logic already inside `find_package()`.

nmoinvaz:
I have removed `find_package()` so there isn't that ambiguity.

Dead2 (approved):
LGTM

PR #1331, opened 2022-08-21
Fixed content already populated error in CMake scripts.

nmoinvaz:
Should only need to use either `FetchContent_MakeAvailable` or
`FetchContent_GetProperties` and `FetchContent_Populate` but not both methods. We
use the later for CMake compatibility with lower versions.

### DIFF
--- a/test/pigz/CMakeLists.txt
+++ b/test/pigz/CMakeLists.txt
@@ -64,9 +64,8 @@ endif()
 FetchContent_Declare(pigz
     GIT_REPOSITORY https://github.com/madler/pigz.git
     GIT_TAG ${PIGZ_TAG})
-FetchContent_MakeAvailable(pigz)
-FetchContent_GetProperties(pigz)
 
+FetchContent_GetProperties(pigz)
 if(NOT pigz_POPULATED)
     FetchContent_Populate(pigz)
 endif()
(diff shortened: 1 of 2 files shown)


mtl1979:
`FetchContent_MakeAvailable` was introduced in cmake 3.14, but the minimum is set to lower, so it can fail.

nmoinvaz:
Fixed.

test/pigz/CMakeLists.txt
```diff
@@ -64,11 +64,11 @@ endif()
 FetchContent_Declare(pigz
     GIT_REPOSITORY https://github.com/madler/pigz.git
     GIT_TAG ${PIGZ_TAG})
-FetchContent_MakeAvailable(pigz)
-FetchContent_GetProperties(pigz)
 
+FetchContent_GetProperties(pigz)
 if(NOT pigz_POPULATED)
     FetchContent_Populate(pigz)
+    add_subdirectory(${pigz_SOURCE_DIR} ${pigz_BINARY_DIR})
```
mtl1979: This line seems to fail.

nmoinvaz: You are right, it is unneeded because we are using `FetchContent` only to retrieve the source code then we build manually because pigz has no CMakeLists.txt.

Dead2 (approved):
LGTM

PR #1335, opened 2022-08-25
Add vcpkg installation instructions

FrankXie05:
`zlib-ng` is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for `zlib-ng` and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build `zlib-ng`, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/zlib-ng/portfile.cmake). We try to keep the library maintained as close as possible to the original library. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,14 +161,16 @@ To install zlib-ng system-wide using the configure script:
 make install
 ```
 
-Alternatively, you can build and install zlib-ng using [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
+### vcpkg
+
+Alternatively, you can build and install zlib-ng using the [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
 
 ```sh or powershell
-    git clone https://github.com/Microsoft/vcpkg.git
-    cd vcpkg
-    ./bootstrap-vcpkg.sh # "./bootstrap-vcpkg.bat" for powershell
-    ./vcpkg integrate install
-    ./vcpkg install zlib-ng
+git clone https://github.com/Microsoft/vcpkg.git
+cd vcpkg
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -161,6 +161,18 @@ To install zlib-ng system-wide using the configure script:
 make install
 ```
 
+Alternatively, you can build and install zlib-ng using [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
+
+```sh or powershell
+    git clone https://github.com/Microsoft/vcpkg.git
```
nmoinvaz: Can you remove beginning whitespace from commands?

README.md
```diff
@@ -161,6 +161,18 @@ To install zlib-ng system-wide using the configure script:
 make install
 ```
 
+Alternatively, you can build and install zlib-ng using [vcpkg](https://github.com/Microsoft/vcpkg/) dependency manager:
```
nmoinvaz: We should have a header `### vcpkg` above this. So we would have CMake, configure, and vcpkg headings.

nmoinvaz: Add "the" after "using"

README.md
FrankXie05: Thanks a lot for your quick answer, I'll keep the format consistent with `zlib-ng`. :)

README.md
FrankXie05: Adding `vcpkg` head will be more intuitive.

Dead2 (approved):
LGTM

PR #1336, opened 2022-08-25
cmake: CPack enable

scivision:
CPack only works with relative install() paths.
This ignores the absolute paths in cache.

It may be that the absolute paths aren't needed anymore since this is built into CMake, unless custom install methods are needed (different absolute prefix for different artifacts).

With CPack, source and binary archives are easily generated from the build directory:

```sh
cpack
```


nmoinvaz:
Should this be incorporated into `release.yaml`?

scivision:
Yes I think CPack would simplify the release.yml. The yaml would just upload the single .zip file (or other archive formats) CPack would create.

Dead2:
Unfortunately this would break for example rpm spec files that already specify *_INSTALL_DIR, for example `BIN_INSTALL_DIR`. Especially redhat and derivatives use this in its macros, and the current names were chosen to be compatible. Maybe add some parsing of the possible input variables as well?

mtl1979:
> Unfortunately this would break for example rpm spec files that already specify *_INSTALL_DIR, for example `BIN_INSTALL_DIR`. Especially redhat and derivatives use this in its macros, and the current names were chosen to be compatible. Maybe add some parsing of the possible input variables as well?

We could easily enable CPack only if `LIB_INSTALL_DIR` isn't defined... We don't really need to check for other variables as library directory is most likely have a custom/different prefix. Other option would be to check for number of common path components at start of each install directory, but I'm not sure if there is such function in CMake, and then strip the common part.

Dead2:
Merged similar PR #1556

PR #1340, opened 2022-09-06
Fix build on FreeBSD/powerpc*

pkubaj:
Use elf_aux_info() instead of getauxval() on FreeBSD.

### DIFF
--- a/arch/power/power_features.c
+++ b/arch/power/power_features.c
@@ -7,6 +7,9 @@
 #ifdef HAVE_SYS_AUXV_H
 #  include <sys/auxv.h>
 #endif
+#ifdef __FreeBSD__
+#include <machine/cpu.h>
+#endif
 #include "../../zbuild.h"
 #include "power_features.h"
 
@@ -17,7 +20,11 @@ Z_INTERNAL int power_cpu_has_arch_3_00 = 0;
 void Z_INTERNAL power_check_features(void) {
 #ifdef PPC_FEATURES
     unsigned long hwcap;
+#ifdef __FreeBSD__
+    elf_aux_info(AT_HWCAP, &hwcap, sizeof(hwcap));
+#else
     hwcap = getauxval(AT_HWCAP);
+#endif
 
     if (hwcap & PPC_FEATURE_HAS_ALTIVEC)
         power_cpu_has_altivec = 1;
@@ -25,7 +32,11 @@ void Z_INTERNAL power_check_features(void) {
 
 #ifdef POWER_FEATURES
     unsigned long hwcap2;
+#ifdef __FreeBSD__
+    elf_aux_info(AT_HWCAP2, &hwcap2, sizeof(hwcap2));
+#else
     hwcap2 = getauxval(AT_HWCAP2);
+#endif
 
     if (hwcap2 & PPC_FEATURE2_ARCH_2_07)
         power_cpu_has_arch_2_07 = 1;
--- a/arch/power/power_features.c
+++ b/arch/power/power_features.c
@@ -8,7 +8,7 @@
 #  include <sys/auxv.h>
 #endif
 #ifdef __FreeBSD__
-#include <machine/cpu.h>
+#  include <machine/cpu.h>
 #endif
 #include "../../zbuild.h"
 #include "power_features.h"
--- a/configure
+++ b/configure
@@ -732,13 +732,20 @@ else
   echo "Checking for strerror... No." | tee -a configure.log
 fi
 
-# check for getauxval() for architecture feature detection at run-time
+# check for getauxval() or elf_aux_info() for architecture feature detection at run-time
 cat > $test.c <<EOF
 #include <sys/auxv.h>
-int main() { return getauxval(0); }
+int main() {
+#ifdef __FreeBSD__
+int test;
+return elf_aux_info(AT_PAGESZ, &test, sizeof(test));
+#else
+return getauxval(0);
+#endif
+}
 EOF
 if try $CC $CFLAGS -o $test $test.c $LDSHAREDLIBC; then
-  echo "Checking for getauxval() in sys/auxv.h... Yes." | tee -a configure.log
+  echo "Checking for getauxval() or elf_aux_info() in sys/auxv.h... Yes." | tee -a configure.log
   CFLAGS="${CFLAGS} -DHAVE_SYS_AUXV_H"
   SFLAGS="${SFLAGS} -DHAVE_SYS_AUXV_H"
 else
@@ -1219,8 +1226,19 @@ EOF
             echo "Checking if -mno-vsx is supported ... No." | tee -a configure.log
         fi
         cat > $test.c << EOF
+#ifdef __FreeBSD__
+#include <machine/cpu.h>
+#endif
 #include <sys/auxv.h>
-int main() { return (getauxval(AT_HWCAP) & PPC_FEATURE_HAS_ALTIVEC); }
+int main() {
+#ifdef __FreeBSD__
+unsigned long hwcap;
+elf_aux_info(AT_HWCAP, &hwcap, sizeof(hwcap));
+return (hwcap & PPC_FEATURE_HAS_ALTIVEC);
+#else
+return (getauxval(AT_HWCAP) & PPC_FEATURE_HAS_ALTIVEC);
+#endif
+}
 EOF
         if try $CC -c $CFLAGS -maltivec $test.c; then
             HAVE_VMX=1
@@ -1234,8 +1252,19 @@ EOF
 check_power8_intrinsics() {
     # Check whether features needed by POWER8 optimisations are available
     cat > $test.c << EOF
+#ifdef __FreeBSD__
+#include <machine/cpu.h>
+#endif
 #include <sys/auxv.h>
-int main() { return (getauxval(AT_HWCAP2) & PPC_FEATURE2_ARCH_2_07); }
+int main() {
+#ifdef __FreeBSD__
+unsigned long hwcap;
+elf_aux_info(AT_HWCAP2, &hwcap, sizeof(hwcap));
+return (hwcap & PPC_FEATURE2_ARCH_2_07);
+#else
+return (getauxval(AT_HWCAP2) & PPC_FEATURE2_ARCH_2_07);
+#endif
+}
 EOF
     if test $buildpower8 -eq 1 && try $CC -c $CFLAGS -mcpu=power8 $test.c; then
         HAVE_POWER8_INTRIN=1
--- a/configure
+++ b/configure
@@ -737,10 +737,10 @@ cat > $test.c <<EOF
 #include <sys/auxv.h>
 int main() {
 #ifdef __FreeBSD__
-int test;
-return elf_aux_info(AT_PAGESZ, &test, sizeof(test));
+  int test;
+  return elf_aux_info(AT_PAGESZ, &test, sizeof(test));
 #else
-return getauxval(0);
+  return getauxval(0);
 #endif
 }
 EOF
@@ -1232,11 +1232,11 @@ EOF
 #include <sys/auxv.h>
 int main() {
 #ifdef __FreeBSD__
-unsigned long hwcap;
-elf_aux_info(AT_HWCAP, &hwcap, sizeof(hwcap));
-return (hwcap & PPC_FEATURE_HAS_ALTIVEC);
+  unsigned long hwcap;
+  elf_aux_info(AT_HWCAP, &hwcap, sizeof(hwcap));
+  return (hwcap & PPC_FEATURE_HAS_ALTIVEC);
 #else
-return (getauxval(AT_HWCAP) & PPC_FEATURE_HAS_ALTIVEC);
+  return (getauxval(AT_HWCAP) & PPC_FEATURE_HAS_ALTIVEC);
 #endif
 }
 EOF
@@ -1258,11 +1258,11 @@ check_power8_intrinsics() {
 #include <sys/auxv.h>
 int main() {
 #ifdef __FreeBSD__
-unsigned long hwcap;
-elf_aux_info(AT_HWCAP2, &hwcap, sizeof(hwcap));
-return (hwcap & PPC_FEATURE2_ARCH_2_07);
+  unsigned long hwcap;
+  elf_aux_info(AT_HWCAP2, &hwcap, sizeof(hwcap));
+  return (hwcap & PPC_FEATURE2_ARCH_2_07);
 #else
-return (getauxval(AT_HWCAP2) & PPC_FEATURE2_ARCH_2_07);
+  return (getauxval(AT_HWCAP2) & PPC_FEATURE2_ARCH_2_07);
 #endif
 }
 EOF
--- a/cmake/detect-intrinsics.cmake
+++ b/cmake/detect-intrinsics.cmake
@@ -289,8 +289,17 @@ macro(check_ppc_intrinsics)
     set(CMAKE_REQUIRED_FLAGS "${PPCFLAGS} ${NATIVEFLAG}")
     check_c_source_compiles(
         "#include <sys/auxv.h>
+        #ifdef __FreeBSD__
+        #include <machine/cpu.h>
+        #endif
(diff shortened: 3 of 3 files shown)


mtl1979:
`cmake/detect-intrinsics.cmake` has `getauxval()` hardcoded in some tests... That file needs to be fixed also.
`configure` has no test for `elf_aux_info()` either.

pkubaj:
`configure` doesn't run correctly for me anyway because of GNU sed vs BSD sed issues. I think it will be ok to leave `configure` as is.
I have added a new commit for the intrinsics test. IMO it's ok not  to test for `elf_aux_info`, it's been available since FreeBSD 12.0-RELEASE and all the supported releases have it.

mtl1979:
I've used *BSD since 1990s and I know `sed` isn't the only command-line tool that behaves differently... However we do already support MacOS which is closer to BSD than GNU. I'm currently running FreeBSD 12, 13 and 14.

pkubaj:
I'm getting the following:
```
root@talos-powerpc64le:~/zlib-ng # ./configure                                                                                                                                                                                       
Checking for compiler... cc                                                                                           
Checking for shared library support... Building shared library libz-ng.so.2.1.0.devel with cc.
Checking for off64_t... Yes.                                                                                          
Checking for fseeko... Yes.                                                                                           
Checking for posix_memalign... Yes.                                                                                   
Checking for strerror... Yes.                                                                                         
Checking for getauxval() in sys/auxv.h... No.                                                                         
Checking for unistd.h... Yes.                                                                                         
Checking for ptrdiff_t... Yes.                                                                                        
Checking for ANSI C compliant compiler...  Yes.                                                                       
Checking for -fno-semantic-interposition... No.                                                                       
Checking for __builtin_ctz ... Yes.                                                                                   
Checking for __builtin_ctzll ... Yes.                                                                                 
Checking for AltiVec intrinsics ... Yes.                                                                              
Checking if -mno-vsx is supported ... Yes.                                                                            
Check whether VMX instructions are available ... No.                                                                  
Check whether POWER8 instructions are available ... No.                                                               
Check whether POWER9 instructions are available ... Yes.                                                              
ARCH: powerpc                                                                                                         
Using arch directory: arch/power                                                                                      
Architecture-specific static object files: compare256_power9.o         
Architecture-specific shared object files: compare256_power9.lo        
sed: 1: "s/# *\include *"\(.*\.h ...": RE error: trailing backslash (\)
```

pkubaj:
I fixed SIMD check in `configure` similarly as the CMake code with:
```
diff --git a/configure b/configure
index 4e73a22..91adc1a 100755
--- a/configure
+++ b/configure
@@ -29,7 +29,7 @@ if [ -n "${CHOST}" ]; then
     CROSS_PREFIX="${CHOST}-"
     ARCH="$(echo "${NORM_CHOST}" | sed -e 's/-.*//')"
 else
-    ARCH="$(uname -m)"
+    ARCH="$(uname -p)"
 fi

 case "${ARCH}" in
@@ -1219,8 +1219,19 @@ EOF
             echo "Checking if -mno-vsx is supported ... No." | tee -a configure.log
         fi
         cat > $test.c << EOF
+#ifdef __FreeBSD__
+#include <machine/cpu.h>
+#endif
 #include <sys/auxv.h>
-int main() { return (getauxval(AT_HWCAP) & PPC_FEATURE_HAS_ALTIVEC); }
+int main() {
+#ifdef __FreeBSD__
+unsigned long hwcap;
+elf_aux_info(AT_HWCAP, &hwcap, sizeof(hwcap));
+return (hwcap & PPC_FEATURE_HAS_ALTIVEC);
+#else
+return (getauxval(AT_HWCAP) & PPC_FEATURE_HAS_ALTIVEC);
+#endif
+}
 EOF
         if try $CC -c $CFLAGS -maltivec $test.c; then
             HAVE_VMX=1
@@ -1234,8 +1245,19 @@ EOF
 check_power8_intrinsics() {
     # Check whether features needed by POWER8 optimisations are available
     cat > $test.c << EOF
+#ifdef __FreeBSD__
+#include <machine/cpu.h>
+#endif
 #include <sys/auxv.h>
-int main() { return (getauxval(AT_HWCAP2) & PPC_FEATURE2_ARCH_2_07); }
+int main() {
+#ifdef __FreeBSD__
+unsigned long hwcap;
+elf_aux_info(AT_HWCAP2, &hwcap, sizeof(hwcap));
+return (hwcap & PPC_FEATURE2_ARCH_2_07);
+#else
+return (getauxval(AT_HWCAP2) & PPC_FEATURE2_ARCH_2_07);
+#endif
+}
 EOF
     if test $buildpower8 -eq 1 && try $CC -c $CFLAGS -mcpu=power8 $test.c; then
         HAVE_POWER8_INTRIN=1
```

But it still fails to run:
```
root@talos-powerpc64le:~/zlib-ng # ./configure
Checking for compiler... cc
Checking for shared library support... Building shared library libz-ng.so.2.1.0.devel with cc.
Checking for off64_t... Yes.
Checking for fseeko... Yes.
Checking for posix_memalign... Yes.
Checking for strerror... Yes.
Checking for getauxval() in sys/auxv.h... No.
Checking for unistd.h... Yes.
Checking for ptrdiff_t... Yes.
Checking for ANSI C compliant compiler...  Yes.
Checking for -fno-semantic-interposition... No.
Checking for __builtin_ctz ... Yes.
Checking for __builtin_ctzll ... Yes.
Checking for AltiVec intrinsics ... Yes.
Checking if -mno-vsx is supported ... Yes.
Check whether VMX instructions are available ... Yes.
Check whether POWER8 instructions are available ... Yes.
Check whether POWER9 instructions are available ... Yes.
ARCH: powerpc64le
Using arch directory: arch/power
Architecture-specific static object files: power_features.o adler32_vmx.o slide_hash_vmx.o adler32_power8.o chunkset_power8.o slide_hash_power8.o crc32_power8.o compare256_power9.o
Architecture-specific shared object files: power_features.lo adler32_vmx.lo slide_hash_vmx.lo adler32_power8.lo chunkset_power8.lo slide_hash_power8.lo crc32_power8.lo compare256_power9.lo
sed: 1: "s/# *\include *"\(.*\.h ...": RE error: trailing backslash (\)
```

Btw, your architecture check in configure is wrong for FreeBSD, `uname -p` should be used instead of `uname -m`.

pkubaj:
After changing `sed` to `gsed` in `incs=$(grep -h include $file | sed -n 's/# *\include *"\(.*\.h\)".*/\1/p' | sort | uniq)`, `configure` runs properly.

configure
```diff
@@ -732,13 +732,18 @@ else
...
+#ifdef __FreeBSD__
+#include <machine/cpu.h>
+int main() { unsigned long hwcap; return elf_aux_info(AT_HWCAP, &hwcap, sizeof(hwcap)); }
+#else
 int main() { return getauxval(0); }
+#endif
 EOF
 if try $CC $CFLAGS -o $test $test.c $LDSHAREDLIBC; then
-  echo "Checking for getauxval() in sys/auxv.h... Yes." | tee -a configure.log
+  echo "Checking for getauxval() or elf_aux_info() in sys/auxv.h... Yes." | tee -a configure.log
```
mtl1979, line 748: Now `in sys/auxv.h` would be misleading... This might be better separated as own code block to avoid confusion.

mtl1979:
> After changing `sed` to `gsed` in `incs=$(grep -h include $file | sed -n 's/# *\include *"\(.*\.h\)".*/\1/p' | sort | uniq)`, `configure` runs properly.

I don't know why it uses `\i` but BSD sed doesn't support that escape code. Best solution would be to just remove the backslash as for me it looks superfluous.

configure
pkubaj: Why is it misleading? `elf_aux_info` is literally in `sys/auxv.h`.

mtl1979: Because for other platforms it is supposed to test if `sys/auxv.h` includes all necessary headers automatically, which isn't always true even for Linux.

pkubaj: Please check now, it doesn't require any headers other than `sys/auxv.h` anymore.

mtl1979:
I think @Dead2 should add CI instance that supports FreeBSD... I've used them for other projects, so I know there is.

configure
mtl1979: I will check next time I boot FreeBSD 13... We still need @Dead2 to enable the CI tests so we know this PR doesn't break build on other platforms.

mtl1979:
Fails on `crc32_fold.c` and `adler32_fold.c` on line containing `memcpy(dst, src, SIZE_MAX)` complaining about last parameter both overflowing and overlapping. These code blocks might need special handling on 32-bit systems. Limit for `memcpy()` is `SSIZE_T`, which is defined in `sys/limits.h`.

I used powerpc-gcc9 with gmake on FreeBSD 13.1.

pkubaj:
How do you check for that? I had the following issue on powerpc (I only tried on powerpc64le before):
```
ld: error: relocation R_PPC_ADDR16_LO cannot be used against symbol _GLOBAL_OFFSET_TABLE_; recompile with -fPIC                                                                                                                             
>>> referenced by adler32.c                                                                                                                                                                                                                 
>>>               adler32.lo:(zng_adler32_z)                                                                                                                                                                                                
```

Recompiling with -fPIC doesn't fix it.

pkubaj:
I was able to build after installing binutils and adding `-fuse-ld=bfd` to CFLAGS. However:
```
# gmake CFLAGS="-fuse-ld=bfd" test
gmake -C test
gmake[1]: Entering directory '/tmp/zlib-ng/test'
hello world
zlib-ng version 2.1.0.devel = 0x02010000, compile flags = 0x95
uncompress(): hello, hello!
gzread(): hello, hello!
gzgets() after gzseek:  hello!
gzgets(): hello, hello!
inflate(): hello, hello!
large_inflate(): OK
large_inflate(): OK
after inflateSync(): hello, hello!
inflate with dictionary: hello, hello!
deflateBound(): OK
deflate_copy(): OK
deflateGetDictionary(): hello, hello!
deflateSetHeader(): OK
deflateTune(): OK
deflatePending(): OK
deflatePrime(): OK
                *** zlib test OK ***
ld-elf.so.1: /tmp/zlib-ng/test/../libz-ng.so.2.1.0.devel: Unsupported relocation type 4 in non-PLT relocations

                *** zlib shared test FAILED ***
gmake[1]: *** [Makefile:39: testshared] Error 1
gmake[1]: Leaving directory '/tmp/zlib-ng/test'
gmake: *** [Makefile:180: test] Error 2
```

mtl1979:
@pkubaj I had to set both `CFLAGS` and `LDFLAGS` before running configure.

nmoinvaz:
Is there a reason why the configure detection code doesn't have the proper formatting?

pkubaj:
What is the proper formatting?

configure
```diff
@@ -732,13 +732,20 @@ else
...
 fi
 
-# check for getauxval() for architecture feature detection at run-time
+# check for getauxval() or elf_aux_info() for architecture feature detection at run-time
 cat > $test.c <<EOF
 #include <sys/auxv.h>
-int main() { return getauxval(0); }
+int main() {
+#ifdef __FreeBSD__
+int test;
```
nmoinvaz: Missing ident these two lines.

configure
```diff
@@ -732,13 +732,20 @@ else
...
+return elf_aux_info(AT_PAGESZ, &test, sizeof(test));
+#else
+return getauxval(0);
```
nmoinvaz: Missing indent.

configure
```diff
@@ -1219,8 +1226,19 @@ EOF
...
         fi
         cat > $test.c << EOF
+#ifdef __FreeBSD__
+#include <machine/cpu.h>
+#endif
 #include <sys/auxv.h>
-int main() { return (getauxval(AT_HWCAP) & PPC_FEATURE_HAS_ALTIVEC); }
+int main() {
+#ifdef __FreeBSD__
+unsigned long hwcap;
```
nmoinvaz: Missing indent next 3 lines.

configure
```diff
@@ -1219,8 +1226,19 @@ EOF
...
+elf_aux_info(AT_HWCAP, &hwcap, sizeof(hwcap));
+return (hwcap & PPC_FEATURE_HAS_ALTIVEC);
+#else
+return (getauxval(AT_HWCAP) & PPC_FEATURE_HAS_ALTIVEC);
```
nmoinvaz: Missing indent.

configure
```diff
@@ -1234,8 +1252,19 @@ EOF
...
     # Check whether features needed by POWER8 optimisations are available
     cat > $test.c << EOF
+#ifdef __FreeBSD__
+#include <machine/cpu.h>
+#endif
 #include <sys/auxv.h>
-int main() { return (getauxval(AT_HWCAP2) & PPC_FEATURE2_ARCH_2_07); }
+int main() {
+#ifdef __FreeBSD__
+unsigned long hwcap;
```
nmoinvaz: Missing indent next 3 lines..

configure
```diff
@@ -1234,8 +1252,19 @@ EOF
...
+elf_aux_info(AT_HWCAP2, &hwcap, sizeof(hwcap));
+return (hwcap & PPC_FEATURE2_ARCH_2_07);
+#else
+return (getauxval(AT_HWCAP2) & PPC_FEATURE2_ARCH_2_07);
```
nmoinvaz: Missing indent.

nmoinvaz:
@pkubaj I have commented about some indents that should make it easier to read. I think perhaps in a previous version it was indented better but maybe it got lost, that is why I was wondering if there was a technical reason it changed, or maybe I am remembering it differently..

configure
pkubaj: done

configure
pkubaj: done

configure
pkubaj: done

configure
pkubaj: done

configure
pkubaj: done

pkubaj:
I changed the previous version since now the FreeBSD variant is added, hence also the necessity for #ifdef.

pkubaj:
Also, the code works fine for powerpc64 and powerpc64le targets. Can it be merged since most people care about 64-bits nowadays?

Dead2:
CI reveals PPC ABI warnings like this:
```
  [C]'function uint32_t adler32_power8(uint32_t, const unsigned char*, size_t)' at adler32_power8.c:55:1 has some indirect sub-type changes:
    parameter 3 of type 'typedef size_t' changed:
      typedef name changed from size_t to uint64_t at stdint-uintn.h:27:1
      underlying type 'unsigned int' changed:
        entity changed from 'unsigned int' to compatible type 'typedef __uint64_t' at types.h:48:1
          type name changed from 'unsigned int' to 'long long unsigned int'
          type size changed from 32 to 64 (in bits)
```

mtl1979:
@Dead2 This PR doesn't touch that file or any header files included by that file. The ABI files are out of date and need to be updated to a new base commit, but that should be done in separate PR, preferably by same person who has previously updated the ABI check script.

Dead2 (approved):
LGTM

arch/power/power_features.c
```diff
@@ -7,6 +7,9 @@
 #ifdef HAVE_SYS_AUXV_H
 #  include <sys/auxv.h>
 #endif
+#ifdef __FreeBSD__
+#include <machine/cpu.h>
```
nmoinvaz: Should have two space indent.

nmoinvaz (approved):

arch/power/power_features.c
pkubaj: Done.

nmoinvaz (approved):

jasonmader:
> > After changing `sed` to `gsed` in `incs=$(grep -h include $file | sed -n 's/# *\include *"\(.*\.h\)".*/\1/p' | sort | uniq)`, `configure` runs properly.
> 
> I don't know why it uses `\i` but BSD sed doesn't support that escape code. Best solution would be to just remove the backslash as for me it looks superfluous.

What does `\i` do in this command anyway?

PR #1341, opened 2022-09-06
msvc/armv7: disable crc32_acle

shuffle2:
msvc compiler targeting 32bit arm supports
only armv7 and lacks these intrinsics

### DIFF
--- a/cmake/detect-intrinsics.cmake
+++ b/cmake/detect-intrinsics.cmake
@@ -2,29 +2,35 @@
 # Licensed under the Zlib license, see LICENSE.md for details
 
 macro(check_acle_compiler_flag)
-    if(CMAKE_C_COMPILER_ID MATCHES "GNU" OR CMAKE_C_COMPILER_ID MATCHES "Clang")
-        if(NOT NATIVEFLAG AND NOT HAVE_ACLE_FLAG)
-            set(ACLEFLAG "-march=armv8-a+crc" CACHE INTERNAL "Compiler option to enable ACLE support")
+    if(MSVC)
+        # Both ARM and ARM64-targeting msvc support intrinsics, but
+        # ARM msvc is missing some intrinsics introduced with ARMv8, e.g. crc32
+        if(MSVC_C_ARCHITECTURE_ID STREQUAL "ARM64")
+            set(HAVE_ACLE_FLAG TRUE)
         endif()
-    elseif(MSVC)
-        set(HAVE_ACLE_FLAG TRUE)
-    endif()
-    # Check whether compiler supports ACLE flag
-    set(CMAKE_REQUIRED_FLAGS "${ACLEFLAG} ${NATIVEFLAG}")
-    check_c_source_compiles(
-        "int main() { return 0; }"
-        HAVE_ACLE_FLAG FAIL_REGEX "not supported")
-    if(NOT NATIVEFLAG AND NOT HAVE_ACLE_FLAG)
-        set(ACLEFLAG "-march=armv8-a+crc+simd" CACHE INTERNAL "Compiler option to enable ACLE support" FORCE)
+    else()
+        if(CMAKE_C_COMPILER_ID MATCHES "GNU" OR CMAKE_C_COMPILER_ID MATCHES "Clang")
+            if(NOT NATIVEFLAG AND NOT HAVE_ACLE_FLAG)
+                set(ACLEFLAG "-march=armv8-a+crc" CACHE INTERNAL "Compiler option to enable ACLE support")
+            endif()
+        endif()
         # Check whether compiler supports ACLE flag
-        set(CMAKE_REQUIRED_FLAGS "${ACLEFLAG}")
+        set(CMAKE_REQUIRED_FLAGS "${ACLEFLAG} ${NATIVEFLAG}")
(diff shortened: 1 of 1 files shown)


shuffle2:
the fixed build is this one: https://github.com/zlib-ng/zlib-ng/runs/8215482281?check_suite_focus=true#logs

CMakeLists.txt
```diff
@@ -587,7 +587,7 @@ if(WITH_OPTIM)
         endif()
         list(APPEND ZLIB_ARCH_HDRS ${ARCHDIR}/arm_features.h)
         list(APPEND ZLIB_ARCH_SRCS ${ARCHDIR}/arm_features.c)
-        if(WITH_ACLE AND NOT "${ARCH}" MATCHES "armv[2-7]")
+        if(WITH_ACLE AND NOT (MSVC AND "${ARCH}" MATCHES "arm") AND NOT "${ARCH}" MATCHES "armv[2-7]")
```
nmoinvaz: Why not add the check into `check_acle_compiler_flag` instead?

nmoinvaz: I don't think we should even be having `AND NOT "${ARCH}" MATCHES "armv[2-7]"` on this if statement. We can let the compiler to run the check for the flag. Because otherwise the `WITH_ACLE` flag is never turned off when printing used features.

nmoinvaz: The check inside of `check_acle_compiler_flag` can use `if(MSVC AND MSVC_C_ARCHITECTURE_ID STREQUAL "ARM64")`

shuffle2: I didn't want to put the check in `check_acle_compiler_flag` because I only want to disable using intrinsics for crc32 instructions. "ACLE" could technically be used for other features (even though that's not currently the case). msvc armv7 does support intrinsics (armv7, not A32).

"ACLE" is a bit of a misnomer on msvc, but it's still the term zlib-ng uses to detect support for intrinsics.

Do you still think it should be inside `check_acle_compiler_flag`?

nmoinvaz: > Do you still think it should be inside check_acle_compiler_flag?

_I_ think so and it would be a good to add a comment to that if statement in `check_acle_compiler_flag` like this:
```
# Missing crc32b instructions in Windows intrinsics headers for ARM 32-bit.
```

nmoinvaz: > "ACLE" is a bit of a misnomer on msvc, but it's still the term zlib-ng uses to detect support for intrinsics.

I agree, we could address that separately when we end up using more instructions.

shuffle2: How about this?
```cmake
macro(check_acle_compiler_flag)
    if(MSVC)
        # Both ARM and ARM64-targeting msvc support intrinsics, but
        # ARM msvc is missing some intrinsics introduced with ARMv8, e.g. crc32
        if(MSVC_C_ARCHITECTURE_ID STREQUAL "ARM64")
            set(HAVE_ACLE_FLAG TRUE)
        endif()
    else()
            if(NOT NATIVEFLAG AND NOT HAVE_ACLE_FLAG)
                set(ACLEFLAG "-march=armv8-a+crc" CACHE INTERNAL "Compiler option to enable ACLE support")
            endif()
        endif()
        # Check whether compiler supports ACLE flag
        set(CMAKE_REQUIRED_FLAGS "${ACLEFLAG} ${NATIVEFLAG}")
        check_c_source_compiles(
            "int main() { return 0; }"
            HAVE_ACLE_FLAG FAIL_REGEX "not supported")
        if(NOT NATIVEFLAG AND NOT HAVE_ACLE_FLAG)
            set(ACLEFLAG "-march=armv8-a+crc+simd" CACHE INTERNAL "Compiler option to enable ACLE support" FORCE)
            # Check whether compiler supports ACLE flag
            set(CMAKE_REQUIRED_FLAGS "${ACLEFLAG}")
            check_c_source_compiles(
                "int main() { return 0; }"
                HAVE_ACLE_FLAG2 FAIL_REGEX "not supported")
            set(HAVE_ACLE_FLAG ${HAVE_ACLE_FLAG2} CACHE INTERNAL "Have compiler option to enable ACLE intrinsics" FORCE)
            unset(HAVE_ACLE_FLAG2 CACHE) # Don't cache this internal variable
        endif()
        set(CMAKE_REQUIRED_FLAGS)
    endif()
endmacro()
```
It seems required to avoid letting msvc hit `check_c_source_compiles`, because it always reports success (at least for the tests done within `check_acle_compiler_flag`). The tests (as written with current code/flags) don't really make sense to do with msvc anyway.

nmoinvaz: Looks good. No need for the outer `else()`. You return early instead:
```
if(MSVC AND MSVC_C_ARCHITECTURE_ID STREQUAL "ARM64")
    set(HAVE_ACLE_FLAG TRUE)
    return()
endif()
```

shuffle2: `return` does not appear to work from a `macro`...the macro gets inlined into the using code, and the return winds up early-returning from wherever that happens to be.

nmoinvaz: In that case it looks ok to me.

nmoinvaz (approved):

Dead2 (approved):
LGTM

PR #1346, opened 2022-09-11
Fixed benchmarks not building

nmoinvaz:
This fixes **Ubuntu GCC Benchmark** CI.

Google benchmark repository removed the `master` tag. I think CMake FetchContent defaults to `GIT_TAG=master` so that is possibly why the failures.


Dead2 (approved):
LGTM

PR #1348, opened 2022-09-14
Various small buildfixes

Dead2:
Found various small issues while doing a lot of manual testing with Clang and [Stabilizer](https://github.com/Dead2/stabilizer)


configure
```diff
@@ -581,16 +581,16 @@ if ($CC -c $CFLAGS $test.c) 2>/dev/null; then
...
+    ( $* ) >> configure.log 2>&1
+    ret=$?
+    if test $ret -ne 0; then
+      echo "(exit code $ret)" >> configure.log
+    fi
+    return $ret
+  }
 fi
 
 tryboth()
```
mtl1979: If `tryboth()` is no longer needed or used, it should be removed. It still stays in the git history if it is later needed again...

PR #1349, opened 2022-09-17
IBM Z DFLTCC: Fix updating strm.adler with inflate()

iii-i:
inflate() does not update strm.adler with DFLTCC.
Add a missing assignment to dfltcc_inflate() to fix this.
Note that deflate() is not affected.
Also add a test to prevent regressions.


Dead2 (approved):
LGTM

PR #1352, opened 2022-09-28
Introduce ZLIB_COMPAT_TESTS

iii-i:
This patch adds the ability to run zlib-ng test suite against the original zlib as follows:

    cmake -DZLIB_COMPAT=ON -DZLIB_COMPAT_TESTS=ON
    LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu make test

The benefit of this is that modifications to the original zlib can be tested with a more extensive zlib-ng's testsuite, and the assumptions that the zlib-ng tests make can be validated against the original zlib.

The following tests fail when run against the original zlib:

- example
- infcover
- test_deflate_prime.cc
- test_inflate_sync.cc

This needs to be investigated, but for now I just indicated the status quo by disabling them when building with ZLIB_COMPAT_TESTS.


test/CMakeLists.txt
```diff
@@ -200,7 +212,12 @@ else()
         target_link_libraries(gtest_zlib external_zlib)
     endif()
 
-    target_link_libraries(gtest_zlib zlibstatic GTest::GTest)
+    if(ZLIB_COMPAT_TESTS)
+        target_link_libraries(gtest_zlib zlib GTest::Main)
```
nmoinvaz, line 219: Why switch between zlib and zlibstatic? Both libraries exist in zlib's CMake.

CMakeLists.txt
```diff
@@ -76,6 +76,7 @@ endif()
 option(WITH_GZFILEOP "Compile with support for gzFile related functions" ON)
 option(ZLIB_COMPAT "Compile with zlib compatible API" OFF)
 option(ZLIB_ENABLE_TESTS "Build test binaries" ON)
+option(ZLIB_COMPAT_TESTS "Test only zlib compatible API" OFF)
```
nmoinvaz: The option name is a bit confusing. What do you think of something like `ZLIBNG_ENABLE_TESTS` and defaulted to `ON`?

nmoinvaz (commented):
I think we should investigate why those other tests are failing before moving forward with this PR, otherwise it should be noted with a comment why the tests are wrapped in `if()` statement which disable them when `ZLIB_COMPAT_TESTS`.

test/CMakeLists.txt
mtl1979: There shouldn't be any reason to change this block of code... Even if there was, there should be at least a comment to make it obvious why changing static to "possibly" dynamic linkage is required or warranted.

CMakeLists.txt
mtl1979: Yeah... It kinda implies the opposite of what is documented... As if it enables testing conformity of zlib compatible API.

iii-i: Ok, I'll rename it.

test/CMakeLists.txt
iii-i: It's needed in order to be able to replace zlib-ng with zlib at runtime. Of course, this may be defined as a separate option, but `ZLIB_COMPAT_TESTS=ON`/`ZLIBNG_ENABLE_TESTS=OFF` is not too useful without it.

Edit: another possibility is to link statically with zlib. However, this is quite unintuitive: we specify `ZLIBNG_ENABLE_TESTS=OFF` and suddenly `make test` don't invoke zlib-ng we just built at all. I think that trying to link dynamically strikes a good balance here.

The only thing is that we might need to guard it more strictly. It requires at least `ZLIB_COMPAT` and `BUILD_SHARED_LIBS`.

iii-i:
> I think we should investigate why those other tests are failing before moving forward with this PR, otherwise it should be noted with a comment why the tests are wrapped in `if()` statement which disable them when `ZLIB_COMPAT_TESTS`.

Ok, I'll investigate the failures and add the comments.

test/CMakeLists.txt
nmoinvaz: I think adding comment would a good idea then, and like you said, you may need to take into account the value of `BUILD_SHARED_LIBS`.

iii-i: Ok, I've implemented this in the commit I pushed.

mtl1979:
Seems to fail in declaration of `deflate_prime_32()`.

`test/test_compress_dual.cc` needs to include `zbuild.h` when using `PREFIX*`  macros.

mtl1979 (changes requested):
Fix build failure.

iii-i:
I moved `deflate_prime_32()` to a new header. `test_compress_dual` seems to be designed to work with the vanilla zlib build environment, so let's keep it this way.

test/test_shared_ng.h
```diff
@@ -0,0 +1,23 @@
+#ifndef TEST_SHARED_NG_H
+#define TEST_SHARED_NG_H
+
+/* Test definitions that can be used in the zlib-ng build environment. */
```
mtl1979: I'm not sure, but "can only be used" would be more accurate due to including `test_shared.h` right below... Also would be more accurate if placed after that include statement.

iii-i: Done.

iii-i:
Is there anything else I need to change here to get this accepted?

`Link / Link zlib-ng compat` and `Package Check / Ubuntu GCC PPC` are failing on the latest develop too.

nmoinvaz (approved):
LGTM

Issue #1347, opened 2022-09-13
busybox gz can't recognize --decompress


Aleksanaa:
```
          9 - minigzip-fireworks.jpg-R (Failed)
         10 - minigzip-fireworks.jpg-h (Failed)
         12 - minigzip-fireworks.jpg-0 (Failed)
         13 - minigzip-fireworks.jpg-1 (Failed)
         14 - minigzip-fireworks.jpg-4 (Failed)
         15 - minigzip-fireworks.jpg-5 (Failed)
         16 - minigzip-fireworks.jpg-F (Failed)
         17 - minigzip-fireworks.jpg-6 (Failed)
         18 - minigzip-fireworks.jpg-9 (Failed)
         19 - minigzip-fireworks.jpg-f (Failed)
         20 - minigzip-lcet10.txt-R (Failed)
         21 - minigzip-lcet10.txt-h (Failed)
         23 - minigzip-lcet10.txt-0 (Failed)
         24 - minigzip-lcet10.txt-1 (Failed)
         25 - minigzip-lcet10.txt-4 (Failed)
         26 - minigzip-lcet10.txt-5 (Failed)
         27 - minigzip-lcet10.txt-F (Failed)
         28 - minigzip-lcet10.txt-6 (Failed)
         29 - minigzip-lcet10.txt-9 (Failed)
         30 - minigzip-lcet10.txt-f (Failed)
         31 - minigzip-paper-100k.pdf-R (Failed)
         32 - minigzip-paper-100k.pdf-h (Failed)
         34 - minigzip-paper-100k.pdf-0 (Failed)
         35 - minigzip-paper-100k.pdf-1 (Failed)
         36 - minigzip-paper-100k.pdf-4 (Failed)
         37 - minigzip-paper-100k.pdf-5 (Failed)
         38 - minigzip-paper-100k.pdf-F (Failed)
         39 - minigzip-paper-100k.pdf-6 (Failed)
         40 - minigzip-paper-100k.pdf-9 (Failed)
         41 - minigzip-paper-100k.pdf-f (Failed)
         42 - minigzip-detect-text-A (Failed)
         43 - minigzip-detect-binary-A (Failed)
         54 - GH-361 (Failed)
         55 - GH-364 (Failed)
         60 - GH-751 (Failed)

```
They've all failed with `43: /usr/bin/gzip: unrecognized option: decompress`

nmoinvaz:
It supports `-d` only?

[...]

PR #1353, opened 2022-09-28
Use short decompress option name for gzip compatibility. 

nmoinvaz:
Long option names in BusyBoxy is an optional feature, so use short option names by default.
See #1347


mtl1979 (approved):
LGTM

PR #1358, opened 2022-10-05
Fix memcpy() overflow in adler32_fold.c and crc32_fold.c

mtl1979:
* On 32-bit platforms, last parameter of memcpy() is limited to SSIZE_MAX, but is likely to overlap if used

### DIFF
--- a/adler32_fold.c
+++ b/adler32_fold.c
@@ -7,14 +7,19 @@
 #include "functable.h"
 #include "adler32_fold.h"
 
+#include <limits.h>
+
 Z_INTERNAL uint32_t adler32_fold_copy_c(uint32_t adler, uint8_t *dst, const uint8_t *src, uint64_t len) {
     adler = functable.adler32(adler, src, len);
-    while (len > SIZE_MAX) {
-        memcpy(dst, src, SIZE_MAX);
-        dst += SIZE_MAX;
-        src += SIZE_MAX;
-        len -= SIZE_MAX;
+/* Test that we don't try to copy more than actually fits in available address space */
+#if INTPTR_MAX > SSIZE_MAX
+    while (len > SSIZE_MAX) {
+        memcpy(dst, src, SSIZE_MAX);
+        dst += SSIZE_MAX;
+        src += SSIZE_MAX;
+        len -= SSIZE_MAX;
     }
+#endif
     if (len) {
         memcpy(dst, src, (size_t)len);
     }
--- a/crc32_fold.c
+++ b/crc32_fold.c
@@ -7,6 +7,8 @@
 
 #include "crc32_fold.h"
 
+#include <limits.h>
+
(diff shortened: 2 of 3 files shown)


adler32_fold.c
```diff
@@ -7,14 +7,18 @@
...
+#include <limits.h>
+
 Z_INTERNAL uint32_t adler32_fold_copy_c(uint32_t adler, uint8_t *dst, const uint8_t *src, uint64_t len) {
     adler = functable.adler32(adler, src, len);
-    while (len > SIZE_MAX) {
-        memcpy(dst, src, SIZE_MAX);
-        dst += SIZE_MAX;
-        src += SIZE_MAX;
-        len -= SIZE_MAX;
+#if INTPTR_MAX > SSIZE_MAX
```
Dead2, line 15: I think this could benefit from a one-liner comment about what this is a workaround for.

crc32_fold.c
```diff
@@ -7,19 +7,23 @@
...
 }
 
 Z_INTERNAL void crc32_fold_copy_c(crc32_fold *crc, uint8_t *dst, const uint8_t *src, uint64_t len) {
     crc->value = functable.crc32(crc->value, src, len);
-    while (len > SIZE_MAX) {
-        memcpy(dst, src, SIZE_MAX);
-        dst += SIZE_MAX;
-        src += SIZE_MAX;
-        len -= SIZE_MAX;
+#if INTPTR_MAX > SSIZE_MAX
```
Dead2, line 20: And here too.

adler32_fold.c
mtl1979: Added explanation.

crc32_fold.c
mtl1979: Added explanation.

Dead2 (approved):
LGTM

nmoinvaz:
Do we really need this PR if we have the other PR that changes everything to size_t?

KungFuJesus:
> Do we really need this PR if we have the other PR that changes everything to size_t?

I'd probably argue not. Memcpy's len arg is also size_t. If it's going to be truncated it's going to be further up the chain.

mtl1979:
> Do we really need this PR if we have the other PR that changes everything to size_t?

Yes, because `ssize_t` can hold half of what `size_t` can.

nmoinvaz:
I see.

mtl1979:
> > Do we really need this PR if we have the other PR that changes everything to size_t?
> 
> I'd probably argue not. Memcpy's len arg is also size_t. If it's going to be truncated it's going to be further up the chain.

The argument is `size_t`, but it will throw compile-time error if you try to pass any constant that overflows `ssize_t` as that would mean copying destination over source, or wrapping around end of user address space, which would crash hard as it would enter kernel address space. At least GCC 9.x on FreeBSD refuses even to compile the code.

PR #1364, opened 2022-10-20
Fix deflate() with Z_BEST_COMPRESSION ignoring the dictionary

iii-i:
Hi!

While working on a DFLTCC improvement, I noticed that using software compression level 9 causes dictionary to be ignored.
I tracked this down to usage of different hashing algorithms.
This patch fixes the issue, but I'm not 100% sure it's correct, in particular:

* Is it safe to mix different hashing algorithms in the same hash table?
* Is it safe to count to `n` here, or do I need to subtract `WANT_MIN_MATCH` or something along these lines?

The problematic test is here: https://github.com/iii-i/zlib-ng/blob/dfltcc-inflate-small-window-20221020/test/test_small_window.cc

Best regards,
Ilya

---

deflate_slow() uses s->quick_insert_string(), while deflateSetDictionary() uses functable.insert_string(). These functions use different hashing algorithms, which leads to deflate_slow() ignoring the dictionary.

Fix by using both functions in deflateSetDictionary().

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -404,7 +404,7 @@ int32_t Z_EXPORT PREFIX(deflateSetDictionary)(PREFIX3(stream) *strm, const uint8
     while (s->lookahead >= STD_MIN_MATCH) {
         str = s->strstart;
         n = s->lookahead - (STD_MIN_MATCH - 1);
-        functable.insert_string(s, str, n);
+        s->insert_string(s, str, n);
         s->strstart = str + n;
         s->lookahead = STD_MIN_MATCH - 1;
         fill_window(s);


mtl1979:
It's definitely unsafe to mix hashing algorithms, or trying to use more than one hashing function at the same time for same stream.

deflate.c
```diff
@@ -404,6 +404,8 @@ int32_t Z_EXPORT PREFIX(deflateSetDictionary)(PREFIX3(stream) *strm, const uint8
     while (s->lookahead >= STD_MIN_MATCH) {
         str = s->strstart;
         n = s->lookahead - (STD_MIN_MATCH - 1);
+        for (i = 0; i < n; i++)
+            s->quick_insert_string(s, str + i);
         functable.insert_string(s, str, n);
```
mtl1979: Using both functions will overwrite the hashtable entries, effectively nullifying effect of the first two lines.

mtl1979: I would just use `s->insert_string(s, str, n);`.

iii-i: `s->insert_string(s, str, n)` worked, thanks!

nmoinvaz (approved):

deflate.c
mtl1979: Only differences between `quick_insert_string` and `insert_string` are that `quick_insert_string` returns the previous entry in the hash table -- so the current string can be compared with another string -- and `insert_string` hashes the remaining bytes of the string, so the latter avoids calling the function inside a loop using indirect jump, when you already know start of the string is a match.

mtl1979 (approved):
LGTM

PR #1365, opened 2022-10-27
Makefile.in: distclean should remove pkgconfig file(s) instead of clean

tachalorah:


mtl1979 (approved):
zlib also does remove *.pc files in distclean.

Issue #1370, opened 2022-11-15
Windows MSYS2 Clang64: build error crc32_braid.c


pashamcr:
https://github.com/zlib-ng/zlib-ng/blob/b3dcf11b4204a16fde71fa8224d7b7054e225b93/crc32_braid.c#L128
Clang 15.0.4
```
zlib/crc32_braid.c:128:24: error: cast to smaller integer type 'unsigned long' from 'const uint8_t *' (aka 'const unsigned char *') [-Werror, -Wpointer-to-int-cast]
        while (len && ((z_size_t)buf & (W - 1)) != 0) {
                       ^~~~~~~~~~~~~
1 error generated.

```

nmoinvaz:
It should probably be `uintptr_t` instead.

pashamcr:
@nmoinvaz, thanks build completed

nmoinvaz:
This is merged in `develop`.

PR #1371, opened 2022-11-19
Fixed cast to smaller integer type on MSYS2 Clang64.

nmoinvaz:
```c
zlib/crc32_braid.c:128:24: error: cast to smaller integer type 'unsigned long'
 from 'const uint8_t *' (aka 'const unsigned char *')
```
See #1370.

### DIFF
--- a/crc32_braid.c
+++ b/crc32_braid.c
@@ -125,7 +125,7 @@ Z_INTERNAL uint32_t crc32_braid(uint32_t crc, const uint8_t *buf, uint64_t len) 
         int k;
 
         /* Compute the CRC up to a z_word_t boundary. */
-        while (len && ((z_size_t)buf & (W - 1)) != 0) {
+        while (len && ((uintptr_t)buf & (W - 1)) != 0) {
             len--;
             DO1;
         }


mtl1979:
This might mask another issue with that compiler. `z_size_t` should not be `unsigned long` for that compiler target.

dzenanz (approved):

dzenanz:
This fixes the compiler warning for me, and all the tests pass. Can it be merged? CI failures seem unrelated.

@mtl1979 `zbuild.h` defines
```c++
#if defined(ZLIB_COMPAT)
...
#  define z_size_t unsigned long
#else
...
#  define z_size_t size_t
#endif
```
It makes me think this is needed for compatibility with zlib.

mtl1979:
@dzenanz zlib didn't even support MSYS2... It had only minimal support for MinGW. For most other operating systems, `unsigned long` works correctly.

dzenanz:
I have the same issue on Windows 11 with Visual Studio, see #1374.

mtl1979:
@Dead2 will merge when he has time. My opinion that we really should try to avoid using `z_size_t` in anything else than function parameter types in `*.c` files doesn't block merging this PR.

PR #1372, opened 2022-11-23
IBM zSystems DFLTCC: Support inflate with small window

iii-i:
There is no hardware control for DFLTCC window size, and because of that supporting small windows for deflate is not trivial: one has to make sure that DFLTCC does not emit large distances, which most likely entails somehow trimming the window and/or input in order to make sure that whave + avail_in <= wsize.

But inflate is much easier: one only has to allocate enough space. Do that in dfltcc_alloc_window(), and also introduce ZCOPY_WINDOW() in order to copy everything, not just what the software implementation cares about.

After this change, software and hardware window formats no longer match: the software will use wbits and wsize, and the hardware will use HB_BITS and HB_SIZE. Unlike deflate, inflate does not switch between software and hardware implementations mid-stream, which leaves only inflateSetDictionary() and inflateGetDictionary() interesting.

### DIFF
--- a/arch/s390/dfltcc_common.c
+++ b/arch/s390/dfltcc_common.c
@@ -20,14 +20,18 @@ void Z_INTERNAL *PREFIX(dfltcc_alloc_window)(PREFIX3(streamp) strm, uInt items, 
     void *w;
 
     /* To simplify freeing, we store the pointer to the allocated buffer right
-     * before the window.
+     * before the window. Note that DFLTCC always uses HB_SIZE bytes.
      */
-    p = ZALLOC(strm, sizeof(void *) + items * size + PAGE_ALIGN, sizeof(unsigned char));
+    p = ZALLOC(strm, sizeof(void *) + MAX(items * size, HB_SIZE) + PAGE_ALIGN, sizeof(unsigned char));
     if (p == NULL)
         return NULL;
     w = ALIGN_UP((char *)p + sizeof(void *), PAGE_ALIGN);
     *(void **)((char *)w - sizeof(void *)) = p;
     return w;
+}
+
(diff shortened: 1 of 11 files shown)


arch/s390/dfltcc_common.c
```diff
@@ -16,20 +16,31 @@
...
     void *w;
 
     /* To simplify freeing, we store the pointer to the allocated buffer right
      * before the window.
      */
-    p = ZALLOC(strm, sizeof(void *) + items * size + PAGE_ALIGN, sizeof(unsigned char));
+    n = items * size;
+    if (n < HB_SIZE)
+      /* DFLTCC always uses HB_SIZE bytes. */
+      n = HB_SIZE;
```
mtl1979: Should use `MAX()` macro, it saves 2 lines without sacrificing readability.

arch/s390/dfltcc_common.c
```diff
@@ -16,20 +16,31 @@
...
     if (p == NULL)
         return NULL;
     w = ALIGN_UP((char *)p + sizeof(void *), PAGE_ALIGN);
     *(void **)((char *)w - sizeof(void *)) = p;
     return w;
 }
 
+void Z_INTERNAL PREFIX(dfltcc_copy_window)(void *dest, const void *src, size_t n) {
+    if (n < HB_SIZE)
+      n = HB_SIZE;
```
mtl1979: Should use `MAX()` macro.

arch/s390/dfltcc_common.c
iii-i: Done.

arch/s390/dfltcc_common.c
iii-i: Done.

PR #1377, opened 2022-12-07
Fix definition of z_size_t to match documentation of legacy zlib API.

mtl1979:
Legacy zlib changed definition of `z_size_t` to smallest unsigned type that can hold a pointer value after we forked, but this was never merged in.

We currently avoid including standard headers in our own headers as some standard headers must be included in specific order, otherwise we could just include `stddef.h` in `zbuild.h`.

Some of the function parameters and return types, and struct members are still limited to `unsigned long` when `ZLIB_COMPAT` is defined.

Blocks: #1356

### DIFF
--- a/inflate.c
+++ b/inflate.c
@@ -1292,8 +1292,8 @@ int32_t Z_EXPORT PREFIX(inflateSync)(PREFIX3(stream) *strm) {
     in = strm->total_in;
     out = strm->total_out;
     PREFIX(inflateReset)(strm);
-    strm->total_in = (z_size_t)in;
-    strm->total_out = (z_size_t)out;
+    strm->total_in = (z_uintmax_t)in; /* Can't use z_size_t here as it will overflow on 64-bit Windows */
+    strm->total_out = (z_uintmax_t)out;
     state->flags = flags;
     state->mode = TYPE;
     return Z_OK;
--- a/zbuild.h
+++ b/zbuild.h
@@ -85,7 +85,11 @@
 #  define PREFIX3(x) z_ ## x
 #  define PREFIX4(x) x ## 64
 #  define zVersion zlibVersion
-#  define z_size_t unsigned long
+#  if defined(_WIN64)
+#    define z_size_t unsigned __int64
+#  else
+#    define z_size_t unsigned long
+#  endif
 #else
 #  define PREFIX(x) zng_ ## x
 #  define PREFIX2(x) ZLIBNG_ ## x
@@ -93,6 +97,13 @@
 #  define PREFIX4(x) zng_ ## x
 #  define zVersion zlibng_version
 #  define z_size_t size_t
+#endif
+
+/* In zlib-compat some functions and types use unsigned long, but zlib-ng use size_t */
+#if defined(ZLIB_COMPAT)
+#  define z_uintmax_t unsigned long
+#else
+#  define z_uintmax_t size_t
 #endif
 
 /* Minimum of a and b. */
--- a/PORTING.md
+++ b/PORTING.md
@@ -43,7 +43,8 @@ certain value will need to be updated.
 - Static library is *libz.a* on Unix and macOS, or *zlib.lib* on Windows
 - Shared library is *libz.so* on Unix, *libz.dylib* on macOS, or *zlib1.dll*
   on Windows
-- Type `z_size_t` is *unsigned long*
+- Type `z_size_t` is *unsigned __int64* on 64-bit Windows, and *unsigned long* on 32-bit Windows, Unix and macOS
+- Type `z_uintmax_t` is *unsigned long* in zlib-compat mode, and *size_t* with zlib-ng API
 
 zlib-ng native mode
 -------------------
--- a/compress.c
+++ b/compress.c
@@ -28,8 +28,8 @@
    memory, Z_BUF_ERROR if there was not enough room in the output buffer,
    Z_STREAM_ERROR if the level parameter is invalid.
 */
-int Z_EXPORT PREFIX(compress2)(unsigned char *dest, z_size_t *destLen, const unsigned char *source,
-                        z_size_t sourceLen, int level) {
+int Z_EXPORT PREFIX(compress2)(unsigned char *dest, z_uintmax_t *destLen, const unsigned char *source,
+                        z_uintmax_t sourceLen, int level) {
     PREFIX3(stream) stream;
     int err;
     const unsigned int max = (unsigned int)-1;
@@ -63,14 +63,14 @@ int Z_EXPORT PREFIX(compress2)(unsigned char *dest, z_size_t *destLen, const uns
         err = PREFIX(deflate)(&stream, sourceLen ? Z_NO_FLUSH : Z_FINISH);
     } while (err == Z_OK);
 
-    *destLen = (z_size_t)stream.total_out;
+    *destLen = stream.total_out;
     PREFIX(deflateEnd)(&stream);
     return err == Z_STREAM_END ? Z_OK : err;
 }
 
 /* ===========================================================================
  */
-int Z_EXPORT PREFIX(compress)(unsigned char *dest, z_size_t *destLen, const unsigned char *source, z_size_t sourceLen) {
+int Z_EXPORT PREFIX(compress)(unsigned char *dest, z_uintmax_t *destLen, const unsigned char *source, z_uintmax_t sourceLen) {
     return PREFIX(compress2)(dest, destLen, source, sourceLen, Z_DEFAULT_COMPRESSION);
 }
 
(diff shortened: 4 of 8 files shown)


Dead2:
This strikes me as complicating the code a bit.
Would it be worth it to instead introduce a new type with the required behavior?
We could call it zng_size_t perhaps, if that is a close enough description to its real sizes?

mtl1979:
> This strikes me as complicating the code a bit. Would it be worth it to instead introduce a new type with the required behavior? We could call it zng_size_t perhaps, if that is a close enough description to its real sizes?

I'm trying to get Mark Adler to fix this in legacy zlib... The best would be to add `_z` version in legacy zlib (and when using `ZLIB_COMPAT` for zlib-ng) that use `size_t` instead of `unsigned long` like other similar functions already have. When not using `ZLIB_COMPAT` we (in zlib-ng) would eventually have only one version, the one that use `size_t` "directly".

Introducing a new type would be quite short-living temporary solution as it would become non-useful after Mark Adler fixes legacy zlib.

mtl1979:
@nmoinvaz Any comments? Still can't ask reviews on my own pull requests, eh...

nmoinvaz:
I think we should see what Mark Adler says about the issue on zlib side. I do like the name `zng_size_t` rather than `z_ulong_t`.

zbuild.h
```diff
@@ -85,7 +85,11 @@
 #  define PREFIX3(x) z_ ## x
 #  define PREFIX4(x) x ## 64
 #  define zVersion zlibVersion
-#  define z_size_t unsigned long
+#  if defined(_WIN64)
```
nmoinvaz, line 88: Why wouldn't you just define `z_size_t` as `size_t`? Since `z_size_t` is just size of pointer. Or even `uintptr_t`?

zbuild.h
```diff
@@ -95,6 +99,13 @@
 #  define z_size_t size_t
 #endif
 
+/* In zlib-compat some functions and types use unsigned long, but zlib-ng use size_t */
+#if defined(ZLIB_COMPAT)
+#  define z_ulong_t unsigned long
```
nmoinvaz: I like name `zng_size_t` like what @Dead2 suggested.

mtl1979: I thought `zng_size_t` would be misleading as it's not supposed to have anything to do with `size_t` when using `ZLIB_COMPAT` and it could be confusing to contributors which one to use if the names are too similar.

zbuild.h
mtl1979: Because using `size_t` or `uintptr_t` would require including a header in `zbuild.h` and that would introduce problems with header ordering in any file that includes `zbuild.h`. I really don't want to include all standard headers in `zbuild.h` just to avoid including them in wrong order in source files. `zbuild.h` is supposed to always be the first header included, so it can't include any headers that might have circular dependency, like `stdint.h` or `stdarg.h`.

mtl1979:
> I think we should see what Mark Adler says about the issue on zlib side. I do like the name `zng_size_t` rather than `z_ulong_t`.

I agree... I just made this PR so I have something to show about how I would fix this myself... With previous PRs I created for legacy zlib, Mark Adler did use parts of the PRs but rewrote it to be more compatible with older compilers. Of course, zlib-ng doesn't need to support 16-bit platforms or compilers that support them.

zbuild.h
Dead2: `zng_size_t` name would be explained by it being `size_t` when building zng, but something else whenever when building zlib-compatible.
`z_ulong_t` has the same kind of confusion but in the opposite direction.

Alternatively we could do `zng_huge_t` or `zng_biggest_t`, that way the name does not suggest any specific type except that it is something big.

mtl1979: The type in legacy zlib was `uLong`, so I chose the closest that still conforms to our own naming style... I'm still waiting for better suggestions... Changing the name to final choice is pretty easy task using `git diff` as a way to ensure all instances are changed correctly.

One possible choice is also `zng_max_t`,  which would follow the style used in `stdint.h`.

mtl1979:
https://github.com/madler/zlib/commit/fa8cd50ada68b873c754766bc73b12080b7b309e

zbuild.h
Dead2: I like `zng_max_t`, that would probably be the least confusing naming, especially since it also follows stdint naming.

Dead2: Or perhaps `zng_uintmax_t` to include the type as well, also closely matching stdint type `uintmax_t`

mtl1979: > Or perhaps `zng_uintmax_t` to include the type as well, also closely matching stdint type `uintmax_t`

If we can find a compromise that you, @nmoinvaz and I can agree is the best, I will rename it and we can merge this and move on.

zbuild.h
nmoinvaz: `zbuild.h` already includes `stdint.h`..

inflate.c
```diff
@@ -1353,7 +1353,7 @@ int32_t Z_EXPORT PREFIX(inflateCopy)(PREFIX3(stream) *dest, PREFIX3(stream) *sou
     }
     copy->next = copy->codes + (state->next - state->codes);
     if (window != NULL) {
-        ZCOPY_WINDOW(window, state->window, 1U << state->wbits);
+        ZCOPY_WINDOW(window, state->window, (size_t)1U << state->wbits);
```
nmoinvaz: @mtl1979  can you break out this small fix into its own PR? If not I can do it for you. It isn't necessarily related to the rest of the changes.

mtl1979: Can do...

mtl1979: Separated as #1429.

zbuild.h
mtl1979: Including `stdint.h` itself is not problem, the problem comes if also `stdarg.h` is included. Also, any header whose contents are not used will be filtered out by preprocessor. We use compiler's internal types intentionally to catch any "bugs" in compiler configuration. In the past it was common for files to be compiled differently inside Visual Studio IDE and command-line using nmake Makefiles. If the types don't match with the ones used in `zlib.h` or `zllb-ng.h`, the error is caught instead of getting masked out.

When `ZLIB_COMPAT` is defined, we're not supposed to include any of the standard headers as if `Z_SOLO` is always defined. I don't think Mark Adler even thought there would some day be platform where `unsigned long` isn't same size as a pointer.

Dead2 (approved):
LGTM

nmoinvaz:
I'm not sure of the need since the description says we cannot add stddef.h to zbuild.h but we can..

mtl1979:
> I'm not sure of the need since the description says we cannot add stddef.h to zbuild.h but we can..

We can do whatever we want when `ZLIB_COMPAT` isn't defined, but this whole PR is about making sure everything is correct when it is defined. Each new clang or gcc version tries to break more circular dependencies, but we still need to make sure our code compiles correctly on older compiler versions. Like I have implied earlier, specific header versions are not tied to specific Visual C++ version as it can target individual Windows builds using different Windows SDKs.

Legacy zlib predates standard headers and standard types, so some of the types it use conflict with types from `stddef.h` and `stdint.h`. This is why we still have and need types with prefix `z_`.

nmoinvaz:
> Each new clang or gcc version tries to break more circular dependencies

What are you referring to here?

> specific header versions are not tied to specific Visual C++ version as it can target individual Windows builds using different Windows SDKs.

I wouldn't expect that Visual C++ would ship new headers for each version of Visual C++.

> Legacy zlib predates standard headers and standard types, so some of the types it use conflict with types from `stddef.h` and `stdint.h`. This is why we still have and need types with prefix z_.

That sounds like a compiler toolchain bug. Why are we trying to fix that in zlib-ng? I thought we were supporting/recommending only Visual C++ 2013 or higher.

Which compiler version and Windows SDK exhibits the problem? Let me know so I can download that specific version and try to reproduce the error. I would rather us not trying to fix something that only "in theory" is a problem without first reproducing it.

mtl1979:
> That sounds like a compiler toolchain bug. Why are we trying to fix that in zlib-ng? I thought we were supporting/recommending only Visual C++ 2013 or higher.
> 
> Which compiler version and Windows SDK exhibits the problem? Let me know so I can download that specific version and try to reproduce the error. I would rather us not trying to fix something that only "in theory" is a problem without first reproducing it.

`unsigned long` has been 32 bits since very first 32-bit Windows version. I asked few days ago how long we want to support legacy zlib API and hell froze solid... Apparently people still think we should support buggy code that was created when 16-bit processors were popular. 

There is a few closed issues in this repository that deal with not using standard integer types for `ZLIB_COMPAT`. Even if compiler can generate code using `size_t`, C++ runtime library of Windows doesn't understand it, so passing the type to functions like `printf` can result in garbage output. 

Visual C++ 2015 was first version that had more comprehensive support for C99 standard. Visual C++ 2017.9 is first usable version that supports the new C++ runtime library (UCRT) that Windows 10 ships with. Visual C++ 2017 broke backwards compatibility with older versions and caused a lot of programs to stop working due to using same filename as Visual C++ 2015, so basically was a dud release.

PR #1384, opened 2022-12-13
Fix ABI checking...

mtl1979:
* Ubuntu 22.04 use different format for ABI files so old ones need to be removed
* Use more recent zlib-ng commit to avoid issues with internal adler32 and crc32 functions


nmoinvaz:
> When new ABI files are uploaded, Ubuntu 22.04 must be used to create them, and CI runs needs to be locked to Ubuntu 22.04.

ABI files can be found in any Package Check workflow that fails; they are uploaded as artifacts. For example, here is from a recent run (older runs may not have retained the artifacts):
https://github.com/zlib-ng/zlib-ng/actions/runs/3790538662

mtl1979:
@nmoinvaz I was waiting for GitHub to upgrade all "latest" workflows to 22.04 first... Otherwise the non-upgraded ones will fail instead due to file format inconsistency.

mtl1979:
Rebased only to include recent fixes to CI runs...

nmoinvaz:
@mtl1979 perhaps all the machines have been updated now?

mtl1979:
> @mtl1979 perhaps all the machines have been updated now?

There is still pull requests that can change our API. Kind of chicken and egg problem.

mtl1979:
Rebased again and fixed issue with visibility of `crc32_pclmulqdq`.

mtl1979:
@nmoinvaz Now there is should be quite nice amount of ABI files...

nmoinvaz:
Wow so much green checkmarks! Only GCC is broken but that is a separate issue where perhaps `<xsaveintrin.h>` needs to be included.

mtl1979:
@nmoinvaz I made a script that tries to build zlib and zlib-ng for all possible targets... Running the script takes ages, but it saves all the ABI files automatically ;)

mtl1979:
@nmoinvaz MinGW build was completely broken... I added separate commit so I can test if we get all greens...

nmoinvaz:
See my comment here: https://github.com/zlib-ng/zlib-ng/pull/1407#discussion_r1082970732. Seems it was recently introduced.

cmake/detect-intrinsics.cmake
```diff
@@ -536,3 +536,21 @@ macro(check_vgfma_intrinsics)
         HAVE_VGFMA_INTRIN FAIL_REGEX "not supported")
     set(CMAKE_REQUIRED_FLAGS)
 endmacro()
+
+macro(check_xsave_intrinsics)
+    if(NOT NATIVE_FLAG)
```
nmoinvaz: Should be `if (NOT NATIVEFLAG AND NOT MSVC)`? Notice extra underscore you added in `NATIVEFLAG`.

configure
```diff
@@ -1224,6 +1241,27 @@ EOF
...
+#else
+#  include <x86gprintrin.h>
+#endif
+int main(void) {
+    return _xgetbv(0);
+}
+EOF
+    if try ${CC} ${CFLAGS} ${xsaveflag} $test.c; then
+        echo "Checking for XSAVE intrinsics ... Yes." | tee -a configure.log
+        HAVE_XSAVE_INTRIN=1
```
nmoinvaz: This `HAVE_XSAVE_INTRIN` does not appear to be used anywhere.

cmake/detect-intrinsics.cmake
mtl1979: Possibly... On MSVC it would just fail and not set the flag...

configure
mtl1979: Fixed.

cmake/detect-intrinsics.cmake
nmoinvaz: We can just return() on MSVC then.

mtl1979:
> See my comment here: [#1407 (comment)](https://github.com/zlib-ng/zlib-ng/pull/1407#discussion_r1082970732). Seems it was recently introduced.

There was other errors with MinGW build too... Basically related to symbol mangling, gzip functions and out-of-tree builds. I got a lot of linker errors while trying to get XSAVE test working with configure.

cmake/detect-intrinsics.cmake
mtl1979: It doesn't hurt to let MSVC test.... Would not be the first time Microsoft messes up with the intrinsics headers and breaks things... One improvement would be to add preprocessor flag so that on non-Windows platforms, it would also use the intrinsic function and not inline assembly... But some gcc versions are known to have buggy intrinsic function...

nmoinvaz:
I think perhaps the xsave stuff should be its own PR.

CMakeLists.txt
```diff
@@ -740,9 +740,14 @@ if(WITH_OPTIM)
             endif()
         endif()
     elseif(BASEARCH_X86_FOUND)
+        check_xsave_intrinsics()
```
nmoinvaz: I think this entire XSave block should go at the end of the if statement. I've tried to keep all the checks alphabetical.

nmoinvaz: But idk, it might be ok as is. Maybe put fallback_bultin's block above this one?

configure
```diff
@@ -452,12 +460,16 @@ if test "$gcc" -eq 1 && ($cc $CFLAGS -c $test.c) >> configure.log 2>&1; then
         LDSHARED=${LDSHARED-"$cc"}
         LDSHAREDFLAGS="-shared -Wl,--out-implib,${IMPORTLIB}"
         LDSHAREDLIBC=""
-        DEFFILE='win32/${LIBNAME2}.def'
+        if test $gzfileops -eq 0; then
```
nmoinvaz: I'm not sure what some of these visibility, RCFLAGS, gzfileops changes are. Perhaps they should be separate commits.

configure
```diff
@@ -1500,6 +1542,12 @@ case "${ARCH}" in
             CFLAGS="${CFLAGS} -DX86_FEATURES"
             SFLAGS="${SFLAGS} -DX86_FEATURES"
 
+            check_xsave_intrinsics
```
nmoinvaz: This should definitely go in alphabetical order, at the bottom of this if statement.

CMakeLists.txt
nmoinvaz: Yeah I think it should be alphabetical, in the same order as configure script.

mtl1979: This has to be before line 749, but otherwise the placement isn't critical...

configure
mtl1979: MinGW doesn't support visibility attribute, but gives only warning, not error, so it makes sense to not check it for configure as we don't parse the compiler output like cmake does.

`RCFLAGS` adds binary/output directory to the include path for `windres` as it doesn't do it automatically. Otherwise it won't find `zlib.h` or `zlib-ng.h`.

configure
mtl1979: I added it here, because it's only relevant to `x86_features.h`. But I can move it...

mtl1979:
> I think perhaps the xsave stuff should be its own PR.

Before, I made separate pull requests for related patches, but this whole PR grew up bigger than I planned due to adding the ABI files and revealing all the build failures behind the noise.

`xsave` tests and all the MinGW build fixes have to be on same PR, otherwise it will be hard to verify that there is no more errors as the build will not finish successfully without including all the changes.

All the MinGW fixes are now separated as #1412. It can only be merged after this PR, so I marked it as a draft.

mtl1979:
@Dead2 Last chance to comment... I want this merged as fast as possible so we can merge all the PRs that depend on ABI checks working correctly.

nmoinvaz (approved):
LGTM.

Dead2 (approved):
LGTM

mtl1979:
Rebasing due to 1ab443812aceab6c80ccd573ecdd06f73a84eef4 and generating new ABI files.

PR #1385, opened 2022-12-14
Update our compatibility to zlib 1.2.13 and merge in remaining bug fixes.

mtl1979:
Requires: #1384

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -59,7 +59,7 @@
 # undef deflateInit2
 #endif
 
-const char PREFIX(deflate_copyright)[] = " deflate 1.2.12 Copyright 1995-2022 Jean-loup Gailly and Mark Adler ";
+const char PREFIX(deflate_copyright)[] = " deflate 1.2.13 Copyright 1995-2022 Jean-loup Gailly and Mark Adler ";
 /*
   If you use the zlib library in a product, an acknowledgment is welcome
   in the documentation of your product. If for some reason you cannot
--- a/deflate.c
+++ b/deflate.c
@@ -662,8 +662,15 @@ unsigned long Z_EXPORT PREFIX(deflateBound)(PREFIX3(stream) *strm, unsigned long
 
     /* if not default parameters, return conservative bound */
     if (DEFLATE_NEED_CONSERVATIVE_BOUND(strm) ||  /* hook for IBM Z DFLTCC */
-            s->w_bits != 15 || HASH_BITS < 15)
+            s->w_bits != 15 || HASH_BITS < 15) {
+        if (s->level == 0) {
+            /* upper bound for stored blocks with length 127 (memLevel == 1) --
+               ~4% overhead plus a small constant */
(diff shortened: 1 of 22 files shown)


trees.c
```diff
@@ -670,7 +670,7 @@ void Z_INTERNAL zng_tr_flush_block(deflate_state *s, char *buf, uint32_t stored_
                 opt_lenb, s->opt_len, static_lenb, s->static_len, stored_len,
                 s->sym_next / 3));
 
-        if (static_lenb <= opt_lenb)
+        if (static_lenb <= opt_lenb || s->strategy == Z_FIXED)
```
nmoinvaz, line 673: The commits from madler/zlib where these changes are made should be copied over, instead of a single lump-sum commit.

mtl1979: I was going to do that after GitHub CI starts working again... It had quite long backlog... I just wanted to do one CI run that actually verifies that there was no ABI breakage between 1.2.12 and 1.2.13...

mtl1979: Now there is still one commit that glues things together, but I cherry-picked and cleaned the original commits for bug fixes from madler/zlib...

nmoinvaz:
It looks like we may be missing 2 commits from Dec from here: https://github.com/madler/zlib/compare/master...develop

mtl1979:
@nmoinvaz Are you saying we should update our compatibility to 1.2.13.1 or were those part of 1.2.13 already?

nmoinvaz:
Actually you were right. I wasn't aware that there was an official 1.2.13 release yet. So those changes from Dec (after 1.2.13) should probably not go in until later.

mtl1979:
> Actually you were right. I wasn't aware that there was an official 1.2.13 release yet. So those changes from Dec (after 1.2.13) should probably not go in until later.

For bug fixes, it really doesn't make a difference, but for something that breaks the ABI or API, it's completely different story... Some of our functions, like deflateBound(), have diverged so much that cherry-picking is quite pointless as whole commit needs to be rewritten.

deflate.c
```diff
@@ -660,6 +660,12 @@ unsigned long Z_EXPORT PREFIX(deflateBound)(PREFIX3(stream) *strm, unsigned long
         wraplen = ZLIB_WRAPLEN;
     }
 
+    if (s->level == 0) {
```
nmoinvaz: This if statement could go inside the if statement on line 670 if we wanted.

mtl1979:
Rebased.

nmoinvaz (approved):
LGTM. Probably want to rebase for abi package changes.

mtl1979:
> LGTM. Probably want to rebase for abi package changes.

More likely if #1412 is merged before this as this was already rebased after ABI fixes, even though this didn't need recreating ABI files as different base commit is used.

Dead2:
Rebase?

mtl1979:
> Rebase?

Done... Just to see everything still passes with flying colours...

Dead2 (approved):

PR #1387, opened 2022-12-15
Tell scan-build ASSERT_TRUE() behaves like assert().

mtl1979:
Fixes false positives with unix.Malloc check.


nmoinvaz:
Perhaps this is better to go in test_shared file.

mtl1979:
> Perhaps this is better to go in test_shared file.

Then I need to move including it below including googletest's headers.

nmoinvaz (approved):

PR #1390, opened 2022-12-22
IBM zSystems DFLTCC: Do not update strm.adler for raw streams

iii-i:
Commit d38dd9240f2d ("IBM Z DFLTCC: Fix updating strm.adler with inflate()") broke libxml2, as can be seen with the repro from [1]:

    $ echo "<a></a>" | gzip >file.xml.gz
    $ python3 -c 'import libxml2; libxml2.parseFile("file.xml.gz")'
    file.xml.gz:1: parser error : Document is empty

This is because libxml2 expects strm.adler to be untouched for raw streams.

Fix this and a similar issue in deflate by adding state->wrap checks. Add tests.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2155328
[2] https://gitlab.gnome.org/GNOME/libxml2/-/blob/v2.10.3/xzlib.c#L607

### DIFF
--- a/arch/s390/dfltcc_deflate.c
+++ b/arch/s390/dfltcc_deflate.c
@@ -240,7 +240,10 @@ again:
         *strm->next_out = (unsigned char)state->bi_buf;
     /* Honor history and check value */
     param->nt = 0;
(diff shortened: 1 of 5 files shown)


Dead2 (approved):
LGTM

Issue #1379, opened 2022-12-08
SPARC64 crash with Ubuntu 22.04 and gcc 11.3.0 


mtl1979:
All tests fail or crash in Ubuntu 22.04 when using gcc 11.3.0 and qemu.
When building with gcc 9.4.0, all tests pass.

nmoinvaz:
I confirmed the issue. It appears to crash even with a simple `main()` that `return 0;`. I upgraded my WSL to Ubuntu 22 and tried GCC 11.3.0 & 10.4.0 & 9.5.0 and they all seg fault.

nmoinvaz:
@mtl1979 do you want to close this issue?

mtl1979:
@nmoinvaz Like with the little-endian 64-bit PowerPC, this might be useful as converted to a discussion, so we don't need to create another one when eventually qemu is fixed... I'm not sure yet if the underlying issue is exactly same block of code for both architectures or is there possibly just some overlap.

Like I said to @Dead2 elsewhere, downgrading to "ubuntu-20.04" runner should be a temporary solution -- to avoid delaying next stable version for too long -- until the real issue is fixed and new packages have been uploaded.

KungFuJesus:
If there is a particular issue you maybe want me to test or explore, I do have an UltraSPARC T4 with Solaris 11.2 (or maybe it was 3) that I can test this on.  I'm fairly certain the T4 is EOL from Oracle at the moment but it's probably a close enough approximation on newer variants.

mtl1979:
@KungFuJesus If it's qemu bug or regression, testing on real hardware doesn't make sense. If it's gcc issue, we need to know what flag is missing or incorrect.

KungFuJesus:
FWIW gtests pass with flying colors on OpenIndiana on an ancient Sun Fire V240.  Though, the symbol versioning doesn't quite seem to be supported with those arguments to the linker, even with GNU's ld.  We should probably look into that.

mtl1979:
> FWIW gtests pass with flying colors on OpenIndiana on an ancient Sun Fire V240. Though, the symbol versioning doesn't quite seem to be supported with those arguments to the linker, even with GNU's ld. We should probably look into that.

You might want to create another issue about the symbol versioning issue with all the relevant logs.

KungFuJesus:
[...]

PR #1394, opened 2022-12-27
Use Ubuntu 20.04 images for SPARC64 and PPC64LE tests

nmoinvaz:
I suspect that qemu or gcc or whatever must be buggy. See #1378 and #1379.


Dead2 (approved):
LGTM

PR #1396, opened 2022-12-29
Symbol versioning support for multiple versions of functions.

Dead2:
This allows us to make changes to functions without losing backwards compatibility.

With this we can:
 - Change types
 - Add/remove parameters
 - Change function or parameter meaning

This means we can have one version of a function bound to 2.0.0 backwards compatibility
while also having a function of the same name bound to 2.1.0 that differs from the old one.

This would then make it easier for distros to upgrade since it would not require a full recompile of all the dependent applications at the same time.

Requires #1421


Dead2:
The `zlibng_version` export was a little more complex than would normally be the case due to it previously having been unversioned. The linker refuses to have the same function name and alias when creating an unversioned alias, so for this to work I had to rename the function to avoid the linker conflict.

mtl1979:
I thought about this, but implementing it on Windows is a little trickier... It does support versioned symbols, but requires re-exporting symbols using thunk dll. When dumping imports of Windows executables, the thunk dll shows up as separate file with version number embedded in filename.

Dead2:
I really don't see any big need to support this on Windows.
The biggest benefit of this is when zlib-ng is included as a system library, where it might need to get updated or the user might attempt to run applications compiled against an older distro.
For windows it is more likely that the application will bundle the zlib-ng dll it is compiled against.

mtl1979:
@Dead2 The issue is mostly Qt applications under Windows as Qt itself links against zlib, but it might not be the same version the application or one of its other dependencies use. 

In the past I have compiled zlib first and then forced Qt to link dynamically against it using non-default build options as the default ones don't include all the necessary functions. Then the application itself will use Qt library as a thunk library to get the zlib symbols.

Dead2:
@mtl1979 I am not saying I reject the idea, just that I don't think it is a requirement for our next version.
I will certainly review a PR for the feature.
This PR ended up a lot smaller and less complex than what I had expected/dreaded when I started down this path, the good thing is we already had versioning (mostly) working with the versioning script, and that can be used as the fallback, so there was much less hackery needed than I had thought.

mtl1979:
@Dead2 I don't see it getting enabled when using `configure`.

nmoinvaz:
The visibility change on some of the functions can easily go into a different PR and merged quicker.

Dead2:
I am not quite certain how to decide whether this should be enabled or not.

The current CMake code requires UNIX and not APPLE in order to enable it, and that probably works in most cases, but there is probably corner cases where it does not. What about HP-UX, Solaris, etc? I have no way of knowing.

But there might be a better way to decide this based on linker or fileformat detection, does anyone have ideas about what would be the Right way to do this?
It has to be implemented in both CMake and Configure in a way that ensures that it is always enabled on the same platforms to avoid having different resulting ABIs.

mtl1979:
@Dead2 We should only enable symbol versioning for operating system and compiler combination(s) that we know support it... If someone wants to enable it for new combination, then the CI should test that it actually results in expected symbol names. If there is no CI runner for that operating system, we shouldn't try to support it until compatible CI runner is added.

Dead2:
I intend this to be an ABI change for 2.1, so we should get this in all the relevant platforms before that.
That means it will be better to have a hard fail on platforms that don't want it, and get bug-reports back that then get fixed as minor versions.
Doing it the other way of gradually enabling them as we confirm platforms that want it will cause ABI change when the minor release enables it, and that is not something we want to do.

I am assuming this will only work on ELF.

The right ™️ way would probably be to do a compile+link during cmake/configure and then verify the output symbol names are correct. Not sure what the most portable way of doing that is though.
Possibly we could go for default enabled on ELF, then the test turns it back off again if it fails. If the test succeeds or is unable to run (missing objdump or other supported tool), it stays enabled and we rely on bug reports if that ever turns out to cause failures somewhere.

mtl1979:
@Dead2 cmake's configurations step and `configure` should try to compile small shared library with versioned symbols and then compare the cleaned result with expected one. If that passes, it should enable it... Obviously it doesn't make sense to test support when building only static library.

Dead2:
> @Dead2 cmake's configurations step and `configure` should try to compile small shared library with versioned symbols and then compare the cleaned result with expected one. If that passes, it should enable it... Obviously it doesn't make sense to test support when building only static library.

Exactly. If anyone wants to contribute a commit adding the proper Configure and/or CMake, I'd appreciate that. I am living in a never-ending brain fog lately (did that last round of covid make it worse? It sure feels like it..), so I haven't been able to put the pieces for this together yet. Otherwise I'll get around to it when I can.
(If anyone does, it can be posted as a PR with ´symver´ branch as target, that way I can accept it directly into this PR and it can be merged as a complete unit)

Dead2:
Rebased and removed commits that have already been merged.

mtl1979:
Needs rebase again ;)

Dead2:
Yeah, but not sure whether to merge this now or wait until we actually need to use it.
I ended up going another route for the problematic version string, but we are likely to want this sooner or later.

mtl1979:
> Yeah, but not sure whether to merge this now or wait until we actually need to use it.
> I ended up going another route for the problematic version string, but we are likely to want this sooner or later.

I'm not sure either... I'm inclined to say we should merge before next stable release, so we can test if it works correctly on all platforms where it gets enabled. That requires keeping the patch up-to-date.

PR #1401, opened 2023-01-13
Fix visibility of a few internal functions.

Dead2:
`crc32_pclmulqdq`, `fold_16_vpclmulqdq_copy` and `fold_16_vpclmulqdq` should be tagged as `Z_INTERNAL`.


nmoinvaz (approved):
LGTM

PR #1405, opened 2023-01-13
Move initialization of functable to init_functable function

pps83:
This PR also fixes an issue where feature detection wasn't done for each of the stub functions (fixes #1172).

### DIFF
--- a/functable.c
+++ b/functable.c
@@ -12,380 +12,378 @@
 #include "cpu_features.h"
 
 static void init_functable(void) {
-    struct functable_s* functable_ptr = &functable;
-    struct functable_s functable;
-    Assert(functable_ptr != &functable,
-        "functable_ptr has to point to global functable");
+    struct functable_s ft;
 
     cpu_check_features();
 
     // update_hash_stub:
-    functable.update_hash = &update_hash_c;
+    ft.update_hash = &update_hash_c;
 #ifdef X86_SSE42_CRC_HASH
     if (x86_cpu_has_sse42)
-        functable.update_hash = &update_hash_sse4;
+        ft.update_hash = &update_hash_sse4;
 #elif defined(ARM_ACLE_CRC_HASH)
     if (arm_cpu_has_crc32)
-        functable.update_hash = &update_hash_acle;
+        ft.update_hash = &update_hash_acle;
 #endif
 
     // insert_string_stub:
-    functable.insert_string = &insert_string_c;
+    ft.insert_string = &insert_string_c;
 #ifdef X86_SSE42_CRC_HASH
     if (x86_cpu_has_sse42)
-        functable.insert_string = &insert_string_sse4;
+        ft.insert_string = &insert_string_sse4;
 #elif defined(ARM_ACLE_CRC_HASH)
     if (arm_cpu_has_crc32)
-        functable.insert_string = &insert_string_acle;
+        ft.insert_string = &insert_string_acle;
 #endif
 
     // quick_insert_string_stub:
-    functable.quick_insert_string = &quick_insert_string_c;
+    ft.quick_insert_string = &quick_insert_string_c;
 #ifdef X86_SSE42_CRC_HASH
     if (x86_cpu_has_sse42)
-        functable.quick_insert_string = &quick_insert_string_sse4;
+        ft.quick_insert_string = &quick_insert_string_sse4;
 #elif defined(ARM_ACLE_CRC_HASH)
     if (arm_cpu_has_crc32)
-        functable.quick_insert_string = &quick_insert_string_acle;
+        ft.quick_insert_string = &quick_insert_string_acle;
 #endif
 
     // slide_hash_stub:
-    functable.slide_hash = &slide_hash_c;
+    ft.slide_hash = &slide_hash_c;
 #ifdef X86_SSE2
 #  if !defined(__x86_64__) && !defined(_M_X64) && !defined(X86_NOCHECK_SSE2)
     if (x86_cpu_has_sse2)
 #  endif
-        functable.slide_hash = &slide_hash_sse2;
+        ft.slide_hash = &slide_hash_sse2;
 #elif defined(ARM_NEON_SLIDEHASH)
 #  ifndef ARM_NOCHECK_NEON
     if (arm_cpu_has_neon)
 #  endif
-        functable.slide_hash = &slide_hash_neon;
+        ft.slide_hash = &slide_hash_neon;
 #endif
 #ifdef X86_AVX2
     if (x86_cpu_has_avx2)
-        functable.slide_hash = &slide_hash_avx2;
+        ft.slide_hash = &slide_hash_avx2;
 #endif
 #ifdef PPC_VMX_SLIDEHASH
     if (power_cpu_has_altivec)
-        functable.slide_hash = &slide_hash_vmx;
+        ft.slide_hash = &slide_hash_vmx;
 #endif
 #ifdef POWER8_VSX_SLIDEHASH
     if (power_cpu_has_arch_2_07)
-        functable.slide_hash = &slide_hash_power8;
+        ft.slide_hash = &slide_hash_power8;
 #endif
 
     // longest_match_stub:
 #ifdef UNALIGNED_OK
 #  if defined(UNALIGNED64_OK) && defined(HAVE_BUILTIN_CTZLL)
-    functable.longest_match = &longest_match_unaligned_64;
+    ft.longest_match = &longest_match_unaligned_64;
 #  elif defined(HAVE_BUILTIN_CTZ)
-    functable.longest_match = &longest_match_unaligned_32;
+    ft.longest_match = &longest_match_unaligned_32;
 #  else
-    functable.longest_match = &longest_match_unaligned_16;
+    ft.longest_match = &longest_match_unaligned_16;
 #  endif
 #else
-    functable.longest_match = &longest_match_c;
+    ft.longest_match = &longest_match_c;
 #endif
 #if defined(X86_SSE2) && defined(HAVE_BUILTIN_CTZ)
     if (x86_cpu_has_sse2)
-        functable.longest_match = &longest_match_sse2;
+        ft.longest_match = &longest_match_sse2;
 #endif
 #if defined(X86_AVX2) && defined(HAVE_BUILTIN_CTZ)
     if (x86_cpu_has_avx2)
-        functable.longest_match = &longest_match_avx2;
+        ft.longest_match = &longest_match_avx2;
 #endif
 #if defined(ARM_NEON) && defined(HAVE_BUILTIN_CTZLL)
     if (arm_cpu_has_neon)
-        functable.longest_match = &longest_match_neon;
+        ft.longest_match = &longest_match_neon;
 #endif
 #ifdef POWER9
     if (power_cpu_has_arch_3_00)
-        functable.longest_match = &longest_match_power9;
+        ft.longest_match = &longest_match_power9;
 #endif
 
     // longest_match_slow_stub:
 #ifdef UNALIGNED_OK
 #  if defined(UNALIGNED64_OK) && defined(HAVE_BUILTIN_CTZLL)
-    functable.longest_match_slow = &longest_match_slow_unaligned_64;
+    ft.longest_match_slow = &longest_match_slow_unaligned_64;
 #  elif defined(HAVE_BUILTIN_CTZ)
-    functable.longest_match_slow = &longest_match_slow_unaligned_32;
+    ft.longest_match_slow = &longest_match_slow_unaligned_32;
 #  else
-    functable.longest_match_slow = &longest_match_slow_unaligned_16;
+    ft.longest_match_slow = &longest_match_slow_unaligned_16;
 #  endif
 #else
-    functable.longest_match_slow = &longest_match_slow_c;
+    ft.longest_match_slow = &longest_match_slow_c;
 #endif
 #if defined(X86_SSE2) && defined(HAVE_BUILTIN_CTZ)
     if (x86_cpu_has_sse2)
-        functable.longest_match_slow = &longest_match_slow_sse2;
+        ft.longest_match_slow = &longest_match_slow_sse2;
 #endif
 #if defined(X86_AVX2) && defined(HAVE_BUILTIN_CTZ)
     if (x86_cpu_has_avx2)
-        functable.longest_match_slow = &longest_match_slow_avx2;
+        ft.longest_match_slow = &longest_match_slow_avx2;
 #endif
 #if defined(ARM_NEON) && defined(HAVE_BUILTIN_CTZLL)
     if (arm_cpu_has_neon)
-        functable.longest_match_slow = &longest_match_slow_neon;
+        ft.longest_match_slow = &longest_match_slow_neon;
 #endif
 #ifdef POWER9
     if (power_cpu_has_arch_3_00)
-        functable.longest_match_slow = &longest_match_slow_power9;
+        ft.longest_match_slow = &longest_match_slow_power9;
 #endif
 
     // adler32_stub:
-    functable.adler32 = &adler32_c;
+    ft.adler32 = &adler32_c;
 #ifdef ARM_NEON_ADLER32
 #  ifndef ARM_NOCHECK_NEON
     if (arm_cpu_has_neon)
 #  endif
-        functable.adler32 = &adler32_neon;
+        ft.adler32 = &adler32_neon;
 #endif
 #ifdef X86_SSSE3_ADLER32
     if (x86_cpu_has_ssse3)
-        functable.adler32 = &adler32_ssse3;
+        ft.adler32 = &adler32_ssse3;
 #endif
 #ifdef X86_AVX2_ADLER32
     if (x86_cpu_has_avx2)
-        functable.adler32 = &adler32_avx2;
+        ft.adler32 = &adler32_avx2;
 #endif
 #ifdef X86_AVX512_ADLER32
     if (x86_cpu_has_avx512)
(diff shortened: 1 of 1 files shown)


pps83:
Note, I didn't try to merge checks, instead, the goal was to make the change less invasive. That is, it was done in a semi-automatic way:

image

same goes for all other stub functions

pps83:
Quite likely, this change may finally allow removing all that `Z_TLS` cruft.

phprus:
Atomic assignment is not guaranteed for pointers.

But, I do not know architectures where it would be dangerous in this ``functable`` case.

pps83:
> Atomic assignment is not guaranteed for pointers.

perhaps on x64 (not even sure), but afaik for 32-bit it's guaranteed. In any case, zlib-ng stands out among all other libs with such dynamic dispatch: I haven't seen a single one that uses tls for this usecase.

phprus:
Most CPU - guaranteed.
C standard - not guaranteed.

Dead2:
I think this looks like a good idea, and I give my tentative thumbs up 👍.

I never liked the TLS kludge, and as a bonus this might even avoid the Stabilizer crash on TLS code (That is a bug in Stabilizer, but avoiding it would be nice).

I still want to actually test this and have a closer look before approving this for merging, and I'd also like others to review this change since it is relatively complex and a tiny error can cause us to incorrectly use a sub-optimal function for example (as we have had happen before).

mtl1979:
The diff is a mess, but it looks like only moving initialization of the functable members from each stub to a new function, which will cause slight slowdown, but that is sub-second range. It really doesn't help readability of the file in my opinion.

functable.c
```diff
@@ -8,450 +8,497 @@
...
 #include "deflate.h"
 #include "deflate_p.h"
-
 #include "functable.h"
-
 #include "cpu_features.h"
 
-Z_INTERNAL Z_TLS struct functable_s functable;
+static void init_functable(void) {
+    uint32_t (* adler32)            (uint32_t adler, const uint8_t *buf, uint64_t len);
```
nmoinvaz: Why not use an intermediate `struct functable_s` instead of all these variables? Then at the end of the function copy one struct to another.

functable.c
```diff
@@ -8,450 +8,497 @@
...
+        compare256 = &compare256_avx2;
 #endif
 #ifdef POWER9
     if (power_cpu_has_arch_3_00)
-        functable.compare256 = &compare256_power9;
-#endif
+        compare256 = &compare256_power9;
+#endif
+
+
```
nmoinvaz: Extra white line

nmoinvaz:
I agree, that readability some what suffers.

functable.c
nmoinvaz: Where did global `struct functable_s functable;` go?

functable.c
```diff
@@ -8,450 +8,497 @@
 #include "crc32_braid_p.h"
 #include "deflate.h"
 #include "deflate_p.h"
-
 #include "functable.h"
-
 #include "cpu_features.h"
 
-Z_INTERNAL Z_TLS struct functable_s functable;
```
nmoinvaz, line 16: Where did global `struct functable_s functable` go?

functable.c
phprus: Copy one struct to another is not atomic on all CPUs.
Copy one pointer to another is atomic on most CPUs.

But I think it's better to use an intermediate ``struct functable_s`` instead of separate variables.

nmoinvaz: > Copy one struct to another is not atomic on all CPUs.
Copy one pointer to another is atomic on most CPUs.

It should have a comment where it copies each one individually telling why.

pps83: Yes, functable_s is a better idea. Copying whole table vs pointer by pointer wouldn't make a difference as long as individual pointers are copied atomically.

functable.c
pps83: It was always at the end of the file. I find it absolutely strange that this code compiles in c, in c++ that's a compilation error.

functable.c
phprus: > Copying whole table vs pointer by pointer wouldn't make a difference as long as individual pointers are copied atomically.

**No.** There are no guarantees how struct copying will be implemented.
The compiler can use ``rep movsb`` (x86) and then the copy will be byte by byte, not pointer to pointer (by pointer-size copy instruction).

phprus: @pps83 
I have a question, can the C compiler replace element-by-element struct copying with memcpy?
If yes, use struct are dangerous.

pps83:
> The diff is a mess, but it looks like only moving initialization of the functable members from each stub to a new function, 

Yes, the code for initialization of functable members was simply moved to standalone function.

> which will cause slight slowdown, but that is sub-second range.

There is no slowdown. Not sure what else to say about it, I guess it's quite clear.

> It really doesn't help readability of the file in my opinion.

The PR is made so that the change is easier to review. Follow up change to `init_functable` might be a good idea so that it follows not the messy way of function by function init, but instead has different blocks for cpu features. Similar to this code: https://github.com/mirror/x264/blob/master/common/x86/mc-c.c#L796

IMO, from the start function pointer init code should have been in a separate function, not spread out over different stub functions.

functable.c
phprus: I think that it can be better to copy pointers as follows:

``__atomic_store`` (gcc-4.7+, clang, other gcc-compatible, https://gcc.gnu.org/onlinedocs/gcc-4.7.0/gcc/_005f_005fatomic-Builtins.html)
Performance note:

https://github.com/boostorg/atomic/blob/c277de6b49804f759ccc429ce56fdc03684050b0/include/boost/atomic/detail/core_ops_gcc_atomic.hpp#L67-L78

or
``InterlockedExchangePointer`` (MSVC, https://learn.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedexchangepointer)

pps83: @phprus interlocked exchange could be used if TLS is removed. In this PR it's not needed because of TLS.

However, it seems that TLS is used because zlib-ng assumes that multiple cores on a CPU might have different features, or, to put it in other words, zlib-ng assumes that CPU tests have to be done for each thread. Not sure if that's a valid or useful assumption, simply because after `functions` table was initialized for the CPU/thread it doesn't mean that the thread will continue running on the same core. Perhaps, @mtl1979 could comment on that.

phprus: @pps83 
Oh! You are absolutely right!

I saw the next line:
```
-Z_INTERNAL Z_TLS struct functable_s functable;
```
 in diff and thought that we don't have TLS anymore and we need pointer copy atomicity.

mtl1979: In some architectures a thread is pinned to specific core, so it's safe to assume that the supported features don't change after initial check when using TLS. This applies to C code only, as with C++ the thread can switch cores if exception is thrown. 

Without TLS, there is also very small window during initialisation of function table, where another thread possibly could call same function and get different result or even invalid pointer. This is why we didn't initialise all the function table members in every stub function as that reduced the window to maximum of few nanoseconds.

ARM is one example of architecture that is known to mix cores with different supported features.

phprus: @mtl1979
C and C++ threads can be run by the OS-scheduler on any available hardware core.
To bind a thread to the hardware core, we must explicitly set the thread affinity. TLS binds data to the program thread, not to the CPU core.
Therefore, using TLS does not protect against CPUs with different feature sets.

mtl1979: zlib or zlib-ng doesn't start threads itself, so it doesn't have to worry about setting the affinity either... Even in my own programs, I haven't ever needed to set the thread affinity myself, it does that automatically.

If program is supposed to run on processor with mismatched cores, it would likely crash even without using zlib or gzip functions if it doesn't set thread affinity either implicitly or explicitly.

mtl1979:
> IMO, from the start function pointer init code should have been in a separate function, not spread out over different stub functions.

There is almost as many opinions as there is contributors to zlib-ng... Sometimes we have to make compromises to get things merged in. There was a reason why each stub only altered own member in the function table as the initial implementation didn't use TLS and many programs did use zlib from more than one thread, so there was quite high chance that one thread could overwrite or corrupt data written by another thread.

functable.c
pps83: That’s what I also thought, that something like windows’  SetThreadAfinity needs to be used to force thread to be executed on specific core.

Arm cpus had low/high power cores for a long time, new x86 CPUs also now have the same stuff. Do these low/high power cores have different CPU features?? I’d expect them to be the same, otherwise this would complicate lots of stuff.

phprus: @pps83 
@mtl1979 

So, if TLS remains anyway...

In this case, there will be no conflicts when using ``functable``.
We can simplify PR by removed the temporary local variable ``struct functable_s ft`` and do the assignment directly to the members of ``functable``.

Using a common ``init_functable`` function will speed up all calls to versioned functions except the only first call of any versioned function in each thread.
The slowdown of the first call will be less than the total slowdown of all functions before RP, since the compiler will most likely combine the same if.

pps83:
> > IMO, from the start function pointer init code should have been in a separate function, not spread out over different stub functions.
> 
> There is almost as many opinions as there is contributors to zlib-ng... Sometimes we have to make compromises to get things merged in. There was a reason why each stub only altered own member in the function table as the initial implementation didn't use TLS and many programs did use zlib from more than one thread, so there was quite high chance that one thread could overwrite or corrupt data written by another thread.

I wanted to edit my reply to explain why, but cancelled the edit. The idea is that initialization of function pointers based on testing cpu features is vastly different concept than actual calling by function pointers. It’s usually better to keep conceptually different stuff in different functions. Because all this was mixed current zlib-ng ended up to have hidden bugs: some of these stub functions do not test cpu features. This PR not only does some clean up, it also fixes this bug.

mtl1979:
> I wanted to edit my reply to explain why, but cancelled the edit. The idea is that initialization of function pointers based on testing cpu features is vastly different concept than actual calling by function pointers. It’s usually better to keep conceptually different stuff in different functions. Because all this was mixed current zlib-ng ended up to have hidden bugs: some of these stub functions do not test cpu features. This PR not only does some clean up, it also fixes this bug.

Mixing cleanup and bug fixes is generally discouraged. Some stub functions are not supposed to be first function table member to be called, they are always called after another member function is called first, so they don't actually need to initialize the feature flags based on `cpuid` or similar kernel feature.

phprus:
> Some stub functions are not supposed to be first function table member to be called, they are always called after another member function is called first, so they don't actually need to initialize the feature flags based on cpuid or similar kernel feature.

This approach complicates the code and adds non-obvious dependencies.
Merge initialization code in single function solves this problem without overhead.

functable.c
phprus: @mtl1979


Single ``init_function`` will help solve another issue.

All CPU feature flags is not TLS:
https://github.com/zlib-ng/zlib-ng/blob/41a5e00e7c9fd8fce5c2135fe412d4e0f3f15eb3/arch/arm/arm_features.c#L74-L75

https://github.com/zlib-ng/zlib-ng/blob/41a5e00e7c9fd8fce5c2135fe412d4e0f3f15eb3/arch/x86/x86_features.c#L21-L32

This makes useless to use TLS for ``functlable``.

After this PR (single ``cpu_check_features()`` call) we replace ``cpu_check_features()`` and ``*_cpu_has_*`` variables with this code:

```c

struct x86_cpu_features
{
    bool x86_cpu_has_...;
};

struct arm_cpu_features
{
    bool arm_cpu_has_...;
};

void x86_check_features(struct x86_cpu_features *f);
void arm_check_features(struct arm_cpu_features *f);

// Platform dependent:
#define cpu_features       x86_cpu_features;
#define cpu_check_features x86_check_features


static void init_functable(void) {

...
cpu_features cpu;
cpu_check_features(&cpu);

...
if (cpu.*_cpu_has_*) {
    ...
}

...

```

This logic ensures that if an external application binds threads to the CPU correctly, then zlib-ng will always choose the correct implementation for the thread/CPU.

mtl1979: @phprus It's true that currently checking the CPU features doesn't utilize thread-local storage or is thread-safe... None of us long-time contributors is using mismatched SMP setup as their development machine, so there wasn't easy way to test the change.

As a sidenote, TLS support can still be disabled, so any change can't break thread-safety more than it is already broken.

mtl1979: > After this PR (single `cpu_check_features()` call) we replace `cpu_check_features()` and `*_cpu_has_*` variables with this code:
> 
> ```c
> struct x86_cpu_features
> {
>     bool x86_cpu_has_...;
> };
> 
> struct arm_cpu_features
> {
>     bool arm_cpu_has_...;
> };
> 
> void x86_check_features(struct x86_cpu_features *f);
> void arm_check_features(struct arm_cpu_features *f);
> 
> // Platform dependent:
> #define cpu_features       x86_cpu_features;
> #define cpu_check_features x86_check_features
> 
> 
> static void init_functable(void) {
> 
> ...
> cpu_features cpu;
> cpu_check_features(&cpu);
> 
> ...
> if (cpu.*_cpu_has_*) {
>     ...
> }
> 
> ...
> ```
> 
> This logic ensures that if an external application binds threads to the CPU correctly, then zlib-ng will always choose the correct implementation for the thread/CPU.

Some could argue that thread-safety of CPU feature checks should indeed be separate PR, but reviewed before instead of after this PR.

pps83:
> Mixing cleanup and bug fixes is generally discouraged.

I won't go into details why I made the change, but while doing it I noticed that zlib-ng had the bug. There is also a [PR that fixes it](https://github.com/zlib-ng/zlib-ng/pull/1172). My change fixes it as well, since there only one init function now.

functable.c
mtl1979: Redefining something to same type is silently ignored as forward declaring in C. This allows resolving possible circular dependency.

mtl1979:
> This approach complicates the code and adds non-obvious dependencies. Merge initialization code in single function solves this problem without overhead.

This part of the code has been neglected for quite a while as people have worked on more "important" bugs in the codebase. I'm just trying to make sure the fix in this PR is the best possible and that means listening to comments by as many people as possible before making the final choice of approving or rejecting. 

Like I've said in most controversial pull requests, at least two of the long-time contributors must approve the PR before it can be merged. Since I started contributing, I've mostly reviewed pull requests that involve Windows (and other non-Linux operating systems, including FreeBSD, OpenBSD and NetBSD) or non-x86 processors, as my knowledge of modern x86 processors and their instruction sets is quite limited.

functable.c
pps83: @phprus you broke my nuclear weapon :).
I recall seeing TLS, made some discussion years ago (perhaps a pr or an issue on gh), saw that was total nonsense. I did remove that mess from my fork. I wanted to see how far that would go this time. Now all the maintainers could be reasonably sure that ALL that discussions, and code related to TLS use was pretty much nonsense, and the only potentially valid reason was atomic pointer swap. IMO, all the stubs should be removed, `functable` should be zero init, `inflateInit`/`deflateInit` should call `functable_init` (to avoid pointless discussion, by `inflateInit`/`deflateInit` I also mean all the functions that could be first-time called by the user code).

pps83: > The slowdown of the first call will be less than the total slowdown of all functions before RP, since the compiler will most likely combine the same if.

I think any reasonably programmer would understand that it's pointless to discuss performance in this case. It just make no sense at all, it distracts from stuff that matters. The code had bugs, and mystical perf is second to correctness, especially when it's crystal clear that this thing has no perf overhead. It's not like it called once per 100-byte block. Even if it was called once per compres/uncompress sequence it's still would be a non-issue.

mtl1979: I've been programmer for few months short of 40 years.... And I see your argument is still overly simplified... I'm not saying this PR is utter waste of time, I'm just saying it can be further improved before we approve it.

functable.c
```diff
@@ -8,447 +8,476 @@
...
+        ft.compare256 = &compare256_avx2;
 #endif
 #ifdef POWER9
     if (power_cpu_has_arch_3_00)
-        functable.compare256 = &compare256_power9;
-#endif
+        ft.compare256 = &compare256_power9;
+#endif
+
+    functable.adler32 = ft.adler32;
```
nmoinvaz, line 368: Add comment above saying, something to the effect of: `// Assign function pointers individually for atomic operation`

functable.c
phprus: @pps83 
I apologize :)

phprus: @mtl1979

> Some could argue that thread-safety of CPU feature checks should indeed be separate PR, but reviewed before instead of after this PR.

I don't understand you.

If we support CPU with different functions on different cores we need TLS.
Then the current implementation is broken since ``cpu_check_features`` does not use TLS. TLS is not free, so it should be used less.
Because of this, I think it's best to have one point of ``cpu_check_features`` call and select for all versioned features.

If we don't we not need TLS.
In this case we need atomic initialization of pointers in ``functable``. And it will also be easier to do it in one place.

So I guess changing ``cpu_check_features`` will be easier after this PR.

mtl1979: I didn't write the current implementation of CPU feature checking... It was written gradually a long time before we even had function table. Like I said elsewhere in this PR, I already know the current implementation is buggy, but instead of introducing more bugs or regressions, we should fix the current ones first...

I've been developer for long enough time to notice when people try to mix opinions with facts. This will only complicate reviewing pull requests as none of us know everything about every part of the code, so when people try to confuse us, we need to research to make sure what information we have is actually factual and not just result of trolling.

A lot of time I might sound like a real bitch or arrogant, but that is just to make sure every pull request gets reviewed properly so we don't need to revert or rewrite them later wasting valuable time. @nmoinvaz and I have been talking about improving the CI to catch even more issues, but those improvements have not been merged in yet.

nmoinvaz (approved):

pps83:
@Dead2 ^^^

Dead2 (approved):

PR #1406, opened 2023-01-18
Match __builtin_ctzl/__builtin_ctzll signatures

pps83:
make sure input/output args match original functions from clang/gcc


nmoinvaz (approved):
LGTM. Checked signature of GCC function [here](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html).

Dead2 (approved):

Dead2:
@pps83 This needs a rebase, could you please do so?

PR #1407, opened 2023-01-18
Do not use cpuid.h on Windows

pps83:
According to msdn docs __cpuid/__cpuidex/_xgetbv are provided by intrin.h and should be used on windows

### DIFF
--- a/arch/x86/x86_features.c
+++ b/arch/x86/x86_features.c
@@ -9,7 +9,7 @@
 
 #include "../../zbuild.h"
 
-#ifdef _MSC_VER
+#ifdef _WIN32
 #  include <intrin.h>
 #else
 // Newer versions of GCC and clang come with cpuid.h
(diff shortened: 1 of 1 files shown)


pps83:
this works equally with mingw/clang-cl. On the contrary, if intrin.h was already included cpuid.h won't work with mingw

mtl1979:
In general, we try to include the smallest usable header file to speed up the compilation... When including headers on Windows, it's not always possible, so we have to include header that pulls in a lot of unnecessary lines or files.

However, supporting more than just Visual C++ has been our plan from very beginning and we have gradually improved support for MinGW, MSYS and Cygwin.

nmoinvaz (approved):

Dead2 (approved):

arch/x86/x86_features.c
```diff
@@ -60,7 +60,7 @@ static inline void cpuidex(int info, int subinfo, unsigned* eax, unsigned* ebx,
 }
 
 static inline uint64_t xgetbv(unsigned int xcr) {
-#ifdef _MSC_VER
```
nmoinvaz, line 63: This change causes build failures on mingw on Ubuntu and Windows.

mtl1979: GCC for MinGW needs "-mxsave" for `_xgetbv`. I will make a commit to fix that.

PR #1408, opened 2023-01-18
Add assert check for tzcnt if building with X86_NOCHECK_TZCNT

pps83:


nmoinvaz (approved):
LGTM

mtl1979:
The whole point for the preprocessor defines is to make feature checks no-op, so I'm not sure adding assert check is what we want. Almost all 64-bit Intel/AMD processors support SSE2, so the assertion failure is very unlikely to trigger.

Dead2:
Hmm, now that the NOCHECK override is only used in the functable stubs, it really does not make as much of a difference having it anymore.
IIRC we used to have those in some hotter functions, and that was a good way to avoid an unnecessary speed penalty.

If those defines also skipped compiling in the C versions of the functions, then it would have still made sense to keep them.

So I guess we should first decide between: Keep them / Remove them / Expand the use to reduce compiled code size.
Personally I would go with options 2 or 3 unless someone has a compelling argument for 1.

mtl1979:
@Dead2 Skipping compiling the C versions might interfere with the benchmarks, but a workaround can be added for that...

Dead2:
> @Dead2 Skipping compiling the C versions might interfere with the benchmarks, but a workaround can be added for that...

Yeah, and we already don't benchmark other functions that we didn't have compiled in (due to missing compiler support), so that should not be a big problem. We that maintain it need to be aware of that distinction of course.

It really does not make sense to have the C version be present in the library, completely unused, when functable is overridden to never use it.

mtl1979:
@Dead2 Only valid reason would be to compare outputs of different variants... But we can precalculate the result already...

pps83:
> Hmm, now that the NOCHECK override is only used in the functable stubs, it really does not make as much of a difference having it anymore.

@Dead2 TYes, makes no sense to have force-sse2. I made a PR for that: [Remove pointless FORCE_SSE2 feature](https://github.com/zlib-ng/zlib-ng/pull/1413).


#1413 should be meged first, then #1408 will be rebased.

pps83:
> The whole point for the preprocessor defines is to make feature checks no-op, so I'm not sure adding assert check is what we want. Almost all 64-bit Intel/AMD processors support SSE2, so the assertion failure is very unlikely to trigger.

I think it's really a good idea to have the assert in there. I'd even say that this assert must have been added when the NOCHECK defines were added. #1413 should remove ` X86_NOCHECK_SSE2` from the PR, but `X86_NOCHECK_TZCNT` should be checked. This way, if CPU doesn't have `tzcnt` assert would trigger with a message.

pps83:
update: removed assert for X86_NOCHECK_SSE2 (should re removed by https://github.com/zlib-ng/zlib-ng/pull/1413)
@Dead2 ^^^

Dead2 (approved):

PR #1412, opened 2023-01-21
MinGW build fix

mtl1979:
Separated from #1384
* Add test for XSAVE intrinsics
* Fix out-of-tree configure build
* Fix configure using incorrect linker definition file
* Fix test binaries not linking against gzip utility functions when zlib-ng doesn't include them in static/shared library


nmoinvaz:
Do we need a `#ifdef X86_XSAVE` in x86_features.c?

mtl1979:
> Do we need a `#ifdef X86_XSAVE` in x86_features.c?

I thought about that, but it's not relevant to MinGW or Windows API specifically.

cmake/detect-intrinsics.cmake
```diff
@@ -536,3 +536,21 @@ macro(check_vgfma_intrinsics)
         HAVE_VGFMA_INTRIN FAIL_REGEX "not supported")
     set(CMAKE_REQUIRED_FLAGS)
 endmacro()
+
+macro(check_xsave_intrinsics)
+    if(NOT NATIVEFLAG AND NOT MSVC)
+        set(XSAVEFLAG "-mxsave")
+    endif()
+    set(CMAKE_REQUIRED_FLAGS "${XSAVEFLAG}")
```
nmoinvaz: This is missing `${NATIVEFLAG}` after `${XSAVEFLAG}`

nmoinvaz:
I think each of these should be its own commit.

* Fix out-of-tree configure build
* Fix configure using incorrect linker definition file
* Fix test binaries not linking against gzip utility functions when zlib-ng doesn't include them in static/shared library

mtl1979:
> I think each of these should be its own commit.
> 
> * Fix out-of-tree configure build
> * Fix configure using incorrect linker definition file
> * Fix test binaries not linking against gzip utility functions when zlib-ng doesn't include them in static/shared library

I will make a copy of the current branch and then try to split it to 4 separate commits.... I will take some time, but I will force push when it's ready...

pps83:
Would it make sense to simply fix `xgetbv` and use 

```
__asm__ ( ".byte 0x0f, 0x01, 0xd0" : "=a"(eax), "=d"(edx) : "c"(xcr));
```
instead of `_xgetbv` when compiling with `__GNUC__`?

mtl1979:
> Would it make sense to simply fix `xgetbv` and use
> 
> ```
> __asm__ ( ".byte 0x0f, 0x01, 0xd0" : "=a"(eax), "=d"(edx) : "c"(xcr));
> ```
> 
> instead of `_xgetbv` when compiling with `__GNUC__`?

Not really... Using inline assembly for 64-bit Windows code is discouraged as it will disable some optimisations and sanity checks. That ASM block is just workaround for old compilers that didn't yet understand the mnemonic for the instruction.

mtl1979:
Rebased.

nmoinvaz:
> [Fix building resource files during out-of-tree build with configure](https://github.com/zlib-ng/zlib-ng/pull/1412/commits/d1d83de238f3dcfb756e643ca3f43ad51cd43194)

I think perhaps we need a CI in the matrix for Configure script on Windows. I think it should work fine if it uses `shell: bash`. Or it might need to be run in a separate shell? Would be a good idea for this or different PR.

mtl1979:
> > [Fix building resource files during out-of-tree build with configure](https://github.com/zlib-ng/zlib-ng/pull/1412/commits/d1d83de238f3dcfb756e643ca3f43ad51cd43194)
> 
> I think perhaps we need a CI in the matrix for Configure script on Windows. I think it should work fine if it uses `shell: bash`. Or it might need to be run in a separate shell? Would be a good idea for this or different PR.

Maybe leave it for next PR... Then it can be tested without delaying these fixes too much...

Dead2 (approved):

PR #1413, opened 2023-01-22
Remove pointless FORCE_SSE2 feature

pps83:


mtl1979:
This is clearly "my way is better" type of pull request... All the `FORCE_*` preprocessor defines serve a clear purpose that contributors who are familiar with our codebase already know.

pps83:
> This is clearly "my way is better" type of pull request... 

🤦‍♂️

> All the `FORCE_*` preprocessor defines serve a clear purpose that contributors who are familiar with our codebase already know.

There is absolutely NO POINT to have this FORCE_SSE2! It disables run-time checks for sse2, which is done ONLY ONCE in stub functions. It's just wrong to have this ability. That's why this PR is made.

Also, see this comment: https://github.com/zlib-ng/zlib-ng/pull/1408#issuecomment-1396975698

mtl1979:
> > This is clearly "my way is better" type of pull request...
> 
> 🤦‍♂️
> 
> > All the `FORCE_*` preprocessor defines serve a clear purpose that contributors who are familiar with our codebase already know.
> 
> There is absolutely NO POINT to have this FORCE_SSE2! It disables run-time checks for sse2, which is done ONLY ONCE in stub functions. It's just wrong to have this ability. That's why this PR is made.
> 
> [#1408 (comment)](https://github.com/zlib-ng/zlib-ng/pull/1408#issuecomment-1396975698)

It doesn't really matter how many times something is done in the code, especially if the code isn't thread-safe. I've seen programs that have hundreds of threads and most of the threads using zlib... There is also processors that have different features for each core, so testing features on just one core doesn't make sense at all...

pps83:
> It doesn't really matter how many times something is done in the code, especially if the code isn't thread-safe. I've seen programs that have hundreds of threads and most of the threads using zlib... There is also processors that have different features for each core, so testing features on just one core doesn't make sense at all...

In any case, the SSE2 check really has 0 cost. Weather you do it once when you run your program, or once when you call deflate function, or once per thread. The cost is ZERO.

mtl1979:
> > It doesn't really matter how many times something is done in the code, especially if the code isn't thread-safe. I've seen programs that have hundreds of threads and most of the threads using zlib... There is also processors that have different features for each core, so testing features on just one core doesn't make sense at all...
> 
> In any case, the SSE2 check really has 0 cost. Weather you do it once when you run your program, or once when you call deflate function, or once per thread. The cost is ZERO.

Stating opinions as facts is just pointless... I've spent last few months fixing the code because most of the contributors have zero knowledge that zlib-ng is supposed to even compile on Windows and what are the main differences between Visual C++ vs. gcc and clang.

pps83:
Sse2 check is zero cost. I won't be trying to convince you to change your opinion.

pps83:
> I've spent last few months fixing the code because most of the contributors have zero knowledge that zlib-ng is supposed to even compile on Windows and what are the main differences between Visual C++ vs. gcc and clang.

I don't see how is that relevant to this PR.

mtl1979:
> Sse2 check is zero cost. I won't be trying to convince you to change your opinion.

You need two out of three long-time contributors to accept the PR, or otherwise it's just waste of time... Besides me, @Dead2 and @nmoinvaz are the only two long-time contributors.

mtl1979:
> > I've spent last few months fixing the code because most of the contributors have zero knowledge that zlib-ng is supposed to even compile on Windows and what are the main differences between Visual C++ vs. gcc and clang.
> 
> I don't see how is that relevant to this PR.

NMake/Windows use `FORCE_*` preprocessor defines as it's impossible to run operating system without specific feature and there is no configure or cmake to detect the support. Thus, zero reason to even try checking the CPU support.

pps83:
> > > I've spent last few months fixing the code because most of the contributors have zero knowledge that zlib-ng is supposed to even compile on Windows and what are the main differences between Visual C++ vs. gcc and clang.
> > 
> > 
> > I don't see how is that relevant to this PR.
> 
> NMake/Windows use `FORCE_*` preprocessor defines as it's impossible to run operating system without specific feature and there is no configure or cmake to detect the support. Thus, zero reason to even try checking the CPU support.

So, then entire checks for SSE2 could be removed then. The point of this PR is clearly described the the PR title: it removes pointless ability to **force SSE2**, nothing else. The scope is narrow and clear, there is no point to introduce unrelated stuff as a reason why FORCE_SSE2 shouldn't be removed

mtl1979:
> So, then entire checks for SSE2 could be removed then. The point of this PR is clearly described the the PR title: it removes pointless ability to **force SSE2**, nothing else. The scope is narrow and clear, there is no point to introduce unrelated stuff as a reason why FORCE_SSE2 shouldn't be removed

We can't remove checks for SSE2 until we remove support for 32-bit 80x86 builds. Even if new code should be built for AMD64/x86_64, there is still cases when 32-bit builds are needed, especially for laptops or other portable devices.

I've already explained why forcing support for some feature check serve a purpose, so removing them will re-introduce issues we already decided to avoid.

pps83:
> I've already explained why forcing support for some feature check serve a purpose, so removing them will re-introduce issues we already decided to avoid.

Let's not deviate from the PR. All I can say, there is no reason to have ability to force SSE2. What purpose/benefit does FORCE_SSE2 feature bring into zlib-ng that makes this PR not acceptable?

nmoinvaz (approved):
This LGTM. There is very little value in skipping the check.

mtl1979:
> > I've already explained why forcing support for some feature check serve a purpose, so removing them will re-introduce issues we already decided to avoid.
> 
> Let's not deviate from the PR. All I can say, there is no reason to have ability to force SSE2. What purpose/benefit does FORCE_SSE2 feature bring into zlib-ng that makes this PR not acceptable?

We can remove the unoptimized C version of the functions to reduce binary size.

pps83:
> We can remove the unoptimized C version of the functions to reduce binary size.

I made a test, and manually removed those _c functions where sse2 versions could be forced with `X86_NOCHECK_SSE2`. x64 didn't have any size diff (I tested with ms compiler), while x86 had 500 extra bytes. Also, `X86_NOCHECK_SSE2` is only relevant to x86, x64 always forces sse2.

mtl1979:
It's no surprise that there is no size change with x64 as compiler has dead code elimination, that's the whole point of having `FORCE_SSE2` in the first place. Removing the `*_c` versions with x86 (32-bit) doesn't make sense at all, as there might not be any other alternative depending on what is the actual processor the compiler defaults to. Generally x86 compilers should default to at least Pentium (80586) or Pentium Pro (80686).

pps83:
> It's no surprise that there is no size change with x64 as compiler has dead code elimination, that's the whole point of having `FORCE_SSE2` in the first place. Removing the `*_c` versions with x86 (32-bit) doesn't make sense at all, as there might not be any other alternative depending on what is the actual processor the compiler defaults to. Generally x86 compilers should default to at least Pentium (80586) or Pentium Pro (80686).

I said this:

> manually removed those _c functions where sse2 versions could be forced with X86_NOCHECK_SSE2

x86 had 117248 -> 116736 size change.

mtl1979:
My point is still that modern processors can utilize same optimizations for 32-bit binaries as they can utilize for 64-bit binaries, as almost all vector instructions can be used in 32-bit mode. However, we don't require that any of the 32-bit platforms we support actually are limited to modern processors... For 32-bit ARM, we support anything from ARMv2 to ARMv8, including processors that don't support unaligned access. For 32-bit PowerPC, we support anything from PowerPC 604 to G4.

pps83:
I’m not sure what you mean. I said this:

> Also, X86_NOCHECK_SSE2 is only relevant to x86, x64 always forces sse2.

x64 builds always force no sse2 check. In other words, removal of force sse feature onky affects x86 builds and has no impact on x64

mtl1979:
> I’m not sure what you mean. I said this:
> 
> > Also, X86_NOCHECK_SSE2 is only relevant to x86, x64 always forces sse2.
> 
> x64 builds always force no sse2 check. In other words, removal of force sse feature onky affects x86 builds and has no impact on x64

I'm talking about builds with `-m32` command-line option.

pps83:
@Dead2 ^^^

Dead2:
I would much prefer seeing this feature expanded on rather than removed.

IE: If `FORCE_SSE2`, then don't even compile in the plain-c version.
And perhaps also enable the -msse2 (and similar) compiler flags, so the compilers can utilize SSE2 instructions anywhere in the code (This already happens automatically on x86-64, but not on x86-32).

pps83:
That's not related to the PR. As I mentioned, sse2 is always forced on x64 builds (eg FORCE_SSE2 is meaningless for x64). Preventing compilation of dead code is unrelated change. zlib-ng is overloaded with deadcode, it's better to remove it than blocking all other changes. One such instance: the TLS for mystical multicore machines with different cpu features: all that is dead code that keeps dragging development and blocks changes all over the place, all while that code in zlib-ng was always dead and broken.

mtl1979:
TLS is not used just for mystical CPUs that consists cores that have different features... It's also used for any CPUs that have more than 1 core as those can run different threads that access zlib functions and cause undefined behaviour due to race conditions between those threads. Like I've said before, there is no thread synchronisation in C.

Like I said earlier, most of those mystical weird CPUs are OEM ARM-based processors that are known to use different features to do different parts of mobile data communication or signal processing. In the old days there was also mix of 486/Pentium or 8086/80186 processors... Mixing different CPU features within CPU cores or CPU sockets was common in Japan long before modern mobile processors (there was even special upgrade boards).

Dead2:
> That's not related to the PR. As I mentioned, sse2 is always forced on x64 builds (eg FORCE_SSE2 is meaningless for x64). Preventing compilation of dead code is unrelated change. zlib-ng is overloaded with deadcode, it's better to remove it than blocking all other changes.

It is very much related to this PR, since this PR does not only remove this switch from x86-64 builds but also from x86-32 builds. It would also be unnecessary code and documentation churn to remove this just to later add it back with a few small additions. This switch used to have a bigger impact, true, and it was a missed opportunity when it was sidelined rather than updated to the way it should have been in the first case when the function selections were taken out of the hot code.
I doubt this switch hurts anyone in its current state, and I don't see what changes it is currently blocking, please enlighten me.

> One such instance: the TLS for mystical multicore machines with different cpu features: all that is dead code that keeps dragging development and blocks changes all over the place, all while that code in zlib-ng was always dead and broken.

TLS was never added for supporting cpus with different features per core, it was added to avoid corrupted pointer reads, ie to prevent one core from reading a value that is in the middle of being written by another core. I never much liked using TLS for this, but there you have the reason.

PR #1420, opened 2023-02-04
Replace __builtin_ctz[ll] fallback functions.

### LINKED ISSUE #1419: TZCNT performance question
A ``__builtin_ctz*`` emulation has been written for Microsoft Visual C++:
https://github.com/zlib-ng/zlib-ng/blob/d144fc06bfa575ddfa8143cedd0873f36a2c7568/fallback_builtins.h#L15-L25

For x86 there is a check for the presence of the ``tzcnt`` instruction. In this case, even if the processor does not support ``tzcnt``, the code will execute correctly and return the correct value (https://stackoverflow.com/questions/43880227/why-does-tzcnt-work-for-my-sandy-bridge-processor).

I don't understand if the overhead of the ``if (x86_cpu_has_tzcnt)`` condition is really less than the performance difference of the real ``tzcnt`` and the old ``rep bsf``?
Maybe it makes sense to completely remove this runtime check and for x86 always call ``tzcnt``? This can simplify the dynamic dispatch code by reducing the number of cpu features flags in global variables.

cc @Dead2, @mtl1979 , @pps83

Dead2:
Replaces __builtin_ctz[ll] fallback functions with branchless implementations.
Added debug assert check for value = 0.
Added more details to the comment to avoid future confusion.
Added fallback logic for older MSVC versions, just in case.

Resolves #1419


fallback_builtins.h
```diff
@@ -4,41 +4,43 @@
...
+ * Performance tzcnt/bsf is identical on Intel cpu, tzcnt is faster than bsf on AMD cpu.
  */
 static __forceinline int __builtin_ctz(unsigned int value) {
-#ifdef X86_FEATURES
-#  ifndef X86_NOCHECK_TZCNT
-    if (x86_cpu_has_tzcnt)
-#  endif
-        return (int)_tzcnt_u32(value);
-#endif
+    assert(value != 0);
```
pps83: 1. `assert(value != 0);` is needed only for `_BitScanForward` part. `tzcnt` does work for value==0. Same goes for `__builtin_ctzll` bellow.
2. I'm not entirely sure if presence of `_tzcnt_u32` depends on _MSC_VER >= 1700.

nmoinvaz: Should we be using `Assert` as defined in `zbuild.h`?

nmoinvaz (commented):
LGTM, just have one question.

fallback_builtins.h
pps83: aren't non-BMI cpus still common? That means that if somebody have 10 year old PC switching to new VS will make your program crash (unless tzcnt is automatically executed as bsf on older CPUs). If it gets executed as bsf, then assert is needed.

mtl1979: Should use `Assert` for consistency.

fallback_builtins.h
```diff
@@ -4,41 +4,43 @@
...
+ * Because of that assumption trailing_zero is not initialized and the return value is not checked.
  */
 static __forceinline int __builtin_ctzll(unsigned long long value) {
-#ifdef X86_FEATURES
-#  ifndef X86_NOCHECK_TZCNT
-    if (x86_cpu_has_tzcnt)
-#  endif
-        return (int)_tzcnt_u64(value);
-#endif
+    assert(value != 0);
```
mtl1979: Should use `Assert` for consistency.

fallback_builtins.h
mtl1979: 10+-year old PC is likely running 32-bit Windows Vista by default, and means it can't run latest Visual Studio which is 64-bit binary. I still use Visual Studio 2013 on my Windows Vista laptop (bought in 2009, Intel Core 2 Duo), as it is the most compatible version. There was no need for 64-bit Windows as the maximum installable memory was 4 GB.

fallback_builtins.h
```diff
@@ -4,41 +4,43 @@
...
-        return (int)_tzcnt_u32(value);
-#endif
+    assert(value != 0);
+# if (_MSC_VER < 1700)
     unsigned long trailing_zero;
     _BitScanForward(&trailing_zero, value);
     return (int)trailing_zero;
+# else
+    return (int)_tzcnt_u32(value);
+# endif
```
phprus: ``tzcnt`` (``_tzcnt_u32``, ``_tzcnt_u64``) is only available on x86:

```c++
#ifdef X86_FEATURES
    return (int)_tzcnt_u32(value);
# else
    unsigned long trailing_zero;
    _BitScanForward(&trailing_zero, value);
    return (int)trailing_zero;
# endif
```

fallback_builtins.h
Dead2: The assert is for the function as a whole because we define the function as unsafe for zero, so better that we get warnings because they would mean a missing check in the caller code.

I reviewed how some opensource projects used tzcnt, and found this one https://github.com/opencv/opencv/issues/7626 , and that seems to be a pretty clear indication that older MSVC likely did not support this. We could have skipped this backwards compatibility, but I figured it hurts nobody by being there.

Dead2: Yes, fixed

Dead2: fixed

fallback_builtins.h
Dead2: fixed

fallback_builtins.h
Dead2: Fixed

fallback_builtins.h
Dead2: Also, yes, non-BMI cpus execute tzcnt as bsf because tzcnt gets interpreted as just another bsf instruction on those cpus due to them being identical except for something (bit / flag? I forget the specifics) that bsf just ignores anyway. The incompatibility between them is only for input values of 0.

fallback_builtins.h
```diff
@@ -9,36 +9,37 @@
...
  */
 static __forceinline int __builtin_ctz(unsigned int value) {
-#ifdef X86_FEATURES
-#  ifndef X86_NOCHECK_TZCNT
-    if (x86_cpu_has_tzcnt)
-#  endif
-        return (int)_tzcnt_u32(value);
-#endif
+    Assert(value != 0);
+# if X86_FEATURES && !(_MSC_VER < 1700)
```
phprus:
```c++
defined(X86_FEATURES)
```
and below.

since the ``X86_FEATURES`` macro is defined without a value:
https://github.com/zlib-ng/zlib-ng/blob/d144fc06bfa575ddfa8143cedd0873f36a2c7568/CMakeLists.txt#L743

Dead2: OMG, should not program before breakfast.. 
Shuffled the code, copy-pasted that in and then completely forgot to add the defined() 🤦‍♂️

phprus:
I think we can remove the ``x86_cpu_has_tzcnt`` variable since it's not used anywhere else:

https://github.com/zlib-ng/zlib-ng/blob/d144fc06bfa575ddfa8143cedd0873f36a2c7568/arch/x86/x86_features.h#L18

https://github.com/zlib-ng/zlib-ng/blob/d144fc06bfa575ddfa8143cedd0873f36a2c7568/arch/x86/x86_features.c#L30

https://github.com/zlib-ng/zlib-ng/blob/d144fc06bfa575ddfa8143cedd0873f36a2c7568/arch/x86/x86_features.c#L98-L100

https://github.com/zlib-ng/zlib-ng/blob/d144fc06bfa575ddfa8143cedd0873f36a2c7568/arch/x86/x86_features.c#L119

Dead2:
@phprus Yes, I also considered removing those (actually I did at first), and I still might.
But I figured it also might as well be discussed first, and added as a separate commit later.
If this PR goes in, I don't really see us using this detection for anything in the future either.

nmoinvaz (approved):

pps83 (approved):

phprus (approved):

fallback_builtins.h
KungFuJesus: Yeah I believe the opcode is basically the same as BSF.  I tested things with a Core 2 Quad a while back and noticed that they are nearly interchangeable.

PR #1423, opened 2023-02-06
Functable cleanup

Dead2:
- Simplified functable.c
- Combined checks that were identical and rewritten some that were not identical but were closely related.
- Made longest_match and compare256 use the X86_NOCHECK_SSE2 override, thus now those are also automatically enabled on x86_64.

There is no easy way to do this, but I tried to split the changes into two commits. Easiest for review is to just look at the resulting file.
I think I have gotten the ordering/priority of the checks right, but please verify if you can.

There is still a whole lot of separate defines that probably could be combined. Just some examples: `ARM_NEON`, `ARM_NEON_ADLER32`, `ARM_NEON_SLIDEHASH` and `ARM_NEON_CHUNKSET`.
I bet those could all have used `ARM_NEON` instead. If we want to have the ability to specifically disable specific optimized functions, I'd have preferred we just use `ARM_NEON` and then check for `ifndef DISABLE_NEON_CHUNKSET` so they don't need to be set by the build system, but can be added manually if someone wanted to disable specifics.
I might make a separate PR for that.

### DIFF
--- a/functable.c
+++ b/functable.c
@@ -16,310 +16,219 @@ static void init_functable(void) {
 
     cpu_check_features();
 
-    // update_hash_stub:
+    // Generic code
     ft.update_hash = &update_hash_c;
-#ifdef X86_SSE42_CRC_HASH
-    if (x86_cpu_has_sse42)
-        ft.update_hash = &update_hash_sse4;
-#elif defined(ARM_ACLE_CRC_HASH)
-    if (arm_cpu_has_crc32)
-        ft.update_hash = &update_hash_acle;
-#endif
-
-    // insert_string_stub:
     ft.insert_string = &insert_string_c;
-#ifdef X86_SSE42_CRC_HASH
-    if (x86_cpu_has_sse42)
-        ft.insert_string = &insert_string_sse4;
-#elif defined(ARM_ACLE_CRC_HASH)
-    if (arm_cpu_has_crc32)
-        ft.insert_string = &insert_string_acle;
-#endif
-
-    // quick_insert_string_stub:
     ft.quick_insert_string = &quick_insert_string_c;
-#ifdef X86_SSE42_CRC_HASH
-    if (x86_cpu_has_sse42)
-        ft.quick_insert_string = &quick_insert_string_sse4;
-#elif defined(ARM_ACLE_CRC_HASH)
-    if (arm_cpu_has_crc32)
-        ft.quick_insert_string = &quick_insert_string_acle;
+    ft.slide_hash = &slide_hash_c;
+    ft.adler32 = &adler32_c;
+    ft.chunksize = &chunksize_c;
+    ft.chunkmemset_safe = &chunkmemset_safe_c;
+    ft.inflate_fast = &inflate_fast_c;
+    ft.adler32_fold_copy = &adler32_fold_copy_c;
+    ft.crc32_fold = &crc32_fold_c;
+    ft.crc32_fold_reset = &crc32_fold_reset_c;
+    ft.crc32_fold_copy = &crc32_fold_copy_c;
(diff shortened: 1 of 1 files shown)


phprus:
I think that grouping by function will reduce the chance of making a error with feature conditions and conditions order.
A grouping of conditions can be make by the compiler.

functable.c
```diff
@@ -56,27 +56,24 @@ static void init_functable(void) {
...
-#endif
-#ifdef X86_SSE2_CHUNKSET
-# if !defined(__x86_64__) && !defined(_M_X64) && !defined(X86_NOCHECK_SSE2)
-    if (x86_cpu_has_sse2)
-# endif
-    {
+#    endif
+#  endif
+#  ifdef X86_SSE2_CHUNKSET
         ft.chunksize = &chunksize_sse2;
```
mtl1979, line 73: `chunksize_sse2` should use `X86_SSE2`, not `X86_SSE2_CHUNKSET`, otherwise the SSE 4.1 variant could possibly use wrong variant of `chunksize`. This also allows leaving out SSE 2 variants when target processor supports SSE 4.1.

Dead2: This PR did not change that, so that does not look like a regression.
But it looks like what you describe would get fixed by #1424.

mtl1979: I know... When that was introduced, I noticed the possible issue... But back then we had no plans to explicitly eliminate out code paths for older instruction sets to reduce code size and dead code. We pretty much relied on compiler doing that implicitly.

functable.c
```diff
@@ -16,310 +16,219 @@ static void init_functable(void) {
 
     cpu_check_features();
 
-    // update_hash_stub:
+    // Generic code
```
nmoinvaz, line 19: I think that the function assignments throughout this function should be in the same order as listed in the struct in functable.h, starting with adler32.. It seems some of them are out of order. Fixing the order in the if statements would be good too.

functable.c
```diff
@@ -16,310 +16,219 @@ static void init_functable(void) {
...
-        ft.chunksize = &chunksize_sse2;
-#endif
-#ifdef X86_AVX_CHUNKSET
-    if (x86_cpu_has_avx2)
-        ft.chunksize = &chunksize_avx;
+    // ARM - NEON
+#if defined(ARM_NEON) && defined(HAVE_BUILTIN_CTZLL)
+    if (arm_cpu_has_neon) {
+        ft.longest_match = &longest_match_neon;
+        ft.longest_match_slow = &longest_match_slow_neon;
```
nmoinvaz, line 155: Are we missing `compare256_neon` here?

functable.c
Dead2: @nmoinvaz Yes, I was thinking about doing that after #1424, just so there is not too much changes at a time.

functable.c
Dead2: @nmoinvaz It sure looks like it, but it was already missing for some reason.. 🤯

nmoinvaz (approved):

iii-i:
I've tested the change on s390, and it turned out that just overriding crc32 is no longer enough, we also need the fold variants for the optimization to be really useful. I will have to look into this. But it has nothing to do with this change, so LGTM.

iii-i (approved):

Dead2:
Rebased on top of develop

iii-i:
Re: crc32 optimization on s390x, that was a false alarm. I ran gtest_zlib with a breakpoint on crc32_le_vgfm_16, and it hit just a couple times. However, this is just because the testsuite don't compress large buffers with gzip too often. Decompressing a large file with minigzip causes this function to be called much more frequently, and the performance gain is still there. So everything is fine and sorry for the noise.

PR #1424, opened 2023-02-06
Further simplify functable and arch-related defines

Dead2:
Based on #1423

- Reduce the amount of different defines required for arch-specific optimizations.
- Also removed a reference to an non-existant adler32_sse41 in test/test_adler32.cc.
- Fix CMake not enabling POWER9_FEATURES, resulting in functable enabling only basic C functions.

Last commit sorts functable contents alphabetically


Dead2:
A further cleanup can be done to alphabetize within blocks, this will improve readability, but I figured it would better to avoid doing that in a PR that also does other things as well.

Dead2:
Rebased on top of Functable Cleanup PR again

Dead2:
Rebased on top of develop

nmoinvaz (approved):
Looks ok to me. Good cleanup.

functable.c
```diff
@@ -148,80 +133,60 @@ static void init_functable(void) {
...
     // Power - VMX
-#ifdef PPC_VMX_SLIDEHASH
-    if (power_cpu_has_altivec)
-        ft.slide_hash = &slide_hash_vmx;
-#endif
-#ifdef PPC_VMX_ADLER32
+#ifdef PPC_VMX
     if (power_cpu_has_altivec)
         ft.adler32 = &adler32_vmx;
+        ft.slide_hash = &slide_hash_vmx;
```
mtl1979: Missing {}...

functable.c
```diff
@@ -148,80 +133,60 @@ static void init_functable(void) {
...
-#endif
-#ifdef POWER8_VSX_CHUNKSET
-    if (power_cpu_has_arch_2_07) {
-        ft.chunksize = &chunksize_power8;
         ft.chunkmemset_safe = &chunkmemset_safe_power8;
+        ft.chunksize = &chunksize_power8;
+        ft.crc32 = &crc32_power8;
         ft.inflate_fast = &inflate_fast_power8;
-    }
+        ft.slide_hash = &slide_hash_power8;
```
mtl1979: Missing {}.

mtl1979 (changes requested):
Fix if statements missing {} around multi-line blocks.

functable.c
Dead2: Fixed

functable.c
Dead2: Fixed

iii-i (approved):
LGTM with respect to s390x.

functable.c
```diff
@@ -148,80 +133,62 @@ static void init_functable(void) {
...
+        ft.chunksize = &chunksize_neon;
         ft.inflate_fast = &inflate_fast_neon;
-    }
+        ft.slide_hash = &slide_hash_neon;
+#  ifdef HAVE_BUILTIN_CTZLL
+        ft.compare256 = &compare256_neon;
+        ft.longest_match = &longest_match_neon;
+        ft.longest_match_slow = &longest_match_slow_neon;
+#  endif
+}
```
phprus: Missing indent.

mtl1979: Yup... Should be at same indentation level as line 140.

functable.c
```diff
@@ -249,103 +216,103 @@ static void init_functable(void) {
...
-static void insert_string_stub(deflate_state* const s, uint32_t str, uint32_t count) {
+static uint32_t adler32_fold_copy_stub(uint32_t adler, uint8_t* dst, const uint8_t* src, size_t len) {
     init_functable();
-    functable.insert_string(s, str, count);
+    return functable.adler32_fold_copy(adler, dst, src, len);
 }
 
-static Pos quick_insert_string_stub(deflate_state* const s, const uint32_t str) {
+static uint8_t* chunkmemset_safe_stub(uint8_t* out, unsigned dist, unsigned len, unsigned left) {
     init_functable();
```
mtl1979, line 233: These Codecov warnings were direct result of moving the functable initialization to single function instead of initializing each functable member in own stub function. The end result is that we have (stub) functions that are never executed by test suite so can possibly (in future) contain bad or dead code.

functable.c
Dead2: Fixed

Dead2:
Fixed erronus partial removal of `POWER8_VSX_CRC32`.

Can this actually not be used on 32bit PPC with VSX? That is what the Cmake/configure tests indicate, so I have kept it like it was.

KungFuJesus:
> Fixed erronus partial removal of `POWER8_VSX_CRC32`.
> 
> Can this actually not be used on 32bit PPC with VSX? That is what the Cmake/configure tests indicate, so I have kept it like it was.

I'd wager that Venn diagram has null intersection given how new VSX is by comparison to the 32 bit version of the ISA.

Dead2:
> I'd wager that Venn diagram has null intersection given how new VSX is by comparison to the 32 bit version of the ISA.

I suspect so as well, I just didn't see any comments about that in the code.

mtl1979:
My 32-bit PowerPC doesn't even have VSX, it only has VMX... So I can't review or test any code that depend on VSX instructions.

PR #1425, opened 2023-02-06
Add official Ubuntu mirror list as fallback for Actions caching proxy

Dead2:
Github Actions caching proxy is repeatedly having problems providing files, sometimes
due to the caching proxy, sometimes due to the upstream server being overloaded.
Lets have a fallback, as re-trying the failing runs is a huge waste of time both for us and for Actions itself.

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -8,6 +8,13 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v3
 
+    - name: Add ubuntu mirrors
+      run: |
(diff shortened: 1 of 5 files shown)


nmoinvaz:
There appear to be some CI failures on runs that use an older version of Ubuntu. I think you can use `lsb_release --codename` to get the codename of Ubuntu.

Dead2:
Fixed, just had a slash the wrong way 🤦‍♂️

Dead2:
All CI passes, this will get merged tomorrow morning unless someone objects.

PR #1426, opened 2023-02-06
Correct inflate_fast function signature

pps83:


nmoinvaz:
That was done on purpose to prevent having to include zlib.h in functable.h. See the CI build failures for the error. Perhaps you know a better way?

pps83:
I did check that, and I think `zbuild.h` is needed, which is included by deflate.h -> zutil.h -> zbuild.h (as PREFIX3 is defined by zbuild.h). Is there any other header that this depends on?

pps83:
Ah, ok, I see that, `stream` itself needs to be fwd declared. I'll update the PR.

pps83:
image

@nmoinvaz I can update these fwd declarations `#ifdef ZLIB_COMPAT`

nmoinvaz:
That should work.

nmoinvaz:
I wonder if we can do the same with deflate_struct..

pps83:
> deflate_struct

sure we can:

`typedef struct internal_state deflate_state;` 

then, all three includes could be removed, only zbuild.h and stdint would be needed.
but that would be kind of unrelated to the PR, as this one attempts to fix invalid signature.

nmoinvaz:
> but that would be kind of unrelated to the PR

Yes you are right, I gather it might end up being more involved.

nmoinvaz:
Build failures again.
```c
/home/actions-runner/_work/zlib-ng/zlib-ng/zbuild.h:85:22: error: unknown type name 'z_stream'
   85 | #  define PREFIX3(x) z_ ## x
      |                      ^~
/home/actions-runner/_work/zlib-ng/zlib-ng/cpu_features.h:98:28: note: in expansion of macro 'PREFIX3'
   98 | extern void inflate_fast_c(PREFIX3(stream) *strm, uint32_t start);
```
I think you will need to reference either `zng_stream` or `z_stream` instead of `PREFIX3(stream)...`.

It might be better as you say to use `#ifdef ZLIB_COMPAT` in both places and just hard code `zng_stream` or `z_stream` in inflate_fast declaration.

pps83:
similar update is needed in cpu_features.h, I'll update

nmoinvaz (approved):
LGTM

pps83:
@Dead2 ^^

Dead2 (approved):
LGTM

PR #1428, opened 2023-02-07
Update actions package dependencies

Dead2:
This hopefully gets rid of the various deprecation warnings we have been seeing lately.
They were up to 56 warnings just for the cmake workflow.
Possibly faster to install when they are all using the same version of Node? IDK.
Some bugs have also been fixed, as well as some new features that we are not yet using.

Notable features I noticed while reading changelogs:
- VS version is selectable https://github.com/ilammy/msvc-dev-cmd/pull/52
- Release uploader now allows setting release name, etc
- Codecov xcode argument support https://github.com/codecov/codecov-action/pull/699
- Codecov gcov argument support https://github.com/codecov/codecov-action/pull/688

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -12,8 +12,8 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: Ubuntu 18.04 GCC
-            os: ubuntu-18.04
+          - name: Ubuntu 20.04 GCC
+            os: ubuntu-20.04
             compiler: gcc
             cxx-compiler: g++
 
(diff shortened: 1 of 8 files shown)


nmoinvaz:
It is nice to know about the codecov gcov args. Right now we are calling gcovr ourselves and creating the report before sending it to codecov.

Dead2:
> It is nice to know about the codecov gcov args. Right now we are calling gcovr ourselves and creating the report before sending it to codecov.

Yeah, that was my first thought too. I didn't want to go down _another_ rabbit hole right now, but if someone wants to have a look at those items now or later, feel free to.

Codecov uploader has behaved a lot better after you changed the workflow to do gcovr separately it seems, at least I don't tear my hair out over codecov not working all the time anymore. That extra step takes about 3-4s on each run (when GHA is not super-slow like yesterday after the downtime), so I figure about 2, possibly 3 sec of that could be saved if it was done in codecov directly (thus avoiding the extra pip install). Due to parallellism that would finish CI about 10-15sec faster (twice that for my PRs).
So it is not a big priority, more of a nice-have because of the simplification and the little time saved, but only IF it actually works right.

Dead2:
All CI passes, this will get merged tomorrow morning unless someone objects.

Issue #1427, opened 2023-02-06
Build failure with recent lld


aaronmondal:
Upstream lld flipped `--undefined-version`(https://reviews.llvm.org/D135402). This causes recent Clang toolchains to break:

Default config:
```bash
ld.lld: error: version script assignment of 'ZLIB_NG_2.0.0' to symbol 'zng_adler32_c' failed: symbol not defined
ld.lld: error: version script assignment of 'ZLIB_NG_2.0.0' to symbol 'zng_zlibng_string' failed: symbol not defined
ld.lld: error: version script assignment of 'ZLIB_NG_2.0.0' to symbol 'zng_version' failed: symbol not defined
ld.lld: error: version script assignment of 'local' to symbol 'zng_inflate_fast' failed: symbol not defined
ld.lld: error: version script assignment of 'local' to symbol 'zng_zcalloc' failed: symbol not defined
ld.lld: error: version script assignment of 'local' to symbol 'zng_zcfree' failed: symbol not defined
ld.lld: error: version script assignment of 'local' to symbol 'zng_gz_error' failed: symbol not defined
ld.lld: error: version script assignment of 'ZLIB_NG_GZ_2.0.0' to symbol 'zng_gzoffset64' failed: symbol not defined
ld.lld: error: version script assignment of 'ZLIB_NG_GZ_2.0.0' to symbol 'zng_gzopen64' failed: symbol not defined
ld.lld: error: version script assignment of 'ZLIB_NG_GZ_2.0.0' to symbol 'zng_gzseek64' failed: symbol not defined
ld.lld: error: version script assignment of 'ZLIB_NG_GZ_2.0.0' to symbol 'zng_gztell64' failed: symbol not defined
```

`ZLIB_COMPAT` config:
```bash
ld.lld: error: version script assignment of 'local' to symbol 'inflate_fast' failed: symbol not defined
ld.lld: error: version script assignment of 'local' to symbol 'inflate_table' failed: symbol not defined
ld.lld: error: version script assignment of 'local' to symbol 'zcalloc' failed: symbol not defined
ld.lld: error: version script assignment of 'local' to symbol 'zcfree' failed: symbol not defined
```

mtl1979:
At least `zng_adler32_c` should not be exported anymore...
`zng_zlibng_string` should be `zlibng_string` and `zng_version` should be `zlibng_version`.
`zng_inflate_fast` and `inflate_fast` might need workaround as those are in functable now.
[...]

PR #1421, opened 2023-02-05
Fix several cases of wrongly exported functions and errors in the version scripts.

Dead2:
- Add missing Z_INTERNAL to some functions that should not be exported.
- Introduced PREFIX5() macro.
- Fixed calloc/cfree prefixing.
- Fixed several errors in the version scripts.
- Fixed several problems with name mangling.

Resolves #1427

PR #1430, opened 2023-02-08
Added CI test for checking undefined symbols in version scripts.

nmoinvaz:
See #1427 and #1421.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -295,6 +295,24 @@ jobs:
             gcov-exec: llvm-cov-11 gcov
             codecov: ubuntu_clang
 
+          # Check for undefined symbols in the version script for the modern api
+          - name: Ubuntu Clang Undefined Symbols
+            os: ubuntu-latest
+            compiler: clang-11
+            cxx-compiler: clang++-11
+            cmake-args: -DCMAKE_SHARED_LINKER_FLAGS="-fuse-ld=lld -Wl,--no-undefined-version" -DZLIBNG_ENABLE_TESTS=OFF
+            build-shared: ON
+            packages: clang-11 llvm-11 lld
+
+          # Check for undefined symbols in the version script for the compat api
+          - name: Ubuntu Clang Undefined Symbols Compat
+            os: ubuntu-latest
+            compiler: clang-11
+            cxx-compiler: clang++-11
+            cmake-args: -DCMAKE_SHARED_LINKER_FLAGS="-fuse-ld=lld -Wl,--no-undefined-version" -DZLIBNG_ENABLE_TESTS=OFF -DZLIB_COMPAT=ON
+            build-shared: ON
+            packages: clang-11 llvm-11 lld
+
           - name: Ubuntu Clang Inflate Strict
             os: ubuntu-latest
             compiler: clang-11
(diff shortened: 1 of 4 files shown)


nmoinvaz:
Quick link to failure job: https://github.com/zlib-ng/zlib-ng/actions/runs/4120456027/jobs/7115205871

mtl1979:
Do we want one for `ZLIB_COMPAT` too?

nmoinvaz:
Good idea, I have added it.

mtl1979:
I didn't expect the `ZLIB_COMPAT`run to fail with `/usr/bin/ld: cannot find -lzlibstatic: No such file or directory`... This kind of revealed another bug...

nmoinvaz:
It looks like it was introduced in https://github.com/zlib-ng/zlib-ng/pull/1352/files @iii-i

mtl1979:
> It looks like it was introduced in https://github.com/zlib-ng/zlib-ng/pull/1352/files @iii-i

Yup... It looks like `ZLIBNG_ENABLE_TESTS` is never set to `OFF` by `CMakeLists.txt` files.

nmoinvaz:
The `gtest_zlib` doesn't support `BUILD_SHARED_LIBS=ON` because it needs access to internal zlib functions for testing.

mtl1979:
> The `gtest_zlib` doesn't support `BUILD_SHARED_LIBS=ON` because it needs access to internal zlib functions for testing.

Some tests will be disabled if both `ZLIB_COMPAT` and `BUILD_SHARED_LIBS` are set...

mtl1979:
There is a typo in one of the commit messages, `intenral` should be `internal`.

mtl1979:
I had to use `-DCMAKE_SHARED_LINKER_FLAGS="-fuse-ld=lld -Wl,--no-undefined-version"` instead of setting `CMAKE_LINKER` to get the expected failure... This also meant installing `lld` as my Ubuntu didn't have it installed.

nmoinvaz:
It already shows expected failure.

mtl1979:
> It already shows expected failure.

Not for the `ZLIB_COMPAT` case.

Dead2:
> I had to use `-DCMAKE_SHARED_LINKER_FLAGS="-fuse-ld=lld -Wl,--no-undefined-version"` instead of setting `CMAKE_LINKER` to get the expected failure... This also meant installing `lld` as my Ubuntu didn't have it installed.

I agree, this is the right way to do it, mainly due to CMake running the linker stage through the compiler instead of the linker even when you specify a linker executable..
I thought about this after I went to bed last night and regretted not re-testing and recommending this way instead (I originally tested this way first, but it failed without any good message, so I changed things before noticing the culprit; lld was not installed along with clang as I had expected)

Not sure why the compat run didn't show the expected failures, and I see nothing wrong in the CI config or the logs, so possibly this is why. CMake probably ran clang to link, and clang might have chosen to use ld instead of lld. Although that would depend on the ld used having already been patched to ignore and not error out on that parameter, because on my 1.5 year old distro it will error out. If that is not why, then I am at a loss. Hard to know when the build doesn't use a verbose makefile so it shows compiler and linker flags =/

mtl1979:
> Not sure why the compat run didn't show the expected failures, and I see nothing wrong in the CI config or the logs, so possibly this is why. CMake probably ran clang to link, and clang might have chosen to use ld instead of lld. Although that would depend on the ld used having already been patched to ignore and not error out on that parameter, because on my 1.5 year old distro it will error out. If that is not why, then I am at a loss. Hard to know when the build doesn't use a verbose makefile so it shows compiler and linker flags =/

I think GNU linker understands the flag, but doesn't check that `local:` symbols actually exist.

nmoinvaz:
Changed to use `-fuse-ld=lld`..

nmoinvaz:
Looks like it is working now. It is much better, showing more errors. Whatever linker it was using before probably wasn't as thorough.

Dead2 (approved):
LGTM

mtl1979 (approved):
I see the expected result in both new CI runs.

PR #1393, opened 2022-12-26
Stable release 2.0.7

Dead2:
Changes since 2.0.6:
- Fix CVE-2022-37434 #1328
- Fix chunkmemset #1196
- Fix deflateBound too small #1236
- Fix Z_SOLO #1263
- Fix ACLE variant of crc32 #1274
- Fix inflateBack #1311
- Fix deflate_quick windowsize #1431 
- Fix DFLTCC bugs related to adler32 #1349 and #1390
- Fix warnings #1194 #1312 #1362
- MacOS build fix #1198
- Add invalid windowBits handling #1293
- Support for Force TZCNT #1186
- Support for aligned_alloc() #1360
- Minideflate improvements #1175 #1238
- Dont use unaligned access for memcpy #1309
- Build system #1209 #1233 #1267 #1273 #1278 #1292 #1316 #1318 #1365
- Test improvements #1208 #1227 #1241 #1353
- Cleanup #1266
- Documentation #1205 #1359
- Misc improvements #1294 #1297 #1306 #1344 #1348
- Backported zlib fixes
- Backported CI workflows from Develop branch

PR #1431, opened 2023-02-08
Remove windowbits change for level 1

### LINKED ISSUE #1417: wbits parameter ignored for compression level 1 on 2.0.x branch
Hi, I maintain python bindings for zlib-ng (https://github.com/pycompression/python-zlib-ng). I run a lot of compatibility tests with Python's zlib module to ensure smooth interoperability.

It turns out that using zng_deflate with compression level 1 overrides  the wbits setting in deflateInit in some way. This results in files that are always compressed with wbits 13 it seems. When I use wbits=-9 , I should also be able to inflate again with wbits=-9, however this is not possible and will throw an invalid window size error. 

This problem does not occur on the develop branch.

rhpvorderman:
This fixes the current 2.0.x branch in this regard. It is a single commit, so it can be cherry-picked for https://github.com/zlib-ng/zlib-ng/pull/1393

Fixes #1417

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -299,11 +299,6 @@ int32_t Z_EXPORT PREFIX(deflateInit2_)(PREFIX3(stream) *strm, int32_t level, int
     if (windowBits == 8)
         windowBits = 9;  /* until 256-byte window bug fixed */
 
-#if !defined(NO_QUICK_STRATEGY) && !defined(S390_DFLTCC_DEFLATE)
-    if (level == 1)
-        windowBits = 13;
-#endif
-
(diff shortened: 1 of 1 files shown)


nmoinvaz:
I think that whole if statement can go bye-bye. Looking back, I think it was commit 951764a1 that allowed any size of window with deflate quick algorithm.

rhpvorderman:
@nmoinvaz 
Done. I amended the commit. Either this commit or https://github.com/zlib-ng/zlib-ng/commit/ce6789c7e093e8e6bb6fc591bbdf0f805999bdb9 can be merged into the latest 2.0.x release to fix the problem.

Dead2 (approved):
LGTM

PR #1432, opened 2023-02-09
Adjust thread counts for compiles and tests

Dead2:
Adjust thread counts for compiles and tests to avoid under-utilization and congestion.
The free Github Actions VMs have 2 cores, the dedicated s390x VM has 4 cores.

This will hopefully shave off a little of the wait time for each PR to complete its CI runs.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -231,6 +231,8 @@ jobs:
             packages: qemu qemu-user gcc-s390x-linux-gnu g++-s390x-linux-gnu libc-dev-s390x-cross
             ldflags: -static
             codecov: ubuntu_gcc_s390x
+            # The dedicated test VM has 4 cores
+            parallels-jobs: 4
 
           - name: Ubuntu GCC S390X No vectorized CRC32 ASAN
             os: ubuntu-latest
@@ -239,6 +241,8 @@ jobs:
             packages: qemu qemu-user gcc-s390x-linux-gnu g++-s390x-linux-gnu libc-dev-s390x-cross
             ldflags: -static
             codecov: ubuntu_gcc_s390x_no_crc32
+            # The dedicated test VM has 4 cores
+            parallels-jobs: 4
 
           - name: Ubuntu GCC S390X DFLTCC ASAN
             os: z15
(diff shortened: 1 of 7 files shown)


.github/workflows/cmake.yml
```diff
@@ -524,7 +534,7 @@ jobs:
           -DLLVM_ENABLE_PROJECTS="libcxx;libcxxabi" \
           -DLLVM_USE_SANITIZER=MemoryWithOrigins \
           -DLLVM_LIBC_ENABLE_LINTING=OFF
-        cmake --build llvm-project/build -- cxx cxxabi
+        cmake --build llvm-project/build -j2 -- cxx cxxabi
```
nmoinvaz, line 537: Do we want `-j ${{ matrix.parallels-jobs || '2' }}` here?

.github/workflows/cmake.yml
```diff
@@ -548,12 +558,12 @@ jobs:
         CI: true
 
     - name: Compile source code
-      run: cmake --build ${{ matrix.build-dir || '.' }} --config ${{ matrix.build-config || 'Release' }}
+      run: cmake --build ${{ matrix.build-dir || '.' }} -j2 --config ${{ matrix.build-config || 'Release' }}
```
nmoinvaz, line 561: Do we want `-j ${{ matrix.parallels-jobs || '2' }}` here?

.github/workflows/cmake.yml
Dead2: I did consider that, but to be really useful it would have to be split into two settings, one for compile and one for tests. (There seems to be more to gain by overloading during testing, likely because of more io-wait and context switching and less cpu-load)

But I figured we might not want to do more than 2 compiles at a time in any case, since it has a tendency to make the build log harder to read when there is an error somewhere. Also, there is not much to be gained from going above 2.
The GHA 2-core VMs would possibly get a very slight boost at 3, if there is any io-wait time in the other threads, but likely those two cores are already contending with other VMs for cpu-time.
The s390x 4-core VMs would probably get much faster at 4-5 threads, but only during the already quite fast compile, and the s390x runs already complete long before any of the other workflows.

So I figured it was not really worth the effort to make it configurable just to save a little bit of time where it doesn't really help us the first place. I am not opposed to a PR making it configurable though.

Dead2:
Comparing compile time, there is some variance, but it seems most runs are about 20-30sec faster after this change.
When it comes to tests, the variance is much bigger, sometimes a full minute faster, sometimes 20sec slower. But it seems to average out to being the same or slightly faster. This is not a big surprise since running 6 parallel threads on 2 cores is going to make for a lot of contention, and reducing that to 3 threads is only freeing up the wasted time taken by doing all the extra context switching. The spurious big slowdowns are probably related to the VM contending with another VM for the use of the actual cpu cores, as there is nothing else here that would cause that.

PR #1434, opened 2023-02-10
[minigzip] Close gzFile before exiting on error.

mtl1979:
ASAN will complain if gzFile is not closed before program exits.


Dead2 (approved):
LGTM

Dead2:
I think a PR for the same change on develop would be a good idea.
Not sure why it didn't already trigger the same error there though.

mtl1979:
> I think a PR for the same change on develop would be a good idea. Not sure why it didn't already trigger the same error there though.

It took me a while to be able to reproduce the issue locally... I had to run each test individually instead of through `ctest`.

Dead2:
I cherry-picked this since it would have otherwise needed a rebase.

PR #1435, opened 2023-02-10
Narrow down the amount of packages installed for Wine in MinGW CI runs.

Dead2:
Before:
```
# x86
12 upgraded, 265 newly installed, 0 to remove and 40 not upgraded.
Need to get 354 MB of archives.
After this operation, 1626 MB of additional disk space will be used.

# x86-64
12 upgraded, 296 newly installed, 0 to remove and 40 not upgraded.
Need to get 545 MB of archives.
After this operation, 2703 MB of additional disk space will be used.
```

After:
```
# x86
2 upgraded, 167 newly installed, 0 to remove and 28 not upgraded.
Need to get 254 MB of archives.
After this operation, 1239 MB of additional disk space will be used.

#x86-64
0 upgraded, 48 newly installed, 0 to remove and 30 not upgraded.
Need to get 318 MB of archives.
After this operation, 1710 MB of additional disk space will be used.
```

This should shave off some of the time spent installing packages, but the variation between runs is too great to see how much. This CI has so far both of the fastest wine install runs I have seen. (Faster by 12sec, 1min, 10min, etc.)

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -276,7 +276,7 @@ jobs:
           - name: Ubuntu MinGW i686
             os: ubuntu-22.04
             cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-mingw-i686.cmake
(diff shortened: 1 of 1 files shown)


nmoinvaz (approved):

mtl1979:
As long as the versions for 32-bit packages and 64-bit packages are forced to be same, I don't see any reason why we can't change what other packages get installed... CI runs should fail quite fast if something needed is changed or removed.

PR #1333, opened 2022-08-22
Inflate: Unconditional refill and fastpath for literals

dougallj:
This is the second of a series of optimisation changes [described on my blog](https://dougallj.wordpress.com/2022/08/20/faster-zlib-deflate-decompression-on-the-apple-m1-and-x86/).

PR #1438, opened 2023-02-12
Use named defines instead of hard coded numbers.

nmoinvaz:


Dead2:
`MAX_BITS` and `MAX_WBITS` are pretty similar looking. Should we change naming to something with a more clear distinction, or is it not needed? Personally I had to do a double-take when reviewing inflate.c, as I misread at first.
`MAX_BITS_WIN` would be more clear, but unfortunate since that would actually mean renaming an existing define.
In any case, this is not a deal-breaker for me, but something I think should be considered before merging.

Other than that I love it, I always hated magic numbers that might or might not mean what you think they do.

nmoinvaz:
Yeah they are very similar; I think the names are okay. I didn't want to deviate too much from madler/zlib named defines. This PR is in preparation for PR #1333.

inffast_tpl.h
```diff
@@ -184,12 +184,13 @@ void Z_INTERNAL INFLATE_FAST(PREFIX3(stream) *strm, uint32_t start) {
...
             if (op & 16) {                      /* distance base */
                 dist = here->val;
-                op &= 15;                       /* number of extra bits */
+                op &= MAX_BITS;                 /* number of extra bits */
                 if (bits < op) {
                     hold |= load_64_bits(in, bits);
                     in += 6;
                     bits += 48;
                 }
+                op &= MAX_BITS;                 /* number of extra bits */
```
mtl1979: There is no explanation why this line was added, and we generally discourage mixing cleanup and code changes.

mtl1979:
I'm generally in favor of merging defines that are used for same purpose in different files and moving them to place where the preprocessor define can be included from in all needed files. I've already said I'm not big fan/supporter of magic numbers and prefer preprocessor defines as those can be documented better and improve readability and reduce confusion if same magic number is used for different purposes.

inffast_tpl.h
nmoinvaz: Good catch, thank you. Fixed.

Dead2 (approved):

PR #1445, opened 2023-02-23
Replace __WIN__ with X86_FEATURES

phprus:
The ``__WIN__`` macro is not standard.

### DIFF
--- a/zbuild.h
+++ b/zbuild.h
@@ -187,7 +187,7 @@
 #  define PREFETCH_L1(addr)     __builtin_prefetch(addr, 0, 3)
 #  define PREFETCH_L2(addr)     __builtin_prefetch(addr, 0, 2)
 #  define PREFETCH_RW(addr)     __builtin_prefetch(addr, 1, 2)
-#elif defined(__WIN__)
+#elif defined(X86_FEATURES)
 #  include <xmmintrin.h>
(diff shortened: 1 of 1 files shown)


nmoinvaz (approved):

Dead2:
This feels like it could potentially bite us later on..
Would it not be better to change the detection here to check for msvc compiler?

Edit: Also, what about ARM-based windows OS? Or perhaps those are already not covered by the `__WIN__` macro..

phprus:
@Dead2 
``__WIN__`` - is undefined macro in all environments.
``_mm_prefetch`` is an intrinsic for all x86 with SSE, not only for MSVC.

Therefore, it is enough to check ``X86_FEATURES`` (or ``X86_FEATURES and X86_SSE2`` if we support x86 without SSE).

Dead2:
> @Dead2 `__WIN__` - is undefined macro in all environments. `_mm_prefetch` is an intrinsic for all x86 with SSE, not only for MSVC.
> 
> Therefore, it is enough to check `X86_FEATURES` (or `X86_FEATURES and X86_SSE2` if we support x86 without SSE).

Thanks. Some of that reasoning could have been in the PR description 👍

Dead2 (approved):

PR #1448, opened 2023-03-07
Add parametrized testing

rhpvorderman:
More tests is better. This simply tests if compressing a small data buffer with all possible combinations of settings is successful.

The test is a good addition, because it already finds one bug which I will make an issue for.  EDIT: #1449 

The C code works properly, I wonder how to best integrate this into the test suite.


nmoinvaz:
We do something similar already in CMake by testing common compression levels and known failure cases. But this looks like it may cover more cases.

It would be best if this was incorporated into Google Test framework. 
See **Value-Parameterized Tests**:
https://github.com/google/googletest/blob/main/docs/advanced.md

Dead2:
@rhpvorderman This needs a rebase on develop, and it would also be nice if the tests were integrated into the existing test framework. Are these things you think you could do to get this merged?

Dead2:
Replaced by #2207

PR #1450, opened 2023-03-09
Shared libraries on cygwin can be named with their version number

fd00:
Shared libraries on cygwin can be named with their version number.

See https://github.com/Kitware/CMake/blob/v3.25.3/Modules/Platform/CYGWIN.cmake#L61

This fix will build `cygz-ng-2.dll`.


Dead2:
I am unable to review this myself, as I have near zero experience with cygwin.

nmoinvaz:
I don't have any experience with it either.

nmoinvaz:
According to the linked cmake file in the description this change should be ok.

nmoinvaz (approved):

Dead2 (approved):

PR #1451, opened 2023-03-18
Switch to the standard CMake module GNUInstallDirs

phprus:
**Benefits:**
1) Relative install() paths by default. Simplify integration with CPack.
2) The default paths correspond the GNU and FHS layouts: ``lib`` or ``lib64`` or ``lib/<arch>`` (debian).
3) Linux distributions do not need to override ``INSTALL_LIB_DIR`` (like this: https://src.fedoraproject.org/rpms/zlib-ng/blob/f38/f/zlib-ng.spec#_33). The current CMake build system hardcodes ``lib``.

**Proposal:**
Remove legacy and unneeded variables:

- ``LIB_INSTALL_DIR``, ``INSTALL_LIB_DIR``
- ``INC_INSTALL_DIR``, ``INSTALL_INC_DIR``


Dead2:
Have not tested this yet, but it looks like a good approach.

nmoinvaz:
A previous discussion related GNUInstallDirs can be found here at PR #881.

Dead2:
I'd like to keep the support for the variables you propose to remove. Some of those are used by distro macros, and `LIB_INSTALL_DIR` was actually added to support the fedora rpm `%cmake` macro after we noticed their use of the workaround. As far as I can see, those are still supported with this PR.

This PR looks like good cleanups, and improvements to things like relative paths, .pc  support and default libdir selection.

But I notice that you have removed the `CACHE PATH` parameter (and their descriptions) for the "legacy variables", is there any actual reasoning for that? I could understand removing the `FORCE` parameter, if that was the intention here.

Dead2:
To be clear, I'd like to see the `CACHE PATH` stuff put back in, unless there is a good reason not to. Otherwise the cache ends up like this:
```
//No help, variable specified on the command line.
LIB_INSTALL_DIR:UNINITIALIZED=/usr/local/lib64
```

phprus:
@Dead2 

> I'd like to keep the support for the variables you propose to remove. Some of those are used by distro macros, and `LIB_INSTALL_DIR` was actually added to support the fedora rpm `%cmake` macro after we noticed their use of the workaround. As far as I can see, those are still supported with this PR.

In the current version of PR, these variables are fully supported.

> But I notice that you have removed the CACHE PATH parameter (and their descriptions) for the "legacy variables"

``CACHE PATH`` and description can be set only if we set the value of deprecated variables. But now all deprecated variables are used as input only.
If these variables are not set by the user, they are not used or set anywhere in cmake.

Dead2:
> `CACHE PATH` and description can be set only if we set the value of deprecated variables. But now all deprecated variables are used as input only. If these variables are not set by the user, they are not used or set anywhere in cmake.

I know, but if you do use them, they still get cached, and I see no reason to not do it properly when they are used.

phprus:
@Dead2 
Done.

Dead2 (approved):

PR #1463, opened 2023-03-30
Remove architecture dependent endian header locations

lawadr:
The header locations are OS specific and not architecture specific. The previous behaviour was to always include machine/endian.h for ARM and AArch64 architectures on non-Windows and non-Linux OSs, causing build failures if the OS uses other locations defined further down the conditional block.


nmoinvaz (approved):

Dead2 (approved):
LGTM

PR #1465, opened 2023-04-04
Fix CMake check for posix_memalign and aligned_alloc

lawadr:
These two functions were being checked using check_function_exists. This CMake macro does not check to see if the given function is declared in any header as it declares its own function prototype and relies on linking to determine function availability. This causes two issues.

Firstly, it will always succeed when the CMake toolchain file sets CMAKE_TRY_COMPILE_TARGET_TYPE to STATIC_LIBRARY as no linking will take place. See: https://gitlab.kitware.com/cmake/cmake/-/issues/18121

Secondly, it will not correctly detect macros or inline functions, or whether the function is even declared in a header at all.

Switch to check_symbol_exists at CMake's recommendation, the logic of which actually matches the same checks in the configure script.


Dead2 (approved):
LGTM

PR #1472, opened 2023-04-19
IBM zSystems: Fix calling deflateBound() before deflateInit()

iii-i:
Even though zlib officialy forbids calling deflateBound() before
deflateInit(), Firefox does this anyway, and it happens to work [1],
but unfortunately not with DFLTCC [2], because the DFLTCC code assumes
that the deflate state is allocated, and segfaults when it isn't.

Bow down before Hyrum's Law and add deflateStateCheck() to
DEFLATE_BOUND_ADJUST_COMPLEN().

Extend the deflate_bound test accordingly.

[1] https://searchfox.org/mozilla-esr102/source/dom/script/ScriptCompression.cpp#97
[2] https://bugzilla.suse.com/show_bug.cgi?id=1210593


Dead2 (approved):

Dead2:
Backported to stable branch

PR #1476, opened 2023-04-25
Fix Coverity warnings

mtl1979:
Fix defects detected by coverity scan.

Personally I think these are corner cases that happen only if the match is very long.


mtl1979:
Rebased.

Dead2:
In absence of reviews, cherry-picked two of the commits

mtl1979:
Rebased.

PR #1477, opened 2023-04-27
Enable RISC-V cross build and test

alexsifivetw:
Add supporting RISC-V cross compilation workflows.
Add riscv cpu info detection.


cmake/toolchain-riscv.cmake
```diff
@@ -0,0 +1,64 @@
...
+set(CMAKE_SYSTEM_NAME "Linux")
+set(CMAKE_SYSTEM_PROCESSOR "riscv64")
+
+option(USE_RVV "Enable riscv vector or not." ON)
+option(USE_AUTO_VECTORIZER "Enable riscv auto vectorizer or not." OFF)
+
+# Avoid to use system path for cross-compile
+set(CMAKE_FIND_USE_CMAKE_SYSTEM_PATH FALSE)
+
+set(TOOLCHAIN_PATH "" CACHE STRING "The toolcahin path.")
```
mtl1979: Spelling mistake... should be `toolchain`

cmake/toolchain-riscv.cmake
```diff
@@ -0,0 +1,64 @@
...
+if(NOT TOOLCHAIN_PATH)
+  set(TOOLCHAIN_PATH ${CMAKE_SOURCE_DIR}/build-toolchain-qemu/riscv-clang)
+endif()
+
+set(TOOLCHAIN_PREFIX "riscv64-unknown-linux-gnu-" CACHE STRING "The toolcahin prefix.")
```
mtl1979: Spelling mistake.

README.md
```diff
@@ -221,6 +221,7 @@ Advanced Build Options
 | WITH_NEON                       | --without-neon        | Build with NEON intrinsics                                          | ON                     |
 | WITH_ALTIVEC                    | --without-altivec     | Build with AltiVec (VMX) intrinsics                                 | ON                     |
 | WITH_POWER8                     | --without-power8      | Build with POWER8 optimisations                                     | ON                     |
+| WITH_RVV                        |                       | Build with RVV                                                      | ON                     |
```
nmoinvaz: Maybe it should say Build with RISC-V Vector intrinsics"? Something more descriptive since it is part of readme.

tools/riscv/README_RISCV.md
```diff
@@ -0,0 +1,59 @@
...
+  -D CMAKE_TOOLCHAIN_FILE=./cmake/toolchain-riscv.cmake \
+  -D CMAKE_INSTALL_PREFIX=./build-riscv/install \
+  -D TOOLCHAIN_PATH={TOOLCHAIN_PATH} \
+  -D QEMU_PATH={QEMU_PATH} \
+  .
+
+cmake --build ./build-riscv --target install
+```
+
+## Enable RISC-V Extensions at Runtime ##
```
nmoinvaz: We don't have enabling of extensions at runtime for any other architecture. Perhaps this should be done in a different PR for all architectures.

arch/riscv/riscv_features.c
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+    return 0;
+}
+
+static inline int riscv_has_rvv() {
+    const char *riscv_cap = getenv("ZLIBNG_RISCVCAP");
+    if (riscv_cap != NULL) {
+        return check_rvv_support(riscv_cap);
+    } else {
```
nmoinvaz: `else is not needed`

arch/riscv/riscv_features.c
```diff
@@ -0,0 +1,67 @@
...
+        ptrdiff_t diff = underscore - rv64_str;                \
+        if (diff <= 0) {                                       \
+            printf("WARNING: ISA string has wrong format.\n"); \
+            return 0;                                          \
+        }                                                      \
+    }                                                          \
+} while (0)
+
+static inline int check_rvv_support(const char *isa_str) {
+    if (isa_str != NULL) {
```
nmoinvaz: Indentation can be reduced with:
```c
if (isa_str != NULL)
    return 0;
```

.github/workflows/riscv.yml
```diff
@@ -0,0 +1,45 @@
...
+        sudo apt-get install -y ninja-build
+
+    - name: Download source and build
+      run: |
+        git clone https://github.com/sifive/prepare-riscv-toolchain-qemu.git
+        chmod +x prepare-riscv-toolchain-qemu/prepare_riscv_toolchain_qemu.sh
+        export GIT_SSL_NO_VERIFY=1
+        ./prepare-riscv-toolchain-qemu/prepare_riscv_toolchain_qemu.sh
+
+    - name: Configure and build zlib-ng
```
nmoinvaz: Why not include this configuration in `cmake.yaml` instead of having it in its own YAML?

CMakeLists.txt
```diff
@@ -882,6 +887,24 @@ if(WITH_OPTIM)
             add_feature_info(XSAVE 1 "Support XSAVE intrinsics using \"${XSAVEFLAG}\"")
             set_property(SOURCE ${ARCHDIR}/x86_features.c PROPERTY COMPILE_FLAGS "${XSAVEFLAG}")
         endif()
+    elseif(BASEARCH_RISCV_FOUND)
```
nmoinvaz: This `elseif` should go above `elseif` for `BASEARCH_S360_FOUND`, because it is sorted alphabetically.

cmake/toolchain-riscv.cmake
alexsifivetw: Thanks, it's fixed.

cmake/toolchain-riscv.cmake
alexsifivetw: Thanks, it's fixed.

README.md
alexsifivetw: Thanks, it's done.

tools/riscv/README_RISCV.md
alexsifivetw: I apologize for using a misleading name. The feature is designed to get risc-v cpu info, because we can't use hwcap to detect risc-v extensions at runtime in linux. We may find better solution in the future.

arch/riscv/riscv_features.c
alexsifivetw: Thanks, it's done.

arch/riscv/riscv_features.c
alexsifivetw: Thanks, it's done.

CMakeLists.txt
alexsifivetw: Thanks, it's done.

.github/workflows/riscv.yml
alexsifivetw: Thanks, it's done.

tools/riscv/README_RISCV.md
mtl1979: Enabling auto-vectorization in compiler and run-time checking of vector instruction support are kind of mutually exclusive, so I would like to see manually (optimized by hand) vectorized functions along with run-time checking.

Dead2:
This looks good for the most part, but there are two things I am not happy about being introduced.
One of those is the use of `getenv()`, a similar approach was proposed years ago both here and in zlib upstream as well, and was rejected because it opens up a new potential attack vector onto any application that uses zlib or uses a library that uses zlib.
The `fopen()` call is another such, although it is a much smaller risk, but it is still something that the library otherwise has never done.

I really hope it is possible to work around this somehow, although that might need hwcap updates to the kernel if I understand correctly?

An alternative is of course to force enable/disable this during compile, that would move the burden onto packagers instead, they could for example compile for multiple variations and only select the correct one during install. For testing or manual compile one could compile for the correct variation directly.

tools/riscv/README_RISCV.md
alexsifivetw: Sure, the `USE_AUTO_VECTORIZER` option is turned off by default, but users can still enable it, if they know what they are doing.

https://github.com/zlib-ng/zlib-ng/blob/915106745a35a8d888c9a78f996648587687a5f3/cmake/toolchain-riscv.cmake#L6

alexsifivetw: Sure, the `USE_AUTO_VECTORIZER` option is turned off by default, but users can still enable it, if they know what they are doing.

https://github.com/zlib-ng/zlib-ng/blob/915106745a35a8d888c9a78f996648587687a5f3/cmake/toolchain-riscv.cmake#L6

alexsifivetw:
> An alternative is of course to force enable/disable this during compile, that would move the burden onto packagers instead, they could for example compile for multiple variations and only select the correct one during install. For testing or manual compile one could compile for the correct variation directly.

I believe that it's reasonable to determine RVV supporting during compile time, and to implement a more graceful solution when the kernel updates hwcaps or hwprobe for risc-v in the features.

alexsifivetw:
> An alternative is of course to force enable/disable this during compile, that would move the burden onto packagers instead, they could for example compile for multiple variations and only select the correct one during install. For testing or manual compile one could compile for the correct variation directly.

I believe that it's reasonable to determine RVV supporting during compile time, and to implement a more graceful solution when the kernel updates hwcaps or hwprobe for risc-v in the features.

mtl1979:
> An alternative is of course to force enable/disable this during compile, that would move the burden onto packagers instead, they could for example compile for multiple variations and only select the correct one during install. For testing or manual compile one could compile for the correct variation directly.

"Force enable" during configuring or compiling is a lot better compromise than relying on user input being sane or some kernel provided text file keeping predictable formatting.

cmake/toolchain-riscv.cmake
```diff
@@ -0,0 +1,64 @@
...
+  endif()
+else()
+  list(APPEND RISCV_COMPILER_FLAGS "-march=rv64gc")
+endif()
+message(STATUS "RISCV_COMPILER_FLAGS: ${RISCV_COMPILER_FLAGS}")
+
+set(CMAKE_C_FLAGS             "${RISCV_COMPILER_FLAGS} ${CMAKE_C_FLAGS}")
+set(CMAKE_CXX_FLAGS           "${RISCV_COMPILER_FLAGS} ${CMAKE_CXX_FLAGS}")
+
+set(RISCV_LINKER_FLAGS "-lstdc++ -lpthread -lm -ldl")
```
mtl1979: Unconditionally adding `-lstdc++` and `-lpthread` might need a comment to explain the reason... zlib-ng library itself is C code, so shouldn't link against C++ standard library.

Usually used C++ library is specified in `CMAKE_CXX_FLAGS_INIT` using `-stdlib=libc++`or `-stdlib=libstdc++` before setting `CMAKE_CXX_FLAGS` when there is multiple choices.

alexsifivetw: Thanks, we should allow users to choose the options based on their conditions.

alexsifivetw:
> "Force enable" during configuring or compiling is a lot better compromise than relying on user input being sane or some kernel provided text file keeping predictable formatting.

Thanks, it's done.

mtl1979:
There is clearly issue with building the dependencies taking too long. It would be a lot easier if there would be official deb packages for the dependencies instead of needing to rebuild them every single time.

alexsifivetw:
> There is clearly issue with building the dependencies taking too long. It would be a lot easier if there would be official deb packages for the dependencies instead of needing to rebuild them every single time.

Thank you for your feedback. It is indeed a big issue, especially for modern platforms like RISC-V, which has not yet been widely used in the world.

We are working on solving this problem, and I will update the message thread with any progress.

Dead2:
Getting qemu and other packages into the repos is the best solution, but that will take time.
I am not really a Docker fan myself, but in this case I think that a QEMU-RISC-V docker image containing all dependencies (pre-compiled) could be a good solution. Just a suggestion for the RISC-V community to consier, I think it would probably be useful for other projects as well.

mtl1979:
> Getting qemu and other packages into the repos is the best solution, but that will take time. I am not really a Docker fan myself, but in this case I think that a QEMU-RISC-V docker image containing all dependencies (pre-compiled) could be a good solution. Just a suggestion for the RISC-V community to consier, I think it would probably be useful for other projects as well.

Docker is just wrapper around `qemu`, so using Docker to run `qemu` is like using `qemu` to run itself. You can essentially achieve same result by making disk image and boot it using `qemu` directly. 

In special cases it might be enough to compile the binaries once, use `tar` to archive the directory structure and then compress the resulting tarball with `gzip`. Then all is needed to download and decompress the archive in CI instance.

Dead2:
Docker does not use qemu emulation to run applications on their native architecture.
But it looks like it can use qemu emulation to run on non-native arches, I didn't know that, that is interesting but does not really matter in this case.

A docker can be just that, a tar download containing a prebuilt environment and any scripts to set things up correctly.
It can just be a little cleaner and easier than maintaining it all directly in our workflows, when you can download and run an externally maintained docker file, but I'll leave the implementation details to the RISC-V people.

cmake/toolchain-riscv.cmake
```diff
@@ -0,0 +1,52 @@
...
+set(CMAKE_C_COMPILER_AR "${TOOLCHAIN_PATH}/bin/llvm-ar")
+set(CMAKE_CXX_COMPILER_AR "${TOOLCHAIN_PATH}/bin/llvm-ar")
+set(CMAKE_C_COMPILER_RANLIB "${TOOLCHAIN_PATH}/bin/llvm-ranlib")
+set(CMAKE_CXX_COMPILER_RANLIB "${TOOLCHAIN_PATH}/bin/llvm-ranlib")
+set(CMAKE_OBJDUMP "${TOOLCHAIN_PATH}/bin/llvm-objdump")
+set(CMAKE_OBJCOPY "${TOOLCHAIN_PATH}/bin/llvm-objcopy")
+
+# compile options
+set(RISCV_COMPILER_FLAGS)
+# enable rvv
```
nmoinvaz, line 33: This should be checked in CMake?

CMakeLists.txt
```diff
@@ -720,6 +725,24 @@ if(WITH_OPTIM)
                 set(WITH_POWER9 OFF)
             endif()
         endif()
+    elseif(BASEARCH_RISCV_FOUND)
+        if(WITH_RVV)
```
nmoinvaz, line 729: There should be `check_rvv_intrinics` function that checks if the compiler supports `-march=rv64gcv`. And then that flag can be removed from the toolchain file.

nmoinvaz: Perhaps the next line `check_c_source_compiles` should be moved into a `check_rvv_intrinics` with the `CMAKE_REQUIRED_FLAGS = -march=rv64gcv`

tools/riscv/README_RISCV.md
```diff
@@ -0,0 +1,38 @@
...
+  -D CMAKE_TOOLCHAIN_FILE=./cmake/toolchain-riscv.cmake \
+  -D CMAKE_INSTALL_PREFIX=./build-riscv/install \
+  -D TOOLCHAIN_PATH={TOOLCHAIN_PATH} \
+  -D QEMU_PATH={QEMU_PATH} \
+  .
+
+cmake --build ./build-riscv
+```
+
+Disable the option if there is no RVV support:
```
nmoinvaz: CMake should determine if the compiler supports RVV. `WITH_RVV` should only be there to person compiling it, to be able to easily enable/disable the feature.

tools/riscv/README_RISCV.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+After running script, clang & qemu are built in `build-toolchain-qemu/riscv-clang/` & `build-toolchain-qemu/riscv-qemu/`.
+
+`build-toolchain-qemu/riscv-clang/` is your `TOOLCHAIN_PATH`.
+`build-toolchain-qemu/riscv-qemu/bin/qemu-riscv64` is your `QEMU_PATH`.
+
+## Cross-Compile for RISC-V Target ##
+
+```
```
nmoinvaz: Put the word `bash` after the three `

nmoinvaz (commented):
Have a few CMake comments.

alexsifivetw:
> In special cases it might be enough to compile the binaries once, use `tar` to archive the directory structure and then compress the resulting tarball with `gzip`. Then all is needed to download and decompress the archive in CI instance.

Thank you for your suggestions.

I have created an automated release workflow for the prebuilt toolchain. Users can now download the prebuilt toolchain and enjoy using it.

Users can also use [the script](https://github.com/sifive/prepare-riscv-toolchain-qemu/blob/main/prepare_riscv_toolchain_qemu.sh) to download and build the toolchain based on their environment.

Apologies for the Docker plan, we cannot provide the RISC-V Docker in the short term. However, I will continue working on implementing this feature.

nmoinvaz:
Actually you can use GitHub CLI to download GH releases. 

https://cli.github.com/manual/gh_release_download

GH cli is already installed on all GH runners.

tools/riscv/README_RISCV.md
alexsifivetw: Thanks, it's done.

CMakeLists.txt
alexsifivetw: Thanks, it's done. We will use `set_property` to add compiler flags "-march=rv64gcv" to the files for rvv optimized kernel in the related pr.

tools/riscv/README_RISCV.md
alexsifivetw: Because we cannot detect rvv support at runtime, running the rvv code on a no-rvv target is a risk. Users could disable the rvv when the target don't support rvv.

We will have a better solution when the kernel updates `hwcap` or `hwprobe` for riscv.

mtl1979:
The RISC-V toolchain doesn't seem to support coverage...

cmake/detect-intrinsics.cmake
```diff
@@ -347,6 +347,19 @@ macro(check_power8_intrinsics)
     set(CMAKE_REQUIRED_FLAGS)
 endmacro()
 
+macro(check_rvv_intrinsics)
+    # Check whether compiler supports RVV
+    set(CMAKE_REQUIRED_FLAGS "-march=rv64gcv")
```
nmoinvaz: This should be similar to `check_power9_intrinsics` where it would do: 
```cmake
        if(NOT NATIVEFLAG)
            set(RVVFLAG "-march=rv64gcv")
        endif()
    endif()
    # Check if we have what we need for RISCV optimizations
    set(CMAKE_REQUIRED_FLAGS "${RVVFLAG} ${NATIVEFLAG}")
    ...
```

cmake/toolchain-riscv.cmake
```diff
@@ -0,0 +1,48 @@
...
+# CMake will just use the host-side tools for the following tools, so we setup them here.
+set(CMAKE_C_COMPILER_AR "${TOOLCHAIN_PATH}/bin/llvm-ar")
+set(CMAKE_CXX_COMPILER_AR "${TOOLCHAIN_PATH}/bin/llvm-ar")
+set(CMAKE_C_COMPILER_RANLIB "${TOOLCHAIN_PATH}/bin/llvm-ranlib")
+set(CMAKE_CXX_COMPILER_RANLIB "${TOOLCHAIN_PATH}/bin/llvm-ranlib")
+set(CMAKE_OBJDUMP "${TOOLCHAIN_PATH}/bin/llvm-objdump")
+set(CMAKE_OBJCOPY "${TOOLCHAIN_PATH}/bin/llvm-objcopy")
+
+# compile options
+set(RISCV_COMPILER_FLAGS "" CACHE STRING "The riscv compiler flags.")
```
nmoinvaz: Not necessary any more.

cmake/toolchain-riscv.cmake
```diff
@@ -0,0 +1,48 @@
...
+
+# disable auto-vectorizer
+add_compile_options(-fno-vectorize -fno-slp-vectorize)
+
+set(CMAKE_C_FLAGS             "${RISCV_COMPILER_FLAGS} ${CMAKE_C_FLAGS}")
```
nmoinvaz: Not necessary any more.

cmake/toolchain-riscv.cmake
```diff
@@ -0,0 +1,48 @@
...
+set(CMAKE_CXX_FLAGS           "${RISCV_COMPILER_FLAGS} ${CMAKE_CXX_FLAGS}")
+
+# link options
+set(RISCV_LINKER_FLAGS "" CACHE STRING "The riscv specific link options.")
```
nmoinvaz: No need for thse linkers flags either? Let's not do them until they are used...

tools/riscv/README_RISCV.md
nmoinvaz: Perhaps at the top of this readme, below, `Building RISC-V Target with Cmake` it should indicate that there is no support for runtime detection atm.

cmake/detect-intrinsics.cmake
alexsifivetw: Thanks, it's done.

cmake/toolchain-riscv.cmake
alexsifivetw: Thanks, the cmake file was over-designed, let’s remove the unused parts.

cmake/toolchain-riscv.cmake
alexsifivetw: It's done.

cmake/toolchain-riscv.cmake
alexsifivetw: It's done.

alexsifivetw:
> The RISC-V toolchain doesn't seem to support coverage...

Do you mean the codecov report?

mtl1979:
> > The RISC-V toolchain doesn't seem to support coverage...
> 
> Do you mean the codecov report?

I mean generating coverage information during compilation... First step of cmake gives warning that the relevant compiler command-line options are not supported.

tools/riscv/README_RISCV.md
```diff
@@ -0,0 +1,45 @@
+# Building RISC-V Target with Cmake #
```
nmoinvaz: Can you rename `tools/riscv/README_RISCV.md` to `arch/riscv/README.md` similar to s390?

nmoinvaz (approved):
Besides renaming the README, everything else lgtm.

nmoinvaz (approved):

tools/riscv/README_RISCV.md
alexsifivetw: Sure, it's done.

Dead2:
This currently fails CI due to a missing github access token. Strangely the downloader doesn't fall back to just using the default access token created for each CI job. https://github.com/zlib-ng/zlib-ng/actions/runs/4930653399/jobs/8855048977?pr=1477

According to this: https://docs.github.com/en/actions/security-guides/automatic-token-authentication
you should be able to fix this by adding:
```
env:
    GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

alexsifivetw:
> According to this: https://docs.github.com/en/actions/security-guides/automatic-token-authentication you should be able to fix this by adding:
> 
> ```
> env:
>     GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
> ```
Thanks, it's done.

Dead2 (approved):
LGTM

PR #1482, opened 2023-05-02
Remove `CMAKE_MACOSX_RPATH`

haampie:
This is ON by default since CMake 3.0


mtl1979:
Should verify with CI, but we deprecated old cmake versions 5 years ago...

nmoinvaz (approved):

mtl1979 (approved):

PR #1487, opened 2023-05-12
Fix pkgcheck and AVX512-VNNI compilation flags

Dead2:
- Fixes for pkgcheck CI jobs.
- Fix AVX512-VNNI compile flags
  CMake required `--mtune=cascadelake`, thus disqualifying some older compilers from compiling with AVX512-VNNI.
  Configure did not specify any mtune flags, likely leading to less optimized AVX512-VNNI builds.
  Both now specify `--mtune=cascadelake` if supported by compiler, otherwise falling back to `--mtune=skylake-avx512`.


Dead2:
@mtl1979 Said he had found some differences in compile flags causing the remaining package check failures.

mtl1979:
This is how I fixed it:
```
diff --git a/configure b/configure
index 1f3ef7a0..77ef2b52 100755
--- a/configure
+++ b/configure
@@ -1570,9 +1570,11 @@ case "${ARCH}" in

             if test ${MTUNE_CASCADELAKE_AVAILABLE} -eq 1; then
                 avx512flag="${avx512flag} -mtune=cascadelake"
+                avx512vnniflag="${avx512vnniflag} -mtune=cascadelake"
             else
                 if test ${MTUNE_SKYLAKE_AVX512_AVAILABLE} -eq 1; then
                     avx512flag="${avx512flag} -mtune=skylake-avx512"
+                    avx512vnniflag="${avx512vnniflag} -mtune=skylake-avx512"
                 fi
             fi
```

configure
```diff
@@ -106,7 +106,7 @@ forcesse2=0
 # For CPUs that can benefit from AVX512, it seems GCC generates suboptimal
 # instruction scheduling unless you specify a reasonable -mtune= target
 avx512flag="-mavx512f -mavx512dq -mavx512bw -mavx512vl"
-avx512vnniflag="-mavx512vnni ${avx512flag}"
+avx512vnniflag="${avx512flag} -mavx512vnni -mtune=cascadelake"
```
mtl1979: Hardcoding the `-mtune` flag here would break older compilers.

Dead2: Yeah, wikichip seemed to suggest cascade lake was the first cpu supporting VNNI, and first gcc release supporting cascade lake also supported VNNI. But wikichip was misleading.
Will have to fix this for cmake as well.

mtl1979: Ice Lake was first Intel CPU to have VNNI instructions. We (currently) use same tuning target for both AVX512 and VNNI.

Dead2: If by currently you mean with this PR then yes. The current develop does not with neither configure nor cmake.

Consider contributing changes to the Microarchitecture support table here: https://en.wikichip.org/wiki/x86/avx512_vnni 😉

configure
```diff
@@ -1570,9 +1570,11 @@ case "${ARCH}" in
 
             if test ${MTUNE_CASCADELAKE_AVAILABLE} -eq 1; then
                 avx512flag="${avx512flag} -mtune=cascadelake"
+                avx512flag="${avx512vnniflag} -mtune=cascadelake"
             else
                 if test ${MTUNE_SKYLAKE_AVX512_AVAILABLE} -eq 1; then
                     avx512flag="${avx512flag} -mtune=skylake-avx512"
+                    avx512flag="${avx512vnniflag} -mtune=skylake-avx512"
```
nmoinvaz, line 1577: Shouldn't this be avx512vnniflag=?

configure
```diff
@@ -1570,9 +1570,11 @@ case "${ARCH}" in
 
             if test ${MTUNE_CASCADELAKE_AVAILABLE} -eq 1; then
                 avx512flag="${avx512flag} -mtune=cascadelake"
+                avx512flag="${avx512vnniflag} -mtune=cascadelake"
```
nmoinvaz, line 1573: Shouldn't this be avx512vnniflag=?

configure
mtl1979: Yes.

configure
mtl1979: Yes.

configure
Dead2: Fix pushed

configure
Dead2: Fix pushed

cmake/detect-intrinsics.cmake
```diff
@@ -100,7 +100,13 @@ macro(check_avx512vnni_intrinsics)
         if(NOT NATIVEFLAG)
             set(AVX512VNNIFLAG "-mavx512f -mavx512dq -mavx512bw -mavx512vl -mavx512vnni")
             if(NOT CMAKE_GENERATOR_TOOLSET MATCHES "ClangCl")
-                set(AVX512VNNIFLAG "${AVX512VNNIFLAG} -mtune=cascadelake")
+                check_c_compiler_flag("-mtune=cascadelake" HAVE_CASCADE_LAKE)
+                if(HAVE_CASCADE_LAKE)
+                    set(AVX512FLAG "${AVX512VNNIFLAG} -mtune=cascadelake")
```
mtl1979, line 105: Wrong variable name here too...

cmake/detect-intrinsics.cmake
```diff
@@ -100,7 +100,13 @@ macro(check_avx512vnni_intrinsics)
...
+                else()
+                    set(AVX512FLAG "${AVX512VNNIFLAG} -mtune=skylake-avx512")
```
mtl1979, line 107: Wrong variable name here too...

PR #1489, opened 2023-05-12
Add tests of file_compress/file_uncompress in minigzip and minideflate.

Dead2:


test/cmake/file-compress.cmake
```diff
@@ -0,0 +1,271 @@
+# compress-and-verify.cmake -- Runs a test against an input file to make sure that the specified
```
nmoinvaz: What is different between this file and `compress-and-verify.cmake`?

Dead2: This file works on files, not streams.

Dead2:
This PR increases coverage, but unfortunately it seems like Github Actions chose to run this PR on VMs without AVX512-VNNI and VPCLMULQDQ support, therefore resulting in a total decrease in coverage.
Ref: https://app.codecov.io/gh/zlib-ng/zlib-ng/pull/1489/indirect-changes

test/cmake/compress-and-verify-fopen.cmake
```diff
@@ -0,0 +1,272 @@
...
+endmacro()
+# Clean up temporary files if not on CI
+macro(cleanup)
+    if(NOT DEFINED ENV{CI})
+        cleanup_always()
+    endif()
+endmacro()
+
+# Show differences between two files
+macro(diff src1 src2)
```
nmoinvaz: Perhaps since both `compare-*.cmake` use this same macro we can move it into a different file like `compare-diff.cmake` and include it in both?

Dead2:
I made some changes so it could be merged into the other script, and that now supports both modes. Also removed the two extra tests, opting instead to run some of the others in filemode.

test/cmake/compress-and-verify-fopen.cmake
Dead2: I merged the two files completely instead 😉

Dead2:
Added back the separate tests again, it was a better solution after all.

Dead2:
It is failing all over now, but I am not sure why it triggers this error:
```
CMake Error at cmake/compress-and-verify.cmake:145 (file):
  file COPY cannot copy file
  "/home/actions-runner/_work/zlib-ng/zlib-ng/test/data/corpora/calgary/book1"
  to
  "/home/actions-runner/_work/zlib-ng/zlib-ng/test/Testing/Temporary/book1":
  No such file or directory.
```
It is copying the file to the temp folder (because we can't control output path/name when compressing a file directly).

But I am at a loss as to why the file would suddenly be missing in this test, then it works again in the very next test that uses the same file but tries to compress using -5 instead of -4..
The failing tests seem to be from test-data.cmake, and a random 5-10 are failing each time.

So this probably is a concurrency issue in relation to running multiple tests at the same time, but the copy should not have any problems with that.. Only cause I can think of is if CMake somehow does a move and then copies the file back, but that would be a pretty stupid way to do it.

Any thoughts?

mtl1979:
> It is failing all over now, but I am not sure why it triggers this error:
> 
> ```
> CMake Error at cmake/compress-and-verify.cmake:145 (file):
>   file COPY cannot copy file
>   "/home/actions-runner/_work/zlib-ng/zlib-ng/test/data/corpora/calgary/book1"
>   to
>   "/home/actions-runner/_work/zlib-ng/zlib-ng/test/Testing/Temporary/book1":
>   No such file or directory.
> ```
> 
> It is copying the file to the temp folder (because we can't control output path/name when compressing a file directly).
> 
> But I am at a loss as to why the file would suddenly be missing in this test, then it works again in the very next test that uses the same file but tries to compress using -5 instead of -4.. The failing tests seem to be from test-data.cmake, and a random 5-10 are failing each time.
> 
> So this probably is a concurrency issue in relation to running multiple tests at the same time, but the copy should not have any problems with that.. Only cause I can think of is if CMake somehow does a move and then copies the file back, but that would be a pretty stupid way to do it.
> 
> Any thoughts?

Instead of copying and renaming, you should copy in single step to avoid concurrency issues...

Dead2:
> Instead of copying and renaming, you should copy in single step to avoid concurrency issues...

That was what I tried to do at first, but ´file(COPY ..´ will create a folder with the destination name and put the file there, instead of letting you specify a filename.
Will try to use ´configure_file()´ instead, it is a bit of an hack but should work at least as good as the old copy and rename.

Dead2:
Passes tests now

test/cmake/compress-and-verify.cmake
```diff
@@ -37,11 +38,22 @@ endif()
 if(NOT DEFINED COMPARE)
     set(COMPARE ON)
 endif()
+if(NOT DEFINED FILEMODE)
+    set(FILEMODE OFF)
+endif()
 if(NOT DEFINED COMPRESS_ARGS)
-    set(COMPRESS_ARGS -c -k)
+    if(FILEMODE)
```
nmoinvaz: Optionally:
```cmake
set(COMPRESS_ARGS -3 -k)
if(NOT FILEMODE)
    list(APPEND COMPRESS_ARGS -c)
endif()
```

test/cmake/compress-and-verify.cmake
```diff
@@ -101,24 +113,47 @@ macro(diff src1 src2)
...
 if(NOT EXISTS ${INPUT})
     message(FATAL_ERROR "Cannot find compress input: ${INPUT}")
 endif()
 
 set(COMPRESS_COMMAND ${COMPRESS_TARGET} ${COMPRESS_ARGS})
 
+set(INPUT_FILE ${OUTPUT_BASE})
+
+# Make CMake copy and rename file in one operation
+configure_file(${INPUT} ${INPUT_FILE} COPYONLY)
```
nmoinvaz, line 143: Why does it need to copy file?

test/cmake/compress-and-verify.cmake
```diff
@@ -101,24 +113,47 @@ macro(diff src1 src2)
...
+
 message(STATUS "Compress ${COMPRESS_COMMAND}")
 message(STATUS "  Input: ${INPUT}")
+message(STATUS "  Input_file: ${INPUT_FILE}")
```
nmoinvaz: Remove underscore?

test/cmake/compress-and-verify.cmake
```diff
@@ -101,24 +113,47 @@ macro(diff src1 src2)
     endif()
 endmacro()
 
+
+macro(exec_streams cmd1 src1 dst1)
```
nmoinvaz: Do we need number 1 after each arg name?

test/cmake/compress-and-verify.cmake
```diff
@@ -101,24 +113,47 @@ macro(diff src1 src2)
...
+        "-DCOMMAND=${cmd1}"
+        -DINPUT=${src1}
+        -DOUTPUT=${dst1}
+        "-DSUCCESS_EXIT=${SUCCESS_EXIT}"
+        -P ${CMAKE_CURRENT_LIST_DIR}/run-and-redirect.cmake
+        RESULT_VARIABLE CMD_RESULT
+        COMMAND_ECHO STDOUT)
+endmacro()
+
+macro(exec_files cmd1 src1)
```
nmoinvaz: Do we need number 1 after each arg name?

test/cmake/compress-and-verify.cmake
```diff
@@ -137,24 +172,27 @@ message(STATUS "Decompress ${DECOMPRESS_COMMAND}")
...
+    exec_streams("${DECOMPRESS_COMMAND}" "${OUTPUT_BASE}.gz" "${OUTPUT_BASE}")
+endif()
 
 if(CMD_RESULT)
     cleanup()
     message(FATAL_ERROR "Decompress failed: ${CMD_RESULT}")
 endif()
 
 if(COMPARE)
+    message(STATUS "xxd comparison")
```
nmoinvaz: What's xxd mean?

nmoinvaz (commented):
Overall I like the changes to use one file instead of the two.

test/cmake/compress-and-verify.cmake
Dead2: Otherwise it would just compress the file in the source path to a file with the input name and only appends '.gz'.
So we need to first copy the file to the current tests temp folder and give it the expected name.

test/cmake/compress-and-verify.cmake
Dead2: `xxd` is the command you use in the diff macro..
It might be misleading though, since it is only used if there is a difference.

test/cmake/compress-and-verify.cmake
Dead2: I am unaware of the list of reserved keywords in cmake, therefore didn't want to name them `cmd` and `src`.
Changing them to `tcmd` and `tsrc`.

test/cmake/compress-and-verify.cmake
Dead2: Changed

test/cmake/compress-and-verify.cmake
Dead2: changed

test/cmake/compress-and-verify.cmake
Dead2: done

nmoinvaz (approved):

PR #1495, opened 2023-05-16
Add `stddef.h` include

haampie:
Which provides `size_t` which is used in the header

### DIFF
--- a/zlib.h.in
+++ b/zlib.h.in
@@ -35,6 +35,7 @@
 #endif
 
 #ifndef RC_INVOKED
+#include <stddef.h>
 #include <stdint.h>
 #include <stdarg.h>
 


mtl1979:
My Ubuntu doesn't have `stddef.h`, so it can't be unconditionally included.

haampie:
It's part of the standard 🤔 maybe you're using the C++ compiler? Or some incomplete clang?

nmoinvaz (approved):

haampie:
fwiw, 

```
# find /usr -name stdint.h -o -name stdarg.h -o -name stddef.h
/usr/lib/gcc/aarch64-linux-gnu/11/include/stdint.h
/usr/lib/gcc/aarch64-linux-gnu/11/include/stddef.h
/usr/lib/gcc/aarch64-linux-gnu/11/include/stdarg.h
```

they get installed with a minimal `apt-get install --no-install-recommends gcc`, I would think that a missing stddef.h is a broken compiler toolchain

Dead2:
Your report from from #1491:
>Sorry, but this is still not good enough 😅
>
>path/to/include/zconf.h:28:9: error: unknown type name 'size_t'
>
>Is it missing #include <cstdint.h>

I am unclear on how you triggered this error. Line 28 of zconf.h does not use size_t, in fact I don't find any references to size_t in zconf.h at all.
Is it actually finding the correct zconf.h file?

Full CMake/Configure log would be nice.

Edit: OOpss.  I was in the wrong branch.
Will fix, but after an errand.

phprus:
@Dead2 
I think PR #1497 is best solution.

mtl1979:
> It's part of the standard 🤔 maybe you're using the C++ compiler? Or some incomplete clang?

clang is C++ compiler... There is just one binary like with Microsoft Visual C++.

Dead2:
An alternative solution has been merged, closing this. 👍

PR #1498, opened 2023-05-17
General Optimizations on compare256 Using RVV

alexsifivetw:
We use upstream toolchain and run it on SiFive internal FPGA /w x280 bitstream:
|                | Speedup (RVV/Scalar) |
|----------------|----------------------|
| compare256/1   | 0.1754               |
| compare256/8   | 0.4614               |
| compare256/64  | 3.1771               |
| compare256/256 | 11.6448              |

### DIFF
--- a/arch/riscv/compare256_rvv.c
+++ b/arch/riscv/compare256_rvv.c
@@ -1,0 +1,47 @@
+/* compare256_rvv.c - RVV version of compare256
+ * Copyright (C) 2023 SiFive, Inc. All rights reserved.
+ * Contributed by Alex Chiang <alex.chiang@sifive.com>
+ * For conditions of distribution and use, see copyright notice in zlib.h
+ */
+
+#ifdef RISCV_RVV
+
+#include "../../zbuild.h"
+#include "fallback_builtins.h"
+
+#include <riscv_vector.h>
+
+static inline uint32_t compare256_rvv_static(const uint8_t *src0, const uint8_t *src1) {
+    uint32_t len = 0;
+    size_t vl;
+    long found_diff;
+    do {
+        vl = __riscv_vsetvl_e8m4(256 - len);
+        vuint8m4_t v_src0 = __riscv_vle8_v_u8m4(src0, vl);
+        vuint8m4_t v_src1 = __riscv_vle8_v_u8m4(src1, vl);
+        vbool2_t v_mask = __riscv_vmsne_vv_u8m4_b2(v_src0, v_src1, vl);
+        found_diff = __riscv_vfirst_m_b2(v_mask, vl);
+        if (found_diff >= 0) 
+            return len + (uint32_t)found_diff;
+        src0 += vl, src1 += vl, len += vl;
+    } while (len < 256);
+
+    return 256;
+}
+
+Z_INTERNAL uint32_t compare256_rvv(const uint8_t *src0, const uint8_t *src1) {
+    return compare256_rvv_static(src0, src1);
+}
+
+#define LONGEST_MATCH       longest_match_rvv
+#define COMPARE256          compare256_rvv_static
+
+#include "match_tpl.h"
+
+#define LONGEST_MATCH_SLOW
+#define LONGEST_MATCH       longest_match_slow_rvv
+#define COMPARE256          compare256_rvv_static
+
+#include "match_tpl.h"
+
+#endif // RISCV_RVV
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -731,6 +731,11 @@ if(WITH_OPTIM)
                 add_definitions(-DRISCV_RVV)
                 list(APPEND ZLIB_ARCH_HDRS ${ARCHDIR}/riscv_features.h)
                 list(APPEND ZLIB_ARCH_SRCS ${ARCHDIR}/riscv_features.c)
+                # FIXME: we will not set compile flags for riscv_features.c when 
+                # the kernels update hwcap or hwprobe for riscv
+                set(RVV_SRCS ${ARCHDIR}/riscv_features.c ${ARCHDIR}/compare256_rvv.c)
+                list(APPEND ZLIB_ARCH_SRCS ${RVV_SRCS})
+                set_property(SOURCE ${RVV_SRCS} PROPERTY COMPILE_FLAGS "${RISCVFLAG} ${NOLTOFLAG}")
             else()
                 set(WITH_RVV OFF)
             endif()
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Features
   * CRC32-B implementation using PCLMULQDQ, VPCLMULQDQ, ACLE, & IBM Z
   * Hash table implementation using CRC32-C intrinsics on x86 and ARM
   * Slide hash implementations using SSE2, AVX2, Neon, VMX & VSX
-  * Compare256 implementations using SSE2, AVX2, Neon, & POWER9
+  * Compare256 implementations using SSE2, AVX2, Neon, POWER9 & RVV
   * Inflate chunk copying using SSE2, SSSE3, AVX, Neon & VSX
   * Support for hardware-accelerated deflate using IBM Z DFLTCC
 * Unaligned memory read/writes and large bit buffer improvements
--- a/cpu_features.h
+++ b/cpu_features.h
@@ -180,6 +180,9 @@ extern uint32_t compare256_neon(const uint8_t *src0, const uint8_t *src1);
 #ifdef POWER9
 extern uint32_t compare256_power9(const uint8_t *src0, const uint8_t *src1);
 #endif
+#ifdef RISCV_RVV
+extern uint32_t compare256_rvv(const uint8_t *src0, const uint8_t *src1);
+#endif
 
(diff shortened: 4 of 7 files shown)


nmoinvaz:
Can you post output of benchmark_zlib with compare256?

arch/riscv/compare256_rvv.c
```diff
@@ -0,0 +1,48 @@
...
+#if RISCV_RVV
+
+#include "../../zbuild.h"
+#include "fallback_builtins.h"
+
+#include <riscv_vector.h>
+
+static inline uint32_t compare256_rvv_static(const uint8_t *src0, const uint8_t *src1) {
+    uint32_t len = 0;
+
```
nmoinvaz: No need for extra white line

arch/riscv/compare256_rvv.c
```diff
@@ -0,0 +1,48 @@
...
+    size_t vl;
+    long found;
+
+    do {
+        vl = __riscv_vsetvl_e8m4(256 - len);
```
nmoinvaz, line 19: Why is it 256 - len? Do we need a comment for it?

arch/riscv/compare256_rvv.c
```diff
@@ -0,0 +1,48 @@
...
+        vuint8m4_t v_src0 = __riscv_vle8_v_u8m4(src0, vl);
+        vuint8m4_t v_src1 = __riscv_vle8_v_u8m4(src1, vl);
+        vbool2_t v_mask = __riscv_vmsne_vv_u8m4_b2(v_src0, v_src1, vl);
+        found = __riscv_vfirst_m_b2(v_mask, vl);
+        if (found >= 0) return len + (uint32_t)found;
```
nmoinvaz: Put `return` on its own line.

arch/riscv/compare256_rvv.c
```diff
@@ -0,0 +1,48 @@
...
+
+    size_t vl;
+    long found;
+
+    do {
+        vl = __riscv_vsetvl_e8m4(256 - len);
+        vuint8m4_t v_src0 = __riscv_vle8_v_u8m4(src0, vl);
+        vuint8m4_t v_src1 = __riscv_vle8_v_u8m4(src1, vl);
+        vbool2_t v_mask = __riscv_vmsne_vv_u8m4_b2(v_src0, v_src1, vl);
+        found = __riscv_vfirst_m_b2(v_mask, vl);
```
nmoinvaz: What is `found`? Can't really easily find docs on these functions.. :-(

arch/riscv/compare256_rvv.c
```diff
@@ -0,0 +1,48 @@
+/* compare256_rvv.c - RVV version of compare256
+ * Copyright (C) 2023 SiFive, Inc. All rights reserved.
+ * Contributed by Alex Chiang <alex.chiang@sifive.com>
+ * For conditions of distribution and use, see copyright notice in zlib.h
+ */
+
+#if RISCV_RVV
```
nmoinvaz: Should be `#ifdef` line in `functable.c` usage?

arch/riscv/compare256_rvv.c
alexsifivetw: Thanks, it's done.

arch/riscv/compare256_rvv.c
alexsifivetw: Thanks, it's done.

arch/riscv/compare256_rvv.c
alexsifivetw: Thanks, it's done.

arch/riscv/compare256_rvv.c
alexsifivetw: 256 is the max length of data that we want to compare, `len` represents the length of the data we have processed, so the length of remaining part is `256 - len`.

arch/riscv/compare256_rvv.c
alexsifivetw: 256 is the max length of data that we want to compare, `len` represents the length of the data we have processed, so the length of remaining part is `256 - len`.

arch/riscv/compare256_rvv.c
alexsifivetw: Renaming it to `found_diff` would be more clear, thanks.

alexsifivetw:
> Can you post output of benchmark_zlib with compare256?

We ran the benchmark on FPGA. The time is not significant for benchmarking, but rather the rvv/scalar speedup.
Sorry about that, we cannot provide a detail report. :(

arch/riscv/compare256_rvv.c
nmoinvaz: Does `v1` need to be reinitialized at the top of each loop? If the vector size is evenly divisible by 256 then it shouldn't correct?

nmoinvaz:
CI outputs:
```bash
740: [----------] 2 tests from compare256
740: [ RUN      ] compare256.c
740: [       OK ] compare256.c (2 ms)
740: [ RUN      ] compare256.rvv
740: [       OK ] compare256.rvv (9 ms)
```
2ms vs 9ms. I wonder if something is wrong. Perhaps it is just qemu is slow for RISCV?

arch/riscv/compare256_rvv.c
nmoinvaz: Perhaps it should just be `size_t v1 = __riscv_vsetvl_e8m4(256);` at the top?

alexsifivetw:
> 2ms vs 9ms. I wonder if something is wrong. Perhaps it is just qemu is slow for RISCV?

QEMU has no good implementation for RVV.
I believe that QEMU is good for testing functionality, rather than performance.

arch/riscv/compare256_rvv.c
alexsifivetw: In RISC-V, one way to handling large number of elements is "stripming" where each iteration of a loop handles some numbers of elemenets, and the iterations continue unitil all elemenets have been processed.

We have to update the `vl` value every time, which means how many elements we want to process in this iteration. We cannot set the `vl` in the top (outside of the for-loop), because `vl` may not be a factor of 256.

However, we indeed use a fixed `vl` under some conditions.

Dead2:
RVV should probably be added to README.md line:
`Compare256 implementations using SSE2, AVX2, Neon, & POWER9`

alexsifivetw:
> RVV should probably be added to README.md line: `Compare256 implementations using SSE2, AVX2, Neon, & POWER9`

Thanks, it's done.

Dead2:
FYI, I am not going to hold the first 2.1 stable release for this, but it will be merged in one of the minor releases after that.

nmoinvaz:
> We ran the benchmark on FPGA. The time is not significant for benchmarking, but rather the rvv/scalar speedup.
Sorry about that, we cannot provide a detail report. :(

You ran `benchmark_zlib --benchmark_filter=compare256*` when compiling `-DWITH_BENCHMARKS=ON`? I don't understand the scalar speedup. If there is a speed up shouldn't it also be present in the Google benchmark results?

arch/riscv/compare256_rvv.c
nmoinvaz: Have you tested whether or not using a fixed value like `32` is faster than `256 - len`, or does that not make sense to do? Sorry to be a pain.

KungFuJesus:
> > Can you post output of benchmark_zlib with compare256?
> 
> We ran the benchmark on FPGA. The time is not significant for benchmarking, but rather the rvv/scalar speedup. Sorry about that, we cannot provide a detail report. :(

This is precisely what the benchmark measures - scalar versus vector.  Even on an FPGA this should give a pretty realistic picture for the improvement in this area.  That is, assuming the FPGA implementation of RVV isn't just serial.

alexsifivetw:
> You ran `benchmark_zlib --benchmark_filter=compare256*` when compiling `-DWITH_BENCHMARKS=ON`? I don't understand the scalar speedup. If there is a speed up shouldn't it also be present in the Google benchmark results?

Yes, we ran the benchmark and got the report:

| Benchmark                   | Time  | CPU   | Iterations |
|-----------------------------|-------|-------|------------|
| compare256/c/problem_size   | X1 ns | X2 ns | X3         |
| compare256/rvv/problem_size | Y1 ns | Y2 ns | Y3         |

The speedup formula would be `X1 / Y1`.

arch/riscv/compare256_rvv.c
alexsifivetw: If the vector register is large and allows users to process more than 32 elements in each iteration, using a fixed value can cause some performance waste.

Using `vl` in RISC-V RVV is a common way to partition a large problem size. However, sometimes we use a fixed `vl` for optimization, depending on our conditions.

nmoinvaz: > We cannot set the vl in the top (outside of the for-loop), because vl may not be a factor of 256.

In what instances will `v1` not be a factor of 256? I believe, but could be wrong that vector sizes on other platforms are evenly divisible by 256.

nmoinvaz (approved):

Dead2 (approved):

PR #1500, opened 2023-05-21
Minor code cleanup in deflate.c.

nmoinvaz:


mtl1979 (approved):
LGTM

deflate.c
```diff
@@ -1234,11 +1227,9 @@ void Z_INTERNAL PREFIX(fill_window)(deflate_state *s) {
             } else if (str >= 1) {
                 s->quick_insert_string(s, str + 2 - STD_MIN_MATCH);
             }
-            unsigned int count;
+            unsigned int count = s->insert;
```
KungFuJesus, line 1230: This might be better expressed as a ternary expression.  E.g.:
```unsigned int count = (s->lookahead == 1) ? s->insert - 1 : s->insert;```
Or, we could forego the branch entirely, even with predicating instructions, by:
```unsigned int count = s->insert - s->lookahead;```
Though, perhaps the second option is a bit less intuitive.  Compiler probably wouldn't take that liberty because it doesn't know that value is expected to be either 1 or 0.

KungFuJesus (changes requested):

deflate.c
mtl1979: Ternary operator would not be code cleanup by definition, as it would alter the jump optimization. Mixing pure cleanup and functional changes is quite discouraged.

KungFuJesus: I don't see why it can't be both.  Conciseness adds clarity to for me.

mtl1979: Functional changes require running benchmarks again and posting the results for comparison and review. Pure cleanup commits do not require benchmarks.

nmoinvaz: I don't plan to make any more changes.

KungFuJesus: Alright, I'll look at opening another PR if it appreciably changes the compiler's decision.

Dead2 (approved):

PR #1502, opened 2023-05-29
Simplify deflate stream/state check.

nmoinvaz:


Dead2:
@nmoinvaz I guess you checked that s-status is always only ever set to one of those defines, right?
If so, then this looks good to me. It is a nice cleanup, even though it makes me a little bit uneasy 😉

nmoinvaz:
There is no & operations that occur on it. It is only checked for equality. And it doesn't appear to be used in any complex calculations.

I just noticed there are some newer zlib-ng functions, where we set `status` to a zlib error. `deflateSetParamPre` and `zng_deflateSetParams`. I'm not sure if that is ok. It might be better to have an explicit `ERROR_STATE` instead of setting an error.

Dead2:
Adding `ERROR_STATE` sounds to me like a good idea, perhaps as a separate commit also modifying the two functions?

I just had a thought though, could this state be accessed by the applications? I mean, I think it is not allowed, but we know how rules are sometimes broken.. Breaking applications that misuse the internals is not so bad, but we should make sure it is actually not allowed.

nmoinvaz:
Actually, I took a look at this again and co-mingling of `*_STATE` and `Z_*_ERROR` seems ok, since `Z_*_ERROR` is `< 0`. I rather it be that the `status` can be less than zero, than lose the granular error value. In my checks I check to see that `status >= INIT_STATE && status <= MAX_STATE`. I think we can merge it as is, if you agree.

nmoinvaz:
Regarding applications that might access `deflate_state->status`, it is possible, but it is not approved usage. We can't really stop people from including `deflate.h` and casting the pointer..

PR #1505, opened 2023-05-31
Make sure file exists before adding it as dependency in configure.

mtl1979:
Should fix #1503.

Requires: #1504

PR #1506, opened 2023-05-31
Add MIPS to GitHub workflows.

mtl1979:
Expected to fail configure during creating dependency information before #1505 is merged.


nmoinvaz (approved):
Nice work LGTM

Dead2:
Interestingly it succeeds the build and self-tests. It does print the grep error, but continues normally it seems.

Dead2 (approved):

mtl1979:
> Interestingly it succeeds the build and self-tests. It does print the grep error, but continues normally it seems.

It only fails if make is run twice, because it won't cleanup build correctly due to missing rule in `arch/generic` and bad dependency rules... I might add new test to run make twice to make sure generated Makefile is sane.

Dead2:
> It only fails if make is run twice, because it won't cleanup build correctly due to missing rule in `arch/generic` and bad dependency rules... I might add new test to run make twice to make sure generated Makefile is sane.

Right. I don't think it is really necessary, at least not for more than 1 CI run. (Not one for each arch, I think we can live without that extra load)

mtl1979:
> > It only fails if make is run twice, because it won't cleanup build correctly due to missing rule in `arch/generic` and bad dependency rules... I might add new test to run make twice to make sure generated Makefile is sane.
> 
> Right. I don't think it is really necessary, at least not for more than 1 CI run. (Not one for each arch, I think we can live without that extra load)

It would affect all builds using configure so would need to be in all such CI runs to catch regressions... Build times would double and that is unacceptable in my professional opinion as we are still going to phase out building with configure in near future.

PR #1512, opened 2023-06-01
Add ABI files for MIPS/MIPS64.

mtl1979:
Missing ABI files after adding CI support for MIPS and MIPS64.

### DIFF
--- a/.github/workflows/pkgcheck.yml
+++ b/.github/workflows/pkgcheck.yml
@@ -27,7 +27,7 @@ jobs:
             os: ubuntu-latest
             chost: arm-linux-gnueabihf
             compiler: arm-linux-gnueabihf-gcc
-            cxx-compiler: g++-arm-linux-gnueabihf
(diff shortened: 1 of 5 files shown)


Dead2 (approved):

nmoinvaz:
These ABI files are such a pain..

Dead2:
> These ABI files are such a pain.

I agree.
It might be possible to auto-update them by saving them to cache after their first generation, and loading them from cache if they already exist. IIRC from reading github actions documentation years ago.

mtl1979:
> > These ABI files are such a pain.
> 
> I agree.
> It might be possible to auto-update them by saving them to cache after their first generation, and loading them from cache if they already exist. IIRC from reading github actions documentation years ago.
> 

We should only need to generate them if we add support for new architectures or if we break the ABI during major version update... Running the generate script does take few minutes of my time but it really makes sense to run it on same machine every time so there is no noise due to paths changing unnecessarily.

Issue #1515, opened 2023-06-07
2.1.2: test stuite want to fetch googletest


kloczek:
https://github.com/zlib-ng/zlib-ng/blob/f75c6856c32a0ade1f33e266fb370bac92d15fac/test/CMakeLists.txt#L113-L257

2.0.7 was possible to build and test against system installed gtest.
Result:
```console
+ /usr/bin/cmake -B x86_64-redhat-linux-gnu -D BUILD_SHARED_LIBS=ON -D CMAKE_AR=/usr/bin/gcc-ar -D CMAKE_BUILD_TYPE=RelWithDebInfo -D CMAKE_C_FLAGS_RELEASE=-DNDEBUG -D CMAKE_CXX_FLAGS_RELEASE=-DNDEBUG -D CMAKE_Fortran_FLAGS_RELEASE=-DNDEBUG -D CMAKE_INSTALL_PREFIX=/usr -D CMAKE_NM=/usr/bin/gcc-nm -D CMAKE_RANLIB=/usr/bin/gcc-ranlib -D CMAKE_VERBOSE_MAKEFILE=ON -D INCLUDE_INSTALL_DIR=/usr/include -D LIB_INSTALL_DIR=/usr/lib64 -D LIB_SUFFIX=64 -D SHARE_INSTALL_PREFIX=/usr/share -D SYSCONF_INSTALL_DIR=/etc -S . -D WITH_GZFILEOP=ON -D WITH_MAINTAINER_WARNINGS=OFF -D WITH_NATIVE_INSTRUCTIONS=OFF -D WITH_SANITIZERS=OFF -D ZLIB_COMPAT=ON -D ZLIB_ENABLE_TESTS=ON
-- Using CMake version 3.26.4
-- ZLIB_HEADER_VERSION: 1.2.13
-- ZLIBNG_HEADER_VERSION: 2.1.2
-- The C compiler identification is GNU 13.1.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Arch detected: 'x86_64'
-- Basearch of 'x86_64' has been detected as: 'x86'
-- Performing Test FNO_LTO_AVAILABLE
-- Performing Test FNO_LTO_AVAILABLE - Success
-- Looking for sys/auxv.h
-- Looking for sys/auxv.h - found
-- Looking for sys/sdt.h
[...]

PR #1519, opened 2023-06-11
Try to find system copy of googletest before trying to download it.

mtl1979:
See #1515

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -111,8 +111,6 @@ if(WITH_BENCHMARKS)
 endif()
 
 if(WITH_GTEST)
-    include(FetchContent)
-
     # Google test requires at least C++11
     set(CMAKE_CXX_STANDARD 11)
 
@@ -125,9 +123,13 @@ if(WITH_GTEST)
         # Must set include and compile options before fetching googletest
         include_directories($ENV{LLVM_BUILD_DIR}/include $ENV{LLVM_BUILD_DIR}/include/c++/v1)
         add_compile_options(-stdlib=libc++ -g)
+    elseif(NOT TARGET GTest::GTest)
+        find_package(GTest)
     endif()
 
     if(NOT TARGET GTest::GTest)
+        include(FetchContent)
+
         # Prevent overriding the parent project's compiler/linker settings for Windows
         set(gtest_force_shared_crt ON CACHE BOOL
             "Use shared (DLL) run-time lib even when Google Test is built as static lib." FORCE)


test/CMakeLists.txt
```diff
@@ -113,6 +113,8 @@ endif()
 if(WITH_GTEST)
     include(FetchContent)
 
+    find_package(GTest)
```
phprus: We need to find a system ``GTest`` only if the cmake target ``GTest::GTest`` is not defined yet (condition on line 130).

mtl1979: When built alone, nothing defines target `GTest::GTest` before that line... I'm still testing if using system version of `googletest` breaks memory sanitizer.

thesamesam: It may be defined if using zlib-ng as a subproject.

mtl1979: > It may be defined if using zlib-ng as a subproject.

True, but it's chicken and egg kind of problem, it can fail and thus we would need to check for existence of the target twice...

phprus: See PR #1277

thesamesam: If the target doesn't exist, call find_package, like we did in minizip-ng: https://github.com/zlib-ng/minizip-ng/issues/654. Then if find_package fails, we fetchcontent instead?

so: target exists (someone imported it for us via subproject) -> find_package -> fetchcontent

phprus: > True, but it's chicken and egg kind of problem, it can fail and thus we would need to check for existence of the target twice...

Yes. We need twice target check.
Suggestion:

Replace:

```cmake
find_package(GTest)
```

with:

```cmake
if(NOT TARGET GTest::GTest)
    find_package(GTest)
endif()
```

mtl1979: Testing for existence of target `GTest::GTest` has been deprecated since cmake 3.20... Preferred way is to use `find_package` which doesn't try to find the dependency twice as the result is cached...

thesamesam: Sorry, I'd completely forgot that `find_package` got substantially smarter!

I got caught out with that before (I even used it in a project, but they needed to support <cmake-3.24). See https://github.com/serge1/ELFIO/pull/113.

mtl1979: We need to support cmake versions down to 3.5, which limits what parameters we can use for some cmake functions.

phprus: > Testing for existence of target `GTest::GTest` has been deprecated since cmake 3.20... Preferred way is to use `find_package` which doesn't try to find the dependency twice as the result is cached...

This is true only for twice find_package call.
If in the parent project GTest is included by a different way (not by find_package), then there may be a library conflict.

We can replace all of ``GTest::GTest`` with ``GTest::gtest`` (https://cmake.org/cmake/help/latest/module/FindGTest.html), but anyway we need add condition before find_package.

mtl1979: As using system copy of googletest breaks with memory sanitizer enabled, I've moved the `find_package()` call later in the file... For now we will do the target test twice to make it work with both standalone build and as a subproject.

phprus (approved):

thesamesam (approved):
Looks right and also tested in a network sandbox, thanks!

test/CMakeLists.txt
```diff
@@ -125,6 +125,8 @@ if(WITH_GTEST)
         # Must set include and compile options before fetching googletest
         include_directories($ENV{LLVM_BUILD_DIR}/include $ENV{LLVM_BUILD_DIR}/include/c++/v1)
         add_compile_options(-stdlib=libc++ -g)
+    elseif(NOT TARGET GTest::GTest)
+        find_package(GTest)
     endif()
 
     if(NOT TARGET GTest::GTest)
```
nmoinvaz, line 130: `include(FetchContent)` can be moved under this `if()`.

mtl1979: Done.

nmoinvaz (approved):

Dead2 (approved):

PR #1520, opened 2023-06-11
[PPC64] Add support for gcc > 12 and clang > 12

mtl1979:
* Use endianess-specific built-in function for gcc < 12 and clang
* Fix build with clang 13, 14, 15 and 16
* Fix detection for availability of Power9 intrinsics


mtl1979:
@KungFuJesus

KungFuJesus:
Hmm I feel like just bumping the gcc version here is not the best answer. I would like to know what gas is doing with {endian} in their intrinsic. Is it possible we just need to just use one of the two now?

Given the intrinsic is supposed to give you the instruction it's thinly veiling I wouldn't love that but it makes some sense maybe. I'd also be curious what clang generates.

mtl1979:
> Hmm I feel like just bumping the gcc version here is not the best answer. I would like to know what gas is doing with {endian} in their intrinsic. Is it possible we just need to just use one of the two now?
> 
> Given the intrinsic is supposed to give you the instruction it's thinly veiling I wouldn't love that but it makes some sense maybe. I'd also be curious what clang generates.

The latest code is basically using just code that is equivalent to one of the intrinsics. The code would get really ugly if we test endianess later in the file just for gcc < 12.

The "old" version didn't support compiling with clang, so there isn't anything to compare against...

Dead2 (approved):

PR #1521, opened 2023-06-16
Update detect-arch.cmake to detect Cortex platform

andreacampanella:
Add Cortex plaform for ARM


andreacampanella:
This is useful for cross-compiling using the Yocto SDK, in my case the architecture is cortexa9t2hf-neon-xilinx but I extended it to all the cortex family.

Dead2 (approved):

Issue #1523, opened 2023-06-20
Stop disabling RPATH on Mac?


dzenanz:
Why do you intentionally disable RPATH for Mac here:
https://github.com/zlib-ng/zlib-ng/blob/afd2a0c323c5c787c1f8a0597568be85f41ab281/CMakeLists.txt#L1136-L1137
@kislinsk reported it as a bug in ITK: https://github.com/InsightSoftwareConsortium/ITK/issues/4084. I would prefer this to be fixed here, as opposed to patching ITK's bundled version.

KungFuJesus:
Git blame goes back to this issue: https://github.com/zlib-ng/zlib-ng/issues/638

dzenanz:
The issues discussed there do not mention RPath directly. Having configure and CMake build systems behaving similarily is not an issue. But then we should switch both to using RPath, instead of both not using it. Do you agree?

mtl1979:
As long as both configure and cmake produce identical binary, changing build parameters is not an issue. Anyone is free to submit a pull request. Not all the long-time contributors have or run MacOS so we have to rely on CI to catch any issues and regressions.

Dead2:
I agree, both configure and CMake should be changed to support `RPATH`. Any takers?

dzenanz:
@hjmjohnson I remember that you use Mac as your main platform. I assume you are also familiar with configure scripts. Do you have time/will to tackle this soon?

hjmjohnson:
@dzenanz I would love to help, but I do not have the extra bandwidth to take this on.  

If someone else makes an initial attempt, I'd be happy to test in a mac environment and report on success/failures, but I do not have the mental bandwidth to address this from scratch.

Sorry :(.

kislinsk:
Regarding the CMake script, all that's necessary can be seen here: https://github.com/MITK/ITK/commit/a08465464b83fb6413e0266369723eb5d2b962b3

Regarding the configure script, the install name seems to be set here: https://github.com/zlib-ng/zlib-ng/blob/develop/configure#L516

So I guess the solution could be to change the line from:

`LDSHAREDFLAGS="-dynamiclib -install_name $libdir/$SHAREDLIBM -compatibility_version $VER1 -current_version $VER3"`

to:

`LDSHAREDFLAGS="-dynamiclib -install_name @rpath/$libdir/$SHAREDLIBM -compatibility_version $VER1 -current_version $VER3"`

I didn't really work with configure files before, though.

[...]

PR #1524, opened 2023-06-23
Re-enable @rpath on OS/X when using relative install prefix

mtl1979:
See #1523.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -963,8 +963,15 @@ else()
   set(PC_INC_INSTALL_DIR "${CMAKE_INSTALL_INCLUDEDIR}")
 endif()
 
+if(APPLE)
+  option(WITH_RPATH "Enable RPATH for shared library" OFF)
+endif()
 if(NOT IS_ABSOLUTE "${CMAKE_INSTALL_LIBDIR}")
-  set(PC_LIB_INSTALL_DIR "\${exec_prefix}/${CMAKE_INSTALL_LIBDIR}")
+  if(APPLE AND WITH_RPATH)
+    set(PC_LIB_INSTALL_DIR "${CMAKE_INSTALL_LIBDIR}")
+  else()
+    set(PC_LIB_INSTALL_DIR "\${exec_prefix}/${CMAKE_INSTALL_LIBDIR}")
+  endif()
 elseif(LIBDIR_POS EQUAL 0)
   string(SUBSTRING "${CMAKE_INSTALL_LIBDIR}" "${INSTALL_PREFIX_LEN}" "-1" LIBDIR_RELATIVE)
   set(PC_LIB_INSTALL_DIR "\${exec_prefix}/${LIBDIR_RELATIVE}")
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -963,8 +963,15 @@ else()
   set(PC_INC_INSTALL_DIR "${CMAKE_INSTALL_INCLUDEDIR}")
 endif()
 
+if(APPLE)
+  option(WITH_RPATH "Enable RPATH for shared library" OFF)
```
nmoinvaz, line 967: Why have an option? Just allow it always.

mtl1979: Because it would break installation to system directory. You can't just change `CMAKE_INSTALL_LIBDIR` to alter RPATH, because the default might match the new value as it's not guaranteed to be absolute path.

dzenanz: Having an option allows backwards compatibility (if someone depended on this). But I think that default should be ON.

dzenanz (commented):
Looks good on a glance. Thanks for working on this.

mtl1979:
> Looks good on a glance. Thanks for working on this.

It wasn't easy to get working the right way as cmake doesn't support paths relative to `DESTDIR`... It only supports paths relative to  `CMAKE_INSTALL_PREFIX`, which defaults to `/usr/local` on Unix-like systems. The trick was to prepend `/` to `CMAKE_INSTALL_LIBDIR` at the latest possible location before defining the install rules but after any rule that generates files.

CMakeLists.txt
mtl1979: The issue with having it ON by default is that there is no way to recover the full RPATH from the information in the cache because the cache isn't built at this stage yet... so the RPATH would be incorrect. With `configure` there is no issue as we don't use cache at all.

It's a lot more common for the `CMAKE_INSTALL_LIBDIR` to contain just `lib` than to contain full path like `/usr/local/lib`. As there is no cache yet, we can't use `CMAKE_INSTALL_FULL_LIBDIR` to determine what the full path would be, as it isn't updated before first `install()` call.

dzenanz: In ITK, we will want to use RPATH by default. I assume the problem with @kislinsk's suggestion (remove custom code) is that configure has slightly different behavior?

mtl1979: The problem is that cmake doesn't like relative paths, so `./lib` must be specified as `/lib` even though the two refer to two distinct paths.

Dead2 (approved):

CMakeLists.txt
Dead2: It does seem pretty cludgy, but I guess if anyone knows of a cleaner way to do this, they can open a PR.

mtl1979: cmake has a lot more limitations than configure script and I'm pretty sure cmake developers are not going to fix every limitation that is by design. I tried whole night different options and either it failed CI tests, or it didn't work as expected.

Only limitation in configure script and its Makefiles affecting this PR is that `DESTDIR` must end with `/´, otherwise relative paths don't work at all...

dzenanz: @bradking Do you have some suggestion for CMake build system?

PR #1532, opened 2023-07-01
Optimize adler32 using rvv

alexsifivetw:
We use upstream toolchain and run it on SiFive internal FPGA /w p470 bitstream:

| Benchmark       | SpeedUp (RVV / Scalar) |
|-----------------|------------------------|
| adler32/2048    | 1.662077018            |
| adler32/4096    | 1.797862857            |
| adler32/32768   | 1.601873082            |
| adler32/262144  | 1.701205598            |
| adler32/2097152 | 1.799026509            |
| adler32/4194304 | 1.790676414            |

### DIFF
--- a/arch/riscv/adler32_rvv.c
+++ b/arch/riscv/adler32_rvv.c
@@ -1,0 +1,110 @@
+/* adler32_rvv.c - RVV version of adler32
+ * Copyright (C) 2023 SiFive, Inc. All rights reserved.
+ * Contributed by Alex Chiang <alex.chiang@sifive.com>
+ * For conditions of distribution and use, see copyright notice in zlib.h
+ */
+
+#ifdef RISCV_RVV
+
+#include <riscv_vector.h>
+#include <stdint.h>
+
+#include "../../zbuild.h"
+#include "../../adler32_p.h"
+
+Z_INTERNAL uint32_t adler32_rvv(uint32_t adler, const uint8_t *buf, size_t len) {
+    /* split Adler-32 into component sums */
+    uint32_t sum2 = (adler >> 16) & 0xffff;
+    adler &= 0xffff;
+
+    /* in case user likes doing a byte at a time, keep it fast */
+    if (len == 1) {
+        return adler32_len_1(adler, buf, sum2);
+    }
+
+    /* initial Adler-32 value (deferred check for len == 1 speed) */
+    if (buf == NULL)
+        return 1L;
+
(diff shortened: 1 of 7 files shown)


KungFuJesus:
Is this implementation maybe held up by a slower multiply accumulate? Other architectures tend to fair bigger improvements from this. It is possible to do it with pure widening adds, saving the multiply accumulate for the end.

arch/riscv/adler32_rvv.c
```diff
@@ -0,0 +1,109 @@
...
+    /* initial Adler-32 value (deferred check for len == 1 speed) */
+    if (buf == NULL)
+        return 1L;
+
+    /* in case short lengths are provided, keep it somewhat fast */
+    if (len < 16) {
+        return adler32_len_16(adler, buf, len, sum2);
+    }
+
+    size_t l = len;
```
nmoinvaz: Can you rename variable to `left` or something more clear?

alexsifivetw:
It cause data error (`Z_DATA_ERROR`) in 2 unit tests. 

```
731 - minideflate-file_compress (Failed)
746 - fuzzer_example_dict (Failed)
```

It's strange..., let me try to fix it first.

arch/riscv/adler32_rvv.c
alexsifivetw: Thanks, it's done.

alexsifivetw:
> Is this implementation maybe held up by a slower multiply accumulate? 

Yes, I believe that the functionality of vector add & mul would not be slow generally.

> It is possible to do it with pure widening adds, saving the multiply accumulate for the end.

Sorry, I cannot move `mul` to the end in the formula.

KungFuJesus:
Looking at your implementation, it looks like you are deferring the multiply-accumulate to the very end (just prior to rebasing to modulo adler base).  

One other thing you could try, assuming your implementation can gang the vector units that wide, is a 512 bit wide version.  Unrolling the loop this way was beneficial for the PowerPC implementation and the AVX512 one as well.

alexsifivetw:
> One other thing you could try, assuming your implementation can gang the vector units that wide, is a 512 bit wide version. Unrolling the loop this way was beneficial for the PowerPC implementation and the AVX512 one as well.

Thank you for your advice. But we don't have enough vector register to do unroll, I almost used all vector register ...

Dead2 (approved):
LGTM

PR #1533, opened 2023-07-06
Get rid of relative #includes in the arch/ code

iii-i:
The arch/ code is built with "-I$SRCDIR".
Remove the ugly ../.. prefixes from #includes.

### DIFF
--- a/arch/arm/adler32_neon.c
+++ b/arch/arm/adler32_neon.c
@@ -7,8 +7,8 @@
  */
 #ifdef ARM_NEON
 #include "neon_intrins.h"
-#include "../../zbuild.h"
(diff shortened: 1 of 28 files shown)


mtl1979 (approved):
LGTM

nmoinvaz:
Didn't we move toward doing relative paths recently? Now we switching back?

iii-i:
I tried to find some discussions about this, but did not succeed. But now that you mentioned there was an explicit switch, I found https://github.com/zlib-ng/zlib-ng/pull/383. Apparently the idea is that the other projects should be able to include the arch/ headers with minimum effort. I think this outweighs the cosmetic concerns, so I'll drop this PR.

PR #1534, opened 2023-07-06
Define an empty __msan_unpoison() without Z_MEMORY_SANITIZER

iii-i:
Currently all the usages of __msan_unpoison() have to be guarded by "#ifdef Z_MEMORY_SANITIZER". Simplify things by defining an empty __msan_unpoison() when the code is compiled without MSan.

### DIFF
--- a/zbuild.h
+++ b/zbuild.h
@@ -245,4 +245,8 @@
 #  endif
 #endif
 
+#ifndef Z_MEMORY_SANITIZER
+#  define __msan_unpoison(a, size) do { Z_UNUSED(a); Z_UNUSED(size); } while (0)
+#endif
+
 #endif
--- a/arch/arm/chunkset_neon.c
+++ b/arch/arm/chunkset_neon.c
@@ -68,10 +68,8 @@ static inline chunk_t GET_CHUNK_MAG(uint8_t *buf, uint32_t *chunk_rem, uint32_t 
     lut_rem_pair lut_rem = perm_idx_lut[dist - 3];
     *chunk_rem = lut_rem.remval;
 
-#ifdef Z_MEMORY_SANITIZER
     /* See note in chunkset_ssse3.c for why this is ok */
     __msan_unpoison(buf + dist, 16 - dist);
-#endif
 
     /* This version of table is only available on aarch64 */
 #if defined(_M_ARM64) || defined(__aarch64__)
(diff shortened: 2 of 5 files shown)


zbuild.h
```diff
@@ -245,4 +245,10 @@
 #  endif
 #endif
 
+#ifndef Z_MEMORY_SANITIZER
+static inline void __msan_unpoison(const volatile void *a, size_t size)
```
nmoinvaz: I think you can probably use a #define instead of empty function?

iii-i: Yes, but an empty function has a couple advantages:
- Variables passed to it are considered used by `-Wunused-variable` (this is relevant for `orig_t2` variable in the DFLTCC code)
- Type checking of arguments (problems introduced by non-MSan builds are found right away and not deferred to CI)

phprus: The ``a`` and ``size`` parameters may generate an "unused variable" warning in the ``__msan_unpoison`` function.

mtl1979: It's better to mark the function parameters "unused" explicitly...

nmoinvaz: I don't like the idea of starting to include static functions in zbuild.h which is used everywhere.
Something like `#define _msan_posion(x, y) 1` is simple and will lead to less warnings.

iii-i: Would `#define _msan_posion(x, y) do { (void)(x); (void)(y); } while (0)` be okay? Otherwise `orig_t2` in `dfltcc()` will need to keep the `#ifdef` guard, making the things not as nice as they could've been.

mtl1979: > I don't like the idea of starting to include static functions in zbuild.h which is used everywhere.
> Something like `#define _msan_posion(x, y) 1` is simple and will lead to less warnings.

`static` here merely means the function is local to current compilation unit... This allows it to be inlined on all common compilers and doesn't throw error due to duplicate symbols.

I'm not sure introducing dead code is the best practice, but I've seen other projects do the same for functions that are used only for some configurations.

nmoinvaz:
> Would #define _msan_posion(x, y) do { (void)(x); (void)(y); } while (0) be okay?

```c
#define _msan_posion(a, size) do { Z_UNUSED(a); Z_UNUSED(size); } while (0)
```

zbuild.h
```diff
@@ -245,4 +245,8 @@
 #  endif
 #endif
 
+#ifndef Z_MEMORY_SANITIZER
+#define __msan_unpoison(a, size) do { Z_UNUSED(a); Z_UNUSED(size); } while (0)
```
nmoinvaz: Two space indent.. `#define `-> `#  define`  sorry

iii-i: My bad, I should've paid attention to the surrounding code.

nmoinvaz (approved):
LGTM

Dead2 (approved):

PR #1541, opened 2023-07-20
Cleanup of instruction set detection and defines.

Dead2:
- Clean up SSE42 vs SSE42_CRC_ASM defines
- Simplify native arch handling in CMake
- Improve native arch handling in Configure
- Move cmake macro to only place it is used.

I recommend reviewing these commits separately.


CMakeLists.txt
```diff
@@ -252,15 +238,25 @@ elseif(CMAKE_C_COMPILER_ID MATCHES "GNU" OR CMAKE_C_COMPILER_ID MATCHES "Clang")
...
+    # Set native march/mcpu
+    if (CMAKE_C_COMPILER_ID MATCHES "GNU" OR CMAKE_C_COMPILER_ID MATCHES "Clang")
+        if(BASEARCH_PPC_FOUND)
+            set(NATIVEFLAG "-mcpu=native")
+        else()
+            set(NATIVEFLAG "-march=native")
+        endif()
+        add_compile_options(${NATIVEFLAG})
+    else()
+        if(WITH_NATIVE_INSTRUCTIONS)
```
nmoinvaz: This will always be true.

configure
```diff
@@ -1607,12 +1594,14 @@ case "${ARCH}" in
             check_sse42_intrinsics
 
             if test ${HAVE_SSE42CRC_INTRIN} -eq 1 || test ${HAVE_SSE42CRC_INLINE_ASM} -eq 1; then
-                CFLAGS="${CFLAGS} -DX86_SSE42"
-                SFLAGS="${SFLAGS} -DX86_SSE42"
-
                 if test ${HAVE_SSE42CRC_INTRIN} -eq 1; then
-                  CFLAGS="${CFLAGS} -DX86_SSE42_CRC_INTRIN"
-                  SFLAGS="${SFLAGS} -DX86_SSE42_CRC_INTRIN"
+                    CFLAGS="${CFLAGS} -DX86_SSE42"
```
nmoinvaz: Indent seems off here.

Dead2:
I also noticed that `configure --native` builds on arm are broken, since for some reason it sets `ARCH=native` early on, and AFAICT that effectively makes configure never activate any of the arch-specific optimizations, making the build only use basic C functions.

I attempted to make a go at fixing this, but it is relatively involved and better suited for a separate PR after this one has been merged. Also it would better if someone who knows the ARM parts of the build/code better, as it might take me a few tries to fix it properly without screwing something up.

configure
Dead2: True, but the wrong indent was actually in the if below.
The old indent here was 2 spaces instead of 4, so this one is correct.

CMakeLists.txt
Dead2: Indeed 👍

configure
nmoinvaz: Yeah I didn't look to see which is the wrong one.

CMakeLists.txt
```diff
@@ -828,13 +806,15 @@ if(WITH_OPTIM)
...
                 set(SSE42_SRCS ${ARCHDIR}/adler32_sse42.c ${ARCHDIR}/insert_string_sse42.c)
                 add_feature_info(SSE42_CRC 1 "Support SSE4.2 optimized CRC hash generation, using \"${SSE42FLAG}\"")
                 list(APPEND ZLIB_ARCH_SRCS ${SSE42_SRCS})
                 set_property(SOURCE ${SSE42_SRCS} PROPERTY COMPILE_FLAGS "${SSE42FLAG} ${NOLTOFLAG}")
                 if(HAVE_SSE42CRC_INTRIN)
-                    add_definitions(-DX86_SSE42_CRC_INTRIN)
+                    add_definitions(-DX86_SSE42)
+                endif()
+                if(HAVE_SSE42CRC_INLINE_ASM)
+                    add_definitions(-DX86_SSE42_CRC_ASM)
```
phprus, line 815: Zlib-ng required gcc 8+, clang 9+ and MSVC 2019+ (https://github.com/zlib-ng/zlib-ng/wiki).
All these compilers support the ``_mm_crc32_u32`` and ``__builtin_ia32_crc32si`` intrinsics.

We can simplify these checks by removing the unused assembler implementation.

Dead2:
Redid some of the first commit, since I now have a better idea of what all the code does and this fixes some of the remaining corner cases.

CMakeLists.txt
phprus: GCC has supported ``_mm_crc32_u32`` since at least version 4.4.
An implementation using ``__builtin_ia32_crc32si`` is redundant.

Dead2: > GCC has supported `_mm_crc32_u32` since at least version 4.4. An implementation using `__builtin_ia32_crc32si` is redundant.

This probably ended up this way due to gcc support being added first using the builtin and then later the MSC variant that uses the more standardized function and nobody actually checked whether GCC/Clang/etc supported that.

CMakeLists.txt
```diff
@@ -827,15 +805,18 @@ if(WITH_OPTIM)
         endif()
         if(WITH_SSE42)
             check_sse42_intrinsics()
-            if(HAVE_SSE42CRC_INLINE_ASM OR HAVE_SSE42CRC_INTRIN)
-                add_definitions(-DX86_SSE42)
-                set(SSE42_SRCS ${ARCHDIR}/adler32_sse42.c ${ARCHDIR}/insert_string_sse42.c)
-                add_feature_info(SSE42_CRC 1 "Support SSE4.2 optimized CRC hash generation, using \"${SSE42FLAG}\"")
-                list(APPEND ZLIB_ARCH_SRCS ${SSE42_SRCS})
-                set_property(SOURCE ${SSE42_SRCS} PROPERTY COMPILE_FLAGS "${SSE42FLAG} ${NOLTOFLAG}")
+            if(HAVE_SSE42CRC_INTRIN OR HAVE_SSE42CRC_INLINE_ASM)
```
mtl1979, line 808: Change on this line is no-op, just reordering the defines.

Dead2: It puts the preferred method first and the fallback second, since that is more logical rather than checking for the fallback first.

CMakeLists.txt
mtl1979: When we started, a lot of people still used gcc 4.x, so it made sense to support all minor versions... Nowadays people use at least gcc 4.8, with a lot of people using at least gcc 6.x series. There is known issue with using gcc 7.x or later with some Xeon processors.

Dead2: > Zlib-ng required gcc 8+, clang 9+ and MSVC 2019+ (https://github.com/zlib-ng/zlib-ng/wiki). All these compilers support the `_mm_crc32_u32` and `__builtin_ia32_crc32si` intrinsics.
> 
> We can simplify these checks by removing the unused assembler implementation.

The asm fallback might in some cases still be valuable while cross-compiling or for compiling in a container. It is very much a corner case, but I don't think it hurts much to keep it around.

CMakeLists.txt
```diff
@@ -827,15 +805,18 @@ if(WITH_OPTIM)
...
                 if(HAVE_SSE42CRC_INTRIN)
-                    add_definitions(-DX86_SSE42_CRC_INTRIN)
+                    add_definitions(-DX86_SSE42)
```
mtl1979, line 810: I'd rather see this kept as `X86_SSE42_CRC_INTRIN`, even if we only use CRC32 intrinsics from SSE42. This way it makes it more obvious that the two choices are mutually exclusive options. We can have inline assembly supported even if the processor doesn't support the instruction set.

CMakeLists.txt
phprus: > The asm fallback might in some cases still be valuable while cross-compiling or for compiling in a container. It is very much a corner case, but I don't think it hurts much to keep it around.

I think the assembler implementation makes the code more confusing, because is not used in practice.

CMakeLists.txt
Dead2: `X86_SSE42` means that SSE42 intrinsics in general are available, just like the similarly named defines do for all the other instruction sets.

CMakeLists.txt
Dead2: After mulling it over some more, I realize you are correct. It does get used in the cases I mentioned, but that does not matter because those are not acutally usable with --native anyways.

Dead2:
I think I'll re-do this PR to make it more organized and incorporate feedback.

PR #1544, opened 2023-07-21
Cleanup handling of march=native

Dead2:
Simplify handling of native in CMake.

As a side note, ARM support in configure is quite messy. There is a lot of feature tests being defined directly in the final "Set arch specific flags" part (line ~1500+). All the other arches have their test functions defined well before that, and only have function calls in that part. Better readability and easier to reuse the test code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,10 +168,6 @@ if(CMAKE_C_COMPILER_ID MATCHES "^Intel")
         set(WARNFLAGS_MAINTAINER /W5)
         set(WARNFLAGS_DISABLE)
     endif()
-    if(WITH_NATIVE_INSTRUCTIONS)
-        message(STATUS "Ignoring WITH_NATIVE_INSTRUCTIONS; not supported on this configuration")
-        set(WITH_NATIVE_INSTRUCTIONS OFF)
-    endif()
 elseif(MSVC)
     # Minimum supported MSVC version is 1800 = Visual Studio 12.0/2013
     # See also https://cmake.org/cmake/help/latest/variable/MSVC_VERSION.html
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -252,15 +238,23 @@ elseif(CMAKE_C_COMPILER_ID MATCHES "GNU" OR CMAKE_C_COMPILER_ID MATCHES "Clang")
     if(MINGW)
         list(APPEND WARNFLAGS_DISABLE -Wno-pedantic-ms-format)
     endif()
-else()
-    if(WITH_NATIVE_INSTRUCTIONS)
+endif()
+
+if(WITH_NATIVE_INSTRUCTIONS)
+    # Set native march/mcpu
+    if (CMAKE_C_COMPILER_ID MATCHES "GNU" OR CMAKE_C_COMPILER_ID MATCHES "Clang")
```
nmoinvaz: Remove space after `if`?

nmoinvaz (commented):
I have questions whether `check_neon_compiler_flag` and `check_acle_compiler_flag` still need to be run if `native`. I can't really tell.. Otherwise looks gtm.

CMakeLists.txt
Dead2: Fixed

Dead2:
Rebased.

Dead2:
> I have questions whether `check_neon_compiler_flag` and `check_acle_compiler_flag` still need to be run if `native`. I can't really tell.. Otherwise looks gtm.

I think the correct thing to do is to rewrite those checks to instead use the native flag, and then verify that the neon/acle/etc code still compiles with native. And let that decide whether we attempt to include those optimized functions in the library or not.

Arm "native" was completely broken in configure before this PR, and with this it is less broken.
Before it would never compile any of the instruction set optimized functions.
With this PR it will enable all optimized functions that the compiler can support, not only the ones the cpu can support, different breakage is expected since much of the now-enabled code has never been tested like this before, only assumed to work.

I am tempted to just rip out the native support from configure completely, for all arches, rather than attempt to fix it and maintain it, and only support such advanced usage in CMake. It is a niche feature, and it is not trivial to get right with all the arch-specific code we have.

nmoinvaz:
Regarding my question, I don't really expect it to be resolved with this PR. It is just a strange scenario.

nmoinvaz (approved):

Dead2:
Rebased and removed commit making changes to Configure. CMake changes remain unchanged.

CMakeLists.txt
```diff
@@ -252,15 +238,23 @@ elseif(CMAKE_C_COMPILER_ID MATCHES "GNU" OR CMAKE_C_COMPILER_ID MATCHES "Clang")
...
+    else()
         message(STATUS "Ignoring WITH_NATIVE_INSTRUCTIONS; not implemented yet on this configuration")
         set(WITH_NATIVE_INSTRUCTIONS OFF)
     endif()
-endif()
-
-# Disable LTO
-if(NOT WITH_NATIVE_INSTRUCTIONS)
+else()
+    # Disable LTO
```
phprus: ``WITH_NATIVE_INSTRUCTIONS`` can be turned off in the previous ``if`` if compiler does not support a native flag.
I think we need to check the ``NOT  WITH_NATIVE_INSTRUCTIONS`` condition again.

Dead2: Technically, you are correct.
But, if you ASK for native, then you should not also ask for LTO since we don't support that combination.
I don't really think it makes sense to allow LTO anyway in the case that native failed to activate, that would be less predictable than just rejecting it like advertised.

phprus: This code doesn't enable LTO, it disable LTO.
If we can't activate native compilation we need to disable LTO.

Dead2: Sorry, I think I confused myself (and you possibly) yesterday.

Legal possibilities:
- No LTO, No native
- No LTO, Native
- LTO, Native

That leaves us with this rule: If you don't ask for native, we always disable LTO.

So that means you are right, there is a bug here; we should disable LTO if native fails to activate.
Thanks for sticking with it 👍

Dead2: @phprus Fixed, please verify.

phprus: @Dead2 
LGTM

PR #1546, opened 2023-07-24
Relocatable pkg config files, `@rpath/<libraryname>` install name on macOS

### LINKED ISSUE #1523: Stop disabling RPATH on Mac?

haampie:
Closes #1523 

1. Make the install name `@rpath/<libz.dylib>` on macOS for CMake and configure. This is what [CMake 3.0+ does by default (CMP0042)](https://cmake.org/cmake/help/latest/policy/CMP0042.html)[^1] and is a very sensible.

2. When using GNUInstallDirs, there's `CMAKE_INSTALL_*` for relative paths, and `CMAKE_INSTALL_FULL_*` for absolute paths (joined with `CMAKE_INSTALL_PREFIX` + some special rules for multi-arch systems). Absolute paths in `CMAKE_INSTALL_*DIR` are allowed but discouraged, since that makes things non-relocatable. This PR drops the complexity of generating relocatable pc files when using absolute paths in those variables. If you want a custom layout, but a relocatable install, use e.g. `CMAKE_INSTALL_[LIB|INCLUDE]DIR=relative/path`, it's as simple as that.

3. In contrast to `CMake` + `GNUInstallDirs`, the configure script expects absolute paths for `--libdir` (or `./configure '--libdir=${prefix}/lib'` literally?). I'm not familiar enough with autotools, but looks like it's standard: [Standard Directory Variables](https://www.gnu.org/software/automake/manual/html_node/Standard-Directory-Variables.html). Or maybe this is to mimic upstream zlib? In any case, with this PR, both the generated autotools & cmake pc file are relocatable by default, unless `--libdir=/absolute/path` or `CMAKE_INSTALL_LIBDIR=/absolute/path` is given.

[^1]: For what it's worth, there are two approaches to make libraries easily locatable in non-standard directories, both are rather hacky:
    
    1. Set the install name to `@rpath/libfoo.dylib` instead of `libfoo.dylib`, so that dependent libraries end up with a load command of `@rpath/libfoo.dylib` after linking, and rpaths can be used automatically.
    2. Set the install name to `/absolute/path/of/libfoo.dylib` so that dependent libraries end up with a load command of `/absolute/path/of/libfoo.dylib`, meaning that the dynamic linker doesn't have to search but can immediately open it.
    
    As far as I know, the second option is the most hacky and non-relocatable (using an absolute path as a library name...), the first option is the default of CMake. Option 1 is very common.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -951,36 +951,8 @@ if(NOT CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_CURRENT_BINARY_DIR)
     endif()
 endif()
 
-# Refer to prefix symbolically to ease relocation by end user,
-# as Makefile-generated .pc file does.
-string(FIND "${CMAKE_INSTALL_INCLUDEDIR}" "${CMAKE_INSTALL_PREFIX}/" INCLUDEDIR_POS)
-string(FIND "${CMAKE_INSTALL_LIBDIR}" "${CMAKE_INSTALL_PREFIX}/" LIBDIR_POS)
-string(LENGTH "${CMAKE_INSTALL_PREFIX}/" INSTALL_PREFIX_LEN)
-
-if(NOT IS_ABSOLUTE "${CMAKE_INSTALL_INCLUDEDIR}")
-  set(PC_INC_INSTALL_DIR "\${prefix}/${CMAKE_INSTALL_INCLUDEDIR}")
-elseif(INCLUDEDIR_POS EQUAL 0)
-  string(SUBSTRING "${CMAKE_INSTALL_INCLUDEDIR}" "${INSTALL_PREFIX_LEN}" "-1" INCLUDEDIR_RELATIVE)
-  set(PC_INC_INSTALL_DIR "\${prefix}/${INCLUDEDIR_RELATIVE}")
-else()
-  set(PC_INC_INSTALL_DIR "${CMAKE_INSTALL_INCLUDEDIR}")
-endif()
-
-if(APPLE)
-  option(WITH_RPATH "Enable RPATH for shared library" OFF)
-endif()
-if(NOT IS_ABSOLUTE "${CMAKE_INSTALL_LIBDIR}")
-  if(APPLE AND WITH_RPATH)
-    set(PC_LIB_INSTALL_DIR "${CMAKE_INSTALL_LIBDIR}")
-  else()
-    set(PC_LIB_INSTALL_DIR "\${exec_prefix}/${CMAKE_INSTALL_LIBDIR}")
-  endif()
-elseif(LIBDIR_POS EQUAL 0)
-  string(SUBSTRING "${CMAKE_INSTALL_LIBDIR}" "${INSTALL_PREFIX_LEN}" "-1" LIBDIR_RELATIVE)
-  set(PC_LIB_INSTALL_DIR "\${exec_prefix}/${LIBDIR_RELATIVE}")
-else()
-  set(PC_LIB_INSTALL_DIR "${CMAKE_INSTALL_LIBDIR}")
-endif()
+set(PC_INC_INSTALL_DIR "${CMAKE_INSTALL_FULL_INCLUDEDIR}")
+set(PC_LIB_INSTALL_DIR "${CMAKE_INSTALL_FULL_LIBDIR}")
 
 #============================================================================
 # zlib
@@ -1142,11 +1114,6 @@ if(NOT DEFINED BUILD_SHARED_LIBS OR BUILD_SHARED_LIBS)
         if(NOT APPLE)
             set_target_properties(zlib PROPERTIES LINK_FLAGS
                 "-Wl,--version-script,\"${CMAKE_CURRENT_SOURCE_DIR}/zlib${SUFFIX}.map\"")
-        elseif(IS_ABSOLUTE "${CMAKE_INSTALL_LIBDIR}" OR NOT WITH_RPATH)
-            # Match configure/make's behavior (i.e. don't use @rpath on mac when using absolute path).
-            set_target_properties(zlib PROPERTIES INSTALL_NAME_DIR "@rpath/${CMAKE_INSTALL_FULL_LIBDIR}")
-        else()
-            set_target_properties(zlib PROPERTIES INSTALL_NAME_DIR "@rpath/${CMAKE_INSTALL_LIBDIR}")
         endif()
     endif()
     if(MSYS)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -1142,11 +1114,6 @@ if(NOT DEFINED BUILD_SHARED_LIBS OR BUILD_SHARED_LIBS)
         if(NOT APPLE)
             set_target_properties(zlib PROPERTIES LINK_FLAGS
                 "-Wl,--version-script,\"${CMAKE_CURRENT_SOURCE_DIR}/zlib${SUFFIX}.map\"")
-        elseif(IS_ABSOLUTE "${CMAKE_INSTALL_LIBDIR}" OR NOT WITH_RPATH)
-            # Match configure/make's behavior (i.e. don't use @rpath on mac when using absolute path).
-            set_target_properties(zlib PROPERTIES INSTALL_NAME_DIR "@rpath/${CMAKE_INSTALL_FULL_LIBDIR}")
```
haampie, line 1147: This was completely bogus... `@rpath/<absolute path>` makes the executable/library that links to zlib-ng add a load command `@rpath/absolute/path/to/zlib-ng/lib`, and `@rpath` is substituted by absolute rpaths set in that dependent. So you end up with for example:

```
/absolute/path/to/zlib-ng/lib/absolute/path/to/zlib-ng/lib/
```

as a search path :|

CMakeLists.txt
```diff
@@ -1183,11 +1150,6 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/zlib${SUFFIX}.h.in
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/gzread.c.in
     ${CMAKE_CURRENT_BINARY_DIR}/gzread.c @ONLY)
 
-# Fix install directory after generating zlib.pc/zlib-ng.pc
-if (NOT IS_ABSOLUTE CMAKE_INSTALL_LIBDIR AND WITH_RPATH)
-    set(CMAKE_INSTALL_LIBDIR "/${CMAKE_INSTALL_LIBDIR}")
```
haampie, line 1188: This also looked very surprising to me.

test/pkgcheck.sh
```diff
@@ -81,7 +81,7 @@ Darwin)
   sysctl -n machdep.cpu.features
   sysctl -n machdep.cpu.leaf7_features
   sysctl -n machdep.cpu.extfeatures
-  CMAKE_ARGS="-DCMAKE_INSTALL_LIBDIR=lib -DPKGCONFIG_INSTALL_DIR=/lib/pkgconfig -DWITH_RPATH=on ${CMAKE_ARGS}"
+  CMAKE_ARGS="-DCMAKE_INSTALL_LIBDIR=lib ${CMAKE_ARGS}"
```
haampie: Not quite sure what the point is of setting `CMAKE_INSTALL_LIBDIR` to its default value...
```suggestion
```

configure
```diff
@@ -513,12 +513,7 @@ if test "$gcc" -eq 1 && ($cc $CFLAGS -c $test.c) >> configure.log 2>&1; then
              SHAREDLIBM=${LIBNAME}.$VER1$shared_ext
              SHAREDTARGET=$SHAREDLIBV
              LDSHARED=${LDSHARED-"$cc"}
-             case ${libdir} in
-             /*)
-                 LDSHAREDFLAGS="-dynamiclib -install_name ${libdir}/${SHAREDLIBM} -compatibility_version ${VER1} -current_version ${VER3}" ;;
-             *)
-                 LDSHAREDFLAGS="-dynamiclib -install_name @rpath/${libdir}/${SHAREDLIBM} -compatibility_version ${VER1} -current_version ${VER3}" ;;
```
haampie, line 520: This was bogus in multiple ways:

1. Relative libdir is not relative to `${prefix}` but current working directory on install.
2. If relative libdir worked, you'd end up with `@rpath/lib/libz.dylib`, which is very non-standard... the expectation is that the dependent uses an rpath of the form `<prefix>/lib`, not `<prefix>`.
3. For absolute libdir, it's still better to use `@rpath/${SHAREDLIBM}`, since that's still more relocatable.

CMakeLists.txt
```diff
@@ -951,35 +951,18 @@ if(NOT CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_CURRENT_BINARY_DIR)
     endif()
 endif()
 
-# Refer to prefix symbolically to ease relocation by end user,
-# as Makefile-generated .pc file does.
-string(FIND "${CMAKE_INSTALL_INCLUDEDIR}" "${CMAKE_INSTALL_PREFIX}/" INCLUDEDIR_POS)
```
haampie, line 956: The idea is to drop this complexity. If a user wants to customize the install layout while keeping things relocatable, then just use `CMAKE_INSTALL_INCLUDEDIR=myincludedir` instead of an absolute path, as suggested by the docs: https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#result-variables

test/pkgcheck.sh
```diff
@@ -81,11 +81,6 @@ Darwin)
   sysctl -n machdep.cpu.features
   sysctl -n machdep.cpu.leaf7_features
   sysctl -n machdep.cpu.extfeatures
-  CMAKE_ARGS="-DCMAKE_INSTALL_LIBDIR=lib -DPKGCONFIG_INSTALL_DIR=/lib/pkgconfig -DWITH_RPATH=on ${CMAKE_ARGS}"
-  CONFIGURE_ARGS="--libdir=lib ${CONFIGURE_ARGS}"
```
haampie, line 85: Not sure how this ever passed tests? `--libdir=lib` with `make DESTDIR=hello/` puts the libraries in `hello/lib/*` instead of `hello/${prefix}/lib/*`. `--libdir` takes absolute paths (or maybe `--libdir='${prefix}/custompath'`)

Dead2:
That is quite the cleanup, I'll have to spend some time later for trying to understand the implications of all the changes. But at first glance it is looking quite neat.

Not sure whether you are aware of `cmake/detect-install-dirs.cmake` or not. I don't immediately see anything that would break, but just want to make sure that you are aware of what is done with paths there as well.
It just checks for non-standard path parameters that are currently in use by some distro packages and/or their standard macros (predating GNUInstallDirs I assume).

Edit: Actually, perhaps some of these changes should also be considered to move into `cmake/detect-install-dirs.cmake`.

haampie:
Hadn't looked at `detect-install-dirs.cmake` yet.

If I understand correctly, it sets defaults for the (not `*FULL*`) GNUInstallDir paths if custom & deprecated variables are set. And it also defines `PKGCONFIG_INSTALL_DIR`.

Looks like it isn't necessary to change any of that: the deprecated variables and pkg config path variables should work whether relative or absolute

haampie:
For what it's worth, [Spack](https://github.com/spack/spack)'s pipeline [is green](https://gitlab.spack.io/spack/spack/-/pipelines/452882) with this patch for macOS (and linux). It uses configure not cmake, and it also builds and links hundreds of dependents of zlib-ng. And most importantly, it relies on rpaths for locating libraries, as it builds packages in funny directories that include a hash, like `opt/zlib-ng-2.1.3-wkvj63jwbhwqovuaqy5frwxezlwujnqn`

Previously it was broken on macOS

CMakeLists.txt
mtl1979: Default prefix is `/usr`, which made it impossible to use directory structure where libraries are located in `lib` subdirectory of the output directory. This primarily affected builds where zlib-ng was used as a dependency.

test/pkgcheck.sh
mtl1979: This worked because `Makefile` doesn't add directory separator after `prefix`. This means that `prefix` also had to always end with directory separator.

In your example, `hello/` is the `prefix`, so should not be duplicated.

PR #1553, opened 2023-08-06
Discard repeated words

DimitriPapadopoulos:


Dead2:
@DimitriPapadopoulos Thanks for the fixes.
Tip for next time though; since all these are cleanups, and each commit is small and has uncontroversial changes, it would be easier to have a single PR but with separate commits for each type of change. 😉

PR #1555, opened 2023-08-12
Remove march=native support from Configure.

Dead2:
This removes the partially broken march=native support from Configure.
CMake will still support this mode of compilation.

Compiling like this is an edge-case, and it is non-trivial to test, support and maintain with a codebase that supports many arch specific optimizations like ours. Removing this from Configure makes Configure simpler and easier to maintain, while letting us spend more effort on improving support in CMake.

### DIFF
--- a/.github/workflows/pkgcheck.yml
+++ b/.github/workflows/pkgcheck.yml
@@ -76,13 +76,6 @@ jobs:
             compiler: clang
             cxx-compiler: clang++
 
-          - name: macOS Clang Native
-            os: macOS-11
-            compiler: clang
-            cxx-compiler: clang++
-            cmake-args: -DWITH_NATIVE_INSTRUCTIONS=ON
-            configure-args: --native
-
           - name: macOS Clang Symbol Prefix
             os: macOS-11
             compiler: clang
--- a/configure
+++ b/configure
@@ -101,7 +101,6 @@ builddfltccdeflate=0
 builddfltccinflate=0
 buildcrc32vx=1
 floatabi=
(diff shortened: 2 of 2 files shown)


nmoinvaz:
I don't have any problems with it. As you know I have been in favor of removing configure and nmake for a long time. 😆

mtl1979:
Makes absolutely no sense at all to remove features from configure script if we are going to remove whole configure script soon... Essentially we break testing "native" CMake build on CI.

It makes more sense to make one big patch for the next major version that removes comparing CMake and configure builds and at the same time removes whole configure script.

Dead2:
@mtl1979 We are not removing configure script..

mtl1979:
> @mtl1979 We are not removing configure script..

Good as my Linux still has "too old" CMake version...

nmoinvaz:
@mtl1979 time to upgrade. Stop living in the past. 🤣

mtl1979:
> @mtl1979 time to upgrade. Stop living in the past. 🤣

Can't upgrade work PC used for building binaries... Have to use oldest possible Linux version so it's compatible with all available releases... Can't use gcc 7 or later due to issue while compiling binaries that cause off by 8 in pointers. As everyone knows, gcc developers are more interested in making new releases than fixing bugs in existing ones.

haampie:
I would not remove the configure script, it only complicates bootstrapping. For example cmake needs zlib & curl, curl needs zlib & openssl, openssl needs zlib & perl (as build dep), perl needs zlib.

PR #1561, opened 2023-08-21
use dash not bash for portability

haampie:
Use `#!/bin/sh` in configure script, since `bash` is not universal.
For example Alpine Linux only has dash aka `/bin/sh` by default.
Also the zlib configure script itself is using `/bin/sh` as shebang.

Also use `"$@"` instead of `$*` so that you could in principle pass
args that contain whitespace. For example

```sh
try ${CC} '-DMY_DEFINE="a b c"' $test.c
```

now invokes the compiler with `'-DMY_DEFINE="a b c"`
literally instead of `'-DMY_DEFINE="a' 'b' 'c'`.

### DIFF
--- a/configure
+++ b/configure
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 # configure script for zlib.
 #
 # Normally configure builds both a static and a shared library.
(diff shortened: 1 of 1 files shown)


gdevenyi:
Should probably use shellcheck and checkbashisms to enforce POSIX-only shell before this is merged so the script doesn't break.

Dead2:
Shellcheck is part of the CI, and this PR also fixes the only remaining warning from that.
Checkbashisms is not AFAIK.

haampie:
`checkbashisms` exits without output, assuming that means it's fine.

`shellcheck` didn't catch all issues, like `$((...) | ...)` being parsed differently; it needs a white space like `$( (...) | ...)`

Dead2 (approved):
LGTM

PR #1578, opened 2023-09-17
Don't run test intrinsic code with native flag in CMake.

nmoinvaz:
Native flag should already determine what code will run on the architecture. This appears to have just been an extra run check with limited benefits. Any compiler that compiles code not available on the native platform is buggy and not our problem.


Dead2 (approved):
👍

PR #1579, opened 2023-09-18
Added instructions for cpack to readme.

nmoinvaz:


README.md
```diff
@@ -131,18 +131,28 @@ LD_PRELOAD=/opt/zlib-ng/libz.so.1.2.13.zlib-ng /usr/bin/program
...
 ```
 
+### CPack
+
+After building the cmake project, it is possible to create an installation package using cpack:
+
+```sh or powershell
+cd build
+cpack --config CPackConfig.cmake
+cpack --config CPackSourceConfig.cmake
```
Dead2, line 155: Might want to add a comment about appending `-G RPM` or `-G DEB` to easily create installable rpm or deb packages.
We default to `TGZ`, so you could add `-G <format>` to each, and have a comment about the different possible choices (the most important ones to mention in our context are probably: TGZ, ZIP, RPM, DEB).


You could also mention the less advanced method (but easier to remember) of using make, it will only do the configured TGZ packages though:
```
make package
make package_source
```

nmoinvaz: I have updated the instructions. Regarding the `make` instructions, I rather leave that out to not make it very long and since it is very limited (only does tgz). If anybody has any questions they can search how to use cpack.

Dead2 (approved):

Issue #1572, opened 2023-09-12
Is it possible to replace madler zlib completely with zlib-ng-compat?


ljavorsk:
Hi,

We're in the middle of the discussion about replacing the "classic" [zlib](https://github.com/madler/zlib) with the zlib-ng (built with the ZLIB_COMPAT=ON flag) in Fedora.

There are a lot of concerns and so I wanted to reach out to you, as you could help us answer some of them. There were a lot of comments already in the [fedora-devel thread](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/CDNPJ4SOTRQMYVCDI3ZSY4SP4FYESCWD/#CJ4KZBTVEHEJ2TOAIWTMFLMBFNJVIYXL).

However, do you think that the compat version of zlib-ng could really fully replace the madler/zlib package? Are there any worries from your side against doing it?

Please let me know anything you think is relevant regarding this topic.

mtl1979:
The obvious is that some existing bad code needs to be fixed before zlib can be replaced with zlib-ng... The long-term goal is to update everything to work with compatibility turned off.

There is small but significant chance that switching to zlib-ng can reveal corner cases where it doesn't produce valid output.

Dead2:
@mtl1979 What bad code are you talking about exactly?
Of course it would be best if applications changed to use the zlib-ng native api, both for performance and individual testing.

@ljavorsk 
For replacing zlib system-wide, that is also possible of course. But to avoid most problems, it should trigger a rebuild of all applications depending on zlib. As @mtl1979 says there is a possibility that applications use zlib in ways we have so far not tested, such as depending on zlib internals that are not promised to be that way for example (nginx is one of these, but has a workaround now).

There is AFAIK currently only one known bug related to using our zlib-compat mode, and it does produce invalid output. The culprit is `pigz`, and that relies on zlib internals and might need a patch in pigz to support zlib-ng #1565

These problems should be few however.
One thing we have thought about, but not yet implemented is a compile-time wrapper that exposes a zlib compatible header file, but links to native zlib-ng. That way packages could more easily be compiled to use zlib-ng without being changed, and without changing the system-wide zlib library as well, making it easier to compile problematic packages with stock zlib.

Dead2:
Just to be clear, if something breaks, then that would be a bug. But it might be a bug in zlib-ng or it might be a bug in how an application (mis)uses the zlib api.

mtl1979:
[...]

PR #1580, opened 2023-09-20
Fix the value of ZLIBNG_VER_STATUS

tuliom:
On C, the value F is usually interpreted as an identifier (e.g. name of a variable, function, etc.).
Force the hexadecimal 0xF in order to avoid compiler errors such as:

    F undeclared (first use in this function)

### DIFF
--- a/zlib.h.in
+++ b/zlib.h.in
@@ -54,7 +54,7 @@ extern "C" {
 #define ZLIBNG_VER_MAJOR 2
 #define ZLIBNG_VER_MINOR 1
 #define ZLIBNG_VER_REVISION 3
-#define ZLIBNG_VER_STATUS F         /* 0=devel, 1-E=beta, F=Release */
+#define ZLIBNG_VER_STATUS 0xF         /* 0=devel, 1-E=beta, F=Release */
 #define ZLIBNG_VER_MODIFIED 0       /* non-zero if modified externally from zlib-ng */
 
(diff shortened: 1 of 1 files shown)


tuliom:
For the record, this fixes and issue when building Perl.
It was found while working on issue #1572 .

nmoinvaz (approved):

Dead2:
This change would break the current users though, for example Perl.. see https://github.com/pmqs/Compress-Raw-Zlib/issues/24

Doing it this way was my mistake, unfortunately nobody caught it in reviews before we made a release with it.

In any case, this value was never defined to be hex, so I think the only way forward is to deprecate it (and remove it at a major release), and introduce a new one that is always hex.

`ZLIBNG_VER_STATUSH` ?
```
#define ZLIBNG_VER_STATUS F         /* DEPRECATED: 0=devel, 1-E=beta, F=Release */
#define ZLIBNG_VER_STATUSH 0xF      /* Hex values: 0=devel, 1-E=beta, F=Release */
```

mtl1979:
Like I said on the other PR, it was not considered a bug that the contents of this macro were not valid in C. Unquoted `F` can be treated both as letter F and hexadecimal value 0xF depending on context... This means we can later introduce macros that will convert the contents to either of the two choices using preprocessor string concatenation.

tuliom:
In that case, I'm closing this.
Thanks!

PR #1581, opened 2023-09-21
Deprecate ZLIBNG_VER_STATUS, add new variant ZLIBNG_VER_STATUSH

Dead2:
Fixes the problem of the value not being defined as hex, and thus causing problems for example with the value F not being easily interpreted correctly by applications.

### DIFF
--- a/zlib-ng.h.in
+++ b/zlib-ng.h.in
@@ -53,7 +53,8 @@ extern "C" {
 #define ZLIBNG_VER_MAJOR 2
 #define ZLIBNG_VER_MINOR 1
 #define ZLIBNG_VER_REVISION 3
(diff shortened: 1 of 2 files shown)


mtl1979:
This is pretty much non-issue as the individual values weren't supposed to be used without string concatenation. There is separate version value wherever numeric comparison is needed.

Clean fix would be to just define `F` as `0xF`, so we don't need to change the name... Same for `A` through `E`. One-letter capitalized tokens are quite rare in normal code as variable names often start with minuscule letters.

Dead2:
> Clean fix would be to just define `F` as `0xF`, so we don't need to change the name... Same for `A` through `E`. One-letter capitalized tokens are quite rare in normal code as variable names often start with minuscule letters.

That is a pretty good idea, but I also think it is a good time to just fix this properly so we have less workarounds to carry around forever.

PR #1583, opened 2023-09-25
Make sure inflateCopy() allocates window with the necessary buffer padding

Dead2:
Make sure inflateCopy() allocates window with the necessary buffer for chunked operations.

Based on Chromium bugfix https://chromium-review.googlesource.com/c/chromium/src/+/4876445

### DIFF
--- a/inflate.c
+++ b/inflate.c
@@ -1319,8 +1319,6 @@ int32_t Z_EXPORT PREFIX(inflateSyncPoint)(PREFIX3(stream) *strm) {
 int32_t Z_EXPORT PREFIX(inflateCopy)(PREFIX3(stream) *dest, PREFIX3(stream) *source) {
     struct inflate_state *state;
     struct inflate_state *copy;
-    unsigned char *window;
-    unsigned wsize;
 
     /* check input */
     if (inflateStateCheck(source) || dest == NULL)
@@ -1331,15 +1329,6 @@ int32_t Z_EXPORT PREFIX(inflateCopy)(PREFIX3(stream) *dest, PREFIX3(stream) *sou
     copy = ZALLOC_INFLATE_STATE(source);
     if (copy == NULL)
         return Z_MEM_ERROR;
-    window = NULL;
-    if (state->window != NULL) {
-        wsize = 1U << state->wbits;
-        window = (unsigned char *)ZALLOC_WINDOW(source, wsize, sizeof(unsigned char));
-        if (window == NULL) {
-            ZFREE_STATE(source, copy);
-            return Z_MEM_ERROR;
-        }
-    }
 
     /* copy state */
     memcpy((void *)dest, (void *)source, sizeof(PREFIX3(stream)));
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Looks ok to me. Alternatively, we could also just call:
 ```c
err = inflate_ensure_window(copy);
if (err != Z_OK) {
    ZFREE_STATE(source, copy);
    return err;
}
ZCOPY_WINDOW(copy->window, state->window, (size_t)1U << state->wbits);
```

**Comment by Dead2** [comment]
@nmoinvaz I agree, that would be cleaner.
I hope this works correctly though, we don't actually have any good tests for `inflatecopy` I think, only the basic usage in `infcover`.

File: inflate.c [hunk 1]
```diff
@@ -1332,13 +1333,11 @@ int32_t Z_EXPORT PREFIX(inflateCopy)(PREFIX3(stream) *dest, PREFIX3(stream) *sou
     if (copy == NULL)
         return Z_MEM_ERROR;
     window = NULL;
```
nmoinvaz: You need to remove `window` reference. And use `copy->window` below.

Dead2: Thanks, completely missed that, as did the `infcover` test..
It fails with a segfault now however, but why.. Time to have a look with gdb I guess.

Dead2: `state->wbits` and `state->wsize` were the culprits, had to reorganize the code somewhat since ensure_window does not have access to the source state, and only had the new blank state.

This passes tests and valgrind now.

Dead2:
As a bonus, this reduces text size quite a bit.

Before:
```
 139932    1384       8  141324   2280c libz-ng.so.2
```
After:
```
 139708    1384       8  141100   2272c libz-ng.so.2
```
224 bytes less to potentially pollute the cache 😄

nmoinvaz (approved):
LGTM. Good now we just have one code path for inflate window creation.

nmoinvaz:
Hmm, CI is failing.

inflate.c
```diff
@@ -1350,10 +1339,15 @@ int32_t Z_EXPORT PREFIX(inflateCopy)(PREFIX3(stream) *dest, PREFIX3(stream) *sou
         copy->distcode = copy->codes + (state->distcode - state->codes);
     }
     copy->next = copy->codes + (state->next - state->codes);
-    if (window != NULL) {
-        ZCOPY_WINDOW(window, state->window, (size_t)1U << state->wbits);
+
+    /* window */
+    copy->window = NULL;
```
nmoinvaz: Possibly need wrap `copy->window` creation in `if (state->window != NULL)` check.

Dead2: You are right, that part of the check is missing now, the old check was dual-purpose, checking both the allocation but also indirectly whether state->window is actually allocated in the first place.

inflate.c
```diff
@@ -1350,10 +1339,17 @@ int32_t Z_EXPORT PREFIX(inflateCopy)(PREFIX3(stream) *dest, PREFIX3(stream) *sou
         copy->distcode = copy->codes + (state->distcode - state->codes);
     }
     copy->next = copy->codes + (state->next - state->codes);
-    if (window != NULL) {
-        ZCOPY_WINDOW(window, state->window, (size_t)1U << state->wbits);
+
+    /* window */
+    if (state->window != NULL)
```
nmoinvaz: Missing `{`

Dead2: Getting too late in the night for this  🤦‍♂️

Dead2:
@iii-i I suspect this also affects DFLTCC systems, as `inflateCopy()` does not seem to call `dfltcc_copy_inflate_state()` before this PR. As far as I can tell it really only means it was missing some padding, but thought I ought to nudge you so you can have a look for yourself.
Interestingly I cannot see `ZCOPY_WINDOW` actually being used before this PR, so perhaps it was misplaced in a merge or something.

nmoinvaz (approved):
LGTM

iii-i:
Hi, sorry, I somehow missed this. For what it's worth, the change looks good to me. `ZCOPY_WINDOW` was used here: https://github.com/zlib-ng/zlib-ng/pull/1583/files#diff-a4d1ad62d5ea3dd46bc58016d6a642b364511e75ac9647c4dd33ff9a906777c7L1353

PR #1585, opened 2023-10-08
Support RVV hwcap detect at runtime

Xeonacid:
Fallback to compile-time detect if kernel version lower than 6.5.

### DIFF
--- a/arch/riscv/README.md
+++ b/arch/riscv/README.md
@@ -1,10 +1,10 @@
 # Building RISC-V Target with Cmake #
 
 > **Warning**
-> We cannot detect rvv support at runtime, running the rvv code on a no-rvv target is a risk. Users should disable the rvv when the target does not support it. 
+> Runtime rvv detection (using `hwcap`) requires linux kernel 6.5 or newer.
 >
-> We will have a better solution when the kernels update `hwcap` or `hwprobe` for risc-v.
-
+> When running on older kernels, we fall back to compile-time detection, potentially this can cause crashes if rvv is enabled at compile but not supported by the target cpu.
+> Therefore if older kernel support is needed, rvv should be disabled if the target cpu does not support it.
 ## Prerequisite: Build RISC-V Clang Toolchain and QEMU ##
 
 If you don't have prebuilt clang and riscv64 qemu, you can refer to the [script](https://github.com/sifive/prepare-riscv-toolchain-qemu/blob/main/prepare_riscv_toolchain_qemu.sh) to get the source. Copy the script to the zlib-ng root directory, and run it to download the source and build them. Modify the content according to your conditions (e.g., toolchain version).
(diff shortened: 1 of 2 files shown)


arch/riscv/README.md
```diff
@@ -1,10 +1,7 @@
 # Building RISC-V Target with Cmake #
 
 > **Warning**
-> We cannot detect rvv support at runtime, running the rvv code on a no-rvv target is a risk. Users should disable the rvv when the target does not support it. 
->
-> We will have a better solution when the kernels update `hwcap` or `hwprobe` for risc-v.
-
+> `hwcap` detect for rvv is only supported on kernel 6.5+. If you are using a lower version, please disable the option `WITH_RVV` to avoid crash on non-rvv machines.
```
Dead2: This seems to be saying the fallback for older kernels is not working and to disable RVV entirely if you intend to use older kernels.
That does not seem to be what the code intends, nor what the commit message says.
So I assume this documentation line is wrong and should be fixed..

arch/riscv/riscv_features.c
```diff
@@ -1,15 +1,46 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <stdbool.h>
```
nmoinvaz: This is the first use of `stdbool.h` in this project. Perhaps we can just make the code similar to the other `features.c` code instead?

arch/riscv/README.md
Xeonacid: Thanks for your review. How does it sound now?

arch/riscv/riscv_features.c
Xeonacid: Thanks for your review. Fixed.

arch/riscv/README.md
```diff
@@ -1,10 +1,9 @@
 # Building RISC-V Target with Cmake #
 
 > **Warning**
-> We cannot detect rvv support at runtime, running the rvv code on a no-rvv target is a risk. Users should disable the rvv when the target does not support it. 
+> `hwcap` detect for rvv is only supported on kernel 6.5+.
 >
-> We will have a better solution when the kernels update `hwcap` or `hwprobe` for risc-v.
-
+> If you are using a lower version, it would fallback to compile-time detection, running the rvv code on a no-rvv target is a risk. Users should disable the rvv when the target does not support it.
```
Dead2: Now I see what you are getting at, but it is a little confusing.

How about something like this? (personally I'd also prefer typing RVV in caps, like we do with other instruction sets)
```
Runtime rvv detection (using `hwcap`) requires linux kernel 6.5 or newer.

When running on older kernels, we fall back to compile-time detection, potentially this can cause crashes if rvv is enabled at compile but not supported by the target cpu.
Therefore if older kernel support is needed, rvv should be disabled if the target cpu does not support it.
```

Xeonacid: Thanks! Modified.

Dead2 (approved):
LGTM

PR #1588, opened 2023-10-15
Fix building benchmarks on 32-bit PowerPC.

mtl1979:
32-bit compiler for PowerPC might know about POWER8, but the CPU doesn't support it, so we disable the benchmark to avoid linker error.

### DIFF
--- a/test/benchmarks/benchmark_crc32.cc
+++ b/test/benchmarks/benchmark_crc32.cc
@@ -59,7 +59,7 @@ BENCHMARK_CRC32(braid, PREFIX(crc32_braid), 1);
 
 #ifdef ARM_ACLE
 BENCHMARK_CRC32(acle, crc32_acle, test_cpu_features.arm.has_crc32);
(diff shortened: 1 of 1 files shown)


nmoinvaz:
I'm surprised the CI didn't catch this.

mtl1979:
> I'm surprised the CI didn't catch this.

I was surprised almost 20 years old development board can run recent enough compiler that allows building the benchmark code... The SoC was released in 2006. Debian runs quite nicely on it, Ubuntu doesn't.

Dead2:
Looks like CI is unable to find and build google test (At least the PPC one I had a look at), thus these tests don't get run.

mtl1979:
> Looks like CI is unable to find and build google test (At least the PPC one I had a look at), thus these tests don't get run.

Building benchmarks is not even enabled on PowerPC CI runs... I see only one CI run where building benchmarks is enabled.

Dead2 (approved):
LGTM

PR #1589, opened 2023-10-16
Fix cross-compiling benchmarks and libpng

mtl1979:
* Can't depend on `ZLIB_ENABLE_TESTS`, there is no zlib for libpng to use when cross-compiling.
* cmake is partially case-sensitive, so project names must be lowercase.
* Variable `PNG_INCLUDE_DIR` isn't set yet in this stage, so we need to set it manually.

### DIFF
--- a/test/benchmarks/CMakeLists.txt
+++ b/test/benchmarks/CMakeLists.txt
@@ -54,10 +54,8 @@ if(WIN32)
     target_link_libraries(benchmark_zlib shlwapi)
 endif()
 
-if(ZLIB_ENABLE_TESTS)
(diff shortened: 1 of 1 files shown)


Dead2:
Please base on `develop`, it makes merging and reviewing of 2.1.4 PR easier.

mtl1979:
> Please base on `develop`, it makes merging and reviewing of 2.1.4 PR easier.

Rebased to remove commits from other PRs...

I usually base on top of most recent branch as that makes sure the CI test results are accurate and there is no "false" errors due to issues that have already been fixed.

nmoinvaz (approved):

PR #1591, opened 2023-10-21
Fix clang-cl warnings

yosh-matsuda:
I have found that if the generator is not Visual Studio (e.g. Ninja), clang-cl is not detected correctly and gives warnings:

```
clang-cl: warning: unknown argument ignored in clang-cl: '-mtune=skylake-avx512' [-Wunknown-argument]
```

Why not use `CMAKE_C_COMPILER_FRONTEND_VARIANT` instead of `CMAKE_GENERATOR_TOOLSET` for clang-cl recognition?


nmoinvaz:
Is this new variable documented?

https://gitlab.kitware.com/cmake/cmake/-/issues/19539

nmoinvaz:
Maybe it makes more sense to have a toolchain-clangcl.cmake just for Ninja..?

Dead2:
If it was introduced in 3.15, then it could be used, but behind a version check, and should have a fallback. So perhaps not the cleanest of solutions =/

yosh-matsuda:
It is documented and introduced in 3.14.
https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_FRONTEND_VARIANT.html

yosh-matsuda:
I think another simple solution is to check if [MSVC](https://cmake.org/cmake/help/latest/variable/MSVC.html) after checking the compiler is clang.

> MSVC
> Set to true when the compiler is some version of Microsoft Visual C++ or **another compiler simulating the Visual C++ cl command-line syntax.**

I confirmed that this also works. How about this approach?

```diff
diff --git a/cmake/detect-intrinsics.cmake b/cmake/detect-intrinsics.cmake
index 250ee44..fb7c735 100644
--- a/cmake/detect-intrinsics.cmake
+++ b/cmake/detect-intrinsics.cmake
@@ -85,7 +85,7 @@ macro(check_avx512_intrinsics)
             # For CPUs that can benefit from AVX512, it seems GCC generates suboptimal
             # instruction scheduling unless you specify a reasonable -mtune= target
             set(AVX512FLAG "-mavx512f -mavx512dq -mavx512bw -mavx512vl")
-            if(NOT CMAKE_GENERATOR_TOOLSET MATCHES "[Cc][Ll][Aa][Nn][Gg][Cc][Ll]")
+            if(NOT MSVC)
                 check_c_compiler_flag("-mtune=cascadelake" HAVE_CASCADE_LAKE)
                 if(HAVE_CASCADE_LAKE)
                     set(AVX512FLAG "${AVX512FLAG} -mtune=cascadelake")
@@ -130,7 +130,7 @@ macro(check_avx512vnni_intrinsics)
     elseif(CMAKE_C_COMPILER_ID MATCHES "GNU" OR CMAKE_C_COMPILER_ID MATCHES "Clang")
         if(NOT NATIVEFLAG)
             set(AVX512VNNIFLAG "-mavx512f -mavx512dq -mavx512bw -mavx512vl -mavx512vnni")
-            if(NOT CMAKE_GENERATOR_TOOLSET MATCHES "[Cc][Ll][Aa][Nn][Gg][Cc][Ll]")
+            if(NOT MSVC)
                 check_c_compiler_flag("-mtune=cascadelake" HAVE_CASCADE_LAKE)
                 if(HAVE_CASCADE_LAKE)
                     set(AVX512VNNIFLAG "${AVX512VNNIFLAG} -mtune=cascadelake")
```

Dead2:
3.11
https://cmake.org/cmake/help/v3.11/variable/MSVC.html
"Set to true when the compiler is some version of Microsoft Visual C++ or another compiler simulating Visual C++. Any compiler defining _MSC_VER is considered simulating Visual C++."

3.10
https://cmake.org/cmake/help/v3.10/variable/MSVC.html
"Set to true when the compiler is some version of Microsoft Visual C++."

Not sure whether the behavior changed for 3.11, or maybe just the documentation changed.

We try to support 3.5, so that would have to be verified.

yosh-matsuda:
[History](https://gitlab.kitware.com/cmake/cmake/-/commits/master/Help/variable/MSVC.rst) of changes to the "MSVC" documentation:

* https://gitlab.kitware.com/cmake/cmake/-/commit/efe22af20e0613308df935a05256d55fc420e42b   
  > The current documentation doesn't clarify that compilers simulating
Visual C++ will also have these variables set.
* https://gitlab.kitware.com/cmake/cmake/-/issues/21651   
  > I'm assuming that the observed behaviour is correct and the docs are wrong -- it's common to see `if(MSVC)` in CMake scripts used to conditionally add cl.exe-style arguments.

They seem just to be documentation changes, and I could not find any changes in the "MSVC" variable in the release notes since 3.5.

mtl1979:
I think it would be informational to know what is the contents of variable `CMAKE_GENERATOR_TOOLSET` when clang-cl is not correctly detected... This would allow us to decide which solution is the best and cleanest.

yosh-matsuda:
OK. My CMake command line is like
```
cmake.exe -DCMAKE_C_COMPILER=clang-cl.exe -G "Ninja Multi-Config" ...
```
where we cannot set `CMAKE_GENERATOR_TOOLSET` with `-T` option since it does not support Ninja, unlike VS.

https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_TOOLSET.html

Then, the value of `CMAKE_GENERATOR_TOOLSET` is empty and will not match the "ClangCL" string in the CMake script.

It's probably a minor warning and not a real problem, but if this PR can improve portability, please consider accepting it.

Dead2:
Can anyone confirm that this PR works as intended with CMake 3.5 in combination with both MSVC and Clangcl?
If so, this looks good to me.

mtl1979:
> cmake.exe -DCMAKE_C_COMPILER=clang-cl.exe -G "Ninja Multi-Config" ..

Needs setting both `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER`, otherwise it fails build.

I tested with MSYS2, MINGW64 and Visual Studio 2022 with v143 toolset to make sure nothing unexpected breaks.

With "old" cmake, hell freezes over, cmake panics even before starting to read CMakeLists.txt... With cmake 3.27.7, everything except benchmarks did build cleanly. I could retry with either older version of Visual Studio or older version of Windows, as cmake version is paired with both Windows and Visual Studio versions. 

Oldest usable combination is cmake 3.12 with Visual Studio 2017.9, but that doesn't come with clang-cl, cmake 3.20 is needed for Visual Studio 2019 that has clang-cl.

```
C:/build/git/zlib-ng/mingw64/_deps/benchmark-src/include/benchmark/benchmark.h:480:3: warning: 'len' may be used uninitialized [-Wmaybe-uninitialized]
  480 |   asm volatile("" : "+m,r"(value) : : "memory");
```

Dead2 (approved):

PR #1595, opened 2023-11-03
pkgconfig file: propagate WITH_GZFILEOP

autoantwort:
if you compile with `WITH_GZFILEOP` you currently manually have to define `WITH_GZFILEOP` again. This should not be needed. Fixes https://github.com/microsoft/vcpkg/discussions/34886

An alternative would be to use `#cmakedefine01 WITH_GZFILEOP` in `zlib_name_mangling-ng.h.in`, but this does not work with Makefiles

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1159,6 +1159,9 @@ else()
 endif()
 
 set(ZLIB_PC ${CMAKE_CURRENT_BINARY_DIR}/zlib${SUFFIX}.pc)
+if(WITH_GZFILEOP)
+    set(PKG_CONFIG_CFLAGS "-DWITH_GZFILEOP")
+endif()
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/zlib.pc.cmakein
     ${ZLIB_PC} @ONLY)
 configure_file(${CMAKE_CURRENT_BINARY_DIR}/zconf${SUFFIX}.h.cmakein
--- a/zlib.pc.cmakein
+++ b/zlib.pc.cmakein
@@ -11,4 +11,4 @@ Version: @ZLIB_FULL_VERSION@
 
 Requires:
 Libs: -L${libdir} -L${sharedlibdir} -lz@SUFFIX@
-Cflags: -I${includedir}
+Cflags: -I${includedir} @PKG_CONFIG_CFLAGS@


Dead2 (approved):
This looks like a pretty logical solution to the linked problem.

Edit: This will break all the pkgcheck CI runs though, since it compares to a version without this defined, so we'll also need to update the definitions after this PR.

mtl1979:
> This looks like a pretty logical solution to the linked problem.
> 
> Edit: This will break all the pkgcheck CI runs though, since it compares to a version without this defined, so we'll also need to update the definitions after this PR.
> 

Updating pkgcheck CI runs is not hard. I've done it few times after some bug fix broke them.

mtl1979 (approved):
LGTM

RJVB:
From the looks of it this is only true when NOT building in ZLIB_COMPAT mode, because `WITH_GZFILEOP` isn't even mentioned in the headerfiles installed in compat mode?

If so, this commit introduces an annoying side effect: the `-DWITH_GZFILEOP` definition weasels itself into the `CFLAGS` for many if not all libraries that depend on libz and that are used as dependencies in more complex projects (like Qt). Which can lead to an expensive rebuild that's pointless for a ditto macro definition.

Example after rebuilding a part of Qt with zlib-ng(compat) providing libz:

```
> fgrep WITH_GZFILEOP -R destroot/opt/local/share/qt5/mkspecs
/opt/local/var/lnxports/build/_opt_local_site-ports_aqua_qt5-kde-devel/qt5-kde-devel/work
destroot/opt/local/share/qt5/mkspecs/modules/qt_lib_gui_private.pri:QMAKE_DEFINES_FREETYPE = WITH_GZFILEOP
destroot/opt/local/share/qt5/mkspecs/modules/qt_lib_gui_private.pri:QMAKE_DEFINES_FONTCONFIG = WITH_GZFILEOP
destroot/opt/local/share/qt5/mkspecs/modules/qt_lib_gui_private.pri:QMAKE_DEFINES_LIBPNG = WITH_GZFILEOP
destroot/opt/local/share/qt5/mkspecs/modules/qt_lib_widgets_private.pri:QMAKE_DEFINES_GTK3 = WITH_GZFILEOP
```

PR #1597, opened 2023-11-07
Add adler32_fold_copy_rvv implementation.

sh1boot:
Implement the fold-copy adler32 sum for RVV.

### DIFF
--- a/arch/riscv/adler32_rvv.c
+++ b/arch/riscv/adler32_rvv.c
@@ -12,23 +12,25 @@
 #include "../../zbuild.h"
 #include "../../adler32_p.h"
 
-Z_INTERNAL uint32_t adler32_rvv(uint32_t adler, const uint8_t *buf, size_t len) {
+static inline uint32_t adler32_rvv_impl(uint32_t adler, uint8_t* restrict dst, const uint8_t *src, size_t len, int COPY) {
     /* split Adler-32 into component sums */
     uint32_t sum2 = (adler >> 16) & 0xffff;
     adler &= 0xffff;
 
     /* in case user likes doing a byte at a time, keep it fast */
     if (len == 1) {
-        return adler32_len_1(adler, buf, sum2);
+        if (COPY) memcpy(dst, src, 1);
+        return adler32_len_1(adler, src, sum2);
     }
 
     /* initial Adler-32 value (deferred check for len == 1 speed) */
-    if (buf == NULL)
+    if (src == NULL)
         return 1L;
 
     /* in case short lengths are provided, keep it somewhat fast */
     if (len < 16) {
-        return adler32_len_16(adler, buf, len, sum2);
+        if (COPY) memcpy(dst, src, len);
+        return adler32_len_16(adler, src, len, sum2);
     }
 
     size_t left = len;
@@ -56,10 +58,12 @@ Z_INTERNAL uint32_t adler32_rvv(uint32_t adler, const uint8_t *buf, size_t len) 
         v_buf16_accu = __riscv_vmv_v_x_u16m2(0, vl);
         size_t subprob = block_size;
         while (subprob > 0) {
-            vuint8m1_t v_buf8 = __riscv_vle8_v_u8m1(buf, vl);
+            vuint8m1_t v_buf8 = __riscv_vle8_v_u8m1(src, vl);
+            if (COPY) __riscv_vse8_v_u8m1(dst, v_buf8, vl);
             v_adler32_prev_accu = __riscv_vwaddu_wv_u32m4(v_adler32_prev_accu, v_buf16_accu, vl);
             v_buf16_accu = __riscv_vwaddu_wv_u16m2(v_buf16_accu, v_buf8, vl);
-            buf += vl;
+            src += vl;
+            if (COPY) dst += vl;
             subprob -= vl;
         }
         v_adler32_prev_accu = __riscv_vmacc_vx_u32m4(v_adler32_prev_accu, block_size / vl, v_buf32_accu, vl);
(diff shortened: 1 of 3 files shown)


sh1boot:
@alexsifivetw FYI

arch/riscv/adler32_rvv.c
```diff
@@ -12,23 +12,25 @@
 #include "../../zbuild.h"
 #include "../../adler32_p.h"
 
-Z_INTERNAL uint32_t adler32_rvv(uint32_t adler, const uint8_t *buf, size_t len) {
+static inline uint32_t adler32_rvv_impl(uint32_t adler, uint8_t* restrict dst, const uint8_t *src, size_t len, int COPY) {
```
KungFuJesus, line 15: These things do end up dispatching to separate function table entries, you can remove the branch here at every loop iteration, even if it is very predictable.

KungFuJesus (changes requested):

arch/riscv/adler32_rvv.c
sh1boot: This is an inline function shared by `adler32_fold_copy_rvv` and `adler32_rvv` at the bottom of the file.  Each of those gets a private copy of this function with the constant value of `COPY` optimised away.

nmoinvaz, line 15: Is there an opportunity to optimize `adler32_avx2` and `adler32_avx512` implementations to use `int COPY`?

KungFuJesus: It's a little less explicit than the templatized variant but assuming the compiler can do the most basic of keyhole optimizations, it'd probably compile to the same result (without the preprocessor mess).

KungFuJesus: Maybe qualifying that argument as const is maybe a stronger assertion that the branch never happens?  The compiler can see the entire scope of copy but it couldn't hurt.  Somewhat polices the function itself, too.

KungFuJesus: Let me give it a whirl.  I don't have a strong feeling one way or the other as long as it works but others might have stronger opinions about a compiler's ability to hoist the variable out entirely.

KungFuJesus: Yeah I don't see anything I'd consider a serious regression, here.  It did slightly reorder the instructions but it definitely doesn't branch:
### Old
``` 
13.24 │ a0:   vmovdqu       (%r8,%rax,1),%ymm0
  0.21 │       vmovdqa       %ymm3,%ymm2
 13.19 │       vpsadbw       %ymm8,%ymm0,%ymm1
  0.38 │       vmovdqu       %ymm0,(%rsi,%rax,1)
  9.13 │       vpmaddubsw    %ymm7,%ymm0,%ymm0
  8.22 │       vpmaddwd      %ymm6,%ymm0,%ymm0
 10.79 │       vpaddd        %ymm1,%ymm3,%ymm1
  0.27 │       vpaddd        %ymm5,%ymm2,%ymm2
 18.89 │       vpaddd        %ymm4,%ymm0,%ymm0
  0.55 │       add           $0x20,%rax
  8.07 │       vmovdqa       %ymm1,%ymm3
  0.26 │       vmovdqa       %ymm2,%ymm5
 13.21 │       vmovdqa       %ymm0,%ymm4
       │       cmp           %rdx,%rax
  0.34 │     ↑ jne           a0
```

### New
```
 11.61 │ a0:   vmovdqu       (%r9,%rdx,1),%ymm0
  0.27 │       vmovdqa       %ymm3,%ymm2
 13.64 │       vpsadbw       %ymm8,%ymm0,%ymm1
  0.32 │       vmovdqu       %ymm0,(%r8,%rdx,1)
 10.22 │       vpmaddubsw    %ymm7,%ymm0,%ymm0
  4.11 │       vpmaddwd      %ymm6,%ymm0,%ymm0
 11.88 │       vpaddd        %ymm1,%ymm3,%ymm1
  0.38 │       vpaddd        %ymm5,%ymm2,%ymm2
 17.76 │       vpaddd        %ymm4,%ymm0,%ymm0
  0.46 │       add           $0x20,%rdx
 10.58 │       vmovdqa       %ymm1,%ymm3
  0.30 │       vmovdqa       %ymm2,%ymm5
 14.54 │       vmovdqa       %ymm0,%ymm4
       │       cmp           %rax,%rdx
  0.25 │     ↑ jne           a0
```

Edit: err actually no these are exactly the same, just a few named registers switched around.  Perf's attribution of instruction just shifted a bit, but that's expected and never exact, especially if instruction alignment maybe shifted a tiny bit.

sh1boot: GCC and Clang support `__attribute__((always_inline))` if you really want to nail it down, but that's a thing you'd want to abstract to cmake to replace it with a portable macro.

Or go nuclear.  Write the whole thing in C++ and template it.

mtl1979: Rewriting C code in C++ is recipe for disaster... C++ might allow a lot fancier code, but it comes with quite a big overhead. It's also common to not install C++ compiler or it's dependencies by default, but still have C compiler. This makes using for example clang really hard in some operating systems as one has to install support libraries for g++ first before trying to use clang, as clang is essentially a C++ compiler just like VC++ on Windows. Some code that is valid with C compiler isn't valid with C++ compiler.

During the 40 years I have been a developer, I've seen a lot of projects just rename *.c files to *.cpp or *.cxx just to pretend the code compiles better and still works as expected. There is also a running joke about a lot of C++ headers just being one line that includes the equivalent C header just to conform the naming guidelines of the C++ standard.

KungFuJesus: I think that was mostly in jest, he wasn't suggesting rewriting this whole thing in templated C++.  Though it _would_ do the job we were doing with function redefines and includes admirably.  Just with cryptic link time behavior and hard to resolve type resolution :).

Dead2 (approved):

PR #1598, opened 2023-11-07
[configure] Propagate -DWITH_GZFILEOP to pkgconfig

mtl1979:
Continuation of #1595.

PR #1599, opened 2023-11-07
Untemplate adlercopy

KungFuJesus:
A strategy was made apparent by @sh1boot in https://github.com/zlib-ng/zlib-ng/pull/1597 that we could simply declare inline functions with extra constant arguments and the compiler would hoist the branches entirely outside of the function, generating effectively the same exact code but without the preprocessor misdirection.  This code is marginally more legible and compiles to the same thing.


KungFuJesus:
Alright so what's the deal with pkgcheck?

nmoinvaz:
Recent changes were made to pkgcheck for cmake only. This PR needs to be merged #1598. 

If you want you can rebase off an older commit.

nmoinvaz (approved):

arch/x86/adler32_avx2.c
```diff
@@ -1,17 +1,154 @@
 /* adler32_avx2.c -- compute the Adler-32 checksum of a data stream
- * Copyright (C) 1995-2011 Mark Adler
+ * Copyright (C) 1995-2022 Mark Adler
```
mtl1979: This change seems bad... Copyright doesn't automagically extend unless original author makes further changes and they are merged in. However Mark Adler isn't mentioned in the list of contributing authors.

KungFuJesus: Ah, fixing now. He isn't an author, except for the core algorithm, anyway.  That was already there originally when I had gotten to it (the 1995-2011 bit).

Dead2:
> Alright so what's the deal with pkgcheck?

Rebase to `develop` head should resolve the pkgcheck CI failures, you caught it between two interdependent changes 😉

arch/x86/adler32_avx2.c
mtl1979: Copyright issues are quite complicated, even for open-source projects... Once you add or update copyright statement, it will be really hard to change it back or correct it. GitHub has been pretty eager to take whole repository down if there is contentious change in copyright statements.

arch/x86/adler32_avx2.c
```diff
@@ -1,17 +1,154 @@
...
+
+         We can, however, do a relatively cheap sequence for horizontal sums.
+         Then, we simply do the integer modulus on the resulting 64 bit GPR, on a scalar value. It was
+         previously thought that casting to 64 bit precision was needed prior to the horizontal sum, but
+         that is simply not the case, as NMAX is defined as the maximum number of scalar sums that can be
+         performed on the maximum possible inputs before overflow
+         */
+
+
+         /* In AVX2-land, this trip through GPRs will probably be unvoidable, as there's no cheap and easy
```
mtl1979: "unavoidable"

arch/x86/adler32_avx512.c
```diff
@@ -8,9 +8,108 @@
...
+                                          38, 39, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, 52, 53, 54, 55,
+                                          56, 57, 58, 59, 60, 61, 62, 63, 64);
+    const __m512i dot3v = _mm512_set1_epi16(1);
+    const __m512i zero = _mm512_setzero_si512();
+    size_t k;
+
+    while (len >= 64) {
+        __m512i vs1 = _mm512_zextsi128_si512(_mm_cvtsi32_si128(adler0));
+        __m512i vs2 = _mm512_zextsi128_si512(_mm_cvtsi32_si128(adler1));
+        vs1_0 = vs1;
```
mtl1979, line 58: This gets optimized away as it is only referenced inside the while loop starting from line 65 below. Moving the line to start of while loop might cause it to not get optimized away as long as line 86 is also removed. Codecov essentially doesn't know which one of these two lines are still in the resulting code after optimizer combines the two lines.

KungFuJesus: I'm not entirely sure what you're proposing here.  We definitely want this assignment outside of the inner while loop.  Are we that worried about codecov for this tight loop?

mtl1979: A lot of times codecov is correct about dead code. It isn't however correct in 100% of the cases, so I usually trace things on paper to see if a line gets optimized away. In this case the line you want to be outside the inner loop gets moved inside due to duplicate code elimination.

arch/x86/adler32_avx512.c
```diff
@@ -8,9 +8,108 @@
...
+            src += 64;
+            k -= 64;
+
+            __m512i vs1_sad = _mm512_sad_epu8(vbuf, zero);
+            __m512i v_short_sum2 = _mm512_maddubs_epi16(vbuf, dot2v);
+            vs1 = _mm512_add_epi32(vs1_sad, vs1);
+            vs3 = _mm512_add_epi32(vs3, vs1_0);
+            __m512i vsum2 = _mm512_madd_epi16(v_short_sum2, dot3v);
+            vs2 = _mm512_add_epi32(vsum2, vs2);
+            vs1_0 = vs1;
```
mtl1979, line 86: This line is no-op on last iteration of the loop and such the instruction gets eliminated due to line 58 getting moved inside inner while loop before the line 82.

arch/x86/adler32_avx512.c
nmoinvaz: I think this PR should only be able the refactoring. I don't suggest adding any functional or optimization changes. To me it looks like that code was already there.

arch/x86/adler32_avx512.c
nmoinvaz: Seems like this code already existed before?

KungFuJesus: Yes, only a minor cosmetic fix to indentation was made here.

Dead2 (approved):
LGTM

Issue #1600, opened 2023-11-12
libgit2 test suite failures with zlib-ng in compat mode


joebonrichie:
https://github.com/libgit2/libgit2#running-tests

Tested libs: 
 - ZLIB-NG: 2.1.4 compat mode
 - zlib 1.3.0 canonical 

Against:
 - libgit2 HEAD
 - libgit2 1.3.0
 - libgit2 1.5.2
 - solbuild (via libgit2)

I have found failing tests in the libgit2 testsuite with zlib-ng (compat mode) installed/ld_preloaded. This propagates to applications using libgit2 such as [solbuild](https://github.com/getsolus/solbuild) which will fail to clone repositories with zlib-ng installed. The tests pass as expected with canonical zlib installed. Please let me know if you can replicate.

Rebuilding libgit2 against zlib-ng-devel in compat mode doesn't seem to resolve the issue.

The following tests fail for me with zlib-ng installed:

1/12 Offline
``` 
  1) Failure:
pack::packbuilder::create_pack [/home/ninya/libgit2/tests/libgit2/pack/packbuilder.c:143]
  String mismatch: hex != "5d410bdf97cf896f9007681b92868471d636954b"
  '5f688cedc3a5132f15fc3352a35bc5dcb6a536a4' != '5d410bdf97cf896f9007681b92868471d636954b' (at byte 1)

  2) Failure:
pack::packbuilder::get_name [/home/ninya/libgit2/tests/libgit2/pack/packbuilder.c:151]
  String mismatch: "7f5fa362c664d68ba7221259be1cbd187434b2f0" != git_packbuilder_name(_packbuilder)
  '7f5fa362c664d68ba7221259be1cbd187434b2f0' != 'f120f6bbad709d0e3c021ca8f9d7c26c3882497b' (at byte 0)

  3) Failure:
pack::packbuilder::write_default_path [/home/ninya/libgit2/tests/libgit2/pack/packbuilder.c:159]
  Expression is not true: git_fs_path_exists("objects/pack/pack-7f5fa362c664d68ba7221259be1cbd187434b2f0.idx")

  4) Failure:
pack::packbuilder::permissions_standard [/home/ninya/libgit2/tests/libgit2/pack/packbuilder.c:184]
  Function call failed: (stat("pack-7f5fa362c664d68ba7221259be1cbd187434b2f0.idx", &statbuf))
  error -1 - no error

  5) Failure:
[...]

Issue #1565, opened 2023-08-28
pigz-ng crc32 mismatch


abmybgx:
I built pigz-ng from pigz-bench last week, and I met below error. I tried WITH_OPTIM=OFF option, but it was no use. After I changed back to pigz, it works.

```
nerdctl pull marvinb/carla_simulator:0.9.14

cd /var/lib/containerd/root/io.containerd.content.v1.content/blobs/sha256

./pigz-ng -cd 918d710d636a791b817a550f8b6eb0b78290d90a813f93812d3b57b022ec7a55 > /dev/null
pigz-ng: skipping: 918d710d636a791b817a550f8b6eb0b78290d90a813f93812d3b57b022ec7a55: corrupted -- crc32 mismatch
```

nmoinvaz:
Can you attach your input files that reproduce the problem? What architecture and platform are you using? Is there a particular zlib-ng commit that it works for? Also what version or commit of pigz are you using?

KungFuJesus:
It also sounds like it could be an issue in pigz-ng, rather than zlib-ng.

nmoinvaz:
@KungFuJesus there is no specific pigz-ng project.

nmoinvaz:
The only thing that the pigz folder provides in pigzbench is a CMake project and Windows shim files.

KungFuJesus:
> The only thing that the pigz folder provides in pigzbench is a CMake project and Windows shim files.

Hmm, the @abmybgx claims they are using "pigz-ng" and the failed test is specifically calling itself that.  It also sounds like when switching back to pigz with zlib-ng's compat mode, it had no issues.  Maybe I'm missing something, though.

KungFuJesus:
Oh, the table on that project clarifies my confusion: https://github.com/zlib-ng/pigzbench

So it's just zlib-ng being tested with pigz.

nmoinvaz:
There is a `a_compile.py` that compiles pigz with different variants of zlib. But I'm not sure if it was used to build the issue in this case.

abmybgx:
My pigz-ng was built by a_compile.py with the newest zlib-ng develop branch, you all could follow my operation to have a test, it's a public image. For decompressed bytes file, there was 1 byte mismatch from pigz-ng to pigz. It's ok for pigz or igzip.

nmoinvaz:
Ok. Maybe somebody else can attach the file, I don't use docker.

mtl1979:
[...]

PR #1602, opened 2023-11-17
Fix an issue with regard to finishing out the window

KungFuJesus:
Evidently we weren't hitting this code path very often in our tests?
If we reach here, there are some circumstances where we never ran any
of the bytes through the checksum. We can easily check for this with a
comparison if the initial checksum bytes.
    
This should fix issue #1600, and possibly #1565 as well.

### DIFF
--- a/inflate.c
+++ b/inflate.c
@@ -1112,11 +1112,12 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
      */
   inf_leave:
     RESTORE();
+    uint32_t check_bytes = out - strm->avail_out;
     if (INFLATE_NEED_UPDATEWINDOW(strm) &&
             (state->wsize || (out != strm->avail_out && state->mode < BAD &&
                  (state->mode < CHECK || flush != Z_FINISH)))) {
         /* update sliding window with respective checksum if not in "raw" mode */
-        if (updatewindow(strm, strm->next_out, out - strm->avail_out, state->wrap & 4)) {
+        if (updatewindow(strm, strm->next_out, check_bytes, state->wrap & 4)) {
             state->mode = MEM;
             return Z_MEM_ERROR;
         }
@@ -1129,8 +1130,13 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
 
     strm->data_type = (int)state->bits + (state->last ? 64 : 0) +
                       (state->mode == TYPE ? 128 : 0) + (state->mode == LEN_ || state->mode == COPY_ ? 256 : 0);
-    if (((in == 0 && out == 0) || flush == Z_FINISH) && ret == Z_OK)
+    if (((in == 0 && out == 0) || flush == Z_FINISH) && ret == Z_OK) {
+        /* when no sliding window is used, hash the output bytes if no CHECK state */
+        if (INFLATE_NEED_CHECKSUM(strm) && !state->wsize && flush == Z_FINISH) {
+            inf_chksum(strm, put - check_bytes, check_bytes);
+        }
         ret = Z_BUF_ERROR;
+    }
     return ret;
 }
 


KungFuJesus:
OK, I think this will work.  It's maybe a bit hackier than I'd like but I'm pretty sure correctness isn't an issue.  This should catch the corner case we were seeing.  The main difference is I think there were unconditional checksums being applied in earlier versions in a lot of places.  I don't see a clean delineation where we clearly aren't applying the checksum but should be, though.

KungFuJesus:
Hmm, I have half a mind to write the buffer being fed through that library's test suite and put it into ours.

KungFuJesus:
Not sure what's up with the clang reduced memory failure.

Ahh, it does look like it reduces the window size to 8k, so perhaps this is making us hit a different corner case.  It also has dawned on me that the checksum will always be the initial CRC32 value until the final fold so that bit might need to be reworked.

KungFuJesus:
I reproduced the reduced mem failure, I'll look into it tonight.

KungFuJesus:
Ok I think I got it.  Still a little hackier than I'd like (I'd be much more thrilled about catching the set of conditions that paints us in this corner) but barring better suggestions or proof that this is not catching all of the corner cases, I say it's ready to merge.

inflate.c
```diff
@@ -1050,12 +1050,25 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
...
+                         *  2.) We have a checksum that does wrap perfectly, in which case, we should end up still
+                         *      correct, anyway */
+                        inf_chksum(strm, put, strm->total_out);
                     }
 #ifdef GUNZIP
                     if (state->flags)
                         strm->adler = state->check = functable.crc32_fold_final(&state->crc_fold);
 #endif
                 }
+
```
mtl1979: Adding this empty line seems unnecessary... Mixing white-space and code changes is frowned upon...

arch/x86/crc32_fold_pclmulqdq_tpl.h
```diff
@@ -182,5 +182,9 @@ Z_INTERNAL void CRC32_FOLD(crc32_fold *crc, const uint8_t *src, size_t len, uint
         partial_fold(len, &xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3, &xmm_crc_part);
     }
 
+    /* Sentinel marker that we use for a comparison later, letting us know if
+     * any residual values have been folded up into this checksum that need to
+     * be finalized. In the C & other implementations, this is the actual checksum. */
+    crc->value = 1;
```
mtl1979: Having a comment to explain the reason for why this is set to `1` is a good thing, but for the purpose of reviewing, it is still unclear why the value is exactly `1` and not some "magic constant" like elsewhere.

It is unclear what is `later`... So I would expand it to say exactly which function checks that this value is `1` and not any other value.

KungFuJesus: How's that?

nmoinvaz: Where is the `crc->value` actually being checked?

inflate.c
```diff
@@ -1050,6 +1050,18 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
                 if (INFLATE_NEED_CHECKSUM(strm) && state->wrap & 4) {
                     if (out) {
                         inf_chksum(strm, put - out, out);
+                    } else if ((!state->flags && strm->adler == ADLER32_INITIAL_VALUE)
+#ifdef GUNZIP
+                    || (state->flags && state->crc_fold.value == CRC32_INITIAL_VALUE)) {
```
nmoinvaz: It would be great if there was a way we could clean up this if statement so it doesn't break like this with the `#define`.

arch/x86/crc32_fold_pclmulqdq_tpl.h
nmoinvaz: Oh I see, it is checked as `crc_fold.value == CRC32_INITIAL_VALUE` below.

inflate.c
```diff
@@ -1050,6 +1050,18 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
...
+#else
+                        ) {
+#endif
+                        /* In the event we reach this block and the checksums are at the initial values, run a checksum
+                         * over the whole stream, as one of two things are true:
+                         *  1.) We never ran any of the bytes through the checksum (testing has found this happens)
+                         *  2.) We have a checksum that does wrap perfectly, in which case, we should end up still
+                         *      correct, anyway */
+                        inf_chksum(strm, put, strm->total_out);
                     }
```
nmoinvaz:
```suggestion
                    } else {
                       bool final_chksum = (!state->flags && strm->adler == ADLER32_INITIAL_VALUE);
#ifdef GUNZIP
                       final_chksum |= (state->flags && state->crc_fold.value == CRC32_INITIAL_VALUE));
#endif
                        /* In the event we reach this block and the checksums are at the initial values, run a checksum
                         * over the whole stream, as one of two things are true:
                         *  1.) We never ran any of the bytes through the checksum (testing has found this happens)
                         *  2.) We have a checksum that does wrap perfectly, in which case, we should end up still
                         *      correct, anyway */
                        if (final_chksum)
                            inf_chksum(strm, put, strm->total_out);
                    }
```

arch/x86/crc32_fold_pclmulqdq_tpl.h
nmoinvaz: Does this assignment need to be wrapped in `#ifdef GUNZIP` too?

KungFuJesus: I don't think so, it's a function only actually called when using the crc32 folding routines, which would be enabled anyway with GUNZIP being defined.  Technically these functions are used in the generic crc32 exported function as well, but the extra assignment is hardly consequential, I think.

inflate.c
KungFuJesus: It does require C99 to do this but I don't think that's a huge ask.  We probably already require it.

Dead2 (approved):
LGTM

Dead2:
Wish we did have a CI test for this, so we could make sure that this is not a problem for all the other alternative code-paths as well. I don't think any other arch based their code on `crc32_fold_pclmulqdq` etc and inherited the problem, did they?

KungFuJesus:
The problem isn't specific to crcs, it's specific to inflate. There's a potential encoding through the compression state machine that arrives at the final check instruction without the current window getting run through the checksum routines.

This seems to catch those corner cases, as when it happens the checksum is the initial unaltered checksum, be that Adler or crc. 

As far as something that catches this, I think integrating libgit2's tests into our CI wouldn't be a bad approach.

mtl1979 (approved):
LGTM... We can add CI test in separate pull request.

inflate.c
```diff
@@ -10,6 +10,7 @@
 #include "inflate_p.h"
 #include "inffixed_tbl.h"
 #include "functable.h"
+#include <stdbool.h>
```
nmoinvaz: @Dead2 we okay using this include? First in the library. I had mentioned it before.

nmoinvaz:
I am testing using `minideflate` and comparing to `libgit2`. The `put` variable sent to `inf_chksum` is different.

image

`minideflate` using a next position in write buffer for each call to `inflate()`, where as `libgit2` starts from scratch from the buffer, so `put` is evaluated correctly.

Here is my test data:
[000000A2CD78EBA0-inflate.gz](https://github.com/zlib-ng/zlib-ng/files/13433188/000000A2CD78EBA0-inflate.gz)
```bash
minideflate -d -k -s 4 -r 25 -w 15 -t 8192 000000A2CD78EBA0-inflate.gz
```

And change `minideflate`:
```c
            err = PREFIX(inflate)(&d_stream, flush);
            if (err == Z_BUF_ERROR && read_buf_size != BUFSIZE) {
                err = Z_OK;
            }
```

So essentially, it only works because in `libgit2` they are passing the same buffer pointer of the previous `inflate()` to the the next call to `inflate()`. I still don't think we should commit this until we add a unit test.

nmoinvaz:
This also seems to work although it can probably be better some how:
```diff
  inf_leave:
    RESTORE();
    if (INFLATE_NEED_UPDATEWINDOW(strm) &&
            (state->wsize || (out != strm->avail_out && state->mode < BAD &&
                 (state->mode < CHECK || flush != Z_FINISH)))) {
        /* update sliding window with respective checksum if not in "raw" mode */
        if (updatewindow(strm, strm->next_out, out - strm->avail_out, state->wrap & 4)) {
            state->mode = MEM;
            return Z_MEM_ERROR;
        }
-   }
+   } else if (!state->wsize && flush == Z_FINISH) {
+       inf_chksum(strm, put - (out - strm->avail_out), (out - strm->avail_out));
+   }
```
Vars: `state->wsize = 0, flush = Z_FINISH, state->mode = CHECK`

In this use case `Z_FINISH` is being used which according to the docs:
image

Normally, `inflate` might already have a window if it had been previously called with another flush type. But in `libgit2` case it is very short length and calling it with `Z_FINISH`, a window doesn't get created or used.

KungFuJesus:
> This also seems to work although it can probably be better some how:
> 
> ```diff
>   inf_leave:
>     RESTORE();
>     if (INFLATE_NEED_UPDATEWINDOW(strm) &&
>             (state->wsize || (out != strm->avail_out && state->mode < BAD &&
>                  (state->mode < CHECK || flush != Z_FINISH)))) {
>         /* update sliding window with respective checksum if not in "raw" mode */
>         if (updatewindow(strm, strm->next_out, out - strm->avail_out, state->wrap & 4)) {
>             state->mode = MEM;
>             return Z_MEM_ERROR;
>         }
> -   }
> +   } else if (!state->wsize && flush == Z_FINISH) {
> +       inf_chksum(strm, put - (out - strm->avail_out), (out - strm->avail_out));
> +   }
> ```
> 
> Vars: `state->wsize = 0, flush = Z_FINISH, state->mode = CHECK`
> 
> In this use case `Z_FINISH` is being used which according to the docs: image
> 
> Normally, `inflate` might already have a window if it had been previously called with another flush type. But in `libgit2` case it is very short length and calling it with `Z_FINISH`, a window doesn't get created or used.

That solution is a lot cleaner assuming it captures every instance of this.  I was hoping to find a solution more like that but couldn't bring myself to find the right spot and conditions to inject that extra forced checksum.  So the right conditions are specifically if the user has specified Z_FINISH, bypassing the window creation entirely and effectively it's moved to the end stage "check" state

KungFuJesus:
I like @nmoinvaz's solution much better, way less of a hack, way fewer lines, fewer comparisons. @nmoinvaz can you file a PR with your minideflate changes and that simple reproducer as a test case?

inflate.c
KungFuJesus: Don't think we need this, I went with your latter suggested approach which rendered it unnecessary.

nmoinvaz:
Yes, I can create a PR with my `minideflate` use case. I looked putting my suggested logic in `CHECK` mode instead, but `CHECK` mode bails early in `NEEDSBITS` with `Z_BUF_ERROR`. So I think at the bottom of the function like I originally suggested is probably the best place for the logic.

inflate.c
```diff
@@ -1120,6 +1120,8 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
             state->mode = MEM;
             return Z_MEM_ERROR;
         }
+    } else if (INFLATE_NEED_CHECKSUM(strm) && !state->wsize && flush == Z_FINISH) {
```
nmoinvaz: I am wondering if it should only be done if also `&& err == Z_BUF_ERROR` -- that is, only when `NEEDBITS()` bails early. I'm not 100% sure.

KungFuJesus: Are there conditions where we'd hit this and err != Z_BUF_ERROR?  I think the spirit of the original code had a checksum here a bit less conditionally (the conditions that applied before were basically if (out)).

nmoinvaz: The reason why I suggest it is: because the only test that fails is when CHECK bails early in NEEDBITS. When CHECK doesn't bail early I assume it will do the chksum in that case statement...?

mtl1979: I think adding check `&& err == Z_BUF_ERROR` might be warranted, but we will need another test for that to be 100% sure... Returning `Z_BUF_ERROR` tells me the stream is truncated in some way and that itself should be enough to not silently ignore the case even if that means there is another bug somewhere else.

inflate.c
```diff
@@ -1120,6 +1120,8 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
...
+        inf_chksum(strm, put - (out - strm->avail_out), (out - strm->avail_out));
```
mtl1979: Using same expression twice hurts readability... even if using temporary variable means adding one more line, this is still quite short delta.

mtl1979 (changes requested):
Needs further testing...

inflate.c
nmoinvaz: To check the other scenario where it doesn't bail early, just don't send `-r 25` to minideflate from my PR.

KungFuJesus: Sorry, was doing some traveling for the holiday.  On it, now.

inflate.c
KungFuJesus: I'm not entirely sure what you mean, none of the conditions being checked here overlap.  It just so happens that NEED_UPDATEWINDOW and NEED_CHECKSUM are the same in most cases, but they wouldn't necessarily have to be.  And inf_chksum does something very different from updatewindow.

inflate.c
KungFuJesus: k, pushed update.

inflate.c
KungFuJesus: Are you referring to the pointer arithmetic?  I guess that can be baked out.  One second.

KungFuJesus: I think that's what you mean, let me know if not.  Resolving for now.

inflate.c
```diff
@@ -1112,14 +1112,18 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
      */
   inf_leave:
     RESTORE();
+    uint32_t checkBytes = out - strm->avail_out;
```
nmoinvaz: I think maybe you meant to do snakecase instead of camelcase.

KungFuJesus: ah, sorry camel is my personal preference that's hard to break. Fixing now.

nmoinvaz: I use it all the time too because like 70% of what I do is in JavaScript these days.

inflate.c
```diff
@@ -1112,14 +1112,18 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
...
             (state->wsize || (out != strm->avail_out && state->mode < BAD &&
                  (state->mode < CHECK || flush != Z_FINISH)))) {
         /* update sliding window with respective checksum if not in "raw" mode */
-        if (updatewindow(strm, strm->next_out, out - strm->avail_out, state->wrap & 4)) {
+        if (updatewindow(strm, strm->next_out, check_bytes, state->wrap & 4)) {
             state->mode = MEM;
             return Z_MEM_ERROR;
         }
+    } else if (INFLATE_NEED_CHECKSUM(strm) && !state->wsize &&
+               flush == Z_FINISH && ret == Z_BUF_ERROR) {
```
nmoinvaz: Shoot, I just realized checking for `ret == Z_BUF_ERROR` at this place will not work, because `PULLBYTE` (called from `NEEDBITS`) only sets `ret = Z_BUF_ERROR;` in `infback.c`, which is weird. The `PULLBYTE` defined in `inflate.c` does not do that.

nmoinvaz: I wonder now if the check should be moved into `ret = Z_BUF_ERROR` below. 
```diff
if (((in == 0 && out == 0) || flush == Z_FINISH) && ret == Z_OK) {
+    /* when no sliding window is used, hash the output bytes when CHECK state does not */
+    if (INFLATE_NEED_CHECKSUM(strm) && !state->wsize && flush == Z_FINISH) {
+        inf_chksum(strm, put - check_bytes, check_bytes);
+    }
    ret = Z_BUF_ERROR;
}
```

KungFuJesus: Yeah, I think it might make more sense there.

inflate.c
nmoinvaz: FYI, I added the other check to the test cases in #1603.

nmoinvaz:
LGTM, now we just need to rebase after #1603 is merged.

inflate.c
mtl1979: I think having both tests is better if just to catch potential regressions... We can't just test existing failures (before fixing them)...

inflate.c
mtl1979: I meant the `(out - strm->avail_out)` part, which you did fix.... I know compiler is smart enough to do duplicate code elimination even on lower optimisation levels, but we really shouldn't rely on it.

Dead2:
Rebase please 👍

Dead2 (approved):
I like this elegant solution 👍

KungFuJesus:
Guessing we can probably ignore that failure to upload issue?

KungFuJesus:
Rerolling the dice anyway for that green check mark.

Dead2:
Looking good, all tests passing. I will probably roll out a hotfix release with this within a day or two, unless something crops up.

mtl1979 (approved):
LGTM

PR #1603, opened 2023-11-22
Added unit test for inflate with Z_FINISH and no window.

nmoinvaz:
This test decompresses a small git pack object with the text: "alone in the dark". By providing inflate with `Z_FINISH`, it should not use a sliding window according to zlib docs. We purposefully limit the read buffer size to trigger `CHECK` mode to bail early.


nmoinvaz:
This PR should fail and PR #1602 should fix it.

Dead2 (approved):
LGTM

mtl1979:
Looks like harakiri code, but for the purpose of simplifying the libgit2 failure, this does require no external code. I'm not sure it is good idea to separate these as two PRs due to chicken and egg dilemma.

nmoinvaz:
The other PR can be rebased after this is merged.

PR #1604, opened 2023-11-26
Prevent tests writing into the source directory

phprus:
Unit test #1603 creates a file in the source directory.

### DIFF
--- a/test/cmake/compress-and-verify.cmake
+++ b/test/cmake/compress-and-verify.cmake
@@ -138,7 +138,11 @@ set(COMPRESS_COMMAND ${COMPRESS_TARGET} ${COMPRESS_ARGS})
 set(INPUT_FILE ${OUTPUT_BASE})
 
 # Make CMake copy and rename file in one operation
-configure_file(${INPUT} ${INPUT_FILE} COPYONLY)
+# The copied file permissions is standard 644 (-rw-r--r--)
+if(NOT CMAKE_VERSION VERSION_LESS "3.19")
+    set(CONFIGURE_NO_SOURCE_PERMISSIONS NO_SOURCE_PERMISSIONS)
+endif()
+configure_file(${INPUT} ${INPUT_FILE} COPYONLY ${CONFIGURE_NO_SOURCE_PERMISSIONS})
 
 message(STATUS "Compress ${COMPRESS_COMMAND}")
 message(STATUS "  Source file: ${INPUT}")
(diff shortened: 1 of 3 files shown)


mtl1979:
We require cmake 3.5.1 or later, so any features that require newer cmake version -- without fallback -- can't be used. 

Several CI tests need to be rerun as either them failed (for unknown reason) or were skipped due to other checks failing.

phprus:
> We require cmake 3.5.1 or later, so any features that require newer cmake version -- without fallback -- can't be used.

Fallback - do not build zlib-ng with the source directory in read-only mode (anyway it's impossible now).
CMake version condition is only for OSB CI tasks.
Or I could remove "Run OSB tests with readonly source directory." commit.

> Several CI tests need to be rerun as either them failed (for unknown reason) or were skipped due to other checks failing.

It seems I don't have permission to restart CI.

phprus:
CI error:

``
[2023-11-27T14:16:48.274Z] ['error'] There was an error running the uploader: Error uploading to https://codecov.io: Error: There was an error fetching the storage URL during POST: 503 - upstream connect error or disconnect/reset before headers. reset reason: connection failure
``

test/cmake/compress-and-verify.cmake
```diff
@@ -138,7 +138,10 @@ set(COMPRESS_COMMAND ${COMPRESS_TARGET} ${COMPRESS_ARGS})
 set(INPUT_FILE ${OUTPUT_BASE})
 
 # Make CMake copy and rename file in one operation
```
Dead2, line 140: This comment is missing info and is a little misleading.

phprus: I haven't changed this comment.
Or do I need to move it before ``configure_file``?

Dead2: You changed the content that the comment describes, so the comment should be updated to describe the new situation.

phprus: Comment is updated.

mtl1979:
> > We require cmake 3.5.1 or later, so any features that require newer cmake version -- without fallback -- can't be used.
> 
> Fallback - do not build zlib-ng with the source directory in read-only mode (anyway it's impossible now).
> CMake version condition is only for OSB CI tasks.
> Or I could remove "Run OSB tests with readonly source directory." commit.


Sometimes it's hard to decide if two related changes should be in same or separate pull requests. In this case there is no controversy in improving how a test is executed, but we already require that source directory is writable, so requiring that it is not written to is backwards.

nmoinvaz (approved):

Dead2 (approved):
LGTM

Issue #1608, opened 2023-12-10
functable overhead


haampie:
I traced zlib-ng with perf and noticed that the `quick_insert_string` function call is the 'hottest' couple instructions in `deflate_medium` when compiled with GCC 13.2.0:

![image](https://github.com/zlib-ng/zlib-ng/assets/194764/65754326-81a4-48ac-b49e-1d8f7d6a433e)

Replacing that function pointer directly with `quick_insert_string_sse42` makes the overhead go away.

Makes me wonder: has zlib-ng considered using ifunc (gcc extension) instead of function pointers? Is there a way to make the dispatch static instead of dynamic when targeting a specific architecture?

Edit: ah, it's actually dynamic cause the implementation depends on certain parameters... In any case, looks like at runtime the function is resolved every call even if it's effectively constant.

phprus:
Try removing ``Z_TLS`` from the ``functable`` definition to avoid tls overhead.
This should be safe to execute on x86(_64), but may generate ThreadSanitizer warnings.

haampie:
That fixes it yeah. I don't think there's any significant gain though, would have to do a proper benchmark.

nmoinvaz:
We could do what we did with `inflate_fast` (in that it uses a template and there is a separate version of it for each instruction set) but do it for all the deflate algorithms as well.

nmoinvaz:
> Replacing that function pointer directly with quick_insert_string_sse42 makes the overhead go away.

What does the overhead equate to in terms of performance?

phprus:
Which is greater: TLS overhead or overhead of calling a function by pointer?

phprus:
If the main overhead is in TLS, then I have one idea on how to remove it for x86 and arm.

KungFuJesus:
If I had to guess it'd probably be TLS overhead. Initializing the tables before an lwp fork would certainly fix that, I would think.

TLS requires this weird indirection through the segment registers on x86.

phprus:
``functable`` without TLS:
https://github.com/phprus/zlib-ng/tree/functable-atomic-1

Requires all CPUs cores to support the same instruction set.

nmoinvaz:
Your branch changes look interesting.

KungFuJesus:
[...]

PR #1609, opened 2023-12-12
``functable`` without TLS

phprus:
Previous discussion in issue #1608.

### DIFF
--- a/functable.c
+++ b/functable.c
@@ -11,6 +11,34 @@
 #include "functable.h"
 #include "cpu_features.h"
 
+#if defined(_MSC_VER)
+#  include <intrin.h>
+#endif
+
+/* Platform has pointer size atomic store */
+#if defined(__GNUC__) || defined(__clang__)
+#  define FUNCTABLE_ASSIGN(VAR, FUNC_NAME) \
+    __atomic_store(&(functable.FUNC_NAME), &(VAR.FUNC_NAME), __ATOMIC_SEQ_CST)
+#  define FUNCTABLE_BARRIER() __atomic_thread_fence(__ATOMIC_SEQ_CST)
+#elif defined(_MSC_VER)
+#  define FUNCTABLE_ASSIGN(VAR, FUNC_NAME) \
+    _InterlockedExchangePointer((void * volatile *)&(functable.FUNC_NAME), (void *)(VAR.FUNC_NAME))
+#  if defined(_M_ARM) || defined(_M_ARM64)
+#    define FUNCTABLE_BARRIER() do { \
+    _ReadWriteBarrier();  \
+    __dmb(0xB); /* _ARM_BARRIER_ISH */ \
+    _ReadWriteBarrier(); \
+} while (0)
+#  else
+#    define FUNCTABLE_BARRIER() _ReadWriteBarrier()
+#  endif
+#else
+#  warning Unable to detect atomic intrinsic support.
+#  define FUNCTABLE_ASSIGN(VAR, FUNC_NAME) \
+    *((void * volatile *)&(functable.FUNC_NAME)) = (void *)(VAR.FUNC_NAME)
+#  define FUNCTABLE_BARRIER() do { /* Empty */ } while (0)
+#endif
+
 static void force_init_empty(void) {
     // empty
 }
@@ -239,24 +267,27 @@ static void init_functable(void) {
 #endif
 
     // Assign function pointers individually for atomic operation
-    functable.force_init = ft.force_init;
-    functable.adler32 = ft.adler32;
-    functable.adler32_fold_copy = ft.adler32_fold_copy;
-    functable.chunkmemset_safe = ft.chunkmemset_safe;
-    functable.chunksize = ft.chunksize;
-    functable.compare256 = ft.compare256;
-    functable.crc32 = ft.crc32;
-    functable.crc32_fold = ft.crc32_fold;
-    functable.crc32_fold_copy = ft.crc32_fold_copy;
-    functable.crc32_fold_final = ft.crc32_fold_final;
-    functable.crc32_fold_reset = ft.crc32_fold_reset;
-    functable.inflate_fast = ft.inflate_fast;
-    functable.insert_string = ft.insert_string;
-    functable.longest_match = ft.longest_match;
-    functable.longest_match_slow = ft.longest_match_slow;
-    functable.quick_insert_string = ft.quick_insert_string;
-    functable.slide_hash = ft.slide_hash;
-    functable.update_hash = ft.update_hash;
+    FUNCTABLE_ASSIGN(ft, force_init);
+    FUNCTABLE_ASSIGN(ft, adler32);
+    FUNCTABLE_ASSIGN(ft, adler32_fold_copy);
+    FUNCTABLE_ASSIGN(ft, chunkmemset_safe);
+    FUNCTABLE_ASSIGN(ft, chunksize);
+    FUNCTABLE_ASSIGN(ft, compare256);
+    FUNCTABLE_ASSIGN(ft, crc32);
+    FUNCTABLE_ASSIGN(ft, crc32_fold);
+    FUNCTABLE_ASSIGN(ft, crc32_fold_copy);
+    FUNCTABLE_ASSIGN(ft, crc32_fold_final);
+    FUNCTABLE_ASSIGN(ft, crc32_fold_reset);
+    FUNCTABLE_ASSIGN(ft, inflate_fast);
+    FUNCTABLE_ASSIGN(ft, insert_string);
+    FUNCTABLE_ASSIGN(ft, longest_match);
+    FUNCTABLE_ASSIGN(ft, longest_match_slow);
+    FUNCTABLE_ASSIGN(ft, quick_insert_string);
+    FUNCTABLE_ASSIGN(ft, slide_hash);
+    FUNCTABLE_ASSIGN(ft, update_hash);
+
+    // Memory barrier for weak memory order CPUs
+    FUNCTABLE_BARRIER();
 }
 
 /* stub functions */
@@ -350,7 +381,7 @@ static uint32_t update_hash_stub(deflate_state* const s, uint32_t h, uint32_t va
 }
 
 /* functable init */
-Z_INTERNAL Z_TLS struct functable_s functable = {
+Z_INTERNAL struct functable_s functable = {
     force_init_stub,
     adler32_stub,
     adler32_fold_copy_stub,
--- a/functable.h
+++ b/functable.h
@@ -37,6 +37,6 @@ struct functable_s {
     uint32_t (* update_hash)        (deflate_state *const s, uint32_t h, uint32_t val);
 };
 
-Z_INTERNAL extern Z_TLS struct functable_s functable;
+Z_INTERNAL extern struct functable_s functable;
 
 #endif
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -417,21 +417,6 @@ if(HAVE_ATTRIBUTE_ALIGNED)
 endif()
 
 #
-# check for _Thread_local() support in the compiler
-#
-check_c_source_compiles(
-    "_Thread_local int test;
-    int main(void) {
-        (void)test;
-        return 0;
-    }"
-    HAVE_THREAD_LOCAL
-)
-if(HAVE_THREAD_LOCAL)
-    add_definitions(-DHAVE_THREAD_LOCAL)
-endif()
-
-#
 # check for __builtin_ctz() support in the compiler
 #
 check_c_source_compiles(
--- a/configure
+++ b/configure
@@ -973,22 +973,6 @@ else
     echo "Checking for attribute(aligned) ... No." | tee -a configure.log
 fi
 
-# Check for _Thread_local support in compiler
-cat > $test.c << EOF
-_Thread_local int test;
-int main(void) {
-    (void)test;
-    return 0;
-}
-EOF
-if try ${CC} ${CFLAGS} $test.c $LDSHAREDLIBC; then
-    echo "Checking for _Thread_local ... Yes." | tee -a configure.log
-    CFLAGS="$CFLAGS -DHAVE_THREAD_LOCAL"
-    SFLAGS="$SFLAGS -DHAVE_THREAD_LOCAL"
-else
-    echo "Checking for _Thread_local ... No." | tee -a configure.log
-fi
-
 # Check for __builtin_ctz() support in compiler
 cat > $test.c << EOF
 long f(unsigned int x) { return __builtin_ctz(x); }
--- a/zbuild.h
+++ b/zbuild.h
@@ -31,20 +31,6 @@
 #  endif
 #endif
 
-/* Determine compiler support for TLS */
-#ifndef Z_TLS
-#  ifdef HAVE_THREAD_LOCAL
-#    define Z_TLS _Thread_local
-#  elif defined(__GNUC__) || defined(__SUNPRO_C)
-#    define Z_TLS __thread
-#  elif defined(_WIN32) && (defined(_MSC_VER) || defined(__ICL))
-#    define Z_TLS __declspec(thread)
-#  else
-#    warning Unable to detect Thread Local Storage support.
-#    define Z_TLS
-#  endif
-#endif
-
 #ifndef Z_HAS_ATTRIBUTE
 #  if defined(__has_attribute)
 #    define Z_HAS_ATTRIBUTE(a) __has_attribute(a)


KungFuJesus:
I _think_ powerpc has 32 bit atomics but not necessarily 64 bit.  This is just a vague recollection, though.

Hmm, quick bit of googling suggests that 32 bit powerpc didn't have 64 bit atomics, so, maybe not a big deal?  Your PR also relies on the implicit 64 bit writes being untearable, rather than something explicit.  Maybe we could use explicit atomic integer storage for the pointers?  The compiler is likely to convert them to plain old stores on architectures that do have this implicit behavior (at least it does with C++'s std::atomic, I would assume as much for the compiler intrinsics in C).

KungFuJesus:
https://en.cppreference.com/w/c/atomic/atomic_store

Requires C11, but a possible solution.

There's also these, though they are more GCC specific.  Maybe we could wrap them in something to shim for everything we currently support: 
https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html

Uglier, but, these sorts of things guarantee correct behavior for weaker memory ordering and more importantly, torn stores.  We'd only need to wrap the stores, the loads of course don't need any sync semantics of any kind because even if not cache coherent, it's either valid or it's null and if it's null, it can just reassign the pointer to the same address on its own.

phprus:
@KungFuJesus 
Not 64bit, only pointer size atomic write is need (4 byte on 32bit platform or 8 byte on 64bit platform).

C11 atomic require "[atomic object types](https://en.cppreference.com/w/c/language/atomic)". For this reason I use the builtin functions of gcc and clang.

phprus:
``functable`` is initialized by stub functions.
Stubs select platform-specific functions, write ``functable``, and call the selected platform-specific function to avoid recursion if weak-memory CPU reorder load and store. Stubs do not read ``functable``.

For this PR, the platform must provide single guarantee: Atomic store of pointer-size variable.

functable.c
```diff
@@ -157,190 +160,195 @@ static void init_functable(void) {
...
+    FUNCTABLE_ASSIGN(ft, insert_string);
+    FUNCTABLE_ASSIGN(ft, longest_match);
+    FUNCTABLE_ASSIGN(ft, longest_match_slow);
+    FUNCTABLE_ASSIGN(ft, quick_insert_string);
+    FUNCTABLE_ASSIGN(ft, slide_hash);
+    FUNCTABLE_ASSIGN(ft, update_hash);
 }
 
+#define STUB_BODY(RET, FUNC_NAME, ...) \
+    struct functable_s ft; \
```
nmoinvaz: Should this be `static`? Am I missing something, but wouldn't this evaluate the entire `functable` for each function?

phprus: No.
A stub may be called more than once only if the first call of versioned functions is parallel. To avoid race we need local variable.

KungFuJesus:
> @KungFuJesus Not 64bit, only pointer size atomic write is need (4 byte on 32bit platform or 8 byte on 64bit platform).
> 
> C11 atomic require "[atomic object types](https://en.cppreference.com/w/c/language/atomic)". For this reason I use the builtin functions of gcc and clang.

Right, it doesn't much matter on ppc32 if your pointer is 32 bit.  The C11 bits are portable, at least to C11 capable compilers, anyway.  The atomic_store macro approach here should do the job just fine, but it would involve wrapping it seven ways to sunday for other platforms. MSVC, GCC, and Clang covers a lot, but there's also Solaris Studio, ICC, XL, Open64, etc.  Admittedly obscure, but if we _could_ rely on C11 for those, it might be multiple birds with one stone.

phprus:
> Admittedly obscure, but if we _could_ rely on C11 for those, it might be multiple birds with one stone.

Quick reply: No.
C11 atomic require special type. Using an C11 atomic pointer will require rewriting all versioned function calls to atomic_load the pointer and call the function at that pointer (overhead and unnecessary atomic ops).
For this PR we need **hardware** guarantees, not compiler guarantees.

If ppc32 guarantees 32-bit write atomicity and ppc64 guarantees 64-bit write atomicity, then PPC support can be added.

KungFuJesus:
Oh sorry I followed the link and see what you're getting at.  We'd need to decorate everything with that atomic keyword, which would implicitly leverage the atomic loads.  

> If ppc32 guarantees 32-bit write atomicity and ppc64 guarantees 64-bit write atomicity, then PPC support can be added.

I _think_ that it does.  

> For this PR we need hardware guarantees, not compiler guarantees.

Right, but, with the compiler guarantees, the hardware atomicity would come with it.  The compiler should be doing the right thing under the hood (for most ISAs, nothing at all, for others, using whatever atomic store is available with cache aware compare and exchanges / mutexes if needed).  The GCC and MSVC paths you have in that macro do, at the cost of only supporting 3 compilers.  I agree though that we should be able to avoid the atomic load, the C11 approach might make that difficult.

phprus:
> The GCC and MSVC paths you have in that macro do, at the cost of only supporting 3 compilers.

GCC-compatible, Clang-compatible and MSVC. This is most of modern compilers.
For example, IBM XL C/C++ compiler defines the ``__clang__`` macro (https://www.ibm.com/docs/ru/xl-c-and-cpp-linux/16.1.0?topic=macros-identify-xl-cc-compiler).

KungFuJesus:
No doubt GCCisms are widely supported. I can think of one tiny minor contribution I made for something that didn't:
https://github.com/libuv/libuv/commit/ef47e8b2127c45ee58cae6b6f8ace5d6fde695b2

KungFuJesus:
I'm also wondering, however difficult, if we could write a test case that tries to induce a torn store to make sure.  I have SPARC, PowerPC, and ARM hardware I can test against for it.  

Unfortunately reliably producing a race with UB is probably impossible to do deterministically.

phprus:
CI errors:
```
[2023-12-12T19:25:57.298Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 502 -
```

KungFuJesus:
> CI errors:
> 
> ```
> [2023-12-12T19:25:57.298Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 502 -
> ```

Yep, been getting that one a lot recently.  Trying again usually resolves it.

phprus:
Converted to draft.
I need to fix one inaccuracy in macros.

functable.c
```diff
@@ -3,6 +3,10 @@
  * For conditions of distribution and use, see copyright notice in zlib.h
  */
 
+#if defined(_MSC_VER)
+#  include <windows.h>
```
mtl1979: Including `windows.h` without limiting what other headers it can include will slow down compiling a lot. At minimum, `WIN32_LEAN_AND_MEAN` and `NOMINMAX` should be defined before including `windows.h`. It should also be the very last header to be included to avoid issues where subsequent headers try to define same macros with incompatible contents. This includes for example `min` and `max` macros that are normally not defined by default in C code, but are defined in zlib-ng.

phprus: Done

Dead2:
I did a few quick benchmarks just to make sure there are no surprises.

On x86-64 it shows a 0.3% average compression speedup, practically no change for decompression.
On Aarch64 it shows a 0.13% compression speedup, well inside margin of error. Also almost a 1% decompression speedup.

Since these speedups are marginal and the tests were not extensive, the most interesting part is that they are not slower and that they always seem to err on the side of faster. I'd say that bodes well for this PR.
I also did a couple variations to provoke different cache alignments, they too turned out to be similarly faster overall.

```
## Baseline 5af152a6a377b45619c8860e2a25c68fa2199ed9  x86-64
 1     44.409%      1.722/1.741/1.753/0.011        0.435/0.451/0.459/0.008       94,127,290
 2     35.518%      2.893/2.906/2.917/0.008        0.445/0.450/0.460/0.005       75,282,961
 3     33.882%      3.430/3.471/3.499/0.024        0.414/0.424/0.433/0.007       71,816,478
 4     33.174%      3.761/3.792/3.827/0.024        0.405/0.419/0.427/0.008       70,315,668
 5     32.660%      4.119/4.177/4.255/0.045        0.402/0.409/0.417/0.006       69,225,542
 6     32.508%      4.657/4.724/4.778/0.047        0.398/0.402/0.405/0.003       68,902,222
 7     32.255%      6.033/6.122/6.195/0.064        0.389/0.405/0.411/0.007       68,366,800
 8     32.167%      8.751/8.851/8.966/0.082        0.401/0.410/0.416/0.005       68,180,776
 9     31.887%   12.260/12.364/12.423/0.050        0.393/0.401/0.409/0.007       67,586,442

 avg1  34.273%                        5.350                          0.419
 tot                                385.190                         30.175      653,804,179

 145452    1384       8  146844   23d9c libz-ng.so.2

## PR #1609
 1     44.409%      1.695/1.706/1.716/0.007        0.439/0.448/0.454/0.005       94,127,290
 2     35.518%      2.840/2.851/2.858/0.006        0.439/0.450/0.456/0.005       75,282,961
 3     33.882%      3.392/3.432/3.462/0.023        0.415/0.425/0.430/0.005       71,816,478
 4     33.174%      3.750/3.782/3.838/0.034        0.412/0.418/0.424/0.004       70,315,668
 5     32.660%      4.092/4.131/4.163/0.022        0.402/0.408/0.418/0.007       69,225,542
 6     32.508%      4.587/4.647/4.696/0.041        0.390/0.405/0.414/0.008       68,902,222
 7     32.255%      6.104/6.161/6.226/0.039        0.399/0.408/0.415/0.005       68,366,800
 8     32.167%      8.734/8.854/8.962/0.076        0.397/0.404/0.412/0.005       68,180,776
 9     31.887%   12.329/12.444/12.666/0.115        0.393/0.405/0.412/0.006       67,586,442

 avg1  34.273%                        5.334                          0.419
 tot                                384.062                         30.163      653,804,179

 145444    1576       8  147028   23e54 libz-ng.so.2

```

```
## Baseline 5af152a6a377b45619c8860e2a25c68fa2199ed9  aarch64
 0    100.008%      0.012/0.021/0.024/0.005        0.028/0.035/0.040/0.004       15,737,543
 1     54.165%      0.473/0.514/0.544/0.027        0.117/0.139/0.159/0.016        8,523,602
 2     43.877%      0.865/0.885/0.903/0.012        0.123/0.168/0.204/0.031        6,904,649
 3     42.389%      1.146/1.175/1.188/0.015        0.129/0.148/0.173/0.014        6,670,476
 4     41.647%      1.285/1.305/1.317/0.011        0.111/0.131/0.163/0.017        6,553,763
 5     41.217%      1.409/1.435/1.446/0.014        0.125/0.135/0.141/0.006        6,485,981
 6     41.037%      1.645/1.666/1.680/0.012        0.116/0.134/0.148/0.012        6,457,780
 7     40.778%      2.096/2.132/2.150/0.018        0.119/0.132/0.143/0.011        6,416,924
 8     40.704%      2.605/2.627/2.643/0.014        0.107/0.121/0.149/0.015        6,405,244
 9     40.409%      3.171/3.188/3.201/0.011        0.120/0.132/0.144/0.011        6,358,951

 avg1  48.623%                        1.495                          0.128
 avg2  54.026%                        1.661                          0.142
 tot                                119.582                         10.202       76,514,913

 119449    1616       8  121073   1d8f1 libz-ng.so.2

## PR #1609
 0    100.008%      0.008/0.019/0.024/0.006        0.016/0.031/0.040/0.008       15,737,543
 1     54.165%      0.494/0.524/0.538/0.015        0.101/0.130/0.163/0.021        8,523,602
 2     43.877%      0.862/0.880/0.899/0.012        0.119/0.159/0.186/0.026        6,904,649
 3     42.389%      1.156/1.173/1.185/0.011        0.103/0.139/0.161/0.018        6,670,476
 4     41.647%      1.286/1.305/1.316/0.010        0.104/0.127/0.146/0.017        6,553,763
 5     41.217%      1.384/1.413/1.432/0.017        0.120/0.137/0.161/0.012        6,485,981
 6     41.037%      1.650/1.673/1.695/0.015        0.129/0.138/0.144/0.006        6,457,780
 7     40.778%      2.104/2.127/2.155/0.014        0.109/0.136/0.154/0.016        6,416,924
 8     40.704%      2.600/2.618/2.632/0.013        0.123/0.131/0.141/0.006        6,405,244
 9     40.409%      3.176/3.197/3.209/0.011        0.112/0.135/0.160/0.018        6,358,951

 avg1  48.623%                        1.493                          0.126
 avg2  54.026%                        1.659                          0.140
 tot                                119.428                         10.107       76,514,913

 118281    1544       8  119833   1d419 libz-ng.so.2

```

functable.h
```diff
@@ -36,6 +36,26 @@ struct functable_s {
     uint32_t (* update_hash)        (deflate_state *const s, uint32_t h, uint32_t val);
 };
 
-Z_INTERNAL extern Z_TLS struct functable_s functable;
+/* Platform and compiler has pointer size atomic store */
+#if defined(X86_FEATURES) || defined(ARM_FEATURES) \
+    || defined(PPC_FEATURES) || defined(POWER_FEATURES) \
```
nmoinvaz: I looked in our code base but can't seem to find any `#ifdef` where we have `||` at the beginning of the line. I think it should go at the end of the previous line.

phprus: Done

phprus:
All modern architectures implement atomic assignment for pointer.
We can remove ``Z_TLS`` completely.

phprus:
Rebased.

phprus:
CI error:
```
[2023-12-15T18:04:14.751Z] ['verbose'] The error stack is: Error: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 502
```

Again...

nmoinvaz:
Unfortunately there is not much we can do about that, except report it to Codecov.

Dead2:
Re-queued the failed run, all clear now.

functable.c
```diff
@@ -157,190 +160,192 @@ static void init_functable(void) {
...
 
+/* Platform has pointer size atomic store */
+#if defined(__GNUC__) || defined(__clang__)
+#  define FUNCTABLE_ASSIGN(VAR, FUNC_NAME) \
+    __atomic_store(&(functable.FUNC_NAME), &(VAR->FUNC_NAME), __ATOMIC_SEQ_CST)
+#elif defined(_MSC_VER)
+#  define FUNCTABLE_ASSIGN(VAR, FUNC_NAME) \
+    _InterlockedExchangePointer((void * volatile *)&(functable.FUNC_NAME), (void *)(VAR->FUNC_NAME))
+#else
+    (void * volatile)(functable.FUNC_NAME) = (void *)(VAR->FUNC_NAME)
```
Dead2: Do you have a reference that confirms that this is safe?

phprus: Sorry, I don't have a public link to the discussion :(

phprus:
New version:

1) Atomic assignment of pointers.
2) After all assignment, a full memory barrier.
3) For unsupported compilers (not MSVC, Clang and GNU compatible), volatile assignment and warning print.

nmoinvaz:
It seems ok to me. It looks like it is just trading an implementation that compiler specific for one that is platform specific.

Dead2 (approved):
LGTM

Dead2:
> https://en.cppreference.com/w/c/atomic/atomic_store
> 
> Requires C11, but a possible solution.

Just a note here, zlib-ng already requires minimum C11 from version 2.1.0 and up (it was upped from C99 to C11 because of `_Thread_local` used in Z_TLS).

So perhaps this is all that is really needed and the selection could be simplified? It is nice to have a fallback in any case though, you never know what some compilers support. But possibly this could also be used for MSVC and M_ARM, if they really do support C11 properly.

Ref: https://github.com/zlib-ng/zlib-ng/wiki

Dead2:
To answer my own question; Yes MSVC supports it, but only added that support in 2022.
https://devblogs.microsoft.com/cppblog/c11-atomics-in-visual-studio-2022-version-17-5-preview-2/

It would be nice if we detected this `stdatomic.h` support and used that if available for all platforms, only falling back to more proprietary solutions if that support is missing. That way every platform will behave as closely to each other as possible, while using a common modern standard, reducing the testing spread and lowers the risk of corner-case problems.

Anyone have thoughts around this?

phprus:
@Dead2
C11 atomic require "atomic object types" (https://en.cppreference.com/w/c/language/atomic).
We will need to mark all function pointers with the ``_Atomic`` keyword.
But, it may add overhead when calling functions by pointers.

pps83:
Finally this TLS nonsense is removed. Thank you [phprus](https://github.com/phprus)!
I've tried to remove it a couple of years ago #1275, and got push back from [mtl1979](https://github.com/mtl1979) based on his believes that this code is needed while I haven't seen any lib with similar run-time cpu detection use TLS.

PR #1610, opened 2023-12-14
Remove duplicate enable tests option

FantasqueX:
There are two identical ZLIB_ENABLE_TESTS option in CMakeLists.txt which means the latter one is meaningless. This issue is introduced in https://github.com/zlib-ng/zlib-ng/commit/ce1a64b5147b874a2dd32920765be333f54e7095 which adds the option to top but forgets to remove the original one.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1205,7 +1205,6 @@ endif()
 # Example binaries
 #============================================================================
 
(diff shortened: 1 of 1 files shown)


nmoinvaz (approved):
Thank you 🙏🏻

Dead2 (approved):
LGTM

PR #1613, opened 2023-12-16
Initialize functable earlier, during DeflateInit

phprus:
``deflate`` captures pointers to stub functions:

1:

https://github.com/zlib-ng/zlib-ng/blob/2baed9faa9437fcd1d0a106da87a540073ad851f/deflate.c#L1125-L1127

Call chain:
```
deflateInit2 -> deflateReset -> lm_init -> lm_set_level
```

If ``deflateInit2`` is called first from zlib-ng, pointers to the stub will be captured.

2:

https://github.com/zlib-ng/zlib-ng/blob/2baed9faa9437fcd1d0a106da87a540073ad851f/deflate_slow.c#L24-L27

This PR should speed up ``deflate``.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -194,6 +194,9 @@ int32_t ZNG_CONDEXPORT PREFIX(deflateInit2)(PREFIX3(stream) *strm, int32_t level
     deflate_state *s;
     int wrap = 1;
 
+    /* Force initialization functable, because deflate captures function pointers from functable. */
+    functable.force_init();
+
     if (strm == NULL)
         return Z_STREAM_ERROR;
 
--- a/functable.c
+++ b/functable.c
@@ -11,6 +11,10 @@
 #include "functable.h"
 #include "cpu_features.h"
 
+static void force_init_empty(void) {
+    // empty
+}
+
 static void init_functable(void) {
     struct functable_s ft;
     struct cpu_features cf;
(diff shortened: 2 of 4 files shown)


Dead2:
This looks like a good idea, especially for applications that does the init before receiving a connection for example, to avoid any extra latency.

But what about doing the same for InflateInit? (a separate commit would be nice though)

phprus:
Inflate functions do not capture function pointers from the ``functable``.
Adding early initialization will not speed up Inflate.

The absence of early initialization in ``deflateInit2`` causes ``init_functable`` to be called multiple times.

Dead2:
> Inflate functions do not capture function pointers from the `functable`. Adding early initialization will not speed up Inflate.

It will not speed up inflate as a whole, but it will move initialization overhead to the init function instead of being part of inflate, and thus would speed up inflate for applications that run inflateinit before they have anything to inflate.

> The absence of early initialization in `deflateInit2` causes `init_functable` to be called multiple times.

Not sure what you mean here..
Edit: Of course I do, I misread this to also be inflate. So I agree there is no bug in inflate as there is in deflate, but I still think it is a good idea to run initialization the same way for both deflate and inflate.

For our native API, perhaps we should at some point export a function to init functable for those that do not use inflate or deflate and just want adler/crc instead. That could reduce latency for their first hash as well, depending on the application.

phprus:
@Dead2
Early initialization is added in ``inflateInit2``.

nmoinvaz (approved):

Dead2 (approved):
LGTM

PR #1617, opened 2023-12-21
Don't attempt ARMv6 detection on AARCH64.

Dead2:
Fixes compile bug #1616 introduced by #1538

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,7 +102,7 @@ set_property(CACHE WITH_SANITIZER PROPERTY STRINGS "Memory" "Address" "Undefined
 if(BASEARCH_ARM_FOUND)
     option(WITH_ACLE "Build with ACLE" ON)
     option(WITH_NEON "Build with NEON intrinsics" ON)
-    option(WITH_ARMV6 "Build with ARMv6 SIMD" ON)
+    cmake_dependent_option(WITH_ARMV6 "Build with ARMv6 SIMD" ON "NOT ARCH STREQUAL \"aarch64\"" OFF)
 elseif(BASEARCH_PPC_FOUND)
     option(WITH_ALTIVEC "Build with AltiVec (VMX) optimisations for PowerPC" ON)
     option(WITH_POWER8 "Build with optimisations for POWER8" ON)
--- a/configure
+++ b/configure
@@ -1604,12 +1604,13 @@ case "${ARCH}" in
 
     # ARM specific optimizations
     arm* | aarch64)
-      case "${ARCH}" in
+        case "${ARCH}" in
             arm*)
                 [ ! -z $CROSS_PREFIX ] && QEMU_ARCH=arm
             ;;
             aarch64)
                 [ ! -z $CROSS_PREFIX ] && QEMU_ARCH=aarch64
+                buildarmv6=0
             ;;
         esac
 


CMakeLists.txt
```diff
@@ -102,7 +102,9 @@ set_property(CACHE WITH_SANITIZER PROPERTY STRINGS "Memory" "Address" "Undefined
 if(BASEARCH_ARM_FOUND)
     option(WITH_ACLE "Build with ACLE" ON)
     option(WITH_NEON "Build with NEON intrinsics" ON)
-    option(WITH_ARMV6 "Build with ARMv6 SIMD" ON)
+    if(NOT ARCH STREQUAL "aarch64")
```
nmoinvaz:
```cmake
include(CMakeDependentOption) # put at top
cmake_dependent_option(WITH_ARMV6 "Build with ARMv6 SIMD" ON "NOT ARCH STREQUAL \"aarch64\"" OFF)
```

mtl1979:
I checked output in CMakeError.log and the test for ARMv6 features works correctly, so this patch is incorrect. It's issue with incompatible command-line options. I tested output with just gcc (no extra command-line options) and neither 32-bit or 64-bit compiler enable 32-bit SIMD feature.

Not all compiler targets support `-march=native` correctly even if the compiler accepts it. I tried compiling with `-DWITH_NATIVE_INSTRUCTIONS=ON` using Clang 14 and it did clearly error out saying it's not supported.

I've run the ARMv6 optimized versions on 64-bit ARM and those did actually run without crashing.

I've had weird errors before when trying to compile with clang version that doesn't support the target architecture... Older clang versions could only generate 32-bit binaries for some processor families.

configure
```diff
@@ -1752,20 +1752,22 @@ EOF
...
+                        if test $HAVE_ARMV6_INTRIN -eq 1; then
+                            CFLAGS="${CFLAGS} -DARM_SIMD_INTRIN"
+                            SFLAGS="${SFLAGS} -DARM_SIMD_INTRIN"
+                        fi
 
-                    ARCH_STATIC_OBJS="${ARCH_STATIC_OBJS} slide_hash_armv6.o"
-                    ARCH_SHARED_OBJS="${ARCH_SHARED_OBJS} slide_hash_armv6.lo"
+                        ARCH_STATIC_OBJS="${ARCH_STATIC_OBJS} slide_hash_armv6.o"
+                        ARCH_SHARED_OBJS="${ARCH_SHARED_OBJS} slide_hash_armv6.lo"
+                    fi
```
mtl1979: We can make this patch a lot shorter if we just force `$buildarmv6` as `0` when targeting AArch64.

mtl1979 (changes requested):

CMakeLists.txt
Dead2: Updated 👍

configure
Dead2: Updated 👍

nmoinvaz:
Honestly I just would prefer it to run the checks on the platform anyways. And we should just fix the checks instead. But if you want this way, the it looks good to me.

Dead2:
@nmoinvaz I don't think it is useful to run checks for things that will never be used, feels a bit like bloat.
But I also think the underlying bug with the tests need to be fixed before this goes in.

mtl1979 (approved):
LGTM

PR #1618, opened 2023-12-21
Check if compiler supports -march=native or -mcpu=native.

mtl1979:
* Clang doesn't always support `-march=native`


CMakeLists.txt
```diff
@@ -252,13 +252,17 @@ endif()
...
+        check_c_compiler_flag(-mcpu=native HAVE_MCPU_NATIVE)
+        if(HAVE_MARCH_NATIVE)
             set(NATIVEFLAG "-march=native")
+        elseif(HAVE_MCPU_NATIVE)
+            set(NATIVEFLAG "-mcpu=native")
+        endif()
+        if(NATIVEFLAG)
+            # Apply flag to all source files and compilation checks
+            add_compile_options(${NATIVEFLAG})
         endif()
```
Dead2, line 263: Perhaps have a similar message as below if compiler does not support native.
Might be best to make it a warning so it is more visible. Or perhaps an error, it is not likely native will be used by distributions, so should not break stuff. Either is good.

mtl1979: I thought about adding warning... But then I also thought about reordering the code that it would just fall through and avoid duplication of the warning message... I wasn't sure which solution would be cleaner...

Dead2 (approved):

nmoinvaz (approved):

CMakeLists.txt
```diff
@@ -252,11 +252,16 @@ endif()
 # Set native march/mcpu
 if(WITH_NATIVE_INSTRUCTIONS)
     if(CMAKE_C_COMPILER_ID MATCHES "GNU" OR CMAKE_C_COMPILER_ID MATCHES "Clang")
-        if(BASEARCH_PPC_FOUND)
-            set(NATIVEFLAG "-mcpu=native")
-        else()
+        check_c_compiler_flag(-march=native HAVE_MARCH_NATIVE)
```
nmoinvaz, line 255:
```suggestion
        check_c_compiler_flag(-march=native HAVE_MARCH_NATIVE)
        if(HAVE_MARCH_NATIVE)
            set(NATIVEFLAG "-march=native")
        else()
            check_c_compiler_flag(-mcpu=native HAVE_MCPU_NATIVE)
            if(HAVE_MCPU_NATIVE)
                set(NATIVEFLAG "-mcpu=native")
            endif()
        endif()
```

PR #1562, opened 2023-08-21
detect intrinsics: ensure intrinsics are not optimized out before assembler is run

haampie:
Closes #1559 

Old binutils may have an assembler that does not support vector instructions
generated by `gcc`.

Although it's recommended to use an up-to-date assembler with gcc, it's happens
in practice that new gcc and old binutils are mixed.

The problem with the current configure script is that it runs tests with `-O2`
cflags, which causes the test inside `main() { ... }` to be optimized out
entirely, and the assembler won't see any of the vector instructions, leading
to false positives in tests.

This PR ensures that most configure tests use a separate function, taking and
returning some `__mm*` type, to avoid optimizations, so that the assembler will
error on seeing unsupported instructions.

Alternatively, `-O0` could be passed, but I don't think that is always
sufficient, since some compilers cannot *not* optimize ;p I believe nvhpc is
one of those.

PR #1619, opened 2023-12-22
Make sure uqsub16 mnemonic doesn't get optimized away.

mtl1979:
See #1616.

### DIFF
--- a/cmake/detect-intrinsics.cmake
+++ b/cmake/detect-intrinsics.cmake
@@ -50,7 +50,7 @@ macro(check_armv6_compiler_flag)
             __asm__ __volatile__ ( \"uqsub16 %0, %1, %2\" : \"=r\" (c) : \"r\" (a), \"r\" (b) );
             return (int)c;
         }
-        int main(void) { return 0; }"
+        int main(void) { return f(1,2); }"
         HAVE_ARMV6_INLINE_ASM
     )
     # Check whether compiler supports ARMv6 intrinsics


Dead2 (approved):

nmoinvaz:
Hmm. This is a problem for all checks? Because the user specified -O3. But it is the wrong way to use CMake?

mtl1979:
> Hmm. This is a problem for all checks? Because the user specified -O3. But it is the wrong way to use CMake?

Even if the user abuse cmake in unexpected way, we should handle the case if even remotely possible... There is two major issues here, one is passing `-O3` in CMAKE_C_FLAGS and the other is using `-flto=thin`, which suppress some errors during compiling, essentially expecting that linker issue the error during LTO phase, but LTO phase happens after dead code elimination, thus no error is triggered.

I noticed some tests like ACLE crc32 support are no-op when `-march=native` is used, so those need to be rewritten anyways.

nmoinvaz:
We should go back and take a look at the PR to see why they were changed initially to be a main function that returned 0. 

IMO passing -O3 is wrong usage for configure. They should just be using --config Release or -CMAKE_BUILD_TYPE=Release.

mtl1979:
> We should go back and take a look at the PR to see why they were changed initially to be a main function that returned 0.
> 
> IMO passing -O3 is wrong usage for configure. They should just be using --config Release or -CMAKE_BUILD_TYPE=Release.

I'm not really fan of going back to old PRs... There is a lot of stuff that I wouldn't have approved if I were the project maintainer, but as we all know, @dead2 is...

Dead2:
An interesting thing to note here is that we always disable LTO, unless the user also enabled WITH_NATIVE_INSTRUCTIONS.
That is because we cannot support LTO with optimized functions for specific instruction sets enabled (Because it would make them all use the same instruction sets). Fixing this would mean making the buildsystem much more custom, by ensuring the optimized functions are each in separate compile units AND linked in separately after the core has been linked with LTO. It could easily become a mess. For GCC and Clang it would require covering every instruction-set optimized function with function attribute __target__, instead of command line parameters to select instruction sets per compile unit.

Whether we want to and should support LTO with NativeInst is a valid question at this point.
The downside is that it massively increases the needed test combinations (if we want good arch+compiler coverage to be sure it works and does not break). It will always be a corner case and it is more fragile.
The upside I am uncertain of, I have not done any performance tests with LTO+NativeInst.


Regardless, I think the best thing for zlib-ng would be to fix all the tests to actually call the test function, or put the test right in main() to reduce the size per test slightly.
Some cleanup is also needed, some of the tests should be renamed from `check_X_compiler_flag` to `check_X_intrinsics`.

nmoinvaz:
Here is the previous PR: https://github.com/zlib-ng/zlib-ng/pull/1562

> The problem with the current configure script is that it runs tests with -O2
cflags, which causes the test inside main() { ... } to be optimized out
entirely, and the assembler won't see any of the vector instructions, leading
to false positives in tests.

> This PR ensures that most configure tests use a separate function, taking and
returning some __mm* type, to avoid optimizations, so that the assembler will
error on seeing unsupported instructions.

Dead2:
> I'm not really fan of going back to old PRs... There is a lot of stuff that I wouldn't have approved if I were the project maintainer, but as we all know, @Dead2 is...

Sometimes I need to make compromises and merge changes that are good enough to be built upon further, not all contributors are interested in revisiting a PR multiple times to fix minor things after reviews. Rather than drive developers away, it is better to reach a compromise. It means we merge things that are not ideal, and sometimes I dislike some of it, but it keeps us moving in the right direction.
Don't be afraid to make PRs with improvements, as long as you can make an argument for why it makes things better.

Dead2:
> Here is the previous PR: #1562
> 
> > The problem with the current configure script is that it runs tests with -O2
> > cflags, which causes the test inside main() { ... } to be optimized out
> > entirely, and the assembler won't see any of the vector instructions, leading
> > to false positives in tests.
> 
> > This PR ensures that most configure tests use a separate function, taking and
> > returning some __mm* type, to avoid optimizations, so that the assembler will
> > error on seeing unsupported instructions.

Right, it would perhaps be better to make the compile tests behave like dlls instead of executables. That way the function will always be required, even if nobody ever calls it. Not sure how CMake likes that though.

mtl1979:
The problem I see with `-O3` and `-flto=thin` is that we need to make sure all vector variables we use are somehow tied to return value of `main()` function... This means last vector variable needs to be truncated somehow to fit 32-bit scalar register... The intrinsics that already return value in scalar register can just be coarsed/passed to return statement of `main()`.

Dead2:
Another alternative is of course to disable LTO for all tests (except the LTO test). It is not useful for the tests, and could break tests like it does now.

nmoinvaz:
> Another alternative is of course to disable LTO for all tests (except the LTO test). It is not useful for the tests, and could break tests like it does now.

But how do we do that when `CMAKE_C_FLAGS` is automatically passed into `check_c_source_compiles`?

Dead2:
> > Another alternative is of course to disable LTO for all tests (except the LTO test). It is not useful for the tests, and could break tests like it does now.
> 
> But how do we do that when `CMAKE_C_FLAGS` is automatically passed into `check_c_source_compiles`?

Same way we currently do it when NativeInst is not enabled. See #1622

nmoinvaz:
But since `CMAKE_C_FLAGS` is passed into `check_c_source_compiles` before `CMAKE_REQUIRED_FLAGS`, then passing `-Od` or `-Od` to checks will make no difference right?

Dead2:
> But since `CMAKE_C_FLAGS` is passed into `check_c_source_compiles` before `CMAKE_REQUIRED_FLAGS`, then passing `-Od` or `-Od` to checks will make no difference right?

From where?
It will have effect from anywhere that ends up in either `CMAKE_C_FLAGS` or `CMAKE_REQUIRED_FLAGS`.
But if we for example specify `-O0` in `CMAKE_REQUIRED_FLAGS` (might also be a good idea), then that will override any previous mention of `-O<N>`.

PR #1620, opened 2023-12-22
Add support for __attribute__((__target__(...))) to overcome limitations of -march=native.

mtl1979:
Support for -march=native for AArch64 is still new, so it doesn't detect all CPU extensions.

Implicitly telling gcc/clang to enable "crc" extension when using -march=native will allow compiling functions with ACLE CRC32 intrinsics.

### DIFF
--- a/arch/arm/insert_string_acle.c
+++ b/arch/arm/insert_string_acle.c
@@ -6,9 +6,7 @@
  */
 
 #ifdef ARM_ACLE
-#ifndef _MSC_VER
-#  include <arm_acle.h>
-#endif
+#include "acle_intrins.h"
 #include "../../zbuild.h"
 #include "../../deflate.h"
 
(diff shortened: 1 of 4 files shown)


Dead2:
I think this can more easily be done by not replacing the crc compiler parameter with `-march=native`, but rather prepending `-march=native` to it. That way it will enable it in addition to whatever the native detection enables.

mtl1979:
> I think this can more easily be done by not replacing the crc compiler parameter with `-march=native`, but rather prepending `-march=native` to it. That way it will enable it in addition to whatever the native detection enables.

I thought about that, but it's not any cleaner solution... There was previous discussion about later processors that always have "crc" extension and people wanting to have clean way to make sure compiler use all the available instructions instead of defaulting to "armv8" basic set.

We already removed "native" support in configure, because it was getting ugly, this patch does fix the underlying issue without changing the configuration step of build and solely relying on compiler-specific solution.

Dead2:
Right, with x86 it is a lot easier to specify the wanted instruction (sets for gcc/clang), I forgot the ARM variant is more messy.

arch/arm/insert_string_acle.c
```diff
@@ -12,15 +12,21 @@
 #include "../../zbuild.h"
 #include "../../deflate.h"
 
+#if defined(__aarch64__)
```
Dead2: I am thinking instead of duplicating this block, perhaps it could be placed in `acle_intrins.h` or `arm_features.h`.
Neither is a perfect match, but I think either would be OK.

mtl1979: I think `acle_intrins.h` would be better so it doesn't pollute other unrelated source files unnecessarily...

mtl1979:
> Right, with x86 it is a lot easier to specify the wanted instruction (sets for gcc/clang), I forgot the ARM variant is more messy.

That might be one reason why `-march=native` wasn't supported for AArch64 before.

Dead2 (approved):
LGTM

nmoinvaz (approved):

PR #1622, opened 2023-12-23
Always run CMake compiler feature tests without LTO.

Dead2:
`NOLTOFLAG` is only used by tests, this makes sure it is always set for gcc/clang to avoid running CMake tests for compiler features with LTO.

### DIFF
--- a/cmake/detect-intrinsics.cmake
+++ b/cmake/detect-intrinsics.cmake
@@ -15,7 +15,7 @@ macro(check_acle_compiler_flag)
             endif()
         endif()
         # Check whether compiler supports ACLE flag
-        set(CMAKE_REQUIRED_FLAGS "${ACLEFLAG} ${NATIVEFLAG}")
+        set(CMAKE_REQUIRED_FLAGS "${ACLEFLAG} ${NATIVEFLAG} ${ZNOLTOFLAG}")
         check_c_source_compiles(
             "int main() { return 0; }"
             HAVE_ACLE_FLAG FAIL_REGEX "not supported")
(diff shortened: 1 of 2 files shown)


nmoinvaz:
I'm not seeing `NOLTOFLAG` being used in `check_compiler_... ` checks. Also will this work when `-DCMAKE_C_FLAG="-flto=thin"` in CMake configure? My reading of the cmake docs for [check_c_source_compiles](https://cmake.org/cmake/help/latest/module/CheckCSourceCompiles.html) is that `CMAKE_C_FLAGS` is prepended to `CMAKE_REQUIRED_FLAGS`.

Dead2:
> I'm not seeing `NOLTOFLAG` being used in `check_compiler_... ` checks. Also will this work when `-DCMAKE_C_FLAG="-flto=thin"` in CMake configure? My reading of the cmake docs for [check_c_source_compiles](https://cmake.org/cmake/help/latest/module/CheckCSourceCompiles.html) is that `CMAKE_C_FLAGS` is prepended to `CMAKE_REQUIRED_FLAGS`.

Prepending is ok, as long as we disable it later on the command line.

But you are right, I will revise the patch.

cmake/detect-intrinsics.cmake
```diff
@@ -220,7 +220,7 @@ macro(check_neon_ld4_intrinsics)
         endif()
     endif()
     # Check whether compiler supports loading 4 neon vecs into a register range
-    set(CMAKE_REQUIRED_FLAGS "${NEONFLAG}")
+    set(CMAKE_REQUIRED_FLAGS "${NEONFLAG} ${NATIVEFLAG} ${ZNOLTOFLAG}")
```
Dead2, line 223: I noticed that this test lacked the NATIVEFLAG, and I don't think it should, since the -march is conditional above.

Dead2: IOW: If NativeInst is enabled, this would always fail unless `-march=native` is already part of cflags from outside of CMake.

nmoinvaz (approved):

PR #1623, opened 2023-12-23
Update to GoogleTest 1.12.1

Dead2:
- Update to GoogleTest 1.12.1
  - Fixes a warning from newer CMake versions about GoogleTest 1.11.0 specifying a CMake version that has been deprecated.
- Add a status message when we start the git download.
- Don't disable GoogleTest when CMake is older than 3.12

GoogleTest 1.12.1 requires minimum CMake 3.5 and C++11, this matches nicely with the zlib-ng 2.1.x requirements of CMake 3.5.1 and C11.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -143,7 +143,7 @@ if(WITH_GTEST)
             if (CMAKE_CXX_COMPILER_ID MATCHES "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS_EQUAL 5.3)
                 set(GTEST_TAG release-1.10.0)
             else()
-                set(GTEST_TAG release-1.11.0)
+                set(GTEST_TAG release-1.12.1)
             endif()
         endif()
 
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -148,6 +148,7 @@ if(WITH_GTEST)
         endif()
 
         # Fetch Google test source code from official repository
+        message(STATUS "Git checking out GoogleTest ${GTEST_TAG}")
         FetchContent_Declare(googletest
             GIT_REPOSITORY ${GTEST_REPOSITORY}
             GIT_TAG ${GTEST_TAG})
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -94,16 +94,11 @@ endif()
 
 if(WITH_GTEST OR WITH_BENCHMARKS)
     if(CMAKE_VERSION VERSION_LESS 3.12)
-        message(WARNING "Minimum cmake version of 3.12 not met for GoogleTest or benchmarks!")
-
-        set(WITH_GTEST OFF)
-        set(WITH_GTEST OFF PARENT_SCOPE)
-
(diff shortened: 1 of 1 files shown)


test/CMakeLists.txt
```diff
@@ -94,11 +94,7 @@ endif()
...
 if(WITH_GTEST OR WITH_BENCHMARKS)
     if(CMAKE_VERSION VERSION_LESS 3.12)
-        message(WARNING "Minimum cmake version of 3.12 not met for GoogleTest or benchmarks!")
-
-        set(WITH_GTEST OFF)
-        set(WITH_GTEST OFF PARENT_SCOPE)
-
+        message(WARNING "Minimum cmake version of 3.12 not met for Google benchmark!")
         set(WITH_BENCHMARKS OFF)
         set(WITH_BENCHMARKS OFF PARENT_SCOPE)
```
mtl1979, line 99: This will need to check for `WITH_BENCHMARKS` second time to behave correctly...

test/CMakeLists.txt
```diff
@@ -94,11 +94,7 @@ endif()
...
     else()
```
mtl1979: This `else()` is unnecessary after the change... CXX should be enabled always.

mtl1979:
```
CMake Warning at test/CMakeLists.txt:97 (message):
  Minimum cmake version of 3.12 not met for Google benchmark!


-- Could NOT find GTest (missing: GTEST_LIBRARY GTEST_MAIN_LIBRARY)
CMake Error at test/CMakeLists.txt:127 (include):
  include could not find load file:

    FetchContent


-- Git checking out GoogleTest release-1.12.1
CMake Error at test/CMakeLists.txt:148 (FetchContent_Declare):
  Unknown CMake command "FetchContent_Declare".


-- Configuring incomplete, errors occurred!
```

Dead2:
FetchContent seems to require CMake 3.10.

test/CMakeLists.txt
Dead2: Fixed

test/CMakeLists.txt
Dead2: Not sure what you meant here, but I think it is fixed now, please explain if not.

Dead2:
It should now use a local GTest down to CMake 3.5 and fetch a git copy if 3.10 or newer.

mtl1979:
Still failing with cmake 3.10.2...

Dead2:
@mtl1979 Off by one error..  It was introduced in 3.11, not 3.10.
Are you able to test with 3.11?

mtl1979:
@Dead2 I think there is still a logic flaw later in the file as it did fail to configure even if I changed the version requirement... Basically it doesn't check that it actually found GoogleTest if FetchContent isn't available.

mtl1979:
> @mtl1979 Off by one error.. It was introduced in 3.11, not 3.10. Are you able to test with 3.11?

Configures correctly with cmake 3.11.0, had to build it from sources as there was no working package in either apt or pip repositories.

nmoinvaz (approved):

PR #1627, opened 2023-12-27
Fixed tab character in ARM ACLE uqsub16 assembly.

nmoinvaz:
Tests in `CMake` and `configure` don't contain this character.


Dead2 (approved):
LGTM

PR #1628, opened 2023-12-28
Fix behavior change in inflateCopy that resulted in data corruption in our application

RyeMutt:
Commit 61e181c8ae93dbf56040336179c9954078bd1399 no longer unconditionally nulled out `copy->window = NULL;` resulting in our application failing to decode most data.


nmoinvaz (approved):

Dead2 (approved):
Good catch!

PR #1629, opened 2023-12-29
Cancel previously running CI jobs for PR or branch.

nmoinvaz:
I have used this for other projects (work) with a lot of success. It prevents having to go through and cancel CI actions a lot. Plus, we run a TON of CI instances so it prevents a lot of CI usage when pushing many changes in a row.

For more info see https://stackoverflow.com/questions/70928424/limit-github-action-workflow-concurrency-on-push-and-pull-request


Dead2 (approved):
LGTM

PR #1630, opened 2024-01-03
Move C fallback functions into arch/generic [Part 1]

Dead2:
Due to the scope of this cleanup, this also includes a major cleanup of how we use includefiles, and also ended up cleaning up unneeded includefiles and also some incorrect dependency rules in win32 makefiles.

This PR is limited in scope to doing the following:
- Simplify how we use includes
- Move adler32 fallbacks
- Move crc32 fallbacks

Down the line, this will make it more easily possible to make support for compiling zlib-ng without including the C fallback functions. The grouping of these objects can also (depending on linker behavior) slightly improve instruction cache efficiency for the rest of the code, since they will no longer be interspersed between more important (for performance) code.


arch/arm/adler32_neon.c
```diff
@@ -7,8 +7,8 @@
  */
 #ifdef ARM_NEON
 #include "neon_intrins.h"
-#include "../../zbuild.h"
-#include "../../adler32_p.h"
+#include "zbuild.h"
```
nmoinvaz, line 10: I thought we made the decision awhile ago to use relative paths?

nmoinvaz:
Looks good. Not sure if `crc32_braid_c` files needs to be suffixed with `_c`.

arch/arm/adler32_neon.c
Dead2: We did, and it avoided some problems early on.
But I am not sure it was really such a good thing. It does complicate things.

But if you or someone else have compelling arguments against this change, I can be convinced to keep it like it is.

Dead2:
> Looks good. Not sure if `crc32_braid_c` files needs to be suffixed with `_c`.

We still have `crc32_braid.c` in the root folder, so it avoids a conflict. Also I figured perhaps we just slap on the `_c` for all the files that are only for the C-fallback functions. This makes it easier to spot what files those are, and easier to avoid the mistake of adding non-C-fallback code to those files.

Dead2:
@nmoinvaz Btw, the linter doesn't like TABs in Makefiles, but the Makefiles break without TABs. Perhaps separate rules can be made for those files?

nmoinvaz:
Perhaps change lint.yaml to:
```
git config core.whitespace blank-at-eol
```

arch/arm/adler32_neon.c
nmoinvaz: I have no compelling argument for relative paths.

nmoinvaz (approved):

PR #1632, opened 2024-01-04
Remove tab-in-indent rule from lint workflow

nmoinvaz:
Makefiles require using tabs as indents. So we need to remove this rule. The `blank-at-eol` rule is what we care about the most.


Dead2 (approved):

mtl1979 (approved):
LGTM... We should only care about mixing spaces and tabs and that is enabled by default.

PR #1633, opened 2024-01-04
Remove unneeded pointer for functable.longest_match in deflate_slow

nmoinvaz:


mtl1979:
Extra indirection is only useful if the function pointer inside struct can change after initial assignment... I'm pretty sure that would be unlikely and something we shouldn't even try to support.

Dead2 (approved):
LGTM

nmoinvaz:
Actually this was changed in https://github.com/zlib-ng/zlib-ng/commit/321028cde93dac9f47b9b92ad186c4cbffb77cf2. Since we call `functable.force_init`, perhaps it is no longer needed? @KungFuJesus 

I can reopen PR if desired.

Dead2:
Well, now that we do functable intialization during deflate_init, maybe this workaround is no longer needed. Is there any way to reach this code without functable having been initialized?

Dead2:
The only possible route I can think of is possibly via `deflateCopy()`. Not quite sure how deflate interacts with functable after that, and whether this new stream can be passed to a non-forked process or not.

phprus:
@nmoinvaz, @Dead2 

I think this change is safe now.

We cannot transfer ``z_stream`` to another process or save it to a file because it contains pointers to allocated memory.
The only way to get a valid ``z_stream`` in process is to call ``deflateInit``, which calls ``force_init``.

After ``force_init()``, a pointer to the optimized function will be stored in ``longest_match``.

KungFuJesus:
Sorry, somehow I missed this last week.  So long as the function is assigned ahead of time, the extra pointer deref should no longer be necessary.  This was mainly to avoid the added direction through the stub every time.

PR #1634, opened 2024-01-04
Added CI instance for WITH_NATIVE_INSTRUCTIONS.

nmoinvaz:


mtl1979:
I noticed this was missing, but generally trying to detect processor features fails miserably if the code is run inside container or emulator, so this is pretty much useful only for detecting regressions.

Dead2 (approved):
LGTM

PR #1635, opened 2024-01-04
Prefer HAVE_ALIGNED_ALLOC when available in zng_alloc

nmoinvaz:
Added some more helpful comments for people who come across this code. I think people would otherwise skip to `APPLE` and forget that different versions of macOS have other aligned memory functions.

### DIFF
--- a/zutil_p.h
+++ b/zutil_p.h
@@ -16,15 +16,17 @@
 
 /* Function to allocate 16 or 64-byte aligned memory */
 static inline void *zng_alloc(size_t size) {
-#ifdef HAVE_POSIX_MEMALIGN
+#ifdef HAVE_ALIGNED_ALLOC
+    return (void *)aligned_alloc(64, size);  /* Defined in C11 */
+#elif defined(HAVE_POSIX_MEMALIGN)
     void *ptr;
     return posix_memalign(&ptr, 64, size) ? NULL : ptr;
 #elif defined(_WIN32)
     return (void *)_aligned_malloc(size, 64);
 #elif defined(__APPLE__)
-    return (void *)malloc(size);     /* MacOS always aligns to 16 bytes */
-#elif defined(HAVE_ALIGNED_ALLOC)
-    return (void *)aligned_alloc(64, size);
+    /* Fallback for when posix_memalign and aligned_alloc are not available.
+     * On macOS, it always aligns to 16 bytes. */
+    return (void *)malloc(size);
 #else
     return (void *)memalign(64, size);
 #endif
--- a/zutil_p.h
+++ b/zutil_p.h
@@ -17,6 +17,8 @@
 /* Function to allocate 16 or 64-byte aligned memory */
 static inline void *zng_alloc(size_t size) {
 #ifdef HAVE_ALIGNED_ALLOC
+    /* Size must be a multiple of alignment */
+    size = (size + (64 - 1)) & ~(64 - 1);
     return (void *)aligned_alloc(64, size);  /* Defined in C11 */
 #elif defined(HAVE_POSIX_MEMALIGN)
     void *ptr;
--- a/test/test_compare256.cc
+++ b/test/test_compare256.cc
@@ -9,7 +9,7 @@
 
 extern "C" {
 #  include "zbuild.h"
-#  include "zutil_p.h"
+#  include "zutil.h"
 #  include "test_cpu_features.h"
 }
 
@@ -26,11 +26,11 @@ static inline void compare256_match_check(compare256_func compare256) {
     uint8_t *str1;
     uint8_t *str2;
 
-    str1 = (uint8_t *)zng_alloc(MAX_COMPARE_SIZE);
+    str1 = (uint8_t *)PREFIX3(alloc_aligned)(NULL, NULL, 1, MAX_COMPARE_SIZE, 64);
     ASSERT_TRUE(str1 != NULL);
     memset(str1, 'a', MAX_COMPARE_SIZE);
 
-    str2 = (uint8_t *)zng_alloc(MAX_COMPARE_SIZE);
+    str2 = (uint8_t *)PREFIX3(alloc_aligned)(NULL, NULL, 1, MAX_COMPARE_SIZE, 64);
     ASSERT_TRUE(str2 != NULL);
     memset(str2, 'a', MAX_COMPARE_SIZE);
 
@@ -45,8 +45,8 @@ static inline void compare256_match_check(compare256_func compare256) {
             str2[i] = 'a';
     }
 
-    zng_free(str1);
-    zng_free(str2);
+    PREFIX3(free_aligned)(NULL, NULL, str1);
+    PREFIX3(free_aligned)(NULL, NULL, str2);
 }
 
 #define TEST_COMPARE256(name, func, support_flag) \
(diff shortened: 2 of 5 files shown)


zutil_p.h
```diff
@@ -19,12 +19,14 @@ static inline void *zng_alloc(size_t size) {
...
+#elif defined(HAVE_ALIGNED_ALLOC)
+    return (void *)aligned_alloc(64, size);  /* Defined in C11 */
 #elif defined(_WIN32)
     return (void *)_aligned_malloc(size, 64);
 #elif defined(__APPLE__)
-    return (void *)malloc(size);     /* MacOS always aligns to 16 bytes */
-#elif defined(HAVE_ALIGNED_ALLOC)
-    return (void *)aligned_alloc(64, size);
+    /* Fallback for when posix_memalign and aligned_alloc are not available. 
+     * On macOS, it always aligns to 16 bytes. */
```
mtl1979, line 30: I'm not sure if people try to compile zlib-ng with older versions than OS X, but changing the capitalization also change the underlying meaning of the comment. With capital M, it meant any version of the operating system, not just the latest versions.

nmoinvaz: I don't think most people would read that much into the capitalization of the OS name.

Dead2: Actually, I think it might be better to prefer `aligned_alloc()` as the first choice and `posix_memalign()` as second choice.
Not because there is anything wrong with the latter one, but the former is after all part of the C11 standard now, and zlib-ng 2.1.x has C11 as a minimum requirement.

Interestingly windows does not seem to support it, ref: [C11 arriving in MSVC](https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/)
`Due to the nature of the Windows heap, aligned_alloc support is missing. The alternative is to use _aligned_malloc`
Makes me wonder what function we use on windows..

nmoinvaz: > Makes me wonder what function we use on windows..

We use `_aligned_malloc` on Windows.

nmoinvaz: I have updated it so `HAVE_ALIGNED_ALLOC` is very first.

KungFuJesus: Yes, I've always hated the fact that Microsoft never cared to write an allocator that handed out aligned heap allocations that could be freed by normal free.  This causes issues when you expect foreign libraries to take ownership over your allocations and basically forces you to provide a "free" wrapper.  

My guess is on Windows we are using our fallback that just over allocates and returns a pointer to the allocated region.  Windows _does_ have _mm_malloc, likely due to Intel insisting its existence very early on, but unlike the glibc versions of this, it can't be freed by normal "free", it needs "_mm_free".

Dead2: > We use `_aligned_malloc` on Windows.

I skipped right over the `_WIN32` section, didn't I

> I have updated it so HAVE_ALIGNED_ALLOC is very first.

LGTM

nmoinvaz:
Weird OSS-Fuzz CI error, maybe it needs to be re-run.

Dead2:
`AddressSanitizer: invalid alignment requested in aligned_alloc: 64, alignment must be a power of two and the requested size 0xe8 must be a multiple of alignment`

Well, that is annoying.. But it also kind of makes sense, at least when it comes to vector intrinsics.

I guess that bug(?) has been hiding there all along.

nmoinvaz:
Last commit was due to this error on macOS:
```bash
error: use of undeclared identifier 'aligned_alloc'
```

https://developer.apple.com/forums/thread/81413
https://github.com/kcat/openal-soft/issues/311

zutil_p.h
```diff
@@ -16,15 +16,19 @@
 
 /* Function to allocate 16 or 64-byte aligned memory */
 static inline void *zng_alloc(size_t size) {
-#ifdef HAVE_POSIX_MEMALIGN
+#ifdef HAVE_ALIGNED_ALLOC
+    /* Size must be a multiple of alignment */
+    size = (size + (64 - 1)) & ~(64 - 1);
```
Dead2, line 21: I am wondering whether perhaps this should rather be an assert, and each alloc should instead be individually padded to the nearest 64-bytes.
This ensures we always have a 64-byte aligned end to avoid hiding bugs, and a predictable allocation size. Except for macos (maybe we should start using `_aligned_alloc()` and `_aligned_free()` there).

The only problem with either solution is that we end up allocing more memory than before, that will cause warnings in Nginx and require it to be patched, but no other application that we know of hard-codes what we can alloc.

I am wondering whether we should consider moving to always making just a single alloc, and then distribute the needed buffers from that. That would avoid Nginx or any other application needing to optimize away our alloc() syscalls, while also letting us more easily ensure that we always have the needed padding and alignment no matter what platform we are on. This also likely reduces memory usage slightly, because each buffer we delegate starts and ends on a 64-byte boundary, thus packing the allocations together more tightly. Optionally we could also make sure we only 64-align the buffers that need  it, letting the remainder be 16-aligned.
If we go this route, we should perhaps go there directly, to avoid multiple changes that affect external applications like Nginx.

nmoinvaz: > I am wondering whether perhaps this should rather be an assert, and each alloc should instead be individually padded to the nearest 64-bytes.

I think that moves the complexity to other parts of the code, where as right here it is nicely contained. However, in that function, I wouldn't mind moving that particular line of code to affect all `malloc` functions.

> I am wondering whether we should consider moving to always making just a single alloc, and then distribute the needed buffers from that.

Sounds similar to what Linux kernel needs, where it is all located on the stack. But I think the `deflate/inflate` structs have to be allocated differently than the `window`, because the `window` can be resized.

I'm unfamiliar with what Nginx needs. If you want to make a new PR to test some of these ideas out, we can wait on this PR.

Dead2: Nginx makes a single allocation for zlib/zlib-ng (different total allocation size) and then returns parts of that for each alloc call zlib/zlib-ng makes. If zlib/zlib-ng allocates too much, it will log a warning and make another real alloc to avoid failing.
The idea is that it makes nginx spend less cpu-time doing alloc calls, leaving more time for real work, and possibly it is also somewhat beneficial to keep the allocations close wrt cache.

How much that actually helps IDK, but each alloc call has some cost, and that cost is probably even higher now that there are so many cpu mitigations in play. And assuming that is true, then that is something that would be beneficial to other applications as well for reducing zlib init latency (web browsers for example).

Hmm, makes me wonder whether we should have a some kind of google benchmark for deflate-init and inflate-init. That would be useful to make sure we have some way to make sure we don't have regressions in how fast init is completed.

mtl1979: Using asserts unnecessarily in production code is generally frowned upon as assertion failure cause termination of the program. It's very common to allocate a buffer that isn't multiple of 64 bytes, so padding is in my professional opinion better solution. We shouldn't forget that we already pad allocations to make sure optimised instructions can read the very last byte of "real" data without resorting to single byte access.

What comes to nginx, it does the allocation already in incompatible, unsupported, way, and as such we shouldn't care if we break it with any change. Until nginx patches are merged upstream, there is no justification for not just making patches for new nginx versions whenever something changes in zlib-ng.

Dead2:
Will pull this after #1654, since that also changes alloc sizes.

Dead2 (approved):

PR #1637, opened 2024-01-08
Add FAR macro to zlib-compat headers to improve compatibility.

Dead2:
Distros that are starting to use zlib-ng in compat mode are finding compile failures due to the missing FAR macro.
This adds that back as an empty define.

### DIFF
--- a/zconf.h.in
+++ b/zconf.h.in
@@ -116,6 +116,9 @@
 #ifndef ZEXPORTVA
 #  define ZEXPORTVA Z_EXPORTVA
 #endif
+#ifndef FAR
+#  define FAR
+#endif
 
 /* Legacy zlib typedefs for backwards compatibility. Don't assume stdint.h is defined. */
 typedef unsigned char Byte;


mtl1979:
`FAR` was used with 16-bit processors... I remember we added back some other defines for compat mode a while ago, but eventually we should encourage distros to drop support for 16-bit processors. Heck... Soon we drop support for some 32-bit architectures as not all 64-bit operating systems can run 32-bit binaries.

Dead2:
@mtl1979 This is not due to distros, but due to old applications that use zlib and still expect the FAR define to exist (even while not really needing it anymore). One known example is minizip.
Distros are currently adding their own patches to add FAR into zlib-ng, so better that we do it upstream to avoid fragmentation and frustration.

mtl1979:
> @mtl1979 This is not due to distros, but due to old applications that use zlib and still expect the FAR define to exist (even while not really needing it anymore). One known example is minizip. Distros are currently adding their own patches to add FAR into zlib-ng, so better that we do it upstream to avoid fragmentation and frustration.

Adding some things upstream is often the best solution... But we really should investigate if `FAR` is the very last one missing, or is there more that old applications, that we really should support, still need...

nmoinvaz (approved):

PR #1638, opened 2024-01-09
Fix building tests with cmake 3.10 and older.

mtl1979:
* if GTest::GTest target is not found, we shouldn't try to build `gtest_zlib`.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -157,99 +157,101 @@ if(WITH_GTEST)
         add_library(GTest::Main ALIAS gtest_main)
     endif()
 
-    set(TEST_SRCS
-        test_compress.cc
-        test_compress_bound.cc
-        test_cve-2003-0107.cc
-        test_deflate_bound.cc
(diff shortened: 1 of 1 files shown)


test/CMakeLists.txt
```diff
@@ -196,8 +196,10 @@ if(WITH_GTEST)
              )
     endif()
 
-    add_executable(gtest_zlib ${TEST_SRCS})
-    configure_test_executable(gtest_zlib)
+    if(TARGET GTest::GTest)
+        add_executable(gtest_zlib ${TEST_SRCS})
+        configure_test_executable(gtest_zlib)
+    endif()
```
Dead2: I think perhaps this if block should encompass the whole rest of the file.
There are multiple messages and references to gtest_zlib that will be invalid if not broken.

As far as I can see, there is no value to keeping those blocks outside the if block. There is also some useless code above (from line 160), but none of that should be visible to the user or break anything.

Dead2 (approved):
LGTM

PR #1643, opened 2024-01-19
Improve x86 intrinsics dependencies

phprus:
AVX512 includes AVX2, ..., SSE(N+1) includes SSE(N).
I have not seen any compilers that violate this chain of inclusions.

We can remove unnecessary checks.


Dead2:
Hmm, I think this looks good.

FYI, there is a plan for making it possible to compile for a minimum arch of for example `x86-64-v2` or `x86-64-v3`, where only the best of the optimized functions needed for that target is compiled in. (So no C fallback functions, and no SSE2 if SSSE3 exists, etc. But AVX512 is still there for better than minimum cpus.) That is why I moved all the C fallback functions to arch/generic, as a first step towards making those completely optional.
The intention is for distros that now target certain arches to be able to drop the code that will never be run in any case.
I think this PR helps clean up some of the requirements for that.

phprus:
@Dead2 

Maybe in this PR we should also remove checks for support for pre-AVX-512 intrinsics, since all supported compilers have these options?
GCC-8.3: https://gcc.gnu.org/onlinedocs/gcc-8.3.0/gcc/x86-Options.html
Clang-9: https://releases.llvm.org/9.0.0/tools/clang/docs/ClangCommandLineReference.html
MSVC: 
    https://learn.microsoft.com/en-us/cpp/build/reference/arch-x86
    https://learn.microsoft.com/en-us/cpp/build/reference/arch-x64


And drop support 32bit MacOS X?


The ``detect-intrinsics.cmake`` file has a comment:

> the pclmul code currently crashes on Mac in 32bit mode. Avoid for now.

that was added in 2017 (https://github.com/zlib-ng/zlib-ng/commit/d265bf7f4be7c6efe52dd719b3633260221f56b4)

Latest version supporting 32 bit applications - macOS Mojave (10.14).
macOS Mojave unsupported as of October 2021.
I can't even test for this issue, because i haven't mac with os prior to Catalina.

phprus:
@Dead2
Last 4 commits in branch https://github.com/phprus/zlib-ng/tree/arch-cond-cleanup

KungFuJesus:
So from what I could tell, the "WITH_*" macros for those intrinsics weren't so much as detecting if they were supported but were for compiling support for those functions.  E.g. you could turn one off if you wanted to.  In general you probably didn't want to, though, as some of them called the lesser variants for shorter widths.

phprus:
@KungFuJesus 
For example, to make build with SSE 4.2, but without SSE 2?
I don’t understand why that an build might be needed?

KungFuJesus:
I'm not saying it's a good idea to do so, but a possible motivating factor might be a smaller binary size. Honestly though that macro is guarding things which only _need_ SSE2 but have no greater SSE4 variants so it's already a bad idea to turn off SSE2 in that case.

KungFuJesus:
I doubt this change has any negative side effects but I'll test it tonight. I'm otherwise good with it at face value

KungFuJesus (approved):
No discernible changes with compilation with this.  I don't have a vpclmulqdq capable CPU to test that, and there's the always problematic "use native" being incompatible with build options, but I don't see anything here as an issue.

_Maybe_ with WITH_NATIVE_INSTRUCTIONS could make cmake only use native capabilities for the WITH_ options, but maybe that warrants a different PR.

phprus:
@KungFuJesus
I have one idea about implementing static function dispatching when using ``WITH_NATIVE_INSTRUCTIONS``:
~~https://github.com/phprus/zlib-ng/tree/functable-native~~
V2: https://github.com/phprus/zlib-ng/tree/functable-native-2

The last commit implements static dispatch using predefined compiler macros.

I plan to propose this changes a bit later after minor tweaks.

Dead2:
> @Dead2
> Maybe in this PR we should also remove checks for support for pre-AVX-512 intrinsics, since all supported compilers have these options? GCC-8.3: https://gcc.gnu.org/onlinedocs/gcc-8.3.0/gcc/x86-Options.html Clang-9: https://releases.llvm.org/9.0.0/tools/clang/docs/ClangCommandLineReference.html MSVC: https://learn.microsoft.com/en-us/cpp/build/reference/arch-x86 https://learn.microsoft.com/en-us/cpp/build/reference/arch-x64

I don't think it is a good idea to completely remove support for older compilers just because they are not actively supported anymore. There are for example binary distributors that do static compiles on older distros so that their binary can be run on pretty much all distros, both old and new. We still try to retain support for older compilers if we can.

But, compilers older than GCC 4.5 is probably few and far between, so I would be open to assuming the compiler supports up to sse4.2. Unless there are arguments against this of course.
But this should be a separate PR in any case.

> And drop support 32bit MacOS X?
> 
> The `detect-intrinsics.cmake` file has a comment:
> 
> > the pclmul code currently crashes on Mac in 32bit mode. Avoid for now.
> 
> that was added in 2017 ([d265bf7](https://github.com/zlib-ng/zlib-ng/commit/d265bf7f4be7c6efe52dd719b3633260221f56b4))
> 
> Latest version supporting 32 bit applications - macOS Mojave (10.14). macOS Mojave unsupported as of October 2021. I can't even test for this issue, because i haven't mac with os prior to Catalina.

Same as above, I don't think removing support is needed. It was released in 2018, so it is not really that old. I don't think anyone should still use that, but that is really none of my business, there will be users and corner-cases. And we don't really win anything by removing that support.

CMakeLists.txt
```diff
@@ -808,7 +808,7 @@ if(WITH_OPTIM)
         endif()
         if(WITH_SSSE3)
             check_ssse3_intrinsics()
-            if(HAVE_SSSE3_INTRIN)
+            if(HAVE_SSSE3_INTRIN AND WITH_SSE2)
```
Dead2, line 811: This fifth commit confuses me. It adds explicit dependencies that are now made redundant by one of the earlier commits [Fix x86 WITH_* options dependencies](https://github.com/zlib-ng/zlib-ng/pull/1643/commits/ccdaab5e09404732047f945fd1f03e242b615c26).

In this line, `WITH_SSE2` will always be true because `WITH_SSSE3` cannot be true without it.
The same kind of thing repeats for many of the changes in this commit.

I am confused, what is the rationale behind these changes?

phprus: ``cmake_dependent_option`` specifies dependencies for build options. We can request support for an instruction set that is not supported by the compiler.

This change defines dependencies for compiler-supported intrinsics. If the compiler does not support ``SSE2`` intrinsics, we disable ``SSSE3``.

See https://github.com/zlib-ng/zlib-ng/pull/1643/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR806

If the compiler does not support ``SSE2`` we set ``WITH_SSE2`` to ``OFF``.

Dead2: Then doesn't this `AND WITH_SSE2` check really belong two lines above, at line 809?
There is no reason to run `check_ssse3_intrinsics()` if its results always gets ignored because of the check on the next line.. I don't see any other reason to have the check where it is, unless you hope to add non-intrinsic SSSE3 support, but that would require ASM, and we try to avoid that.

phprus: Only for simplification.

My code (we need set ``WITH_SSSE3`` to ``OFF`` if ``WITH_SSE2`` is ``OFF`` or compiler not support ``SSSSE3``):

```cmake
        if(WITH_SSSE3)
            check_ssse3_intrinsics()
            if(HAVE_SSSE3_INTRIN AND WITH_SSE2)
                add_definitions(-DX86_SSSE3)
                set(SSSE3_SRCS ${ARCHDIR}/adler32_ssse3.c ${ARCHDIR}/chunkset_ssse3.c)
                add_feature_info(SSSE3_ADLER32 1 "Support SSSE3-accelerated adler32, using \"${SSSE3FLAG}\"")
                list(APPEND ZLIB_ARCH_SRCS ${SSSE3_SRCS})
                set_property(SOURCE ${SSSE3_SRCS} PROPERTY COMPILE_FLAGS "${SSSE3FLAG} ${NOLTOFLAG}")
            else()
                set(WITH_SSSE3 OFF)
            endif()
        endif()
```

Your suggestion:

```cmake
        if(WITH_SSSE3 AND WITH_SSE2)
            check_ssse3_intrinsics()
            if(HAVE_SSSE3_INTRIN)
                add_definitions(-DX86_SSSE3)
                set(SSSE3_SRCS ${ARCHDIR}/adler32_ssse3.c ${ARCHDIR}/chunkset_ssse3.c)
                add_feature_info(SSSE3_ADLER32 1 "Support SSSE3-accelerated adler32, using \"${SSSE3FLAG}\"")
                list(APPEND ZLIB_ARCH_SRCS ${SSSE3_SRCS})
                set_property(SOURCE ${SSSE3_SRCS} PROPERTY COMPILE_FLAGS "${SSSE3FLAG} ${NOLTOFLAG}")
            else()
                set(WITH_SSSE3 OFF)
            endif()
        else()
            set(WITH_SSSE3 OFF)
        endif()
```

Yes, ``check_ssse3_intrinsics`` will be called unnecessarily, but I think it simplifies the code.

Dead2: I understand, your suggestion makes it more clear what kind of instruction set the if block is about.
That could also be done by adding a comment on the same line or the line above, but I am ok with this solution.

Dead2 (approved):

PR #1644, opened 2024-01-19
Split cpu_features.h by architecture.

nmoinvaz:


CMakeLists.txt
```diff
@@ -532,10 +532,11 @@ if(WITH_REDUCED_MEM)
     message(STATUS "Configured for reduced memory environment")
 endif()
 
+set(ARCHDIR "arch/generic")
 
 set(ZLIB_ARCH_SRCS)
-set(ZLIB_ARCH_HDRS)
-set(ARCHDIR "arch/generic")
+set(ZLIB_ARCH_HDRS ${ARCHDIR}/generic_features.h)
```
Dead2: I think this should instead be a separate variable set to `arch/generic` (`GENERIC_ARCHDIR`?), since `ARCHDIR` is only set to `arch/generic` temporarily as a fallback. And the setting of `ARCHDIR` should be moved down to line 554, in the else() block.

nmoinvaz: Done.

Dead2:
Just wondering, is it useful to have `extern` definitions behind ìfdef`? It makes it look more cluttered, but do we gain anything by doing it?
Perhaps it moves some potential errors from linking to compile? Anything else?

nmoinvaz:
I added a new commit to test it to see if removing `extern` causes any problems.

Dead2:
@nmoinvaz Is this one good to go?
I don't see any reason to object.

nmoinvaz:
Yes 👍🏼

phprus:
Maybe mark all architecture specific functions as ``Z_INTERNAL``?

nmoinvaz:
Hmm.. I was hesitant to even remove `extern` since these functions are also used by the testing framework..

nmoinvaz:
Rebased

Dead2 (approved):

nmoinvaz:
Yay!

PR #1626, opened 2023-12-27
When compiling with native instructions bypass functable.

nmoinvaz:

PR #1645, opened 2024-01-21
Proposal: Disable dynamic function dispatching for native or arch-specific builds

phprus:
My vision of PR #1626 (@nmoinvaz)

~~Based on my PR #1643~~

For X86, PPC, RISC-V, s390 the conditions are correct.
**For ARM, they require verification.**

Microsoft Visual Studio does not support compile time detection of CPU features for ``/arch`` before ``AVX``. Missing predefined macros (``__SSE*__``).

### DIFF
--- a/arch/arm/arm_features.h
+++ b/arch/arm/arm_features.h
@@ -53,4 +53,45 @@ uint32_t crc32_acle(uint32_t crc, const uint8_t *buf, size_t len);
 
 #endif
 
+
+#ifdef DISABLE_RUNTIME_CPU_DETECTION
+//// ARM - SIMD
+#  if (defined(ARM_SIMD) && defined(__ARM_FEATURE_SIMD32)) || defined(ARM_NOCHECK_SIMD)
+#    undef native_slide_hash
+#    define native_slide_hash slide_hash_armv6
+#  endif
+// ARM - NEON
+#  if (defined(ARM_NEON) && (defined(__ARM_NEON__) || defined(__ARM_NEON))) || ARM_NOCHECK_NEON
+#    undef native_adler32
+#    define native_adler32 adler32_neon
+#    undef native_chunkmemset_safe
+#    define native_chunkmemset_safe chunkmemset_safe_neon
+#    undef native_chunksize
+#    define native_chunksize chunksize_neon
+#    undef native_inflate_fast
+#    define native_inflate_fast inflate_fast_neon
+#    undef native_slide_hash
+#    define native_slide_hash slide_hash_neon
+#    ifdef HAVE_BUILTIN_CTZLL
+#      undef native_compare256
+#      define native_compare256 compare256_neon
+#      undef native_longest_match
+#      define native_longest_match longest_match_neon
+#      undef native_longest_match_slow
+#      define native_longest_match_slow longest_match_slow_neon
+#    endif
+#  endif
+// ARM - ACLE
+#  if defined(ARM_ACLE) && defined(__ARM_ACLE) && defined(__ARM_FEATURE_CRC32)
+#    undef native_crc32
+#    define native_crc32 crc32_acle
+#    undef native_insert_string
+#    define native_insert_string insert_string_acle
+#    undef native_quick_insert_string
+#    define native_quick_insert_string quick_insert_string_acle
+#    undef native_update_hash
+#    define native_update_hash update_hash_acle
+#  endif
+#endif
+
 #endif /* ARM_H_ */
--- a/functable.h
+++ b/functable.h
@@ -35,4 +35,11 @@ struct functable_s {
 
 Z_INTERNAL extern struct functable_s functable;
 
+
+/* Explicitly indicate functions are conditionally dispatched.
+ */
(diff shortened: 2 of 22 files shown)


nmoinvaz:
It will be good to see what it looks after PR #1643 and #1644 lands.

nmoinvaz:
How do we address any of the comments by @Dead2 here https://github.com/zlib-ng/zlib-ng/pull/1626#issuecomment-1875230082?

phprus:
@nmoinvaz, @Dead2

I agree with @Dead2 (https://github.com/zlib-ng/zlib-ng/pull/1626#issuecomment-1875230082) if we use ``-march=***`` and dynamic dispatching.
If we use static/compile-time dispatch (this PR), then functions for this architecture will be selected but on compile time. This PR essentially disables newer and older instruction sets.

I'm thinking of adding restrictions on the compilation of those functions that will not be called during static dispatch.

phprus:
Rebased onto ``develop``.

phprus:
Conditions have been updated for PPC (without changes), RISC-V, s390.

**Verification of the conditions for ARM is required.**

phprus:
@nmoinvaz, please remove the "PR Dependency" label.

phprus:
Rebased.

functable.h
```diff
@@ -6,12 +6,285 @@
...
+#include "zbuild.h"
+#include "zendian.h"
 #include "deflate.h"
 #include "crc32_fold_c.h"
 #include "adler32_fold_c.h"
 #include "inftrees.h"
 #include "inflate.h"
 
+
+#if defined(UNALIGNED_OK) && BYTE_ORDER == LITTLE_ENDIAN
```
nmoinvaz: I think we should see what all these `#define` changes in `functable.h` look like if they were moved into their associated arch/x86/*.h headers. Perhaps you can add it as a separate commit, that way it can be easily backed out or squashed at the end. But I think it might clean all this up a bit more.

phprus: I think it is a bad idea.
Currently there are two dispatching points:
``functable.h`` - for compile-time
``functable.c`` - for runtime.

Moving ``#define`` into the ``arch/x86/*.h`` headers will require multiple checks of the same conditions in headers.

nmoinvaz: Well, I mean for non-x86 those functions would be moved into the header for their own arch folder too.  All the arch ifdef logic would be kept in x86_features or arm_features.h ...

phprus: @nmoinvaz 
I understood your idea!

Dead2: lol, I had just written a big comment about there probably being a misunderstanding between you two, and trying to explain it better. Was about to press send when your new comment popped up 😄 

You still seem to think it is a problem though, so please explain what the problem is, because I don't see it currently.

---

I don't think that would increase the compiler load by a noticeable degree, so I am wondering whether @phprus misunderstood the suggestion..?

Basically @nmoinvaz suggested splitting this file into arch-dependent files, only one of those would ever be active for a compiler to see.
The C-fallback parts could also move to arch/generic/ where those functions currently live.

Then this file would mainly be the struct and a few other bits and pieces that are functable-specific.

In that case this file would include the generic header file and the arch-specific header file.
If that is still a problem, I am afraid I do not understand what the problem you are describing is @phprus?

phprus: Splitted.

nmoinvaz: Sorry to be a pain. Anyways, it looks good. Now future changes to x86 code should even more so only touch arch/x86 files, same for other archs too.

phprus: We need split the runtime dispatcher too, but this is more difficult task.

I didn't succeed, because the ``*_features.h`` headers contain a lot of non-obvious dependencies.
For example:

```c
#ifdef CPU_FEATURES_H_
```

or 
```c
#ifdef INFLATE_H_
```

arch/arm/arm_features.h
```diff
@@ -53,4 +53,45 @@ uint32_t crc32_acle(uint32_t crc, const uint8_t *buf, size_t len);
 
 #endif
 
+
+#ifdef DISABLE_RUNTIME_CPU_DETECTION
+//// ARM - SIMD
+#  if (defined(ARM_SIMD) && 0 /* unknown predefined macro */ ) || defined(ARM_NOCHECK_SIMD)
```
phprus: We must find the macros required for this check (ARM).

functable.h
phprus: Another implementation of this PR: #1659

Main differences:
- Separate CPU-specific function prototypes and CPU-arch features checker.
- Reducing dependencies between header files.

Separating headers reduces dependencies on implementations (check internal macro ``CPU_FEATURES_H_`` or ``INFLATE_H_ ``, ...)

arch/arm/arm_features.h
mtl1979: ```__ARM_FEATURE_SIMD32```

phprus:
All compile time CPU features checks is implemented.
I think PR #1659 is more promising.

phprus:
Rebased.

@Dead2, what do you think about this version or version 2 (PR #1659)?

PR #1649, opened 2024-01-24
Fix regression caused by 2fa631e029084b75acd81db5d33fd4aa802fd082

mtl1979:
* POWER8/9 feature checks were enabled even if the toolchain didn't support AT_HWCAP2
* Add detection if we need to include <linux/auxvec.h>

See #1648.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -327,6 +327,14 @@ endif()
 check_include_file(unistd.h    HAVE_UNISTD_H)
 
 #
+# Check for Linux includes
+#
+check_include_file(linux/auxvec.h HAVE_LINUX_AUXVEC_H)
+if(HAVE_LINUX_AUXVEC_H)
+    add_definitions(-DHAVE_LINUX_AUXVEC_H)
+endif()
+
+#
 # Check to see if we have large file support
 #
 set(CMAKE_REQUIRED_DEFINITIONS -D_LARGEFILE64_SOURCE=1 -D__USE_LARGEFILE64)
@@ -685,8 +693,13 @@ if(WITH_OPTIM)
         if(WITH_POWER9)
             check_power9_intrinsics()
         endif()
-        if(HAVE_VMX OR HAVE_POWER8_INTRIN OR HAVE_POWER9_INTRIN)
+        if(POWER8_NEED_AUXVEC_H OR POWER9_NEED_AUXVEC_H)
+            add_definitions(-DPOWER_NEED_AUXVEC_H)
+        endif()
+        if(HAVE_POWER8_INTRIN OR HAVE_POWER9_INTRIN)
             add_definitions(-DPOWER_FEATURES)
+        endif()
+        if(HAVE_VMX OR HAVE_POWER8_INTRIN OR HAVE_POWER9_INTRIN)
             list(APPEND ZLIB_ARCH_HDRS ${ARCHDIR}/power_features.h)
             list(APPEND ZLIB_ARCH_SRCS ${ARCHDIR}/power_features.c)
         endif()
--- a/cmake/detect-intrinsics.cmake
+++ b/cmake/detect-intrinsics.cmake
@@ -364,6 +364,21 @@ macro(check_power8_intrinsics)
         }"
         HAVE_POWER8_INTRIN
     )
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -326,6 +326,15 @@ if(HAVE_SYS_SDT_H)
 endif()
 check_include_file(unistd.h    HAVE_UNISTD_H)
 
+#
+# Check for Linux includes
+#
+
```
nmoinvaz: Extra line here?

cmake/detect-intrinsics.cmake
```diff
@@ -364,6 +364,21 @@ macro(check_power8_intrinsics)
         }"
         HAVE_POWER8_INTRIN
     )
+    if (NOT HAVE_POWER8_INTRIN AND HAVE_LINUX_AUXVEC_H)
```
nmoinvaz: No space after `if`

cmake/detect-intrinsics.cmake
```diff
@@ -364,6 +364,21 @@ macro(check_power8_intrinsics)
...
+        check_c_source_compiles(
+            "#include <sys/auxv.h>
+            #include <linux/auxvec.h>
+            int main() {
+                return (getauxval(AT_HWCAP2) & PPC_FEATURE2_ARCH_2_07);
+            }"
+            HAVE_POWER8_INTRIN2
+        )
+        if(HAVE_POWER8_INTRIN2)
+            set(POWER8_NEED_AUXVEC_H 1)
```
nmoinvaz, line 377: Set to `ON` instead of `1` like we have many places elsewhere?

cmake/detect-intrinsics.cmake
```diff
@@ -422,6 +437,21 @@ macro(check_power9_intrinsics)
...
+        check_c_source_compiles(
+            "#include <sys/auxv.h>
+            #include <linux/auxvec.h>
+            int main() {
+                return (getauxval(AT_HWCAP2) & PPC_FEATURE2_ARCH_3_00);
+            }"
+            HAVE_POWER9_INTRIN2
+        )
+        if(HAVE_POWER9_INTRIN2)
+            set(POWER9_NEED_AUXVEC_H 1)
```
nmoinvaz, line 450: Set to `ON` instead of `1` like we have many places elsewhere?

cmake/detect-intrinsics.cmake
nmoinvaz: Also why even set this `POWER9_NEED_AUXVEC_H` variable, it is not used further down in the file..

cmake/detect-intrinsics.cmake
```diff
@@ -364,6 +364,21 @@ macro(check_power8_intrinsics)
...
+            #include <linux/auxvec.h>
+            int main() {
+                return (getauxval(AT_HWCAP2) & PPC_FEATURE2_ARCH_2_07);
+            }"
+            HAVE_POWER8_INTRIN2
+        )
+        if(HAVE_POWER8_INTRIN2)
+            set(POWER8_NEED_AUXVEC_H 1)
+            set(HAVE_POWER8_INTRIN ${HAVE_POWER8_INTRIN} CACHE INTERNAL "Have POWER8 intrinsics" FORCE)
+            unset(HAVE_POWER8_INTRIN2 CACHE)
```
nmoinvaz, line 379: Why unset the value?

cmake/detect-intrinsics.cmake
mtl1979: It's used in root CMakeLists.txt to avoid adding same define twice if either POWER8 or POWER9 optimizations are disabled. I didn't use `ON` because this isn't a switch, it's internal variable.

cmake/detect-intrinsics.cmake
mtl1979: I've done the same in other similar tests... So it's just for consistency to avoid adding the variable to cache.

cmake/detect-intrinsics.cmake
mtl1979: Actually we already use `1` for internal variables like this.

cmake/detect-intrinsics.cmake
nmoinvaz: Ok I will review it again

Dead2 (approved):
LGTM

PR #1651, opened 2024-01-25
Add VPCLMULQDQ crc32 tests to Google benchmarks

nmoinvaz:


Dead2 (approved):
LGTM 👍

nmoinvaz:
Strange, it seems this potentially discovered a crash on GCC benchmark with VPCLMULQDQ. @KungFuJesus maybe you can take a look?

Dead2:
> Strange, it seems this potentially discovered a crash on GCC benchmark with VPCLMULQDQ.

The test machine doesn't have AVX512, so it shouldn't try to use VPCLMULQDQ at all. So looks like the cpu check is possibly at fault somehow.

nmoinvaz:
Good observation I didn't think about that possibility.

nmoinvaz:
I think I missed some of the logic for the benchmark condition. I found existing logic elsewhere and just copied it.

nmoinvaz:
That doesn't seem to have fixed it. So maybe there is a bug in CPU runtime detection..

nmoinvaz:
For example, here is what running the same crc32 tests on my desktop (without AVX512) using MSVC shows:
image

nmoinvaz:
Looks like that fixed it.

nmoinvaz:
I'm not sure why the code coverage would be less... 
image
Seems ok?

mtl1979:
> I'm not sure why the code coverage would be less... image Seems ok?

There is a lot of partial coverage on the lines with preprocessor macros... This suggests that some of the expanded code is not reached.

nmoinvaz:
Ah yea those Google testing/benchmark macros have a lot of boilerplate.

PR #1654, opened 2024-01-29
Sync changes with zlib 1.3.1

nmoinvaz:
A lot of the other changes didn't apply to us because either we don't have those contribs or we have already made such changes.


gzguts.h
```diff
@@ -139,6 +139,7 @@ void Z_INTERNAL gz_error(gz_state *, int, const char *);
 /* GT_OFF(x), where x is an unsigned value, is true if x > maximum z_off64_t
    value -- needed when comparing unsigned to z_off64_t, which is signed
    (possible z_off64_t types off_t, off64_t, and long are all signed) */
+unsigned Z_INTERNAL gz_intmax(void);
```
phprus, line 139: This commit is not required.
``gz_intmax`` is not exported in zlib (``gz_intmax`` in ``local`` section of https://github.com/madler/zlib/blob/develop/zlib.map).
In zlib-ng it is removed from the ``local`` section of ``zlib.map`` and ``zlib-ng.map`` (commit: https://github.com/zlib-ng/zlib-ng/commit/b1ebcc0ee988b398580bced4a90050d41ec38133).
It's does not affect the ABI.

nmoinvaz: But wouldn't it still be necessary for ZLIB_COMPAT when compiling statically?

phprus: Interesting question!


If we need to maintain compatibility of internal (non exported) interfaces for static library then I suggest making this change like:

```c
#ifdef ZLIB_COMPAT
unsigned Z_INTERNAL gz_intmax(void);
#endif
```

```c
#ifdef ZLIB_COMPAT

/* portably return maximum value for an int (when limits.h presumed not
   available) -- we need to do this to cover cases where 2's complement not
   used, since C standard permits 1's complement and sign-bit representations,
   otherwise we could just use ((unsigned)-1) >> 1 */
unsigned Z_INTERNAL gz_intmax(void) {
    return INT_MAX;
}

#endif
```

Or add the special ``ZLIB_COMPAT_STATIC`` macros only for a ``ZLIB_COMPAT``-static library target.

nmoinvaz: Fixed.

nmoinvaz:
I have added the `LIT_MEM` changes. It can be set via `-D CMAKE_C_FLAGS=-DLIT_MEM`. I ensured that `gtest_zlib` passed with `LIT_MEM` enabled.

nmoinvaz:
## PR
```
OS: Windows 10 10.0.22631 AMD64
CPU: Intel64 Family 6 Model 151 Stepping 2, GenuineIntel
Tool: ../zlib-ng/build/Release/minigzip.exe Size: 99,328 B
Levels: 0-9

 Level   Comp   Comptime min/avg/max/stddev   Compressed size
 0    100.008%      0.086/0.088/0.089/0.001       211,973,953
 1     44.409%      0.691/0.694/0.695/0.001        94,127,290
 2     35.518%      1.140/1.143/1.145/0.002        75,282,961
 3     33.882%      1.529/1.533/1.536/0.002        71,816,478
 4     33.174%      1.759/1.763/1.765/0.002        70,315,668
 5     32.660%      1.975/1.979/1.981/0.002        69,225,542
 6     32.508%      2.357/2.362/2.366/0.002        68,902,222
 7     32.255%      3.122/3.129/3.133/0.003        68,366,800
 8     32.167%      5.001/5.026/5.048/0.013        68,180,776
 9     31.887%      5.449/5.465/5.482/0.008        67,586,442

 avg1  40.847%                        2.318
 avg2  45.385%                        2.576
 tot                                695.437       865,778,132
```

## PR LIT_MEM
`-D CMAKE_C_FLAGS=-DLIT_MEM`
```
OS: Windows 10 10.0.22631 AMD64
CPU: Intel64 Family 6 Model 151 Stepping 2, GenuineIntel
Tool: ../zlib-ng/build/Release/minigzip-litmem.exe Size: 98,816 B
Levels: 0-9

 Level   Comp   Comptime min/avg/max/stddev   Compressed size
 0    100.008%      0.087/0.088/0.089/0.001       211,973,953
 1     44.409%      0.704/0.707/0.709/0.001        94,127,290
 2     35.518%      1.102/1.104/1.106/0.001        75,282,961
 3     33.882%      1.469/1.472/1.475/0.002        71,816,478
 4     33.174%      1.705/1.710/1.712/0.002        70,315,668
 5     32.660%      1.949/1.952/1.955/0.002        69,225,542
 6     32.508%      2.331/2.337/2.339/0.002        68,902,222
 7     32.255%      3.087/3.097/3.100/0.003        68,366,800
 8     32.167%      4.972/4.989/5.001/0.008        68,180,776
 9     31.887%      5.418/5.429/5.436/0.006        67,586,442

 avg1  40.847%                        2.289
 avg2  45.385%                        2.543
 tot                                686.561       865,778,132
```
Compression speed improvement of ~1.276% which is within the range of 1-2% mentioned by Mark in his commit:
>A bug fix in zlib 1.2.12 resulted in a slight slowdown (1-2%) of
deflate. This commit provides the option to #define LIT_MEM, which
uses more memory to reverse most of that slowdown. The memory for
the pending buffer and symbol buffers is increased by 25%, which
increases the total memory usage with the default parameters by
about 6%.

sym_buf changes from 65536 -> 81920 (+16,384)

nmoinvaz:
We might want to always enable `LIT_MEM` by default, except when CMake option `WITH_REDUCED_MEM=ON`. @Dead2?

Dead2:
> We might want to always enable `LIT_MEM` by default, except when CMake option `WITH_REDUCED_MEM=ON`. @Dead2?

I agree. I don't see any reason to disable this by default in our case. 16KiB is not a whole lot for the systems we target.

I am also a little surprised that your LIT_MEM .exe file is 512bytes smaller than without (but 512b might be the step size of your compiler/linker, I think GCC/LD uses 16 byte steps IIRC).  Didn't expect it to get smaller by that change at all.

nmoinvaz:
I enabled `-DLIT_MEM` by default.

> I am also a little surprised that your LIT_MEM .exe file is 512bytes smaller than without

It's called the "MSVC special".

deflate.h
```diff
@@ -22,6 +22,10 @@
 #  define GZIP
 #endif
 
+/* define LIT_MEM to slightly increase the speed of deflate (order 1% to 2%) at
+   the cost of a larger memory footprint */
+/* #define LIT_MEM */
```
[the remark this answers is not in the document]
Dead2: Perhaps remove the last line here with the #define. It really serves no purpose for us.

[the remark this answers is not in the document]
Dead2: Perhaps change this to be like the above GZIP define to invert the activation.
That way it is always active unless NO_LIT_MEM is defined, more clean that way.
That also means you don't have to change any of the win32 makefiles, only CMake and Configure offer the reduced memory settings AFAIK.

Dead2:
I wonder why pkgcheck fails, I don't see anything that would cause CMake and Configure builds to become different.

deflate.h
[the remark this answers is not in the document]
nmoinvaz: Done.

gzlib.c
```diff
@@ -523,3 +523,13 @@ void Z_INTERNAL gz_error(gz_state *state, int err, const char *msg) {
     }
     (void)snprintf(state->msg, strlen(state->path) + strlen(msg) + 3, "%s%s%s", state->path, ": ", msg);
 }
+
+#ifdef ZLIB_COMPAT
+/* portably return maximum value for an int (when limits.h presumed not
+   available) -- we need to do this to cover cases where 2's complement not
+   used, since C standard permits 1's complement and sign-bit representations,
+   otherwise we could just use ((unsigned)-1) >> 1 */
```
mtl1979: The rationale here doesn't match the code below...

nmoinvaz: Yeah I can probably just remove the comment.

nmoinvaz: Fixed.

Dead2 (approved):
LGTM

PR #1655, opened 2024-01-29
Removing some outdated comments

KungFuJesus:
These were left on my part, the inline copy + checksum is the very thing the function is doing.


Dead2:
Good cleanup :)

PR #1662, opened 2024-02-08
Allow overwrite NATIVEFLAG value.

phprus:


phprus:
CI error:
``[2024-02-08T14:12:30.411Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 503 - upstream connect error or disconnect/reset before headers. reset reason: connection failure``

nmoinvaz:
Seems even though we collect all coverage reports and upload them at once, that codecov.io's servers still have problems from time to time.

nmoinvaz (approved):
Looks like this makes some important corrections. LGTM

.github/workflows/cmake.yml
```diff
@@ -41,6 +41,13 @@ jobs:
             cmake-args: -DWITH_NATIVE_INSTRUCTIONS=ON
             codecov: ubuntu_gcc_native_inst
 
+          - name: Ubuntu GCC Native Instructions (AVX)
+            os: ubuntu-latest
+            compiler: gcc
+            cxx-compiler: g++
+            cmake-args: -DWITH_NATIVE_INSTRUCTIONS=ON -DNATIVEFLAG=-mavx
```
Dead2: I am not sure it is a good idea to export `NATIVEFLAG` directly like this.

I think exporting for example `NATIVE_ARCH_OVERRIDE` instead, and setting `NATIVEFLAG` based on it would be better. And perhaps print a warning in/after the CMake summary (Like CPack does on line :-4 at the end of CMakeLists.txt).
`message (WARNING "WITH_NATIVE_INSTRUCTIONS enabled, but running with NATIVE_ARCH_OVERRIDE: XX")`

The option is great for testing, and potentially good for making very narrow builds for a different machine, but also very much an advanced option that can also break things in subtle or not-so-subtle ways.
With the renamed parameter and the warning, I think it makes it more clear to potential users that this is not for the casual tinkerer and also easier to understand what its purpose is.

What do you think @phprus?

phprus: @Dead2 
``NATIVE_ARCH_OVERRIDE`` is a good name.

I think it's better to print ``message(STATUS ...)`` instead of ``message(WARNING ...)``.

PR updated.

phprus:
Fix gcc flags for AVX.

phprus:
Ready for review.
Now support multiple flags in ``NATIVE_ARCH_OVERRIDE``

phprus:
@Dead2
rerun CI actions please.

Dead2 (approved):

PR #1663, opened 2024-02-11
Fix `deflate_state` alignment with MS or clang-cl compilers

pps83:
When building with clang-cl, compiler produces the following warning:

```
zlib-ng/deflate.h(287,3): warning : attribute 'align' is ignored, place it after "struct" to apply attribute to type declaration [-Wignored-attributes]
zlib-ng/zbuild.h(196,34): note: expanded from macro 'ALIGNED_'
```
Repositioning align attribute after "struct" fixes the warning and aligns `deflate_state` correctly.

PR #1664, opened 2024-02-12
Fix xsave intrinsic test for clang, and gcc 8.2 or newer, and icc.

mtl1979:
* clang needs `<immintrin.h>` for `_xgetbv()` intrinsic function
* gcc 8 needs `<xsaveintrin.h>`
* gcc 9 and newer need `<immintrin.h>`
* icc needs `<immintrin.h>`


phprus:
PR with ``internal_state`` ``ALIGNED_(16)``: #1663

mtl1979:
> PR with `internal_state` `ALIGNED_(16)`: #1663

I have that user blocked for a good reason...

Dead2 (approved):

arch/x86/x86_features.c
```diff
@@ -15,6 +15,13 @@
 #else
 // Newer versions of GCC and clang come with cpuid.h
 #  include <cpuid.h>
+#  ifdef X86_HAVE_XSAVE_INTRIN
+#    if __GNUC__ == 8
```
Dead2, line 19: So this is only needed for 8.x, right? Not 9.x and newer or 7.x and older?

Also, this place you only check for 8.x, the other places you check that minor is 1 or more as well.

Intentional?

phprus: ``X86_HAVE_XSAVE_INTRIN`` is not defined is gcc < 8

phprus, line 19: Suggest:
```c
#    if defined(__GNUC__) && __GNUC__ == 8
```

And in other places.

mtl1979: > ``X86_HAVE_XSAVE_INTRIN`` is not defined is gcc < 8

It's not defined for gcc < 8.2 as the implementation is buggy...

mtl1979: > Suggest:
> ```c
> #    if defined(__GNUC__) && __GNUC__ == 8
> ```
> 
> And in other places.

It's not necessary as we already rely on anything undefined to be interpreted as `0`.

mtl1979: @Dead2 AFAIK, Clang sets `__GNUC__` to `4`, Intel icc/icpc sets it to `6` or `7` depending on what operating system is used (`6` for Ubuntu 16.04, `7` for Ubuntu 18.04, essentially to same as default gcc/g++ version on the specific distribution)

PR #1668, opened 2024-02-14
Fix include paths

pps83:
zlb-ng shouldn't require to have arch/generic in the include path


pps83:
if you search entire project, in many places you'll find: `#include "arch/generic/...`. These are the only places that for some reason want to have arch/generic in the include path.
There is also two places `#include "../generic/` that should perhaps be changed to `#include "arch/generic/...`

Dead2 (approved):

PR #1671, opened 2024-02-16
Add more result variables to the cmake package configuration

Marc-Aldorasi-Imprivata:
[FindZLIB provides `ZLIB_INCLUDE_DIRS` and `ZLIB_LIBRARIES`](https://cmake.org/cmake/help/latest/module/FindZLIB.html#result-variables), so the package configuration should as well.


mtl1979:
Instead of implementing everything third-party CMake modules have, we generally try to implement what stock zlib does have... As such I was waiting to see how Mark Adler implements the package configuration before adding more variables, so we don't need to revise anything breaking backwards compatibility if it takes long time for Mark Adler to merge necessary improvements.

I'm not generally against improving something that I have originally implemented on zlib-ng. Sometimes I just wait until I have read all the documentation, have all the facts and received enough feedback to make the best version.

Dead2:
@mtl1979 I see no harm in adding the include dir at least.
About the `set(ZLIB_LIBRARIES ZLIB::ZLIB)` I don't know, although I don't really see how it could hurt.

What are you objecting to/worried about specifically?

mtl1979:
> @mtl1979 I see no harm in adding the include dir at least. About the `set(ZLIB_LIBRARIES ZLIB::ZLIB)` I don't know, although I don't really see how it could hurt.
> 
> What are you objecting to/worried about specifically?

That the added lines have no effect due to the Cmake module already adding them. There isn't really easy way to test what variables are actually set as it would require compiling zlib-ng before the configure step is executed, thus chicken and egg dilemma.

As far as I know it runs the version check scripts first for each package, and after that finishes, it runs `FindZLIB` module to set the variables.

Marc-Aldorasi-Imprivata:
> As far as I know it runs the version check scripts first for each package, and after that finishes, it runs FindZLIB module to set the variables.

This is not the case. [Cmake runs _either_ `Find<PackageName>.cmake` or `<lowercasePackageName>-config.cmake`, it does not run both](https://cmake.org/cmake/help/latest/command/find_package.html#search-modes). In this case, since the code is in `zlib-config.cmake`, cmake would not run `FindZLIB.cmake`, and so `FindZLIB.cmake` couldn't set any variables.

In general, if you want to see where a cmake variable is being set, you can run cmake with `--trace-expand` and it will show every line of code that it executes, which includes all the calls to `set`.

Dead2 (approved):

mtl1979:
> This is not the case. [Cmake runs _either_ `Find<PackageName>.cmake` or `<lowercasePackageName>-config.cmake`, it does not run both](https://cmake.org/cmake/help/latest/command/find_package.html#search-modes). In this case, since the code is in `zlib-config.cmake`, cmake would not run `FindZLIB.cmake`, and so `FindZLIB.cmake` couldn't set any variables.

This logic was changed in recent versions of CMake... Older versions did explicitly call `Find<PackageName>.cmake` first and it was responsibility for that script to do all the logic for version and module checking (zlib-ng has no modules). Obviously it couldn't do any version checking if the specific zlib/zlib-ng version didn't advertise any version number.

We still have to support wide variety of CMake versions.

PR #1673, opened 2024-02-19
Cleanup and update NMake Makefiles.

mtl1979:
* ARM and ARM64 Makefiles had dependencies for X86 specific headers
* Some dependencies were mentioned more than once for same object file
* Use only `$(TOP)` as NMake doesn't change directories while building, `$(SRCDIR)` was redundant
* Sort the dependencies and separate dependencies for test binaries.

### DIFF
--- a/win32/Makefile.a64
+++ b/win32/Makefile.a64
@@ -29,8 +29,8 @@ WFLAGS  = \
 	-D_ARM64_WINAPI_PARTITION_DESKTOP_SDK_AVAILABLE=1 \
 	-D_CRT_SECURE_NO_DEPRECATE \
 	-D_CRT_NONSTDC_NO_DEPRECATE \
-	-DARM_NEON_HASLD4 \
 	-DARM_FEATURES \
+	-DARM_NEON_HASLD4 \
 	#
 LDFLAGS = -nologo -debug -incremental:no -opt:ref -manifest
 ARFLAGS = -nologo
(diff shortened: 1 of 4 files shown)


nmoinvaz:
Did you make a script to do this or did you do it manually?

mtl1979:
> Did you make a script to do this or did you do it manually?

I made an executable, I will make another PR/commit when I have integrated it to CI...

nmoinvaz:
I see. Or, if you want you can just make a GitHub gist and paste the code there.

mtl1979:
> I see. Or, if you want you can just make a GitHub gist and paste the code there.

https://gist.github.com/mtl1979/83d0691c0c2a03154ffbea473a85a6c3

Dead2 (approved):

Dead2:
@mtl1979 This needs a rebase

phprus:
@mtl1979
Please provide a simple example of using ``depcheck.cpp`` to check NMake dependencies.

mtl1979:
> @mtl1979 Please provide a simple example of using `depcheck.cpp` to check NMake dependencies.

I run it from root of the repository using for example `depcheck.exe win32\Makefile.msc .` ... 

Running the check on Makefiles one by one is easier for a human than running `nmake -f win32\Makefile.msc depcheck` which will check all three Makefiles in batch, as it will be harder to track which one of the three Makefiles has missing or extra dependencies... When it detects an error, the log should be read from bottom to top to see which line has the error. It will use prefix `File:` when it starts reading a new line. It will currently only check direct dependencies. I've whitelisted some known indirect dependencies until there is enough to warrant checking all the dependencies, not just the source file.

mtl1979:
`depcheck.exe` will now print current Makefile and object file when it encounters either missing or unnecessary dependency.

phprus (approved):

PR #1675, opened 2024-02-19
[CI] Fix creating list of artifacts for CodeCov.

mtl1979:
* Remove trailing comma from list of CodeCov artifacts to avoid warning and subsequent error state.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -674,7 +674,7 @@ jobs:
       shell: bash
       run: |
         python3 -u -m pip install --user gcovr==5.0
-        python3 -m gcovr -j 3 --verbose \
+        python3 -m gcovr ${{ matrix.build-dir || '' }} -j 3 --verbose \
           --exclude-unreachable-branches \
           --gcov-executable "${{ matrix.gcov-exec || 'gcov' }}" \
           --root ${{ matrix.build-src-dir || '.' }} \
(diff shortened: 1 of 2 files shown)


.github/workflows/cmake.yml
```diff
@@ -674,7 +674,7 @@ jobs:
       shell: bash
       run: |
         python3 -u -m pip install --user gcovr==5.0
-        python3 -m gcovr -j 3 --verbose \
+        python3 -m gcovr ${{ matrix.build-dir || '' }} -j 3 --verbose \
```
nmoinvaz, line 677: Why is this needed? I thought that is the purpose of the `--root` arg below.

mtl1979: For out-of-source builds, the build/binary directory needs to be sub-directory of the source directory for `--root` argument to work, it's limitation of filtering in `gcovr`... However it is possible to pass binary directory separately without that restriction.

Dead2 (approved):

.github/workflows/cmake.yml
nmoinvaz: Why was out of source builds previously working fine?

mtl1979: I don't know why but recently I noticed that the size of coverity artifact dropped from few tens of kilobytes on normal builds to about 300 bytes on OSB builds. It was pretty clear hint that no coverity data was found or collected.

PR #1640, opened 2024-01-13
Fixed S390X CI builds not running in forked repos.

nmoinvaz:
Since MSAN is not able to run in emulated environment, I have just made it to become a Compat test in forks. Without this PR, then workflows in forks will not finish and will timeout.

PR #1678, opened 2024-02-20
Cleanup z15 configure workflows.

mtl1979:
* Cleanup extra white space
* Rely on runner architecture instead of current repository like z15 cmake workflows

### DIFF
--- a/.github/workflows/configure.yml
+++ b/.github/workflows/configure.yml
@@ -171,8 +171,8 @@ jobs:
             os: ${{ github.repository == 'zlib-ng/zlib-ng' && 'z15' || 'ubuntu-latest' }}
             compiler: ${{ github.repository == 'zlib-ng/zlib-ng' && 'gcc' || 's390x-linux-gnu-gcc' }}
             configure-args: --warn --static --with-dfltcc-deflate --with-dfltcc-inflate
-            chost: ${{ github.repository != 'zlib-ng/zlib-ng' &&  's390x-linux-gnu' || '' }}
-            packages: ${{ github.repository != 'zlib-ng/zlib-ng' && 'qemu qemu-user gcc-s390x-linux-gnu g++-s390x-linux-gnu libc-dev-s390x-cross' || '' }}
+            chost: ${{ github.repository != 'zlib-ng/zlib-ng' && 's390x-linux-gnu' || '' }}
+            packages: qemu qemu-user gcc-s390x-linux-gnu g++-s390x-linux-gnu libc-dev-s390x-cross
             cflags: ${{ github.repository != 'zlib-ng/zlib-ng' && '-static' || '' }}
             ldflags: ${{ github.repository != 'zlib-ng/zlib-ng' && '-static' || '' }}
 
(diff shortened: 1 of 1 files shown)


nmoinvaz:
@Dead2 I previously did this but it was not merged properly into the `develop` branch in #1640.
https://github.com/zlib-ng/zlib-ng/pull/1640/commits/eb41c8ce9b4d855c0abbdb56e6dd4f5f515009fd

nmoinvaz:
@mtl1979 your PR is missing some parameters I had in my commit.

mtl1979:
> @mtl1979 your PR is missing some parameters I had in my commit.

I wasn't trying to copy your PR... I just made minimal changes so the build passes... Adding `-static` manually isn't really a good idea when `--static` is already passed to `configure`.

Dead2:
@mtl1979 I merged #1640 since it was already reviewed and was supposed to already be in.

You can rebase on top of that if there are changes you still want to merge 👍

mtl1979:
> @mtl1979 I merged #1640 since it was already reviewed and was supposed to already be in.
> 
> You can rebase on top of that if there are changes you still want to merge 👍

Rebased... Now it's mostly style cleanup left, so adjusted the commit message and PR description.

.github/workflows/configure.yml
```diff
@@ -216,15 +216,15 @@ jobs:
         show-progress: 'false'
 
     - name: Add ubuntu mirrors
-      if: runner.os == 'Linux' && matrix.packages
+      if: runner.os == 'Linux' && matrix.packages && !contains(matrix.os, 'z15')
```
Dead2, line 219: **thought:** 
These are both valid ways to do this, but personally I prefer the existing way as it doesn't add special handling for specific actions runners to these tasks.

These are both skipped when `matrix.packages` is empty already.

.github/workflows/configure.yml
```diff
@@ -171,17 +171,17 @@ jobs:
...
+            packages: qemu qemu-user gcc-s390x-linux-gnu g++-s390x-linux-gnu libc-dev-s390x-cross
             cflags: ${{ github.repository != 'zlib-ng/zlib-ng' && '-static' || '' }}
             ldflags: ${{ github.repository != 'zlib-ng/zlib-ng' && '-static' || '' }}
 
           - name: Ubuntu GCC S390X DFLTCC Compat
             os: ${{ github.repository == 'zlib-ng/zlib-ng' && 'z15' || 'ubuntu-latest' }}
             compiler: ${{ github.repository == 'zlib-ng/zlib-ng' && 'gcc' || 's390x-linux-gnu-gcc' }}
             configure-args: --warn --zlib-compat --static --with-dfltcc-deflate --with-dfltcc-inflate
-            chost: ${{ github.repository != 'zlib-ng/zlib-ng' &&  's390x-linux-gnu' || '' }}
-            packages: ${{ github.repository != 'zlib-ng/zlib-ng' && 'qemu qemu-user gcc-s390x-linux-gnu g++-s390x-linux-gnu libc-dev-s390x-cross' || '' }}
```
Dead2, line 184: **thought:** 

All these conditional statements are a bit messy, I agree, but there is no gain from getting rid of just one of them.
One option to simplify this could perhaps be to preprocess the check for repo name, and store into a variable like `qemu_s390x_dfltcc`. Then these checks would all be shorter and a little easier to read, and also easier for a third party to change if they wanted to run on real s390x.

mtl1979: I just wanted to make minimal changes and keep configure and Cmake workflows as similar as possible, so making further changes is easier... When more forks start running on real hardware, it will be easier to make the required changes.

.github/workflows/configure.yml
mtl1979: I know the end result is the same, but there really isn't any justification for configure workflow to do it in one way and cmake workflow to do in another way.

Dead2: **suggestion:**
True enough, then perhaps the cmake workflow should be changed instead.

mtl1979: > **suggestion:**
> True enough, then perhaps the cmake workflow should be changed instead.

You are free to open own PR for that but that doesn't mean everyone will agree that it will be any cleaner or better solution.

PR #1687, opened 2024-02-21
Add dependabot for github actions

Dead2:
Lets us get automated monthly PRs updating github actions if needed

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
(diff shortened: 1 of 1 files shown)


nmoinvaz (approved):

mtl1979:
I used to use dependabot, but the PRs it made were incomplete and thus unusable... Then I made my own version, which actually worked better...

nmoinvaz:
We use dependabot with npm, yarn3, and python3 and it works great. Sometimes there are PRs that I have to manually fix, but it is rare.

PR #1688, opened 2024-02-21
Don't export git/github-related files in tar/zip archives

Dead2:
These files are useless outside of git and github, so don't distribute them to users downloading/exporting tar.gz, zip or other archives. Doing this is relatively common for well-known projects.


nmoinvaz (approved):

mtl1979 (approved):
LGTM

PR #1698, opened 2024-03-12
Fix building with NVHPC

skosukhin:
This makes it possible to build the library with the Nvidia compiler (NVHPC).

1. The value of the `__GNUC__` macro defined by the compiler depends on the version of `gcc` it is configured to work with.
2. Ever since version `20.7` (which looks like [the oldest one](https://developer.nvidia.com/nvidia-hpc-sdk-releases)), the compiler comes with its own headers that declare/define the required functions.

### DIFF
--- a/arch/x86/x86_intrins.h
+++ b/arch/x86/x86_intrins.h
@@ -7,7 +7,7 @@
 #ifdef __AVX2__
 #include <immintrin.h>
 
-#if (!defined(__clang__) && defined(__GNUC__) && __GNUC__ < 10) \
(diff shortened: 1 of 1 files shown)


skosukhin:
The failed tests are not related to these changes, aren't they? At least, I have no idea how that is possible.

Dead2:
No, see #1699

mtl1979:
I'm all for supporting new compiler frontends as long as there is no serious bugs that affect building zlib-ng... IMO, it doesn't really matter if we land this before or after #1699.

nmoinvaz (approved):

Dead2 (approved):

PR #1699, opened 2024-03-15
Fix kernel ASLR breaking LLVM ASAN/MSAN builds

Dead2:
Ubuntu image 20240310 contained a more secure kernel build with increased kernel address randomization settings.
This is unfortunately incompatible with the provided LLVM build, and makes MSAN/ASAN crash with segfaults.

Image update fixing this is ETA March 22.

Ref: https://github.com/actions/runner-images/issues/9491#issuecomment-1989718917

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -568,6 +568,12 @@ jobs:
             cmake-args: -DWITH_NATIVE_INSTRUCTIONS=ON
 
     steps:
+    - name: Kernel ASLR
+      if: runner.os == 'Linux' && !contains(matrix.os, 'z15')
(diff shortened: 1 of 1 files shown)


mtl1979:
We might want to wait until the updated image is available as I don't think we will merge that many pull requests that would possibly cause ASAN/MSAN builds to break and not the others.

Adding and (soon) later reverting the workaround seems counterproductive.

mtl1979:
Is this still needed?

After new Clang/LLVM packages are added to repository, we might consider pinning the minimum versions to make sure all CI instances update to compatible version. FYI, this is similar to what we have previously done with incompatible versions of `wine` dependencies due to `apt` not correctly matching versions across architectures.

PR #1700, opened 2024-03-17
arch/riscv/riscv_features.c: fix uclibc build

ffontaine:
Fix the following uclibc build failure raised since version 2.1.4 and https://github.com/zlib-ng/zlib-ng/commit/6ff8b52cefe56a824fae1d53fdd687bcde2e53c9:

```
/home/autobuild/autobuild/instance-2/output-1/build/zlib-ng-2.1.6/arch/riscv/riscv_features.c:4:10: fatal error: sys/auxv.h: No such file or directory
    4 | #include <sys/auxv.h>
      |          ^~~~~~~~~~~~
```

Fixes:
 - http://autobuild.buildroot.org/results/06a7d8e59ec4de7c711d3f4a4624f67b97d78afe

### DIFF
--- a/arch/riscv/riscv_features.c
+++ b/arch/riscv/riscv_features.c
@@ -1,8 +1,11 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
-#include <sys/auxv.h>
 #include <sys/utsname.h>
+
+#if defined(__linux__) && defined(HAVE_SYS_AUXV_H)
+#  include <sys/auxv.h>
+#endif
 
 #include "zbuild.h"
 #include "riscv_features.h"
@@ -33,7 +36,11 @@ void Z_INTERNAL riscv_check_features_compile_time(struct riscv_cpu_features *fea
 }
 
 void Z_INTERNAL riscv_check_features_runtime(struct riscv_cpu_features *features) {
+#if defined(__linux__) && defined(HAVE_SYS_AUXV_H)
     unsigned long hw_cap = getauxval(AT_HWCAP);
+#else
+    unsigned long hw_cap = 0;
+#endif
     features->has_rvv = hw_cap & ISA_V_HWCAP;
 }
 


mtl1979:
Not all toolchains have `sys/auxv.h`, but the same information might be available in other headers... Just disabling the feature detection doesn't really count as a "proper" fix, it should be the last resort if anything else doesn't work.

ffontaine:
Sorry but I don't understand your comment, I'm basically patching `riscv_features.c` as already done for other architectures in e1baa6b439c06f3d0e6be3e381cc746ffb55aee1. If `getauxval` is not available, there is no way to do any runtime detection. Did I missed something?

mtl1979:
> Sorry but I don't understand your comment, I'm basically patching `riscv_features.c` as already done for other architectures in [e1baa6b](https://github.com/zlib-ng/zlib-ng/commit/e1baa6b439c06f3d0e6be3e381cc746ffb55aee1). If `getauxval` is not available, there is no way to do any runtime detection. Did I missed something?

Latest uclibc release is very old, uclibc-ng does have `sys/auxv.h`. As long as the Linux distribution uses ELF binaries, auxiliary vector (auxv) does exist, even with latest available uclibc version. The issue is only getting the start address of auxiliary vector.

ffontaine:
`getauxval` and `sys/auxv.h` have been added very recently to uclibc-ng. It was done 15 months ago by https://github.com/wbx-github/uclibc-ng/commit/d869bb1600942c01a77539128f9ba5b5b55ad647. This means that most of the "not so old" uclibc-ng toolchains used for embedded systems  will raise a build failure. If this PR is not acceptable, I can also disable zlib-ng with uclibc-ng on buildroot side.

mtl1979:
> `getauxval` and `sys/auxv.h` have been added very recently to uclibc-ng. It was done 15 months ago by [wbx-github/uclibc-ng@d869bb1](https://github.com/wbx-github/uclibc-ng/commit/d869bb1600942c01a77539128f9ba5b5b55ad647). This means that most of the "not so old" uclibc-ng toolchains used for embedded systems will raise a build failure. If this PR is not acceptable, I can also disable zlib-ng with uclibc-ng on buildroot side.

The real issue is that run-time detection in zlib-ng heavily depends on proper detection of CPU features. If people try to use too old toolchain, there might also be other issues that are fixed or might get fixed in later releases. If we short-circuit run-time detection, it means either that the binary crashes if RVV isn't supported, or we always disable RVV even if available.

I'm not completely against as adding hardcoded `0` as last resort fallback, but the real fix in this case really seems to be upgrading the toolchain. We don't support all toolchain versions even on other "more important" architectures. RISC-V support is quite new addition to zlib-ng and it was initially decided that "we" don't support older toolchains or kernels due to having no easy access to auxiliary vector in user applications or libraries.

Dead2 (approved):
Looks ok to me, confined to RiscV with little risk of negative effects.

mtl1979:
There still exist small issue for systems that have neither `getauxval` function or `ISA_V_HWCAP` defined, which would essentially be very old systems (uclibc is very old, uclibc-ng is more recent) or possibly any non-Linux system running on RISC-V (We pretty much only support BSD-compatible or Linux on non-x86/ARM platforms). I'm not really worried about supporting non-ELF systems on RISC-V with this PR... As I said before, this seems incomplete fix.

As I've said before preprocessor should evaluate undefined preprocessor tokens in expressions as `0`, but this depends on if the warnings are suppressed or not, or treated as warnings or errors.

PR #1701, opened 2024-03-20
Allow disabling runtime CPU features detection in tests and benchmarks

phprus:
The ``cpu_check_features`` and arch-specific ``*_check_features`` functions are no longer used and not included in the build if runtime checks are disabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -679,8 +679,12 @@ if(WITH_OPTIM)
                 endif()
             endif()
         endif()
-        list(APPEND ZLIB_ARCH_HDRS ${ARCHDIR}/arm_features.h ${ARCHDIR}/arm_functions.h)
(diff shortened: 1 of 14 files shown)


mtl1979 (approved):
LGTM. Unoptimized C version and one optimized version is enough when run-time detection is disabled.

phprus:
> Unoptimized C version and one optimized version is enough when run-time detection is disabled.

This is non-trivial because some optimized functions depend on their less optimized variants.

mtl1979:
> non-trivial

I meant that having unoptimized C-version as baseline for benchmark is enough... we don't need all the other variants as individual benchmarks when run-time detection is disabled. The other option would have been to completely disable building benchmarks.

nmoinvaz (approved):

Dead2 (approved):

PR #1707, opened 2024-04-05
CMake: Replace `;` by `$<SEMICOLON>` in generator-expression

DenizThatMenace:
CMake generator-expressions should not contain semicolons, especially if they might end up in a CMake list, because a semicolon (`;`) would be interpreted as list-item separator and therefore render the generator-expression invalid. The generator-expression `$<SEMICOLON>` should be used instead.


nmoinvaz:
Looks ok to me

nmoinvaz (approved):

Dead2 (approved):

mtl1979 (approved):

PR #1711, opened 2024-04-09
Update dependencies for 32-bit MinGW CI run.

mtl1979:
ODBC packages were updated, so version mismatch was introduced between libodbc1 and libodbc2. Now we pin versions for both of them.


nmoinvaz (approved):

nmoinvaz:
Thanks for looking into this.

mtl1979:
> Thanks for looking into this.

I've updated the dependencies in the past when they updated the packages... I'm still waiting for them to fix the multi-arch dependency resolving in `apt`.

Dead2 (approved):

PR #1712, opened 2024-04-11
.gitattributes: Do no longer mess with line-endings and leave as is

DenizThatMenace:
Although Git is able to automatically modify line-endings during checkin and checkout, this brings a lot of trouble, especially when trying to use a repository from different platforms (as Windows and Linux).  
**Therefore, Git should leave line-endings as is and should not automatically modify them during checkin and checkout.**

_It is the committer's responsibility to commit files with correct line-endings (preferably `LF`) and the reviewer's responsibility to ensure that no wrong line-endings are introduced through pull-requests._

Probably, some tool like `clang-format` or an `.editorconfig` file for text-editors and IDEs should enforce line-endings.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,11 +1,12 @@
-* text=auto
-*.abi text eol=lf
-*.c text
-*.h text
-*.sh text eol=lf
+# Do NO line-ending conversion on any file (by default)!
+* -text
+# Note: If one wants a specific line-ending on a file, one has to make sure to commit it
+#       with the correct line-endings! If at some point all files are committed with
+#       correct line-endings one should use some other mechanism (e.g. clang-format or
+#       .editorconfig) to ensure that no different line-endings are introduced.
+
+
 crc32_braid_tbl.h hooks-max-size=1000000
-Makefile text
-configure text eol=lf
 
 # Don't export git/github-related files in tar/zip archives
 /.github export-ignore  


mtl1979:
This is generally a bad idea... When the line-endings are wrong, it is not guaranteed that the repository has the expected line-ending for shell scripts and other files that don't work with CR+LF combination. We would need to store even text files as binary files, which would make comparing changes next to impossible. In the worst case, the diff between two revisions will show whole file as replaced with completely different content.

DenizThatMenace:
> This is generally a bad idea...
> ...
> In the worst case, the diff between two revisions will show whole file as replaced with completely different content.

That would mean that line-endings modification slipped through the review.

> When the line-endings are wrong, it is not guaranteed that the repository has the expected line-ending for shell scripts and other files that don't work with CR+LF combination.

So you mean you want to enforce that all (non-binary) files have `LF` line-endings in the repository (aka what gets pushed to Github)?  
Then, of course, using the `* text=auto` rule is correct.

However, I hope, you don't want to have different line-endings for all these (non-binary) files in the workspace then when on Windows or Linux? Because that is what we currently are facing. And that makes a patch created on one platform not apply on the other platform and vice-versa.

The reason is that the above rule only changes line-endings to `LF` during **checkin**, but during **checkout** instead considers the user's `.gitconfig` and in particular `core.autocrlf`. And if that is set to `false` (because the user does not want automatic line-endings conversion to `CR+LF`) the value of `core.eol` is used instead which defaults to `native`. And that again results in automatic conversion to `CR+LF` on Windows (which is undesirable).

What you should use instead is to also provide a value for `eol`:
```
* text=auto eol=lf
```

Thereby, files that Git considers as non-binary will get their line-endings changed to `LF` during checkin and will keep these during checkout. (The only requirement is to use at least Git 2.10 or newer, because older versions had a bug where the `=auto` was silently dropped from the rule.)


> We would need to store even text files as binary files, which would make comparing changes next to impossible.

I don't understand what you mean here.  
We were talking about automatic line-ending conversions done by Git which is what the `text` attribute is used for. Marking all files as "binary" is nonsense, because then you effectively unset the `diff` attribute (by using `-diff` in the rule).

Note: Unsetting the `text` attribute (by using `-text` in the rule) does not mark the matching files as binary but instead just prevents any automatic line-ending conversions on these files.

nmoinvaz:
No project I have ever worked on with git has had to specify `text` for `.c` & `.h` files. I proposed removing similar lines from .gitattributes before. It was because I ran into issues with Windows using Unix LF mode where certain files were being checked out still in Windows mode CRLF due to `text`.

image
image
image

.gitattributes
```diff
@@ -1,11 +1,12 @@
-* text=auto
-*.abi text eol=lf
-*.c text
-*.h text
-*.sh text eol=lf
+# Do NO line-ending conversion on any file (by default)!
+* -text
```
nmoinvaz, line 2: Why not just remove this?

Dead2:
Honestly, I would love if we could just let .c/.h files go `text eol=lf`, if windows and mac users are OK with that.
Any other files we could lock down to a single kind of line endings is also good with me.

Something I am opposed to is removing all handling of line endings and leaving it up to reviewers to spot any wrong line endings, because sooner or later that will probably lead to stray incorrect line endings.
IMO the best solution is one where git ensures it is committed correctly. A solution where just CI fails due to incorrect LE is too late in the process, since it requires reviewers and the committer to do extra work to fix it (and possibly it then gets fixed in a different commit than it was introduced, etc).

I don't think the current proposal in this PR helps unfortunately.

DenizThatMenace:
> Honestly, I would love if we could just let .c/.h files go `text eol=lf`, if windows and mac users are OK with that. Any other files we could lock down to a single kind of line endings is also good with me.
> ...
> I don't think the current proposal in this PR helps unfortunately.

I opened #1715 as an alternative approach.

Dead2:
#1715 merged instead

PR #1713, opened 2024-04-23
Rewrite deflate memory allocation.

### LINKED ISSUE #1708: Linux x64: segfault by Unity/Mono
Replacing zlib with the zlib-ng compatibility library consistently causes a segfault in the game "Kerbal Space Program" (built on Unity 2019.4) in certain situations.

Coredump backtrace:
```
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007fd9344ab163 in __pthread_kill_internal (threadid=<optimized out>, signo=6) at pthread_kill.c:78
#2  0x00007fd93445365e in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007fd93443b902 in __GI_abort () at abort.c:79
#4  0x00007fd9350eef0f in HandleSignal(int, siginfo_t*, void*) () from /home/jonas/games/steam/steamapps/common/Kerbal Space Program/UnityPlayer.so
#5  0x00007fd8dcf51a42 in ?? () from /home/jonas/games/steam/steamapps/common/Kerbal Space Program/KSP_Data/MonoBleedingEdge/x86_64/libmonobdwgc-2.0.so
#6  0x00007fd8dce5b9f4 in ?? () from /home/jonas/games/steam/steamapps/common/Kerbal Space Program/KSP_Data/MonoBleedingEdge/x86_64/libmonobdwgc-2.0.so
#7  <signal handler called>
#8  0x00007fd9344b9d05 in __GI___libc_free (mem=0x10011) at malloc.c:3375
#9  0x00007fd9338b8964 in z_free_aligned (ptr=<optimized out>, opaque=<optimized out>, zfree=<optimized out>)
    at /usr/src/debug/zlib-ng-2.1.6-2.fc40.x86_64/zutil.c:158
#10 deflateEnd (strm=0x6fd5fa0) at /usr/src/debug/zlib-ng-2.1.6-2.fc40.x86_64/deflate.c:1018
#11 0x00007fd7bf61abfa in CloseZStream ()
   from /home/jonas/games/steam/steamapps/common/Kerbal Space Program/KSP_Data/MonoBleedingEdge/x86_64/libMonoPosixHelper.so
#12 0x00000000416daead in ?? ()
#13 0x0000000000000000 in ?? ()
```
(Sorry, this is a proprietary app, so several symbols do not have debug information.)

System:
GNU/Linux / Fedora 40 (Beta) / Kernel 6.8.4
CPU: AMD Ryzen 7 7840HS

Please tell me what further information I can provide.

Dead2:
This PR currently lacks s390x support entirely. @iii-i is working on updating that support.

This PR moves all allocations to come from a single buffer allocated during init, this has many positive sides such as being faster overall, and in the case of inflate; ensuring that as much as possible of the startup cost is done during init, leading to more predictable behavior and performance of inflate itself.
I have also added debug logging to the allocation functions, as it is now simple and self-contained, although it could likely use some improvements to be more portable.
This should also fix #1708 and similar errors, although that is not the motivation for these changes.

## Deflate
Deflate used to call allocate 5 times during init.

- 5 calls to external alloc function now becomes 1
- Handling alignment of allocated buffers is simplified
  - Efforts to align the allocated buffer now needs to happen only once.
  - Individual buffers are ordered so that they have natural sequential alignment.
- Due to reduced losses to alignment, we allocate less memory in total.
- While doing alloc(), we now store pointer to corresponding free(), avoiding crashes
  with applications that incorrectly set alloc/free pointers after running init function.
- Removed need for extra padding after window, chunked reads can now go beyond the window
  buffer without causing a segfault.
 
 ## Inflate
  Inflate used to allocate state during init, but window would be allocated
when/if needed and could be resized and that required a new free/alloc round.

- Now, we allocate state and a 32K window during init, allowing the latency cost
  of allocs to be done during init instead of at one or more times later.
- Total memory allocation is about the same when requesting a 32K window, but
  if now window or a smaller window was requested, then it is an increase.
- While doing alloc(), we now store pointer to corresponding free(), avoiding crashes
  with applications that incorrectly set alloc/free pointers after running init function.
- After init has succeeded, inflate will no longer possibly fail due to a failing malloc.

## Performance
Preliminary benchmarks show that compress() of 1 byte is ~9% faster.
Compress() of 32KiB is 3.6% faster.
Bigger files show less improvements due to most of the savings happening during deflate init().
Benchmark used is in PR #1721


zbuild.h
```diff
@@ -202,6 +202,22 @@
...
+#ifdef HAVE_BUILTIN_ASSUME_ALIGNED
+#  define HINT_ALIGNED(p,n) __builtin_assume_aligned((void *)(p),(n))
+#else
+#  define HINT_ALIGNED(p,n) (p)
+#endif
+#define HINT_ALIGNED_64(p) HINT_ALIGNED((p),64)
+#define HINT_ALIGNED_16(p) HINT_ALIGNED((p),16)
+
+/* PADSZ returns needed bytes to pad bpos to pad size
+ * PAD_NN caluclates pad size and adds it to bpos, returning the result.
```
mtl1979: "calculates"

Dead2:
@iii-i Any chance you have the time for commenting on the feasibility for porting the DFLTCC code to using this allocator and removing most of the custom allocation code? Any obvious problems that need to be caught early?

I have looked at the code, and I think it is doable, but I don't really know that part of the code at all, and it'd probably require trial-and-error to implement this I think. If you want to contribute a patch, that'd be awesome too. Worst case (I really want to avoid doing that though) is doing a point release without s390x DFLTCC support while it gets brought up to speed.

inflate.c
```diff
@@ -120,21 +120,98 @@ int32_t Z_EXPORT PREFIX(inflateReset2)(PREFIX3(stream) *strm, int32_t windowBits
...
+    LOGSZP("state", state_size, PAD_64(currsize), PADSZ(currsize,64));
+    int state_pos = PAD_64(currsize);
+    currsize = state_pos + state_size;
+
+    LOGSZP("alloc", alloc_size, PAD_16(currsize), PADSZ(currsize,16));
+    int alloc_pos = PAD_16(currsize);
+    currsize = alloc_pos + alloc_size;
+
+    /* Add 64-1 or 4096-1 to allow window alignment, and round size of buffer up to multiple of 64 */
+    int totalsize = PAD_64(currsize + (WINDOW_PAD_SIZE - 1));
```
nmoinvaz: total_size?

inflate.c
```diff
@@ -120,21 +120,98 @@ int32_t Z_EXPORT PREFIX(inflateReset2)(PREFIX3(stream) *strm, int32_t windowBits
...
+
+    /* Allocate buffer, align to 64-byte cacheline, and zerofill the resulting buffer */
+    char *origbuff = strm->zalloc(strm->opaque, 1, totalsize);
```
nmoinvaz: `original_buf`?

deflate.c
```diff
@@ -71,14 +71,15 @@ const char PREFIX(deflate_copyright)[] = " deflate 1.3.1 Copyright 1995-2024 Jea
  */
 #ifdef S390_DFLTCC_DEFLATE
 #  include "arch/s390/dfltcc_deflate.h"
+#  define PAD_WINDOW            PAD_4096
+#  define WINDOW_PAD_SIZE       4096
```
nmoinvaz, line 76: Why is S390 pad size so high? Maybe a comment about it?

test/benchmarks/benchmark_compress.cc
```diff
@@ -0,0 +1,67 @@
...
+    }
+};
+
+#define BENCHMARK_COMPRESS(name) \
+    BENCHMARK_DEFINE_F(compress, name)(benchmark::State& state) { \
+        Bench(state); \
+    } \
+    BENCHMARK_REGISTER_F(compress, name)->Arg(1)->Arg(8)->Arg(16)->Arg(32)->Arg(64)->Arg(512)->Arg(4<<10)->Arg(32<<10);
+
+BENCHMARK_COMPRESS(compress);
```
nmoinvaz: Usually benchmark would test multiple versions of similar functions, but here there is only one. What is the purpose of this file?

Dead2: Before this PR:
```
compress/compress/1           6157 ns         6147 ns       105599
compress/compress/8           6666 ns         6654 ns       100625
compress/compress/16          6873 ns         6862 ns        99180
compress/compress/32          7663 ns         7650 ns        89458
compress/compress/64          8265 ns         8251 ns        83929
compress/compress/512         8208 ns         8194 ns        82918
compress/compress/4096        9570 ns         9553 ns        72661
compress/compress/32768      19308 ns        19272 ns        36328
```

With this PR:
```
compress/compress/1           5618 ns         5610 ns       113507
compress/compress/8           5984 ns         5974 ns       109259
compress/compress/16          6097 ns         6087 ns       111210
compress/compress/32          6828 ns         6817 ns       100269
compress/compress/64          7435 ns         7424 ns        91821
compress/compress/512         7417 ns         7405 ns        91981
compress/compress/4096        8785 ns         8771 ns        79245
compress/compress/32768      18541 ns        18510 ns        37689
```

PS: Neither of these are up to date, but they do illustrate the point of using them for comparisons across different commits.

deflate.c
Dead2: Window needs to be page-aligned for DFLTCC instruction it seems.
A comment is a good idea.

test/benchmarks/benchmark_compress.cc
Dead2: Benchmarking this PR with deflatebench is pretty useless, it has too much overhead to reliably distinguish speed differences for compressing/decompressing such small files. It can be seen on the total runtime over hundreds of runs on small files.
This is a much better benchmark for that kind of thing since it runs the benchmark a lot closer to the library itself and without all the forking and external monitoring and such.

inflate.c
Dead2: I am not sure about this one.
`totalsize` is related to the "counter" `currsize`.
`*_size` are sizes for individual buffers.

It might be a silly way to differentiate between them, since it is more usual to choose one style or the other.
Alternatively I could rename them to `tsize` and `csize`, but I feel like that loses some of the context.

I can change it to `total_size` and `curr_size` if that is more readable. Thoughts?

deflate.c
iii-i: I can confirm that window needs to be page-aligned, so 4k is the right value.

iii-i:
Hi, I think there are no huge compatibility problems for DFLTCC in this PR. I currently have a draft commit that adapts the code and survives the regtest: https://github.com/iii-i/zlib-ng/releases/tag/single-malloc-dfltcc-20240430.

Dead2:
> Hi, I think there are no huge compatibility problems for DFLTCC in this PR. I currently have a draft commit that adapts the code and survives the regtest: [iii-i/zlib-ng@`single-malloc-dfltcc-20240430` (release)](https://github.com/iii-i/zlib-ng/releases/tag/single-malloc-dfltcc-20240430).

That is great. Maybe you should do a PR for that based on top of this.
When looking through it, I notice it looks like a pretty substantial cleanup 👍.
I only noticed one thing I'd prefer changed; in deflate.h, move `arch_deflate_state arch;` ~10 lines up, to right below `deflate_allocs *alloc_bufs;`. That group has a natural 16-byte alignment, it should not have any negative effect whether it is there or not, and it avoids having it dangling at the end after the padding bytes 😄.
There might also be a potential for getting rid of `dfltcc_common.c`, as `ZCOPY_WINDOW` is only used from inflate.

iii-i:
Will do. Unfortunately I found some issues which I don't fully understand yet; but now I have at least a couple bite-sized commits that make sense on their own, which I will post as separate PRs (here is the first one: https://github.com/zlib-ng/zlib-ng/pull/1717).

inflate.c
nmoinvaz: `total_size` and `curr_size` is more readable imo. I always think that `_` underscore should separate words. It also makes it easy say if you were in the future to try to convert from when casing style to another. (hypothetical scenario).

nmoinvaz:
Interesting removal of `ZCOPY_DEFLATE_STATE` and the one for inflate also. Is it no longer necessary? I wonder what kind of optimization DFLTCC was doing.

Configure script changes LGTM.

Dead2:
> Interesting removal of `ZCOPY_DEFLATE_STATE` and the one for inflate also. Is it no longer necessary? I wonder what kind of optimization DFLTCC was doing.

I was initially uncertain whether to keep those around or not, since I was unsure what DFLTCC needed to do. But when @iii-i posted his initial patch that also removed those, I figured I might as well do so in this patch where it fits with the other changes in the same areas, and then the s390x changes will be more self-contained and easier to review as well.

Dead2:
Rebased

iii-i:
`ZCOPY_DEFLATE_STATE` is not an optimization - DFLTCC needs to maintain state of its own, it is allocated alongside deflate/inflate state, and needs to be copied on `deflateCopy()` / `inflateCopy()`. In order to make things as unintrusive as possible, in my zlib patch I chose not to modify the structs, but rather to introduce hooks and do pointer arithmetics there. This rewrite is an opportunity to simplify things (https://github.com/zlib-ng/zlib-ng/pull/1718) at the cost of making DFLTCC support a little bit more intrusive, which is hopefully okay.

iii-i:
P.S. I wonder in which order this and #1718 should go in? I'd prefer to land #1718 first, since it's more local and as far as I'm concerned is ready.

inflate.c
Dead2: Changed

phprus:
Fix for ``arch/s390/dfltcc_detail.h:222:12: error: call to undeclared function 'ZALLOC'``: #1728
Fix for CI: #1726, #1725

KungFuJesus:
So, putting on some exploit mitigation goggles, I do worry about this one a tiny bit:

> While doing alloc(), we now store pointer to corresponding free(), avoiding crashes with applications that incorrectly set alloc/free pointers after running init function.

Doesn't this mean we effectively are giving a user controlled pointer for every allocation?  Granted we were already doing that at library initialization time, I'm just wondering if giving this to something possibly controllable by the user, not just the developer, becomes a bit dangerous.  I get that this makes us resilient to a pretty brutal crash from mono but...it just seems like it might be easy to craft a gzip or png file that could readily exploit this.  Yes, I realize that we are the allocators here and not the user, I'm just wondering if a well crafted input file could lie about remaining lengths such that we end up writing a malicious location behind a buffer. This is now several pointers that are vulnerable, rather than just one in the init structure that is outside of the purview of the user.  Am I wrong or overly paranoid here?

Dead2:
@KungFuJesus I'm not sure this PR really changes that.

The free pointer is set during init() both now and before, but now we copy it to somewhere where the application using zlib can't easily change it after init() (at least without accessing the structs and buffers that have no exported interfaces to make sense of them).

> Doesn't this mean we effectively are giving a user controlled pointer for every allocation? Granted we were already doing that at library initialization time, I'm just wondering if giving this to something possibly controllable by the user, not just the developer, becomes a bit dangerous.

How does this become controllable by the user? The developer of the application using zlib needs to set this through the `z_stream` struct during init(). Who is the user here?

> This is now several pointers that are vulnerable, rather than just one in the init structure that is outside of the purview of the user.

I am not sure what you mean by the init structure is outside the purview of the user. The "init structure" or `z_stream` is passed to zlib for every function call from the application, and contains the pointer both to `free`, `malloc` and to the `internal state` struct. (And the `internal state` struct also contains a pointer back to `z_stream`).

We also have other function pointers in structs in various places.
I'm not sure what makes this pointer to `free` any worse than those.


I fear we'd have to do a lot more invasive changes if we wanted to harden all our pointers from attacks. Possibly we could use `mproctect()` / `VirtualProtect()`, but that'd mean moving all function pointers to a separate page-sized+aligned buffer. Also it'd interfere with things like changing deflate levels.

KungFuJesus:
Sorry I realize some of this was poorly articulated. Looking at the code now, I see it's set in the strm struct, not per allocation as I had anticipated based on that summary. We're no worse off for that. Had it been a pointer at the beginning of the allocation it might have been another story (though, I think the inflate code always writes forward in the buffer, never backward). 

So as it stands now (after this PR), if the user tries to switch out the zalloc/zfree functions after init (the situation that is causing the grief, of course), we ignore the values and use the matching free that it should be?

Dead2:
> So as it stands now (after this PR), if the user tries to switch out the zalloc/zfree functions after init (the situation that is causing the grief, of course), we ignore the values and use the matching free that it should be?

Yes, assuming of course that the application doesn't set a mismatching `free` in the first place, but that would be a very clear bug in the application.

Dead2:
@iii-i I am getting ready to merge this, did you get anywhere with updating the DFLTCC support?

iii-i:
I'm currently rebasing my DFLTCC change and I believe I found a bug with my fuzzer (see #1731 for a reproducer): `inflateCopy()` reuses the old window pointer. I propose the following fixup to this PR:

```
--- a/inflate.c
+++ b/inflate.c
@@ -1405,7 +1405,6 @@ int32_t Z_EXPORT PREFIX(inflateCopy)(PREFIX3(stream) *dest, PREFIX3(stream) *sou
     if (alloc_bufs == NULL)
         return Z_MEM_ERROR;
     copy = alloc_bufs->state;
-    copy->window = alloc_bufs->window;
 
     /* copy state */
     memcpy(copy, state, sizeof(struct inflate_state));
@@ -1416,6 +1415,7 @@ int32_t Z_EXPORT PREFIX(inflateCopy)(PREFIX3(stream) *dest, PREFIX3(stream) *sou
     }
     copy->alloc_bufs = alloc_bufs;
+    copy->window = alloc_bufs->window;
 
     /* window */
     memcpy(copy->window, state->window, INFLATE_ADJUST_WINDOW_SIZE((size_t)state->wsize));
```

Edit: the current version is here: https://github.com/iii-i/zlib-ng/releases/tag/single-malloc-dfltcc-20240521. It's based on this PR + fixup + cleanup from @phprus + the reproducer. I think it would be better if you could just squash the DFLTCC commit into this PR, since otherwise bisect will be broken on s390x in the future.

Dead2:
@iii-i Merged in your changes, and squashed one of my commits into the others to try to make things always be in a working state for each commit. Please have a look in case I made any merging errors.

CMakeLists.txt
```diff
@@ -484,6 +484,22 @@ if(HAVE_ATTRIBUTE_ALIGNED)
     add_definitions(-DHAVE_ATTRIBUTE_ALIGNED)
 endif()
 
+#
+# Check for __builtin_assume_aligned(x,n) support in the compiler
+#
+check_c_source_compiles(
+    "char *test(char *buffer) {
+         char *abuffer = __builtin_assume_aligned(buffer,64);
```
iii-i: Inconsistent padding.

deflate.c
```diff
@@ -177,12 +183,106 @@ static const config configuration_table[10] = {
...
+    /* Define sizes */
+    int window_size = DEFLATE_ADJUST_WINDOW_SIZE((1 << windowBits) * 2);
+    int prev_size = (1 << windowBits) * sizeof(Pos);
+    int head_size = HASH_SIZE * sizeof(Pos);
+    int pending_size = lit_bufsize * LIT_BUFS;
+    int state_size = sizeof(deflate_state);
+    int alloc_size = sizeof(deflate_allocs);
+
+    /* Calculate relative buffer positions and paddings */
+    LOGSZP("window", window_size, PAD_WINDOW(curr_size), PADSZ(curr_size,WINDOW_PAD_SIZE));
```
iii-i, line 214: Missing space after `curr_size,`.

iii-i (approved):
I tested the PR on s390x and everything look good. Going over code, I noticed a couple nits, but otherwise LGTM.

CMakeLists.txt
Dead2: Is it though? I don't see it.

deflate.c
Dead2: That was intentional and is used for all those LOGSZP() lines below as well. Since PADSZ() is nested within LOGSZP() I found it easier to see what parameters belong to what macro when the contents of PADSZ() is grouped together.
If you really think it needs to be changed, I can do so, it just seems neater to my preferences.

CMakeLists.txt
iii-i: I see 9 spaces before `char *abuffer` and 8 before `return`.

deflate.c
iii-i: I don't have a strong opinion here, I just noticed a discrepancy in style. Let's leave it as is.

CMakeLists.txt
Dead2: There it was, thanks 🤦‍♂️

nmoinvaz:
Recent changes look good.

nmoinvaz:
I think `PORTING.md` should be updated to indicate that we will always allocate enough memory for the max window size and the reason why.

Dead2:
Should be ready for final reviews now

PORTING.md
```diff
@@ -4,6 +4,38 @@ Porting applications to use zlib-ng
...
+Zlib-ng's deflate will however allocate a lower amount of memory depending
+on compression level and window size.
+
+Zlib-ng also allocates one "big" buffer instead of doing multiple smaller
+allocations. This is faster, can lead to better cache locality and reduces
+space lost to alignment padding.
+
+At the time of writing, by default zlib-ng allocates the following amounts
+of memory on a 64-bit system (except on S390x that requires ~4KiB more):
+-  Deflate: 350.272 Bytes
```
nmoinvaz, line 23: This is for the struct only or for the window and pending buffers?

deflate.c
```diff
@@ -177,12 +183,106 @@ static const config configuration_table[10] = {
...
+
+    LOGSZP("prev", prev_size, PAD_64(curr_size), PADSZ(curr_size,64));
+    int prev_pos = PAD_64(curr_size);
+    curr_size = prev_pos + prev_size;
+
+    LOGSZP("head", head_size, PAD_64(curr_size), PADSZ(curr_size,64));
+    int head_pos = PAD_64(curr_size);
+    curr_size = head_pos + head_size;
+
+    LOGSZP("pending", pending_size, PAD_64(curr_size), PADSZ(curr_size,64));
```
nmoinvaz, line 226: Consider putting pending after the window? Maybe it will be better it doesn't have to seek backwards?

PORTING.md
Dead2: This is all we allocate, everything.
```
window is 65536 bytes, offset 0, padded 0
prev is 65536 bytes, offset 65536, padded 0
head is 131072 bytes, offset 131072, padded 0
pending is 81920 bytes, offset 262144, padded 0
state is 6080 bytes, offset 344064, padded 0
alloc is 56 bytes, offset 350144, padded 0
Buffer alloc is 350272 bytes, offset 0, padded 8
```

deflate.c
Dead2: This made no discernible difference that I can see, maybe because they are already bigger than pagesize so they each take up multiple TLBs already.

Not sure what you mean by seek backwards though.

PORTING.md
nmoinvaz: I see, I misread the period since I am so used to seeing a comma..

nmoinvaz (approved):

PR #1715, opened 2024-04-24
.gitattributes: Enforce LF line-endings on all non-binary files

DenizThatMenace:
Although Git is able to automatically modify line-endings during checkin and checkout, this brings a lot of trouble, especially when trying to use a repository from different platforms (as Windows and Linux). This is due to the fact that Git consults different local gitconfig settings and uses problematic defaults if not set.

Therefore, Git should enfoce one type of line-ending (LF) and not consult the local config, which is what the change from this commit does.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,11 +1,7 @@
-* text=auto
-*.abi text eol=lf
-*.c text
-*.h text
-*.sh text eol=lf
+# By default, enforce LF line-endings on all files that are not considered binary files!
+* text=auto eol=lf
+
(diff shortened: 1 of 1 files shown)


Dead2:
This looks good to me, but I'd like to hear from someone that use mac and windows as well.

nmoinvaz:
I use Windows primary and it seems ok to me. I already have a local setting that forces use of LF and haven't had any problems with it.

Dead2 (approved):

PR #1717, opened 2024-05-02
IBM zSystems DFLTCC: Extend sanitizer checks

iii-i:
Currently the DFLTCC sanitizer instrumentation is limited to MSAN-unpoisoning the parameter block. Add ASAN and MSAN checks; also MSAN-unpoison the window.

Introduce the generic instrument_read(), instrument_write() and instrument_read_write() macros, that are modeled after the repsective functions in the Linux kernel.

### DIFF
--- a/arch/s390/dfltcc_detail.h
+++ b/arch/s390/dfltcc_detail.h
@@ -162,6 +162,30 @@ typedef enum {
 #define HB_BITS 15
 #define HB_SIZE (1 << HB_BITS)
 
+/* Return lengths of high (starting at param->ho) and low (starting at 0) fragments of the circular history buffer. */
+static inline void get_history_lengths(struct dfltcc_param_v0 *param, size_t *hl_high, size_t *hl_low) {
+    *hl_high = MIN(param->hl, HB_SIZE - param->ho);
+    *hl_low = param->hl - *hl_high;
+}
+
+/* Notify instrumentation about an upcoming read/write access to the circular history buffer. */
+static inline void instrument_read_write_hist(struct dfltcc_param_v0 *param, void *hist) {
+    size_t hl_high, hl_low;
+
+    get_history_lengths(param, &hl_high, &hl_low);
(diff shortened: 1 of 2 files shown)


arch/s390/dfltcc_detail.h
```diff
@@ -162,6 +162,30 @@ typedef enum {
 #define HB_BITS 15
 #define HB_SIZE (1 << HB_BITS)
 
+static inline void get_history_lengths(struct dfltcc_param_v0 *param, size_t *hl_high, size_t *hl_low)
+{
```
nmoinvaz: nit: These braces should be on previous line..  for the next 3 functions too.

zbuild.h
```diff
@@ -254,7 +279,28 @@
 #endif
 
 #ifndef Z_MEMORY_SANITIZER
+#  define __msan_check_mem_is_initialized(a, size) do { Z_UNUSED(a); Z_UNUSED(size); } while (0)
 #  define __msan_unpoison(a, size) do { Z_UNUSED(a); Z_UNUSED(size); } while (0)
 #endif
 
+#define instrument_read(a, size) do {             \
```
Dead2, line 287: **suggestion:** 
I think these three functions could use a comment to clarify what they do/what they are for.
I am not familiar with these from the linux kernel, so I only have a vague ide of what I think they do.

arch/s390/dfltcc_detail.h
```diff
@@ -162,6 +162,27 @@ typedef enum {
 #define HB_BITS 15
 #define HB_SIZE (1 << HB_BITS)
 
+static inline void get_history_lengths(struct dfltcc_param_v0 *param, size_t *hl_high, size_t *hl_low) {
```
Dead2, line 166: **nitpick:** 
These three functions are not in such a central part of the code, and are somewhat easier to figure out, but I'd still appreciate if they got a one-liner comment each.

Dead2 (approved):
LGTM

PR #1718, opened 2024-05-02
IBM zSystems DFLTCC: Inline DLFTCC states into zlib states

iii-i:
Currently DFLTCC states are allocated using hook macros, complicating memory management. Inline them into zlib states and remove the hooks.

### DIFF
--- a/deflate.h
+++ b/deflate.h
@@ -14,6 +14,11 @@
 #include "zendian.h"
 #include "crc32.h"
 
+#ifdef S390_DFLTCC_DEFLATE
+#  include "arch/s390/dfltcc_common.h"
+#  define HAVE_ARCH_DEFLATE_STATE
+#endif
+
 /* define NO_GZIP when compiling if you want to disable gzip header and
    trailer creation by deflate().  NO_GZIP would be used to avoid linking in
    the crc code when it is not needed.  For shared libraries, gzip encoding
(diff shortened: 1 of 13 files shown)


deflate.h
```diff
@@ -14,6 +14,12 @@
 #include "zendian.h"
 #include "crc32.h"
 
+#ifdef S390_DFLTCC_DEFLATE
+#  include "arch/s390/dfltcc_common.h"
+#else
+typedef struct {} arch_deflate_state;
```
phprus: Empty struct is not allowed in C.

Add
```c++
char empty;
```
as in a similar case:

https://github.com/zlib-ng/zlib-ng/blob/cd721383bf502c76a00b07e465bc1bc8eaa6c5be/cpu_features.h#L23-L37

deflate.h
```diff
@@ -297,6 +303,8 @@ struct ALIGNED_(16) internal_state {
     /* Reserved for future use and alignment purposes */
     char *reserved_p;
 
+    arch_deflate_state arch;      /* architecture-specific extensions */
```
phprus, line 306: Or do not add ``arch`` member if ``arch_deflate_state`` is empty.

inflate.h
```diff
@@ -13,6 +13,12 @@
 
 #include "crc32.h"
 
+#ifdef S390_DFLTCC_INFLATE
+#  include "arch/s390/dfltcc_common.h"
+#else
+typedef struct {} arch_inflate_state;
```
phprus: Similar issue.

iii-i:
I didn't know about the empty struct restriction, thanks! I guess even if it were allowed, the compiler would still have to implicitly make it at least 1 byte in size in order to give it a distinct address. So I've added an `#ifdef` to omit the arch field.

Dead2 (approved):
LGTM

PR #1720, opened 2024-05-03
Fix macOS Github Actions

### LINKED ISSUE #1719: MacOS CI builds failing
Seems an updated Github Actions MacOS image has broken our CI

From a quick look, I noticed the following:
- Configure fails without any good info.
- pip fails and suggests using pipx or venv.

phprus:
Fixes #1719 

1) New macOS pipelines, based on
``macos-13`` - x86 tests
``macos-latest`` - ARM64 tests

2) Fix ``configure`` script (replaced unconditional add ``-std=c11``).

3) Switch to ``python`` ``venv``.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -712,7 +712,9 @@ jobs:
       if: matrix.codecov
       shell: bash
       run: |
-        python3 -u -m pip install --user gcovr==5.0
+        python3 -u -m venv ${{ matrix.build-dir || '.' }}/venv
+        source ${{ matrix.build-dir || '.' }}/venv/${{ runner.os == 'Windows' && 'Scripts' || 'bin' }}/activate
+        python3 -u -m pip install gcovr==5.0
         python3 -m gcovr ${{ matrix.build-dir || '' }} -j 3 --verbose \
           --exclude-unreachable-branches \
           --gcov-executable "${{ matrix.gcov-exec || 'gcov' }}" \
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -546,20 +546,39 @@ jobs:
             codecov: win64_gcc_compat_no_opt
 
           - name: macOS Clang ASAN
-            os: macos-latest
+            os: macos-13
             compiler: clang
             cxx-compiler: clang++
             cmake-args: -DWITH_SANITIZER=Address
             codecov: macos_clang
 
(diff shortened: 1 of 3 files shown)


phprus:
@Dead2, @nmoinvaz Review this PR please.

Dead2 (approved):
LGTM

.github/workflows/cmake.yml
```diff
@@ -546,21 +546,38 @@ jobs:
...
+            compiler: clang
+            cxx-compiler: clang++
+            cmake-args: -DWITH_SANITIZER=Address
+            codecov: macos_clang_arm64
+
+          - name: macOS GCC UBSAN
+            os: macos-13
             compiler: gcc-10
             cxx-compiler: g++-10
+            ldflags: -ld_classic
```
nmoinvaz, line 568: Why `-ld_classic`?

.github/workflows/cmake.yml
```diff
@@ -546,21 +546,38 @@ jobs:
...
 
-          - name: macOS GCC UBSAN
+          - name: macOS Clang ASAN (ARM64)
             os: macos-latest
+            compiler: clang
+            cxx-compiler: clang++
+            cmake-args: -DWITH_SANITIZER=Address
+            codecov: macos_clang_arm64
+
+          - name: macOS GCC UBSAN
```
nmoinvaz, line 562: Maybe specify "GCC 10"?

phprus: Other pipelines do not show GCC version in it names.

nmoinvaz: Yeah, it looks like even for Clang. Ok. Nvm.

.github/workflows/cmake.yml
phprus: Xcode 15 uses a new linker that is not compatible with GCC.
``-ld_classic`` switches to the old linker.

Homebrew ``gcc@11`` and later have a built-in workaround for this issue:

https://github.com/Homebrew/homebrew-core/blob/c0218d50084e300cba26da84028acfd4917ce623/Formula/g/gcc%4011.rb#L92-L96

nmoinvaz: Oh wow. Ok. Can you add a comment in the yaml about that?

phprus: PR updated.

nmoinvaz (approved):

PR #1721, opened 2024-05-06
Add small compress() benchmark

Dead2:
This minimal benchmark lets us test how long time it takes to do compress a small file.

The 1-byte compression will show the minimal time to initialize, compress and end, something we could use to calculate the maximum count we can do per second.. Lets call it COPS (Compression Operations Per Second).
The bigger sizes will spend more time doing payload compression, useful for adding some context to how much of the time is spent just doing init and end.

This PR lets us compare before/after code changes to see whether it affects COPS.

I intended to make one for decompress() as well, but never got around to it. It would need to compress multiple buffers during SetUp, to be able to do decompression on multiple sizes. It doesn't really need many different sizes though, I suggest `[1, 4096, 32768]`. If anyone wants to make that, I'd appreciate it.

### DIFF
--- a/test/benchmarks/CMakeLists.txt
+++ b/test/benchmarks/CMakeLists.txt
@@ -38,6 +38,7 @@ add_executable(benchmark_zlib
     benchmark_adler32_copy.cc
     benchmark_compare256.cc
     benchmark_compare256_rle.cc
+    benchmark_compress.cc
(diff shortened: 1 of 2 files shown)


Dead2:
Rebased

nmoinvaz (approved):

PR #1723, opened 2024-05-14
Fix invalid instruction usage in Xeon Phi x200 processors

tuliom:
The Xeon Phi x200 family of processors (Knights Landing) supports AVX512 (F, CD, ER, PF) but does not support AVX512 (VL, DQ, BW).

Because of processors like this, the Intel Software Developer's Manual suggests the bits AVX512 (DQ,BW) are also tested in EBX together with AVX512F before deciding to run AVX512 (DQ,BW) instructions.

Considering that adler32_avx512 and adler32_fold_copy_avx512 are both built with -mavx512dq -mavx512bw, it's recommended that both CPUID bits are tested before executing these functions.

This has been reported downstream as
https://bugzilla.redhat.com/show_bug.cgi?id=2280347 .

### DIFF
--- a/arch/x86/x86_features.c
+++ b/arch/x86/x86_features.c
@@ -99,7 +99,16 @@ void Z_INTERNAL x86_check_features(struct x86_cpu_features *features) {
 
         // check AVX512 bits if the OS supports saving ZMM registers
         if (features->has_os_save_zmm) {
-            features->has_avx512 = ebx & 0x00010000;
+            features->has_avx512f = ebx & 0x00010000;
+            if (features->has_avx512f) {
+                // According to the Intel Software Developer's Manual, AVX512F must be enabled too in order to enable
+                // AVX512(DQ,BW,VL).
+                features->has_avx512dq = ebx & 0x00020000;
+                features->has_avx512bw = ebx & 0x40000000;
+                features->has_avx512vl = ebx & 0x80000000;
+            }
+            features->has_avx512_common = features->has_avx512f && features->has_avx512dq && features->has_avx512bw \
+              && features->has_avx512vl;
             features->has_avx512vnni = ecx & 0x800;
         }
     }
--- a/arch/x86/x86_features.h
+++ b/arch/x86/x86_features.h
@@ -8,7 +8,11 @@
 
 struct x86_cpu_features {
     int has_avx2;
-    int has_avx512;
+    int has_avx512f;
+    int has_avx512dq;
+    int has_avx512bw;
+    int has_avx512vl;
+    int has_avx512_common; // Enabled when AVX512(F,DQ,BW,VL) are all enabled.
     int has_avx512vnni;
     int has_sse2;
     int has_ssse3;
--- a/functable.c
+++ b/functable.c
@@ -125,7 +125,7 @@ static void init_functable(void) {
 #endif
     // X86 - AVX512 (F,DQ,BW,Vl)
 #ifdef X86_AVX512
-    if (cf.x86.has_avx512) {
+    if (cf.x86.has_avx512_common) {
         ft.adler32 = &adler32_avx512;
         ft.adler32_fold_copy = &adler32_fold_copy_avx512;
     }
@@ -138,7 +138,7 @@ static void init_functable(void) {
 #endif
     // X86 - VPCLMULQDQ
 #ifdef X86_VPCLMULQDQ_CRC
-    if (cf.x86.has_pclmulqdq && cf.x86.has_avx512 && cf.x86.has_vpclmulqdq) {
+    if (cf.x86.has_pclmulqdq && cf.x86.has_avx512_common && cf.x86.has_vpclmulqdq) {
         ft.crc32 = &crc32_vpclmulqdq;
         ft.crc32_fold = &crc32_fold_vpclmulqdq;
         ft.crc32_fold_copy = &crc32_fold_vpclmulqdq_copy;
--- a/test/benchmarks/benchmark_adler32.cc
+++ b/test/benchmarks/benchmark_adler32.cc
@@ -91,7 +91,7 @@ BENCHMARK_ADLER32(ssse3, adler32_ssse3, test_cpu_features.x86.has_ssse3);
 BENCHMARK_ADLER32(avx2, adler32_avx2, test_cpu_features.x86.has_avx2);
 #endif
 #ifdef X86_AVX512
-BENCHMARK_ADLER32(avx512, adler32_avx512, test_cpu_features.x86.has_avx512);
+BENCHMARK_ADLER32(avx512, adler32_avx512, test_cpu_features.x86.has_avx512_common);
 #endif
 #ifdef X86_AVX512VNNI
 BENCHMARK_ADLER32(avx512_vnni, adler32_avx512_vnni, test_cpu_features.x86.has_avx512vnni);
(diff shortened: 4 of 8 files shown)


KungFuJesus:
Good catch, these niche CPUs I think are a rare example where all of those extensions didn't come along for the ride. 

A more appropriate fix might be to store flags for each feature and and all of those together in the function table.

arch/x86/x86_features.c
```diff
@@ -99,7 +99,10 @@ void Z_INTERNAL x86_check_features(struct x86_cpu_features *features) {
 
         // check AVX512 bits if the OS supports saving ZMM registers
         if (features->has_os_save_zmm) {
-            features->has_avx512 = ebx & 0x00010000;
+            // We need a minimum of AVX512(F,DQ,BW) to be available because we
+            // our AVX512 functions use their relative compiler flags, causing
+            // builds to potentially use these instructions.
```
KungFuJesus: Not just potentially, definitely. We do several 8 bit sums.

tuliom:
> A more appropriate fix might be to store flags for each feature and and all of those together in the function table.

OK.
@KungFuJesus Let me confirm if I understood you: are you suggesting to replace `features->has_avx512` with the following 3 new features? `features->has_avx512f`, `features->has_avx512dq` and `features->has_avx512bw`.

KungFuJesus:
> > A more appropriate fix might be to store flags for each feature and and all of those together in the function table.
> 
> OK. @KungFuJesus Let me confirm if I understood you: are you suggesting to replace `features->has_avx512` with the following 3 new features? `features->has_avx512f`, `features->has_avx512dq` and `features->has_avx512bw`.

Correct.  You'll see where the function gets assigned in the function table.  Instead of assigning based on avx512f, check for all 3 of those features with a logical AND.  This paves the way to maybe having avx512f-only methods (though, we do rely pretty heavily on byte level operations).

phprus:
> OK. @KungFuJesus Let me confirm if I understood you: are you suggesting to replace `features->has_avx512` with the following 3 new features? `features->has_avx512f`, `features->has_avx512dq` and `features->has_avx512bw`.

I think this is not necessary.
All AVX512 optimizations require DQ & BW extensions.

KungFuJesus:
> > OK. @KungFuJesus Let me confirm if I understood you: are you suggesting to replace `features->has_avx512` with the following 3 new features? `features->has_avx512f`, `features->has_avx512dq` and `features->has_avx512bw`.
> 
> I think this is not necessary. All AVX512 optimizations require DQ & BW extensions.

Currently, yes. I'm more thinking for future scenarios.  Realistically this niche CPU is basically the only one of its kind that doesn't support the rest of the family of extensions, and there might be arguably some places that have 32 or 64 bit operands where we could write a version that supports it.  Whether or not it's worth it for a Phi or an accelerator card...well, debatable.  There's also the unfortunate fact that many of those places are more memory bandwidth than instruction bound (e.g. slide hash), and a 512 bit version of that just hurts performance due to the imposed clock penalty.

KungFuJesus:
A possible AVX512F-only version of this might involve starting from the AVX2 code and then widening (which do appear to be in the F set of instructions) after the multiply-add.  Arguably the latency on those conversions are mighty high and a large part of the throughput is the ability to pump 64 madd's through at once. It's feasible, though.  I haven't tried to scoreboard any possible solution, and I sort of expect no matter what things are going to be bound up on the shuffle port when taking an approach like that, but maybe not entirely hopeless.

phprus (approved):

tuliom:
In this new version, I implemented the new features.

KungFuJesus (approved):
LGTM. @Dead2?

phprus:
@tuliom 
We also need to check the availability of ``avx512vl``

https://github.com/zlib-ng/zlib-ng/blob/a31e6ddecf2fd4ad98a859d73c15db5a19d3ea94/arch/x86/x86_functions.h#L144

https://github.com/zlib-ng/zlib-ng/blob/a31e6ddecf2fd4ad98a859d73c15db5a19d3ea94/cmake/detect-intrinsics.cmake#L80

arch/x86/x86_features.c
```diff
@@ -100,6 +100,13 @@ void Z_INTERNAL x86_check_features(struct x86_cpu_features *features) {
         // check AVX512 bits if the OS supports saving ZMM registers
         if (features->has_os_save_zmm) {
             features->has_avx512 = ebx & 0x00010000;
+            if (features->has_avx512) {
+                // According to the Intel Software Developer's Manual,
+                // AVX512F must be enabled too in order to enable
+                // AVX512(DQ,BW).
```
phprus: We also need to check the availability of ``avx512vl``

tuliom: Good point! I've just updated the commit in order to include this feature test too.

arch/x86/x86_features.c
```diff
@@ -100,6 +100,14 @@ void Z_INTERNAL x86_check_features(struct x86_cpu_features *features) {
         // check AVX512 bits if the OS supports saving ZMM registers
         if (features->has_os_save_zmm) {
             features->has_avx512 = ebx & 0x00010000;
+            if (features->has_avx512) {
```
Dead2: **suggestion:** 
- This comment can fit in 2 lines, we have line-length 120 in this repo.
- `has_avx512` should be renamed to `has_avx512f` to avoid being confusing.

Also, to simplify the checks throughout the code, perhaps it'd be good to add a features field for `has_avx512_common` or something like that, and have that depend on F, VL, DQ and BW.

Related to that, we could also (in a separate PR) add CD since that is AFAIK always included on cpus that implement F, and then allow the compiler to utilize the CD instructions if it wants to (collision detection for vectorizing loops). It might not actually be useful to us, but it could be interesting to look into whether the compiler chooses to use it for something, and it would be a logical thing to include in `has_avx512_common`.
Btw, we test making a compare256 variant that uses the avx512cd LZCNT instruction? It might be slower than AVX2 on early model cpus though. I still don't have access to any AVX512 cpus unfortunately.

KungFuJesus: The lzcnt isn't as useful as it sounds, it will not count into the 512 bit vector, it will count per 32 or 64 bit lane.  Getting info out of that requires another reduction that has to check each individual value to know where the first nonzero bit occurred.  Additionally comparisons come back in mask registers, typically, at which point you really can't do _much_ fancy with it other than copy it back out to a GPR.  

Using march=native, it is using the masked comparisons and kortest, but I suspect it's not really going to buy us a ton. The real magic is after it goes from a register to a GPR, where the scalar tzcnt's are performed to directly count how far into the buffer based on a direct bit count.

tuliom: >    This comment can fit in 2 lines, we have line-length 120 in this repo.

Fixed.

>    has_avx512 should be renamed to has_avx512f to avoid being confusing.

Done

> Also, to simplify the checks throughout the code, perhaps it'd be good to add a features field for has_avx512_common or something like that, and have that depend on F, VL, DQ and BW.

Done.

tuliom:
This 4th version implements the changes suggested by @Dead2 .
It also changes the code for `crc32_vpclmulqdq`, which was fixed in my initial implementation, but was missing in v2 and v3 of this patch.
I updated the commit message accordingly.

Dead2 (approved):
LGTM

PR #1727, opened 2024-05-17
Improved ACLE check

Un1q32:
Isolated a part of https://github.com/zlib-ng/zlib-ng/pull/1697 that I was looking forward too, since there's been no activity there for a while.  Fixes a false positive when building for armv6 Darwin with clang and possibly other systems as well, saves me from having to explicitly pass `--without-acle` to make the build work.


cmake/detect-intrinsics.cmake
```diff
@@ -2,35 +2,34 @@
...
+    if(NOT NATIVEFLAG)
+        if(CMAKE_C_COMPILER_ID MATCHES "GNU" OR CMAKE_C_COMPILER_ID MATCHES "Clang")
+            check_c_compiler_flag("-march=armv8-a+crc" HAVE_MARCH_ARMV8_CRC)
+            if(HAVE_MARCH_ARMV8_CRC)
                 set(ACLEFLAG "-march=armv8-a+crc" CACHE INTERNAL "Compiler option to enable ACLE support")
+            else()
+                check_c_compiler_flag("-march=armv8-a+crc+simd" HAVE_MARCH_ARMV8_CRC_SIMD)
+                if(HAVE_MARCH_ARMV8_CRC_SIMD)
+                    set(ACLEFLAG "-march=armv8-a+crc+simd" CACHE INTERNAL "Compiler option to enable ACLE support")
+                endif()
```
Dead2, line 14: **question:** 
Would it make sense to instead test for `-march=armv8-a+crc+simd` before falling back to `-march=armv8-a+crc` instead of preferring to not use the +simd option?
I am not intimately familiar with what the +simd option gives us in practice, but it seems logical to instead use the one without it as the fallback here.

Un1q32: I'm not too familiar with cmake, the changes should be functionally the same as the ones to the configure script.

Dead2: True enough. I don't know why it is like that there either. 🤷‍♂️

Un1q32: Should I mark this as resolved or would you like me to switch the order of the flags?

Dead2: I am not familiar enough with this, so I was hoping someone who is would add their thoughts on the matter.

The PR is also waiting for a review by someone else, I try not to merge things until someone else in addition to me has reviewed it, and especially when I am not feeling confident in whether it is correct or not.

nmoinvaz: I'm not sure about it either. It does look like a simplification to some degree.

mtl1979: Testing `crc+simd` first would mean we can't disable NEON support... Some compilers require both extensions to be enabled to be able to target either one because NEON uses different calling convention.

Dead2: @mtl1979 Thanks, that makes sense. But do we take this into account when do the runtime tests as well?

And would it be possible for you to do a review of this PR? I'd like at least one other approval before I merge it. To me it looks ok, so I will merge it eventually even without any other reviews.

Dead2 (approved):

cmake/detect-intrinsics.cmake
mtl1979: @Dead2 Runtime check isn't really affected as the issue is just with the command-line parameters... Some compilers don't enable FPU, which means NEON can't be used. By default FPU should be enabled when available.

I usually try to review every commit that changes something I have written or modified earlier, but sometimes I'm busy with other projects and sometimes I only have small remarks about inline comments or non-code text changes that don't look quite right. If there is something seriously wrong, I'm usually loud enough that everyone notice.

configure
```diff
@@ -1091,20 +1091,33 @@ check_acle_compiler_flag() {
...
+    # Check whether compiler supports ARMv8 intrinsics
+    cat > $test.c << EOF
+#include <arm_acle.h>
+unsigned int f(unsigned int a, unsigned int b) {
+    return __crc32w(a, b);
+}
+int main(void) { return 0; }
+EOF
+    if try ${CC} ${CFLAGS} ${acleflag} $test.c; then
+        echo "Checking for ARMv8 intrinsics ... Yes." | tee -a configure.log
```
mtl1979: This is still ambiguous as all new intrinsics after ARMv7 are eventually implemented in both ARMv8 and subsequent processor families... "ARMv8 CRC intrinsics" would be less ambiguous.

cmake/detect-intrinsics.cmake
```diff
@@ -2,35 +2,34 @@
...
-            set(CMAKE_REQUIRED_FLAGS "${ACLEFLAG}")
-            check_c_source_compiles(
-                "int main() { return 0; }"
-                HAVE_ACLE_FLAG2 FAIL_REGEX "not supported")
-            set(HAVE_ACLE_FLAG ${HAVE_ACLE_FLAG2} CACHE INTERNAL "Have compiler option to enable ACLE intrinsics" FORCE)
-            unset(HAVE_ACLE_FLAG2 CACHE) # Don't cache this internal variable
-        endif()
-        set(CMAKE_REQUIRED_FLAGS)
     endif()
+    # Check whether compiler supports ARMv8 intrinsics
```
mtl1979: "ARMv8 CRC intrinsics" would be less ambiguous, see my other comment...

PR #1729, opened 2024-05-18
Codecov fixes

Dead2:
- Make codecov uploads happen even if some jobs have failed.
- Fix error in generating list of reports to upload.
- Clean up the implementation a bit.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -744,6 +744,7 @@ jobs:
     name: Upload Coverage Reports
     runs-on: ubuntu-latest
     needs: cmake
+    if: cancelled() == false
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/cmake.yml
```diff
@@ -744,6 +744,7 @@ jobs:
     name: Upload Coverage Reports
     runs-on: ubuntu-latest
     needs: cmake
+    if: ${{ always() }}
```
nmoinvaz: In our work repos we have recently changed it to:

cancelled() == false

This is essentially same as always() except it doesn't do it if a workflow has been manually cancelled.

== false is necessary because linters complain about ! at the start of a value.

Also no need for ${{}} in if statements.

Dead2: Sounds like a good idea, lets test it 👍

.github/workflows/cmake.yml
```diff
@@ -758,16 +759,14 @@ jobs:
     - name: Display all coverage artifacts
       run: |
         ls -R *.xml
-        echo "CODECOV_REPORTS=`ls -p *.xml | grep -v / | tr '\n' ',' | sed 's/,$//g`" >> $GITHUB_ENV
+        echo "CODECOV_REPORTS=`ls -p *.xml | grep -v / | tr '\n' ',' | sed 's/,$//g'`" >> $GITHUB_ENV
 
     - name: Upload reports
       uses: codecov/codecov-action@v4
-      if: (env.CODECOV_TOKEN != '' || github.repository == 'zlib-ng/zlib-ng')
```
nmoinvaz, line 765: Shouldn't this still be `env.CODECOV_TOKEN != ''` for forks? Or maybe `secrets.CODECOV_TOKEN != ''`?

Dead2: I assume you mean to disable this job for forks that don't have CODECOV_TOKEN set?
The problem with that is that it then also hinders forks from being able to have codecov with tokenless uploads as required for pull requests from 3rd parties.

So this way forks without secrets.CODECOV_TOKEN set will get a failed upload CI job (I'd say that is a minor inconvenience at worst), but those that do care about codecov and want to be able to receive pull requests from others can have that work as well.
It could maybe be improved with tests for pushes vs pull requests though, but you'd still get a failed upload for pull requests to repos with no associated codecov account.

Only having `if: (env.CODECOV_TOKEN != ''` would break on pull requests to all repos, also ours.

Not sure there is a way to please everyone..

nmoinvaz: Why remove this line at all? Seems like || github.repository covers tokenless.

Dead2: The problem with keeping it is that tokenless uploads for all downstream repos remains broken, and tokenless is required for pull requests. So codecov would remain broken for pull-requests to downstream repos.

Dead2:
Rebased.

.github/workflows/cmake.yml
Dead2: Reinstated the if as it was so that we can land this PR. That does mean that codecov remains broken for PRs to all downstream repos.

PR #1734, opened 2024-05-23
Simplify inflate window checks

Dead2:
Based on PR #1713, needs rebase when that has gone in. Didn't want to add it to that PR this late, although it really belongs there.

Since there is no need to worry about alloc calls failing after init has been done, we can
simplify the initialization and error checking of windows.

Optimization wasn't the purpose here, but in my preliminary tests, this improved inflate speed on big files by ~0.4-0.67%, so that is looking pretty good too.

PS: `inflateReset2` is called during init and whenever the windows needs a reset and its size set to 0.


Dead2:
According to codecov, this and #1713 together remove 26 branches, that is nearly 1% of the branches in the whole codebase (including tests). That doesn't necessarily mean much, it was just an interesting tidbit I noticed 😃

Dead2:
Rebased.

nmoinvaz (approved):

PR #1736, opened 2024-05-26
Improve Z_NULL compatibility with zlib

### LINKED ISSUE #1733: `Z_NULL` isn't fully compatible with zlib
In zlib, `Z_NULL` is defined as `0` where in zlib-ng it is defined as `NULL`. This can cause type errors.

phprus:
Fix #1733

### DIFF
--- a/zlib.h.in
+++ b/zlib.h.in
@@ -220,7 +220,7 @@ typedef gz_header *gz_headerp;
 #define Z_DEFLATED   8
 /* The deflate compression method (the only one supported in this version) */
 
-#define Z_NULL  NULL  /* for compatibility with zlib, was for initializing zalloc, zfree, opaque */
+#define Z_NULL  0  /* for compatibility with zlib, was for initializing zalloc, zfree, opaque */
 
 #define zlib_version zlibVersion()
 /* for compatibility with versions < 1.0.2 */


nmoinvaz (approved):

mtl1979 (approved):
LGTM... NULL was originally defined only for C++, but later got added to C too... Later C++ compilers use `nullptr` instead, so there is no "compatibility" reason to keep this defined as `NULL`.

ac000:
Of course this is now triggered by `-Wzero-as-null-pointer-constant` (enabled for C in GCC 15, not enabled by either of Wall or Wextra... [_yet_](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117059)).

```
src/http/modules/ngx_http_gunzip_filter_module.c:307:26: error: zero as null pointer constant [-Werror=zero-as-null-pointer-constant]
  307 |     ctx->zstream.next_in = Z_NULL;
      |                          ^
```

phprus:
zlib also declare ``Z_NULL`` macro as ``0``:

https://github.com/madler/zlib/blob/5a82f71ed1dfc0bec044d9702463dbdf84ea3b71/zlib.h#L216

This macro is only for compatibility with zlib.

Use ``NULL`` instead of ``Z_NULL``.

ac000:
I'm thinking of the case where you have a code base that was developed against zlib but may now build with either zlib or zlib-ng-comat as Linux distributions are now switching to...

No biggie... just an unfortunate legacy decision...

phprus:
This change was made because defining the ``Z_NULL`` macro as ``NULL`` broke other legacy code bases :(

There is no reason to use the ``Z_NULL`` macro in the code now. This is zlib-compatibility macro.

PR #1744, opened 2024-06-08
Use setup-python action to setup python environment.

nmoinvaz:


phprus:
The ``actions/python-versions`` does not support architecture ``s390x``.

```
Error: The version '3.x' with architecture 's390x' was not found for this operating system.
The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

mtl1979:
Has someone tried to ask maintainer of `python-versions` to add support for `s309x`? Sometimes actions have newer version for packages than the native package manager, which means some (security) issues have been fixed. Obviously there is (small) chance that newer version introduce new issues, but that's just digging a hole under oneself...

nmoinvaz:
It looks like they have no interest in supporting it. 
https://github.com/actions/setup-python/issues/197#issuecomment-845961956

Issue #1748, opened 2024-06-13
Failed to configure when cross compiling shared library with Android SDK


t3sserakt:
version 1.2.11

export NDK=/home/user/Android/Sdk/ndk/26.1.10909125
export TOOLCHAIN=$NDK/toolchains/llvm/prebuilt/linux-x86_64
export TARGET=$1
export API=$2
export AR=$TOOLCHAIN/bin/llvm-ar
export CC=$TOOLCHAIN/bin/$TARGET$API-clang
export AS=$CC
export CXX=$TOOLCHAIN/bin/$TARGET$API-clang++
export LD=$TOOLCHAIN/bin/ld
export RANLIB=$TOOLCHAIN/bin/llvm-ranlib
export STRIP=$TOOLCHAIN/bin/llvm-strip

PATH=$PATH:$TOOLCHAIN/bin/

./configure --shared

In the configure.log I got

No shared library support; try without defining CC and CFLAGS

I executed the test from line 421 of the configure script

/home/user/Android/Sdk/ndk/26.1.10909125/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android25-clang -shared -Wl,-soname,libz.so.1,--version-script,zlib.map -g -O0 -Wall -fPIC  -o ztest1181.so test.o
ld.lld: error: version script assignment of 'ZLIB_1.2.0' to symbol 'compressBound' failed: symbol not defined
ld.lld: error: version script assignment of 'ZLIB_1.2.0' to symbol 'deflateBound' failed: symbol not defined
ld.lld: error: version script assignment of 'ZLIB_1.2.0' to symbol 'inflateBack' failed: symbol not defined
ld.lld: error: version script assignment of 'ZLIB_1.2.0' to symbol 'inflateBackEnd' failed: symbol not defined
ld.lld: error: version script assignment of 'ZLIB_1.2.0' to symbol 'inflateBackInit_' failed: symbol not defined
ld.lld: error: version script assignment of 'ZLIB_1.2.0' to symbol 'inflateCopy' failed: symbol not defined
ld.lld: error: version script assignment of 'local' to symbol 'deflate_copyright' failed: symbol not defined
ld.lld: error: version script assignment of 'local' to symbol 'inflate_copyright' failed: symbol not defined
ld.lld: error: version script assignment of 'local' to symbol 'inflate_fast' failed: symbol not defined
[...]

PR #1749, opened 2024-06-13
[configure] Try stripping "--version-script" from LDSHAREDFLAGS if testing shared library support fails

mtl1979:
See #1748.

### DIFF
--- a/configure
+++ b/configure
@@ -600,9 +600,13 @@ int hello() {return getchar();}
 EOF
 if test $shared -eq 1; then
   printf "Checking for shared library support... " | tee -a configure.log
+  # Workaround for lld not liking --version-script
+  LDDSHAREDFLAGS="$(echo $LDSHAREDFLAGS | sed 's/,--version-script,[^ ]\+//')"
   # we must test in two steps (cc then ld), required at least on SunOS 4.x
   if try $CC -c $SFLAGS $test.c &&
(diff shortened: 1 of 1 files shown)


haampie:
I didn't see this and submitted #1750, which is more reliable.

Better would be to test if `-Wl,--version-script` is supported at all.

configure
```diff
@@ -600,10 +600,14 @@ int hello() {return getchar();}
 EOF
 if test $shared -eq 1; then
   printf "Checking for shared library support... " | tee -a configure.log
+  # Workaround for lld not liking --version-script
+  LDDSHAREDFLAGS="$(echo $LDSHAREDFLAGS | sed 's/,--version-script,[^ ]\+//')"
```
haampie, line 604: I would not use sed to remove linker flags. It fails on: `-Wl,--version-script,...`, leaving `-Wl`, causing the compiler to complain about an unrecognized flag.

haampie (changes requested):

configure
mtl1979: It's supposed to fail if `--version-script` is specified alone... In zlib-ng's configure script it is never specified alone.

mtl1979:
> I didn't see this and submitted #1750, which is more reliable.
> 
> Better would be to test if `-Wl,--version-script` is supported at all.

True... But that would require writing "dummy" version script specific to the test, otherwise it would fail just like with the current version script.

PR #1750, opened 2024-06-22
don't use zlib-ng's -Wl,--version-script in tests

haampie:
lld 18 errors when a version script assigns a version to a symbol that
is not defined in the object files. Therefore configure scripts should
not use the -Wl,--version-script,zlib.map script -- all tests that create a
dynamic library will fail


configure
```diff
@@ -630,6 +630,29 @@ fi
 
 echo >> configure.log
 
+# check for version script support
+cat > $test.c <<EOF
+int foo() { return 0; }
```
haampie:
```suggestion
int foo(void) { return 0; }
```

Dead2 (approved):
LGTM

haampie:
@Dead2 do you do backports for 2.1.x?

Dead2:
@haampie Yes, I was thinking of backporting this to the `2.1.x` branch.

PR #1756, opened 2024-07-04
Fix several issues with configure

arter97:
This PR fixes some breakages when building zlib-ng against Ubuntu with multiple architectures.

### DIFF
--- a/configure
+++ b/configure
@@ -1644,11 +1644,11 @@ case "${ARCH}" in
 int main() { return 0; }
 EOF
         if try $CC -c $SFLAGS $test.c -mfloat-abi=softfp &&
-           try $LDSHARED $LDSHAREDFLAGS $LDFLAGS -o $test$shared_ext $test.o $LDSHAREDLIBC; then
+           try $LDSHARED $LDSHAREDFLAGS $LDFLAGS -mfloat-abi=softfp -o $test$shared_ext $test.o $LDSHAREDLIBC; then
             floatabi="-mfloat-abi=softfp"
         else
             if try $CC -c $SFLAGS $test.c -mfloat-abi=hard &&
-               try $LDSHARED $LDSHAREDFLAGS $LDFLAGS -o $test$shared_ext $test.o $LDSHAREDLIBC; then
+               try $LDSHARED $LDSHAREDFLAGS $LDFLAGS -mfloat-abi=hard -o $test$shared_ext $test.o $LDSHAREDLIBC; then
                 floatabi="-mfloat-abi=hard"
             fi
         fi
--- a/configure
+++ b/configure
@@ -347,7 +347,7 @@ if test "$gcc" -eq 1 && ($cc $CFLAGS -c $test.c) >> configure.log 2>&1; then
     CFLAGS="${CFLAGS} ${ARCHS}"
     LDFLAGS="${LDFLAGS} ${ARCHS}"
   fi
-  CFLAGS="${CFLAGS} -Wall"
+  CFLAGS="${CFLAGS} -Wall -Werror=implicit-function-declaration"
   SFLAGS="${CFLAGS} -fPIC"
   if test "$warn" -eq 1; then
     CFLAGS="${CFLAGS} -Wextra"


Dead2:
The fixes look ok to me, but please note that `-flto` is not supported at all by `configure` and therefore it forces `-fno-lto`.
In CMake it is supported, but ONLY if `WITH_NATIVE_INSTRUCTIONS` is enabled.

It is not possible to compile zlib-ng with runtime instruction-set detection and LTO, because a side-effect of LTO is that it makes all the instruction-set optimized functions be compiled with the same instruction-set, thus causing `SIGILL` unless the runtime cpu supports all the instruction sets zlib-ng was compiled with.

configure
```diff
@@ -347,7 +347,7 @@ if test "$gcc" -eq 1 && ($cc $CFLAGS -c $test.c) >> configure.log 2>&1; then
     CFLAGS="${CFLAGS} ${ARCHS}"
     LDFLAGS="${LDFLAGS} ${ARCHS}"
   fi
-  CFLAGS="${CFLAGS} -Wall"
+  CFLAGS="${CFLAGS} -Wall -Werror=implicit-function-declaration"
```
nmoinvaz, line 350: Why is this necessary?

arter97: It basically forbids compilation if a function prototype was not found before. It's a very common flag to have, Linux kernel does it too.

Also, see commit 8943aa3b6f012da9386a8347a7f6dd24debb9e82.

mtl1979: In theory this is a good idea as only in C++ prototypes are actually required... In C, any undefined function is expected to return `int`, which might for example mask truncation if the actual return type is larger.

haampie: Not all compilers may support this flag, so you gotta test for it.

arter97: This really should be in all sane C compilers by now. It's first introduced in GCC in 1997: https://github.com/gcc-mirror/gcc/blob/releases/gcc-2.8.0/gcc/ChangeLog#L1868

The Linux kernel also unconditionally enforces it on all compilers.

But if the maintainers insist, I'm find guarding this with $gcc.

haampie: > The Linux kernel also unconditionally enforces it on all compilers.

There aren't many compilers in that list ;)

See for example https://github.com/spack/spack/issues/23830. Test, don't assume... It's easy enough.

Dead2: It would have to have a test to make sure it actually works.
You could also split that into a separate PR, that way we could merge the softfp abi fix.

mtl1979: > It would have to have a test to make sure it actually works. You could also split that into a separate PR, that way we could merge the softfp abi fix.

If we are not going to support LTO with configure, the softfp ABI fix will be redundant and unnecessary.

Dead2: Ok, then this PR can be focused on just adding the function declaration error.

nekopsykose:
> It is not possible to compile zlib-ng with runtime instruction-set detection and LTO, because a side-effect of LTO is that it makes all the instruction-set optimized functions be compiled with the same instruction-set, thus causing SIGILL unless the runtime cpu supports all the instruction sets zlib-ng was compiled with.

are you sure? that works completely fine, at least with latest gcc and clang. there's caveats and it's not guaranteed to work if the multiple implementations are in the same TU, but when they're in separate files lto doesn't prevent that from working and i've seen a lot of projects work fine with that, doing the same thing zlib-ng is doing.

i remember seeing some amount of bugfixes in e.g. clang around exactly that, where compiler inlining with lto could mistakenly inline a specific instruction-set-optimised impl into a place where it was supposed to be generic, which gave exactly that issue (runtime sigill instead of working detection), but with current toolchains this works fine- i can boot an avx-disabled qemu vm with a zlib-ng built with lto+instruction detection and it uses just sse4.2, with no sigills.

Dead2:
@nekopsykose I am not sure, and I have been unable to find any GCC documentation stating whether it does or does not support this.

When I previously looked this up (some years ago), I found statements by some GCC developers that said it did not keep track of different command-line parameters (like `-msse4` or `-march/-mtune`) per .c file and instead just compiled everything with the same flags during linking.
AFAIK it did however support this on a per-function level if the parameters were instead set as function attributes.

It might very well be that they have since introduced tracking of per-file parameters and apply the right ones to the code coming from those files, but I have unfortunately been unable to find any documentation of this at all.

If GCC now does support this, we'd have to know from what version this is safe.

Clang I have not looked at, but I would not be surprised if they supported this as they seem to have been focusing more on LTO than GCC did.

nekopsykose:
right, that makes sense :) i sadly lack the search-fu currently to figure out if that got changed..

the benefit of lto for something like zlib-ng is probably very little give how hand-optimised it is anyway, so it was more of a passing question, thanks for answering!

Dead2:
Closing this as unneeded, since LTO is not supported by configure

PR #1761, opened 2024-07-24
Add option to turn off ZLIB_DEBUG

### LINKED ISSUE #1760: Windows error C4242: 'function': Conversion from 'unsigned int' to 'Pos', possible loss of data
In the https://github.com/dotnet/runtime repo, we are getting a warning in Windows that was fixed back in 2019 by @nmoinvaz: https://github.com/zlib-ng/zlib-ng/issues/288

The errors:

```
C:\Users\calope\source\repos\runtime\src\native\external\zlib-ng\deflate_fast.c(61): error C4242: 'function': conversion from 'unsigned int' to 'Pos', possible loss of data
C:\Users\calope\source\repos\runtime\src\native\external\zlib-ng\deflate_quick.c(105): error C4242: 'function': conversion from 'unsigned int' to 'Pos', possible loss of data
C:\Users\calope\source\repos\runtime\src\native\external\zlib-ng\deflate_rle.c(61): error C4242: 'function': conversion from 'unsigned int' to 'Pos', possible loss of data
```

The offending code lines are:
- https://github.com/zlib-ng/zlib-ng/blob/d54e3769be0c522015b784eca2af258b1c026107/deflate_fast.c#L61
- https://github.com/zlib-ng/zlib-ng/blob/d54e3769be0c522015b784eca2af258b1c026107/deflate_quick.c#L105
- https://github.com/zlib-ng/zlib-ng/blob/d54e3769be0c522015b784eca2af258b1c026107/deflate_rle.c#L61

Which call `check_match` defined here, but has not been modified in the last 3 years: https://github.com/zlib-ng/zlib-ng/blame/develop/deflate_p.h#L14-L48

We're currently [consuming](https://github.com/dotnet/runtime/blob/main/src/native/external/zlib-ng-version.txt#L1) release https://github.com/zlib-ng/zlib-ng/releases/tag/2.1.6 (from January 10th)

This is how we build zlib-ng: https://github.com/dotnet/runtime/blob/main/src/native/external/zlib-ng.cmake

Note that we're suppressing wd4242 and wd4244 with a compile option but we _need_ to remove them as we should not be hiding memory warnings: https://github.com/dotnet/runtime/blob/efb09a57ddebfe213f43eea4289608031fe7795a/src/native/external/zlib-ng.cmake#L18-L19

I see there are a few newer releases that came after 2.1.6, and had several bug fixes, but I was unable to find a PR that looked like it fixed the issue. Is this a known problem? Any ideas how can I fix it?

carlossanlop:
Fixes https://github.com/zlib-ng/zlib-ng/issues/1760

Twin PR of https://github.com/dotnet/runtime/pull/105433

WD4242 is a critical compiler warnings that should not be suppressed, but we had to, temporarily in order to be able to compile. It would only show up in Windows when building in Debug configuration and the `check_match` method is invoked.

`check_match` gets a proper method definition when the `ZLIB_DEBUG` property is turned on and resolves to an empty macro that does nothing when `ZLIB_DEBUG` is turned off.

`ZLIB_DEBUG` is always turned on by zlib-ng when the Debug configuration is detected. Unfortunately, the property is not configurable by the user from with the existing zlib-ng cmake options, so I am proposing the addition of an option with the same name that will be turned ON by default, and when the Debug Configuration is used, the property will be enabled. When the user turns the option OFF, then the property does not get defined, and `check_match` gets an empty definition that does not throw those two warnings.

Tested locally with ZLIB_DEBUG turned ON and OFF in zlib-ng.cmake, to confirm it behaves as expected.

I also confirmed that the Linux build behavior did not get affected, with ON or OFF.


carlossanlop:
Closing. Will submit the upstream fix after I get the runtime feedback addressed.

PR #1762, opened 2024-07-25
Fix various loss of data warnings

### LINKED ISSUE #1760: Windows error C4242: 'function': Conversion from 'unsigned int' to 'Pos', possible loss of data

carlossanlop:
Fixes https://github.com/zlib-ng/zlib-ng/issues/1760

Fixed in the dotnet/runtime repo copy of zlib-ng: https://github.com/dotnet/runtime/pull/105433

WD4242 and WD4244 are compiler warnings that should not be suppressed because the warn about possible loss of data.

- WD4242 shows up in `/arch/*/slide_hash*.c` files and comes from the arguments passed to the `slide_hash_chain` method.
- WD4244 happens in Windows when building in Debug configuration, in various `deflate*.c` files, and comes from the arguments passed to the `check_match` method.
- Also fixed a couple additional failures that showed up after upgrading to v2.2.1.

Fixed by:
- Adding asserts to verify the values are below the maximum allowed for their type.
- Casting them the proper type before passing them as arguments to their methods.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -204,7 +204,7 @@ Z_INTERNAL deflate_allocs* alloc_deflate(PREFIX3(stream) *strm, int windowBits, 
 
     /* Define sizes */
     int window_size = DEFLATE_ADJUST_WINDOW_SIZE((1 << windowBits) * 2);
-    int prev_size = (1 << windowBits) * sizeof(Pos);
+    int prev_size = (1 << windowBits) * (int)sizeof(Pos);
     int head_size = HASH_SIZE * sizeof(Pos);
     int pending_size = lit_bufsize * LIT_BUFS;
     int state_size = sizeof(deflate_state);
--- a/deflate_p.h
+++ b/deflate_p.h
@@ -78,8 +78,10 @@ static inline int zng_tr_tally_dist(deflate_state* s, uint32_t dist, uint32_t le
     /* dist: distance of matched string */
     /* len: match length-STD_MIN_MATCH */
 #ifdef LIT_MEM
-    s->d_buf[s->sym_next] = dist;
-    s->l_buf[s->sym_next++] = len;
+    Assert(dist <= UINT16_MAX, "dist should fit in uint16_t");
+    Assert(len <= UINT8_MAX, "len should fit in uint8_t");
+    s->d_buf[s->sym_next] = (uint16_t)dist;
+    s->l_buf[s->sym_next++] = (uint8_t)len;
 #else
     s->sym_buf[s->sym_next++] = (uint8_t)(dist);
     s->sym_buf[s->sym_next++] = (uint8_t)(dist >> 8);
--- a/deflate_rle.c
+++ b/deflate_rle.c
@@ -58,7 +58,8 @@ Z_INTERNAL block_state deflate_rle(deflate_state *s, int flush) {
 
         /* Emit match if have run of STD_MIN_MATCH or longer, else emit literal */
         if (match_len >= STD_MIN_MATCH) {
-            check_match(s, s->strstart, s->strstart - 1, match_len);
+            Assert(s->strstart <= UINT16_MAX, "strstart should fit in uint16_t");
+            check_match(s, (Pos)s->strstart, (Pos)(s->strstart - 1), match_len);
 
             bflush = zng_tr_tally_dist(s, 1, match_len - STD_MIN_MATCH);
 
--- a/arch/arm/slide_hash_armv6.c
+++ b/arch/arm/slide_hash_armv6.c
@@ -39,7 +39,8 @@ static inline void slide_hash_chain(Pos *table, uint32_t entries, uint16_t wsize
 }
 
 Z_INTERNAL void slide_hash_armv6(deflate_state *s) {
-    unsigned int wsize = s->w_size;
+    Assert(s->w_size <= UINT16_MAX, "w_size should fit in uint16_t");
+    uint16_t wsize = (uint16_t)s->w_size;
 
     slide_hash_chain(s->head, HASH_SIZE, wsize);
     slide_hash_chain(s->prev, wsize, wsize);
--- a/arch/arm/slide_hash_neon.c
+++ b/arch/arm/slide_hash_neon.c
@@ -38,7 +38,8 @@ static inline void slide_hash_chain(Pos *table, uint32_t entries, uint16_t wsize
 }
 
 Z_INTERNAL void slide_hash_neon(deflate_state *s) {
-    unsigned int wsize = s->w_size;
+    Assert(s->w_size <= UINT16_MAX, "w_size should fit in uint16_t");
+    uint16_t wsize = (uint16_t)s->w_size;
 
     slide_hash_chain(s->head, HASH_SIZE, wsize);
     slide_hash_chain(s->prev, wsize, wsize);
--- a/arch/generic/crc32_braid_c.c
+++ b/arch/generic/crc32_braid_c.c
@@ -192,6 +192,7 @@ Z_INTERNAL uint32_t PREFIX(crc32_braid)(uint32_t crc, const uint8_t *buf, size_t
 #endif
 #endif
         words += N;
+        Assert(comb <= UINT32_MAX, "comb should fit in uint32_t");
         c = (uint32_t)ZSWAPWORD(comb);
 
         /* Update the pointer to the remaining bytes to process. */
--- a/arch/power/slide_ppc_tpl.h
+++ b/arch/power/slide_ppc_tpl.h
@@ -24,7 +24,8 @@ static inline void slide_hash_chain(Pos *table, uint32_t entries, uint16_t wsize
 }
 
 void Z_INTERNAL SLIDE_PPC(deflate_state *s) {
-    uint16_t wsize = s->w_size;
+    Assert(s->w_size <= UINT16_MAX, "w_size should fit in uint16_t");
+    uint16_t wsize = (uint16_t)s->w_size;
 
     slide_hash_chain(s->head, HASH_SIZE, wsize);
     slide_hash_chain(s->prev, wsize, wsize);
--- a/arch/riscv/slide_hash_rvv.c
+++ b/arch/riscv/slide_hash_rvv.c
@@ -23,6 +23,7 @@ static inline void slide_hash_chain(Pos *table, uint32_t entries, uint16_t wsize
 }
 
 Z_INTERNAL void slide_hash_rvv(deflate_state *s) {
+    Assert(s->w_size <= UINT16_MAX, "w_size should fit in uint16_t");
     uint16_t wsize = (uint16_t)s->w_size;
 
     slide_hash_chain(s->head, HASH_SIZE, wsize);
--- a/arch/x86/slide_hash_avx2.c
+++ b/arch/x86/slide_hash_avx2.c
@@ -31,6 +31,7 @@ static inline void slide_hash_chain(Pos *table, uint32_t entries, const __m256i 
 }
 
 Z_INTERNAL void slide_hash_avx2(deflate_state *s) {
+    Assert(s->w_size <= UINT16_MAX, "w_size should fit in uint16_t");
     uint16_t wsize = (uint16_t)s->w_size;
     const __m256i ymm_wsize = _mm256_set1_epi16((short)wsize);
 
--- a/arch/x86/slide_hash_sse2.c
+++ b/arch/x86/slide_hash_sse2.c
@@ -52,6 +52,7 @@ next_chain:
 }
 
 Z_INTERNAL void slide_hash_sse2(deflate_state *s) {
+    Assert(s->w_size <= UINT16_MAX, "w_size should fit in uint16_t");
     uint16_t wsize = (uint16_t)s->w_size;
     const __m128i xmm_wsize = _mm_set1_epi16((short)wsize);
 
--- a/deflate_fast.c
+++ b/deflate_fast.c
@@ -58,7 +58,9 @@ Z_INTERNAL block_state deflate_fast(deflate_state *s, int flush) {
         }
 
         if (match_len >= WANT_MIN_MATCH) {
-            check_match(s, s->strstart, s->match_start, match_len);
+            Assert(s->strstart <= UINT16_MAX, "strstart should fit in uint16_t");
+            Assert(s->match_start <= UINT16_MAX, "match_start should fit in uint16_t");
+            check_match(s, (Pos)s->strstart, (Pos)s->match_start, match_len);
 
             bflush = zng_tr_tally_dist(s, s->strstart - s->match_start, match_len - STD_MIN_MATCH);
 
--- a/deflate_quick.c
+++ b/deflate_quick.c
@@ -102,7 +102,8 @@ Z_INTERNAL block_state deflate_quick(deflate_state *s, int flush) {
                         if (UNLIKELY(match_len > STD_MAX_MATCH))
                             match_len = STD_MAX_MATCH;
 
-                        check_match(s, s->strstart, hash_head, match_len);
+                        Assert(s->strstart <= UINT16_MAX, "strstart should fit in uint16_t");
+                        check_match(s, (Pos)s->strstart, hash_head, match_len);
 
                         zng_tr_emit_dist(s, static_ltree, static_dtree, match_len - STD_MIN_MATCH, (uint32_t)dist);
                         s->lookahead -= match_len;
--- a/deflate_slow.c
+++ b/deflate_slow.c
@@ -78,7 +78,8 @@ Z_INTERNAL block_state deflate_slow(deflate_state *s, int flush) {
             unsigned int max_insert = s->strstart + s->lookahead - STD_MIN_MATCH;
             /* Do not insert strings in hash table beyond this. */
 
-            check_match(s, s->strstart-1, s->prev_match, s->prev_length);
+            Assert((s->strstart-1) <= UINT16_MAX, "strstart-1 should fit in uint16_t");
+            check_match(s, (Pos)(s->strstart - 1), s->prev_match, s->prev_length);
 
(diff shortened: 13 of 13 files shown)


deflate_rle.c
```diff
@@ -58,7 +58,9 @@ Z_INTERNAL block_state deflate_rle(deflate_state *s, int flush) {
 
         /* Emit match if have run of STD_MIN_MATCH or longer, else emit literal */
         if (match_len >= STD_MIN_MATCH) {
-            check_match(s, s->strstart, s->strstart - 1, match_len);
+            Assert((s->strstart - 1) <= UINT16_MAX, "strstart-1 should fit in uint16_t");
+            Assert(s->match_start <= UINT16_MAX, "match_start should fit in uint16_t");
+            check_match(s, (Pos)s->strstart, (Pos)(s->strstart - 1), match_len);
```
mtl1979: No `match_start` here, only `strstart` needs to fit in uint16_t.

deflate_rle.c
```diff
@@ -58,7 +58,9 @@ Z_INTERNAL block_state deflate_rle(deflate_state *s, int flush) {
 
         /* Emit match if have run of STD_MIN_MATCH or longer, else emit literal */
         if (match_len >= STD_MIN_MATCH) {
-            check_match(s, s->strstart, s->strstart - 1, match_len);
+            Assert((s->strstart - 1) <= UINT16_MAX, "strstart-1 should fit in uint16_t");
```
carlossanlop:
```suggestion
            Assert(s->strstart <= UINT16_MAX, "strstart should fit in uint16_t");
```

deflate_rle.c
```diff
@@ -58,7 +58,9 @@ Z_INTERNAL block_state deflate_rle(deflate_state *s, int flush) {
...
+            Assert(s->match_start <= UINT16_MAX, "match_start should fit in uint16_t");
```
carlossanlop:
```suggestion
```

deflate_rle.c
carlossanlop: Right, and one of the asserts is not needed. Fixing it now.

Dead2:
The datatype changes in some of the slide_hash functions would need to be checked for performance impacts.

As a side note, wsize cannot be bigger than 32768 without breaking with the Deflate format (Deflate64 however can be ~64k, but we don't support that). So we could in theory make s->window_size an uint_16 as well and avoid some checking, although that would have to be investigated as it could present challenges other places, and could end up costing us performance instead.

What do you think @nmoinvaz?

mtl1979:
`unsigned short` and `unsigned char` are slower than `unsigned int`. That's why we originally changed from using `Pos` in some places to using `unsigned int`. It also helped with aligning struct members. The issue with non-inlined function is that either compiler does the casting of function parameter implicitly or the developer does it explicitly.

mtl1979 (approved):
LGTM.

nmoinvaz:
Changing s->window_size to uint32_t might be a good optimization.

KungFuJesus:
> Changing s->window_size to uint32_t might be a good optimization.

At the cost of halving the throughput of slide_hash, though.

nmoinvaz:
Sorry, I meant uint16_t, I was responding to @Dead2.

nmoinvaz:
These commits should be squashed and commit message cleaned up, maybe use the same message as the name & description of this PR.

KungFuJesus:
> Sorry, I meant uint16_t, I was responding to @Dead2.

Ah.  Yeah making it a 16 bit field would at least eliminate the branch at the top of the function.  And it'd shrink the struct to boot.  So long as we aren't violating API compatibility I don't see a reason not to do that.

carlossanlop:
@nmoinvaz @Dead2 @mtl1979 - I'm updating dotnet/runtime to 2.2.1 and I found a few more loss of data bugs. Before squashing as requested by @nmoinvaz, I would like to include the fixes for the new data loss issues in this same PR.

Here is the PR updating dotnet/runtime to 2.2.1 https://github.com/dotnet/runtime/pull/105771  . The errors are:

```
C:\Users\calope\source\repos\runtime\src\native\external\zlib-ng\deflate.c(207): 
    warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?)
C:\Users\calope\source\repos\runtime\src\native\external\zlib-ng\arch\generic\crc32_braid_c.c(195): 
    error C4242: '=': conversion from 'z_word_t' to 'uint32_t', possible loss of data
C:\Users\calope\source\repos\runtime\src\native\external\zlib-ng\deflate_p.h(81): 
    error C4242: '=': conversion from 'uint32_t' to 'uint16_t', possible loss of data
C:\Users\calope\source\repos\runtime\src\native\external\zlib-ng\deflate_p.h(82): 
    error C4242: '=': conversion from 'uint32_t' to 'unsigned char', possible loss of data
```

Now I'm wondering: Is zlib-ng might be suppressing C4242 and C4334? Why aren't these errors being caught in this repo? It's outside the scope of this PR (meant for only fixing the build warnings/errors), and I don't have the expertise or time to apply and verify such changes.

carlossanlop:
> Changing s->window_size to uint32_t might be a good optimization.

> > Sorry, I meant uint16_t, I was responding to @Dead2.
> 
> Ah. Yeah making it a 16 bit field would at least eliminate the branch at the top of the function. And it'd shrink the struct to boot. So long as we aren't violating API compatibility I don't see a reason not to do that.

Can someone else please take care of these improvements after this PR gets merged?

carlossanlop:
I squashed all the commits, and I included extra fixes from similar problems found in v2.2.1 coming from crc32_braid_c.c, deflate_p.h and deflate.c.

deflate.c
```diff
@@ -204,7 +204,7 @@ Z_INTERNAL deflate_allocs* alloc_deflate(PREFIX3(stream) *strm, int windowBits,
 
     /* Define sizes */
     int window_size = DEFLATE_ADJUST_WINDOW_SIZE((1 << windowBits) * 2);
-    int prev_size = (1 << windowBits) * sizeof(Pos);
+    int prev_size = ((unsigned long long)1 << windowBits) * sizeof(Pos);
```
mtl1979: `UINT64_C(1)`... Cast is never used with constant literals. Bit width of left side must match bit width of right side when bit-shifting, otherwise VC++ will issue warning.

32-bit build might need `UINT32_C(1)` instead as there is no `SIZE_C` macro in most of the toolchains I know (only 1 has). All numeric literals are signed by default, so all unsigned operations need to explicitly use suffix macros.

mtl1979 (changes requested):

mtl1979:
> > Sorry, I meant uint16_t, I was responding to @Dead2.
> 
> Ah. Yeah making it a 16 bit field would at least eliminate the branch at the top of the function. And it'd shrink the struct to boot. So long as we aren't violating API compatibility I don't see a reason not to do that.

It will alter the alignment of the following struct members, so extra padding is required. It will also slow down access a lot as I already mentioned. In some of the supported architectures copying less than 32 bits to a register needs explicit zeroing of the unused bits.

deflate.c
nmoinvaz: This is the only way I could get the warning to go away..
```c
int prev_size = (1 << windowBits) * (int)sizeof(Pos);
```

KungFuJesus:
> It will alter the alignment of the following struct members, so extra padding is required. It will also slow down access a lot as I already mentioned. In some of the supported architectures copying less than 32 bits to a register needs explicit zeroing of the unused bits.

So move where it's declared in the struct?  I'm sure there are several 8 bit fields or other 16 bit fields to put it near that could result in the same sized struct or even smaller.  Inflating the padding up to 7 bytes doesn't really seem like nearly as much a net loss as removing a branch for this function every time is a gain.

mtl1979:
> > It will alter the alignment of the following struct members, so extra padding is required. It will also slow down access a lot as I already mentioned. In some of the supported architectures copying less than 32 bits to a register needs explicit zeroing of the unused bits.
> 
> So move where it's declared in the struct?  I'm sure there are several 8 bit fields or other 16 bit fields to put it near that could result in the same sized struct or even smaller.  Inflating the padding up to 7 bytes doesn't really seem like nearly as much a net loss as removing a branch for this function every time is a gain.
> 

The structs were originally optimized for 32-bit pointers, so things get ugly if we want the layout be clean for both 32-bit and 64-bit targets. In worst case we need only 4 bytes of padding.

Dead2:
I did a quick test with just changing `s->window_size` to be `uint16_t`, but at least on my x86-64 test this resulted in a noticeable reduction in performance and an increased code size. So that probably means it increased the need for casts elsewhere, and it could perhaps be fixed by finding those regressions and improve the related code.

@carlossanlop This is certainly not something we expect you to do. We just sometimes have ideas for further/deeper improvements and we encourage freely sharing and discussing those. It can make the PR discussions a little more cluttered, but usually it is ok.

This PR is looking pretty good so far, but after the last changes it throws an error
```
[ 17%] Building C object CMakeFiles/zlib.dir/deflate.c.o
In file included from /home/runner/work/zlib-ng/zlib-ng/deflate.c:52:
/home/runner/work/zlib-ng/zlib-ng/deflate_p.h:82:19: error: use of undeclared identifier 'UCHAR_MAX'
    Assert(len <= UCHAR_MAX, "len should fit in unsigned char");
                  ^
1 error generated.
```

KungFuJesus:
> I did a quick test with just changing `s->window_size` to be `uint16_t`, but at least on my x86-64 test this resulted in a noticeable reduction in performance and an increased code size. So that probably means it increased the need for casts elsewhere, and it could perhaps be fixed by finding those regressions and improve the related code.

That's maybe a little surprising but it might have to do with the fact that going from a 32 bit value to a 16 bit is virtually free on x86 but in the opposite direction it requires an explicit move as 16 bit writes I think leave the second half of a register alone.  I suspect you're correct that in changing every access to be 16 bit will fix that.

carlossanlop:
> @carlossanlop This is certainly not something we expect you to do. We just sometimes have ideas for further/deeper improvements and we encourage freely sharing and discussing those. It can make the PR discussions a little more cluttered, but usually it is ok.

Noted, and thank you for the clarification. Then feel free to discuss all you want, hope the final decisions bring fruitful results!

> This PR is looking pretty good so far, but after the last changes it throws an error
> 
> ```
> [ 17%] Building C object CMakeFiles/zlib.dir/deflate.c.o
> In file included from /home/runner/work/zlib-ng/zlib-ng/deflate.c:52:
> /home/runner/work/zlib-ng/zlib-ng/deflate_p.h:82:19: error: use of undeclared identifier 'UCHAR_MAX'
>     Assert(len <= UCHAR_MAX, "len should fit in unsigned char");
>                   ^
> 1 error generated.
> ```


@Dead2 Where are you seeing this error? In the dotnet/runtime repo, the PR CI passed without failures, and the build also passed in my local machine. We do have the warnings enabled in that repo.

Related question: are we suppressing the warnings in this repo somewhere? I am assuming that whatever I fix and push here, won't show up as a warning or error in the local build or in the CI, hence why I kept catching them in the dotnet/runtime repo. What cmake change should I make to make sure we get those build warnings and errors from now on?

deflate.c
mtl1979: > This is the only way I could get the warning to go away..
> 
> ```c
> int prev_size = (1 << windowBits) * (int)sizeof(Pos);
> ```

That might actually be the cleanest way as it avoids the need for truncating the whole expression to fit in `int`... `sizeof́()` returning `size_t` has been causing a lot of issues since 64-bit processors were introduced.

mtl1979:
> > I did a quick test with just changing `s->window_size` to be `uint16_t`, but at least on my x86-64 test this resulted in a noticeable reduction in performance and an increased code size. So that probably means it increased the need for casts elsewhere, and it could perhaps be fixed by finding those regressions and improve the related code.
> 
> That's maybe a little surprising but it might have to do with the fact that going from a 32 bit value to a 16 bit is virtually free on x86 but in the opposite direction it requires an explicit move as 16 bit writes I think leave the second half of a register alone. I suspect you're correct that in changing every access to be 16 bit will fix that.

As I said before, only 32-bit operations zero the upper bits automatically, explaining why 32-bit operations have smaller performance penalty than 8- and 16-bit operations... Also, some operations will widen implicitly if used with narrower than 32 bits. Shifts will widen to largest possible bit width that can handle the whole expression, not just operands of the shift itself.

deflate.c
carlossanlop: Addressed. I squashed the commits too.

carlossanlop:
Last remaining suggestion addressed. If this is ready, feel free to approve and merge.

carlossanlop:
And hopefully we can consider enabling the warnings so they do show up in this repo. It could become too time consuming (and even unsafe) to keep addressing the issues in the dotnet/runtime repo every time we upgrade (weeks or months after a new release comes out), and then have to come back here to submit the upstream fix.

mtl1979:
> And hopefully we can consider enabling the warnings so they do show up in this repo. It could become too time consuming (and even unsafe) to keep addressing the issues in the dotnet/runtime repo every time we upgrade (weeks or months after a new release comes out), and then have to come back here to submit the upstream fix.

We currently use `/W3` for Visual C++, we don't treat any warnings as errors... It will be easy to make a pull request that treats some common warnings as errors to avoid introducing regressions. Just enabling certain warnings will result in most of developers just ignoring them as they don't read the CI logs if the build passes.

deflate_p.h
```diff
@@ -78,8 +78,10 @@ static inline int zng_tr_tally_dist(deflate_state* s, uint32_t dist, uint32_t le
     /* dist: distance of matched string */
     /* len: match length-STD_MIN_MATCH */
 #ifdef LIT_MEM
-    s->d_buf[s->sym_next] = dist;
-    s->l_buf[s->sym_next++] = len;
+    Assert(dist <= UINT16_MAX, "dist should fit in uint16_t");
+    Assert(len <= UCHAR_MAX, "len should fit in unsigned char");
+    s->d_buf[s->sym_next] = (uint16_t)dist;
+    s->l_buf[s->sym_next++] = (unsigned char)len;
```
mtl1979: Should be `uint8_t` to match line 88.

carlossanlop: Fixed.

carlossanlop:
> Just enabling certain warnings will result in most of developers just ignoring them as they don't read the CI logs if the build passes.

Yep. This is one of the reasons why we now treat all warnings as errors in the dotnet/runtime repo.

mtl1979:
> > Just enabling certain warnings will result in most of developers just ignoring them as they don't read the CI logs if the build passes.
> 
> Yep. This is one of the reasons why we now treat all warnings as errors in the dotnet/runtime repo.

I've been developer for over 40 years... I'm not afraid to say aloud things that I think help make certain project look more professional. That doesn't mean I always suggest the best course of action or that I don't make mistakes... The whole point of open-source community is that there is always other people to review changes and suggest improvements. zlib-ng pretty much started as one-man project, but after few years it made the stock zlib look like something written by child in kindergarten. Last few months there has been a lot of activity in stock zlib, but that doesn't change the fact that it was unmaintained for over 5 years.

carlossanlop:
> That doesn't mean I always suggest the best course of action or that I don't make mistakes... The whole point of open-source community is that there is always other people to review changes and suggest improvements

I fully appreciate what you guys have achieved with zlib-ng. The impact of this project has been impressive to say the least. I became such a big fan of zlib-ng when I discovered it that I brought it to the attention of the .NET team so we would consider switching to it instead of stock zlib, and it makes me very happy to say that we are finally going to include it in our next release.

I'll try to help with improvements whenever possible, and for the things I am unable to help with (due to lack of time or knowledge), I will always make sure to mention it in the comments or open issues so other maintainers become aware.

Dead2 (approved):
LGTM

PR #1766, opened 2024-08-09
IBM zSystems: Hardcode HWCAP_S390_VXRS

iii-i:
Compiling zlib-ng with glibc 2.17 (minimum version still supported by crosstool-ng) fails due to the lack of HWCAP_S390_VX - it was introduced in glibc 2.23.

Strictly speaking, this is a problem with the feature detection logic in cmake. However, it's not worth disabling the s390x vectorized CRC32 if the hwcap constant is missing and the compiler intrinsics are available.

So fix by hardcoding the constant. It's a part of the kernel ABI, which does not change.

### DIFF
--- a/arch/s390/s390_features.c
+++ b/arch/s390/s390_features.c
@@ -6,7 +6,7 @@
 #endif
 
 #ifndef HWCAP_S390_VXRS
-#define HWCAP_S390_VXRS HWCAP_S390_VX
+#define HWCAP_S390_VXRS (1 << 11)
 #endif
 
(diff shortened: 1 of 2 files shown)


iii-i:
`CMake / macOS GCC UBSAN (ARM64)` fails for unrelated reasons.
`codecov/project` failure looks bogus - it points to, e.g., `arch/x86/chunkset_avx2.c`, which has nothing to do with this change.

mtl1979 (approved):
LGTM

Dead2 (approved):

Issue #1705, opened 2024-03-27
RVV support enabled accidentally and leads to SIGILL on older kernels


xctan:
PR https://github.com/zlib-ng/zlib-ng/pull/1585 has added runtime detection of RISC-V vector support for kernels newer than or equal to 6.5, and if the kernel is too old, zlib-ng would use compile time compiler support as fallback. However, this behavior is not safe for older kernels in conjunction with new compilers. In my case, I was using Linux 6.1.61 and GCC 13.2.1 when trying to compile [starship](https://github.com/starship/starship) 1.18.1 with default features enabled, including zlib-ng. Some git test cases of starship would fail because of the SIGILL signal coming from `adler32_rvv`.

I suggest using a more conservative assumption about runtime support of RVV when hwcap is not available. When the kernel is too old, or hwcap is disabled by the sandbox mechanism, RVV support should be disabled.

mtl1979:
We enable everything by default if we can't detect availability of specific instructions... This is to encourage people to upgrade to at least oldest usable kernel and C library versions... Flipping the default to not enable everything would require the change for all architectures. We still allow disabling every optimisation to make it possible to compile for older kernels.

ncopa:
I bumped into this on Alpine Linux with the Mlik-v pioneer machine, with sophgo 2042 and RVV 0.7.1.

In alpine we have previously bumped into same issue with [ffmpeg](https://gitlab.alpinelinux.org/alpine/aports/-/commit/9766a57f0f7c3b4d5704301a25792b563bc1dbe9) and [highway](https://github.com/google/highway/issues/2125).

ncopa:
We are using this workaround based on the [fix from highway](https://github.com/google/highway/pull/2127)

```diff
diff --git a/arch/riscv/riscv_features.c b/arch/riscv/riscv_features.c
index b066f42..259a63a 100644
--- a/arch/riscv/riscv_features.c
+++ b/arch/riscv/riscv_features.c
@@ -42,4 +42,20 @@ void Z_INTERNAL riscv_check_features(struct riscv_cpu_features *features) {
         riscv_check_features_runtime(features);
     else
         riscv_check_features_compile_time(features);
+    if (features->has_rvv) {
+       size_t e8m1_vec_len;
+       int64_t vtype_reg_val;
+       // Check that a vuint8m1_t vector is at least 16 bytes and that tail
+       // agnostic and mask agnostic mode are supported
+       //
+       __asm__ volatile(
[...]

Issue #1769, opened 2024-08-26
SSH connection to RISCV failed.


chLFF:
## What happened?
I try to ssh connect to my RISC-V device by python asyncssh, but some how failed.

Python3 asyncssh code:
```python3
import asyncio
import asyncssh

if __name__ == "__main__":

    async def run_coro() -> None:
        """ Test coroutine showing the Connection reset by peer bug """

        async with asyncssh.connect("10.0.90.162", port=22, username="device", password="device", known_hosts=None) as conn:
            result = await conn.run("ls /")
            print(" --- ".join(result.stdout.splitlines()))

    asyncio.run(run_coro())
```

Fail log:
```
Traceback (most recent call last):
  File "/home/user/workspace/ssh.py", line 13, in <module>
    asyncio.run(run_coro())
  File "/home/user/miniconda3/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/home/user/miniconda3/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/miniconda3/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/user/workspace/ssh.py", line 9, in run_coro
    async with asyncssh.connect("10.0.90.162", port=22, username="bianbu", password="bianbu", known_hosts=None) as conn:
  File "/home/user/miniconda3/lib/python3.11/site-packages/asyncssh/misc.py", line 298, in __aenter__
    self._coro_result = await self._coro
                        ^^^^^^^^^^^^^^^^
  File "/home/user/miniconda3/lib/python3.11/site-packages/asyncssh/connection.py", line 8811, in connect
    return await asyncio.wait_for(
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/miniconda3/lib/python3.11/asyncio/tasks.py", line 442, in wait_for
    return await fut
           ^^^^^^^^^
  File "/home/user/miniconda3/lib/python3.11/site-packages/asyncssh/connection.py", line 453, in _connect
    await options.waiter
asyncssh.misc.ConnectionLost: Connection lost
```
## What i found?
[...]

PR #1770, opened 2024-08-26
[RISCV] Better run-time detection of RVV vector instruction support

mtl1979:
Original version posted by @ncopa in #1705.

### DIFF
--- a/arch/riscv/riscv_features.c
+++ b/arch/riscv/riscv_features.c
@@ -14,7 +14,10 @@
 
 int Z_INTERNAL is_kernel_version_greater_or_equal_to_6_5() {
     struct utsname buffer;
-    uname(&buffer);
+    if (uname(&buffer) == -1) {
+        // uname failed
+        return 0;
+    }
 
     int major, minor;
     if (sscanf(buffer.release, "%d.%d", &major, &minor) != 2) {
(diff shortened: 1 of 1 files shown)


mtl1979:
@alexsifivetw Can you tag some RISC-V users who could test or review the PR...

alexsifivetw (approved):
LGTM.

arch/riscv/riscv_features.c
```diff
@@ -14,7 +14,10 @@
 
 int Z_INTERNAL is_kernel_version_greater_or_equal_to_6_5() {
     struct utsname buffer;
-    uname(&buffer);
+    if (uname(&buffer) == -1) {
+        // uname failed
+        return 0;
+    }
```
mtl1979, line 20: Only way this can fail is if the host name is too long, usually longer than 64 characters reserved for each of the fields in the struct. Even then some versions of `uname` function will just truncate the host name and leave NULL-termination out. However if `-1` is returned, the other fields are guaranteed to be uninitialised.

Not ignoring the return value makes it clear that the call to `uname` and `buffer` in stack can't be optimised out, which would cause the contents of `buffer` later below to be uninitialised. Some compilers will issue warning if the contents of struct member can be uninitialised (not guaranteed to be zero-filled), but that might require enabling "higher" warning level or static analysis.

Dead2:
@mtl1979 Remember to take this out of Draft status when you feel it is ready.

mtl1979:
> @mtl1979 Remember to take this out of Draft status when you feel it is ready.

I still have one pending issue where the fix likely overlaps the same area of code... I rather combine everything to single PR so I don't have to revise this one right after...

mtl1979:
I haven't seen any activity on #1769 for two weeks now, so I'm not delaying this PR for longer... I'm not completely sure if it even is issue in zlib-ng or the way the binary itself, or C library and kernel is compiled.

Dead2 (approved):

PR #1771, opened 2024-08-28
Fix new Windows SDK build break

am11:
Upstream port of @jkotas' https://github.com/dotnet/runtime/pull/106578.

https://github.com/dotnet/runtime/pull/106578#issuecomment-2295597357

### DIFF
--- a/functable.c
+++ b/functable.c
@@ -4,14 +4,14 @@
  */
 #ifndef DISABLE_RUNTIME_CPU_DETECTION
 
+#if defined(_MSC_VER)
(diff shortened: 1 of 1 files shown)


mtl1979:
`zbuild.h` should always be the first header to be included, however there is no rule about order of any other zlib-ng headers.

am11:
I think that would work since the problematic defines are coming from `functable.h`.

nmoinvaz (approved):

Dead2 (approved):

PR #1773, opened 2024-09-09
Don't use chunkunroll for inflateBack

### LINKED ISSUE #1772: inflate_fast_avx2 causes corrupted -- crc32 mismatch with pigz
Opening as a follow-up for https://github.com/madler/pigz/issues/123 (it contains more details)

Originally reported here: https://github.com/containers/podman/issues/23822

pigz fails with a  CRC32 mismatch error when decompressing a blob file. The same file passed the integrity check when tested with gzip.

Steps to reproduce:

Pull the container image with skopeo:
```
$ skopeo copy docker://docker.io/ollama/ollama:0.3.8-rocm oci:/var/tmp/ollama
```

Attempt to decompress the layer 1db71b1d7c67607172266cd839e3e429bf523aab5b0df4761fc0d05ec55dc727:

```
$ pigz -d < /var/tmp/ollama/blobs/sha256/1db71b1d7c67607172266cd839e3e429bf523aab5b0df4761fc0d05ec55dc727 > /dev/null
pigz: skipping: <stdin>: corrupted -- crc32 mismatch
```

The same file passes the integrity test with gzip:

```
$ gzip --test /var/tmp/ollama/blobs/sha256/1db71b1d7c67607172266cd839e3e429bf523aab5b0df4761fc0d05ec55dc727 && echo ok
ok
```

If I apply the following patch to disable using `inflate_fast_avx2`, the issue does not happen, `--without-unaligned` at configure time does not seem to make any difference:

```diff
diff --git a/functable.c b/functable.c
index dd8f7731..10abb358 100644
--- a/functable.c
+++ b/functable.c
@@ -115,7 +115,7 @@ static void init_functable(void) {
         ft.adler32_fold_copy = &adler32_fold_copy_avx2;
         ft.chunkmemset_safe = &chunkmemset_safe_avx2;
         ft.chunksize = &chunksize_avx2;
-        ft.inflate_fast = &inflate_fast_avx2;
+        //ft.inflate_fast = &inflate_fast_avx2;
         ft.slide_hash = &slide_hash_avx2;
 #  ifdef HAVE_BUILTIN_CTZ
         ft.compare256 = &compare256_avx2;
```

nmoinvaz:
If the output buffer and the window buffer are the same memory allocation, we cannot make the assumptions that `chunkunroll` does, that it is okay to overwrite the output buffer.

Fixes #1772

### DIFF
--- a/inffast_tpl.h
+++ b/inffast_tpl.h
@@ -249,7 +249,8 @@ void Z_INTERNAL INFLATE_FAST(PREFIX3(stream) *strm, uint32_t start) {
                     if (op < len) {             /* still need some from output */
                         len -= op;
                         out = chunkcopy_safe(out, from, op, safe);
-                        out = CHUNKUNROLL(out, &dist, &len);
+                        if (!extra_safe)
+                            out = CHUNKUNROLL(out, &dist, &len);
                         out = chunkcopy_safe(out, out - dist, len, safe);
                     } else {
                         out = chunkcopy_safe(out, from, len, safe);


Dead2 (approved):
LGTM

giuseppe:
thanks, tested locally and it solves the issue I've reported

PR #1774, opened 2024-09-10
Fixed false positive HAVE_ARMV6_INTRIN value on old ARM platforms

asmorkalov:
The check returns 1, even if `__uqsub16` is not provided by compiler. The issue was reproduced with jetson TK1 (armv7 neon), GCC 5.4.0 and `-DWITH_RUNTIME_CPU_DETECTION=ON`

### DIFF
--- a/cmake/detect-intrinsics.cmake
+++ b/cmake/detect-intrinsics.cmake
@@ -66,7 +66,7 @@ macro(check_armv6_compiler_flag)
             return __uqsub16(a, b);
         #endif
         }
-        int main(void) { return 0; }"
+        int main(void) { return f(1,2); }"
         HAVE_ARMV6_INTRIN
     )
     set(CMAKE_REQUIRED_FLAGS)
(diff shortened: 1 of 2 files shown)


Dead2 (approved):
LGTM

nmoinvaz:
Does this need to be fixed in the configure script too?

Dead2:
It probably should be changed for all the tests in both CMake and Configure.

Although the compiler _should_ balk at the code, it is known that is will not do so if for example LTO is enabled, because at linktime the compiler can discard the code as unused and never detect/ignore that it contained code it would not be able to/allowed to compile (given the current instruction set limitations).

We disable LTO during these tests in part for this reason, but there may well be other compilers/settings/conditions that will cause the same issue, now or in the future.

asmorkalov:
Added to the configure script too. Looks like it's not the only place in configure script where the test function is not called in main.

nmoinvaz:
> Looks like it's not the only place in configure script where the test function is not called in main.

Those changes were made in b7ac8479353f8e64c71cace5f8f76494fed6fad9 to "ensure instructions are not optimized out before assembler is run". https://github.com/zlib-ng/zlib-ng/pull/1562.

nmoinvaz:
My question is, why is your compiler compiling for that test okay if the instruction does not exist?

PR #1764, opened 2024-08-05
Silence possible loss of data warning in crc32_braid ZSWAPWORD call

nmoinvaz:
MSVC
```
Warning	C4244
'=': conversion from 'z_word_t' to 'uint32_t', possible loss of data zlib
crc32_braid_c.c	195
```

PR #1763, opened 2024-07-30
[CI] Upgrade MacOS ARM64 UBSAN to use gcc-13.

mtl1979:
* GCC 12.4 and 13.3 are minimum supported due to [GCC bug #114007](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114007); prefer GCC 13.3
* gcovr 7.x is required for parsing coverage output of GCC 13.x; gcovr 7.2 is currently latest

PR #1779, opened 2024-09-15
2.2.2 Release

Dead2:
This release fixes a corruption bug with the inflateback implementation, this was
detected by Docker using pigz w/zlib-ng to decompress a 25GB image file and failing CRC.
Since this is so far the only known way to trigger the bug, it seems to be hard to hit.

Most of the rest are minor changes to avoid triggering warnings in MSVC or optional
warnings in other compilers, as well as a few minor fixes to the buildsystem and CI.

### Important Fixes
- Don't use chunkunroll for inflateBack #1773

### Buildsystem
- Enable warning C4242 and treat warnings as errors for Visual C++. #1768
- Fixed false positive HAVE_ARMV6_INTRIN value on old ARM platforms #1774

### CI/Test
- Upgrade MacOS ARM64 UBSAN to use gcc-13. #1763

### Misc
- Fix compiler warnings #1762 #1764 #1765 
- Fix new Windows SDK build break #1771
- Prepare to make use of unaligned loads on big endian in insert_string #1695

### S390x
- IBM zSystems: Hardcode HWCAP_S390_VXRS #1766

### RISC-V
- Better run-time detection of RVV vector instruction support #1770


mtl1979:
Personally I think #1773 alone warranted a new release, but I'm happy that we managed to get some fixes for minor issues also included. I have seen quite a few open issues where the original reporter went quiet after few days, so there might still be fixes left for next release.

PR #1780, opened 2024-09-15
2.1.8 Release

Dead2:
This backport release fixes a corruption bug with the inflateback implementation, this was
detected by Docker using pigz w/zlib-ng to decompress a 25GB image file and failing CRC.
Since this is so far the only known way to trigger the bug, it seems to be hard to hit.

- Don't use chunkunroll for inflateBack #1773

Most of the rest are minor changes to avoid triggering warnings in MSVC or optional
warnings in other compilers, as well as a few minor fixes to the buildsystem and CI.

### Important Fixes
- Don't use chunkunroll for inflateBack #1773

### Buildsystem
- Enable warning C4242 and treat warnings as errors for Visual C++. #1768
- Fixed false positive HAVE_ARMV6_INTRIN value on old ARM platforms #1774
- don't use zlib-ng's -Wl,--version-script in tests #1750

### CI/Test
- Upgrade MacOS ARM64 UBSAN to use gcc-13. #1763

### Misc
- Fix compiler warnings #1762 #1764
- Fix new Windows SDK build break #1771

### S390x
- IBM zSystems: Hardcode HWCAP_S390_VXRS #1766

### RISC-V
- Better run-time detection of RVV vector instruction support #1770


mtl1979:
Only thing I currently think still might need some fixes is MacOS CI, but that pretty much requires someone who is very familiar with the internals of the toolchain.

PR #1784, opened 2024-09-17
Use target include instead of raw include

FantasqueX:
CMake recommends using `target_include_directories` instead of `include_directories`. https://cmake.org/cmake/help/latest/command/include_directories.html

Some downstream projects such as opencv integrate zlib-ng to their source code. In those cases, less `include_directories` is better.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1174,10 +1174,6 @@ if (ZLIB_COMPAT)
     endif()
 endif()
 
-include_directories(${CMAKE_CURRENT_SOURCE_DIR})
-include_directories(${ARCHDIR})
-include_directories(${CMAKE_CURRENT_SOURCE_DIR}/arch/generic)
-
 foreach(ZLIB_INSTALL_LIBRARY ${ZLIB_INSTALL_LIBRARIES})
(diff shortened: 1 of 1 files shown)


nmoinvaz:
I'm not sure about removing `${CMAKE_CURRENT_SOURCE_DIR}` but the CI should say if there is a problem.

FantasqueX:
> I'm not sure about removing `${CMAKE_CURRENT_SOURCE_DIR}` but the CI should say if there is a problem.

```
target_include_directories(${ZLIB_INSTALL_LIBRARY} PUBLIC
        "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}$<SEMICOLON>${CMAKE_CURRENT_SOURCE_DIR}>"
        "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>")
```

The code above does the same thing as `include_directories(${CMAKE_CURRENT_SOURCE_DIR})` in compile time.

nmoinvaz (approved):

Dead2 (approved):
LGTM

PR #1786, opened 2024-09-17
Disable MSVC warning 4819

FantasqueX:
Currently, zlib-ng doesn't compile on my machine due to warning C4819: The file contains a character that cannot be represented in the current code page (936). Save the file in Unicode format to prevent data loss.

The non-ascii characters live in zbuild.h.
```
 * gcc says: "warning: implicit declaration of function ‘...’"
 * g++ says: "error: new declaration ‘...’ ambiguates built-in declaration ‘...’"
```

Because zlib-ng treats warning as error, the project doesn't compile. If this warning is disabled, zlib-ng compiles well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,7 +193,7 @@ elseif(MSVC)
     # /Oi ?
     set(WARNFLAGS /W3 /w34242 /WX)
     set(WARNFLAGS_MAINTAINER /W4)
-    set(WARNFLAGS_DISABLE /wd4206 /wd4054)
+    set(WARNFLAGS_DISABLE /wd4206 /wd4054 /wd4819)
     if(BASEARCH_ARM_FOUND)
         add_definitions(-D_ARM_WINAPI_PARTITION_DESKTOP_SDK_AVAILABLE)
         if(NOT "${ARCH}" MATCHES "aarch64")


nmoinvaz:
Is it possible to remove the character that is none ascii?

FantasqueX:
> Is it possible to remove the character that is none ascii?

It's a possible solution if maintainers think removing those characters is OK.

mtl1979:
As far as I know full UTF-8 support (without requiring BOM at start of file) only exists in later versions of Windows 10, and Windows 11, so if the file was edited in Windows, it would default to ISO-8859-15 (Latin-15, or Latin-1 + Euro sign), which is the default in a lot of English-speaking countries in Europe.

I'm all for making sure the source files either use UTF-8 or 7-bit ASCII, which are safe in all regions except some Asian countries using for example EUC...

nmoinvaz:
What is the characters or lines of code?

mtl1979:
> What is the characters or lines of code?

Lines 278-279 in current develop branch...

See also: https://learn.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-170

nmoinvaz:
If it is in a comment, then the comment should be fixed to be ascii characters instead of changing warning flag.

FantasqueX:
> If it is in a comment, then the comment should be fixed to be ascii characters instead of changing warning flag.

OK, I'll create a PR to replace non-ascii characters later today.

FantasqueX:
https://github.com/zlib-ng/zlib-ng/pull/1791

PR #1790, opened 2024-09-18
[CI] Don't try to use macOS 11 as it's no longer supported.

mtl1979:
* Add support for Apple M1 in pkgcheck


Dead2 (approved):
LGTM

nmoinvaz:
Is the ZERO_AR_DATE thing no longer and issue on the newer SDK?

mtl1979:
> Is the ZERO_AR_DATE thing no longer and issue on the newer SDK?

I tried few different SDK versions and this one was first one to pass... Not sure if it's because of switch to AArch64 or generic fix in toolchain... With previous SDK version it would fail long before `ar` was even used.

nmoinvaz (approved):

PR #1789, opened 2024-09-18
Force Visual C++ to treat source files as UTF-8.

mtl1979:
See #1786.

PR #1791, opened 2024-09-19
Replace non-ascii characters to fix MSVC warning

FantasqueX:
https://github.com/zlib-ng/zlib-ng/pull/1786#issuecomment-2359182303


nmoinvaz (approved):

mtl1979:
Even though #1789 should fix the issue for Visual C++, the issue would still exist for other compilers (and even text editors) when UTF-8 locale isn't used.

Dead2 (approved):

PR #1798, opened 2024-09-26
Fix Haiku default directories.

mtl1979:
Haiku is open-source operating system forked from BeOS. It doesn't use same file layout as Linux or *BSD.

We have to move uname check earlier, so defaults can be selected depending on operating system.

### DIFF
--- a/cmake/detect-install-dirs.cmake
+++ b/cmake/detect-install-dirs.cmake
@@ -16,6 +16,9 @@ if (DEFINED LIB_INSTALL_DIR)
     set(CMAKE_INSTALL_LIBDIR "${LIB_INSTALL_DIR}")
 elseif (DEFINED INSTALL_LIB_DIR)
     set(CMAKE_INSTALL_LIBDIR "${INSTALL_LIB_DIR}")
+elseif (CMAKE_INSTALL_PREFIX STREQUAL "/boot/system")
+    # Haiku development libraries
+    set(CMAKE_INSTALL_LIBDIR "develop/lib")
 endif()
 
 # Determine installation directory for include files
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -77,12 +80,24 @@ RCOBJS=
 STRIP=
 ARCHS=
 PC_CFLAGS=
-prefix=${prefix-/usr/local}
+if test "${uname}" = "Haiku" || test "${uname}" = "haiku"; then
+  prefix=${prefix-/boot/system}
```
jmairboeck, line 84: The Haiku equivalent of `/usr/local` is `/boot/system/non-packaged`. `/boot/system` only works for packages, because it is read-only.

configure
```diff
@@ -77,12 +80,24 @@ RCOBJS=
...
+else
+  prefix=${prefix-/usr/local}
+fi
 exec_prefix=${exec_prefix-'${prefix}'}
 bindir=${bindir-'${exec_prefix}/bin'}
-libdir=${libdir-'${exec_prefix}/lib'}
+if test "${exec_prefix}" = "/boot/system" || (test "${prefix}" = "/boot/system" && test "${exec_prefix}" = '${prefix}'); then
```
jmairboeck, line 90: This check doesn't work like this. Packages use a prefix like `/packages/$packageName/.self` (which at runtime is a symlink to where the package is installed, usually `/boot/system`), or it is a `non-packaged` path (see above).

The same applies also everywhere else where you check the prefix.

configure
mtl1979: This is just an early patch, to get configure and cmake make identical binaries, after that I start working on package recipe. Supporting both non-packaged and packaged build makes things too complicated and suits better for downstream patch if needed.

configure
mtl1979: Can't just evaluate `exec_prefix` because it isn't always variable reference, hence single quotes, it can be a literal path. When packaging, `prefix` gets overwritten.

cmake/detect-install-dirs.cmake
```diff
@@ -24,6 +27,20 @@ if (DEFINED INC_INSTALL_DIR)
...
+
+# Determine installation directory for man files
+if (DEFINED MAN_INSTALL_DIR)
+    set(MAN_INSTALL_DIR "${MAN_INSTALL_DIR}" CACHE PATH "Installation directory for manuals (Deprecated)" FORCE)
+    set(CMAKE_INSTALL_MANDIR "${MAN_INSTALL_DIR}")
+elseif (DEFINED INSTALL_MAN_DIR)
+    set(CMAKE_INSTALL_MANDIR "${INSTALL_MAN_DIR}")
+elseif (CMAKE_INSTALL_PREFIX STREQUAL "/boot/system")
+    # Haiku system manuals
+    set(CMAKE_INSTALL_MANDIR "documentation/man")
```
phprus, line 43: Why do we need to add ``MAN_INSTALL_DIR`` and ``INSTALL_MAN_DIR`` if they were not used before, instead of using the standard ``CMAKE_INSTALL_MANDIR`` variable?


``MAN_INSTALL_DIR`` has been removed as unused: https://github.com/zlib-ng/zlib-ng/pull/885

mtl1979: This is just to document all the directories, so I can run pkgcheck and see it pass... There is still a lot of unused code in both configure and CMakeLists.txt for backwards compatibility.

PR #1801, opened 2024-10-03
Compute the "safe" distance properly

KungFuJesus:
The safe pointer that is computed is an exclusive, not inclusive bounds. While we were probably rarely ever bit this, if ever, it still makes sense to apply the limit, properly.

## Summary by CodeRabbit

- **Bug Fixes**
	- Adjusted memory operation parameters for improved performance in the decompression process. 

- **Documentation**
	- Maintained existing comments and documentation for clarity on function purpose and behavior.


Dead2 (approved):
LGTM

PR #1803, opened 2024-10-09
Reorder variables in inflate functions to reduce padding holes

Dead2:
Reorder variables in inflate functions to reduce padding holes due to variable alignment requirements.

This is more a case of doing things the recommended way rather than an optimization.

This is a very very minor change, and while it is probable to have a positive impact on performance, it is so small that I can not reliably benchmark it as it is well within the run-to-run variability.
That said, it seems to trend towards being faster by around 0.02%.


nmoinvaz (approved):

KungFuJesus (approved):
Even if the performance wins are only theoretical, reducing the memory footprint is always a plus, especially when it's no cost like this.

PR #1808, opened 2024-11-03
Add support for sending "assume" hints to compilers

Dead2:
- Add support for giving compilers (and static analyzers) hints about value ranges of variables, allowing them to better optimize the code.
- It is very important that Assume hints are always true, therefore make debug builds check the values and throw errors if the assumption is not true.
- Make Assert() do Assume() in release builds.
- Add hints about value ranges coming into inffast
- Remove always-true Assert in zng_tr_tally_lit()

Just letting existing Asserts() calls go to Assume() reduces compiled code size by 104 bytes on x86-64.
Performance seems to be unchanged with these hints (within 0.01%).

The intentions behind adding this:
- Improve static analyzer hinting, letting them avoid false-positives.
- Possibly letting the compilers better optimize code and for example not try to optimize for big lengths where we already know that we will only have short lengths (for example 258bytes is the max length of matches, so there is no need to optimize for the 64kB lengths that the datatype would allow).
- Make adding asserts/assume to new code more attractive.
- This would then also make code more self-documenting and make it easier for anyone to understand the limitations of a function.

### DIFF
--- a/zbuild.h
+++ b/zbuild.h
@@ -331,7 +331,7 @@
 #  define Assume(cond)          Assert(cond, "Value assumption failed")
 #else
 #  define Assert(cond, msg)
-#  define AssertHint(cind, msg) z_assume(cond)
+#  define AssertHint(cond, msg) z_assume(cond)
 #  define Assume(cond)          z_assume(cond)
 #endif
 
--- a/zbuild.h
+++ b/zbuild.h
@@ -284,19 +284,55 @@
 #ifdef ZLIB_DEBUG
    extern int Z_INTERNAL z_verbose;
    extern void Z_INTERNAL z_error(const char *m);
-#  define Assert(cond, msg) {int _cond = (cond); if (!(_cond)) z_error(msg);}
 #  define Trace(x) {if (z_verbose >= 0) fprintf x;}
 #  define Tracev(x) {if (z_verbose > 0) fprintf x;}
 #  define Tracevv(x) {if (z_verbose > 1) fprintf x;}
 #  define Tracec(c, x) {if (z_verbose > 0 && (c)) fprintf x;}
 #  define Tracecv(c, x) {if (z_verbose > 1 && (c)) fprintf x;}
 #else
-#  define Assert(cond, msg)
 #  define Trace(x)
 #  define Tracev(x)
 #  define Tracevv(x)
 #  define Tracec(c, x)
 #  define Tracecv(c, x)
+#endif
+
+/* Prepare compiler-specific assume intrinsics */
+#if defined(__clang__) && __has_builtin(__builtin_assume)
+  #define z_assume(...) __builtin_assume(__VA_ARGS__)
+#elif defined(__GNUC__) && __GNUC__ >= 13
+  #define z_assume(...) __attribute__((assume(__VA_ARGS__)))
+#elif defined(_MSC_VER) && _MSC_VER >= 1700
+  #define z_assume(...) __assume(__VA_ARGS__)
+#else
+  #define z_assume(...)
+#endif
+/*
+   Assume     - Run check in Debug builds, send hint to compiler in Release builds.
+   Assert     - Run check in Debug builds
+   AssertHint - Run check in Debug builds, send hint to compiler in Release builds.
+   Both Assert types require a second parameter msg containing an error message.
+
+   The Assume(cond) and AssertHint(cond,msg) macros provides hints to compiler and
+   static analyzers about what value range is valid for a variable, helping them
+   better understand and optimize for real use.
+   To ensure this does not hide bugs, compilation in debug mode will replace hinting
+   with assert checks of the hint, and will throw an error on failed checks.
+
+   Example:
+    void process_buffer(char* buf, size_t len) {
+      Assume(len <= 1024);  // Hint to compiler about maximum size
+      ...
+    }
+*/
+#ifdef ZLIB_DEBUG
+#  define Assert(cond, msg)     {int _cond = (cond); if (!(_cond)) z_error(msg);}
+#  define AssertHint(cond, msg) Assert(cond,msg)
+#  define Assume(cond)          Assert(cond, "Value assumption failed")
+#else
+#  define Assert(cond, msg)
+#  define AssertHint(cind, msg) z_assume(cond)
+#  define Assume(cond)          z_assume(cond)
 #endif
 
 /* OPTIMAL_CMP values determine the comparison width:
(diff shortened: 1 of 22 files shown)


zbuild.h
```diff
@@ -211,6 +211,28 @@
...
+    #endif
+  #else
+    #define __assume(...)
+  #endif
+#endif
+#ifdef ZLIB_DEBUG
+#  define Assume(...) do { if (!(cond)) z_error("Value assumption failed");; } while (0)
+#else
+#  define Assume(...) __assume(__VA_ARGS__);
+#endif
```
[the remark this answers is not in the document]
Dead2: Fixed.

zbuild.h
```diff
@@ -211,6 +211,26 @@
...
+  #define z_assume(...) do { __builtin_assume(__VA_ARGS__); } while(0)
+#elif defined(__GNUC__) && __GNUC__ >= 13
+  #define z_assume(...) __attribute__((__assume__(__VA_ARGS__)))
+#elif defined(_MSC_VER) && _MSC_VER >= 1900
+  #define z_assume(...) do { __assume(__VA_ARGS__); } while(0)
+#else
+  #define z_assume(...)
+#endif
+#ifdef ZLIB_DEBUG
+#  define Assume(...) do { if (!(__VA_ARGS__)) z_error("Value assumption failed");; } while (0)
```
phprus: ``;;`` instead of ``;``

Dead2: Fixed.

Dead2:
MSVC had support for this before the stated version, but I found some references to bugs, so ended up setting VS 2012 as lower end.
Clang had support from 4.5 or thereabouts, but I set version 5 as lowest for the same reason.

It is possible to emulate this using unreachable attributes in older GCC versions, however the semantics for that differ a little, therefore I chose to only use the proper assume hint at this time.

Dead2:
I'll have to redo this PR, as I had initially suspected, not all the asserts are in fact usable as assume hints (some variables do not exist in release builds, so those cause failures).

I need to consider whether to add all Assume hints separately, or whether to introduce another Assert() that does Assume in release builds, while the normal one does not (name suggestions?). Thoughts?

We could possibly do it like this if it looks clean, not sure.
```
Assert(buf != NULL, "lost buf"); Assume(buf != NULL);
```

mtl1979:
> I'll have to redo this PR, as I had initially suspected, not all the asserts are in fact usable as assume hints (some variables do not exist in release builds, so those cause failures).
> 
> I need to consider whether to add all Assume hints separately, or whether to introduce another Assert() that does Assume in release builds, while the normal one does not (name suggestions?). Thoughts?

Something like `Verify()` for release builds would work... It makes clear that we don't assume the condition holds, we want to make sure it holds.

zbuild.h
```diff
@@ -211,6 +211,32 @@
...
+   check the assumption and throw an error if it fails.
+
+   Example:
+    void process_buffer(char* buf, size_t len) {
+      Assume(len <= 1024);  // Hint to compiler about maximum size
+      ...
+    }
+ */
+#if defined(__clang__) && __clang_major__ >= 4
+  #define z_assume(...) do { __builtin_assume(__VA_ARGS__); } while(0)
```
nmoinvaz: `#define -> #  define`

Dead2:
> > I need to consider whether to add all Assume hints separately, or whether to introduce another Assert() that does Assume in release builds, while the normal one does not (name suggestions?). Thoughts?
> 
> Something like `Verify()` for release builds would work... It makes clear that we don't assume the condition holds, we want to make sure it holds.

Not sure what you mean `Verify` would do. We don't want to add actual checks to the release builds.
That is why the hint is named assume, meaning it is completely safe for the compiler to assume that the hint is always true. Using an assert would give compilers and static checkers the same hint but also perform the check and thus cost performance and add instructions to the compiled code. Assume gives the hint but does not add any verification code to the build, if you assumed wrong it would more than likely break due to incorrect optimizations.

mtl1979:
> > > I need to consider whether to add all Assume hints separately, or whether to introduce another Assert() that does Assume in release builds, while the normal one does not (name suggestions?). Thoughts?
> 
> > 
> 
> > Something like `Verify()` for release builds would work... It makes clear that we don't assume the condition holds, we want to make sure it holds.
> 
> 
> 
> Not sure what you mean `Verify` would do. We don't want to add actual checks to the release builds.
> 
Just suggesting names... Maybe use Verify() for Debug builds and Assume() for Release builds.

nmoinvaz:
`_assume` might also be valuable in adler32 short length functions to tell the compiler that length is expected to be less than < # of bytes for unrolling.

I don't know if it would really make a difference but we could do `_assume(len <= 256)` in the compare256 functions too. Or maybe it works in reverse.. in that instance.

Dead2:
> `_assume` might also be valuable in adler32 short length functions to tell the compiler that length is expected to be less than < # of bytes for unrolling.
> 
> I don't know if it would really make a difference but we could do `_assume(len <= 256)` in the compare256 functions too. Or maybe it works in reverse.. in that instance.

Assume is only useful for letting the compiler know something it cannot already figure out from the code. Since most of the call sites do a len check before calling the small adler32/crc32 functions, and those functions are inlined, the compiler can already figure most of that out.
It does reduce code size a bit on the adler32/crc32 functions (because it can put MAX_LEN in relation to len), but the benchmarks are inconclusive on whether it is any faster or not.

Ex:
```
adler32/c/1                                  2.91 ns         2.91 ns    722457796
adler32/c/8                                  4.48 ns         4.47 ns    469600443
adler32/c/12                                 5.60 ns         5.59 ns    375684607
adler32/c/16                                 8.00 ns         7.99 ns    262691702
adler32/c/32                                 12.2 ns         12.2 ns    171684390
adler32/c/64                                 20.7 ns         20.7 ns    101360773
```
assume:
```
adler32/c/1                                  2.91 ns         2.91 ns    722440306
adler32/c/8                                  4.48 ns         4.47 ns    469603481
adler32/c/12                                 5.60 ns         5.59 ns    375681324
adler32/c/16                                 7.99 ns         7.98 ns    263506694
adler32/c/32                                 12.2 ns         12.2 ns    172136405
adler32/c/64                                 20.7 ns         20.6 ns    102222248
```

It probably is a tiny bit faster, but too close to really tell.

Adding the assume hint to both crc32 and adler32 small functions reduced code size by 64bytes, so it did help somewhat.
I am considering pulling the first two commits from this PR into a new PR, so we can add Assume hints as we go, where we are certain it is safe.

PR #1810, opened 2024-11-21
Fix cmake 3.10 warning with latest version

am11:
We are seeing warnings like these with cmake 3.31 in github actions:

```
CMake Deprecation Warning at D:/a/runtime/runtime/runtime/src/native/external/zlib-ng/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.
```

e.g. https://github.com/dotnet/runtime/actions/runs/11948322805/job/33305725079?pr=109537

## Summary by CodeRabbit

- **New Features**
	- Enhanced support for various architectures including ARM, PPC, RISC-V, S390, and X86.
	- Improved error handling for unsupported configurations.
	- More detailed status messages during the configuration process.

- **Bug Fixes**
	- Updated validation checks for C standard options to ensure compatibility.

- **Documentation**
	- Revised installation section for clarity on file placements.

- **Chores**
	- Updated minimum required CMake version across multiple projects to 3.12 for improved compatibility and features.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,9 +1,11 @@
-cmake_minimum_required(VERSION 3.5.1)
 if(CMAKE_VERSION VERSION_LESS 3.12)
+    cmake_minimum_required(VERSION 3.5.1)
     cmake_policy(VERSION ${CMAKE_VERSION})
 else()
-    cmake_policy(VERSION 3.5.1...3.29.0)
+    cmake_minimum_required(VERSION 3.12.0)
+    cmake_policy(VERSION 3.12.0...3.31.0)
 endif()
+
 message(STATUS "Using CMake version ${CMAKE_VERSION}")
 
 if(POLICY CMP0169)
--- a/test/add-subdirectory-project/CMakeLists.txt
+++ b/test/add-subdirectory-project/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5.1)
+cmake_minimum_required(VERSION 3.12)
 
 project(zlib-ng-add-subdirecory-test C)
 
--- a/test/fuzz/CMakeLists.txt
+++ b/test/fuzz/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5.1)
+cmake_minimum_required(VERSION 3.12)
 
 if(CMAKE_C_COMPILER_ID MATCHES "Clang")
     enable_language(CXX)
--- a/test/pigz/CMakeLists.txt
+++ b/test/pigz/CMakeLists.txt
@@ -18,7 +18,7 @@
 #   PTHREADS4W_ROOT     - Path to pthreads4w source directory on Windows.
 #                         If not specified then threading will be disabled.
 
-cmake_minimum_required(VERSION 3.11)
+cmake_minimum_required(VERSION 3.12)
 
 include(CheckCCompilerFlag)
 include(FeatureSummary)


mtl1979:
There is a lot of systems that don't have cmake 3.12 but can compile zlib-ng without tests and successfully run it. This includes at least systems running Ubuntu 18.04, which is last release to support certain enterprise software. There has been a lot of discussion in the past about bugs in later versions of Ubuntu and essential packages bundled with various Linux distributions.

am11:
Good point. I have lowered it to 3.10 for ubuntu 18.04 support (cmake/bionic-updates 3.10.2-1ubuntu2.18.04.2 amd64). 

BTW, standard support of ubuntu 18.04 ended over a year ago (4.23). Funnily even debian buster (which is also out of standard support) has 3.13 available in olstable.

Also, for someone running old systems can always resort to kitware's binary distribution., direct download or script download. e.g. the latest 3.31 works on systems as old as CentOS 7:
  ```sh
  $ curl -o cmake-install.sh -sSL https://github.com/Kitware/CMake/releases/download/v3.31.0/cmake-3.31.0-Linux-x86_64.sh
  $ sh cmake-install.sh --skip-license --prefix=/usr/local
  $ cmake --version # CMake Installer Version: 3.31.0, Copyright (c) Kitware
  ```

am11:
> There is a lot of systems that don't have cmake 3.12 but can compile zlib-ng without tests and successfully run it. This includes at least systems running Ubuntu 18.04, which is last release to support certain enterprise software. There has been a lot of discussion in the past about bugs in later versions of Ubuntu and essential packages bundled with various Linux distributions.

Good point. I have lowered it to 3.10 for ubuntu 18.04 support (cmake/bionic-updates 3.10.2-1ubuntu2.18.04.2 amd64). 

BTW, standard support of ubuntu 18.04 ended over a year ago (4.23). Funnily, even debian buster (which is also out of standard support) has 3.13 available in oldstable.

Also, for someone running old systems can always resort to kitware's binary distribution., direct download or script download. e.g. the latest 3.31 works on systems as old as CentOS 7 (glibc 2.17):
  ```sh
  $ curl -o cmake-install.sh -sSL https://github.com/Kitware/CMake/releases/download/v3.31.0/cmake-3.31.0-Linux-x86_64.sh
  $ sh cmake-install.sh --skip-license --prefix=/usr/local
  $ cmake --version # CMake Installer Version: 3.31.0, Copyright (c) Kitware
  ```

mtl1979:
> BTW, standard support of ubuntu 18.04 ended over a year ago (4.23). Funnily, even debian buster (which is also out of standard support) has 3.13 available in oldstable.
> 
That applies only to normal users with no subscription... Even small companies using Ubuntu have subscription, which means security updates continue to be supported for 18.04... This is essential for servers that must run 24/7 without downtime. Upgrading to 20.04 or later would only take few minutes, but there is no guarantee that third-party software would run without recompiling and that could take hours.

am11:
> This is essential for servers that must run 24/7 without downtime

Do we expect such production server to be used to:
1. "compile" the latest version of zlib-ng?
2. use "cmake" to compile it, when the more traditional `make` option is available?

Anyway, I've lowered to 3.10 which is available in ubuntu 18.04 and it fixes the deprecation warnings with cmake 3.31 which GitHub Actions has picked up on windows machines. The chances of someone running old OS and requiring to build latest zlib-ng using cmake are unlikely to nil IMHO, but let me know if this is still a concern.

CMakeLists.txt
```diff
@@ -1,9 +1,5 @@
-cmake_minimum_required(VERSION 3.5.1)
-if(CMAKE_VERSION VERSION_LESS 3.12)
-    cmake_policy(VERSION ${CMAKE_VERSION})
-else()
-    cmake_policy(VERSION 3.5.1...3.29.0)
-endif()
+cmake_minimum_required(VERSION 3.10)
```
phprus: What do you think about using the solution:

```cmake
    cmake_minimum_required(VERSION 3.5.1)
    cmake_policy(VERSION ${CMAKE_VERSION})
else()
    cmake_minimum_required(VERSION 3.12.0)
    cmake_policy(VERSION 3.12.0...3.31.0)
endif()
```

and about increasing the cmake version requirements to 3.12 in tests?

Like this: https://github.com/phprus/zlib-ng/commit/0b96d6f6a0e958c0c48b98ef56fe7d65a5c09f05

@mtl1979

nmoinvaz:
Why are the pigz tests now failing to link?

CMakeLists.txt
am11: This sounds better. 👍

am11:
@nmoinvaz, apparently we need to keep test versions the same for google tests. I have applied minimal change, now pigz is green.

am11:
Windows arm failures are unrelated. EL9 legs timeout after 2-3 days based on the runs on other PRs: https://github.com/zlib-ng/zlib-ng/actions/runs/11941852517.

phprus:
I think we can increase the cmake version requirement to 3.11 or 3.12 in the following files:

- ``test/add-subdirectory-project/CMakeLists.txt``
- ``test/fuzz/CMakeLists.txt``
- ``test/pigz/CMakeLists.txt``

am11:
> I think we can increase the cmake version requirement to 3.11 or 3.12 in the following files:
> 
>     * `test/add-subdirectory-project/CMakeLists.txt`
> 
>     * `test/fuzz/CMakeLists.txt`
> 
>     * `test/pigz/CMakeLists.txt`

That fails pigz linkage https://github.com/zlib-ng/zlib-ng/pull/1810#issuecomment-2496098350.

phprus:
> That fails pigz linkage [#1810 (comment)](https://github.com/zlib-ng/zlib-ng/pull/1810#issuecomment-2496098350).

I think the error could be due to downgrading the cmake version (3.11 -> 3.10):

https://github.com/zlib-ng/zlib-ng/commit/428e990615913e791827e4aae7a8cb0c1f36c9e1

My tests with suggested changes pass without errors, but if we don't increase the cmake versions in the tests, then when I compile them, cmake will print warnings.

blowekamp:
Please consider this alternative: #1812

mtl1979:
> Please consider this alternative: #1812

Both alternatives have pros and cons... We shouldn't expect minimum version of cmake being available and instead always use explicit version check on code that use new or non-backwards compatible features. Cmake is known to force feed new versions like Microsoft does and we shouldn't fall in such trap.

am11:
Right, it was just a periodic version bump more than anything (with reference to the previous similar version bump) https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9875. There is no behavior change AFAICT. cmake features used by zlib-ng project are compatible with 3.5 all the way up to 3.31.

Lets take one of these changes to get rid of the warning. Ideally we should be setting the max version and if we must, it should be the latest at the time of the change (3.31 as done here). GitHub actions VMs are on 3.31 plan which was the whole point of this change.

am11:
Right, it was just a periodic version bump more than anything (with reference to the previous similar version bump) https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9875. There is no behavior change AFAICT. cmake features used by zlib-ng project are compatible with 3.5 all the way up to 3.31.

Lets take one of these changes to get rid of the warning. Ideally we should not be setting the max version and if we must, it should be the latest at the time of the change (3.31 as done here). GitHub actions VMs are on 3.31 plan which was the whole point of this change.

blowekamp:
> Ideally we should not be setting the max version and if we must, it should be the latest at the time of the change (3.31 as done here).

The CMake documentation may provide clarification for the behavior of the second argument for `cmake_minimum_requred`:
https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html

The second argument is not a "max" version. It will still work with CMake versions newer than the second "max_policy" argument. The second argument sets the max policy, in a more efficient and way less prone to emit warnings.

The following are equivalent:

A.
```
cmake_minimum_required( VERSION X...Y )
```
B.
```
cmake_minimum_required( VERSION X )
cmake_policy(VERSION X...Y)
```

However, B may generate a warning when the behavior of X may be deprecated/remove in CMake. The structure of A, should be preferred. And there are caveat with version of CMake before 3.12, hence the need to retain that if block.

The compatibility of the current CMake code with 3.31 policies was beyond the scope of my PR.

am11:
I've used @blowekamp's version. Lets use that patch. Closing.

PR #1811, opened 2024-11-24
Force use of latest Windows SDK with 32-bit ARM support

phprus:
Fix for CI error:

``
      C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Microsoft\VC\v170\Microsoft.Cpp.WindowsSDK.targets(49,5): error MSB8037: The Windows SDK version 10.0.26100.0 for Desktop C++ ARM Apps was not found. Install the required version of Windows SDK or change the SDK version in the project property pages or by right-clicking the solution and selecting "Retarget solution".  [D:\a\zlib-ng\zlib-ng\CMakeFiles\3.31.0\VCTargetsPath.vcxproj]
``

## Summary by CodeRabbit

- **New Features**
	- Added version specification for ARM architecture in CMake builds.
	- Enhanced flexibility of the NMake workflow by allowing dynamic SDK version referencing.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -502,7 +502,7 @@ jobs:
           - name: Windows MSVC ARM No Test
             os: windows-latest
             compiler: cl
-            cmake-args: -A ARM
+            cmake-args: -A ARM,version=10.0.22621.0
 
           - name: Windows MSVC ARM64 No Test
             os: windows-latest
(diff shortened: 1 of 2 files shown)


nmoinvaz (approved):

nmoinvaz:
It's weird that cmake doesn't select an existing version as the default.

phprus:
> It's weird that cmake doesn't select an existing version as the default.

Version 10.0.26100.0 is exist on runner. But it does not support 32bit ARM.

mtl1979:
Might be that Microsoft is slowly phasing out support for 32-bit operating systems... 32-bit ARM targeted ARMv7, which is quite old already. Windows 11 is supposed to be 64-bit only.

See also: https://github.com/actions/runner-images/issues/10981

Dead2 (approved):
LGTM

PR #1812, opened 2024-11-26
Address deprecated cmake version warning.

blowekamp:
Use cmake_minimum_required(VERSION <min>...<policy_max>) syntax to set the policy at the same time as the compatibile CMake version.

The CMake documentation for this call is here:
https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html

Of note is the following:
```
The optional <policy_max> version, if specified, must be at least the <min> version and affects policy settings as described in [Policy Settings](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#policy-settings). If the running version of CMake is older than 3.12, the extra ... dots will be seen as version component separators, resulting in the ...<max> part being ignored and preserving the pre-3.12 behavior of basing policies on <min>.
```

## Summary by CodeRabbit

- **New Features**
	- Enhanced configurability with new options for building features such as gzFile support, zlib compatibility, and CPU architecture detection.
	- Introduced a feature summary section for better visibility of enabled features during configuration.

- **Bug Fixes**
	- Updated minimum CMake version for broader compatibility with newer features.

- **Documentation**
	- Improved clarity on project configurations and options available for users during the build process.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,6 @@
-cmake_minimum_required(VERSION 3.5.1)
+cmake_minimum_required(VERSION 3.5.1...3.29.0)
 if(CMAKE_VERSION VERSION_LESS 3.12)
     cmake_policy(VERSION ${CMAKE_VERSION})
-else()
-    cmake_policy(VERSION 3.5.1...3.29.0)
 endif()
 message(STATUS "Using CMake version ${CMAKE_VERSION}")
 
--- a/test/add-subdirectory-project/CMakeLists.txt
+++ b/test/add-subdirectory-project/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5.1)
+cmake_minimum_required(VERSION 3.5.1...3.29.0)
 
 project(zlib-ng-add-subdirecory-test C)
 
--- a/test/fuzz/CMakeLists.txt
+++ b/test/fuzz/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5.1)
+cmake_minimum_required(VERSION 3.5.1...3.29.0)
 
 if(CMAKE_C_COMPILER_ID MATCHES "Clang")
     enable_language(CXX)


CMakeLists.txt
```diff
@@ -1,8 +1,6 @@
-cmake_minimum_required(VERSION 3.5.1)
+cmake_minimum_required(VERSION 3.5.1...3.29.0)
 if(CMAKE_VERSION VERSION_LESS 3.12)
     cmake_policy(VERSION ${CMAKE_VERSION})
```
dzenanz, line 3: I think this is not necessary with minimum version specified in x...y format.

blowekamp: From: https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html

"If the running version of CMake is older than 3.12, the extra ... dots will be seen as version component separators, resulting in the ...<max> part being ignored and preserving the pre-3.12 behavior of basing policies on <min>."

dzenanz: Also, it might make sense to start requiring CMake 3.12 as a minimum? Up to maintainers here.

mtl1979: We test zlib-ng on wide range of Linux versions and some of them don't have cmake 3.12... We avoid adding unnecessary binary packages or recompiling from sources as the current CI workflows are already pushing limits of the runners and any unnecessary additions would cause unnecessary failure. People say building newer cmake works if glibc is at least some version without actually checking what glibc version is used on each distribution and release... I've seen more than once that the bundled version is one less than what people assume it would be.

It is out of scope of each PR to repeat all the unfixed bugs in new releases when we have already downgraded CI instances to use older releases that actually can build zlib-ng successfully.

blowekamp: This change is compatible with versions of cmake before 3.12. Jus to be clear.

Dead2:
A few questions;
First off, why not set the "max" version to 3.31.0? Any particular reason?
Also, what about `test/pigz/CMakeLists.txt`? Perhaps add `cmake_policy(VERSION 3.11.0...3.29.0)` to it?

To be clear, I ask out of ignorance, not because I think anything is actually wrong.

I think that this PR would be better for backwards compatibility, so unless someone has objections I will probably be merging this one instead of #1810.

blowekamp:
> A few questions; First off, why not set the "max" version to 3.31.0? Any particular reason? 

The intention of this patch was to address the warning and not change behavior. CMake version >3.29 will still work with the CMake code but the new changes in behavior or policies of CMake will not be in effect. This preserved the current behavior of CMake and makiesthis PR a bug  fix and not change in behavior.

>Also, what about `test/pigz/CMakeLists.txt`? Perhaps add `cmake_policy(VERSION 3.11.0...3.29.0)` to it?

This CMake file had not policies applied. I did not investigate further to determine what would be correct behavior of the file.

> 
> To be clear, I ask out of ignorance, not because I think anything is actually wrong.
> 
> I think that this PR would be better for backwards compatibility, so unless someone has objections I will probably be merging this one instead of #1810.

blowekamp:
Are the failing builds chronic? Does the PR need to be rebased onto master?

Dead2:
> Are the failing builds chronic? Does the PR need to be rebased onto master?

The S390X builds are down and requires some manual work to get it work again after a long pause where a working github actions runner was unavailable on S390X.

The rest would hopefully be fixed with a rebase.

Dead2 (approved):

PR #1813, opened 2024-11-27
add env variable to control dfltcc usage

fneddy:
this patch change the hard-coded dfltcc control variables to be sources from env, just like its done in the zlib patch. It is a nearly 1-1 copy from the mechanism used [in the zlib dfltcc patch](https://github.com/iii-i/zlib/blob/dfltcc/contrib/s390/dfltcc.c#L697).

reasons for this patch:
- most distributions that use zlib [use this patch](https://github.com/madler/zlib/pull/410) to enable ibms DFLTCC on s390 platform. this patch also sources this 7 env variables at runtime. ([e.g. OpenSuse](https://build.opensuse.org/projects/openSUSE:Slowroll/packages/zlib/files/zlib-1.3-IBM-Z-hw-accelerated-deflate-s390x.patch?expand=1))
- gzip also [sources the DFLTCC env variable](https://git.savannah.gnu.org/cgit/gzip.git/tree/dfltcc.c#n151) to control the usage of the feature.
- ibms DFLTCC implementation **should NOT be used for reproducible output**. (see [here](https://github.com/zlib-ng/zlib-ng/blob/develop/arch/s390/README.md?plain=1#L39) and [here](https://github.com/zlib-ng/zlib-ng/blob/develop/arch/s390/dfltcc_deflate.c#L47)). in case of zlib or gzip, if the user wants reproducible output, they are advised to set env `DFLTCC=0`. 
- the env variables allow for finer grained switching between [software deflate and dfltcc](https://github.com/zlib-ng/zlib-ng/blob/develop/arch/s390/dfltcc_deflate.c#L41). 

all this culminates in: it would be nice to have this env control variables also in zlib-ng, so the experience is the same when using any of the tools that use DFLTCC. 

## Summary by CodeRabbit

- **New Features**
	- Introduced support for IBM Z DEFLATE CONVERSION CALL (DFLTCC) functionality, enhancing compression and decompression capabilities.
	- Added dynamic configuration options for DEFLATE parameters via environment variables.
	- Enhanced architecture-specific optimizations in the configuration script.

- **Bug Fixes**
	- Improved error handling and state management for DFLTCC decompression processes.

- **Documentation**
	- Enhanced documentation for runtime tuning parameters related to DFLTCC functionality.

- **Chores**
	- Updated the `configure` script to support new architecture checks and modularity improvements.


arch/s390/dfltcc_inflate.c
```diff
@@ -30,6 +31,9 @@ int Z_INTERNAL PREFIX(dfltcc_can_inflate)(PREFIX3(streamp) strm) {
     struct inflate_state *state = (struct inflate_state *)strm->state;
     struct dfltcc_state *dfltcc_state = &state->arch.common;
 
+	if(env_dfltcc_disabled)
+		return 0;
+
```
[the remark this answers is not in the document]
fneddy: `env_dfltcc_disabled` is only read and set **ONCE** during library [constructor run](https://github.com/fneddy/zlib-ng/blob/develop/arch/s390/dfltcc_common.c#L34). it will not change after that and no need to concurrent access is needed.

arch/s390/dfltcc_common.h
```diff
@@ -3,6 +3,17 @@
...
+/*
+   Runtime Tuning parameters.
+ */
+extern int env_dfltcc_disabled;
+extern int env_dfltcc_source_date_epoch;
+extern unsigned long env_dfltcc_level_mask;
+extern unsigned long env_dfltcc_block_size;
+extern unsigned long env_dfltcc_block_threshold;
+extern unsigned long env_dfltcc_dht_threshold;
+extern unsigned long env_dfltcc_ribm;
```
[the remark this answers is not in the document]
fneddy: resolved in new commit

arch/s390/dfltcc_common.c
```diff
@@ -0,0 +1,68 @@
...
+#endif
+    env_dfltcc_dht_threshold = xstrtoul(getenv("DFLTCC_DHT_MIN_SAMPLE_SIZE"),
+                                 DFLTCC_DHT_MIN_SAMPLE_SIZE);
+
+#ifndef DFLTCC_RIBM
+#define DFLTCC_RIBM 0
+#endif
+    env_dfltcc_ribm = xstrtoul(getenv("DFLTCC_RIBM"), DFLTCC_RIBM);
+
+}
```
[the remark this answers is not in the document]
fneddy: added sanity check for env vars where it makes sense

arch/s390/dfltcc_common.c
```diff
@@ -0,0 +1,80 @@
...
+    result = strtoul(s, &endptr, 0);
+    return (errno || *endptr) ? _default : result;
+}
+
+__attribute__((constructor)) static void init_globals(void)
+{
+    const char *env;
+
+    env = getenv("DFLTCC");
+    env_dfltcc_disabled = env && !strcmp(env, "0");
```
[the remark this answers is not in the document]
fneddy: this is the wrong place and way to communicate it. env_dfltcc_disabled can be either unset 0 or anything else then 0. all values are valid and should not produce an error message.

arch/s390/dfltcc_common.c
```diff
@@ -0,0 +1,80 @@
...
+#endif
+    env_dfltcc_block_threshold = xstrtoul(getenv("DFLTCC_FIRST_FHT_BLOCK_SIZE"),
+                                   DFLTCC_FIRST_FHT_BLOCK_SIZE);
+
+#ifndef DFLTCC_DHT_MIN_SAMPLE_SIZE
+#define DFLTCC_DHT_MIN_SAMPLE_SIZE 4096
+#endif
+    env_dfltcc_dht_threshold = xstrtoul(getenv("DFLTCC_DHT_MIN_SAMPLE_SIZE"),
+                                 DFLTCC_DHT_MIN_SAMPLE_SIZE);
+
```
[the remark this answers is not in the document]
fneddy: the threshold values can be set to arbitrary values. sanity checking is not needed here.

CMakeLists.txt
```diff
@@ -873,6 +873,9 @@ if(WITH_OPTIM)
             add_definitions(-DS390_DFLTCC_INFLATE)
             list(APPEND ZLIB_ARCH_SRCS ${ARCHDIR}/dfltcc_inflate.c)
         endif()
+		if(WITH_DFLTCC_DEFLATE OR WITH_DFLTCC_INFLATE)
```
iii-i: Here and in a few other places there are tabs instead of spaces.

arch/s390/dfltcc_common.c
```diff
@@ -0,0 +1,80 @@
...
+    errno = 0;
+    result = strtoul(s, &endptr, 0);
+    return (errno || *endptr) ? _default : result;
+}
+
+__attribute__((constructor)) static void init_globals(void)
+{
+    const char *env;
+
+    env = getenv("DFLTCC");
```
iii-i, line 34: zlib has secure_getenv() here and in other places.

arch/s390/dfltcc_common.c
```diff
@@ -0,0 +1,80 @@
...
+    env = getenv("SOURCE_DATE_EPOCH");
+    env_dfltcc_source_date_epoch = !!env;
+
+#ifndef DFLTCC_LEVEL_MASK
+#define DFLTCC_LEVEL_MASK 0x2
+#endif
+    env_dfltcc_level_mask = xstrtoul(getenv("DFLTCC_LEVEL_MASK"),
+                              DFLTCC_LEVEL_MASK);
+    if(env_dfltcc_level_mask > 10) {
+        fprintf(stderr, "DFLTCC_LEVEL_MASK %lu out of range. Allowed range is 0-10. Resetting to default(%d)\n",
```
iii-i, line 46: While having diagnostics is good, I don't think we should be printing to console from a widely used shared library. This may create nasty consequences if it's used in software which treats stderr as machine-readable.

arch/s390/dfltcc_common.c
```diff
@@ -0,0 +1,80 @@
...
+            env_dfltcc_level_mask, DFLTCC_LEVEL_MASK);
+        env_dfltcc_level_mask = DFLTCC_LEVEL_MASK;
+    }
+
+#ifndef DFLTCC_BLOCK_SIZE
+#define DFLTCC_BLOCK_SIZE 1048576
+#endif
+    env_dfltcc_block_size = xstrtoul(getenv("DFLTCC_BLOCK_SIZE"),
+                              DFLTCC_BLOCK_SIZE);
+    if( (env_dfltcc_block_size & (env_dfltcc_block_size - 1) /* power of 2 */)
```
iii-i: I think it's okay to have any value here. 256k is optimal, but if a user wants a smaller one for whatever reason, we shouldn't prevent that.

arch/s390/dfltcc_common.h
```diff
@@ -3,6 +3,72 @@
...
+  0 or unset: use DFLTCC_CMPR
+  1: disable DFLTCC_CMPR
+ */
+extern int env_dfltcc_source_date_epoch;
+
+/*
+  disable DFLTCC for specific compression levels
+
+  env variable: DFLTCC_LEVEL_MASK
+  valid range: 0-10 / 0x0-0xA
```
iii-i, line 33: It's a bitmask, so the valid range is [0, (1<<10)).

arch/s390/dfltcc_deflate.c
```diff
@@ -59,6 +62,9 @@ static inline int dfltcc_can_deflate_with_params(PREFIX3(streamp) strm, int leve
 int Z_INTERNAL PREFIX(dfltcc_can_deflate)(PREFIX3(streamp) strm) {
     deflate_state *state = (deflate_state *)strm->state;
 
+	if(env_dfltcc_disabled)
```
iii-i: Here and in a few other places: missing space after if.

CMakeLists.txt
fneddy: will fix

arch/s390/dfltcc_common.c
fneddy: secure_getenv is behind the `#define _GNU_SOURCE` is it ok to pull it in?

arch/s390/dfltcc_common.h
fneddy: ~actually its a [shift for a mask](https://github.com/zlib-ng/zlib-ng/blob/develop/arch/s390/dfltcc_deflate.c#L41).~
~I see three possibilities to fix this:~

~1. just fix the comment~
~2. rename env var to DFLTCC_LEVEL_MASK_SHIFT~
~3. change it to allow a 10 bit mask~

oh my bad what i wrote is wrong

arch/s390/dfltcc_deflate.c
fneddy: will fix

arch/s390/dfltcc_common.c
fneddy: i see this options:

1. only print if `isatty() == 1) 
2. only print with special DEBUG compile flag
3. don't print at all

what would be an acceptable solution?

arch/s390/dfltcc_common.c
fneddy: will fix

arch/s390/dfltcc_common.c
iii-i: There is a precedent with ZLIB_DEBUG in deflate_p.h, so I guess doing the same would be acceptable.

arch/s390/dfltcc_common.h
fneddy: make totally sense. I'll fix the comment and also add a check for a valid mask when reading the env variable.

arch/s390/dfltcc_common.c
iii-i: I think _GNU_SOURCE is okay: https://stackoverflow.com/questions/5582211/what-does-define-gnu-source-imply. The resulting code may (but not necessarily will) be incompatible with musl though.

iii-i: Another possibility is to open-code it using `getauxval(AT_SECURE)`.

configure
```diff
@@ -1885,6 +1885,12 @@ EOF
                 ARCH="${ARCH}+dfltcc-inflate"
             fi
 
+            if test $builddfltccinflate -eq 1 -o $builddfltccdeflate -eq 1; then
+                ARCH_STATIC_OBJS="${ARCH_STATIC_OBJS} dfltcc_common.o"
+                ARCH_SHARED_OBJS="${ARCH_SHARED_OBJS} dfltcc_common.lo"
+                ARCH="${ARCH}+dfltcc-common"
```
Dead2, line 1891: **question:** 
Is this actually something that needs to be added to the ARCH string?

Dead2:
I am honestly very uncertain about this one.
Zlib rejected parsing of env variables when the DFLTCC patchset was initially suggested, and so did we, with the reasoning being that it is adding another possible attack vector.

There is no way for an application to disable/detect/override this if it wanted (for extra security or whatever).

So, pending other reviews and opinions, what I'd suggest is the following:
1. This needs to be opt-in during compile, not default enabled.
2. Don't hide the defines from `dfltcc_detail.h` in `dfltcc_common.c`, either keep them there or move them to `dfltcc_common.h` if needed, and add a comment that they can be overridden if env parsing is enabled.

configure
fneddy: No, as far as I see the $ARCH string is only used during `./configure` time. AFAIK it has no actual effect on the build. I added `-common` to it for consistency, because the other dfltcc options also append to `$ARCH` if they are enabled. 

But because only the `s390` target has this behavior, the more consisted solution would be to [remove also all other appends](https://github.com/zlib-ng/zlib-ng/blob/develop/configure#L1864-L1898).

iii-i:
Thanks for the review! What do you think about putting the environment variable handling under `ZLIB_COMPAT`? While zlib does not officially support DFLTCC at all (in 5 years there was no reaction from Mark Adler), de-facto it's distributed as a distro patch, which includes the environment variable support. So it would match the spirit of `ZLIB_COMPAT`.

configure
iii-i: I don't remember why I added them and I won't object to removing them - they seem to only cause confusion.

arch/s390/dfltcc_common.c
mtl1979: We shouldn't use environment variables unless absolutely necessary. Just because other projects are using doesn't mean we should use them. There are several security issues related to using environment variables. If we must use environment variables, we should validate all values but silently ignore invalid input instead of crashing.

Availability of `secure_getenv` should be checked at configure time if it is used.

arch/s390/dfltcc_common.c
mtl1979: Either 2 or 3 is fine for me...

configure
mtl1979: There are other architectures that require adding suffixes to `ARCH` and won't work without the addition. It seems to be specific to compiler version. Some compiler versions enable extensions by default, others enable them only when FP or vector unit is explicitly enabled on command line.

mtl1979:
> I am honestly very uncertain about this one. Zlib rejected parsing of env variables when the DFLTCC patchset was initially suggested, and so did we, with the reasoning being that it is adding another possible attack vector.

This is exactly why I'm against using environment variables. Adding explicit validation to parsing them adds unnecessary complexity to the zlib-ng code with very little or no benefit.

But every decision should be a vote in my opinion and there is enough long-time contributors to have a fair vote. As said earlier, new contributors shouldn't participate in the vote but can justify their point of view before the votes are cast.

mtl1979:
> Thanks for the review! What do you think about putting the environment variable handling under `ZLIB_COMPAT`? While zlib does not officially support DFLTCC at all (in 5 years there was no reaction from Mark Adler), de-facto it's distributed as a distro patch, which includes the environment variable support. So it would match the spirit of `ZLIB_COMPAT`.

Until the support is merged to zlib, there is not much point to hide it behind `ZLIB_COMPAT`... It could however be own configuration option, but even then it would make the code (a little) less readable. I won't be reviewing the code before all the concerns are addressed. I will comment when changing something might make me shift to thinking the patch is clean and safe enough to be good idea to merge.

configure
iii-i: I see now why they are needed elsewhere, thanks. In this particular case they definitely don't serve this function. So let's drop them (perhaps in a separate PR?).

iii-i:
One question regarding the security considerations, what is the zlib-ng's threat model with respect to environment variables?

My particular concern is that if the threat models allows an attacker to set arbitrary environment variables, one can already cause a DoS even without zlib-ng:
```
$ GLIBC_TUNABLES=glibc.rtld.optional_static_tls=999999999999 ls; echo $?
cannot allocate TLS data structures for initial thread
127
```
and I suspect that someone more skilled than I am can find a way to perform arbitrary code execution this way (e.g., by creatively abusing things like `LD_PRELOAD` or `BASH_FUNC_ls%%`).

mtl1979:
> One question regarding the security considerations, what is the zlib-ng's threat model with respect to environment variables?
> 
> My particular concern is that if the threat models allows an attacker to set arbitrary environment variables, one can already cause a DoS even without zlib-ng:
> ```

We shouldn't worry about other applications or libraries crashing due to malicious environment variables... Environment variables generally persist, so any application can set malicious variable and cause zlib-ng to crash itself and any application using it... Like I said earlier, proper validation must be performed to avoid things like wrong type, out of range values or stack overflow.

fneddy:
i think it is obvious that this subject needs much more thought and the pr needs rework. For now I will mark this PR ad draft and rework it.

PR #1820, opened 2024-11-30
Add a fallback to ALIGNED_ macro for other compilers

winspool:
Since the affectes compiler does not support SSE/AVX,
a bigger alignment is not needed. 

## Summary by CodeRabbit

- **New Features**
	- Introduced a fallback definition for the `ALIGNED_` macro to enhance compatibility with various compilers.

- **Bug Fixes**
	- Improved macro flexibility for future compiler-specific implementations.

### DIFF
--- a/zbuild.h
+++ b/zbuild.h
@@ -200,6 +200,9 @@
 #  define ALIGNED_(x) __attribute__ ((aligned(x)))
 #elif defined(_MSC_VER)
 #  define ALIGNED_(x) __declspec(align(x))
+#else
+/* TODO: Define ALIGNED_ for your compiler */
+#  define ALIGNED_(x)
 #endif
 
 #ifdef HAVE_BUILTIN_ASSUME_ALIGNED


Dead2:
@winspool Just wondering, what compiler is hitting this?

winspool:
> @winspool Just wondering, what compiler is hitting this?

OpenWatcom v2 on linux

TODO:
*  a patch related to cpuid()
   (Fallback:  returning 0 in the variables with the registernames)

* a patch related to xgetbv
  (Fallback: returning 0 should be ok)

* a patch relatet to zng_alloc
  OW does not support posix_memalign() and alligned_alloc() yet
  Using the normal malloc() should work, 
  as the intinsics are not supported
  and the higher alignment is probably not required.


Yes, all together the result is a lot slower as possible,
but at a first step, it has to work.

winspool:
> @winspool Just wondering, what compiler is hitting this?

My  `_build` script is here:
 
https://github.com/winspool/_build.git

I  build with autotools, which results in `configure  --static  --32` 
and for OpenWatcom, i have to overwrite AR and ARFLAGS:

```
HOST=" "  AR="wlib"  ARFLAGS="-n"  _build  --static  --32
```

Btw, cross-compiling on linux for _WIN32 fails,
but cross-compiling  ( including `make test` ) works here
for DOS with the causeway extender.
(works also with the dos4gw extender)

winspool:
Proposed additional portability-patches to allow OpenWatcom and tcc to compile zlib-ng:

Use a fallback for cpuid related functions: 
[0003-portability-Add-a-fallback-for-cpuid-related-functio.txt](https://github.com/user-attachments/files/17999685/0003-portability-Add-a-fallback-for-cpuid-related-functio.txt)

Use a fallback for memalign:
[0002-portability-Check-for-memalign-when-available.-Other.txt](https://github.com/user-attachments/files/17999688/0002-portability-Check-for-memalign-when-available.-Other.txt)

------

When patch 0002 is an acceptable style, i can update `configure` and `CMakeList.txt`
in a similar style for the cpuid related functions/includes.

This will also fix cross compiling for windows (tested with `i386-win32-tcc` and `x86_64-win32-tcc`)

winspool:
> @winspool Just wondering, what compiler is hitting this?

Forgot to mention, that tcc is also affected

mtl1979:
Generally I'm not against adding support for new compilers, but it seems these compilers target "configurations" that are not highest priority for zlib-ng as most of the optimisations get disabled.

winspool:
> Generally I'm not against adding support for new compilers,

Great.
The improved speed look interesting,
but requiring to use one of the 3 mainstream compiler
for the speedup is not needed.

I usually use `autotools`, but the provided `configure` script
is simple enough to be easily extended.
And for `cmake`, i test with gcc and clanng.
I have no MSVC installed, and hope, that zlib-ng is able to extended and test the related parts it in a similar way.

> but it seems these compilers target "configurations" 
> that are not highest priority for zlib-ng as most of the optimisations get disabled.

Get it runing without optimization is only the first step.
I hope to get the optimized code running without depending on a specific toolchain, when possible (thanks to the intrinsics).

The `C++` tests won't work, when using tcc (Tiny C Compiler: only C)
or when using OpenWatcom (Missing Many C++11 features),
but the simple C test work on all compiler, which i have tested.

winspool:
Please give me some infos, what i can do,
to get my suggested portability changes in the main tree?

Is an update to this simple PR needed?
Are the other attached proposed changes OK?
How many portability changes in one PR is acceptable?
What else?

winspool:
I didn't looked at the CodeCov integration yet,
but i have additional changes to test the different memory allocation strategies locally.

mtl1979:
> Please give me some infos, what i can do, to get my suggested portability changes in the main tree?
> 
> Is an update to this simple PR needed? Are the other attached proposed changes OK? How many portability changes in one PR is acceptable? What else?

Adding support for one compiler per pull request is usually good idea as it makes reviewing the necessary changes easier... If two or more compilers benefit for overlapping changes, then adding the overlapping part first as own commit in PR makes it clear what changes overlap and what are specific to each compiler. If just the overlapping part doesn't result in successful build, then combine it with enough code for one compiler configuration that results in successful build (smaller initial commit is better). 

We allow incremental patches that span multiple pull requests, but generally single PR shouldn't break the build for any configuration, or revert changes in previous PR in same patch series. Reverting older changes is acceptable as long as the older change was clearly bad or it is no longer needed due to other changes committed in between.

As far as I know latest stable version of Watcom compiler is quite old (it was released the same day my older girl was born), but I haven't yet checked what newer features are planned for the current beta version or when the feature freeze is for that branch.

I know almost nothing about Tiny C Compiler, so can't comment on that yet...

mtl1979:
Patch 3 doesn't look quite right, but the first two have nothing seriously bad that I can notice... Clang as far as I know defines `__GNUC__` as it reuses parts of GNU toolchain for binary compatibility. This is why to test for GCC, it is necessary to test that `__GNUC__` is defined, but the preprocessor defines for other forked compilers are not defined, including for example Clang and Intel compilers.

Dead2:
In general, we don't support this compiler, but I am not against merging patches like this one that really has no downside.
In this case, I think we can actually call this a bug, the macro should have had a fallback.

Dead2 (approved):

winspool:
> In general, we don't support this compiler, but I am not against merging patches like this one that really has no downside.
> In this case, I think we can actually call this a bug, the macro should have had a fallback.


Thanks.
What's next for this pull requests?

KungFuJesus:
Curiously this macro was causing problems for me with CMake, GCC, and PowerPC.  I never got to the bottom of it, though.  We actually _do_ need alignment for VMX, otherwise you get the fun side effect of a load happening at the next alignment boundary, silently.

Dead2:
> Curiously this macro was causing problems for me with CMake, GCC, and PowerPC. I never got to the bottom of it, though. We actually _do_ need alignment for VMX, otherwise you get the fun side effect of a load happening at the next alignment boundary, silently.

@KungFuJesus I am a little uncertain about this comment, so I hope you can clarify; Is this a Pro/Con on this PR, or a suggestion for later, or just a comment?

KungFuJesus:
> > Curiously this macro was causing problems for me with CMake, GCC, and PowerPC. I never got to the bottom of it, though. We actually _do_ need alignment for VMX, otherwise you get the fun side effect of a load happening at the next alignment boundary, silently.
> 
> @KungFuJesus I am a little uncertain about this comment, so I hope you can clarify; Is this a Pro/Con on this PR, or a suggestion for later, or just a comment?

Mostly just a comment.  The issue likes somewhere in CMake, I think.  There's a pending PR for PPC that suddenly breaks the machinery around how we specify this macro: https://github.com/zlib-ng/zlib-ng/pull/1690
It's not entirely tangential, it might make things silently fail if that other PR is merged.  I really need to get to the bottom of that.

Dead2:
> Mostly just a comment. The issue likes somewhere in CMake, I think. There's a pending PR for PPC that suddenly breaks the machinery around how we specify this macro: #1690 It's not entirely tangential, it might make things silently fail if that other PR is merged. I really need to get to the bottom of that.

Ah #1690, now I get it 👍

PR #1823, opened 2024-12-04
added in-tree build artifacts to .gitignore

fneddy:
when build with traditional `configure && make` in-tree, some build artifacts will pop up that are not excluded by `.gitignore`. 

## Summary by CodeRabbit

- **Chores**
	- Updated the `.gitignore` file to exclude additional files and directories, enhancing version control management.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,9 @@
 /switchlevels
 /zlib.pc
 /zlib-ng.pc
+/zconf-ng.h.included
+/zlib_name_mangling-ng.h
+/zlib-ng.h
 
 .DS_Store
 *_fuzzer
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -61,6 +61,10 @@ zconf-ng.h.cmakein
 ztest*
 /test/CTestTestfile.cmake
 /test/cmake_install.cmake
+/zconf-ng.h.included
```
mtl1979: Would be cleaner to have this few lines above where similar filenames are excluded. Remove initial `/` to match style of other similar filenames.

fneddy: i'll do that

Dead2 (approved):
LGTM

PR #1826, opened 2024-12-14
adler32_rvv: Fix two overflow problems

Icenowy:
There are currently two overflow problems in adler32_rvv implementation, which can lead to wrong results for some input, and these problems could be easily exhibited when running `git fsck` with zlib-ng suitituting the system zlib on a big git repository.

These problems and the solutions are the following:

- When the input data is long enough, the v_buf32_accu can overflow too. Add it to the modulo code that happens per ~NMAX bytes.
- When the remaining data less than vl bytes, the code falls back to a scalar implementation; however the sum2 and alder2 values are just reduced from vectors and could be very big that makes sum2 overflows in the scalar code. Modulo them before the scalar code to prevent such overflow (because vl is surely quite smaller than NMAX).

## Summary by CodeRabbit

- **New Features**
	- Enhanced the Adler-32 checksum calculations to ensure values remain within defined bounds.

- **Bug Fixes**
	- Adjusted checksum accumulation logic to prevent exceeding limits during calculations.

### DIFF
--- a/arch/riscv/adler32_rvv.c
+++ b/arch/riscv/adler32_rvv.c
@@ -72,6 +72,7 @@ static inline uint32_t adler32_rvv_impl(uint32_t adler, uint8_t* restrict dst, c
         /* do modulo once each block of NMAX size */
         if (++cnt >= nmax_limit) {
             v_adler32_prev_accu = __riscv_vremu_vx_u32m4(v_adler32_prev_accu, BASE, vl);
+            v_buf32_accu = __riscv_vremu_vx_u32m4(v_buf32_accu, BASE, vl);
             cnt = 0;
         }
     }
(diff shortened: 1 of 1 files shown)


Icenowy:
Added another overflow fix about the reduction from vector to scalar.

nmoinvaz (commented):
Is it possible to add unit tests for this or is it not practical?

Icenowy:
For the first problem, I have a 2M random .bin file that can lead to this issue being exhibited when running adler32 on it.
For the second/third problem, I think maybe carefully crafted data is needed, and I currently do not have suitable test data (I tested and found these problems by running `git fsck` on https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git )

Icenowy:
For the second problem (in the commit message now), I think crafting a dataset with very long `len - left` (well technically this means `len / vl * vl`, I think), a big enough initial adler32 value (0xFFF0FFF0 ? 0xFFF0 equals to 65520) could make it easily exhibit?

Then for the third, attaching (vl - 1) 0xFF's to a data that will be calculated to very big sum2 and adler may exhibit the problems?

Well these two dataset construction methods both depend on a pre-determined vl ...

(My testing platform is SpacemiT K1, which should have VLEN=256 and the vl is 32 in this situation)

Icenowy:
And surely the behavior between QEMU and K1 differs on these problems -- QEMU defaults to VLEN=128 (vl = 16).

Dead2 (approved):
LGTM

nmoinvaz (approved):

PR #1831, opened 2024-12-20
Make big endians first class citizens again

KungFuJesus:
No longer do the big iron on yore which lack SIMD optimized loads need to search strings a byte at a time like primitive machines of the vax era. This guard here was mostly due to the fact that the string comparison was searched with "count trailing zero", which assumes an endianness.  We can just conditionally use leading zeros when on big endian and stop using the extremely naive C implementation. This makes things a tad bit faster.

## Summary by CodeRabbit

- **New Features**
	- Introduced unaligned memory comparison functions for 16, 32, and 64 bits.
  
- **Bug Fixes**
	- Enhanced conditional compilation for various functions, improving compatibility across architectures.

- **Documentation**
	- Updated macro definitions for clarity and consistency regarding compiler features and architecture-specific optimizations.

### DIFF
--- a/arch/generic/compare256_c.c
+++ b/arch/generic/compare256_c.c
@@ -57,7 +57,8 @@ Z_INTERNAL uint32_t compare256_c(const uint8_t *src0, const uint8_t *src1) {
 
 #include "match_tpl.h"
 
-#if BYTE_ORDER == LITTLE_ENDIAN && OPTIMAL_CMP >= 32
+#if OPTIMAL_CMP >= 32
+
 /* 16-bit unaligned integer comparison */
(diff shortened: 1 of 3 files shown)


nmoinvaz (approved):

Dead2 (approved):
LGTM

KungFuJesus:
Hmm the failed CI is odd... Gtest passed for me, but I'll try to reproduce it. Is the qemu target little or big endian?

KungFuJesus:
@Dead2 Ahhh, it's the RLE methods that are broken because they are slightly different.  Ok, I'm working on a fix.

PR #1833, opened 2024-12-21
Fix unaligned access in ACLE based crc32

KungFuJesus:
This fixes a rightful complaint from the alignment sanitizer that we alias memory in an unaligned fashion. A nice added bonus is that this improves performance a tiny by on the larger buffers, perhaps due to loops that idiomatically decrement a count and increment a single buffer pointer rather than the maze of conditional pointer reassignments.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved buffer handling for CRC computation, ensuring accurate processing of data.
  - Enhanced testing for alignment issues in CRC32 calculations.

- **New Features**
  - Streamlined the CRC calculation process by utilizing local variables for buffer data, enhancing performance.
  - Introduced a new test class to verify CRC32 functionality with varying alignment offsets.

### DIFF
--- a/arch/arm/crc32_acle.c
+++ b/arch/arm/crc32_acle.c
@@ -11,9 +11,9 @@
 
 Z_INTERNAL Z_TARGET_CRC uint32_t crc32_acle(uint32_t crc, const uint8_t *buf, size_t len) {
     Z_REGISTER uint32_t c;
-    Z_REGISTER const uint16_t *buf2;
-    Z_REGISTER const uint32_t *buf4;
-    Z_REGISTER const uint64_t *buf8;
+    Z_REGISTER uint16_t buf2;
+    Z_REGISTER uint32_t buf4;
+    Z_REGISTER uint64_t buf8;
 
     c = ~crc;
 
@@ -29,45 +29,43 @@ Z_INTERNAL Z_TARGET_CRC uint32_t crc32_acle(uint32_t crc, const uint8_t *buf, si
             len--;
         }
 
-        if ((len >= sizeof(uint16_t)) && ((ptrdiff_t)buf & sizeof(uint16_t))) {
-            buf2 = (const uint16_t *) buf;
-            c = __crc32h(c, *buf2++);
+        if ((len >= sizeof(uint16_t)) && ((ptrdiff_t)buf & (sizeof(uint32_t) - 1))) {
+            buf2 = *((uint16_t*)buf);
+            c = __crc32h(c, buf2);
+            buf += sizeof(uint16_t);
             len -= sizeof(uint16_t);
-            buf4 = (const uint32_t *) buf2;
-        } else {
-            buf4 = (const uint32_t *) buf;
         }
 
-        if ((len >= sizeof(uint32_t)) && ((ptrdiff_t)buf & sizeof(uint32_t))) {
-            c = __crc32w(c, *buf4++);
+        if ((len >= sizeof(uint32_t)) && ((ptrdiff_t)buf & (sizeof(uint64_t) - 1))) {
+            buf4 = *((uint32_t*)buf);
+            c = __crc32w(c, buf4);
             len -= sizeof(uint32_t);
+            buf += sizeof(uint32_t);
         }
 
-        buf8 = (const uint64_t *) buf4;
-    } else {
-        buf8 = (const uint64_t *) buf;
     }
 
     while (len >= sizeof(uint64_t)) {
-        c = __crc32d(c, *buf8++);
+        buf8 = *((uint64_t*)buf);
+        c = __crc32d(c, buf8);
         len -= sizeof(uint64_t);
+        buf += sizeof(uint64_t);
     }
 
     if (len >= sizeof(uint32_t)) {
-        buf4 = (const uint32_t *) buf8;
-        c = __crc32w(c, *buf4++);
+        buf4 = *((uint32_t*)buf);
+        c = __crc32w(c, buf4);
         len -= sizeof(uint32_t);
-        buf2 = (const uint16_t *) buf4;
-    } else {
-        buf2 = (const uint16_t *) buf8;
+        buf += sizeof(uint32_t);
     }
 
     if (len >= sizeof(uint16_t)) {
-        c = __crc32h(c, *buf2++);
+        buf2 = *((uint16_t*)buf);
+        c = __crc32h(c, buf2);
         len -= sizeof(uint16_t);
+        buf += sizeof(uint16_t);
     }
 
-    buf = (const unsigned char *) buf2;
     if (len) {
         c = __crc32b(c, *buf);
     }
--- a/test/test_crc32.cc
+++ b/test/test_crc32.cc
@@ -8,6 +8,8 @@
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
+#include "zutil.h"
+#include "zutil_p.h"
 
 extern "C" {
 #  include "zbuild.h"
@@ -195,6 +197,22 @@ public:
     }
 };
 
+/* Specifically to test where we had dodgy alignment in the acle CRC32
+ * function. All others are either byte level access or use intrinsics
+ * that work with unaligned access */
+class crc32_align : public ::testing::TestWithParam<int> {
+public:
+    void hash(int param, crc32_func crc32) {
+        uint8_t *buf = (uint8_t*)zng_alloc(sizeof(uint8_t) * (128 + param));
+        if (buf != NULL) {
+            (void)crc32(0, buf + param, 128);
+        } else {
+            FAIL();
+        }
+        zng_free(buf);
+    }
+};
+
 INSTANTIATE_TEST_SUITE_P(crc32, crc32_variant, testing::ValuesIn(tests));
 
 #define TEST_CRC32(name, func, support_flag) \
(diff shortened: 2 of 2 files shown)


ccawley2011:
The code is supposed to ensure that the pointers are aligned by performing smaller loads until alignment is reached, so if that isn't happening it's likely that there's something else wrong.

Using `memcpy` will likely hurt performance on operating systems where unaligned access is forbidden, since it'll need to load each word one byte at a time before combining them.

KungFuJesus:
> The code is supposed to ensure that the pointers are aligned by performing smaller loads until alignment is reached, so if that isn't happening it's likely that there's something else wrong.
> 
> Using `memcpy` will likely hurt performance on operating systems where unaligned access is forbidden, since it'll need to load each word one byte at a time before combining them.

It doesn't though, as the compiler optimizes memcpys of fixed lengths to call builtins that translate the calls to direct unaligned loads on architectures that support it (which ARM does).  This is more to ensure that we're calling well defined behavior within C.  The sanitizer detected unaligned access at runtime, so the ladder of ifs had a lapse in proper alignment checks.  Actually I think I see where the alignment checks are failing because the & trick for powers of 2 requires a subtraction of one.  This is being done for the first one but not subsequent ones.

KungFuJesus:
Now the access should actually be aligned to the data type being read (for the slight benefit of speed on any particularly ARM chip).

ccawley2011:
> It doesn't though, as the compiler optimizes memcpys of fixed lengths to call builtins that translate the calls to direct unaligned loads on architectures that support it (which ARM does).

The behaviour on ARMv6 and newer is dependent on how the operating system configures alignment exceptions - if they're enabled (or ARMv4/v5 behaviour is enabled on ARMv6), applications and libraries will need to be compiled without unaligned access, and the behaviour observed in PR #1548 will occur instead. RISC OS is an example of a platform that both supports ARMv8 with CRC extensions and has alignment exceptions enabled by default.

KungFuJesus:
Hmm, so it sounds like the may_alias attribute is needed for those cases.  I can write this so that it just dereferences a pointer but it's hard to imagine a case where the compiler wouldn't treat that the same way as us calling a fixed sized memcpy as it sees we're assigning the pointer from something that only has 1 byte alignment.  FWIW, with the full test corpora this passes without memcpy and simply just deref'ing the pointers as the respective types, so you shouldn't be seeing unaligned access regardless.  The memcpys are more there as a guard rail (though it sounds like one you can't rely on if you're on RISC OS, lest you want to compile your binaries with forced memcpy's there).

KungFuJesus:
@ccawley2011 It's now memcpy free.  We'll have to rely on the alignment sanitizer to police it here.

nmoinvaz (approved):

Dead2:
### Aarch64 rpi5
#### Before PR
```
crc32/acle/1              2.50 ns         2.50 ns    279907009
crc32/acle/8              4.17 ns         4.17 ns    167944184
crc32/acle/12             4.17 ns         4.17 ns    167945907
crc32/acle/16             4.58 ns         4.58 ns    152674019
crc32/acle/32             6.67 ns         6.67 ns    104957581
crc32/acle/64             8.34 ns         8.34 ns     83971953
crc32/acle/512            35.1 ns         35.1 ns     19920130
crc32/acle/4096            226 ns          226 ns      3092795
crc32/acle/32768          1720 ns         1720 ns       406891
crc32/acle/262144        14683 ns        14683 ns        47675
crc32/acle/4194304      240527 ns       240526 ns         2905
```
#### After PR
```
crc32/acle/1              2.50 ns         2.50 ns    279910077
crc32/acle/8              4.17 ns         4.17 ns    167945750
crc32/acle/12             4.17 ns         4.17 ns    167946288
crc32/acle/16             4.58 ns         4.58 ns    152674642
crc32/acle/32             6.67 ns         6.67 ns    104964779
crc32/acle/64             8.34 ns         8.34 ns     83969360
crc32/acle/512            35.7 ns         35.7 ns     19554344
crc32/acle/4096            226 ns          226 ns      3092795
crc32/acle/32768          1720 ns         1720 ns       406898
crc32/acle/262144        14666 ns        14665 ns        47730
crc32/acle/4194304      240605 ns       240604 ns         2906
```
Performance here seems to be pretty much identical.

Dead2 (approved):
LGTM

phprus:
Unaligned access is not fixed

Ubuntu GCC AARCH64 No NEON UBSAN (https://github.com/zlib-ng/zlib-ng/actions/runs/12447135765/job/34750296246?pr=1833):

```
/home/runner/work/zlib-ng/zlib-ng/arch/arm/crc32_acle.c:40:18: runtime error: load of misaligned address 0x00550046cc52 for type 'uint32_t', which requires 4 byte alignment
```

Dead2:
> Unaligned access is not fixed
> 
> Ubuntu GCC AARCH64 No NEON UBSAN ([zlib-ng/zlib-ng/actions/runs/12447135765/job/34750296246?pr=1833](https://github.com/zlib-ng/zlib-ng/actions/runs/12447135765/job/34750296246?pr=1833)):
> 
> ```
> /home/runner/work/zlib-ng/zlib-ng/arch/arm/crc32_acle.c:40:18: runtime error: load of misaligned address 0x00550046cc52 for type 'uint32_t', which requires 4 byte alignment
> ```

Thanks for noticing, we have had many months of failing CIs due to github not supporting s390x themselves, so we have gotten too used to CI always failing.

KungFuJesus:
Ugh, I ran through the whole data test corpus without issue but alignment is just kind of a chance thing if we can't generate unaligned input. Maybe I'll need to test it that way, forcing unaligned input.

arch/arm/crc32_acle.c
```diff
@@ -29,45 +29,43 @@ Z_INTERNAL Z_TARGET_CRC uint32_t crc32_acle(uint32_t crc, const uint8_t *buf, si
             len--;
         }
 
-        if ((len >= sizeof(uint16_t)) && ((ptrdiff_t)buf & sizeof(uint16_t))) {
-            buf2 = (const uint16_t *) buf;
-            c = __crc32h(c, *buf2++);
+        if ((len >= sizeof(uint16_t)) && ((ptrdiff_t)buf & (sizeof(uint16_t) - 1))) {
```
phprus:
```c
if ((len >= sizeof(uint16_t)) && ((ptrdiff_t)buf & (sizeof(uint32_t) - 1))) {
```

arch/arm/crc32_acle.c
```diff
@@ -29,45 +29,43 @@ Z_INTERNAL Z_TARGET_CRC uint32_t crc32_acle(uint32_t crc, const uint8_t *buf, si
...
+            buf += sizeof(uint16_t);
             len -= sizeof(uint16_t);
-            buf4 = (const uint32_t *) buf2;
-        } else {
-            buf4 = (const uint32_t *) buf;
         }
 
-        if ((len >= sizeof(uint32_t)) && ((ptrdiff_t)buf & sizeof(uint32_t))) {
-            c = __crc32w(c, *buf4++);
+        if ((len >= sizeof(uint32_t)) && ((ptrdiff_t)buf & (sizeof(uint32_t) - 1))) {
```
phprus:
```c
if ((len >= sizeof(uint32_t)) && ((ptrdiff_t)buf & (sizeof(uint64_t) - 1))) {
```

This should resolve the alignment issue.

KungFuJesus: Ahhh that makes sense, we still want to see if we're a modulus of 8 bytes, the prior adjustments in the loop are what guarantee we're aligned to the next steps up.  I'm testing this with a more reliable test (controlling the allocation in the benchmark input).  I should have the fix ready shortly if it proves correct for all alignments.

KungFuJesus: Thanks for that, I was going crazy trying to figure out how this ever would have worked.  The trick is that even if there isn't a combination of aligned loads that get us to 8 byte alignment the order in which that's done means that it will at least be aligned to the next biggest size for the tail of the remainder peeling.

phprus:
Test for unaligned access (pseudocode):

```c
char buf[32];
size_t len = sizeof(buf);
for (size_t i = 0; i < len; i++)
{
    buf[i] = 'a' + (i % 26);
}

for (size_t offset = 0; offset < sizeof(uint64_t); offset++)
{
    for (size_t current_len = 1; current_len < (current_len - offset); current_len++)
    {
        c = crc(0, buf + offset, current_len);
        ignore(c);
    }
}
```

KungFuJesus:
> Test for unaligned access (pseudocode):
> 
> ```c
> char buf[32];
> size_t len = sizeof(buf);
> for (size_t i = 0; i < len; i++)
> {
>     buf[i] = 'a' + (i % 26);
> }
> 
> for (size_t offset = 0; offset < sizeof(uint64_t); offset++)
> {
>     for (size_t current_len = 1; current_len < (current_len - offset); current_len++)
>     {
>         c = crc(0, buf + offset, current_len);
>         ignore(c);
>     }
> }
> ```

Yep that's more or less how I tested it, I just did it in the benchmark code where we start with a 64 byte aligned address and walked values 1-7 for that.

phprus:
> Yep that's more or less how I tested it, I just did it in the benchmark code where we start with a 64 byte aligned address and walked values 1-7 for that.

Maybe add this test to zlib-ng?

KungFuJesus:
Rebased to go all green.

KungFuJesus:
> > Yep that's more or less how I tested it, I just did it in the benchmark code where we start with a 64 byte aligned address and walked values 1-7 for that.
> 
> Maybe add this test to zlib-ng?

I mean my test consisted of running that in conjunction with ubsan.  If not compiled with ubsan it's not much of a test but I suppose I could write it as at least one runner will run it with that.  I've never written gtests before, though.

KungFuJesus:
Ok, I added a gtest specifically for this.

test/test_crc32.cc
```diff
@@ -195,7 +201,22 @@ class crc32_variant : public ::testing::TestWithParam<crc32_test> {
...
+class crc32_align : public ::testing::TestWithParam<int> {
+public:
+    void hash(int param, crc32_func crc32) {
+        uint8_t *buf = (uint8_t*)zng_alloc(sizeof(uint8_t) * (128 + param));
+        if (buf != NULL) {
+            (void)crc32(0, buf + param, 128);
+        }
+        zng_free(buf);
+    }
+};
```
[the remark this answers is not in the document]
phprus: ~~We need to use single buffer for all offsets to ensure that all addresses are checked.~~

``zng_alloc`` guarantees alignment. So everything will work. I made a mistake.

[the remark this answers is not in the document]
KungFuJesus: We don't actually care what the memory contents are, we are simply testing varying alignments against the alignment sanitizer, here.  Additionally, if an allocation somehow fails we have bigger problems and it will crash, causing a failure.

KungFuJesus:
Ok, I think this is now good to go.

test/test_crc32.cc
```diff
@@ -195,6 +197,20 @@ class crc32_variant : public ::testing::TestWithParam<crc32_test> {
...
```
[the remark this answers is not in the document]
KungFuJesus: Our alignment offset won't be negative because we control the test parameters and it would never make sense to do so, as all non zero modulus values of 8 are tested.

PR #1837, opened 2024-12-24
Chorba

samrussell:
Implements the algorithms from https://arxiv.org/abs/2412.16398 for AVX1 and non-accelerated.

Should give a 2x speedup over braiding for data over 256KB and a smaller speedup for >256B

AVX1 should get a 5-20% speedup depending on data length. Same principle applies to AVX2/AVX512 with diminishing returns, haven't tested NEON yet. possibly this can also be applied to CRC32 opcodes on ARM etc

## Summary by CodeRabbit

- **New Features**
	- Enhanced CRC-32 computation with optimized logic for handling various input sizes.
	- Introduced methods for efficient CRC calculation tailored for larger datasets.
	- Improved performance of CRC32 folding computation through chunked processing.
	- Added new implementations of the Chorba algorithm for CRC32 calculations.
	- New source files and object files added to enhance CRC32 functionality.
	- New benchmark registration for the generic CRC32 implementation.
	- New test case added for the generic CRC32 implementation.
	- Introduced a new internal function for streamlined CRC calculations.
	- Renamed and reorganized CRC-related functions for clarity.

- **Bug Fixes**
	- Ensured proper post-conditioning of CRC values before returning results.

### DIFF
--- a/arch/generic/crc32_braid_c.c
+++ b/arch/generic/crc32_braid_c.c
@@ -59,11 +59,7 @@ static z_word_t crc_word(z_word_t data) {
 #endif /* W */
 
 /* ========================================================================= */
-Z_INTERNAL uint32_t PREFIX(crc32_braid)(uint32_t crc, const uint8_t *buf, size_t len) {
-    uint32_t c;
-
-    /* Pre-condition the CRC */
-    c = (~crc) & 0xffffffff;
+Z_INTERNAL uint32_t crc32_braid_internal(uint32_t c, const uint8_t *buf, size_t len) {
 
 #ifdef W
     /* If provided enough bytes, do a braided CRC calculation. */
@@ -210,6 +206,15 @@ Z_INTERNAL uint32_t PREFIX(crc32_braid)(uint32_t crc, const uint8_t *buf, size_t
         len--;
         DO1;
     }
+
+    /* Return the CRC, post-conditioned. */
+    return c;
+}
+
+uint32_t PREFIX(crc32_braid)(uint32_t c, const uint8_t *buf, size_t len) {
+    c = (~c) & 0xffffffff;
+
+    c = crc32_braid_internal(c, buf, len);
 
     /* Return the CRC, post-conditioned. */
     return c ^ 0xffffffff;
--- a/arch/generic/crc32_c.c
+++ b/arch/generic/crc32_c.c
@@ -1,0 +1,49 @@
+#include "zbuild.h"
+#include "crc32_braid_p.h"
+#include "crc32_c.h"
+
+/* Size thresholds for Chorba algorithm variants */
+#define CHORBA_LARGE_THRESHOLD (sizeof(z_word_t) * 64 * 1024)
+#define CHORBA_MEDIUM_UPPER_THRESHOLD 32768
+#define CHORBA_MEDIUM_LOWER_THRESHOLD 8192
+#define CHORBA_SMALL_THRESHOLD_64BIT 72
+#define CHORBA_SMALL_THRESHOLD_32BIT 80
+
+uint32_t PREFIX(crc32_c)(uint32_t crc, const uint8_t *buf, size_t len) {
+    uint32_t c;
+    uint64_t* aligned_buf;
+    size_t aligned_len;
+
+    c = (~crc) & 0xffffffff;
+#ifndef WITHOUT_CHORBA
+    unsigned long algn_diff = ((uintptr_t)8 - ((uintptr_t)buf & 0xF)) & 0xF;
+    if (algn_diff < len) {
+        if (algn_diff) {
+            c = crc32_braid_internal(c, buf, algn_diff);
+        }
+        aligned_buf = (uint64_t*) (buf + algn_diff);
+        aligned_len = len - algn_diff;
+        if(aligned_len > CHORBA_LARGE_THRESHOLD)
+            c = crc32_chorba_118960_nondestructive(c, (z_word_t*) aligned_buf, aligned_len);
+#  if W == 8
+        else if (aligned_len > CHORBA_MEDIUM_LOWER_THRESHOLD && aligned_len <= CHORBA_MEDIUM_UPPER_THRESHOLD)
+            c = crc32_chorba_32768_nondestructive(c, (uint64_t*) aligned_buf, aligned_len);
+        else if (aligned_len > CHORBA_SMALL_THRESHOLD_64BIT)
+            c = crc32_chorba_small_nondestructive(c, (uint64_t*) aligned_buf, aligned_len);
+#  else
+        else if (aligned_len > CHORBA_SMALL_THRESHOLD_32BIT)
+            c = crc32_chorba_small_nondestructive_32bit(c, (uint32_t*) aligned_buf, aligned_len);
+#  endif
+        else
+            c = crc32_braid_internal(c, (uint8_t*) aligned_buf, aligned_len);
+    }
+    else {
+        c = crc32_braid_internal(c, buf, len);
+    }
+#else
+    c = crc32_braid_internal(c, buf, len);
+#endif /* WITHOUT_CHORBA */
+
+    /* Return the CRC, post-conditioned. */
+    return c ^ 0xffffffff;
+}
--- a/arch/x86/crc32_fold_pclmulqdq_tpl.h
+++ b/arch/x86/crc32_fold_pclmulqdq_tpl.h
@@ -105,6 +105,226 @@ Z_INTERNAL void CRC32_FOLD(crc32_fold *crc, const uint8_t *src, size_t len, uint
     }
 #endif
 
+#ifndef WITHOUT_CHORBA
+        /* Implement Chorba algorithm from https://arxiv.org/abs/2412.16398
+         * We interleave the PCLMUL-base folds with 8x scaled generator
+         * polynomial copies; we read 8x QWORDS and then XOR them into
+         * the stream at the following offsets: 6, 9, 10, 16, 20, 22,
+         * 24, 25, 27, 28, 30, 31, 32 - this is detailed in the paper
+         * as "generator_64_bits_unrolled_8" */
+        while (len >= 512 + 64 + 16*8) {
+            __m128i chorba8 = _mm_loadu_si128((__m128i *)src);
+            __m128i chorba7 = _mm_loadu_si128((__m128i *)src + 1);
+            __m128i chorba6 = _mm_loadu_si128((__m128i *)src + 2);
+            __m128i chorba5 = _mm_loadu_si128((__m128i *)src + 3);
+            __m128i chorba4 = _mm_loadu_si128((__m128i *)src + 4);
+            __m128i chorba3 = _mm_loadu_si128((__m128i *)src + 5);
+            __m128i chorba2 = _mm_loadu_si128((__m128i *)src + 6);
+            __m128i chorba1 = _mm_loadu_si128((__m128i *)src + 7);
+#ifdef COPY
+            _mm_storeu_si128((__m128i *)dst, chorba8);
+            _mm_storeu_si128((__m128i *)dst + 1, chorba7);
+            _mm_storeu_si128((__m128i *)dst + 2, chorba6);
+            _mm_storeu_si128((__m128i *)dst + 3, chorba5);
+            _mm_storeu_si128((__m128i *)dst + 4, chorba4);
+            _mm_storeu_si128((__m128i *)dst + 5, chorba3);
+            _mm_storeu_si128((__m128i *)dst + 6, chorba2);
+            _mm_storeu_si128((__m128i *)dst + 7, chorba1);
+            dst += 16*8;
+#else
+            XOR_INITIAL128(chorba8);
+#endif
+            chorba2 = _mm_xor_si128(chorba2, chorba8);
+            chorba1 = _mm_xor_si128(chorba1, chorba7);
+            src += 16*8;
+            len -= 16*8;
+
+            xmm_t0 = _mm_loadu_si128((__m128i *)src);
+            xmm_t1 = _mm_loadu_si128((__m128i *)src + 1);
+            xmm_t2 = _mm_loadu_si128((__m128i *)src + 2);
+            xmm_t3 = _mm_loadu_si128((__m128i *)src + 3);
+
+            fold_12(&xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3);
+#ifdef COPY
+            _mm_storeu_si128((__m128i *)dst, xmm_t0);
+            _mm_storeu_si128((__m128i *)dst + 1, xmm_t1);
+            _mm_storeu_si128((__m128i *)dst + 2, xmm_t2);
+            _mm_storeu_si128((__m128i *)dst + 3, xmm_t3);
+            dst += 64;
+#endif
+            xmm_t0 = _mm_xor_si128(xmm_t0, chorba6);
+            xmm_t1 = _mm_xor_si128(_mm_xor_si128(xmm_t1, chorba5), chorba8);
+            xmm_t2 = _mm_xor_si128(_mm_xor_si128(_mm_xor_si128(xmm_t2, chorba4), chorba8), chorba7);
+            xmm_t3 = _mm_xor_si128(_mm_xor_si128(_mm_xor_si128(xmm_t3, chorba3), chorba7), chorba6);
+            xmm_crc0 = _mm_xor_si128(xmm_t0, xmm_crc0);
+            xmm_crc1 = _mm_xor_si128(xmm_t1, xmm_crc1);
+            xmm_crc2 = _mm_xor_si128(xmm_t2, xmm_crc2);
+            xmm_crc3 = _mm_xor_si128(xmm_t3, xmm_crc3);
+
+            xmm_t0 = _mm_loadu_si128((__m128i *)src + 4);
+            xmm_t1 = _mm_loadu_si128((__m128i *)src + 5);
+            xmm_t2 = _mm_loadu_si128((__m128i *)src + 6);
+            xmm_t3 = _mm_loadu_si128((__m128i *)src + 7);
+
+            fold_4(&xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3);
+#ifdef COPY
+            _mm_storeu_si128((__m128i *)dst, xmm_t0);
+            _mm_storeu_si128((__m128i *)dst + 1, xmm_t1);
+            _mm_storeu_si128((__m128i *)dst + 2, xmm_t2);
+            _mm_storeu_si128((__m128i *)dst + 3, xmm_t3);
+            dst += 64;
+#endif
+
+            xmm_t0 = _mm_xor_si128(_mm_xor_si128(_mm_xor_si128(xmm_t0, chorba2), chorba6), chorba5);
+            xmm_t1 = _mm_xor_si128(_mm_xor_si128(_mm_xor_si128(xmm_t1, chorba1), chorba4), chorba5);
+            xmm_t2 = _mm_xor_si128(_mm_xor_si128(xmm_t2, chorba3), chorba4);
+            xmm_t3 = _mm_xor_si128(_mm_xor_si128(xmm_t3, chorba2), chorba3);
+            xmm_crc0 = _mm_xor_si128(xmm_t0, xmm_crc0);
+            xmm_crc1 = _mm_xor_si128(xmm_t1, xmm_crc1);
+            xmm_crc2 = _mm_xor_si128(xmm_t2, xmm_crc2);
+            xmm_crc3 = _mm_xor_si128(xmm_t3, xmm_crc3);
+
+            xmm_t0 = _mm_loadu_si128((__m128i *)src + 8);
+            xmm_t1 = _mm_loadu_si128((__m128i *)src + 9);
+            xmm_t2 = _mm_loadu_si128((__m128i *)src + 10);
+            xmm_t3 = _mm_loadu_si128((__m128i *)src + 11);
+
+            fold_4(&xmm_crc0, &xmm_crc1, &xmm_crc2, &xmm_crc3);
+#ifdef COPY
+            _mm_storeu_si128((__m128i *)dst, xmm_t0);
+            _mm_storeu_si128((__m128i *)dst + 1, xmm_t1);
+            _mm_storeu_si128((__m128i *)dst + 2, xmm_t2);
+            _mm_storeu_si128((__m128i *)dst + 3, xmm_t3);
+            dst += 64;
+#endif
+
+            xmm_t0 = _mm_xor_si128(_mm_xor_si128(_mm_xor_si128(xmm_t0, chorba1), chorba2), chorba8);
+            xmm_t1 = _mm_xor_si128(_mm_xor_si128(xmm_t1, chorba1), chorba7);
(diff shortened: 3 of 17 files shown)


nmoinvaz:
Can't say I understood all of it, but your PDF looks very well written.

arch/x86/crc32_fold_pclmulqdq_tpl.h
```diff
@@ -105,6 +105,205 @@ Z_INTERNAL void CRC32_FOLD(crc32_fold *crc, const uint8_t *src, size_t len, uint
...
+            xmm_crc2 = _mm_xor_si128(xmm_t2, fold_high3);
+            xmm_crc2 = _mm_xor_si128(xmm_crc2, fold_low3);
+            fold_high4 = _mm_clmulepi64_si128(xmm_crc3, shift544_shift480, 0x11);
+            fold_low4 = _mm_clmulepi64_si128(xmm_crc3, shift544_shift480, 0x00);
+            xmm_crc3 = _mm_xor_si128(xmm_t3, fold_high4);
+            xmm_crc3 = _mm_xor_si128(xmm_crc3, fold_low4);
+
+            len -= 512;
+            src += 512;
+        }
```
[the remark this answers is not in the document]
nmoinvaz: This may not be necessary.

arch/x86/crc32_fold_pclmulqdq_tpl.h
```diff
@@ -105,6 +105,205 @@ Z_INTERNAL void CRC32_FOLD(crc32_fold *crc, const uint8_t *src, size_t len, uint
...
+            __m128i bonus4 = _mm_loadu_si128(src + (16 * 8));
+            __m128i bonus3 = _mm_loadu_si128(src + (16 * 9));
+            __m128i bonus2 = _mm_loadu_si128(src + (16 * 10)) ^ bonus8;
+            __m128i bonus1 = _mm_loadu_si128(src + (16 * 11)) ^ bonus7;
+            src += 16*8;
+            len -= 16*8;
+            xmm_t0 = _mm_loadu_si128(src + (16 * 4)) ^ bonus6;
+            xmm_t1 = _mm_loadu_si128(src + (16 * 5)) ^ bonus5 ^ bonus8;
+            xmm_t2 = _mm_loadu_si128(src + (16 * 6)) ^ bonus4 ^ bonus8 ^ bonus7;
+            xmm_t3 = _mm_loadu_si128(src + (16 * 7)) ^ bonus3 ^ bonus7 ^ bonus6;
```
KungFuJesus: There's a high probability those XORs won't compile for MSVC, I think only GCC and clang support binary operations on SIMD operands.  With AVX512 there's also an opportunity here to use ternary logic here, though we'll need to pollute things with an ifdef macro like we do in some other places.

arch/generic/crc32_braid_c.c
```diff
@@ -211,6 +207,539 @@ Z_INTERNAL uint32_t PREFIX(crc32_braid)(uint32_t crc, const uint8_t *buf, size_t
...
 
+    /* Return the CRC, post-conditioned. */
+    return c;
+}
+
+#define bitbuffersizebytes (128 * 1024)
+#define bitbuffersizeqwords (bitbuffersizebytes / sizeof(uint64_t))
+
+uint32_t chorba_118960_nondestructive (uint32_t crc, const uint8_t* input, size_t len) {
+    uint64_t* bitbuffer = calloc(bitbuffersizeqwords, sizeof(uint64_t));
```
KungFuJesus: I think we probably want zcalloc here (mostly for alignment).  It may also make sense to time this using stack backed memory, instead, since it's just 128k and fixed at compile time.

arch/generic/crc32_braid_c.c
```diff
@@ -211,6 +207,539 @@ Z_INTERNAL uint32_t PREFIX(crc32_braid)(uint32_t crc, const uint8_t *buf, size_t
...
+        bitbuffer[outoffset2 + 16] = in27;
+        bitbuffer[outoffset2 + 17] = in28;
+        bitbuffer[outoffset2 + 18] = in29;
+        bitbuffer[outoffset2 + 19] = in30;
+        bitbuffer[outoffset2 + 20] = in31;
+        bitbuffer[outoffset2 + 21] = in32;
+    }
+
+    for(; i + 118960 + 512 < len; i += 256) {
+        uint64_t in1, in2, in3, in4, in5, in6, in7, in8;
```
KungFuJesus: This is some pretty heavy unrolling. Is there anyway to maybe roll the loop body up and get the same performance?  I'm just thinking this is going to be generating a lot of program code and it doesn't seem like it needs this much to fill most pipelines.  Do you have slightly less unrolled variants we could test?

arch/x86/crc32_fold_pclmulqdq_tpl.h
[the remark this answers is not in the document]
samrussell: I'm okay with abstracting out the pclmuls, it's just the loads where we XOR in Chorba pattern

arch/x86/crc32_fold_pclmulqdq_tpl.h
samrussell: That's good to know, I did the binary operations to make the code clearer but happy to rewrite with intrinsics

arch/generic/crc32_braid_c.c
samrussell: Yes for alignment, also the calloc/zalloc is not needed because the first 2 passes write back directly (instead of XOR) so we don't need memory to be empty. Happy to do this with stack too if you're okay with it.

arch/generic/crc32_braid_c.c
samrussell: The issue is less about filling pipelines and more about maximising register usage and reducing memory writes. What the algorithm does is it copies each QWORD 7, 11 and 22 places forward (56, 88 and 176 bytes respectively), so we need to keep 22 local regs to get the benefit of this without either dumping into a buffer or writing back onto the original stream. Unrolling 4-8 times gives an additional speed boost, and then since we're just doing reads/xors/stores it actually takes something like a 50% performance hit when the ring buffer isn't a power of 2 (DIV vs AND), and if the ring buffer is a power of 2 then we need the loop counter to also be a power of 2. On x64 we get register spills but on ARM we get more benefit out of it.

This is not my field of expertise so I am completely open to alternatives, it's possible that writing this with an array or some other approach gets the same performance.

As for the 3 variants of the loop, the `memset()` ends up being quite a large performance hit if we're working in the 256KiB-1MiB range, so this means we still get performance here (and it is proportionally tiny once we hit >10MiB).

arch/x86/crc32_fold_pclmulqdq_tpl.h
samrussell: New commit replaces xors with `_mm_xor_si128()` intrinsic

arch/generic/crc32_braid_c.c
```diff
@@ -211,6 +207,539 @@ Z_INTERNAL uint32_t PREFIX(crc32_braid)(uint32_t crc, const uint8_t *buf, size_t
...
+    uint32_t c;
+
+    c = (~crc) & 0xffffffff;
+    if(len > 512 * 1024)
+        c = chorba_118960_nondestructive(c, buf, len);
+    else if (len > 72)
+        c = chorba_small_nondestructive(c, buf, len);
+    else {
+        c = crc32_braid(c, buf, len);
+    }
```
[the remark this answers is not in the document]
samrussell: note to self, should probably factor this into its own file and rename

arch/generic/crc32_braid_c.c
```diff
@@ -211,6 +207,539 @@
...
+    *((uint64_t*) (final + (3*8))) ^= next4;
+    *((uint64_t*) (final + (4*8))) ^= next5;
+
+    for(int j = 0; j<len-i; j++) {
+        crc = crc_table[(crc ^ final[j] ^ bitbufferbytes[(j+i) % bitbuffersizebytes]) & 0xff] ^ (crc >> 8);
+    }
+
+    free(bitbuffer);
+    return crc;
+}
```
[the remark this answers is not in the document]
samrussell: note to self: this needs a bswap for big endian (only when we're doing shifts, the ones where we just read/write whole QWORDS is fine)

arch/generic/crc32_braid_c.c
```diff
@@ -211,6 +207,538 @@ Z_INTERNAL uint32_t PREFIX(crc32_braid)(uint32_t crc, const uint8_t *buf, size_t
...
+        uint64_t d1, d2, d3, d4;
+
+        uint64_t out1;
+        uint64_t out2;
+        uint64_t out3;
+        uint64_t out4;
+        uint64_t out5;
+
+        in1 = *((uint64_t*) (input + i)) ^ next1 ^ bitbuffer[(i/8) % bitbuffersizeqwords];
+        in2 = *((uint64_t*) (input + i + (8*1))) ^ next2 ^ bitbuffer[(i/8 + 1) % bitbuffersizeqwords];
```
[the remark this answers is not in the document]
samrussell: not necessary because we are taking the modulus over the buffer size

arch/x86/crc32_fold_pclmulqdq_tpl.h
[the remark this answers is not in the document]
samrussell: I moved the chorba xors down to the #ifdef COPY section so nearly all of the fold blocks are identical and can be abstracted no problem, will do this later

nmoinvaz:
CodeRabbits nitpicks look worthy of inclusion.

arch/generic/crc32_braid_c.c
KungFuJesus: If the scope of your variables are that long lived then rolling the loop up a bit more probably doesn't help that situation. I was thinking if this loop were rolled up a bit more the lifetimes of these variables would be short lived enough that they could simply be clobbered. Register moves on nearly every modern architecture go through move elimination and end up being almost free.

KungFuJesus: If you could reduce things to something like 16 live variables to reduce stack spills on x86.that might be beneficial but your stashing values into a stack or heap backed buffer anyhow so you're ultimately going to be limited by that.

I'll be curious to try this on pre nehalem hardware that lack the carryless multiplies.

arch/generic/crc32_braid_c.c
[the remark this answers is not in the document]
samrussell: ok this is actually a thing, i need to double check my input reading to make sure it's safe

arch/generic/crc32_braid_c.c
samrussell: yeah the stack spill is unavoidable unless someone has a clever trick, this is the densest polynomial i could find (so 22 is the minimum and then 32 total to avoid the DIV penalty)

arch/x86/crc32_fold_pclmulqdq_tpl.h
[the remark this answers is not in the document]
samrussell: reusing `fold_4()` and added `fold_12()` to help tidy it up

arch/generic/crc32_braid_c.c
```diff
@@ -211,6 +207,596 @@ Z_INTERNAL uint32_t PREFIX(crc32_braid)(uint32_t crc, const uint8_t *buf, size_t
...
+        in1 = ZSWAP64(in1);
+        in2 = ZSWAP64(in2);
+#endif
+        in1 ^= next1;
+        in2 ^= next2;
+
+        a1 = (in1 << 17) ^ (in1 << 55);
+        a2 = (in1 >> 47) ^ (in1 >> 9) ^ (in1 << 19);
+        a3 = (in1 >> 45) ^ (in1 << 44);
+        a4 = (in1 >> 20);
```
ccawley2011: I haven't run any benchmarks yet, but shifts of 64-bit integers tend to be more complex on 32-bit systems compared to other operations. If it's not possible to do this trivially with 32-bit integers (where necessary), would it be possible for specific architectures to always use the braided version instead?

samrussell:
That's a good point. I haven't had a chance to bench this on 32-bit either,
but there are a few options to avoid the 64-bit shifts. Register space is a
lot tighter on x86 too, only 6 free regs vs 14 for x64.

arch/generic/crc32_braid_c.c
```diff
@@ -211,6 +207,1063 @@ Z_INTERNAL uint32_t PREFIX(crc32_braid)(uint32_t crc, const uint8_t *buf, size_t
...
     }
 
+    /* Return the CRC, post-conditioned. */
+    return c;
+}
+
+/* Implement Chorba algorithm from https://arxiv.org/abs/2412.16398 */
+#define bitbuffersizebytes (16 * 1024 * sizeof(z_word_t))
+#define bitbuffersizezwords (bitbuffersizebytes / sizeof(z_word_t))
+#define bitbuffersizeqwords (bitbuffersizebytes / sizeof(uint64_t))
```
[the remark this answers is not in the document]
samrussell: business requirement to use stack instead of heap alloc

arch/generic/crc32_braid_c.c
```diff
@@ -211,6 +207,1063 @@ Z_INTERNAL uint32_t PREFIX(crc32_braid)(uint32_t crc, const uint8_t *buf, size_t
         DO1;
     }
 
+    /* Return the CRC, post-conditioned. */
+    return c;
+}
+
+/* Implement Chorba algorithm from https://arxiv.org/abs/2412.16398 */
```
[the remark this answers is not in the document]
samrussell: I would like to see a draft documentation. The paper is at https://arxiv.org/pdf/2412.16398

arch/generic/crc32_braid_c.c
```diff
@@ -211,6 +207,1063 @@ Z_INTERNAL uint32_t PREFIX(crc32_braid)(uint32_t crc, const uint8_t *buf, size_t
...
+        a2 = (in1 >> 47) ^ (in1 >> 9) ^ (in1 << 19);
+        a3 = (in1 >> 45) ^ (in1 << 44);
+        a4 = (in1 >> 20);
+        
+        b1 = (in2 << 17) ^ (in2 << 55);
+        b2 = (in2 >> 47) ^ (in2 >> 9) ^ (in2 << 19);
+        b3 = (in2 >> 45) ^ (in2 << 44);
+        b4 = (in2 >> 20);
+
+        in3 = input[(i + 16) / sizeof(z_word_t)] ^ bitbuffer[(i / sizeof(uint64_t) + 2) % bitbuffersizeqwords];
```
nmoinvaz: Does this get optimized to:
Const ix = i / sizeof(z_word_t)
input[ix + (16 / sizeof(z_word_t)]
? In other words is division only done once?

KungFuJesus: I imagine the compiler is doing that but even if it isn't it's a division by a power of 2 so it simply becomes a bit shift.

samrussell: input is a `uint64_t*` so the division disappears completely (mem lookups have to be a byte offset) and the +2 in the bitbuffer lookup actually ends up being a +16 const in the final asm, so no mult/division should happen here

the important part is the `% bitbuffersizeqwords`, if bitbuffersizeqwords isn't a power of 2 this actually takes up something like 30% of the cycles for the loop (DIV is much more expensive than an AND bitmask), so we're wasting a bit of memory here for a massive time saving. in the bigger function we actually precompute these at the start because the modulo starts to get expensive when we do it 24 times.

arch/generic/crc32_braid_c.c
samrussell: i've just added a 32-bit version using the same polynomial, im getting a 30% speedup on a i5-4570 in 32-bit mode, what difference are you seeing?

arch/generic/crc32_braid_c.c
```diff
@@ -211,6 +207,1279 @@ Z_INTERNAL uint32_t PREFIX(crc32_braid)(uint32_t crc, const uint8_t *buf, size_t
...
+        next5 = out5;
+        next6 = out6;
+        next7 = out7;
+        next8 = out8;
+        next9 = out9;
+        next10 = out10;
+
+    }
+#if BYTE_ORDER == BIG_ENDIAN
+    next1 = ZSWAP64(next1);
```
KungFuJesus: These should be ZSWAP32.

samrussell: thanks, fixed

arch/generic/crc32_braid_c.c
```diff
@@ -211,6 +207,1279 @@ Z_INTERNAL uint32_t PREFIX(crc32_braid)(uint32_t crc, const uint8_t *buf, size_t
...
+    final[7] ^= next8;
+    final[8] ^= next9;
+    final[9] ^= next10;
+
+    crc = crc32_braid_base(crc, (uint8_t*) final, len-i);
+
+    return crc;
+}
+
+Z_INTERNAL uint32_t PREFIX(crc32_braid)(uint32_t crc, const uint8_t *buf, size_t len) {
```
nmoinvaz: I propose:
* `crc32_braid` should be renamed `crc32_c` and moved into `arch\generic\crc32_c.c`
* `crc32_braid_base` should be renamed back to `crc32_braid`.
* `chorba_*` functions should be moved to `arch\generic\crc32_chorba_c.c`
  * Possibly all `chorba_*` functions prefixed with `crc32_`

@Dead2 thoughts?

samrussell:
makes perfect sense to me, this was just the shortest path to getting it
working

arch/generic/crc32_braid_c.c
samrussell: Done in 6ff14c317d187bef8e5b1539b795f0f3f55a80bd

test/benchmarks/benchmark_crc32.cc
```diff
@@ -56,7 +56,7 @@ class crc32: public benchmark::Fixture {
     } \
     BENCHMARK_REGISTER_F(crc32, name)->Arg(1)->Arg(8)->Arg(12)->Arg(16)->Arg(32)->Arg(64)->Arg(512)->Arg(4<<10)->Arg(32<<10)->Arg(256<<10)->Arg(4096<<10);
 
-BENCHMARK_CRC32(braid, PREFIX(crc32_braid), 1);
```
nmoinvaz, line 59: Do we want to benchmark crc32_braid along side crc32_c? It might be a good idea.

arch/generic/crc32_braid_c.c
```diff
@@ -211,6 +207,538 @@ Z_INTERNAL uint32_t PREFIX(crc32_braid)(uint32_t crc, const uint8_t *buf, size_t
         DO1;
     }
 
+    /* Return the CRC, post-conditioned. */
+    return c;
+}
+
+#define bitbuffersizebytes (128 * 1024)
+#define bitbuffersizeqwords (bitbuffersizebytes / sizeof(uint64_t))
+
```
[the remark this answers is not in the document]
samrussell: business requirement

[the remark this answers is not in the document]
samrussell: we cannot add memory allocations but we can use stack mem even though this is a large allocation

test/benchmarks/benchmark_crc32.cc
samrussell: done

arch/generic/crc32_c.c
```diff
@@ -0,0 +1,38 @@
...
+#include "crc32_braid_p.h"
+#include "crc32_c.h"
+
+uint32_t PREFIX(crc32_c)(uint32_t crc, const uint8_t *buf, size_t len) {
+    uint32_t c;
+    uint64_t* aligned_buf;
+    size_t aligned_len;
+
+    c = (~crc) & 0xffffffff;
+    unsigned long algn_diff = ((uintptr_t)8 - ((uintptr_t)buf & 0xF)) & 0xF;
```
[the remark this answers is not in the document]
nmoinvaz: Not really necessary imo

arch/generic/crc32_c.c
```diff
@@ -0,0 +1,38 @@
...
+            c = crc32_chorba_118960_nondestructive(c, (z_word_t*) aligned_buf, aligned_len);
+#if W == 8
+        else if (aligned_len > 8192 && aligned_len <= 32768)
+            c = crc32_chorba_32768_nondestructive(c, (uint64_t*) aligned_buf, aligned_len);
+        else if (aligned_len > 72)
+            c = crc32_chorba_small_nondestructive(c, (uint64_t*) aligned_buf, aligned_len);
+#else
+        else if (aligned_len > 80)
+            c = crc32_chorba_small_nondestructive_32bit(c, (uint32_t*) aligned_buf, aligned_len);
+#endif
```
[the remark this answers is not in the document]
Dead2: These functions are already declared in crc32_c.h

Dead2:
@samrussell Would you please rebase this? It should pass all CI tests then.

Also, the ai suggestion about changing magic numbers with defines would be very nice if you could work in as well.
```
+/* Size thresholds for Chorba algorithm variants */
+#define CHORBA_LARGE_THRESHOLD (sizeof(z_word_t) * 64 * 1024)
+#define CHORBA_MEDIUM_THRESHOLD 32768
+#define CHORBA_SMALL_THRESHOLD 72

 if(aligned_len > (sizeof(z_word_t) * 64) * 1024)
-    c = crc32_chorba_118960_nondestructive(c, (z_word_t*) aligned_buf, aligned_len);
+    c = crc32_chorba_118960_nondestructive(c, (z_word_t*) aligned_buf, aligned_len);
 #if W == 8
-else if (aligned_len > 8192 && aligned_len <= 32768)
+else if (aligned_len > 8192 && aligned_len <= CHORBA_MEDIUM_THRESHOLD)
     c = crc32_chorba_32768_nondestructive(c, (uint64_t*) aligned_buf, aligned_len);
-else if (aligned_len > 72)
+else if (aligned_len > CHORBA_SMALL_THRESHOLD)
     c = crc32_chorba_small_nondestructive(c, (uint64_t*) aligned_buf, aligned_len);
```

Other than that, I am ready to merge this next.

samrussell:
Rebased and magic numbers given names as per AI feedback

> @samrussell Would you please rebase this? It should pass all CI tests then.
> 
> Also, the ai suggestion about changing magic numbers with defines would be very nice if you could work in as well.
> 
> ```
> +/* Size thresholds for Chorba algorithm variants */
> +#define CHORBA_LARGE_THRESHOLD (sizeof(z_word_t) * 64 * 1024)
> +#define CHORBA_MEDIUM_THRESHOLD 32768
> +#define CHORBA_SMALL_THRESHOLD 72
> 
>  if(aligned_len > (sizeof(z_word_t) * 64) * 1024)
> -    c = crc32_chorba_118960_nondestructive(c, (z_word_t*) aligned_buf, aligned_len);
> +    c = crc32_chorba_118960_nondestructive(c, (z_word_t*) aligned_buf, aligned_len);
>  #if W == 8
> -else if (aligned_len > 8192 && aligned_len <= 32768)
> +else if (aligned_len > 8192 && aligned_len <= CHORBA_MEDIUM_THRESHOLD)
>      c = crc32_chorba_32768_nondestructive(c, (uint64_t*) aligned_buf, aligned_len);
> -else if (aligned_len > 72)
> +else if (aligned_len > CHORBA_SMALL_THRESHOLD)
>      c = crc32_chorba_small_nondestructive(c, (uint64_t*) aligned_buf, aligned_len);
> ```
> 
> Other than that, I am ready to merge this next.

Dead2 (approved):
LGTM

KungFuJesus:
Awesome work on this, can't wait to get around to vectorizing it again.

PR #1843, opened 2025-01-02
CI changes for Ubuntu 24

nmoinvaz:
## Summary by CodeRabbit

- **Chores**
	- Updated GitHub Actions workflows for multiple repositories.
	- Standardized package management for cross-compilation environments.
	- Upgraded Clang compiler versions from 11 to 15.
	- Simplified package dependencies by removing redundant `qemu` packages.
	- Updated cross-compilation package names for various architectures.
	- Renamed job configurations for clarity and consistency.
	- Updated LLVM package versions across various workflows.
	- Enhanced consistency in coverage reporting across multiple job configurations.
	- Added `gcov-exec` specifications for improved coverage reporting.
	- Modified variable types in macros to accommodate larger input data sizes.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -272,9 +272,10 @@ jobs:
             packages: qemu-user clang binutils-powerpc64-linux-gnu libgcc-11-dev-ppc64-cross libc-dev-ppc64-cross libstdc++-11-dev-ppc64-cross
 
           - name: Ubuntu GCC PPC64LE
-            os: ubuntu-latest
+            # qemu appears to be broken in newer versions of Ubuntu (see issue 1378)
+            os: ubuntu-20.04
             cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-powerpc64le.cmake
-            packages: qemu-user crossbuild-essential-ppc64el
+            packages: qemu qemu-user gcc-powerpc64le-linux-gnu g++-powerpc64le-linux-gnu libc-dev-ppc64el-cross
             gcov-exec: powerpc64le-linux-gnu-gcov
             codecov: ubuntu_gcc_ppc64le
 
@@ -298,6 +299,7 @@ jobs:
             packages: qemu-user crossbuild-essential-ppc64el
 
           - name: Ubuntu GCC SPARC64
+            # qemu appears to be broken in newer versions of Ubuntu (see issue 1378)
             os: ubuntu-20.04
             cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-sparc64.cmake
             packages: qemu qemu-user gcc-sparc64-linux-gnu g++-sparc64-linux-gnu libc-dev-sparc64-cross
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -194,31 +194,31 @@ jobs:
             codecov: ubuntu_gcc_armhf_compat_no_opt
 
           - name: Ubuntu GCC AARCH64 ASAN
-            os: ubuntu-latest
+            os: ubuntu-22.04
             cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-aarch64.cmake -DWITH_SANITIZER=Address
             asan-options: detect_leaks=0
-            packages: qemu-user crossbuild-essential-arm64
+            packages: qemu-user gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libc-dev-arm64-cross
             gcov-exec: aarch64-linux-gnu-gcov
             codecov: ubuntu_gcc_aarch64
 
           - name: Ubuntu GCC AARCH64 No ACLE UBSAN
-            os: ubuntu-latest
+            os: ubuntu-22.04
             cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-aarch64.cmake -DWITH_ACLE=OFF -DWITH_SANITIZER=Undefined
-            packages: qemu-user crossbuild-essential-arm64
+            packages: qemu-user gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libc-dev-arm64-cross
             gcov-exec: aarch64-linux-gnu-gcov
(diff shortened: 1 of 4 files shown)


nmoinvaz:
We are getting some illegal instructions on Ubuntu GCC PCC64LE:
```
qemu-ppc64le;-cpu;power8;-L;/usr/powerpc64le-linux-gnu/;/home/runner/work/zlib-ng/zlib-ng/minigzip;-k;-c;-2
18:   failed: Illegal instruction
```

Dead2:
> We are getting some illegal instructions on Ubuntu GCC PCC64LE:
> 
> ```
> qemu-ppc64le;-cpu;power8;-L;/usr/powerpc64le-linux-gnu/;/home/runner/work/zlib-ng/zlib-ng/minigzip;-k;-c;-2
> 18:   failed: Illegal instruction
> ```

If I understand the CI and toolchain differences correctly, the failure happens only with the Power8 + Little Endian + VSX combination.
GCC fails on Power8 LE with VSX.
GCC succeeds on Power8 LE without VSX.
GCC succeeds on Power8 BE with VSX.
Clang does not seem to be tested with Power8, only Power9.

So I am guessing something might be wrong with Power8 VSX code, somehow affected by LE endianness.
Maybe @KungFuJesus has some more insight.

nmoinvaz:
I commented out all the pc*/power* assigns in functable.h and errors still happen.

nmoinvaz:
We need to ignore gcov errors. Seems they were previously ignored anyways.
https://github.com/linux-test-project/lcov/issues/341#issuecomment-2548282709

nmoinvaz:
[Ubuntu GCC AARCH64](https://github.com/nmoinvaz/zlib-ng/actions/runs/12620816900/job/35166940379#logs) builds failing due to segmentation fault when running gtest_zlib. Perhaps that was mentioned somewhere else?

nmoinvaz:
I have a Raspberry Pi 5 with Ubuntu 24.10 and I was able to run gtest_zlib after compiling with both clang 19 and gcc 14. So perhaps it is an issue with qemu. I'm inclined to just force those tests to use Ubuntu 22.

Dead2:
> I have a Raspberry Pi 5 with Ubuntu 24.10 and I was able to run gtest_zlib after compiling with both clang 19 and gcc 14. So perhaps it is an issue with qemu. I'm inclined to just force those tests to use Ubuntu 22.

Yeah, how about we have those do Ubuntu 22 and just add an issue about it being broken, that way we might remember it and could recheck it at a later time, or someone else might take it upon themselves to investigate it further.

mtl1979:
> We are getting some illegal instructions on Ubuntu GCC PCC64LE:
> 
> ```
> qemu-ppc64le;-cpu;power8;-L;/usr/powerpc64le-linux-gnu/;/home/runner/work/zlib-ng/zlib-ng/minigzip;-k;-c;-2
> 18:   failed: Illegal instruction
> ```

Support for Power8 was dropped, you need older version of toolchain. This was mentioned earlier.

nmoinvaz:
Where was it mentioned?

nmoinvaz:
I found it.. https://github.com/zlib-ng/zlib-ng/issues/1378

nmoinvaz:
Is it just me or does qemu appear to be getting worse and worse?

nmoinvaz:
MSAN fails because zlib-ng repo isn’t updated to use Ubuntu 24 as ubuntu-latest yet. In my branch it works.

mtl1979:
> Is it just me or does qemu appear to be getting worse and worse?

It has been broken since around 6.0 and worse performance wise since they ditched the old instruction translator that was based on old GCC.

mtl1979:
> MSAN fails because zlib-ng repo isn’t updated to use Ubuntu 24 as ubuntu-latest yet. In my branch it works.

In most cases it is better to use explicit version instead of "latest" as during deploying of new Ubuntu release every 2 years, a lot of package versions change and we rely on specific matching package versions for certain package combinations.

nmoinvaz:
Rebased

Dead2:
Still some failed runs, did not spot anything obvious, other than the Power8 issue you discussed above.

nmoinvaz:
Looks like my commit for PPC64LE was overwritten on rebase.

nmoinvaz:
Not sure about the "Package check" ones.

inflate.c
```diff
@@ -387,7 +387,7 @@ static void updatewindow(PREFIX3(stream) *strm, const uint8_t *end, uint32_t len
     do { \
         if (have == 0) goto inf_leave; \
         have--; \
-        hold += ((unsigned)(*next++) << bits); \
+        hold += ((uint64_t)(*next++) << bits); \
```
phprus: Is this the correct fix?
The ``hold`` variable is of type ``uint32_t`` and may overflow when assigned with addition.

nmoinvaz: `hold` is `uint64_t`. The warning goes away, so I assume it is correct.

nmoinvaz: There are two separate warnings for two different lines. I think they are non-issues, and AI suggested something that looks rather complicated.

phprus: > `hold` is `uint64_t`. The warning goes away, so I assume it is correct.

https://github.com/zlib-ng/zlib-ng/blob/8ff036958987f2acbc92daaca32066ade20e63b0/inflate.c#L482

https://github.com/zlib-ng/zlib-ng/blob/8ff036958987f2acbc92daaca32066ade20e63b0/infback.c#L157

??

nmoinvaz: It looks like it is `uint64_t` in `inffast_tpl.h`.

phprus: Potentially dangerous:

``ZSWAP32(hold)``

on this line and below:

https://github.com/zlib-ng/zlib-ng/blob/e6fca08c04c82e03c802ca6107de9e70abd64a0a/inflate.c#L707

nmoinvaz: Hmm. I think I will just let somebody else fix this static analysis problem.

Dead2 (approved):
LGTM.
While this does not fix all the current failures, it makes it actually possible to see the rest of the failures instead of them drowning in a sea of other failures 👍

PR #1850, opened 2025-01-09
cmake: disable LTO for some configure checks

### LINKED ISSUE #1841: Some CMake configure checks may be broken with LTO
Some of zlib-ng's configure tests define a function expecting it to be compiled but don't call that function, or don't use its return value. This is risky with LTO where the whole thing may be optimised out.

Just skimming the diff between 2.1.7 and 2.2.3, I see the following:

```
# This doesn't use f, and f may be optimised out as a result
+    check_c_source_compiles(
+        "#if defined(_MSC_VER)
+        #include <intrin.h>
+        #else
+        #include <arm_acle.h>
+        #endif
+        unsigned int f(unsigned int a, unsigned int b) {
+            return __crc32w(a, b);
+        }
+        int main(void) { return 0; }"
+        HAVE_ACLE_FLAG
+    )
```

```
# This doesn't use test, and test may be optimised out as a result (see my CMake ref below, this is really similar)
+check_c_source_compiles(
+    "char *test(char *buffer) {
+        char *abuffer = __builtin_assume_aligned(buffer,64);
+        return abuffer;
+    }
+    int main() {
+        return 0;
+    }"
+    HAVE_BUILTIN_ASSUME_ALIGNED)
+if(HAVE_BUILTIN_ASSUME_ALIGNED)
+    add_definitions(-DHAVE_BUILTIN_ASSUME_ALIGNED)
+endif()
```

It may work to just disable LTO for these (or force CMake to only compile them, not link).

This also came up in https://github.com/zlib-ng/zlib-ng/issues/1616 (https://github.com/zlib-ng/zlib-ng/pull/1622).

See also https://gitlab.kitware.com/cmake/cmake/-/issues/26103.

thesamesam:
Some of zlib-ng's configure tests define a function expecting it to be compiled but don't call that function, or don't use its return value. This is risky with LTO where the whole thing may be optimised out, which has happened before:
* https://github.com/zlib-ng/zlib-ng/issues/1616
* https://github.com/zlib-ng/zlib-ng/pull/1622
* https://gitlab.kitware.com/cmake/cmake/-/issues/26103

Closes: https://github.com/zlib-ng/zlib-ng/issues/1841

## Summary by CodeRabbit

- **Chores**
	- Updated CMake configuration flags to improve compiler compatibility
	- Modified compiler flag settings for Link Time Optimization (LTO) checks
	- Adjusted compilation settings to ensure consistent function detection

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -301,7 +301,7 @@ if(NOT WITH_RUNTIME_CPU_DETECTION)
         message(STATUS "WARNING: Microsoft Visual Studio does not support compile time detection of CPU features for \"/arch\" before \"AVX\"")
         # Workaround for MSVC. By default MSVC does not define the __SSE*__ macros.
         # Fix it if AVX is enabled.
-        set(CMAKE_REQUIRED_FLAGS "${NATIVEFLAG}")
+        set(CMAKE_REQUIRED_FLAGS "${NATIVEFLAG} ${ZNOLTOFLAG}")
         check_c_source_compiles(
             "#ifndef __AVX__
             #  error \"AVX is not enabled.\"
@@ -491,6 +491,7 @@ endif()
 #
 # Check for __builtin_assume_aligned(x,n) support in the compiler
 #
+set(CMAKE_REQUIRED_FLAGS ${ZNOLTOFLAG})
 check_c_source_compiles(
     "char *test(char *buffer) {
         char *abuffer = __builtin_assume_aligned(buffer,64);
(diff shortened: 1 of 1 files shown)


thesamesam:
I diffed the generated CMake cache files on amd64 + x86 and saw no change as expected (given this is a defensive change).

nmoinvaz (approved):

CMakeLists.txt
```diff
@@ -301,7 +301,7 @@ if(NOT WITH_RUNTIME_CPU_DETECTION)
         message(STATUS "WARNING: Microsoft Visual Studio does not support compile time detection of CPU features for \"/arch\" before \"AVX\"")
         # Workaround for MSVC. By default MSVC does not define the __SSE*__ macros.
         # Fix it if AVX is enabled.
-        set(CMAKE_REQUIRED_FLAGS "${NATIVEFLAG}")
+        set(CMAKE_REQUIRED_FLAGS "${NATIVEFLAG} ${ZNOLTOFLAG}")
```
Dead2, line 304: **question (if-minor):** Does this actually make any difference? AFAIK there is no support for LTO in MSVC, so the variable would be empty.

thesamesam: I wondered the same -- I figured I'd be consistent given someone might copy it in future, or MSVC might gain it at some point.

I googled a bit and found https://stackoverflow.com/questions/42230141/how-to-write-lto-enabled-code which implies it has it under the name "LTCG", but honestly, I'm no MSVC expert at all. Happy to drop it if you want.

Dead2 (approved):

PR #1851, opened 2025-01-13
Remove unused HAVE_CHUNKUNROLL define check

pps83:
## Summary by CodeRabbit

- **Refactor**
	- Removed a specific chunk memory operation function, simplifying the memory handling approach in the codebase.

### DIFF
--- a/chunkset_tpl.h
+++ b/chunkset_tpl.h
@@ -49,7 +49,6 @@ static inline uint8_t* CHUNKCOPY(uint8_t *out, uint8_t const *from, unsigned len
    This assumption holds because inflate_fast() starts every iteration with at
    least 258 bytes of output space available (258 being the maximum length
    output from a single token; see inflate_fast()'s assumptions below). */
-#ifndef HAVE_CHUNKUNROLL
 static inline uint8_t* CHUNKUNROLL(uint8_t *out, unsigned *dist, unsigned *len) {
     unsigned char const *from = out - *dist;
     chunk_t chunk;
@@ -62,7 +61,6 @@ static inline uint8_t* CHUNKUNROLL(uint8_t *out, unsigned *dist, unsigned *len) 
     }
     return out;
 }
-#endif
 
 #ifndef HAVE_CHUNK_MAG
 /* Loads a magazine to feed into memory of the pattern */


pps83:
as a side note, I'm not sure what's going on, but seems like entire git world is broken. I recall, if I simply searched in my tortoise git for `HAVE_CHUNKUNROLL` I saw all the diffs to track where it was added/removed to understand why this thing isn't there. For some reason, it didn't work, searching in git didn't give me any results. So, I simply went to github.com and it also gave ZERO results for this thing: https://github.com/search?q=repo%3Azlib-ng%2Fzlib-ng%20HAVE_CHUNKUNROLL&type=code 

No, commits that contained it, which is obviously complete bs... so, is git/github this broken silently, or there is something that I missed?? :)


Just to make sure, searching [zlib-ng for 0x020202F0L](https://github.com/search?q=repo%3Azlib-ng%2Fzlib-ng%200x020202F0L&type=code) gives no results (except this actual message), while one of the [latest commits](https://github.com/zlib-ng/zlib-ng/commit/cbb6ec1d74e8061efdf7251f8c2dae778bed14fd) did remove `0x020202F0L`

pps83:
> ## Summary by CodeRabbit
> * **Refactor**
>   
>   * Removed a specific chunk memory operation function, simplifying the memory handling approach in the codebase.

@codeRabbit, that's pretty wrong

pps83:
@KungFuJesus ping.
@Dead2 the code removes define that's not used/referenced anywhere.

KungFuJesus (commented):
I think the original intent of this macro was to allow for a custom architecture specific chunkunroll if it was advantageous to do so. While removing it here doesn't hurt anything I don't think it helps any, either.

Those macros exist for the other chunk functions and I do plan to make use of them for a vmx method (at least for the chunked copy) if I can get a performant one out.

pps83:
Ok, should I close the PR?

My understanding was that the code was added, but after some changes these ifdefs became stale.
In any case, it's always better to remove unused code, will be easy to add when it's needed.

For some reason, searching github now shows no results when the line was added, which makes it difficult to track when/why was it added.

Dead2:
@KungFuJesus That is what I was thinking, thank you.

@pps83 This file is a template file, so IMO it is kind of expected for it to be like that. Even if it is not currently used, it demonstrates the possibilities for architectures to implement their own optimized version of the specific functions.

It has no practical effect currently, but does serve the small function of self-documenting and facilitating the intention. We have several cases of optimizations arriving as a PR without the author consulting us about the possibilities first, so keeping it will make it easier for them to figure out what they can/can't do there.

I suggest we keep this as is for now.

Issue #1840, opened 2025-01-01
Workflows incompatibility with ubuntu 24.04


phprus:
Github updates ubuntu-latest to ubuntu 24.04 (https://github.com/actions/runner-images/issues/10636).

This causes a lot of errors in CI:

https://github.com/phprus/zlib-ng/actions/runs/12568999002
https://github.com/phprus/zlib-ng/actions/runs/12568998935
https://github.com/phprus/zlib-ng/actions/runs/12568998934
https://github.com/phprus/zlib-ng/actions/runs/12568998931
https://github.com/phprus/zlib-ng/actions/runs/12568998928

Errors:
1) Old dependencies (e.g. clang-11). I suggest explicitly declaring the ubuntu version for such tests.
2) Problems due to ABI change in ubuntu 24.04 on ARM, MIPS, PPC: https://github.com/phprus/zlib-ng/actions/runs/12485475129/job/34844303420
3) Static analysis warnings/errors from the new version of clang: https://github.com/phprus/zlib-ng/actions/runs/12568998935/job/35037197233
4) Incompatibility in gcovr and new GCC (``Unrecognized GCOV output``).

nmoinvaz:
I'm taking a look at a few of these.

mtl1979:
Those ABI failures get fixed when we regenerate the cached files using Ubuntu 24.04... It takes a while to regenerate them as there can't be any pending pull requests that affect the CI workflows.

nmoinvaz:
This should be resolved now.

PR #1857, opened 2025-01-26
[abicheck] Regenerate ABI files for zlib

mtl1979:
* Generate using Ubuntu 24.04.1 LTS to fix mismatch in function signatures of gzseek() and gztell()

See #1840.


Dead2 (approved):
LGTM

PR #1859, opened 2025-01-26
Fix shift overflow in inflate and send_code.

mtl1979:
Size of `hold` was inconsistent between different inflate versions. Make all use 64-bit variable.
In `send_bits` there was no guard against shifting by variable bit width, which will overflow.

## Release Notes

- **Performance Improvements**
  - Enhanced bit buffer handling by expanding variable types to support larger data values.
  - Improved data processing capacity in compression and decompression functions.

- **Bug Fixes**
  - Added safety checks to prevent potential buffer overflow in bit manipulation operations.

- **Technical Updates**
  - Updated internal data type representations to increase range and precision of bit-level operations.
  - Reduced memory footprint in the `inflate_state` structure by modifying the `padding` array size.
  - Refined control flow for bit processing, ensuring proper handling of buffer states.

### DIFF
--- a/inflate.h
+++ b/inflate.h
@@ -121,7 +121,7 @@ struct ALIGNED_(64) inflate_state {
     uint32_t chunksize;         /* size of memory copying chunk */
 
         /* bit accumulator */
-    uint32_t hold;              /* input bit accumulator */
+    uint64_t hold;              /* input bit accumulator */
     unsigned bits;              /* number of bits in "in" */
         /* fixed and dynamic code tables */
     unsigned lenbits;           /* index bits for lencode */
@@ -141,7 +141,7 @@ struct ALIGNED_(64) inflate_state {
     code *next;                 /* next available space in codes[] */
 
 #if defined(_M_IX86) || defined(_M_ARM)
-    uint32_t padding[2];
+    uint32_t padding[1];
 #endif
     struct crc32_fold_s ALIGNED_(16) crc_fold;
 
--- a/trees_emit.h
+++ b/trees_emit.h
@@ -38,6 +38,10 @@ extern Z_INTERNAL const int base_dist[D_CODES];
 /* If not enough room in bi_buf, use (valid) bits from bi_buf and
  * (64 - bi_valid) bits from value, leaving (width - (64-bi_valid))
  * unused bits in value.
+ *
+ * NOTE: Static analyzers can't evaluate value of total_bits, so we
+ *       also need to make sure bi_valid is within acceptable range,
+ *       otherwise the shifts will overflow.
  */
 #define send_bits(s, t_val, t_len, bi_buf, bi_valid) {\
     uint64_t val = (uint64_t)t_val;\
@@ -45,10 +49,10 @@ extern Z_INTERNAL const int base_dist[D_CODES];
     uint32_t total_bits = bi_valid + len;\
     send_bits_trace(s, val, len);\
     sent_bits_add(s, len);\
-    if (total_bits < BIT_BUF_SIZE) {\
+    if (total_bits < BIT_BUF_SIZE && bi_valid < BIT_BUF_SIZE) {\
         bi_buf |= val << bi_valid;\
         bi_valid = total_bits;\
-    } else if (bi_valid == BIT_BUF_SIZE) {\
+    } else if (bi_valid >= BIT_BUF_SIZE) {\
         put_uint64(s, bi_buf);\
         bi_buf = val;\
         bi_valid = len;\
--- a/infback.c
+++ b/infback.c
@@ -102,7 +102,7 @@ int32_t Z_EXPORT PREFIX(inflateBackInit_)(PREFIX3(stream) *strm, int32_t windowB
     do { \
         PULL(); \
         have--; \
-        hold += ((unsigned)(*next++) << bits); \
+        hold += ((uint64_t)(*next++) << bits); \
         bits += 8; \
     } while (0)
 
@@ -154,7 +154,7 @@ int32_t Z_EXPORT PREFIX(inflateBack)(PREFIX3(stream) *strm, in_func in, void *in
     z_const unsigned char *next; /* next input */
     unsigned char *put;          /* next output */
     unsigned have, left;         /* available input and output */
-    uint32_t hold;               /* bit buffer */
+    uint64_t hold;               /* bit buffer */
     unsigned bits;               /* bits in bit buffer */
     unsigned copy;               /* number of stored or match bytes to copy */
     unsigned char *from;         /* where to copy match bytes from */
--- a/inflate.c
+++ b/inflate.c
@@ -291,7 +291,7 @@ int32_t Z_EXPORT PREFIX(inflatePrime)(PREFIX3(stream) *strm, int32_t bits, int32
     if (bits > 16 || state->bits + (unsigned int)bits > 32)
         return Z_STREAM_ERROR;
     value &= (1L << bits) - 1;
-    state->hold += (unsigned)value << state->bits;
+    state->hold += (uint64_t)value << state->bits;
     state->bits += (unsigned int)bits;
     return Z_OK;
 }
@@ -387,7 +387,7 @@ static void updatewindow(PREFIX3(stream) *strm, const uint8_t *end, uint32_t len
     do { \
         if (have == 0) goto inf_leave; \
         have--; \
-        hold += ((unsigned)(*next++) << bits); \
+        hold += ((uint64_t)(*next++) << bits); \
         bits += 8; \
     } while (0)
 
@@ -479,7 +479,7 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
     unsigned char *put;         /* next output */
     unsigned char *from;        /* where to copy match bytes from */
     unsigned have, left;        /* available input and output */
-    uint32_t hold;              /* bit buffer */
+    uint64_t hold;              /* bit buffer */
     unsigned bits;              /* bits in bit buffer */
     uint32_t in, out;           /* save starting available input and output */
     unsigned copy;              /* number of stored or match bytes to copy */
@@ -577,7 +577,7 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
         case TIME:
             NEEDBITS(32);
             if (state->head != NULL)
-                state->head->time = hold;
+                state->head->time = (unsigned)(hold);
             if ((state->flags & 0x0200) && (state->wrap & 4))
                 CRC4(state->check, hold);
             INITBITS();
@@ -704,7 +704,7 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
 #endif
         case DICTID:
             NEEDBITS(32);
-            strm->adler = state->check = ZSWAP32(hold);
+            strm->adler = state->check = ZSWAP32((unsigned)hold);
             INITBITS();
             state->mode = DICT;
             Z_FALLTHROUGH;
@@ -1128,7 +1128,7 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
 #ifdef GUNZIP
                      state->flags ? hold :
 #endif
-                     ZSWAP32(hold)) != state->check) {
+                     ZSWAP32((unsigned)hold)) != state->check) {
                     SET_BAD("incorrect data check");
                     break;
                 }


inflate.h
```diff
@@ -121,7 +121,7 @@ struct ALIGNED_(64) inflate_state {
     uint32_t chunksize;         /* size of memory copying chunk */
 
         /* bit accumulator */
-    uint32_t hold;              /* input bit accumulator */
+    uint64_t hold;              /* input bit accumulator */
```
phprus, line 124: Below is the struct member defined:
```c
uint32_t padding[2];
```

Maybe we should decrease it to 1?

mtl1979: Will test...

trees_emit.h
```diff
@@ -53,7 +53,9 @@ extern Z_INTERNAL const int base_dist[D_CODES];
         bi_buf = val;\
         bi_valid = len;\
     } else {\
-        bi_buf |= val << bi_valid;\
+        if (bi_valid < BIT_BUF_SIZE) {\
```
nmoinvaz: I think the `else` should be `else if (bi_valid < BIT_BUF_SIZE) {`

mtl1979: Fails on line 49 with that change...

nmoinvaz: Hmm. I read it wrong.

mtl1979: For some reason `bi_valid` ends up being more than `BIT_BUF_SIZE`, but I can't trace why or where it happens when using static analyzer...

trees_emit.h
```diff
@@ -53,9 +53,15 @@ extern Z_INTERNAL const int base_dist[D_CODES];
...
+            bi_buf |= val << bi_valid;\
+        }\
         put_uint64(s, bi_buf);\
-        bi_buf = val >> (BIT_BUF_SIZE - bi_valid);\
+        if (bi_valid < BIT_BUF_SIZE) {\
+            bi_buf = val >> (BIT_BUF_SIZE - bi_valid);\
+        } else {\
+            bi_buf = val;\
```
mtl1979: Here `bi_valid = len;` will cause static analyzer to trip on line 49 during next call... I suspect `t_len` might be either negative or very large positive number that results `total_bits` to wrap around to less than `BIT_BUF_SIZE` while `bi_valid` is 64 (or more).

We assume that `total_bits` is between 64 and 128 if we reach here.

trees_emit.h
```diff
@@ -42,6 +42,14 @@ extern Z_INTERNAL const int base_dist[D_CODES];
...
     uint64_t val = (uint64_t)t_val;\
     uint32_t len = (uint32_t)t_len;\
+    if (bi_valid > BIT_BUF_SIZE) {\
+        bi_buf = 0;\
+        bi_valid = 0;\
+    }\
+    if (len > BIT_BUF_SIZE) {\
+        val = 0;\
+        len = 0;\
+    }\
```
mtl1979: These should make sure `total_bits` is between 0 and 128.

trees_emit.h
phprus: If this is an impossible state, maybe make sense to add an ``assert``?

Dead2:
Not sure why Github thought I wanted to remove a reviewer, that was not my intention.

I do want to see reviews before I merge this though, next stable release is waiting for this currently.

nmoinvaz:
Should probably performance test the send_bits. To me it seems like a lot of extra if statements just to hide warnings.

phprus:
I don't see any problems in PR now.
Tomorrow I will try to analyze this PR in more detail.

phprus (approved):
I think we should wait for @nmoinvaz performance tests, but I don't think these are false positive warnings.

I think this PR needs to be merged.

mtl1979:
> Should probably performance test the send_bits. To me it seems like a lot of extra if statements just to hide warnings.

I'm not quite sure if compilers can optimize out some of the if statements especially when there is more than one macro invocation in series. The current version is in my opinion a lot more readable than the previous one.

As far as I managed to check, running without static analyzer never hits the added code and neither of the two length variables is ever zero with valid stream.

Dead2:
I had some gremlins causing benchmarks to be highly random (+-5% run to run) recently, finally figured it out, turns out I should not disable deeper C-states on the idling hyperthread while the other one is running the benchmark. I am thinking this change in behavior might have been caused by a kernel or cpu microcode update. 🤷‍♂️
(As some of you know, I force the benchmark to run on a specific core, disable turbo, force the core to run at a specific Mhz and use cpuset and a few kernel tricks to force other tasks off that core and its HT sibling.)

## Develop
```
 1     54.185%      0.078/0.082/0.083/0.001        0.027/0.030/0.031/0.001        8,526,745
 2     43.871%      0.144/0.149/0.151/0.001        0.027/0.030/0.031/0.001        6,903,702
 3     42.390%      0.178/0.182/0.183/0.001        0.025/0.028/0.030/0.001        6,670,664
 4     41.644%      0.206/0.212/0.214/0.002        0.024/0.028/0.030/0.001        6,553,205
 5     41.215%      0.235/0.238/0.241/0.002        0.024/0.028/0.029/0.001        6,485,659
 6     41.032%      0.277/0.285/0.290/0.004        0.023/0.027/0.029/0.002        6,456,912
 7     40.778%      0.381/0.388/0.392/0.003        0.024/0.028/0.029/0.001        6,416,941
 8     40.704%      0.492/0.501/0.506/0.004        0.022/0.027/0.030/0.002        6,405,249
 9     40.409%      0.588/0.595/0.599/0.003        0.024/0.027/0.028/0.001        6,358,951

 avg1  42.914%                        0.292                          0.028
 tot                                 78.931                          7.601       60,778,028

 130504    1344       8  131856   20310 libz-ng.so.2

compress_bench/compress_bench/1                 5347 ns         5338 ns       124032
compress_bench/compress_bench/8                 5631 ns         5621 ns       117456
compress_bench/compress_bench/16                5669 ns         5660 ns       120083
compress_bench/compress_bench/32                6035 ns         6025 ns       112324
compress_bench/compress_bench/64                6344 ns         6334 ns       106719
compress_bench/compress_bench/512               6809 ns         6798 ns       100918
compress_bench/compress_bench/4096             12010 ns        11988 ns        58512
compress_bench/compress_bench/32768            46215 ns        46130 ns        15176
uncompress_bench/uncompress_bench/1              150 ns          149 ns      4683675
uncompress_bench/uncompress_bench/64             310 ns          309 ns      2264460
uncompress_bench/uncompress_bench/1024           484 ns          483 ns      1449716
uncompress_bench/uncompress_bench/16384         4637 ns         4629 ns       150867
uncompress_bench/uncompress_bench/131072       26841 ns        26795 ns        26045
uncompress_bench/uncompress_bench/1048576     207574 ns       207153 ns         3372
```

## PR #1859 eca360a1ce671e29c2a808e915367ae5573fb8e7
```
 1     54.185%      0.085/0.090/0.092/0.002        0.027/0.030/0.031/0.001        8,526,745
 2     43.871%      0.150/0.154/0.156/0.002        0.025/0.030/0.032/0.001        6,903,702
 3     42.390%      0.180/0.185/0.188/0.002        0.026/0.029/0.031/0.001        6,670,664
 4     41.644%      0.209/0.216/0.219/0.003        0.024/0.029/0.031/0.002        6,553,205
 5     41.215%      0.237/0.242/0.245/0.002        0.026/0.029/0.031/0.001        6,485,659
 6     41.032%      0.282/0.291/0.294/0.003        0.026/0.028/0.029/0.001        6,456,912
 7     40.778%      0.378/0.388/0.394/0.004        0.024/0.029/0.031/0.002        6,416,941
 8     40.704%      0.492/0.501/0.508/0.004        0.026/0.029/0.030/0.001        6,405,249
 9     40.409%      0.585/0.594/0.603/0.005        0.024/0.028/0.029/0.001        6,358,951

 avg1  42.914%                        0.296                          0.029
 tot                                 79.869                          7.831       60,778,028

 132680    1344       8  134032   20b90 libz-ng.so.2

compress_bench/compress_bench/1                 5313 ns         5305 ns       122435
compress_bench/compress_bench/8                 5500 ns         5491 ns       119030
compress_bench/compress_bench/16                5526 ns         5517 ns       121097
compress_bench/compress_bench/32                6119 ns         6109 ns       115237
compress_bench/compress_bench/64                6483 ns         6473 ns       107138
compress_bench/compress_bench/512               7012 ns         7000 ns        98111
compress_bench/compress_bench/4096             12357 ns        12338 ns        56095
compress_bench/compress_bench/32768            47198 ns        47114 ns        14858
uncompress_bench/uncompress_bench/1              147 ns          147 ns      4759144
uncompress_bench/uncompress_bench/64             310 ns          309 ns      2263728
uncompress_bench/uncompress_bench/1024           485 ns          484 ns      1446700
uncompress_bench/uncompress_bench/16384         4650 ns         4644 ns       150566
uncompress_bench/uncompress_bench/131072       27031 ns        26985 ns        26080
uncompress_bench/uncompress_bench/1048576     207922 ns       207502 ns         3369
```

The PR adds 2176 bytes of code.
Compression is 1.2% slower on average.
* Level 1 9.76% slower
* Level 2 3.36% slower
* Level 6, 2.1% slower
* Levels 7-9 unchanged, lower levels up to 2.1% slower.
Decompression 3% slower on average.


## PR + added UNLIKELY() hints to all three if checks.
```
 1     54.185%      0.083/0.087/0.088/0.001        0.027/0.031/0.032/0.002        8,526,745
 2     43.871%      0.152/0.156/0.158/0.002        0.027/0.031/0.032/0.001        6,903,702
 3     42.390%      0.182/0.186/0.188/0.002        0.026/0.029/0.030/0.001        6,670,664
 4     41.644%      0.212/0.217/0.220/0.002        0.024/0.029/0.030/0.001        6,553,205
 5     41.215%      0.235/0.242/0.245/0.002        0.022/0.028/0.030/0.002        6,485,659
 6     41.032%      0.284/0.290/0.294/0.003        0.024/0.028/0.029/0.001        6,456,912
 7     40.778%      0.381/0.389/0.395/0.004        0.026/0.029/0.031/0.001        6,416,941
 8     40.704%      0.491/0.502/0.509/0.004        0.026/0.029/0.031/0.001        6,405,249
 9     40.409%      0.588/0.595/0.599/0.003        0.025/0.028/0.029/0.001        6,358,951

 avg1  42.914%                        0.296                          0.029
 tot                                 79.953                          7.808       60,778,028

 132360    1344       8  133712   20a50 libz-ng.so.2

compress_bench/compress_bench/1                 5294 ns         5286 ns       121578
compress_bench/compress_bench/8                 5575 ns         5566 ns       118902
compress_bench/compress_bench/16                5632 ns         5623 ns       119614
compress_bench/compress_bench/32                5921 ns         5912 ns       114086
compress_bench/compress_bench/64                6263 ns         6253 ns       108188
compress_bench/compress_bench/512               6775 ns         6765 ns       100570
compress_bench/compress_bench/4096             11951 ns        11931 ns        58166
compress_bench/compress_bench/32768            46690 ns        46609 ns        15017
uncompress_bench/uncompress_bench/1              150 ns          150 ns      4669256
uncompress_bench/uncompress_bench/64             309 ns          308 ns      2269007
uncompress_bench/uncompress_bench/1024           483 ns          483 ns      1449211
uncompress_bench/uncompress_bench/16384         4657 ns         4650 ns       150678
uncompress_bench/uncompress_bench/131072       26900 ns        26855 ns        26199
uncompress_bench/uncompress_bench/1048576     208094 ns       207662 ns         3368
```
Unlikely hints shrink the code by 320 bytes.
Performance numbers are not accurate enough to tell for sure. compress_bench>512 seems to be faster.


IOW, performance takes a noticeable hit with these changes, at least in my tests on an i9-9900K (x86-64 w/AVX2).

Dead2:
Random thought, without looking at the code, just wanted to make sure it does not end up forgotten;
Something we could perhaps do to at some point to regain some of the lost performance is to split this macro into two. send_bits() and send_bits_safe(), then call the old one where we know the new checks are not going to be needed. That is, if we can distinguish that at all.

phprus:
> As far as I managed to check, running without static analyzer never hits the added code and neither of the two length variables is ever zero with valid stream.

Do you think this is a false positive? (I'm not sure, but I can't prove it)
If so, then we need to add assert's (maybe instead of conditions).
Assert's may tell the static analyzer that this is an impossible path.

nmoinvaz:
Because of the performance hit, I would suggest we try to fix it another way. That code path is called a lot to emit bits. We can try and use the Dead2's `assume` PR? Or even downgrade the clang static analyzer version.

I have no problems with the `uint64_t` changes. It is only the `if` statements that are the problem.

mtl1979:
> > As far as I managed to check, running without static analyzer never hits the added code and neither of the two length variables is ever zero with valid stream.
> 
> 
> 
> Do you think this is a false positive? (I'm not sure, but I can't prove it)
> 
> If so, then we need to add assert's (maybe instead of conditions).
> 
> Assert's may tell the static analyzer that this is an impossible path.

I thought about using an assert but those don't generally trap on release code, so no easy way to verify if it is false positive. I wanted to see that the code pass all tests on CI first.

mtl1979:
> Because of the performance hit, I would suggest we try to fix it another way. That code path is called a lot to emit bits. We can try and use the Dead2's `assume` PR? Or even downgrade the clang static analyzer version.
> 
> 
> 
> I have no problems with the `uint64_t` changes. It is only the `if` statements that are the problem.

I'm open for suggestions.

nmoinvaz:
Referring to:
```
/home/runner/work/zlib-ng/zlib-ng/trees.c:569:5: warning: Left shift overflows the capacity of 'uint64_t' [core.BitwiseShift]
  569 |     send_bits(s, lcodes-257, 5, bi_buf, bi_valid); /* not +255 as stated in appnote.txt */
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/zlib-ng/zlib-ng/trees_emit.h:56:23: note: expanded from macro 'send_bits'
   56 |         bi_buf |= val << bi_valid;\
      |                   ~~~~^~~~~~~~~~~
```
Idk if this will work:
```c
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wbitwise-shift"
bi_buf |= val << bi_valid;
#pragma clang diagnostic pop
```
There is no real case where `bi_valid` will ever be greater than `BIT_BUF_SIZE` (64)

nmoinvaz:
We could also attempt something like this:
```c
bi_buf |= val << (bi_valid & BIT_BUF_SIZE);
```
Which I think is the same as `max(bi_valid, BIT_BUF_SIZE)`

mtl1979:
> We could also attempt something like this:
> 
> ```c
> bi_buf |= val << (bi_valid & BIT_BUF_SIZE);
> ```
> 
> Which I think is the same as `max(bi_valid, BIT_BUF_SIZE)`

That doesn't work... You can use "bitwise and" only with numbers where either lower or higher bits are all ones.

mtl1979:
> Idk if this will work:
> 
> ```c
> #pragma clang diagnostic push
> #pragma clang diagnostic ignored "-Wbitwise-shift"
> bi_buf |= val << bi_valid;
> #pragma clang diagnostic pop
> ```
> 
> There is no real case where `bi_valid` will ever be greater than `BIT_BUF_SIZE` (64)

You can't use pragmas within preprocessor macros.

trees_emit.h
```diff
@@ -45,13 +45,13 @@ extern Z_INTERNAL const int base_dist[D_CODES];
     uint32_t total_bits = bi_valid + len;\
     send_bits_trace(s, val, len);\
     sent_bits_add(s, len);\
-    if (total_bits < BIT_BUF_SIZE) {\
-        bi_buf |= val << bi_valid;\
-        bi_valid = total_bits;\
-    } else if (bi_valid == BIT_BUF_SIZE) {\
+    if (bi_valid >= BIT_BUF_SIZE) {\
```
mtl1979: Moving this check earlier avoids need for checking range of `len` as on next iteration it will be `bi_valid` here. We can't rely on `total_bits` when using static analyser as exact value can't be determined on compile time. We need to use `>=` here due to not knowing exact value of `bi_valid` on compile time, when using static analyser it might be either less or more than `BIT_BUF_SIZE`.

nmoinvaz: This change looks better. I think there might still be slight performance issue because the most common path is not the first if statement, but perhaps @Dead2 can do another test.

mtl1979: I wasn't sure if using `UNLIKELY()` here would help... That needs proper benchmarking to verify... It takes really long time for CI to finish all tests, so I didn't want to experiment too much while I was testing on Linux.

Dead2:
Retested with updated PR

For reference, Develop:
```
 1     54.185%      0.078/0.082/0.083/0.001        0.027/0.030/0.031/0.001        8,526,745
 2     43.871%      0.144/0.149/0.151/0.001        0.027/0.030/0.031/0.001        6,903,702
 3     42.390%      0.178/0.182/0.183/0.001        0.025/0.028/0.030/0.001        6,670,664
 4     41.644%      0.206/0.212/0.214/0.002        0.024/0.028/0.030/0.001        6,553,205
 5     41.215%      0.235/0.238/0.241/0.002        0.024/0.028/0.029/0.001        6,485,659
 6     41.032%      0.277/0.285/0.290/0.004        0.023/0.027/0.029/0.002        6,456,912
 7     40.778%      0.381/0.388/0.392/0.003        0.024/0.028/0.029/0.001        6,416,941
 8     40.704%      0.492/0.501/0.506/0.004        0.022/0.027/0.030/0.002        6,405,249
 9     40.409%      0.588/0.595/0.599/0.003        0.024/0.027/0.028/0.001        6,358,951

 avg1  42.914%                        0.292                          0.028
 tot                                 78.931                          7.601       60,778,028

 130504    1344       8  131856   20310 libz-ng.so.2

compress_bench/compress_bench/1                 5347 ns         5338 ns       124032
compress_bench/compress_bench/8                 5631 ns         5621 ns       117456
compress_bench/compress_bench/16                5669 ns         5660 ns       120083
compress_bench/compress_bench/32                6035 ns         6025 ns       112324
compress_bench/compress_bench/64                6344 ns         6334 ns       106719
compress_bench/compress_bench/512               6809 ns         6798 ns       100918
compress_bench/compress_bench/4096             12010 ns        11988 ns        58512
compress_bench/compress_bench/32768            46215 ns        46130 ns        15176
uncompress_bench/uncompress_bench/1              150 ns          149 ns      4683675
uncompress_bench/uncompress_bench/64             310 ns          309 ns      2264460
uncompress_bench/uncompress_bench/1024           484 ns          483 ns      1449716
uncompress_bench/uncompress_bench/16384         4637 ns         4629 ns       150867
uncompress_bench/uncompress_bench/131072       26841 ns        26795 ns        26045
uncompress_bench/uncompress_bench/1048576     207574 ns       207153 ns         3372
```

This PR:
```
 1     54.185%      0.083/0.087/0.088/0.001        0.028/0.030/0.031/0.001        8,526,745
 2     43.871%      0.149/0.152/0.154/0.002        0.028/0.031/0.033/0.001        6,903,702
 3     42.390%      0.179/0.184/0.186/0.002        0.026/0.029/0.031/0.001        6,670,664
 4     41.644%      0.207/0.213/0.215/0.002        0.024/0.029/0.030/0.002        6,553,205
 5     41.215%      0.236/0.240/0.243/0.002        0.024/0.028/0.030/0.001        6,485,659
 6     41.032%      0.278/0.287/0.291/0.003        0.025/0.028/0.029/0.001        6,456,912
 7     40.778%      0.381/0.389/0.394/0.003        0.027/0.028/0.030/0.001        6,416,941
 8     40.704%      0.493/0.501/0.506/0.003        0.027/0.029/0.030/0.001        6,405,249
 9     40.409%      0.583/0.594/0.598/0.003        0.025/0.028/0.029/0.001        6,358,951

 avg1  42.914%                        0.294                          0.029
 tot                                 79.421                          7.776       60,778,028

 130376    1344       8  131728   20290 libz-ng.so.2

compress_bench/compress_bench/1                 5336 ns         5327 ns       123379
compress_bench/compress_bench/8                 5646 ns         5606 ns       119203
compress_bench/compress_bench/16                5863 ns         5854 ns       113485
compress_bench/compress_bench/32                6057 ns         6047 ns       112427
compress_bench/compress_bench/64                6357 ns         6347 ns       108111
compress_bench/compress_bench/512               6950 ns         6939 ns        98693
compress_bench/compress_bench/4096             11944 ns        11923 ns        58593
compress_bench/compress_bench/32768            46457 ns        46378 ns        15090
uncompress_bench/uncompress_bench/1              151 ns          151 ns      4643511
uncompress_bench/uncompress_bench/64             309 ns          309 ns      2268901
uncompress_bench/uncompress_bench/1024           485 ns          484 ns      1445545
uncompress_bench/uncompress_bench/16384         4663 ns         4655 ns       150376
uncompress_bench/uncompress_bench/131072       26968 ns        26924 ns        26216
uncompress_bench/uncompress_bench/1048576     207662 ns       207227 ns         3374
```

It is now 128 bytes smaller than Develop.
Compression is now only 0.6% slower on average.
Level 1 is 6% slower, resulting in this having a larger slowdown in ms, not only in percent.
Level 2 is 2% slower
Levels 3-6 are around 0.5-1% slower
Levels 7-9 no change.

Confirmed these benchmarks with another run of before/after compiled with increased function alignment, showing the same trends as the above (0.58% slower on average).

trees_emit.h
Dead2: Adding unlikely hint shows another decrease in code size, but unfortunately results in a further slowdown in my (repeated) tests.

mtl1979: > Adding unlikely hint shows another decrease in code size, but unfortunately results in a further slowdown in my (repeated) tests.

Then we don't add `UNLIKELY` hint... I'm not sure we can completely avoid slowdown as even without static analyser, we need the `bi_valid == BIT_BUF_SIZE` test to be first to get valid generated assembly code. The other two tests have significant overlap, `bi_valid` essentially being `total_bits` modulo 64.

nmoinvaz: What about using `bi_buf |= val << (bi_valid > BIT_BUF_SIZE ? BIT_BUF_SIZE: bit_valid)`? The extra if statement stays in the path that needs it.

mtl1979: > What about using `bi_buf |= val << (bi_valid > BIT_BUF_SIZE ? BIT_BUF_SIZE: bit_valid)`? The extra if statement stays in the path that needs it.

Shifting by exactly `BIT_BUF_SIZE` already overflows...

Dead2:
I benchmarked this alternative:
```diff
--- a/trees_emit.h
+++ b/trees_emit.h
@@ -45,10 +45,10 @@ extern Z_INTERNAL const int base_dist[D_CODES];
     uint32_t total_bits = bi_valid + len;\
     send_bits_trace(s, val, len);\
     sent_bits_add(s, len);\
-    if (total_bits < BIT_BUF_SIZE) {\
+    if (total_bits < BIT_BUF_SIZE && bi_valid < BIT_BUF_SIZE) {\
         bi_buf |= val << bi_valid;\
         bi_valid = total_bits;\
-    } else if (bi_valid == BIT_BUF_SIZE) {\
+    } else if (bi_valid >= BIT_BUF_SIZE) {\
         put_uint64(s, bi_buf);\
         bi_buf = val;\
         bi_valid = len;\
```
Wouldn't this also fix the problem? It keeps the block ordering more optimal.

```
 1     54.185%      0.082/0.085/0.087/0.001        0.027/0.030/0.031/0.001        8,526,745
 2     43.871%      0.148/0.151/0.152/0.001        0.026/0.030/0.032/0.002        6,903,702
 3     42.390%      0.179/0.181/0.183/0.001        0.025/0.028/0.030/0.001        6,670,664
 4     41.644%      0.206/0.213/0.215/0.003        0.026/0.028/0.029/0.001        6,553,205
 5     41.215%      0.233/0.239/0.242/0.002        0.024/0.028/0.029/0.001        6,485,659
 6     41.032%      0.283/0.288/0.291/0.002        0.024/0.027/0.029/0.002        6,456,912
 7     40.778%      0.384/0.388/0.391/0.002        0.023/0.026/0.029/0.002        6,416,941
 8     40.704%      0.493/0.500/0.505/0.004        0.025/0.028/0.030/0.001        6,405,249
 9     40.409%      0.589/0.597/0.603/0.004        0.024/0.027/0.029/0.001        6,358,951

 avg1  42.914%                        0.294                          0.028
 tot                                 79.269                          7.571       60,778,028

 130184    1344       8  131536   201d0 libz-ng.so.2
```
Surprisingly this results in even smaller code, -320bytes, the compiler must see some optimization opportunity that I don't.
On average 0.43% slower compression.

nmoinvaz:
That does look a bit better. I think the logic is probably ok. Is it necessary to change `bi_valid == BIT_BUF_SIZE` to `bi_valid >= BIT_BUF_SIZE`?

mtl1979:
> I benchmarked this alternative:
> 
> ```diff
> --- a/trees_emit.h
> +++ b/trees_emit.h
> @@ -45,10 +45,10 @@ extern Z_INTERNAL const int base_dist[D_CODES];
>      uint32_t total_bits = bi_valid + len;\
>      send_bits_trace(s, val, len);\
>      sent_bits_add(s, len);\
> -    if (total_bits < BIT_BUF_SIZE) {\
> +    if (total_bits < BIT_BUF_SIZE && bi_valid < BIT_BUF_SIZE) {\
>          bi_buf |= val << bi_valid;\
>          bi_valid = total_bits;\
> -    } else if (bi_valid == BIT_BUF_SIZE) {\
> +    } else if (bi_valid >= BIT_BUF_SIZE) {\
>          put_uint64(s, bi_buf);\
>          bi_buf = val;\
>          bi_valid = len;\
> ```
> 
> Wouldn't this also fix the problem? It keeps the block ordering more optimal.
> 

That's exactly the code what I thought first, but it's kinda less readable than the current "alternative"... If it performs better, I can change... I didn't like the extra test at top as every extra check usually means penalty somewhere and rely on optimizer to mask the penalty.

mtl1979:
> That does look a bit better. I think the logic is probably ok. Is it necessary to change `bi_valid == BIT_BUF_SIZE` to `bi_valid >= BIT_BUF_SIZE`?

Static analyser does always require that and we really try to not mess too much with it to avoid the risk of replacing bad code with another bad code...

Dead2:
Interestingly, testing the code in godbolt (although as a function, not a macro, so less chances for the compiler to optimize) seems to indicate that changing position on the two if checks would allow it to only have 2 extra code blocks instead of 3.
It is of course quite logical now that I think about it.
There is no code size change, and performance is similar, compression is only 0.31% slower, so it is possibly slightly faster than the other way around.

https://godbolt.org/z/91e5WT1sP

```
 1     54.185%      0.081/0.085/0.086/0.001        0.028/0.030/0.031/0.001        8,526,745
 2     43.871%      0.147/0.151/0.153/0.001        0.025/0.029/0.030/0.001        6,903,702
 3     42.390%      0.179/0.183/0.184/0.002        0.025/0.028/0.030/0.001        6,670,664
 4     41.644%      0.210/0.214/0.216/0.001        0.026/0.028/0.030/0.001        6,553,205
 5     41.215%      0.233/0.239/0.242/0.002        0.023/0.027/0.029/0.001        6,485,659
 6     41.032%      0.282/0.287/0.291/0.003        0.025/0.028/0.029/0.001        6,456,912
 7     40.778%      0.381/0.387/0.391/0.003        0.023/0.027/0.029/0.001        6,416,941
 8     40.704%      0.495/0.500/0.505/0.003        0.024/0.028/0.029/0.002        6,405,249
 9     40.409%      0.589/0.594/0.599/0.003        0.025/0.027/0.028/0.001        6,358,951

 avg1  42.914%                        0.293                          0.028
 tot                                 79.177                          7.564       60,778,028

 130184    1344       8  131536   201d0 libz-ng.so.2
```

```diff
--- a/trees_emit.h
+++ b/trees_emit.h
@@ -45,10 +45,10 @@ extern Z_INTERNAL const int base_dist[D_CODES];
     uint32_t total_bits = bi_valid + len;\
     send_bits_trace(s, val, len);\
     sent_bits_add(s, len);\
-    if (total_bits < BIT_BUF_SIZE) {\
+    if (bi_valid < BIT_BUF_SIZE && total_bits < BIT_BUF_SIZE) {\
         bi_buf |= val << bi_valid;\
         bi_valid = total_bits;\
-    } else if (bi_valid == BIT_BUF_SIZE) {\
+    } else if (bi_valid >= BIT_BUF_SIZE) {\
         put_uint64(s, bi_buf);\
         bi_buf = val;\
         bi_valid = len;\
```

mtl1979:
> Interestingly, testing the code in godbolt (although as a function, not a macro, so less chances for the compiler to optimize) seems to indicate that changing position on the two if checks would allow it to only have 2 extra code blocks instead of 3. It is of course quite logical now that I think about it. There is no code size change, and performance is similar, compression is only 0.31% slower, so it is possibly slightly faster than the other way around.

This is how I "parse" the logic.

1. Check if there is room for more bits, if there is, add as many bits as possible
2. Check if there is 64 used bits and flush if there is. After that check if we still have more bits to add and add them if there is
3. Update `bi_valid` by taking lowest 6 bits of `bi_valid + len`

Dead2:
> That's exactly the code what I thought first, but it's kinda less readable than the current "alternative"... If it performs better, I can change... I didn't like the extra test at top as every extra check usually means penalty somewhere and rely on optimizer to mask the penalty.

I guess the extra jump is more expensive than the extra test.
IIRC, on Broadwell for example, a tight loop of only jumps (worst case) is one jump every 12 cycles. In real code a large part of that latency can be hidden if the instruction and the target has already been decoded. Not sure what best case is, but I think ~4 cycles.

mtl1979:
> > That's exactly the code what I thought first, but it's kinda less readable than the current "alternative"... If it performs better, I can change... I didn't like the extra test at top as every extra check usually means penalty somewhere and rely on optimizer to mask the penalty.
> 
> I guess the extra jump is more expensive than the extra test. IIRC, on Broadwell for example, a tight loop of only jumps (worst case) is one jump every 12 cycles. In real code a large part of that latency can be hidden if the instruction and the target has already been decoded. Not sure what best case is, but I think ~4 cycles.

Modern compilers and processors are pretty good at guessing if jump is taken most of the time and reorder the code to make it as fast as possible. Like I said earlier, I'm open for suggestions from anyone, but I'm still worried if @nmoinvaz has trouble decoding the logic, then how can we expect that new or less experienced contributors can decode the logic and avoid breaking it later on... Honestly thinking a lot of the people who have contributed regularly are more experienced coders than I am.

Dead2:
> ```diff
> --- a/trees_emit.h
> +++ b/trees_emit.h
> @@ -45,10 +45,10 @@ extern Z_INTERNAL const int base_dist[D_CODES];
>      uint32_t total_bits = bi_valid + len;\
>      send_bits_trace(s, val, len);\
>      sent_bits_add(s, len);\
> -    if (total_bits < BIT_BUF_SIZE) {\
> +    if (bi_valid < BIT_BUF_SIZE && total_bits < BIT_BUF_SIZE) {\
>          bi_buf |= val << bi_valid;\
>          bi_valid = total_bits;\
> -    } else if (bi_valid == BIT_BUF_SIZE) {\
> +    } else if (bi_valid >= BIT_BUF_SIZE) {\
>          put_uint64(s, bi_buf);\
>          bi_buf = val;\
>          bi_valid = len;\
> ```

Also there should not be any need for checking `bi_valid < BIT_BUF_SIZE` in addition to `total_bits < BIT_BUF_SIZE`, right?
After all, `total_bits = bi_valid + len`, so if that is smaller than BIT_BUF_SIZE, then bi_valid will automatically also be smaller.

That would make the diff only
```
--- a/trees_emit.h
+++ b/trees_emit.h
@@ -48,7 +48,7 @@ extern Z_INTERNAL const int base_dist[D_CODES];
     if (total_bits < BIT_BUF_SIZE) {\
         bi_buf |= val << bi_valid;\
         bi_valid = total_bits;\
-    } else if (bi_valid == BIT_BUF_SIZE) {\
+    } else if (bi_valid >= BIT_BUF_SIZE) {\
         put_uint64(s, bi_buf);\
         bi_buf = val;\
         bi_valid = len;\
```

This also reduces the code size further.
```
 130056    1344       8  131408   20150 libz-ng.so.2
```

mtl1979:
> Also there should not be any need for checking `bi_valid < BIT_BUF_SIZE` in addition to `total_bits < BIT_BUF_SIZE`, right? After all, `total_bits = bi_valid + len`, so if that is smaller than BIT_BUF_SIZE, then bi_valid will automatically also be smaller.
> 

Static analyser doesn't work that way... `total_bits` is not evaluated at that point yet, because the value can't be evaluated at build time (during code tree generation). First time it is evaluated is after value `bi_valid` is replaced with `len`, because that is often a constant. Essentially it's an explicit hint (not assume hint) to static analyser of what the valid range is.

Dead2:
> > Also there should not be any need for checking `bi_valid < BIT_BUF_SIZE` in addition to `total_bits < BIT_BUF_SIZE`, right? After all, `total_bits = bi_valid + len`, so if that is smaller than BIT_BUF_SIZE, then bi_valid will automatically also be smaller.
> 
> Static analyser doesn't work that way... `total_bits` is not evaluated at that point yet, because the value can't be evaluated at build time (during code tree generation). First time it is evaluated is after value `bi_valid` is replaced with `len`, because that is often a constant. Essentially it's an explicit hint (not assume hint) to static analyser of what the valid range is.

Right, I had completely changed focus from static analyzer to optimizing the code while retaining correctness.
It is tricky if we cannot give range hints.

mtl1979:
> Right, I had completely changed focus from static analyzer to optimizing the code while retaining correctness. It is tricky if we cannot give range hints.

I have Clang 18 and that doesn't yet support range hints... Next big update will be April 2026 and then maybe the default becomes either Clang 19 or Clang 20 depending on if Clang 20 is stable enough.

Visual C++ has supported range hints for quite a while already. I haven't tested gcc 13 and later as those are broken for me.

trees_emit.h
```diff
@@ -45,10 +45,10 @@ extern Z_INTERNAL const int base_dist[D_CODES];
     uint32_t total_bits = bi_valid + len;\
     send_bits_trace(s, val, len);\
     sent_bits_add(s, len);\
-    if (total_bits < BIT_BUF_SIZE) {\
+    if (total_bits < BIT_BUF_SIZE && bi_valid < BIT_BUF_SIZE) {\
```
Dead2, line 52: **suggestion:** 
I think it would be helpful to have a comment explaining why we check bi_valid when it is not really needed for the code to work. Inline or above the macro would both be fine I think.
Perhaps something like this?
/* To help static checkers understand the code, we need to have an extra check for bi_valid  */ \

/* To avoid static checker warnings, we need to check for >= instead of == */ \

Dead2 (approved):
Code looks good. Unless there is further discussion/changes, I'll merge this once comments have been added.

trees_emit.h
mtl1979: I think above the macro is the best place...

Issue #1861, opened 2025-02-03
MSVC 2015 Debug x86 build occasionally crashing with AVX2 enabled (2.2.3 regression)


GerbilSoft:
After updating my copy of zlib-ng from 2.2.2 to 2.2.3, I've noticed some crashes in MSVC 2015 Debug x86 builds. (These crashes don't happen in Release or x64 builds, or with MSVC 2022.)

It's specifically crashing here: https://github.com/zlib-ng/zlib-ng/blob/2.2.3/arch/x86/chunkset_avx2.c#L34-L36

`from` in one case where it crashes is 0x02b930c9, which implies it's failing due to alignment issues. This doesn't make sense because the `_mm_loadu_si128()` intrinsic is used.

Disassembly:
```asm
    *chunk = _mm256_broadcastsi128_si256(_mm_loadu_si128((__m128i*)from));
   5DEE4D97  mov         eax,dword ptr [from]  
   5DEE4D9A  vmovdqu     xmm0,xmmword ptr [eax]  
=> 5DEE4D9E  vmovdqa     xmmword ptr [ebp-10h],xmm0  
   5DEE4DA3  vbroadcasti128 ymm0,oword ptr [ebp-10h]  
```

It appears MSVC 2015 put an aligned vmovdqa here. Starting with MSVC 2017, the compiler never emits aligned SSE/AVX2 instructions, which explains why it doesn't crash in MSVC 2022, but it doesn't explain why it's only crashing in the Debug x86 build with MSVC 2015.

Not sure how to fix this other than disabling AVX2 (and maybe AVX-512) when building with MSVC 2015.

KungFuJesus:
That intrinsic is specifically unaligned, I'm not sure exactly what to do but chalk this up to a compiler bug. Maybe file a bug with Microsoft?

KungFuJesus:
Actually there is an unaligned load there. I think there's an intermediate load being generated there by the compiler that's getting optimized out. I'm not entirely sure _why_ it wants to make that trip to the stack, that broadcast I think should support register, register operands. Or maybe it doesn't and that's the real issue?

GerbilSoft:
Filing a bug with MS won't do any good because MSVC 2015 is no longer supported. If it's a compiler bug, then I'll just have to disable AVX2 when using MSVC 2015 (and I suppose earlier versions too).

(Well, technically "extended" end of support is 2025/10/14, but that requires having an extended support contract.)

[...]

PR #1862, opened 2025-02-03
Fix an unfortunate bug with Visual Studio 2015

KungFuJesus:
Evidently this instruction, despite the intrinsic having a register operand,
is a memory-register instruction. There seems to be no alignment requirement
for the source operand. Because of this, compilers when not optimized are doing
the unaligned load and then dumping back to the stack to do the broadcasting load.
In doing this, MSVC seems to be dumping to the stack with an aligned move at an
unaligned address, causing a segfault.  GCC does not seem to make this mistake, as
it stashes to an aligned address.

If we're on Visual Studio 2015, let's just do the longer 9 cycle sequence of a 128
bit load followed by a vinserti128. This _should_ fix this (issue https://github.com/zlib-ng/zlib-ng/issues/1861).

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced internal memory handling to resolve performance issues on legacy compiler environments, ensuring consistent efficiency and stability across different systems.

### DIFF
--- a/arch/x86/chunkset_avx2.c
+++ b/arch/x86/chunkset_avx2.c
@@ -32,7 +32,13 @@ static inline void chunkmemset_8(uint8_t *from, chunk_t *chunk) {
 }
 
 static inline void chunkmemset_16(uint8_t *from, chunk_t *chunk) {
+    /* See explanation in chunkset_avx512.c */
+#if defined(_MSC_VER) && _MSC_VER <= 1900
+    halfchunk_t half = _mm_loadu_si128((__m128i*)from);
+    *chunk = _mm256_inserti128_si256(_mm256_castsi128_si256(half), half, 1);
+#else
     *chunk = _mm256_broadcastsi128_si256(_mm_loadu_si128((__m128i*)from));
+#endif
 }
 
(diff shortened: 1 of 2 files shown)


arch/x86/chunkset_avx2.c
```diff
@@ -32,7 +32,7 @@ static inline void chunkmemset_8(uint8_t *from, chunk_t *chunk) {
 }
 
 static inline void chunkmemset_16(uint8_t *from, chunk_t *chunk) {
-    *chunk = _mm256_broadcastsi128_si256(_mm_loadu_si128((__m128i*)from));
+    *chunk = _mm256_broadcastsi128_si256(*((__m128i*)from));
```
nmoinvaz: An inline comment might be a good idea so nobody tries it again.

KungFuJesus: This change actually has the effect of doing exactly what we didn't want - forcing an aligned load at an unaligned offset. I'll need to redo this PR completely.

Short of inline assembly to do what the intrinsic should have been defined as all along the best we could do is use inserti128 if we detect the broken compiler.

nmoinvaz (commented):
Lgtm

Dead2 (approved):
LGTM

PR #1865, opened 2025-02-05
Fix pkgconfig for cmake mingw builds

### LINKED ISSUE #1864: Windows: zlib.pc linker flags do not match library name
Hello!

When building a shared mingw32 lib, the build outputs a `libzlib.dll.a` linker library.
However the `zlib.pc` has linker flags `Libs: -L${libdir} -L${sharedlibdir} -lz`.

Either the `.pc`-file should list `-lzlib` or the static linker library for the shared lib should be named
`libz.dll.a` (ie be named unix style).

Opinions which is better? I'd tend to alter the `.pc` files for the windows case as this leaves hard-coded linker-flags in consumers in a working state while consumers utilizing pkg-config will pick up the change with no problems (I hope).
 
I would have a go on fixing this if we decided on a way to handle this.

hlewin:
This fixes https://github.com/zlib-ng/zlib-ng/issues/1864 for cmake builds. It does not for stuff build by `configure` - I don't know about that and will not touch it.

## Summary by CodeRabbit

- **Chores**
	- Improved build configuration to automatically adjust library naming for various build setups, ensuring smoother library integration across environments.


hlewin:
Reviewing this I found another discrepancy: The static lib under mingw32 is really called `libz.a` which is different from the name for the shared lib. This cannot be fixed without breaking consumers. The shared lib should be named `libz` also.
I'll fix this for my use-case with patches. Do whatever you want with this PR

PR #1867, opened 2025-02-09
Added --installnamedir

radarhere:
Rather than having the `install_name` always set to `@rpath/${SHAREDLIBM}` on macOS

https://github.com/zlib-ng/zlib-ng/blob/43b27034358d657a2ef6926009d3b5553ad3799c/configure#L473

this PR suggests making it possible to replace `@rpath` via `--installnamedir`.

This is helpful as setting it can remove a need for `DYLD_LIBRARY_PATH` - recent versions of macOS have System Integrity Protection (SIP) which prevents `DYLD_LIBRARY_PATH` from being passed into subshells.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new command-line option that allows users to specify a custom installation directory during build.
	- Updated help messaging to clearly present this configurable option.
	- Enhanced logging to record the selected installation directory for future reference.

### DIFF
--- a/configure
+++ b/configure
@@ -83,6 +83,7 @@ bindir=${bindir-'${exec_prefix}/bin'}
 libdir=${libdir-'${exec_prefix}/lib'}
 sharedlibdir=${sharedlibdir-'${libdir}'}
 includedir=${includedir-'${prefix}/include'}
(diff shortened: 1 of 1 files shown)


Dead2 (approved):
LGTM

PR #1870, opened 2025-02-12
ARM Neon: Fold a copy into the adler32 function for UPDATEWINDOW

KungFuJesus:
So a lot of alterations had to be done to make this not worse and so far, it's not really better, either. I had to force inlining for the adler routine, I had to remove the x4 load instruction otherwise pipelining stalled, and I had to use restrict pointers with a copy idiom for GCC to inline a copy routine for the tail.

Still, we see a small benefit in benchmarks, particularly when done with size of our window or larger. There's also an added benefit that this will fix #1824.

## Summary by CodeRabbit

- **New Features**
  - Enhanced checksum processing on ARM platforms leveraging advanced SIMD support with an optional data copying mode.
  - Introduced a new inline macro for improved function inlining across different compilers.

- **Tests**
  - Updated performance benchmarks to evaluate the enhanced checksum routine, ensuring robust performance across supported architectures.

### DIFF
--- a/arch/arm/adler32_neon.c
+++ b/arch/arm/adler32_neon.c
@@ -10,17 +10,141 @@
 #include "zbuild.h"
 #include "adler32_p.h"
 
-static void NEON_accum32(uint32_t *s, const uint8_t *buf, size_t len) {
-    static const uint16_t ALIGNED_(16) taps[64] = {
-        64, 63, 62, 61, 60, 59, 58, 57,
-        56, 55, 54, 53, 52, 51, 50, 49,
-        48, 47, 46, 45, 44, 43, 42, 41,
-        40, 39, 38, 37, 36, 35, 34, 33,
-        32, 31, 30, 29, 28, 27, 26, 25,
-        24, 23, 22, 21, 20, 19, 18, 17,
-        16, 15, 14, 13, 12, 11, 10, 9,
-        8, 7, 6, 5, 4, 3, 2, 1 };
+static const uint16_t ALIGNED_(64) taps[64] = {
+    64, 63, 62, 61, 60, 59, 58, 57,
+    56, 55, 54, 53, 52, 51, 50, 49,
+    48, 47, 46, 45, 44, 43, 42, 41,
+    40, 39, 38, 37, 36, 35, 34, 33,
+    32, 31, 30, 29, 28, 27, 26, 25,
+    24, 23, 22, 21, 20, 19, 18, 17,
+    16, 15, 14, 13, 12, 11, 10, 9,
+    8, 7, 6, 5, 4, 3, 2, 1 };
 
+static Z_FORCEINLINE void NEON_accum32_copy(uint32_t *s, uint8_t *dst, const uint8_t *buf, size_t len) {
+    uint32x4_t adacc = vdupq_n_u32(0);
+    uint32x4_t s2acc = vdupq_n_u32(0);
+    uint32x4_t s2acc_0 = vdupq_n_u32(0);
+    uint32x4_t s2acc_1 = vdupq_n_u32(0);
+    uint32x4_t s2acc_2 = vdupq_n_u32(0);
+
+    adacc = vsetq_lane_u32(s[0], adacc, 0);
+    s2acc = vsetq_lane_u32(s[1], s2acc, 0);
+
+    uint32x4_t s3acc = vdupq_n_u32(0);
+    uint32x4_t adacc_prev = adacc;
+
+    uint16x8_t s2_0, s2_1, s2_2, s2_3;
+    s2_0 = s2_1 = s2_2 = s2_3 = vdupq_n_u16(0);
+
+    uint16x8_t s2_4, s2_5, s2_6, s2_7;
+    s2_4 = s2_5 = s2_6 = s2_7 = vdupq_n_u16(0);
+
+    size_t num_iter = len >> 2;
+    int rem = len & 3;
+
+    for (size_t i = 0; i < num_iter; ++i) {
+        uint8x16_t d0 = vld1q_u8(buf);
+        uint8x16_t d1 = vld1q_u8(buf + 16);
+        uint8x16_t d2 = vld1q_u8(buf + 32);
+        uint8x16_t d3 = vld1q_u8(buf + 48);
+
+        vst1q_u8(dst, d0);
+        vst1q_u8(dst + 16, d1);
+        vst1q_u8(dst + 32, d2);
+        vst1q_u8(dst + 48, d3);
+        dst += 64;
+
+        /* Unfortunately it doesn't look like there's a direct sum 8 bit to 32
+         * bit instruction, we'll have to make due summing to 16 bits first */
+        uint16x8x2_t hsum, hsum_fold;
+        hsum.val[0] = vpaddlq_u8(d0);
+        hsum.val[1] = vpaddlq_u8(d1);
+
+        hsum_fold.val[0] = vpadalq_u8(hsum.val[0], d2);
+        hsum_fold.val[1] = vpadalq_u8(hsum.val[1], d3);
+
+        adacc = vpadalq_u16(adacc, hsum_fold.val[0]);
+        s3acc = vaddq_u32(s3acc, adacc_prev);
+        adacc = vpadalq_u16(adacc, hsum_fold.val[1]);
+
+        /* If we do straight widening additions to the 16 bit values, we don't incur
+         * the usual penalties of a pairwise add. We can defer the multiplications
+         * until the very end. These will not overflow because we are incurring at
+         * most 408 loop iterations (NMAX / 64), and a given lane is only going to be
+         * summed into once. This means for the maximum input size, the largest value
+         * we will see is 255 * 102 = 26010, safely under uint16 max */
+        s2_0 = vaddw_u8(s2_0, vget_low_u8(d0));
+        s2_1 = vaddw_high_u8(s2_1, d0);
+        s2_2 = vaddw_u8(s2_2, vget_low_u8(d1));
+        s2_3 = vaddw_high_u8(s2_3, d1);
+        s2_4 = vaddw_u8(s2_4, vget_low_u8(d2));
+        s2_5 = vaddw_high_u8(s2_5, d2);
+        s2_6 = vaddw_u8(s2_6, vget_low_u8(d3));
+        s2_7 = vaddw_high_u8(s2_7, d3);
+
+        adacc_prev = adacc;
+        buf += 64;
+    }
+
+    s3acc = vshlq_n_u32(s3acc, 6);
+
+    if (rem) {
+        uint32x4_t s3acc_0 = vdupq_n_u32(0);
+        while (rem--) {
+            uint8x16_t d0 = vld1q_u8(buf);
+            vst1q_u8(dst, d0);
+            dst += 16;
+            uint16x8_t adler;
+            adler = vpaddlq_u8(d0);
+            s2_6 = vaddw_u8(s2_6, vget_low_u8(d0));
+            s2_7 = vaddw_high_u8(s2_7, d0);
+            adacc = vpadalq_u16(adacc, adler);
+            s3acc_0 = vaddq_u32(s3acc_0, adacc_prev);
+            adacc_prev = adacc;
+            buf += 16;
+        }
+
+        s3acc_0 = vshlq_n_u32(s3acc_0, 4);
+        s3acc = vaddq_u32(s3acc_0, s3acc);
+    }
+
+    uint16x8x4_t t0_t3 = vld1q_u16_x4(taps);
+    uint16x8x4_t t4_t7 = vld1q_u16_x4(taps + 32);
+
+    s2acc = vmlal_high_u16(s2acc, t0_t3.val[0], s2_0);
+    s2acc_0 = vmlal_u16(s2acc_0, vget_low_u16(t0_t3.val[0]), vget_low_u16(s2_0));
+    s2acc_1 = vmlal_high_u16(s2acc_1, t0_t3.val[1], s2_1);
+    s2acc_2 = vmlal_u16(s2acc_2, vget_low_u16(t0_t3.val[1]), vget_low_u16(s2_1));
+
+    s2acc = vmlal_high_u16(s2acc, t0_t3.val[2], s2_2);
+    s2acc_0 = vmlal_u16(s2acc_0, vget_low_u16(t0_t3.val[2]), vget_low_u16(s2_2));
+    s2acc_1 = vmlal_high_u16(s2acc_1, t0_t3.val[3], s2_3);
+    s2acc_2 = vmlal_u16(s2acc_2, vget_low_u16(t0_t3.val[3]), vget_low_u16(s2_3));
+
+    s2acc = vmlal_high_u16(s2acc, t4_t7.val[0], s2_4);
+    s2acc_0 = vmlal_u16(s2acc_0, vget_low_u16(t4_t7.val[0]), vget_low_u16(s2_4));
+    s2acc_1 = vmlal_high_u16(s2acc_1, t4_t7.val[1], s2_5);
+    s2acc_2 = vmlal_u16(s2acc_2, vget_low_u16(t4_t7.val[1]), vget_low_u16(s2_5));
+
+    s2acc = vmlal_high_u16(s2acc, t4_t7.val[2], s2_6);
+    s2acc_0 = vmlal_u16(s2acc_0, vget_low_u16(t4_t7.val[2]), vget_low_u16(s2_6));
+    s2acc_1 = vmlal_high_u16(s2acc_1, t4_t7.val[3], s2_7);
+    s2acc_2 = vmlal_u16(s2acc_2, vget_low_u16(t4_t7.val[3]), vget_low_u16(s2_7));
+
+    s2acc = vaddq_u32(s2acc_0, s2acc);
+    s2acc_2 = vaddq_u32(s2acc_1, s2acc_2);
+    s2acc = vaddq_u32(s2acc, s2acc_2);
+
+    uint32x2_t adacc2, s2acc2, as;
+    s2acc = vaddq_u32(s2acc, s3acc);
+    adacc2 = vpadd_u32(vget_low_u32(adacc), vget_high_u32(adacc));
+    s2acc2 = vpadd_u32(vget_low_u32(s2acc), vget_high_u32(s2acc));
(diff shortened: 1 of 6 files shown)


neheb:
Nope

```
[4/41] Compiling C object libneon.a.p/arch_arm_slide_hash_neon.c.o
FAILED: libneon.a.p/arch_arm_slide_hash_neon.c.o 
clang -Ilibneon.a.p -I. -I.. -fcolor-diagnostics -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=c11 -O2 -g -fPIC -D_LARGEFILE64_SOURCE=1 -D__USE_LARGEFILE64 -DHAVE_LINUX_AUXVEC_H -DHAVE_SYS_AUXV_H -DHAVE_ARM_ACLE_H -DHAVE_BUILTIN_ASSUME_ALIGNED -DHAVE_BUILTIN_CTZ -DHAVE_BUILTIN_CTZLL -DHAVE_ALIGNED_ALLOC -DHAVE_POSIX_MEMALIGN -DHAVE_ATTRIBUTE_ALIGNED -DHAVE_VISIBILITY_HIDDEN -DHAVE_VISIBILITY_INTERNAL -DWITH_GZFILEOP -DARM_FEATURES -DARM_AUXV_HAS_CRC32 -DARM_ASM_HWCAP -DARM_AUXV_HAS_NEON -DARM_ACLE -DARM_NEON -DARM_NEON_HASLD4 -mfpu=neon -MD -MQ libneon.a.p/arch_arm_slide_hash_neon.c.o -MF libneon.a.p/arch_arm_slide_hash_neon.c.o.d -o libneon.a.p/arch_arm_slide_hash_neon.c.o -c ../arch/arm/slide_hash_neon.c
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
0.	Program arguments: clang -Ilibneon.a.p -I. -I.. -fcolor-diagnostics -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=c11 -O2 -g -fPIC -D_LARGEFILE64_SOURCE=1 -D__USE_LARGEFILE64 -DHAVE_LINUX_AUXVEC_H -DHAVE_SYS_AUXV_H -DHAVE_ARM_ACLE_H -DHAVE_BUILTIN_ASSUME_ALIGNED -DHAVE_BUILTIN_CTZ -DHAVE_BUILTIN_CTZLL -DHAVE_ALIGNED_ALLOC -DHAVE_POSIX_MEMALIGN -DHAVE_ATTRIBUTE_ALIGNED -DHAVE_VISIBILITY_HIDDEN -DHAVE_VISIBILITY_INTERNAL -DWITH_GZFILEOP -DARM_FEATURES -DARM_AUXV_HAS_CRC32 -DARM_ASM_HWCAP -DARM_AUXV_HAS_NEON -DARM_ACLE -DARM_NEON -DARM_NEON_HASLD4 -mfpu=neon -MD -MQ libneon.a.p/arch_arm_slide_hash_neon.c.o -MF libneon.a.p/arch_arm_slide_hash_neon.c.o.d -o libneon.a.p/arch_arm_slide_hash_neon.c.o -c ../arch/arm/slide_hash_neon.c 
1.	<eof> parser at end of file
2.	Code generation
3.	Running pass 'Function Pass Manager' on module '../arch/arm/slide_hash_neon.c'.
4.	Running pass 'ARM Instruction Selection' on function '@slide_hash_neon'
clang: error: clang frontend command failed due to signal (use -v to see invocation)


[1/37] Compiling C object libneon.a.p/arch_arm_adler32_neon.c.o
FAILED: libneon.a.p/arch_arm_adler32_neon.c.o 
clang -Ilibneon.a.p -I. -I.. -fcolor-diagnostics -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=c11 -O2 -g -fPIC -D_LARGEFILE64_SOURCE=1 -D__USE_LARGEFILE64 -DHAVE_LINUX_AUXVEC_H -DHAVE_SYS_AUXV_H -DHAVE_ARM_ACLE_H -DHAVE_BUILTIN_ASSUME_ALIGNED -DHAVE_BUILTIN_CTZ -DHAVE_BUILTIN_CTZLL -DHAVE_ALIGNED_ALLOC -DHAVE_POSIX_MEMALIGN -DHAVE_ATTRIBUTE_ALIGNED -DHAVE_VISIBILITY_HIDDEN -DHAVE_VISIBILITY_INTERNAL -DWITH_GZFILEOP -DARM_FEATURES -DARM_AUXV_HAS_CRC32 -DARM_ASM_HWCAP -DARM_AUXV_HAS_NEON -DARM_ACLE -DARM_NEON -DARM_NEON_HASLD4 -mfpu=neon -MD -MQ libneon.a.p/arch_arm_adler32_neon.c.o -MF libneon.a.p/arch_arm_adler32_neon.c.o.d -o libneon.a.p/arch_arm_adler32_neon.c.o -c ../arch/arm/adler32_neon.c
PLEASE submit a bug report to https://bugs.llvm.org/ and include the crash backtrace, preprocessed source, and associated run script.
Stack dump:
0.	Program arguments: clang -Ilibneon.a.p -I. -I.. -fcolor-diagnostics -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -std=c11 -O2 -g -fPIC -D_LARGEFILE64_SOURCE=1 -D__USE_LARGEFILE64 -DHAVE_LINUX_AUXVEC_H -DHAVE_SYS_AUXV_H -DHAVE_ARM_ACLE_H -DHAVE_BUILTIN_ASSUME_ALIGNED -DHAVE_BUILTIN_CTZ -DHAVE_BUILTIN_CTZLL -DHAVE_ALIGNED_ALLOC -DHAVE_POSIX_MEMALIGN -DHAVE_ATTRIBUTE_ALIGNED -DHAVE_VISIBILITY_HIDDEN -DHAVE_VISIBILITY_INTERNAL -DWITH_GZFILEOP -DARM_FEATURES -DARM_AUXV_HAS_CRC32 -DARM_ASM_HWCAP -DARM_AUXV_HAS_NEON -DARM_ACLE -DARM_NEON -DARM_NEON_HASLD4 -mfpu=neon -MD -MQ libneon.a.p/arch_arm_adler32_neon.c.o -MF libneon.a.p/arch_arm_adler32_neon.c.o.d -o libneon.a.p/arch_arm_adler32_neon.c.o -c ../arch/arm/adler32_neon.c 
1.	<eof> parser at end of file
2.	Code generation
3.	Running pass 'Function Pass Manager' on module '../arch/arm/adler32_neon.c'.
4.	Running pass 'ARM Instruction Selection' on function '@adler32_neon'
clang: error: clang frontend command failed due to signal (use -v to see invocation)
Debian clang version 11.0.1-2
Target: armv7l-unknown-linux-gnueabihf
Thread model: posix
InstalledDir: /usr/bin
clang: note: diagnostic msg: 
********************

PLEASE ATTACH THE FOLLOWING FILES TO THE BUG REPORT:
Preprocessed source(s) and associated run script(s) are located at:
clang: note: diagnostic msg: /tmp/adler32_neon-19e03b.c
clang: note: diagnostic msg: /tmp/adler32_neon-19e03b.sh
clang: note: diagnostic msg: 

```

KungFuJesus:
Hmm, the slide hash function should be pointed at the buffer window, which _should_ be properly aligned.  Do you have more error context for that one?

As for the adler, the only thing we're using the multi-register variant with ld1 with is something that we control the alignment of on the stack.  Everything else is using plain old ld1 (really 2x ldp's, as that's usually what both clang and GCC compile to):

```
 ldp    q4, q3, [x2]
 ldp    q2, q1, [x2, #32]
```

Of course this is aarch64, not necessarily armv7.

neheb:
@KungFuJesus nope. This is Debian 11 with clang-11.

Interestingly enough, if I use an Alpine Linux container with distrobox, it compiles and it works fine after this PR. Before this PR, example.c would crash when ran.

KungFuJesus:
Ok so the failing to compile is probably a separate issue.  There's probably something weird going on with your build toolchain.

neheb:
huh? it's stock Debian. clang was installed with sudo apt install clang.

KungFuJesus:
Let me try to reproduce with my potato pi on a similar vintage of debian:
```
pi@pitemp:/mnt/nasshare/storage/armv7-zlib-ng $ lsb_release  -a
No LSB modules are available.
Distributor ID:	Raspbian
Description:	Raspbian GNU/Linux 11 (bullseye)
Release:	11
Codename:	bullseye
pi@pitemp:/mnt/nasshare/storage/armv7-zlib-ng $ clang --version
Raspbian clang version 11.0.1-2+rpi1
Target: armv6k-unknown-linux-gnueabihf
Thread model: posix
InstalledDir: /usr/bin
```
The armv6 abi but I'll give it a go.

KungFuJesus:
Hmm, this might be too potato, it's just compiling functions with the fallback v6 asm:
```
[ 23%] Building C object CMakeFiles/zlibstatic.dir/arch/arm/arm_features.c.o
[ 25%] Building C object CMakeFiles/zlibstatic.dir/arch/arm/crc32_acle.c.o
[ 25%] Building C object CMakeFiles/zlibstatic.dir/arch/arm/slide_hash_armv6.c.o
```
I don't really have any aarch32 installs to test this on, everything I have that's aarch64 capable is running on aarch64.  I might have a second gen pi laying around somewhere but I'd have to track it down.  Suffice to say though if the compiler is crashing and spitting a stack trace out it's probably a compiler bug.  Either that or your clang binaries are somehow corrupt.

neheb:
Correct. That stack trace is indeed an internal LLVM bug.

arch/arm/adler32_neon.c
```diff
@@ -141,22 +157,29 @@ static void NEON_handle_tail(uint32_t *pair, const uint8_t *buf, size_t len) {
...
+    if (len == 1) {
+        if (COPY)
+           *dst = *src;
+        return adler32_len_1(adler, src, sum2);
+    }
 
     /* initial Adler-32 value (deferred check for len == 1 speed) */
-    if (buf == NULL)
+    if (src == NULL)
         return 1L;
```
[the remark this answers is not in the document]
KungFuJesus: This is not a function that's exposed in the API and is only consumed internally.  There's no reason anybody would ever call this with copy being true and dist being NULL.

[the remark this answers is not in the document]
KungFuJesus: We can in a relatively inexpensive way control the COPY variable from the caller.  As a precaution I'll check dst there.

zbuild.h
```diff
@@ -70,6 +70,15 @@
...
 #endif
 
+/* A forced inline decorator */
+#if defined(_MSC_VER)
+#define Z_INLINE __forceinline
+#elif defined(__GNUC__)
+#define Z_INLINE __attribute__((always_inline))
+#else
+#define Z_INLINE
+#endif
```
phprus: Suggestion for fix MSVC error:


```
D:\a\zlib-ng\zlib-ng\arch\arm\adler32_neon.c(14,8): error C2220: the following warning is treated as an error [D:\a\zlib-ng\zlib-ng\zlib.vcxproj]
D:\a\zlib-ng\zlib-ng\arch\arm\adler32_neon.c(14,8): warning C4141: 'inline': used more than once [D:\a\zlib-ng\zlib-ng\zlib.vcxproj]
D:\a\zlib-ng\zlib-ng\arch\arm\adler32_neon.c(162,8): warning C4141: 'inline': used more than once [D:\a\zlib-ng\zlib-ng\zlib.vcxproj]
```

```c
/* A forced inline decorator */
#if defined(_MSC_VER)
#  define Z_FORCEINLINE __forceinline
#elif defined(__GNUC__)
#  define Z_FORCEINLINE inline __attribute__((always_inline))
#else
#  define Z_FORCEINLINE inline
#endif
```


And use 
```c
static Z_FORCEINLINE
```

instead of

```c
Z_INLINE
static inline
```

zbuild.h
```diff
@@ -70,6 +70,15 @@
...
+
```
Dead2, line 82: **suggestion (non-blocking):** 
Wrapping this in a test for `WITHOUT_FORCEINLINE` or `NO_FORCEINLINE` could be helpful for testing and debugging, and maybe also for the few ones that want to minimize library size for use in embedded and such.

KungFuJesus: I can see how that might be useful if we come to use it in many places.  In this circumstance, there's a lot of stack churn going into and out of NEON_accum32, it's quite a bit slower without inlining.  Maybe we can do that in a later PR.

mtl1979: Instead of allowing to disable forced inlining, we should consider using x4 versions only for non-copy version... This would mean we would need to duplicate the function instead of using a parameter to decide if we need to copy... On a lot of known platforms where x4 versions don't work, we already convert them to equivalent 4 x1 calls... For the remaining, we need to fix the issue anyways instead of masking it.

Easiest way to avoid inlining is to build debug version as it will disable all inlining.

Having two versions of the function also makes it easier to test for regressions as there is always "spare" version to compare against especially when the plain C version is not included for reducing binary size.

KungFuJesus: We are avoiding that due to the 64 byte alignment requirement.

mtl1979: The 64-byte alignment requirement seems to be dictated by limitation of specific ABI, not generic restriction... On default ABI there is no such restriction and using x1 version should be somewhat slower and rely too much on compiler's ability to reorder code to avoid unnecessary pipeline stalls and stack spills.

As mentioned elsewhere, the other non-ABI related issues with x4 intrinsics were caused by using incompatible combination of compiler command-line options. As such, fix for them warrants own PR.

KungFuJesus:
@neheb could you check to see if my iOS check fixes your issue?  I don't have a cross compilation setup to test this sort of thing, it'd be good to know.

I had half a mind to force 32 byte alignment unconditionally for the non-copying variant of the checksum.  Then we actually could actually not use the fallback 4x loads.  But this would probably make a lot of small checksums really slow.

zbuild.h
KungFuJesus: I mean, the x4 load improvement is quite minor to begin with but if we're going to have 2 separate function paths anyway, I don't see much real harm.  I do find it funny that coderabbit tried to suggest adding it back, claiming regurgitated benefits from whatever lore it may have read from stack overflow, but then failed to do it because its copy of clang to verify that claims didn't support it.  I've added it back and injected what I hope is a good workaround for detecting iOS.  That should be all the more that's needed.

We could instead enforce 32 byte alignment in all scenarios but then we're going to have 1-31 byte checksums done a byte at a time which is less than ideal.  The alignment benefits everywhere are touted to be transparent and automatically realized by many ARM SIPs, yet the x2,3,&4 variants of this instruction still support encoding an alignment hint into the instruction so...clearly something wants that. In any case clang and GCC will not ever put anything there when targeting Linux, even with an alignment suggestion (__builtin_assume_aligned).

arch/arm/neon_intrins.h
```diff
@@ -25,8 +25,12 @@
     out.val[3] = vqsubq_u16(a.val[3], b); \
 } while (0)
 
-#  if defined(__clang__) && defined(__arm__) && defined(__ANDROID__)
-/* Clang for 32-bit Android has too strict alignment requirement (:256) for x4 NEON intrinsics */
+#if (defined(TARGET_OS_IPHONE) || defined(__IPHONE_OS_VERSION_MIN_REQUIRED))
```
Un1q32: `TARGET_OS_IPHONE` is defined when building for any Darwin platform, it is defined to 1 when building for iOS and 0 when building for any other Darwin system.

KungFuJesus: Ah, should be fixed now.

zbuild.h
mtl1979: My knowledge of ARM architecture is pretty much limited to compilers for embedded programming and that implies utilizing limited set of instructions... As I've mentioned before, sometimes we need to have to find middle ground between using as many registers (in a single instruction or code block) as possible and avoiding stalls. This pretty much means we should in such cases benchmark 1x, 2x and 4x variations, due to us deducing that 3x variants cause more trouble than have benefits. Compilers will almost always try to reorder the code if there is chance of performance penalty and the optimisation level prefers speed over size.

What works for 32-bit ARM might not be the best option for 64-bit ARM and that was one of the reasons I didn't back then want to unify the code even though a lot of code was shared between 32-bit and 64-bit versions.

arch/arm/adler32_neon.c
```diff
@@ -171,12 +303,33 @@ Z_INTERNAL uint32_t adler32_neon(uint32_t adler, const uint8_t *buf, size_t len)
...
-    unsigned int align_offset = ((uintptr_t)buf & 15);
-    unsigned int align_adj = (align_offset) ? 16 - align_offset : 0;
+     * though there's no explicit aligned load instruction. Note:
+     * on Android and iOS, their ABIs specify stricter alignment
+     * requirements for the 2,3,4x register ld1 variants. Clang for
+     * these platforms emits an alignment hint in the instruction for exactly
+     * 256 bits. Several ARM SIPs have small penalties for cacheline
+     * crossing loads as well (so really 512 bits is the optimal alignment
+     * of the buffer). 32 bytes should strike a balance, though. Clang and
+     * GCC on linux will not emit this hint in the encoded instruction and
```
mtl1979: Just a nitpick, but "linux" as a generic term kind of implies also any other operating system based on Linux... For clarity, it should be always capitalized.

arch/arm/adler32_neon.c
```diff
@@ -10,17 +10,142 @@
...
+        uint8x16_t d0 = vld1q_u8(buf);
+        uint8x16_t d1 = vld1q_u8(buf + 16);
+        uint8x16_t d2 = vld1q_u8(buf + 32);
+        uint8x16_t d3 = vld1q_u8(buf + 48);
+
+        vst1q_u8(dst, d0);
+        vst1q_u8(dst + 16, d1);
+        vst1q_u8(dst + 32, d2);
+        vst1q_u8(dst + 48, d3);
+        //vst1q_u8_x4(dst, d0_d3);
```
mtl1979: This commented line should be removed to avoid confusion. As we have two versions of the function, we don't need to preserve it for historical purposes.

neheb:
> neheb

as is, this PR fixes runtime on Alpine Linux + clang.

Dead2 (approved):
LGTM

Un1q32:
Hey, this seems to not be needed with Clang 20.  Can anyone else confirm?

neheb:
correct.

I just tested versions 2.2.4 and 2.2.3. Both succeed.

PR #1872, opened 2025-02-16
Explicit SSE2 vectorization of Chorba CRC method

KungFuJesus:
The version that's currently in the generic implementation for 32768 byte buffers leverages the stack. It manages to autovectorize but unfortunately the trips to the stack hurt its performance for CPUs which need this the most. This version is explicitly SIMD vectorized and doesn't use trips to the stack.  In my testing it's ~10% faster than the "small" variant, and about 42% faster than the "32768" variant.

## Summary by CodeRabbit

- **New Features**
  - Introduced an optimized CRC32 routine that leverages advanced SIMD (SSE2) instructions for improved performance on modern x86 systems.
  - Added support for a new CRC32 calculation method, `crc32_chorba_sse2`, enhancing efficiency for larger datasets.
  - Updated the build configuration to include the new `chorba_sse2` functionality across different environments.

- **Tests**
  - Added new benchmarks and test cases to validate the performance and reliability of the optimized CRC32 functionality.
  - Introduced a benchmark for the `chorba_sse2` implementation.
  - Added a test case for the `chorba_sse2` CRC32 functionality.
  - Added a new function declaration for `crc32_braid_internal` to enhance CRC32 functionality.


functable.c
```diff
@@ -82,6 +82,7 @@ static void init_functable(void) {
         ft.longest_match = &longest_match_sse2;
         ft.longest_match_slow = &longest_match_slow_sse2;
 #  endif
+        ft.crc32 = &crc32_chorba_sse;
```
Dead2: **nitpick:**  This should be on line 78, after chunksize ;)

configure
```diff
@@ -1570,8 +1570,8 @@ case "${ARCH}" in
             if test ${HAVE_SSE2_INTRIN} -eq 1; then
                 CFLAGS="${CFLAGS} -DX86_SSE2"
                 SFLAGS="${SFLAGS} -DX86_SSE2"
```
phprus, line 1600: Missing:
``-DX86_CHORBA_SSE_CRC``

arch/x86/x86_functions.h
```diff
@@ -89,6 +92,8 @@ uint32_t crc32_vpclmulqdq(uint32_t crc32, const uint8_t *buf, size_t len);
 #      define native_longest_match longest_match_sse2
 #      undef native_longest_match_slow
 #      define native_longest_match_slow longest_match_slow_sse2
+#      undef native_crc32
+#      define native_crc32 crc32_chorba_sse
```
phprus:
```c
#ifdef X86_CHORBA_SSE_CRC
```
Because the function is defined only if macro ``X86_CHORBA_SSE_CRC`` is defined.

functable.c
phprus:
```c
#ifdef X86_CHORBA_SSE_CRC
```
Because the function is defined only if macro ``X86_CHORBA_SSE_CRC`` is defined.

test/benchmarks/benchmark_crc32.cc
```diff
@@ -58,6 +58,7 @@ class crc32: public benchmark::Fixture {
 
 BENCHMARK_CRC32(braid, PREFIX(crc32_braid), 1);
 BENCHMARK_CRC32(generic, PREFIX(crc32_c), 1);
+BENCHMARK_CRC32(chorba_sse, PREFIX(crc32_chorba_sse), 1);
```
phprus:
```c
#ifdef X86_CHORBA_SSE_CRC
```
Because the function is defined only if macro ``X86_CHORBA_SSE_CRC`` is defined.

arch/x86/chorba_sse2.c
```diff
@@ -0,0 +1,835 @@
+#ifdef X86_CHORBA_SSE_CRC
+
+#include "zbuild.h"
+#include "crc32_braid_p.h"
+#include "crc32_braid_tbl.h"
+#include "crc32_c.h"
+#include <emmintrin.h>
+#include <smmintrin.h>
+#include <x86_intrins.h>
```
phprus: CI error:
```
/home/runner/work/zlib-ng/zlib-ng/arch/x86/chorba_sse2.c:9:10: error: 'x86_intrins.h' file not found with <angled> include; use "quotes" instead
#include <x86_intrins.h>
         ^~~~~~~~~~~~~~~
         "x86_intrins.h"
1 error generated.
```

arch/x86/x86_functions.h
```diff
@@ -89,6 +92,10 @@ uint32_t crc32_vpclmulqdq(uint32_t crc32, const uint8_t *buf, size_t len);
 #      define native_longest_match longest_match_sse2
 #      undef native_longest_match_slow
 #      define native_longest_match_slow longest_match_slow_sse2
+#      undef native_crc32
+#      ifdef X86_CHORBA_SSE_CRC
+#          define native_crc32 crc32_chorba_sse
+#      endif
```
phprus:
```c
#      ifdef X86_CHORBA_SSE_CRC
#          undef native_crc32
#          define native_crc32 crc32_chorba_sse
#      endif
```

arch/x86/chorba_sse2.c
```diff
@@ -0,0 +1,836 @@
...
+    final[2] ^= next3;
+    final[3] ^= next4;
+    final[4] ^= next5;
+
+    crc = crc32_braid_internal(crc, (uint8_t*) final, len-i);
+
+    return crc;
+}
+
+Z_INTERNAL uint32_t PREFIX(crc32_chorba_sse)(uint32_t crc, const uint8_t *buf, size_t len) {
```
phprus: ``crc32_chorba_sse`` or  ``PREFIX(crc32_chorba_sse)``?

In some places ``crc32_chorba_sse`` is used, in others ``PREFIX(crc32_chorba_sse)``.

KungFuJesus: Yeah the prefix bit isn't needed, we aren't exporting it to the api

arch/x86/x86_intrins.h
```diff
@@ -89,4 +89,31 @@ static inline __m512i _mm512_zextsi128_si512(__m128i a) {
...
+#if !defined(_M_AMD64)
+static inline int64_t _mm_cvtsi128_si64x(__m128i a) {
+    return ((int64_t*)&a)[0];
+}
+#define _mm_cvtsi64x_si128(a) _mm_set_epi64x(0, a)
+#endif
+#endif
+
+
+#if defined(__clang__)
```
phprus, line 114:
```c
#if defined(__clang__) && !defined(_MSC_VER)
```

Error: https://github.com/zlib-ng/zlib-ng/actions/runs/13362966244/job/37315662228?pr=1872

arch/x86/x86_intrins.h
```diff
@@ -89,4 +89,31 @@ static inline __m512i _mm512_zextsi128_si512(__m128i a) {
...
+#define _mm_cvtsi64x_si128(v) _mm_cvtsi64_si128(v)
+#define _mm_cvtsi128_si64x(v) _mm_cvtsi128_si64(v)
+#endif
+
+/* For some reason GCC doesn't define these intrinsics for 32 bit */
+#if defined(__GNUC__) && !defined( __x86_64__)
```
phprus:
```c
#if defined(__GNUC__) && !defined( __x86_64__) && !defined(__clang__)
```

Because clang define macro ``__GNUC__``.

CMakeLists.txt
```diff
@@ -912,7 +912,7 @@ if(WITH_OPTIM)
             check_sse2_intrinsics()
             if(HAVE_SSE2_INTRIN)
                 add_definitions(-DX86_SSE2)
-                set(SSE2_SRCS ${ARCHDIR}/chunkset_sse2.c ${ARCHDIR}/compare256_sse2.c ${ARCHDIR}/slide_hash_sse2.c)
+                set(SSE2_SRCS ${ARCHDIR}/chunkset_sse2.c ${ARCHDIR}/chorba_sse2.c ${ARCHDIR}/compare256_sse2.c ${ARCHDIR}/slide_hash_sse2.c)
```
phprus, line 941: Missing:

```cmake
add_definitions(-DX86_CHORBA_SSE_CRC)
```

test/benchmarks/benchmark_crc32.cc
```diff
@@ -59,6 +59,10 @@ class crc32: public benchmark::Fixture {
 BENCHMARK_CRC32(braid, PREFIX(crc32_braid), 1);
 BENCHMARK_CRC32(generic, PREFIX(crc32_c), 1);
 
+#ifdef X86_CHORBA_SSE_CRC
+BENCHMARK_CRC32(chorba_sse, PREFIX(crc32_chorba_sse), 1);
```
phprus: Remove ``PREFIX()``

win32/Makefile.msc
```diff
@@ -60,6 +61,7 @@ OBJS = \
 	chunkset_avx2.obj \
 	chunkset_sse2.obj \
 	chunkset_ssse3.obj \
+    chorba_sse2.obj \
```
phprus, line 63: Spaces vs tabs

Dead2:
I am thinking this should not introduce `X86_CHORBA_SSE_CRC`, but instead check for `WITHOUT_CHORBA` and `X86_SSE2`.
Having feature enabled/disabled messages in CMake output is still nice though.

This is more in line with what the other optimizations do, makes it default enabled, and also cuts down on the number of possible configurations.

arch/x86/chorba_sse2.c
```diff
@@ -0,0 +1,836 @@
...
+        a = _mm_load_si128((__m128i*)(in + off / sizeof(uint64_t))); \
+        b = _mm_load_si128((__m128i*)(in + off / sizeof(uint64_t) + 2)); \
+        } while (0);
+
+#define NEXT_ROUND(invec, a, b, c, d) do { \
+        a = _mm_xor_si128(_mm_slli_epi64(invec, 17), _mm_slli_epi64(invec, 55)); \
+        b = _mm_xor_si128(_mm_xor_si128(_mm_srli_epi64(invec, 47), _mm_srli_epi64(invec, 9)), _mm_slli_epi64(invec, 19)); \
+        c = _mm_xor_si128(_mm_srli_epi64(invec, 45), _mm_slli_epi64(invec, 44)); \
+        d  = _mm_srli_epi64(invec, 20); \
+        } while (0);
```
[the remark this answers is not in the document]
KungFuJesus: I think the names themselves are pretty self explanatory, as are the commented out scalar versions of the code that remain.

KungFuJesus:
Ugh no idea how this is segfaulting on this test on this very particular msvc and no other. Something tells me it might be a stack alignment issue induced by the compiler itself, though...

mtl1979:
> Ugh no idea how this is segfaulting on this test on this very particular msvc and no other. Something tells me it might be a stack alignment issue induced by the compiler itself, though...

Fails on line: 
```
a4_ = _mm_unpacklo_epi64(next56, ab4);
```

Looking at disassembly, it is indeed trying to load from stack, but as release code doesn't show all variables, it's not apparent why it thinks the address isn't aligned when it clearly is when I calculate using the register contents. The faulting instruction in disassembler is `pxor`, so to me it's clear that it has combined and reordered some intrinsics.

KungFuJesus:
ok, so, is this yet another case of an assumed align load from the stack when it's not?  It sounds like it's calling the memory-register variant of the function and assuming the load is aligned (a requirement for doing that), so I guess it's a question of whether or not we generated that address or MSVC did.  My _bet_ is that visual studio did and this is yet another compiler bug. The ab4 and next56 are register scoped variables so if it is that instruction, this is MSVC generating a stack spill and failing to properly align the address yet again. 

Do we file a ticket with Microsoft?  Some wider testing would be appreciated here, it's definitely a worthwhile win on the potato Core 2 Quad (and really this should be a win as far back as pentium 4s).

arch/x86/x86_functions.h
```diff
@@ -89,6 +91,11 @@ uint32_t crc32_vpclmulqdq(uint32_t crc32, const uint8_t *buf, size_t len);
 #      define native_longest_match longest_match_sse2
 #      undef native_longest_match_slow
 #      define native_longest_match_slow longest_match_slow_sse2
+#      undef native_crc32
+#      ifndef WITHOUT_CHORBA
+#          undef native_crc32
```
phprus: Duplicaded ``undef``.
Only the second is needed.

phprus:
I think this is a bug in MSVC.

The 32-bit architecture contains only 8 SSE registers (the 64-bit architecture contains 16 registers) and this can trigger a bug in MSVC.

arch/x86/chorba_sse2.c
```diff
@@ -0,0 +1,853 @@
...
+    final[2] ^= next3;
+    final[3] ^= next4;
+    final[4] ^= next5;
+
+    crc = crc32_braid_internal(crc, (uint8_t*) final, len-i);
+
+    return crc;
+}
+
+Z_INTERNAL uint32_t crc32_chorba_sse(uint32_t crc, const uint8_t *buf, size_t len) {
```
ccawley2011:
```suggestion
Z_INTERNAL uint32_t crc32_chorba_sse2(uint32_t crc, const uint8_t *buf, size_t len) {
```

arch/x86/chorba_sse2.c
```diff
@@ -0,0 +1,853 @@
...
+        } while (0);
+
+#define NEXT_ROUND(invec, a, b, c, d) do { \
+        a = _mm_xor_si128(_mm_slli_epi64(invec, 17), _mm_slli_epi64(invec, 55)); \
+        b = _mm_xor_si128(_mm_xor_si128(_mm_srli_epi64(invec, 47), _mm_srli_epi64(invec, 9)), _mm_slli_epi64(invec, 19)); \
+        c = _mm_xor_si128(_mm_srli_epi64(invec, 45), _mm_slli_epi64(invec, 44)); \
+        d  = _mm_srli_epi64(invec, 20); \
+        } while (0);
+
+Z_INTERNAL uint32_t chorba_small_nondestructive_sse(uint32_t crc, const uint64_t* buf, size_t len) {
```
ccawley2011:
```suggestion
Z_INTERNAL uint32_t chorba_small_nondestructive_sse2(uint32_t crc, const uint64_t* buf, size_t len) {
```

arch/x86/chorba_sse2.c
```diff
@@ -0,0 +1,853 @@
+#if !defined(WITHOUT_CHORBA) && (defined(X86_SSE2) || defined(X86_NOCHECK_SSE2))
```
phprus: We do not need use ``X86_NOCHECK_SSE2`` macro outside of runtime/compile-time dispatcher.
Checking the ``X86_SSE2`` macro is enough.

test/benchmarks/benchmark_crc32.cc
```diff
@@ -58,6 +58,9 @@ class crc32: public benchmark::Fixture {
 
 #ifndef WITHOUT_CHORBA
 BENCHMARK_CRC32(generic_chorba, crc32_c, 1);
+#   if defined(X86_SSE2) || defined(X86_NOCHECK_SSE2)
+    BENCHMARK_CRC32(chorba_sse, crc32_chorba_sse, 1);
+#   endif
```
phprus: Suggestion:
```c
#   if defined(X86_SSE2)
    BENCHMARK_CRC32(chorba_sse, crc32_chorba_sse, test_cpu_features.x86.has_sse2);
#   endif
```

test/test_crc32.cc
```diff
@@ -267,5 +267,8 @@ TEST_CRC32(pclmulqdq, crc32_pclmulqdq, test_cpu_features.x86.has_pclmulqdq)
 #ifdef X86_VPCLMULQDQ_CRC
 TEST_CRC32(vpclmulqdq, crc32_vpclmulqdq, (test_cpu_features.x86.has_pclmulqdq && test_cpu_features.x86.has_avx512_common && test_cpu_features.x86.has_vpclmulqdq))
 #endif
+#if !defined(WITHOUT_CHORBA) && (defined(X86_SSE2) || defined(X86_NOCHECK_SSE2))
```
phprus: Suggestion:
```c
#if !defined(WITHOUT_CHORBA) && defined(X86_SSE2)
```

arch/x86/chorba_sse2.c
```diff
@@ -0,0 +1,853 @@
...
+
+#include "zbuild.h"
+#include "crc32_braid_p.h"
+#include "crc32_braid_tbl.h"
+#include "crc32.h"
+#include <emmintrin.h>
+#include <immintrin.h>
+#include <smmintrin.h>
```
ccawley2011: Is it necessary to have the headers for later SSE versions?

KungFuJesus: Sorry, was trying to figure out what was going on with cvtsi64x_si128 et al.  Evidently they just remove them from 32 bit compilations to make our lives more difficult.

arch/x86/chorba_sse2.c
```diff
@@ -0,0 +1,852 @@
...
+}
+
+Z_INTERNAL uint32_t crc32_chorba_sse2(uint32_t crc, const uint8_t *buf, size_t len) {
+    uint32_t c;
+    uint64_t* aligned_buf;
+    size_t aligned_len;
+
+    c = (~crc) & 0xffffffff;
+    unsigned long algn_diff = ((uintptr_t)16 - ((uintptr_t)buf & 15)) & 15;
+    if (algn_diff < len) {
```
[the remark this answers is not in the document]
KungFuJesus: The arithmetic is bound to values between 16 and 1, it can't overflow.

arch/x86/chorba_sse2.c
```diff
@@ -0,0 +1,836 @@
...
+#include <emmintrin.h>
+#include <immintrin.h>
+#include <smmintrin.h>
+#include "x86_intrins.h"
+
+#define bitbuffersizebytes (16 * 1024 * sizeof(z_word_t))
+#define bitbuffersizezwords (bitbuffersizebytes / sizeof(z_word_t))
+#define bitbuffersizeqwords (bitbuffersizebytes / sizeof(uint64_t))
+
+extern uint32_t crc32_braid_base(uint32_t c, const uint8_t *buf, size_t len);
```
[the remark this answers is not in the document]
KungFuJesus: This header is the same directory, we do not need to do a path relative to the compilation root.  The name of the function is also pretty self explanatory.

arch/x86/chorba_sse2.c
```diff
@@ -0,0 +1,836 @@
...
+        next34 = _mm_xor_si128(b4c4, cd3);
+        next34 = _mm_xor_si128(next34, d2_);
+        next56 = _mm_unpackhi_epi64(cd4, _mm_setzero_si128());
+    }
+
+    next1 = _mm_cvtsi128_si64(next12);
+    next2 = _mm_cvtsi128_si64(_mm_unpackhi_epi64(next12, next12));
+    next3 = _mm_cvtsi128_si64(next34);
+    next4 = _mm_cvtsi128_si64(_mm_unpackhi_epi64(next34, next34));
+    next5 = _mm_cvtsi128_si64(next56);
```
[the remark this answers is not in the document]
KungFuJesus: _mm_extract_epi64 is not available in SSE2.  Duh.  Also extract takes more cycles than the cvtsi128_si64 generated movq.

arch/x86/chorba_sse2.c
```diff
@@ -0,0 +1,853 @@
...
+        a = _mm_load_si128((__m128i*)(in + off / sizeof(uint64_t))); \
+        b = _mm_load_si128((__m128i*)(in + off / sizeof(uint64_t) + 2)); \
+        } while (0);
+
+#define NEXT_ROUND(invec, a, b, c, d) do { \
+        a = _mm_xor_si128(_mm_slli_epi64(invec, 17), _mm_slli_epi64(invec, 55)); \
+        b = _mm_xor_si128(_mm_xor_si128(_mm_srli_epi64(invec, 47), _mm_srli_epi64(invec, 9)), _mm_slli_epi64(invec, 19)); \
+        c = _mm_xor_si128(_mm_srli_epi64(invec, 45), _mm_slli_epi64(invec, 44)); \
+        d  = _mm_srli_epi64(invec, 20); \
+        } while (0);
```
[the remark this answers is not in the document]
KungFuJesus: I had this to debug early on.  The code that calls this ensures alignment, this should no longer be needed.

arch/x86/chorba_sse2.c
```diff
@@ -0,0 +1,853 @@
...
+
+Z_INTERNAL uint32_t chorba_small_nondestructive_sse(uint32_t crc, const uint64_t* buf, size_t len) {
+    const uint64_t* input = buf;
+    ALIGNED_(16) uint64_t final[9] = {0};
+    uint64_t next1 = crc;
```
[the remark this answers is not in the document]
KungFuJesus: The input to this function is forced to be a modulus of 16, which itself is already an even modulus of 8.  This is not needed.

arch/x86/chorba_sse2.c
```diff
@@ -0,0 +1,852 @@
...
+#define NEXT_ROUND(invec, a, b, c, d) do { \
+        a = _mm_xor_si128(_mm_slli_epi64(invec, 17), _mm_slli_epi64(invec, 55)); \
+        b = _mm_xor_si128(_mm_xor_si128(_mm_srli_epi64(invec, 47), _mm_srli_epi64(invec, 9)), _mm_slli_epi64(invec, 19)); \
+        c = _mm_xor_si128(_mm_srli_epi64(invec, 45), _mm_slli_epi64(invec, 44)); \
+        d  = _mm_srli_epi64(invec, 20); \
+        } while (0);
+
+Z_INTERNAL uint32_t chorba_small_nondestructive_sse2(uint32_t crc, const uint64_t* buf, size_t len) {
+    const uint64_t* input = buf;
+    ALIGNED_(16) uint64_t final[9] = {0};
```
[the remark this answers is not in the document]
KungFuJesus: The calling function ensures this.

arch/x86/x86_intrins.h
```diff
@@ -89,4 +89,44 @@ static inline __m512i _mm512_zextsi128_si512(__m128i a) {
 #  undef _mm512_extracti32x4_epi32
 #  define _mm512_extracti32x4_epi32(v1, e1) _mm512_maskz_extracti32x4_epi32(UINT8_MAX, v1, e1)
 #endif
+
+#if defined(_MSC_VER) && !defined(__clang__)
+#include <intrin.h>
+/* For whatever reason this intrinsic is 64 it only with MSVC? Hopefully this generates
```
Dead2: **typo (if-minor):** `64 bit`?

arch/x86/x86_functions.h
```diff
@@ -6,6 +6,12 @@
 #ifndef X86_FUNCTIONS_H_
 #define X86_FUNCTIONS_H_
 
+/* So great news, your compiler is broken and causes stack smashing. Rather than
+ * notching out its compilation we'll just remove the assignment in the functable */
+#if defined(_MSC_VER) && !defined(_M_AMD64) && _MSC_VER >= 1920 && _MSC_VER <= 1929
+#define NO_CHORBA_SSE2
```
Dead2, line 14: **thought:** I would absolutely accept a `NO_CHORBA_SSE2` define for disabling it completely, so this can be done from CMake if you want to implement it. This should probably stay though, since it would avoid non-CMake builds like the VS-project makefiles from failing or having to always exlude it (not sure whether they can check stuff like this and do it conditionally).

arch/x86/x86_functions.h
```diff
@@ -89,6 +97,10 @@ uint32_t crc32_vpclmulqdq(uint32_t crc32, const uint8_t *buf, size_t len);
 #      define native_longest_match longest_match_sse2
 #      undef native_longest_match_slow
 #      define native_longest_match_slow longest_match_slow_sse2
+#      ifndef WITHOUT_CHORBA
```
Dead2: **issue (blocking):** This would need to check the `NO_CHORBA_SSE2` define as well, since it bypasses functable completely.

mtl1979:
> Do we file a ticket with Microsoft? Some wider testing would be appreciated here, it's definitely a worthwhile win on the potato Core 2 Quad (and really this should be a win as far back as pentium 4s).

I've filed a ticket with Microsoft... Everyone with Microsoft account can see, vote and comment on the ticket.

arch/x86/x86_functions.h
mtl1979: NMake can conditionally include/exclude files, but Visual Studio project files can't when created through CMake... However it's possible to manually write project file that excludes certain files from specific configurations.

Visual Studio project file generator in CMake doesn't support generator expressions in earlier CMake versions.

phprus: Maybe make sense add a link to the bug report in Visual Studio in comment?

Link: https://developercommunity.visualstudio.com/t/Stack-corruption-with-v142-toolchain-whe/10853479

KungFuJesus:
Out of morbid curiosity I decided to give this a whirl on a wider range of x86 hardware.  On my Haswell-U laptop this is marginally better than the vanilla scalar version that's autovectorizing but it's within spitting distance.  Here's on a Cascade Lake:

```
---------------------------------------------------------------------
Benchmark                           Time             CPU   Iterations
---------------------------------------------------------------------
crc32/generic_chorba/32768       4403 ns         4404 ns       158397
crc32/chorba_sse2/32768          4680 ns         4681 ns       149538
crc32/braid/32768               17278 ns        17282 ns        40489
crc32/pclmulqdq/32768            1384 ns         1384 ns       504698
crc32/vpclmulqdq/32768     ERROR OCCURRED: 'CPU does not support vpclmulqdq'
```
So basically the same, but this isn't who this is for, anything modern is going to use the carryless multiplication variant.

Here's on a Core 2 Quad Q9650:

```
---------------------------------------------------------------------
Benchmark                           Time             CPU   Iterations
---------------------------------------------------------------------
crc32/generic_chorba/32768      16171 ns        16170 ns        43272
crc32/chorba_sse2/32768         10215 ns        10213 ns        68667
crc32/braid/32768               15025 ns        15025 ns        46619
```

And finally, with a little bit of necromancy and seemingly only one working memory channel, a Northwood P4:

```
---------------------------------------------------------------------
Benchmark                           Time             CPU   Iterations
---------------------------------------------------------------------
crc32/generic_chorba/32768      70446 ns        70389 ns        10272
crc32/chorba_sse2/32768         36515 ns        36484 ns        19405
crc32/braid/32768              147004 ns       146829 ns         4560
crc32/pclmulqdq/32768      ERROR OCCURRED: 'CPU does not support pclmulqdq'
```

Err well, debian spawned on me unattended upgrades during this (and with only a single CPU core with HT, of course there's a huge penalty).  Here's a retake:

```
---------------------------------------------------------------------
Benchmark                           Time             CPU   Iterations
---------------------------------------------------------------------
crc32/generic_chorba/32768      32985 ns        32982 ns        21224
crc32/chorba_sse2/32768         27261 ns        27259 ns        25682
crc32/braid/32768               83003 ns        82991 ns         8330
crc32/pclmulqdq/32768      ERROR OCCURRED: 'CPU does not support pclmulqdq'
```

Not quite as dramatic but still pretty good.  With only half the vector registers we are spilling to the stack more.  But, we are benefiting from effectively the quadruple wide shifts with SSE2.  The ensured alignment helps, too (I suspect that's a lot of the gain on the Q9650).

arch/x86/chorba_sse2.c
```diff
@@ -0,0 +1,852 @@
...
+            c = crc32_braid_internal(c, (uint8_t*) aligned_buf, aligned_len);
+        }
+    }
+    else {
+        c = crc32_braid_internal(c, buf, len);
+    }
+
+    /* Return the CRC, post-conditioned. */
+    return c ^ 0xffffffff;
+}
```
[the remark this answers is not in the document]
KungFuJesus: We don't need this, the caller is already checking the bounds.  The magic number bit has been addressed.

arch/x86/x86_intrins.h
```diff
@@ -89,4 +89,44 @@ static inline __m512i _mm512_zextsi128_si512(__m128i a) {
...
+#if defined(__GNUC__) && !defined( __x86_64__) && !defined(__clang__)
+/* This may be sub-optimal, we can't move directly into a GPR but such
+ * is the 32 bit life */
+static inline int64_t _mm_cvtsi128_si64x(__m128i a) {
+    union { __m128i v; int64_t i; } u;
+    u.v = a;
+    return u.i;
+}
+#define _mm_cvtsi64x_si128(a) _mm_set_epi64x(0, a)
+#endif
```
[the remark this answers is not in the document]
KungFuJesus: We are looking out for clang.  MSVC does not and probably never will define `__GNUC__`.

arch/x86/x86_intrins.h
```diff
@@ -89,4 +89,40 @@ static inline __m512i _mm512_zextsi128_si512(__m128i a) {
...
+   return _mm_set_epi64x(0, a);
+}
+#endif
+#endif
+
+
+#if defined(__clang__)
+#define _mm_cvtsi64x_si128(v) _mm_cvtsi64_si128(v)
+#define _mm_cvtsi128_si64x(v) _mm_cvtsi128_si64(v)
+#endif
```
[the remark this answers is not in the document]
KungFuJesus: I think this is safe, regardless.

arch/x86/x86_intrins.h
```diff
@@ -89,4 +89,40 @@ static inline __m512i _mm512_zextsi128_si512(__m128i a) {
...
+#if !defined(_M_AMD64)
+/* So, while we can't move directly to a GPR, hopefully this move to
+ * a stack resident variable doesn't equate to something awful */
+static inline int64_t _mm_cvtsi128_si64x(__m128i a) {
+    union { __m128i v; int64_t i; } u;
+    u.v = a;
+    return u.i;
+}
+
+static inline __m128i _mm_cvtsi64x_si128(int64_t a) {
```
mtl1979, line 107: v142 toolchain can't handle inlining this function... It works without `inline` but compiles to really slow code. I tried converting it to preprocessor macro, but that didn't work either.

Only code that I could get to work when inlined was:

```
   ALIGNED_(16) int64_t res[2] = { a, 0 };
   return _mm_load_si128((__m128i*)res);
```

The elements need to be backwards when spilled to stack, the intrinsic did automatically swap the element order.

KungFuJesus: Any word yet if and when Microsoft will actually fix this?

mtl1979: Microsoft closed the bug report as WONTFIX due to Visual Studio 2019 being out of support. I guess the best solution is to use the workaround I posted above if we want to enable SSE2 version of Chorba CRC for all Windows users.

Dead2 (approved):
Although we found this to be a little slower than the generic Chorba on cpus with double-pumped SSE2 (Like Zacate for example), it is however good for the rest of the SSE2+.

KungFuJesus:
I just removed an extra function call or two (one to memcpy, the other to braid).  It was faster for me on modern hardware, I suspect it'll be the same on older.

KungFuJesus:
> Although we found this to be a little slower than the generic Chorba on cpus with double-pumped SSE2 (Like Zacate for example), it is however good for the rest of the SSE2+.

Did we want to try to notch out that variant when assigning from the function table or did we just want to issue that caveat so those who are bit by that can build without it?

Dead2:
> > Although we found this to be a little slower than the generic Chorba on cpus with double-pumped SSE2 (Like Zacate for example), it is however good for the rest of the SSE2+.
> 
> Did we want to try to notch out that variant when assigning from the function table or did we just want to issue that caveat so those who are bit by that can build without it?

I don't think we need to bother with that really. First of all it is an extremely narrow userbase, quite possibly zero users will use zlib-ng on this really old cpu. Also it is not that much slower, so combined with the rest of the SSE2 optimizations it is nevertheless a solid win.
So adding complexity to cpuid checking for enabling only partial SSE2, for something that very few or possibly nobody will benefit from, I don't think it is actually worth it.

PR #1882, opened 2025-03-06
fix the url of the s390x actions worker patch

fneddy:
gaplib changed their patch names. This fixes the path and pins the patch to a specific git hash instead of the main branch.

## Summary by CodeRabbit

- **Chores**
  - Updated the source URL for a key patch, which may enhance build stability and improve actions runner functionality.

### DIFF
--- a/arch/s390/self-hosted-builder/actions-runner.Dockerfile
+++ b/arch/s390/self-hosted-builder/actions-runner.Dockerfile
@@ -15,7 +15,7 @@ RUN     cd /tmp && \
         git clone -q https://github.com/actions/runner && \
         cd runner && \
         git checkout $(git describe --tags $(git rev-list --tags --max-count=1)) -b build && \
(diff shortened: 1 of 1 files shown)


Dead2:
Name change is a good fix, but I am not so sure about locking it to a commit. It was locked previously and it masked a problem for a long time.

If we could lock actions runner to a specific version, then this could perhaps have been useful, but unfortunately it keeps changing and the updated runner is required to be implemented within a couple weeks (or less if they deem a change important) to keep the runner activated with github actions. Also, making a PR in zlib-ng every time gaplib changes their patch because the runner has introduced an incompatible change is not very useful to us.

To ensure that we actually detect breaking changes (like with the recent filename change), the gaplib docker is rebuilt by cron every week, so hopefully we detect the problem sooner than we did previously. This way we only notice when something that cannot be automatically resolved breaks.

Dead2 (approved):

PR #1884, opened 2025-03-07
Disable MSVC optimizations for AVX512 GET_CHUNK_MAG

nmoinvaz:
MSVC compiler (VS 17.11.x) incorrectly optimizes the GET_CHUNK_MAG code on older versions. Appears to be resolved in VS 17.13.2. The compiler would optimize the code in such a way that it would cause a decompression failure. It only happens when /Os flag is set.

## Summary by CodeRabbit

- **Bug Fixes**
  - Addressed a compiler optimization issue affecting builds with older versions of Microsoft Visual Studio, ensuring a more reliable and stable build process without changing the application's functionality.

### DIFF
--- a/arch/x86/chunkset_avx512.c
+++ b/arch/x86/chunkset_avx512.c
@@ -98,6 +98,10 @@ static inline uint8_t* CHUNKCOPY(uint8_t *out, uint8_t const *from, unsigned len
     return out;
 }
 
+/* MSVC compiler decompression bug when optimizing for size */
+#if defined(_MSC_VER) && _MSC_VER < 1943
+#  pragma optimize("", off)
+#endif
 static inline chunk_t GET_CHUNK_MAG(uint8_t *buf, uint32_t *chunk_rem, uint32_t dist) {
     lut_rem_pair lut_rem = perm_idx_lut[dist - 3];
     __m256i ret_vec;
@@ -128,6 +132,9 @@ static inline chunk_t GET_CHUNK_MAG(uint8_t *buf, uint32_t *chunk_rem, uint32_t 
 
     return ret_vec;
 }
+#if defined(_MSC_VER) && _MSC_VER < 1943
+#  pragma optimize("", on)
+#endif
 
 static inline void storehalfchunk(uint8_t *out, halfchunk_t *chunk) {
     _mm_storeu_si128((__m128i *)out, *chunk);


arch/x86/chunkset_avx512.c
```diff
@@ -98,6 +98,10 @@ static inline uint8_t* CHUNKCOPY(uint8_t *out, uint8_t const *from, unsigned len
     return out;
 }
 
+/* MSVC compiler decompression bug when optimizing for size */
+#if defined(_MSC_VER) && _MSC_VER < 1943
+#  pragma optimize("", off)
```
mtl1979, line 103: Does it still work if the first parameter is `"t"`, as it would not disable optimizations when `/Ot` is enabled.

nmoinvaz:
@mtl1979 I tried it but it doesn’t work right.

arch/x86/chunkset_avx512.c
nmoinvaz: I tried it but it does not work right.

mtl1979: `/Ot` optimizes for speed, so the comment is then kind of inaccurate...

KungFuJesus:
I'm wondering if we could find a lower bound for the bug, too. It'd be great to bookend this check between versions.

arch/x86/chunkset_avx512.c
nmoinvaz: Don't quite understand your last comment. That optimization string argument only works if you are starting from enabling optimization, not starting from disabling optimization.

nmoinvaz:
That is a lot of work. There is no easy way to switch back to an older version of MSVC like there is with Linux packages.

arch/x86/chunkset_avx512.c
mtl1979: `/Ot` and `/Os` are mutually exclusive, however both also enable for example `/Og` (common sub-expression elimination). With current patch, it will disable all optimizations and thus degrade performance even if optimizing for speed.

Basically what I was saying was, that disable `/Os` and enable `/Ot` instead... might need two pragma lines to do that...

Something like:

```
#  pragma optimize("s", off)
#  pragma optimize("t", on)
```

mtl1979:
> That is a lot of work. There is no easy way to switch back to an older version of MSVC like there is with Linux packages.

We could try to clamp it to specific toolchain. As said reverting to older build of same toolchain version is too much hassle. If both v142 and v143 have the bug then we don't need to worry about v141 or older as those are out of support too. Not sure which toolchain introduced AVX512 support as my CPU does not support it.

Visual Studio Installer can install multiple builds of same toolchain but I don't know how CMake can switch between them.

phprus:
Specify the toolset version:

```
cmake ... -T "v142,version=14.29.30133"
```

https://cmake.org/cmake/help/v3.25/variable/CMAKE_GENERATOR_TOOLSET.html

mtl1979:
> Specify the toolset version:
> 
> ```
> cmake ... -T "v142,version=14.29.30133"
> ```
> 
> https://cmake.org/cmake/help/v3.25/variable/CMAKE_GENERATOR_TOOLSET.html

That works once per build directory... Need to use different directory or remove the cache and generated files before switching to another toolchain version/build...

nmoinvaz:
https://x.com/i/grok/share/XwMErXewB6EjJK3X2owak0A6U

I don't really have plans to check older versions as this change is limited to one tiny function. If somebody else wants to do it, I would appreciate it, but I don't think it's really necessary.

Dead2 (approved):

pps83:
@nmoinvaz I've seen ms compiler fail in in similar cases with heavy intrisic use. When I tried to debug them, I noticed that by placing some printf's such issues often evaporated. After careful "optimization" of such "fix", it looks like often placing a `__nop();` somewhere in the middle of the function for affected versions of ms compiler is enough. Here's one such fix: https://github.com/abseil/abseil-cpp/commit/25fb731837188648e0726d223154ed3821465365

PR #1885, opened 2025-03-17
Add parameterized testing

Dead2:
Based on #1448 by @rhpvorderman.

This replaces the old `test-data.cmake` that partly does the same thing.
The old one tests 10 parameter combinations per input file, the new one by default tests 540 combinations.
In my local tests this still ends up with a much faster runtime, due to the very high overhead of cmake executing cmake for each parameter combination. The new test does it all internally.

Originally this would do more than 10k combination per input file, this would be prohibitively slow, therefore I adjusted this a bit by adding the `WITH_EXTRA_TESTS` define that enables it to actually run all the combinations. Currently this enables 9450 combinations per file.

Changes compared to #1448:
- Removed compression level -1, since it maps to 6 and is also tested.
- Removed wbits level 8, since it is and probably always has been broken in upstream zlib.
- Fixed some datatype mismatches.
- Added define `WITH_EXTRA_TESTS` to enable all 9450 tests, otherwise run the most important 540 tests.
- Added to CMake and Configure
- Added configurable max filesize, instead of 4KiB it now defaults to 128KiB or 4MiB depending on extra tests being enabled.
- Replaced `test-data.cmake` with code using this new tester, split out `test-detect.cmake` as this functionality is not replaced.

Also added a small unrelated "fix" for an annoyance while doing this, to stop running gbench as part of gtest. It can take minutes to complete benchmarking, but it hides the results so it is not so useful to run as a gtest. The benchmarks also do not verify correctness of output.

In total, this cuts down the time it takes to run the whole testsuite from ~100sec to ~30sec on Github CI (tests running in emulators are slower than this, but also complete much faster than before).
This cuts down the total time it takes to run all the github CI for a PR by a lot.

### Pros:
- Faster testing and CI.
- Tests more combinations of deflate settings, 540 instead of 10.

### Cons:
- Does not test all the corpora files completely, by default now only testing the first 128KiB.

## Summary by CodeRabbit

- **Tests**
	- Introduced a robust parameterized testing framework covering a wide range of compression and decompression configurations.
	- Added new test targets for diverse input cases, including an integrated help option to guide usage.
	- Removed an outdated benchmark test to streamline the testing suite.
- **Chores**
	- Updated configuration scripts to automatically trigger extended tests on supported platforms, enhancing overall quality assurance.

### DIFF
--- a/configure
+++ b/configure
@@ -2274,7 +2274,7 @@ sed < $SRCDIR/arch/generic/Makefile.in "
 # Emscripten does not support large amounts of data via stdin/out
 # https://github.com/emscripten-core/emscripten/issues/16755#issuecomment-1102732849
 if test "$CHOST" != "wasm32"; then
-    TEST="${TEST} ghtests"
+    TEST="${TEST} ghtests parameterized"
 fi
 
 # Determine emulator to use when running tests
--- a/test/benchmarks/CMakeLists.txt
+++ b/test/benchmarks/CMakeLists.txt
@@ -62,9 +62,6 @@ if(WIN32)
     target_link_libraries(benchmark_zlib shlwapi)
 endif()
 
-add_test(NAME benchmark_zlib
-    COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:benchmark_zlib>)
-
 if(WITH_BENCHMARK_APPS)
     option(BUILD_ALT_BENCH "Link against alternative zlib implementation" OFF)
 
--- a/test/parameterized_tester.c
+++ b/test/parameterized_tester.c
@@ -1,0 +1,194 @@
+/* parameterized_tester.c - Test deflate() and inflate() on a chunk of the
+ * input data file, combining many deflate() settings to improve testing coverage.
+ */
+#include "zbuild.h"
+#ifdef ZLIB_COMPAT
+#  include "zlib.h"
+#else
+#  include "zlib-ng.h"
+#endif
+
+#include <stdio.h>
+
+/*
+ * WITH_EXTRA_TESTS results in 9450 permutations per input file.
+ * Normal mode results in 540.
+ *
+ * WITH_EXTRA_TESTS also tests a bigger chunk of the files.
+ */
+
+#ifdef WITH_EXTRA_TESTS
+const int valid_wbits[] = {
+    -9, -10, -11, -12, -13, -14, -15,
+     9,  10,  11,  12,  13,  14,  15,
+#  ifdef WITH_GZFILEOP
+    25,  26,  27,  28,  29,  30,  31,
+#  endif
+};
+
+const int valid_compression_levels[] = {
+    0, 1, 2, 3, 4, 5, 6, 7, 8, 9
+};
+
+const int valid_mem_levels[] = {
+    1, 2, 3, 4, 5, 6, 7, 8, 9,
+};
+
+#define MAX_FILESIZE_KB 4096
+
+#  ifdef WITH_GZFILEOP
+#    define NUM_W 21
+#  else
+#    define NUM_W 14
+#  endif
+#  define NUM_C 10
+#  define NUM_M 9
+
+#else // !WITH_EXTRA_TESTS
+
+const int valid_wbits[] = {
+    -9, -15,
+     9,  15,
+#  ifdef WITH_GZFILEOP
+    25,  31,
+#  endif
+};
+
+const int valid_compression_levels[] = {
+    0, 1, 2, 4, 6, 9
+};
+
+const int valid_mem_levels[] = {
+    1, 8, 9,
+};
+
+#define MAX_FILESIZE_KB 128
+
+#  ifdef WITH_GZFILEOP
+#    define NUM_W 6
+#  else
+#    define NUM_W 4
+#  endif
+#  define NUM_C 6
+#  define NUM_M 3
+#endif
+
+#define MAX_FILESIZE MAX_FILESIZE_KB * 1024
+#define MAX_OUTSIZE MAX_FILESIZE_KB * (1024 + 128)
+
+const int valid_strategies[] = {
+    Z_FILTERED,
+    Z_HUFFMAN_ONLY,
+    Z_RLE,
+    Z_FIXED,
+    Z_DEFAULT_STRATEGY,
+};
+
+int main (int argc, char *argv[])  {
+    PREFIX3(stream) c_stream;
+    PREFIX3(stream) d_stream;
+    uint8_t data_in[MAX_FILESIZE];
+    uint8_t compressed_out[MAX_OUTSIZE];
+    uint8_t decompressed_out[MAX_OUTSIZE];
+    const char *error_format = "Error at %s\n"
+                               "level: %d\n"
+                               "wbits: %d\n"
+                               "memLevel: %d\n"
+                               "strategy: %d\n"
+                               "exit: %d\n"
+                               "message: %s\n";
+
+    if (argc != 2) {
+        fputs("A data input must be given as argument", stderr);
+        return EXIT_FAILURE;
+    }
+
+    if ((argc == 2 && strcmp(argv[1], "--help") == 0)) {
+        printf("Usage: test_parameterized <FILENAME>\n\n");
+        return 0;
+    }
+
+    FILE *in_file = fopen(argv[1], "r");
+    uint32_t read_size = (uint32_t)fread(data_in, 1, MAX_FILESIZE, in_file);
+    fclose(in_file);
+
+    if (!read_size) {
+        fprintf(stderr, "File is empty or error: %s", argv[1]);
+        return EXIT_FAILURE;
+    }
+
+    for (int wbits_index = 0; wbits_index < NUM_W; wbits_index += 1) {
+        for (int complevel_index = 0; complevel_index < NUM_C; complevel_index += 1) {
+            for (int memlevel_index = 0; memlevel_index < NUM_M; memlevel_index += 1) {
+                for (int strat_index = 0; strat_index < 5; strat_index += 1) {
+
+                    int level = valid_compression_levels[complevel_index];
+                    int wbits = valid_wbits[wbits_index];
+                    int memlevel  = valid_mem_levels[memlevel_index];
+                    int strategy = valid_strategies[strat_index];
+                    memset(&c_stream, 0, sizeof(PREFIX3(stream)));
+                    memset(&d_stream, 0, sizeof(PREFIX3(stream)));
+
+                    int err = PREFIX(deflateInit2)(&c_stream, level, Z_DEFLATED, wbits, memlevel, strategy);
+                    if (err != Z_OK) {
+                        fprintf(stderr, error_format, "deflateInit2", level, wbits, memlevel, strategy, c_stream.msg);
+                        return EXIT_FAILURE;
+                    }
+
+                    c_stream.next_in = data_in;
+                    c_stream.avail_in = read_size;
+                    c_stream.next_out = compressed_out;
+                    c_stream.avail_out = MAX_OUTSIZE;
+
+                    err = PREFIX(deflate)(&c_stream, Z_FINISH);
+                    if (err != Z_STREAM_END) {
+                        fprintf(stderr, error_format, "deflate", level, wbits, memlevel, strategy, c_stream.msg);
+                        return EXIT_FAILURE;
+                    }
+
+                    err = PREFIX(inflateInit2)(&d_stream, wbits);
+                    if (err != Z_OK) {
+                        fprintf(stderr, error_format, "inflateInit2", level, wbits, memlevel, strategy, err, d_stream.msg);
+                        return EXIT_FAILURE;
+                    }
+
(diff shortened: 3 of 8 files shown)


test/benchmarks/CMakeLists.txt
```diff
@@ -62,9 +62,6 @@ if(WIN32)
     target_link_libraries(benchmark_zlib shlwapi)
 endif()
 
-add_test(NAME benchmark_zlib
-    COMMAND ${CMAKE_CROSSCOMPILING_EMULATOR} $<TARGET_FILE:benchmark_zlib>)
-
```
phprus, line 67: Please tell me why you removed the ``benchmark_zlib`` cmake-test?

Dead2: It is in the PR description.

```
Also added a small unrelated "fix" for an annoyance while doing this, to stop running
gbench as part of gtest. It can take minutes to complete benchmarking, but it hides
the results so it is not so useful to run as a gtest.
```

Dead2: `benchmark_zlib` can of course still be run directly from the command line.

phprus: Thanks for your answer.

rhpvorderman:
Whoops, I see you asked me to rebase in https://github.com/zlib-ng/zlib-ng/pull/1448 and I didn't respond. That one slipped through. My apologies.

Dead2:
> Whoops, I see you asked me to rebase in #1448 and I didn't respond. That one slipped through. My apologies.

No worries, it needed quite a bit of  integration to the build system as well. 😄

Dead2:
Added configurable max filesize, instead of only testing up to 4KiB of data it now defaults to 128KiB or 4MiB depending on extra tests being enabled.

configure
```diff
@@ -2274,7 +2274,7 @@ sed < $SRCDIR/arch/generic/Makefile.in "
 # Emscripten does not support large amounts of data via stdin/out
 # https://github.com/emscripten-core/emscripten/issues/16755#issuecomment-1102732849
 if test "$CHOST" != "wasm32"; then
-    TEST="${TEST} ghtests"
```
nmoinvaz, line 2277: parametrized -> parameterized. I think it i miss spelled everywhere.

test/parametrized_tester.c
```diff
@@ -0,0 +1,194 @@
...
+};
+
+const int valid_mem_levels[] = {
+    1, 8, 9,
+};
+
+#define MAX_FILESIZE_KB 128
+
+# ifdef WITH_GZFILEOP
+#  define NUM_W 6
```
nmoinvaz: two space indent?

test/parametrized_tester.c
```diff
@@ -0,0 +1,194 @@
...
+# endif
+# define NUM_C 10
+# define NUM_M 9
+
+#else // !WITH_EXTRA_TESTS
+
+const int valid_wbits[] = {
+    -9, -15,
+     9,  15,
+# ifdef WITH_GZFILEOP
```
nmoinvaz: Two space indent?

test/parametrized_tester.c
```diff
@@ -0,0 +1,194 @@
...
+                               "strategy: %d\n"
+                               "exit: %d\n"
+                               "message: %s\n";
+
+    if (argc != 2) {
+        fputs("A data input must be given as argument", stderr);
+        return EXIT_FAILURE;
+    }
+
+    if ((argc == 2 && strcmp(argv[1], "--help") == 0)) {
```
nmoinvaz: No need for extra braces.

test/parametrized_tester.c
```diff
@@ -0,0 +1,194 @@
...
+    FILE *in_file = fopen(argv[1], "r");
+    uint32_t read_size = (uint32_t)fread(data_in, 1, MAX_FILESIZE, in_file);
+    fclose(in_file);
+
+    if (!read_size) {
+        fprintf(stderr, "File is empty or error: %s", argv[1]);
+        return EXIT_FAILURE;
+    }
+
+    for (int wbits_index=0; wbits_index < NUM_W; wbits_index+=1) {
```
nmoinvaz: Spaces between operators?

test/parametrized_tester.c
```diff
@@ -0,0 +1,194 @@
...
+};
+
+const int valid_mem_levels[] = {
+    1, 2, 3, 4, 5, 6, 7, 8, 9,
+};
+
+#define MAX_FILESIZE_KB 4096
+
+# ifdef WITH_GZFILEOP
+#  define NUM_W 21
```
nmoinvaz: NUM_W? Consider a better name? Same with NUM_C, NUM_M.

test/parametrized_tester.c
```diff
@@ -0,0 +1,194 @@
...
+const int valid_strategies[] = {
+    Z_FILTERED,
+    Z_HUFFMAN_ONLY,
+    Z_RLE,
+    Z_FIXED,
+    Z_DEFAULT_STRATEGY,
+};
+
+int main (int argc, char *argv[])  {
+    PREFIX3(stream) zst_deflate;
```
nmoinvaz: Can we use `d_stream` and `c_stream` like elsewhere in our example code?

test/parametrized_tester.c
Dead2: I did not like these names, but I couldn't come up with any that I liked. Ideas?

Dead2:
As a few datapoints for the time it takes to run the CI testing itself (setup/build times not included):

`EL9 Clang S390X DFLTCC MSAN` run goes from 9min 12sec to 10min 14sec.
A bit slower for this already extremely slow testing.

`Ubuntu 20.04 GCC` goes from 1min 37sec to 32sec.
This is much faster, as are most of the native tests it seems.

`Ubuntu GCC PPC` goes from 4min 20sec to 4min 11sec
A little faster for this Qemu emulated testing.

So the tradeoff between not testing the complete corpora files vs testing a lot more compression parameters (10->540) on each seems to be pretty good from a CI time spent point of view.
Tweaks could be done to the blocksizes, and we could potentially enable the extra tests for a few of the currently fast ci runs Although it would be about 500-600x slower (assuming files are 4+MB), so even the 32sec run above would probably take several hours. So it might be better for datasets with smaller files (even extra tests would only test 4MB of the ~50MB mozilla tar(?)file).
Automated daily/weekly offline testing of develop could be an alternative, especially if anyone are willing to provide access to a machine or two that can do testing. Worth thinking about at least, would love to improve our testing.

I guess we could also use optipng/pngcrush to do the same kind of testing on png files, those also try all possible parameter combinations, but to find the best compression for a specific image. But we'd need to modify it to instead verify each output for data loss/crc.

configure
Dead2: Fixed.

test/parametrized_tester.c
Dead2: Fixed

test/parametrized_tester.c
Dead2: Fixed.

test/parametrized_tester.c
Dead2: Fixed.

test/parametrized_tester.c
Dead2: Changed.

test/parameterized_tester.c
```diff
@@ -0,0 +1,194 @@
...
+
+const int valid_compression_levels[] = {
+    0, 1, 2, 3, 4, 5, 6, 7, 8, 9
+};
+
+const int valid_mem_levels[] = {
+    1, 2, 3, 4, 5, 6, 7, 8, 9,
+};
+
+#define MAX_FILESIZE_KB 4096
```
nmoinvaz, line 37: Do we need this?

Why not just allocate dynamically enough memory for the entire file?

We already open it using fopen so we can fseek and ftell to get the size and then allocate memory.

test/parameterized_tester.c
```diff
@@ -0,0 +1,194 @@
...
+};
+
+const int valid_mem_levels[] = {
+    1, 8, 9,
+};
+
+#define MAX_FILESIZE_KB 128
+
+#  ifdef WITH_GZFILEOP
+#    define NUM_W 6
```
nmoinvaz, line 68: NUM_WINDOW_BITS
NUM_COMPRESS_LEVEL
NUM_MEM_LEVEL?

test/parameterized_tester.c
Dead2: Running 9450 iterations on 50+MB files would be extremely timeconsuming.

test/parameterized_tester.c
mtl1979: I agree that the preprocessor token names should be self-explanatory... This avoids need for adding extra comments for readability.

test/parameterized_tester.c
mtl1979: Time consuming or not, I would expect at least one iteration to fail without any code changes... That would warrant adding complex testing that essentially limits the corner cases we have tested before. Personally I'm not confident enough that limiting file size to just 4 MB (4096 kB) is enough.

test/parameterized_tester.c
```diff
@@ -0,0 +1,194 @@
...
+        fputs("A data input must be given as argument", stderr);
+        return EXIT_FAILURE;
+    }
+
+    if ((argc == 2 && strcmp(argv[1], "--help") == 0)) {
+        printf("Usage: test_parameterized <FILENAME>\n\n");
+        return 0;
+    }
+
+    FILE *in_file = fopen(argv[1], "r");
```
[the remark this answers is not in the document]
mtl1979: Opened test files likely reside in different directory, so the suggested patch is incorrect.

test/parameterized_tester.c
```diff
@@ -0,0 +1,194 @@
...
+#define MAX_FILESIZE_KB 4096
+
+#  ifdef WITH_GZFILEOP
+#    define NUM_W 21
+#  else
+#    define NUM_W 14
+#  endif
+#  define NUM_C 10
+#  define NUM_M 9
+
```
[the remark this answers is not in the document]
mtl1979: Binary mode is pretty much mandatory for at least Windows as it's the only supported platform that use two characters (CR+LF) for EOL.

test/parameterized_tester.c
```diff
@@ -0,0 +1,194 @@
...
+    if ((argc == 2 && strcmp(argv[1], "--help") == 0)) {
+        printf("Usage: test_parameterized <FILENAME>\n\n");
+        return 0;
+    }
+
+    FILE *in_file = fopen(argv[1], "r");
+    uint32_t read_size = (uint32_t)fread(data_in, 1, MAX_FILESIZE, in_file);
+    fclose(in_file);
+
```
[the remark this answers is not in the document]
mtl1979: The patch is incorrect, it doesn't match the original code. However testing for failure in fopen call is a good idea.

test/parameterized_tester.c
```diff
@@ -0,0 +1,194 @@
...
+    if (!read_size) {
+        fprintf(stderr, "File is empty or error: %s", argv[1]);
+        return EXIT_FAILURE;
+    }
+
+    for (int wbits_index = 0; wbits_index < NUM_W; wbits_index += 1) {
```
mtl1979, line 120: `wbits_index += 1` is same as `wbits_index++`, but less readable.

test/parameterized_tester.c
```diff
@@ -0,0 +1,194 @@
...
+        for (int complevel_index = 0; complevel_index < NUM_C; complevel_index += 1) {
+            for (int memlevel_index = 0; memlevel_index < NUM_M; memlevel_index += 1) {
+                for (int strat_index = 0; strat_index < 5; strat_index += 1) {
```
mtl1979, line 123: Same as above, nitpick coding style.

test/parameterized_tester.c
```diff
@@ -0,0 +1,194 @@
...
+    uint8_t data_in[MAX_FILESIZE];
+    uint8_t compressed_out[MAX_OUTSIZE];
+    uint8_t decompressed_out[MAX_OUTSIZE];
+    const char *error_format = "Error at %s\n"
+                               "level: %d\n"
+                               "wbits: %d\n"
+                               "memLevel: %d\n"
+                               "strategy: %d\n"
+                               "exit: %d\n"
+                               "message: %s\n";
```
mtl1979, line 99: For readability, all lines except the first one should be either indented or bulleted.

test/parameterized_tester.c
Dead2: We might roll back to using the original testing method instead then, and let this tool be something developers can run manually, like it was in #1448.
Because doing complete testing like this would just not be possible through github actions, the jobs would likely take more than the max 6hr job limit, and we would for sure quickly tie up our 10 runners and block all CI progress.

nmoinvaz: We could run this on a few files in our CI tests that way it gets coverage.

Dead2:
Replaced by #2207

rhpvorderman:
Sorry for being unresponsive on this one. There is a lot of stuff going on and this one fell through. My apologies.

PR #1890, opened 2025-03-25
ci: fixes for RISC-V

Icenowy:
Remove the currently broken RISC-V CI flow and add two test targets (GCC and Clang) with toolchain from ubuntu-latest system.

Tests passed with my fixes of RVV-related SIGSEGV -- see https://github.com/Icenowy/zlib-ng/actions/runs/14062430701/job/39376366365 (Clang) and https://github.com/Icenowy/zlib-ng/actions/runs/14054987248/job/39352378603 (GCC) . 

## Summary by CodeRabbit

- **New Features**
  - Enhanced CI pipelines now support RISC-V builds on Ubuntu with both GCC and Clang toolchains.
  - Introduced an updated cross-compilation configuration to optimize RISC-V build performance.

- **Chores**
  - Streamlined the build process by removing outdated CI configurations and redundant steps.


nmoinvaz:
Seems like a simplification that relies more heavily on qemu, which is nice. I don't know enough about the platform though.

Icenowy:
@nmoinvaz It always relies on QEMU -- but I switched from QEMU built (and packaged as tarball) by SiFive (which is already not available to be used in Actions because of restrictions on their GitHub org) to the one shipped by Ubuntu.

Dead2 (approved):
LGTM

PR #1892, opened 2025-03-28
CI: Changes to running benchmark

Dead2:
Running the benchmarks takes a long time, and when run through gtest it will not print the bench results.
Running it during CI increases coverage, but since the benchmarks do not actually verify that results are correct, the coverage it provides is worthless and misleading.

Removing it lets us see where we have actual CI test coverage, and speeds up CI runs as well.

Add a CI step to quickly crashtest the benchmarks, without collecting coverage data.

## Summary by CodeRabbit

- **Tests**
  - Updated the registration command for the `benchmark_zlib` test to include a minimum benchmark time.
  - Added a new job step in the GitHub Actions workflow to run benchmarks without collecting coverage data, enhancing testing capabilities across multiple configurations.
  - Updated job configurations to enable benchmarks across various platforms.
  - Improved memory management in the `slide_hash` benchmark to prevent memory leaks.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -19,15 +19,9 @@ jobs:
             os: ubuntu-latest
             compiler: gcc
             cxx-compiler: g++
-            cmake-args: -DWITH_SANITIZER=Address
+            cxxflags: -Wno-maybe-uninitialized
+            cmake-args: -DWITH_SANITIZER=Address -DWITH_BENCHMARKS=ON
             codecov: ubuntu_gcc
-
-          - name: Ubuntu GCC Benchmark
-            os: ubuntu-latest
-            compiler: gcc
-            cxx-compiler: g++
-            cmake-args: -DWITH_BENCHMARKS=ON
-            codecov: ubuntu_gcc_benchmark
 
           - name: Ubuntu GCC Native Instructions
             os: ubuntu-latest
@@ -88,7 +82,7 @@ jobs:
             os: ubuntu-latest
             compiler: gcc
             cxx-compiler: g++
-            cmake-args: -DCMAKE_C_FLAGS=-m32 -DCMAKE_CXX_FLAGS=-m32 -DCMAKE_SHARED_LINKER_FLAGS=-m32 -DCMAKE_EXE_LINKER_FLAGS=-m32
+            cmake-args: -DCMAKE_C_FLAGS=-m32 -DCMAKE_CXX_FLAGS=-m32 -DCMAKE_SHARED_LINKER_FLAGS=-m32 -DCMAKE_EXE_LINKER_FLAGS=-m32 -DWITH_BENCHMARKS=ON
             packages: gcc-multilib g++-multilib
             codecov: ubuntu_gcc_m32
 
@@ -166,7 +160,8 @@ jobs:
 
           - name: Ubuntu GCC ARM HF ASAN
             os: ubuntu-latest
-            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-armhf.cmake -DWITH_SANITIZER=Address
+            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-armhf.cmake -DWITH_SANITIZER=Address -DWITH_BENCHMARKS=ON
+            cxxflags: -Wno-psabi -Wno-maybe-uninitialized
             asan-options: detect_leaks=0
             packages: qemu-user crossbuild-essential-armhf
             gcov-exec: arm-linux-gnueabihf-gcov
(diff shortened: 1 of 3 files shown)


nmoinvaz (approved):

KungFuJesus:
I'm all for speeding up CI tests but it would be good to catch when a change breaks the benchmarks, particularly for architectures the contribute doesn't have access to.  Is there maybe a way we could instead reduce how many runs it does?  We only really need one trial per benchmark to make sure things still work.

`./benchmark_zlib --benchmark_min_time=1x` will make things very fast while still exercising all that code.  

```
	Command being timed: "./benchmark_zlib --benchmark_min_time=1x"
	User time (seconds): 2.86
```

Dead2:
> I'm all for speeding up CI tests but it would be good to catch when a change breaks the benchmarks, particularly for architectures the contribute doesn't have access to. Is there maybe a way we could instead reduce how many runs it does? We only really need one trial per benchmark to make sure things still work.
> 
> `./benchmark_zlib --benchmark_min_time=1x` will make things very fast while still exercising all that code.
> 
> ```
> 	Command being timed: "./benchmark_zlib --benchmark_min_time=1x"
> 	User time (seconds): 2.86
> ```

Sure, that solves the speed. But it still gives false test coverage all over the library.
Not sure what the best solution would be to run the benchmarks but not collect any coverage on it.

Would it be possible to run the benchmark after coverage has been uploaded perhaps? @nmoinvaz what do you think?

Edit: When benchmarking, it would have been ideal to collect coverage for the benchmark code itself, but not for the library code. That way it does not show false test coverage, but does let us collect coverage for the benchmark code so we can also see that it has been run.

nmoinvaz:
Well it gives increases coverage for repository but not necessarily the library. I mean if you exclude test files and benchmarks in codecov then that is also I way to get more accurate library coverage.

If this works, I'd be in favor of it:
./benchmark_zlib --benchmark_min_time=1x

Dead2:
> Well it gives increases coverage for repository but not necessarily the library. I mean if you exclude test files and benchmarks in codecov then that is also I way to get more accurate library coverage.

I think you misunderstand, running the benchmarks gives code coverage for library code as it hits some code that we don't touch using the tests themselves.
The problem with that is that although the library code is run, we don't catch incorrect data unless it crashes. So it is false coverage for the library itself.

nmoinvaz:
I see. I think the value of crash coverage is still important. Even though it isn't tested for accuracy. It's just my opinion. I don't mind either way.

Dead2:
> I see. I think the value of crash coverage is still important. Even though it isn't tested for accuracy. It's just my opinion. I don't mind either way.

Well, it makes it difficult to see what new tests we need to implement to have actual test coverage of that code, since codecov will tell us it is already covered.

Thats why I am thinking about possibilities to run it for crash-testing purposes, but discard either all coverage or only the library coverage.

nmoinvaz:
Perhaps we should just not run it for coverage. So if the binaries are built for coverage, then don't add benchmark_zlib to ctest. In other configurations it would still run, for crash coverage.

mtl1979:
We should test building benchmarks on at least one 32-bit architecture, as there has been at least one benchmark that failed to build on 32-bit architecture due to mismatched types.

.github/workflows/cmake.yml
```diff
@@ -171,7 +165,7 @@ jobs:
 
           - name: Ubuntu GCC ARM HF ASAN
             os: ubuntu-latest
-            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-armhf.cmake -DWITH_SANITIZER=Address
+            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-armhf.cmake -DWITH_SANITIZER=Address -DWITH_BENCHMARKS=ON
```
mtl1979, line 163: For this to work correctly, we need `-Wno-psabi` in `CXXFLAGS`.

.github/workflows/cmake.yml
```diff
@@ -18,22 +18,16 @@ jobs:
...
             compiler: gcc
+            cmake-args: -DWITH_BENCHMARKS=ON
             cxx-compiler: g++
 
           - name: Ubuntu GCC ASAN
             os: ubuntu-latest
             compiler: gcc
             cxx-compiler: g++
-            cmake-args: -DWITH_SANITIZER=Address
+            cmake-args: -DWITH_SANITIZER=Address -DWITH_BENCHMARKS=ON
```
mtl1979, line 23: Seems to need `-Wno-maybe-uninitialized` to build without warnings/errors.

.github/workflows/cmake.yml
Dead2: Added 👍

.github/workflows/cmake.yml
Dead2: Added 👍

Dead2:
Added to `Ubuntu 32-bit`, and removed from Sparc due to errors.

.github/workflows/cmake.yml
```diff
@@ -18,22 +18,17 @@
           - name: Ubuntu 20.04 GCC
             os: ubuntu-20.04
             compiler: gcc
+            cmake-args: -DWITH_BENCHMARKS=ON
```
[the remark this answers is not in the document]
Dead2: This is not actually a refactor suggestion though..

.github/workflows/cmake.yml
```diff
@@ -166,7 +160,8 @@ jobs:
 
           - name: Ubuntu GCC ARM HF ASAN
             os: ubuntu-latest
-            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-armhf.cmake -DWITH_SANITIZER=Address
+            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-armhf.cmake -DWITH_SANITIZER=Address -DWITH_BENCHMARKS=ON
+            cxxflags: -Wno-psabi -Wno-maybe-uninitialized
```
mtl1979, line 164: This seem to be ignored due to `cmake/detect-coverage.cmake`, see #1902.

PR #1893, opened 2025-03-28
SSE4.1 optimized chorba

KungFuJesus:
This is ~25-30% faster than the SSE2 variant on a core2 quad. The main reason for this has to do with the fact that, while incurring far fewer shifts, an entirely separate stack buffer has to be managed that is the size of the L1 cache on most CPUs. This was one of the main reasons the 32k specialized function was slower for the scalar counterpart, despite auto vectorizing. The auto vectorized loop was setting up the stack buffer at unaligned offsets, which is detrimental to performance pre-nehalem. Additionally, we were losing a fair bit of time to the zero initialization, which we are now doing more selectively.

There are a ton of loads and stores happening, and for sure we are bound on the fill buffer + store forwarding. An SSE2 version of this code is probably possible by simply replacing the shifts with unpacks with zero and the palignr's with shufpd's. I'm just not sure it'll be all that worth it, though. We are gating against SSE4.1 not because we are using specifically a 4.1 instruction but because that marks when Wolfdale came out and palignr became a lot faster.

## Summary by CodeRabbit

- **New Features**
	- Added support for SSE4.1 instruction set on x86 processors, enabling optimized CRC32 checksum calculations for improved performance on compatible hardware.
- **Bug Fixes**
	- None.
- **Tests**
	- Introduced new benchmarks and test cases for the SSE4.1-optimized CRC32 implementation, including tests for smaller buffer sizes.
- **Chores**
	- Updated build scripts and configuration tools to detect and enable SSE4.1 support when available.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,7 +131,8 @@ elseif(BASEARCH_S360_FOUND)
 elseif(BASEARCH_X86_FOUND)
     option(WITH_SSE2 "Build with SSE2" ON)
     cmake_dependent_option(WITH_SSSE3 "Build with SSSE3" ON "WITH_SSE2" OFF)
-    cmake_dependent_option(WITH_SSE42 "Build with SSE42" ON "WITH_SSSE3" OFF)
+    cmake_dependent_option(WITH_SSE41 "Build with SSE41" ON "WITH_SSSE3" OFF)
+    cmake_dependent_option(WITH_SSE42 "Build with SSE42" ON "WITH_SSE41" OFF)
     cmake_dependent_option(WITH_PCLMULQDQ "Build with PCLMULQDQ" ON "WITH_SSE42" OFF)
     cmake_dependent_option(WITH_AVX2 "Build with AVX2" ON "WITH_SSE42" OFF)
     cmake_dependent_option(WITH_AVX512 "Build with AVX512" ON "WITH_AVX2" OFF)
@@ -151,7 +152,7 @@ mark_as_advanced(FORCE
     WITH_DFLTCC_INFLATE
     WITH_CRC32_VX
     WITH_AVX2 WITH_SSE2
-    WITH_SSSE3 WITH_SSE42
+    WITH_SSSE3 WITH_SSE41 WITH_SSE42
     WITH_PCLMULQDQ
     WITH_ALTIVEC
     WITH_POWER8
@@ -997,9 +998,20 @@ if(WITH_OPTIM)
                 set(WITH_SSSE3 OFF)
             endif()
         endif()
+        if(WITH_SSE41)
+            check_sse41_intrinsics()
+            if(HAVE_SSE41_INTRIN AND WITH_SSSE3)
+                add_definitions(-DX86_SSE41)
+                set(SSE41_SRCS ${ARCHDIR}/chorba_sse41.c)
+                list(APPEND ZLIB_ARCH_SRCS ${SSE41_SRCS})
+                set_property(SOURCE ${SSE41_SRCS} PROPERTY COMPILE_FLAGS "${SSE41FLAG} ${NOLTOFLAG}")
+            else()
+                set(WITH_SSE41 OFF)
+            endif()
+        endif()
         if(WITH_SSE42)
             check_sse42_intrinsics()
-            if(HAVE_SSE42_INTRIN AND WITH_SSSE3)
+            if(HAVE_SSE42_INTRIN AND WITH_SSE41)
                 add_definitions(-DX86_SSE42)
                 set(SSE42_SRCS ${ARCHDIR}/adler32_sse42.c)
                 add_feature_info(SSE42_CRC 1 "Support SSE4.2 optimized adler32 hash generation, using \"${SSE42FLAG}\"")
@@ -1488,6 +1500,7 @@ elseif(BASEARCH_X86_FOUND)
     add_feature_info(WITH_AVX512VNNI WITH_AVX512VNNI "Build with AVX512 VNNI")
     add_feature_info(WITH_SSE2 WITH_SSE2 "Build with SSE2")
     add_feature_info(WITH_SSSE3 WITH_SSSE3 "Build with SSSE3")
+    add_feature_info(WITH_SSE41 WITH_SSE41 "Build with SSE41")
     add_feature_info(WITH_SSE42 WITH_SSE42 "Build with SSE42")
     add_feature_info(WITH_PCLMULQDQ WITH_PCLMULQDQ "Build with PCLMULQDQ")
     add_feature_info(WITH_VPCLMULQDQ WITH_VPCLMULQDQ "Build with VPCLMULQDQ")
--- a/arch/x86/chorba_sse41.c
+++ b/arch/x86/chorba_sse41.c
@@ -1,0 +1,342 @@
+#if !defined(WITHOUT_CHORBA) && defined(X86_SSE41)
+
+#include "zbuild.h"
+#include "crc32_braid_p.h"
+#include "crc32_braid_tbl.h"
+#include "crc32.h"
+#include <emmintrin.h>
+#include <smmintrin.h>
+#include "arch/x86/x86_intrins.h"
+#include "arch/generic/generic_functions.h"
+#include <assert.h>
+
+extern uint32_t crc32_braid_base(uint32_t c, const uint8_t *buf, size_t len);
+extern uint32_t chorba_small_nondestructive_sse2(uint32_t c, uint64_t *aligned_buf, size_t aligned_len);
+
+#define READ_NEXT(in, off, a, b) do { \
+        a = _mm_load_si128((__m128i*)(in + off / sizeof(uint64_t))); \
+        b = _mm_load_si128((__m128i*)(in + off / sizeof(uint64_t) + 2)); \
+        } while (0);
+
+#define NEXT_ROUND(invec, a, b, c, d) do { \
+        a = _mm_xor_si128(_mm_slli_epi64(invec, 17), _mm_slli_epi64(invec, 55)); \
+        b = _mm_xor_si128(_mm_xor_si128(_mm_srli_epi64(invec, 47), _mm_srli_epi64(invec, 9)), _mm_slli_epi64(invec, 19)); \
+        c = _mm_xor_si128(_mm_srli_epi64(invec, 45), _mm_slli_epi64(invec, 44)); \
+        d  = _mm_srli_epi64(invec, 20); \
+        } while (0);
+
+#define REALIGN_CHORBA(in0, in1, in2, in3, out0, out1, out2, out3, out4, shift) do { \
+        out0 = _mm_slli_si128(in0, shift); \
+        out1 = _mm_alignr_epi8(in1, in0, shift); \
+        out2 = _mm_alignr_epi8(in2, in1, shift); \
+        out3 = _mm_alignr_epi8(in3, in2, shift); \
+        out4 = _mm_srli_si128(in3, shift); \
+        } while (0)
+
+#define STORE4(out0, out1, out2, out3, out) do { \
+        _mm_store_si128(out++, out0); \
+        _mm_store_si128(out++, out1); \
+        _mm_store_si128(out++, out2); \
+        _mm_store_si128(out++, out3); \
+    } while (0)
+
+#define READ4(out0, out1, out2, out3, in) do { \
+    out0 = _mm_load_si128(in++); \
+    out1 = _mm_load_si128(in++); \
+    out2 = _mm_load_si128(in++); \
+    out3 = _mm_load_si128(in++); \
+    } while (0)
+
+/* This is intentionally shifted one down to compensate for the deferred store from
+ * the last iteration */
+#define READ4_WITHXOR(out0, out1, out2, out3, xor0, xor1, xor2, xor3, in) do { \
+    out0 = _mm_xor_si128(in[1], xor0); \
+    out1 = _mm_xor_si128(in[2], xor1); \
+    out2 = _mm_xor_si128(in[3], xor2); \
+    out3 = _mm_xor_si128(in[4], xor3); \
+    } while (0)
+
+static Z_FORCEINLINE uint32_t crc32_chorba_32768_nondestructive_sse41(uint32_t crc, const uint64_t* buf, size_t len) {
+    const uint64_t* input = buf;
+    ALIGNED_(16) uint64_t bitbuffer[32768 / sizeof(uint64_t)];
+    __m128i *bitbuffer_v = (__m128i*)bitbuffer;
+    const uint8_t* bitbufferbytes = (const uint8_t*) bitbuffer;
+    __m128i z = _mm_setzero_si128();
+
+    __m128i *bitbuf128 = &bitbuffer_v[64];
+    __m128i *bitbuf144 = &bitbuffer_v[72];
+    __m128i *bitbuf182 = &bitbuffer_v[91];
+    __m128i *bitbuf210 = &bitbuffer_v[105];
+    __m128i *bitbuf300 = &bitbuffer_v[150];
+    __m128i *bitbuf0 = bitbuf128;
+    __m128i *inptr = (__m128i*)input;
+
+    /* We only need to zero out the bytes between the 128'th value and the 144th
+     * that are actually read */
+    __m128i *z_cursor = bitbuf128;
+    for (size_t i = 0; i < 2; ++i) {
+        STORE4(z, z, z, z, z_cursor);
+    }
+
+    /* We only need to zero out the bytes between the 144'th value and the 182nd that
+     * are actually read */
+    z_cursor = bitbuf144 + 8;
+    for (size_t i = 0; i < 11; ++i) {
+        _mm_store_si128(z_cursor++, z);
+    }
+
+    /* We only need to zero out the bytes between the 182nd value and the 210th that
+     * are actually read. */
+    z_cursor = bitbuf182;
+    for (size_t i = 0; i < 4; ++i) {
+        STORE4(z, z, z, z, z_cursor);
+    }
+
+    /* We need to mix this in */
+    __m128i init_crc = _mm_cvtsi64x_si128(crc);
+    crc = 0;
+
+    size_t i = 0;
+
+    /* Previous iteration runs carried over */
+    __m128i buf144 = z;
+    __m128i buf182 = z;
+    __m128i buf210 = z;
+
+    for(; i + 300*8+64 < len && i < 22 * 8; i += 64) {
+        __m128i in12, in34, in56, in78,
+                in_1, in23, in45, in67, in8_;
+
+        READ4(in12, in34, in56, in78, inptr);
+
+        if (i == 0) {
+            in12 = _mm_xor_si128(in12, init_crc);
+        }
+
+        REALIGN_CHORBA(in12, in34, in56, in78,
+                       in_1, in23, in45, in67, in8_, 8);
+
+        __m128i a = _mm_xor_si128(buf144, in_1);
+
+        STORE4(a, in23, in45, in67, bitbuf144);
+        buf144 = in8_;
+
+        __m128i e = _mm_xor_si128(buf182, in_1);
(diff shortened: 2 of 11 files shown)


KungFuJesus:
```
2025-03-28T18:16:29-04:00
Running ./benchmark_zlib
Run on (4 X 3003 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x4)
  L1 Instruction 32 KiB (x4)
  L2 Unified 6144 KiB (x2)
Load Average: 1.91, 1.36, 0.75
---------------------------------------------------------------------
Benchmark                           Time             CPU   Iterations
---------------------------------------------------------------------
crc32/generic_chorba/32768      15754 ns        15762 ns        44401
crc32/braid/32768               24084 ns        24095 ns        29072
crc32/chorba_sse2/32768          8991 ns         8994 ns        77933
crc32/chorba_sse41/32768         6264 ns         6267 ns       111684
```

win32/Makefile.msc
```diff
@@ -61,6 +62,7 @@ OBJS = \
 	chunkset_sse2.obj \
 	chunkset_ssse3.obj \
     chorba_sse2.obj \
+    chorba_sse41.obj \
```
nmoinvaz: I think these TWO lines need tabs instead of spaces.

nmoinvaz:
There are a bunch of these errors:
image

KungFuJesus:
> There are a bunch of these errors: image

I mean that is intentional, 2 64 bit integers is a stride through a single __m128i.  Maybe I modify the pointer type to appease codeQL, though.

Err actually no, that's me intentionally advancing an __m128i pointer.  What else would it have me do? The intrinsic _requires_ a __m128i pointer.

KungFuJesus:
Can we whitelist this thing with codeQL or something?  The alternative looks like this:

```
_mm_load_si128((__m128i*)((uint64_t*)bitbuffer182 + 2 * 2));
```

Hmm, evidently derefing with a [] is what it wants? https://codeql.github.com/codeql-query-help/cpp/cpp-suspicious-pointer-scaling/

That's rather silly, but whatever, it's the same thing.

CMakeLists.txt
```diff
@@ -126,6 +126,7 @@ elseif(BASEARCH_X86_FOUND)
     option(WITH_SSE2 "Build with SSE2" ON)
     cmake_dependent_option(WITH_SSSE3 "Build with SSSE3" ON "WITH_SSE2" OFF)
     cmake_dependent_option(WITH_SSE42 "Build with SSE42" ON "WITH_SSSE3" OFF)
+    cmake_dependent_option(WITH_SSE41 "Build with SSE41" ON "WITH_SSSE3" OFF)
```
phprus:
```suggestion
    cmake_dependent_option(WITH_SSE41 "Build with SSE41" ON "WITH_SSSE3" OFF)
    cmake_dependent_option(WITH_SSE42 "Build with SSE42" ON "WITH_SSE41" OFF)
```

CMakeLists.txt
```diff
@@ -145,7 +146,7 @@ mark_as_advanced(FORCE
     WITH_DFLTCC_INFLATE
     WITH_CRC32_VX
     WITH_AVX2 WITH_SSE2
-    WITH_SSSE3 WITH_SSE42
+    WITH_SSSE3 WITH_SSE42 WITH_SSE41
```
phprus:
```suggestion
    WITH_SSSE3 WITH_SSE41 WITH_SSE42
```

CMakeLists.txt
```diff
@@ -991,6 +992,17 @@ if(WITH_OPTIM)
...
+                set(SSE41_SRCS ${ARCHDIR}/chorba_sse41.c)
+                list(APPEND ZLIB_ARCH_SRCS ${SSE41_SRCS})
+                set_property(SOURCE ${SSE41_SRCS} PROPERTY COMPILE_FLAGS "${SSE41FLAG} ${NOLTOFLAG}")
+            else()
+                set(WITH_SSE41 OFF)
+            endif()
+        endif()
         if(WITH_SSE42)
             check_sse42_intrinsics()
             if(HAVE_SSE42_INTRIN AND WITH_SSSE3)
```
phprus:
```suggestion
            if(HAVE_SSE42_INTRIN AND WITH_SSE41)
```

CMakeLists.txt
```diff
@@ -1481,6 +1493,7 @@ elseif(BASEARCH_X86_FOUND)
     add_feature_info(WITH_SSE2 WITH_SSE2 "Build with SSE2")
     add_feature_info(WITH_SSSE3 WITH_SSSE3 "Build with SSSE3")
     add_feature_info(WITH_SSE42 WITH_SSE42 "Build with SSE42")
+    add_feature_info(WITH_SSE41 WITH_SSE41 "Build with SSE41")
```
phprus:
```suggestion
    add_feature_info(WITH_SSE41 WITH_SSE41 "Build with SSE41")
    add_feature_info(WITH_SSE42 WITH_SSE42 "Build with SSE42")
```

arch/x86/x86_functions.h
```diff
@@ -114,6 +120,10 @@ uint32_t crc32_vpclmulqdq(uint32_t crc32, const uint8_t *buf, size_t len);
 #    undef native_inflate_fast
 #    define native_inflate_fast inflate_fast_ssse3
 #  endif
+#  if !defined(WITHOUT_CHORBA) && defined(X86_SSE41) && !defined(NO_CHORBA_SSE2)
```
phprus: Maybe rename ``NO_CHORBA_SSE2`` to ``NO_CHORBA_SSE``?

functable.c
```diff
@@ -95,6 +95,16 @@ static void init_functable(void) {
         ft.inflate_fast = &inflate_fast_ssse3;
     }
 #endif
+
+    // X86 - SSE41
```
phprus:
```suggestion
    // X86 - SSE4.1
```

arch/x86/chorba_sse41.c
```diff
@@ -0,0 +1,407 @@
+#if !defined(WITHOUT_CHORBA) && defined(X86_SSE41) && defined(X86_SSE2)
```
phprus:
```suggestion
#if !defined(WITHOUT_CHORBA) && defined(X86_SSE41)
```

CMakeLists.txt
KungFuJesus: So things do get a little strange here - namely there are CPUs with SSE4a that do not have SSSE3 (I think AMD fits that bill for a lot of those).  However, I think for the full 4.1 set this is correct in that 4.2 implies 4.1.

CMakeLists.txt
KungFuJesus: Heh, actually I think this should be WIHT_SSE42 since we do build with -msse42 and give the compiler leeway to generate 4.2 instructions (though, I'm near certain it isn't).  Still, it could one day so we better.  SSE4.2 was a pivotal point when Intel released Nehalem and unaligned loads were pipelined out.

CMakeLists.txt
phprus: I didn't understand you.
SSE4a is a separate instruction set from AMD and it has a dedicated CPUID bit.

This changes use SSE4.1 and SSE4.1 CPUID bit.

arch/x86/x86_functions.h
```diff
@@ -114,6 +120,10 @@ uint32_t crc32_vpclmulqdq(uint32_t crc32, const uint8_t *buf, size_t len);
 #    undef native_inflate_fast
 #    define native_inflate_fast inflate_fast_ssse3
 #  endif
+#  if !defined(WITHOUT_CHORBA) && defined(X86_SSE41) && !defined(NO_CHORBA_SSE)
```
phprus: For fix native build we need to use predef macro ``__SSE4_1__`` to detect sse4.1 is present:

```suggestion
#  if !defined(WITHOUT_CHORBA) && defined(X86_SSE41) && defined(__SSE4_1__) && !defined(NO_CHORBA_SSE)
```

phprus (changes requested):

CMakeLists.txt
```diff
@@ -991,9 +992,20 @@ if(WITH_OPTIM)
                 set(WITH_SSSE3 OFF)
             endif()
         endif()
+        if(WITH_SSE41)
+            check_sse41_intrinsics()
+            if(HAVE_SSE41_INTRIN AND WITH_SSE41)
```
phprus: SSE4.1 is enabled if compiler support it and SSSE3 is present (this is declare dependencies between instructions set: SSE4.1 is "superset" of SSSE3):

```suggestion
            if(HAVE_SSE41_INTRIN AND WITH_SSSE3)
```

CMakeLists.txt
```diff
@@ -991,9 +992,20 @@ if(WITH_OPTIM)
...
+                list(APPEND ZLIB_ARCH_SRCS ${SSE41_SRCS})
+                set_property(SOURCE ${SSE41_SRCS} PROPERTY COMPILE_FLAGS "${SSE41FLAG} ${NOLTOFLAG}")
+            else()
+                set(WITH_SSE41 OFF)
+            endif()
+        endif()
         if(WITH_SSE42)
             check_sse42_intrinsics()
-            if(HAVE_SSE42_INTRIN AND WITH_SSSE3)
+            if(HAVE_SSE42_INTRIN AND WITH_SSE42)
```
phprus: SSE4.2 is enabled if compiler support it and SSE4.1 is present (this is declare dependencies between instructions set: SSE4.2 is "superset" of SSE4.1):

```suggestion
            if(HAVE_SSE42_INTRIN AND WITH_SSE41)
```

phprus (changes requested):

KungFuJesus:
Anybody have any idea why the 32 bit exported functions differ now?

CMakeLists.txt
KungFuJesus: Yes, and that's the one we want, I was mostly musing that SSE4a, which is a minor subset of SSE41 (albeit with no actual vector operations) was exposed on AMD CPUs and didn't include SSSE3.  SSE42 I believe does mean that SSE41 is fully supported (and I believe the compiler implies this as well).

phprus: These checks are added to prevent the user from building an impossible configuration.
For example, with SSE4.2 but without SSE2 (ran cmake with flags ``-DWITH_SSE2=OFF -DWITH_SSE42=ON``).

arch/x86/chorba_sse41.c
```diff
@@ -0,0 +1,344 @@
...
+        } while (0);
+
+#define REALIGN_CHORBA(in0, in1, in2, in3, out0, out1, out2, out3, out4, shift) do { \
+        out0 = _mm_slli_si128(in0, shift); \
+        out1 = _mm_alignr_epi8(in1, in12, shift); \
+        out2 = _mm_alignr_epi8(in2, in34, shift); \
+        out3 = _mm_alignr_epi8(in3, in56, shift); \
+        out4 = _mm_srli_si128(in3, shift); \
+        } while (0)
+
```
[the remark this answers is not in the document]
KungFuJesus: Good call, you're not terrible sometimes, robot.

arch/x86/x86_functions.h
```diff
@@ -36,6 +36,12 @@ uint8_t* chunkmemset_safe_ssse3(uint8_t *out, uint8_t *from, unsigned len, unsig
 void inflate_fast_ssse3(PREFIX3(stream) *strm, uint32_t start);
 #endif
 
+#ifdef X86_SSE41
+#   if !defined(WITHOUT_CHORBA) && defined(X86_SSE2)
```
phprus:
```suggestion
#   if !defined(WITHOUT_CHORBA)
```

We do not need check ``defined(X86_SSE2)`` if ``X86_SSE41`` macro is defined

phprus (commented):
One small suggestion for code.

Other code for CPU features check and implementation selector: LGTM

arch/x86/chorba_sse41.c
```diff
@@ -0,0 +1,344 @@
...
+    out3 = _mm_load_si128(in++); \
+    } while (0)
+
+/* This is intentionally shifted one down to compensate for the deferred store from
+ * the last iteration */
+#define READ4_WITHXOR(out0, out1, out2, out3, xor0, xor1, xor2, xor3, in) do { \
+    out0 = _mm_xor_si128(_mm_load_si128(&in[1]), xor0); \
+    out1 = _mm_xor_si128(_mm_load_si128(&in[2]), xor1); \
+    out2 = _mm_xor_si128(_mm_load_si128(&in[3]), xor2); \
+    out3 = _mm_xor_si128(_mm_load_si128(&in[4]), xor3); \
```
mtl1979: Taking address of an array element of vector type and then loading it with intrinsic is redundant as dereferencing with `[]` already implies loading to register.

arch/x86/chorba_sse41.c
```diff
@@ -0,0 +1,344 @@
...
+static Z_FORCEINLINE uint32_t crc32_chorba_32768_nondestructive_sse41(uint32_t crc, const uint64_t* buf, size_t len) {
+    const uint64_t* input = buf;
+    ALIGNED_(16) uint64_t bitbuffer[32768 / sizeof(uint64_t)];
+    const uint8_t* bitbufferbytes = (const uint8_t*) bitbuffer;
+    __m128i z = _mm_setzero_si128();
+
+    __m128i *bitbuf128 = (__m128i*)(bitbuffer + 128);
+    __m128i *bitbuf144 = (__m128i*)(bitbuffer + 144);
+    __m128i *bitbuf182 = (__m128i*)(bitbuffer + 182);
+    __m128i *bitbuf0 = bitbuf128;
```
mtl1979: This is backwards... Last line gets moved to top by compiler when it does common subexpression elimination. Should use array notation for three top lines.

arch/x86/chorba_sse41.c
```diff
@@ -0,0 +1,344 @@
...
+    /* We need to mix this in */
+    __m128i init_crc = _mm_cvtsi64x_si128(crc);
+    crc = 0;
+
+    size_t i = 0;
+    __m128i *inptr = (__m128i*)input;
+    bitbuf144 = (__m128i*)(bitbuffer + 144);
+    bitbuf182 = (__m128i*)(bitbuffer + 182);
+    __m128i *bitbuf210 = (__m128i*)(bitbuffer + 210);
+    __m128i *bitbuf300 = (__m128i*)(bitbuffer + 300);
```
mtl1979: Should use temporary variable to avoid excessive casting, array notation as above...

arch/x86/chorba_sse41.c
KungFuJesus: If you're asking why I don't just dereference the pointer - that was mainly to convey to the compiler and the reader that this is an explicitly aligned load.  Doing pointer arithmetic here makes CodeQL lose its mind and whine that I might not know what I'm doing so I had to add the & + [].

phprus:
Missing ``X86_SSE41`` macro in https://github.com/zlib-ng/zlib-ng/blob/develop/win32/Makefile.msc (``WFLAGS`` variable)

KungFuJesus:
> Missing `X86_SSE41` macro in https://github.com/zlib-ng/zlib-ng/blob/develop/win32/Makefile.msc (`WFLAGS` variable)

Are you looking at the wrong branch?  It's definitely in there:
https://github.com/KungFuJesus/zlib-ng/blob/sse41_chorba/win32/Makefile.msc#L36

phprus:
@KungFuJesus 
My apologies!
I got it wrong tabs in the browser where branches were open.

arch/x86/chorba_sse41.c
mtl1979: Using `[]` without `&` and `__mm_load_si128()` around it would do the same as `[]` already implies aligned load.

The current version violates the rule that forbids taking an address of something that should be stored in a register. It's essentially using LEA (load effective address) semantics, which disable any possible optimisations, including dead code elimination. Without LEA semantics, it would use indexed load, which avoids use of extra register.

KungFuJesus: Fair enough.  Changed.

arch/x86/chorba_sse41.c
```diff
@@ -0,0 +1,343 @@
...
+    __m128i *z_cursor = bitbuf128;
+    for (size_t i = 0; i < 2; ++i) {
+        STORE4(z, z, z, z, z_cursor);
+    }
+
+    /* We only need to zero out the bytes between the 144'th value and the 182nd that
+     * are actually read */
+    z_cursor = bitbuf144;
+    for (size_t i = 0; i < 11; ++i) {
+        _mm_store_si128(&z_cursor[8], z);
```
mtl1979: This is same as `z_cursor[8] = z;`, but less readable. Taking address of an array element just for assignment is unnecessary when the write length is same as length of the array element. Index semantics are faster than LEA semantics due to risk of unnecessary spilling when there is shortage of general-purpose registers.

arch/x86/chorba_sse41.c
```diff
@@ -0,0 +1,343 @@
...
+    z_cursor = bitbuf182;
+    for (size_t i = 0; i < 4; ++i) {
+        STORE4(z, z, z, z, z_cursor);
+    }
+
+    /* We need to mix this in */
+    __m128i init_crc = _mm_cvtsi64x_si128(crc);
+    crc = 0;
+
+    size_t i = 0;
```
mtl1979, line 99: Duplicate variable name with no restricting scope.

arch/x86/chorba_sse41.c
```diff
@@ -0,0 +1,343 @@
...
+        e = _mm_xor_si128(buf182, in_1);
+        READ4_WITHXOR(f, g, h, buf182, in23, in45, in67, in8_, bitbuf182);
+        STORE4(e, f, g, h, bitbuf182);
+
+        __m128i n, o, p;
+        __m128i m = _mm_xor_si128(buf210, in_1);
+
+        /* Couldn't tell you why but despite knowing that this is always false,
+         * removing this branch with GCC makes things significantly slower. Some
+         * loop bodies must be being joined or something */
```
mtl1979, line 219: Abusing `for` loop for ranges with unknown start value disables some optimisations as compiler can't deduce how many iterations are run. The `if` block(s) pretty much give the missing information even though the conditions are reversed.

Optimising compiler should in theory duplicate the loop with known start values from `if` loops.

arch/x86/chorba_sse41.c
KungFuJesus: I'm not sure what you mean, the loops above are the restricted scope of i, it does not live outside of the for loop declaration.

arch/x86/chorba_sse41.c
KungFuJesus: Hmm, possibly.  I haven't compared the compiled code to see exactly what's happening but yeah it could be that.

KungFuJesus: Yeah, removing the condition there seems to be generating a lot more code so yeah I'm guessing one or more of the loops are being joined together.

Dead2 (approved):
LGTM

Dead2:
@KungFuJesus Sorry, this needs a rebase now.

KungFuJesus:
Rebased

arch/x86/chorba_sse41.c
mtl1979: In C, outer scope starts when function starts. To be valid ANSI C, it would require explicit block scope. Some compilers will issue warning if variable in inner scope shadows variable in outer scope, but that warning can be suppressed.

KungFuJesus: I'm fairly certain declaring the variables in the loop parameters is the explicit scope block.  C89 at one point wouldn't allow you to declare variables in that but I don't know that we are trying to cater that far back.  GNU's C extension and afterward anything C99 compliant accepts variable declarations there.  

```C
#include <stdlib.h>

int main(void)
{
    for (int i = 0; i < 20; ++i) {};
    return i;
}
```

```bash
[adam@thinkpad ~]$ gcc test_scope.c -o test.out
test_scope.c: In function ‘main’:
test_scope.c:6:12: error: ‘i’ undeclared (first use in this function)
    6 |     return i;
      |            ^
test_scope.c:6:12: note: each undeclared identifier is reported only once for each function it appears in
```

PR #1894, opened 2025-04-02
CI: Preinstall packages needed for testing and benchmark.

Dead2:
Avoids having to compile gtest and google benchmark in every CI job. To make sure we also test downloading and building ourself, don't install for jobs that specify any matrix.packages.

In my spotchecks, it typically spends ~7-8sec extra to run apt install, while the compile is ~22-40sec faster.

## Summary by CodeRabbit

- **Chores**
  - Refined the continuous integration conditions to streamline the setup process.
  - Added a default package installation fallback to ensure essential dependencies are reliably installed during automated processes.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -653,7 +653,7 @@ jobs:
       run: sudo dpkg --add-architecture i386
 
     - name: Add ubuntu mirrors
-      if: runner.os == 'Linux' && matrix.packages && !contains(matrix.os, 'z15')
+      if: runner.os == 'Linux' && !contains(matrix.os, 'z15')
       # Github Actions caching proxy is at times unreliable
       run: |
         echo -e 'http://azure.archive.ubuntu.com/ubuntu\tpriority:1\n' | sudo tee /etc/apt/mirrors.txt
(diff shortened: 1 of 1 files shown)


nmoinvaz (approved):

.github/workflows/cmake.yml
```diff
@@ -653,18 +653,19 @@ jobs:
...
         sudo sed -i 's#http://azure.archive.ubuntu.com/ubuntu/#mirror+file:/etc/apt/mirrors.txt#' /etc/apt/sources.list
 
     - name: Install packages (Ubuntu)
-      if: runner.os == 'Linux' && matrix.packages && !contains(matrix.os, 'z15')
+      if: runner.os == 'Linux' && !contains(matrix.os, 'z15')
       run: |
         sudo apt-get update
-        sudo apt-get install -y --allow-downgrades --no-install-recommends ${{ matrix.packages }}
+        sudo apt-get install -y --allow-downgrades --no-install-recommends \
+            ${{ matrix.packages || 'libgtest-dev libbenchmark-dev' }}
```
nmoinvaz, line 668:
```bash
sudo apt-get install -y --allow-downgrades --no-install-recommends \
    libgtest-dev libbenchmark-dev ${{ matrix.packages }}
```
No need for || ?

Dead2: It is if we still want to also test the download-and-compile-ourselves code.

PR #1896, opened 2025-04-04
Pass POSIX_C_SOURCE for std::alligned_alloc try_compile checks

cvijdea-bd:
On FreeBSD 11, defining POSIX_C_SOURCE to a lower level has the efect of inhibiting the language level (__ISO_C_VISIBLE) to be lower than C11, even in the presence of -std=c11

Since the check_symbol_exists runs without setting POSIX_C_SOURCE, this means that we will spuriously define HAVE_ALIGNED_ALLOC, while in the actual build it is not going to be defined

ref: https://github.com/freebsd/freebsd-src/blob/stable/11/sys/sys/cdefs.h#L738

## Summary by CodeRabbit

Our system's build process has been updated to improve compatibility and stability regarding memory allocation handling.

- **Chores**
  - Enhanced the build configuration to support both POSIX and C11 standards for aligned memory allocation, contributing to improved compliance and system reliability. This update ensures a more stable build process and better overall performance.


Dead2 (approved):
LGTM

Dead2:
Actually, this should be changed in Configure as well, so they are kept in sync.

@cvijdea-bd Would you mind adding the defines to Configure as well?
https://github.com/zlib-ng/zlib-ng/blob/e88471f3795170e6fe63e384b77b71ab5c46d9ad/configure#L730
https://github.com/zlib-ng/zlib-ng/blob/e88471f3795170e6fe63e384b77b71ab5c46d9ad/configure#L751

cvijdea-bd:
Yes that makes sense, I pushed the modified commit

PR #1897, opened 2025-04-04
Fix pointer type mismatch.

mtl1979:
* When passing pointer as function parameter, we need to make sure the underlying type is exact match.

## Summary by CodeRabbit

- **Refactor**
  - Enhanced internal handling of compression and decompression operations to support larger data sizes, which helps ensure robustness and scalability for demanding use cases.


Dead2 (approved):
LGTM

nmoinvaz (approved):

PR #1902, opened 2025-04-11
Fix CXXFLAGS when coverage enabled.

mtl1979:
This allows CI to add flags for C++ compiler even when coverage is enabled. According to CMake documentation `CXXFLAGS` is only used if `CMAKE_CXX_FLAGS` is unset. Enable C++ support early enough so `CMAKE_CXX_FLAGS` is correctly populated.

## Summary by CodeRabbit

- **Chores**
	- Enhanced the build configuration to include new options for enabling C++ support, improving flexibility in the build process. Options added include support for Google Test, fuzz tests, and benchmarks.


phprus:
> According to CMake documentation `CFLAGS` is only used if `CMAKE_C_FLAGS` is unset and `CXXFLAGS` is only used if `CMAKE_CXX_FLAGS` is unset.

https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS.html:

> CMAKE_C_FLAGS: **Initialized by** the CFLAGS environment variable.
> CMAKE_CXX_FLAGS: **Initialized by** the CXXFLAGS environment variable.

As I see, ``CMAKE_*_FLAGS`` are overridden only in:

https://github.com/zlib-ng/zlib-ng/blob/e4a31e0538277a7960d5e3e78023db38fa301f9a/test/pkgcheck.sh#L13

https://github.com/zlib-ng/zlib-ng/blob/e4a31e0538277a7960d5e3e78023db38fa301f9a/.github/workflows/pkgcheck.yml#L24

https://github.com/zlib-ng/zlib-ng/blob/e4a31e0538277a7960d5e3e78023db38fa301f9a/.github/workflows/cmake.yml#L91

Maybe it makes sense to fix these workflows instead of duplicate ``C*FLAGS`` value in ``CMAKE_*_FLAGS``?

mtl1979:
> Maybe it makes sense to fix these workflows instead of duplicate C*FLAGS value in CMAKE_*_FLAGS?

`pkgcheck.sh` use both configure and cmake, so it is special case.

The issue is with cmake populating `CMAKE_C_FLAGS` and `CMAKE_CXX_FLAGS` too late. So only way to set the variables early enough is to use `CFLAGS` and `CXXFLAGS` when coverity scan is enabled. Without the workaround both variables are always set to `-O0` despite what was passed to cmake.

When coverity scan is not enabled, there is no issue and the workaround is skipped.

phprus:
> The issue is with cmake populating `CMAKE_C_FLAGS` and `CMAKE_CXX_FLAGS` too late.

What do you mean by too late?
I thought that it was initialized during the project(LANGUAGE) call. Or was I wrong?

mtl1979:
> > The issue is with cmake populating `CMAKE_C_FLAGS` and `CMAKE_CXX_FLAGS` too late.
> 
> What do you mean by too late? I thought that it was initialized during the project(LANGUAGE) call. Or was I wrong?

Clearly when I and dead2 tested, those variables were empty when cmake hit the function in the PR.

See for example https://github.com/zlib-ng/zlib-ng/actions/runs/14375470240/job/40306746506?pr=1892

nmoinvaz:
This may be why I had to turn off gtest for PPC64LE ci run.

mtl1979:
> This may be why I had to turn off gtest for PPC64LE ci run.

You can verify that locally by re-enabling gtest for PPC64LE after checking out branch from this PR.

nmoinvaz:
> You can verify that locally by re-enabling gtest for PPC64LE after checking out branch from this PR.

Nope this PR doesn't fix that.

mtl1979:
> Nope this PR doesn't fix that.

At least this workaround allows enabling benchmarks on 32-bit ARM with hard-float on CI.

Dead2 (approved):
LGTM

phprus:
> See for example https://github.com/zlib-ng/zlib-ng/actions/runs/14375470240/job/40306746506?pr=1892

I have one idea. I'll check it now and write the result...

phprus:
> > > The issue is with cmake populating `CMAKE_C_FLAGS` and `CMAKE_CXX_FLAGS` too late.
> > 
> > 
> > What do you mean by too late? I thought that it was initialized during the project(LANGUAGE) call. Or was I wrong?
> 
> Clearly when I and dead2 tested, those variables were empty when cmake hit the function in the PR.
> 
> See for example https://github.com/zlib-ng/zlib-ng/actions/runs/14375470240/job/40306746506?pr=1892

@mtl1979 , @Dead2 , this is incorrect PR!

The C++ language is enabled too late. After the ``CMAKE_CXX_*`` variables are used.

We need to enable C++ directly in ``project()``.
Or conditionally, but BEFORE the line:
https://github.com/zlib-ng/zlib-ng/blob/e4a31e0538277a7960d5e3e78023db38fa301f9a/CMakeLists.txt#L50

See fix commit https://github.com/phprus/zlib-ng/commit/06678084e81e827a523bbf256cb2e56436116012
and CI: https://github.com/phprus/zlib-ng/actions/runs/14418138299/job/40437352696

mtl1979:
> The C++ language is enabled too late. After the `CMAKE_CXX_*` variables are used.
> 
> We need to enable C++ directly in `project()`. Or conditionally, but BEFORE the line:
> 
> https://github.com/zlib-ng/zlib-ng/blob/e4a31e0538277a7960d5e3e78023db38fa301f9a/CMakeLists.txt#L50
> 
> See fix commit [phprus@0667808](https://github.com/phprus/zlib-ng/commit/06678084e81e827a523bbf256cb2e56436116012) and CI: https://github.com/phprus/zlib-ng/actions/runs/14418138299/job/40437352696

A lot of operating systems don't come with C++ compiler installed by default. I'd rather enable C++ only when it's actually needed...

phprus:
> A lot of operating systems don't come with C++ compiler installed by default. I'd rather enable C++ only when it's actually needed...

Yes, you are right.
I suggest enabling C++ earlier only if needed: https://github.com/phprus/zlib-ng/commit/b0832f2e49ba46fe525c30c14d447f8c558aa4c9

The previous commit was a proof of concept only.

CMakeLists.txt
```diff
@@ -47,6 +47,11 @@ include(CMakeDependentOption)
 include(CMakePackageConfigHelpers)
 include(FeatureSummary)
 
+# We need to enable C++ before trying to check for coverage
+if(WITH_GTEST OR WITH_BENCHMARKS)
```
phprus:
```suggestion
if(WITH_GTEST OR WITH_FUZZERS OR WITH_BENCHMARKS)
```

test/fuzz can also use C++:

https://github.com/zlib-ng/zlib-ng/blob/e4a31e0538277a7960d5e3e78023db38fa301f9a/test/fuzz/CMakeLists.txt#L4

phprus: Flags enabled by default are defined later this ``if``
For example ``WITH_GTEST``.

I think we need reorder code like this:

https://github.com/phprus/zlib-ng/commit/b0832f2e49ba46fe525c30c14d447f8c558aa4c9

mtl1979: I'd like to keep the patch as short as possible, so moving only the "used" options before the check...

Dead2 (approved):
Yeah, this looks like a better solution 👍

nmoinvaz (approved):

mtl1979:
> Yeah, this looks like a better solution 👍

Once this PR is merged, #1892 needs to be rebased for CI to pass on all except EL9...

PR #1903, opened 2025-04-14
Remove late enable_language calls.

phprus:
Cleanup after #1902

## Summary by CodeRabbit

- **Chores**
  - Revised the build configuration by removing explicit language activation commands for C and C++. This update affects how language-specific code is processed in scenarios like tests, benchmarks, and fuzzing, so users may need to verify that their build environments meet the necessary compilation requirements.


Dead2 (approved):
LGTM

PR #1912, opened 2025-05-12
Fix 32bit large chorba

### LINKED ISSUE #1910: crc32 fails with 32bit mingw32 builds (chorba issue)
I just updated zlib-ng to the latest, and my tests started to fail. All kinds of tests pass, but specifically 32bit mingw32 builds fail because crc32 values are incorrect. I tried to take slightly older version and it was ok. If I `-DWITHOUT_CHORBA` then it's ok in the latest version. Seems like something isn't right.

Note, the latest code passes tests if I use msvc2022, clang-cl or intel ixc compiler (all 32bit builds).
It's specifically, either 32bit gcc (I didn't try 32bit gcc outsize of mingw32), or mingw32 only

samrussell:
Fixes #1910

Tested on WSL/Ubuntu 24.04.1 gcc with --m32

## Summary by CodeRabbit

- **Refactor**
	- Improved internal data processing by optimizing memory access to use 64-bit word operations for enhanced performance and consistency.
	- Updated byte order handling and data copying methods to align with the new memory access approach.

No changes to user-facing features or functionality.

### DIFF
--- a/arch/generic/crc32_chorba_c.c
+++ b/arch/generic/crc32_chorba_c.c
@@ -34,13 +34,15 @@ Z_INTERNAL uint32_t crc32_chorba_118960_nondestructive (uint32_t crc, const z_wo
     ALIGNED_(16) z_word_t bitbuffer[bitbuffersizezwords];
 #endif
     const uint8_t* bitbufferbytes = (const uint8_t*) bitbuffer;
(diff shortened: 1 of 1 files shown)


Dead2:
@samrussell Rebase please? So we can see that the CI succeeds.

Dead2 (approved):

PR #1913, opened 2025-05-12
Add large 1mb buffer test for crc32 hashing.

nmoinvaz:
## Summary by CodeRabbit

- **Tests**
  - Added new tests to verify CRC32 implementations on large 512KB buffers for improved coverage.
  - Extended existing test macros to include large buffer test cases for each CRC32 variant.

### DIFF
--- a/test/test_crc32.cc
+++ b/test/test_crc32.cc
@@ -227,6 +227,29 @@ public:
     }
 };
 
+/* Test large 1MB buffer with known CRC32 */
+class crc32_large_buf : public ::testing::Test {
+protected:
+    static uint8_t *buffer;
+    static const size_t buffer_size = 1024 * 1024;
+
+    static void SetUpTestSuite() {
+        buffer = (uint8_t*)zng_alloc(buffer_size);
+        memset(buffer, 0x55, buffer_size);
+    }
+
+    static void TearDownTestSuite() {
+        zng_free(buffer);
+    }
+
+public:
+    void hash(crc32_func crc32) {
+        EXPECT_EQ(crc32(0, buffer, buffer_size), 0x0026D5FB);
+    }
+};
+
+uint8_t *crc32_large_buf::buffer = nullptr;
+
 INSTANTIATE_TEST_SUITE_P(crc32, crc32_variant, testing::ValuesIn(tests));
 
 #define TEST_CRC32(name, func, support_flag) \
@@ -236,6 +259,13 @@ INSTANTIATE_TEST_SUITE_P(crc32, crc32_variant, testing::ValuesIn(tests));
             return; \
         } \
         hash(GetParam(), func); \
+    } \
+    TEST_F(crc32_large_buf, name) { \
+        if (!(support_flag)) { \
+            GTEST_SKIP(); \
+            return; \
+        } \
+        hash(func); \
     }
 
 #ifndef WITHOUT_CHORBA


test/test_crc32.cc
```diff
@@ -227,6 +227,32 @@ class crc32_align : public ::testing::TestWithParam<int> {
...
+
+public:
+    void hash(crc32_func crc32) {
+        const uint32_t actual_hash = crc32(0, buffer, buffer_size);
+        EXPECT_EQ(actual_hash, expected_hash);
+    }
+};
+
+const uint32_t crc32_large_buf::expected_hash;
+uint8_t *crc32_large_buf::buffer = nullptr;
```
mtl1979: I don't see why these shouldn't be declared and initialized inside the class like `buffer_size` is. Also, `crc32_large_buf::expected_hash` is forward declaration that happens after the real declaration, which makes it redundant. The way `buffer` is initialized here can result in memory leak if `SetupTestSuite` is run more than once.

nmoinvaz: I’m not really a C++ programmer so..

nmoinvaz: > The way buffer is initialized here can result in memory leak if SetupTestSuite is run more than once.

That would be a bug in Google test framework if it happened.

nmoinvaz: Regarding forward declaring, the linker just complained and I did what it said so idk. 🤷‍♂️

nmoinvaz: Maybe I need `constexpr` but I didn't want to be the first to introduce it.

mtl1979: > > The way buffer is initialized here can result in memory leak if SetupTestSuite is run more than once.
> 
> That would be a bug in Google test framework if it happened.

Google test framework does quite a lot of voodoo that breaks on some compilers, so I wouldn't be surprised there is some bugs or code that rely on undefined behaviour in some compilers.

mtl1979: > I’m not really a C++ programmer so..

Neither am I... I started with Z80 machine code and moved on to Pascal... I'm just saying that if I can't understand the logic, then most likely new contributors can't either... Without knowing the exact error messages, it will be hard to know if there is better solution... It's my personal opinion, but I really don't think we should introduce code that gets reverted or replaced right after it is committed... My professional opinion is that since this fails Emscripten WASM32 CI in a way that is unexpected, it should be rewritten...

According to clang 19, the construct is valid for C++17, but we set C++11 as minimum...

mtl1979:
``` diff
diff --git a/test/test_crc32.cc b/test/test_crc32.cc
index 00abd5f6..dc8c497e 100644
--- a/test/test_crc32.cc
+++ b/test/test_crc32.cc
@@ -232,7 +232,7 @@ class crc32_large_buf : public ::testing::Test {
 protected:
     static uint8_t *buffer;
     static const size_t buffer_size = 1024 * 1024;
-    static const uint32_t expected_hash = 0x0026D5FB;
+    static const uint32_t expected_hash;

     static void SetUpTestSuite() {
         buffer = (uint8_t*)zng_alloc(buffer_size);
@@ -250,7 +250,8 @@ public:
     }
 };

-const uint32_t crc32_large_buf::expected_hash;
+// inline static declaration requires C++17
+const uint32_t crc32_large_buf::expected_hash = 0x0026D5FB;
 uint8_t *crc32_large_buf::buffer = nullptr;

 INSTANTIATE_TEST_SUITE_P(crc32, crc32_variant, testing::ValuesIn(tests));
```

mtl1979: Still unsure why `buffer_size` doesn't trigger warning or error...

nmoinvaz: Grok told me it may be that `EXPECT_EQ` macro is referencing the `expected_hash` variable in such a way that it takes the address of it. So that is possibly why it wants it for `expected_hash` but not `buffer_size`. I don't care enough to find out for sure, but it seems plausable.

mtl1979: I'm not really familiar with static const class members as I haven't really used C++17 features much... I did briefly read the relevant section of the standard, but that's quite confusing...

In my own code, I usually just use global non-member variables when static class members need to access same variables as non-static class members and there is no instance of the class that could be passed to the static member function.

nmoinvaz:
@Dead2 this is ready to merge.

Dead2 (approved):

PR #1915, opened 2025-05-20
[WebAssembly] Fix stack overflow in crc32_chorba_118960_nondestructive.

mtl1979:
* Currently default stack size on emscripten is 64 kB, which is too small for Chorba.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility with Emscripten by updating memory allocation for temporary buffers, ensuring stable operation in web environments.


mtl1979:
Emscripten doesn't by default need NULL check for allocation failure as it will abort on out-of-memory condition.

nmoinvaz (approved):

Dead2 (approved):

PR #1916, opened 2025-05-20
manually align and reduce memsize internal_state for 64-bit platforms

GermanAizek:
Using pahole memory analyzer from Red Hat Linux. Struct's reserve field can be removed. I could comment field that it is easier for the compiler to work with structures where there is no explicit definition of fields. But I thought I could break alignment in arm and x86.
- reduce 6144 bytes to 6024 bytes because move field reserved for future to arm and x86
- remove aligned keywords

## Summary by CodeRabbit

- **Refactor**
  - Simplified struct alignment annotations and adjusted conditional inclusion of padding fields for improved consistency.

### DIFF
--- a/deflate.h
+++ b/deflate.h
@@ -142,7 +142,7 @@ typedef struct deflate_allocs_s {
     Pos             *head;
 } deflate_allocs;
 
-struct ALIGNED_(64) internal_state {
+struct internal_state {
     PREFIX3(stream)      *strm;            /* pointer back to this zlib stream */
     unsigned char        *pending_buf;     /* output still pending */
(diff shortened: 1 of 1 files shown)


Dead2:
I don't see any reason to do this unfortunately.

The aligned hints are there to let the compiler know that the pointers point to aligned buffers, enabling aligned reads/writes where possible.

The reserved bytes are there to let us add/remove parameters without changing the size of the struct, thereby avoiding changes to the size of our alloc calls. Some applications allocate those up-front and expect those not to change (Nginx for example). Additionally the reserved bytes pad the struct out to end on a cacheline boundary.

PR #1919, opened 2025-06-06
Fix cyclic sharedlibdir variable in zlib.pc.in for autotools builds.

### LINKED ISSUE #1918: Stray "-L" in pkgconfig file causes cURL build (and probably others) to fail!
The **`zlib.pc`** generated by zlib-ng looks like this:
```
...
Requires:
Libs: -L${libdir} -L${sharedlibdir} -lz
```

...but since `${sharedlibdir}` apparently isn't defined anywhere (i.e. expands to an empty string), we get a *stray* `-L` flag 😩 

(Actually, `sharedlibdir` is defined as `${bindir}`, the latter of which is blank/undefined)

**The result:**
```
pkgconf --libs zlib
-LC:/Source/cURL-build-win32/build/x86/_deps/lib -L -lz
                                                  ^
                                                  └─ Whoops !!!!
```

This in turn causes cURL build to fail, because `libtool` isn't happy!
```
/bin/sh ../libtool  --tag=CC   --mode=link gcc  -fvisibility=hidden -march=i586 -mtune=generic -isystem /c/Source/cURL-build-win32/build/x86/_deps/include -Werror-implicit-function-declaration -O2 -Wno-system-headers   -version-info 12:0:8 -no-undefined   -export-symbols-regex '^curl_.*'  -lnghttp3 -lngtcp2_crypto_ossl -lngtcp2 -lnghttp2 -lidn2 -lrtmp -lz -lssl -lcrypto -lssh2 -lws2_32 -lssl -lcrypto -lz -lgsasl -lpsl -lbcrypt -ladvapi32 -lcrypt32 -lssl -lcrypto -lssl -lcrypto -lgdi32 -lwldap32 -lzstd -lzstd -lbrotlidec -lbrotlidec -lz -lws2_32 -liconv -lcrypt32 -lwinmm -lbrotlicommon -mconsole -Wl,--trace -static -no-pthread -L/c/Source/cURL-build-win32/build/x86/_deps/lib -LC:/Source/cURL-build-win32/build/x86/_deps/lib -L -LC:/Source/cURL-build-win32/build/x86/_deps/lib [...]
libtool:   error: require no space between '-L' and '-LC:/Source/cURL-build-win32/build/x86/_deps/lib'
```

This was tested on Windows with latest mingw-w64 (msys2) and using zlib-ng [2.2.4](https://github.com/zlib-ng/zlib-ng/releases/tag/2.2.4).

My config:
```
./configure --zlib-compat --static --without-optimizations
```

Regards.

satmandu:
Fixes #1918
Syncs with value in https://github.com/zlib-ng/zlib-ng/blob/stable/zlib.pc.cmakein :
https://github.com/zlib-ng/zlib-ng/blob/860e4cff7917d93f54f5d7f0bc1d0e8b1a3cb988/zlib.pc.cmakein#L5

## Summary by CodeRabbit

- **Chores**
  - Updated the directory path used for shared libraries in the configuration template to improve consistency with standard library locations.

### DIFF
--- a/zlib.pc.in
+++ b/zlib.pc.in
@@ -2,7 +2,7 @@ prefix=@prefix@
 exec_prefix=@exec_prefix@
 symbol_prefix=@symbol_prefix@
 libdir=@libdir@
-sharedlibdir=@sharedlibdir@
+sharedlibdir=@libdir@
 includedir=@includedir@
 
 Name: zlib@SUFFIX@


Dead2:
Should perhaps libdir be set to the same as cmake does?
Otherwise the configure workflow now sets -L with the same path twice, and that seems unnecessary and possibly wrong(?).

CMake:
```
libdir=${exec_prefix}/lib
sharedlibdir=${libdir}
*snip*
Libs: -L${libdir} -L${sharedlibdir} -lz-ng
```

Configure after PR:
```
libdir=@libdir@
sharedlibdir=@libdir@
*snip*
Libs: -L${libdir} -L${sharedlibdir} -lz@SUFFIX@
```

satmandu:
> Should perhaps libdir be set to the same as cmake does? Otherwise the configure workflow now sets -L with the same path twice, and that seems unnecessary and possibly wrong(?).
> 

AFAIK, listing an `-L<path>` entry twice during builds, while likely being bad form, doesn't break the build.

> CMake:
CMake uses `@PC_LIB_INSTALL_DIR@` for `libdir`, which would expand to `${exec_prefix}/lib` or `${exec_prefix}/lib64` depending upon the system. So I'm wary of being too clever.

Maybe the more correct fix would be to remove `sharedlibdir` entirely  from both `zlib.pc.in` and `zlib.pc.cmakein` if it isn't used? I haven't done a deeper dive into that, though.

phprus:
@Dead2 
I think this bug is more complex.

The ``sharedlibdir`` value should be set in ``zlib.pc`` by the configure script:

https://github.com/zlib-ng/zlib-ng/blob/2c98ece180e1b541f0311220e38937dd0a94d08e/configure#L2552

Additional information is needed on the issue #1918

phprus:
Fix and description: #1920

PR #1924, opened 2025-06-18
Simplify compatibility intrinsics for x86

phprus:
1) Switched from Microsoft-specific intrinsic names to common (gcc, msvc, clang, intel) names.
2) Removed macros for clang.
3) Replaced ``!defined(__x86_64__)`` with ``defined(__i386__)`` for compatibility with MCST LCC.

## Summary by CodeRabbit

- **Refactor**
  - Updated naming conventions for certain internal operations to improve consistency and clarity. No changes to application behavior or user-facing functionality.

### DIFF
--- a/arch/x86/chorba_sse2.c
+++ b/arch/x86/chorba_sse2.c
@@ -33,7 +33,7 @@ Z_INTERNAL uint32_t chorba_small_nondestructive_sse2(uint32_t crc, const uint64_
     uint64_t next4 = 0;
     uint64_t next5 = 0;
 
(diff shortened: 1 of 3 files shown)


mtl1979:
Initially those X-suffixed versions were supposed to differentiate between versions using GPR (suffix) and MMX registers (no suffix) for 64-bit integers, but not all instructions had MMX versions, so there was two versions using GPR registers instead. As time has passed, most compilers don't even try to use MMX registers anymore as there is SSE equivalent versions and 64-bit processors have no need for MMX registers for 64-bit operations.

Dead2 (approved):
LGTM

PR #1927, opened 2025-07-04
Replace usage of aligned alloc implementations with malloc

Dead2:
Removes detection and usage of aligned alloc implementations and instead use malloc and handle alignment internally.

Since we already always have to do the alignment checks (because we have to support external alloc implementations), this does not cost us anything but does simplify maintenance, especially compatibility across various operating systems, C libraries and compilers.

## Summary by CodeRabbit

* **Refactor**
  * Simplified memory allocation logic to always use standard allocation functions, removing all platform-specific aligned memory allocation checks and implementations.
  * Updated related documentation and comments for clarity.

* **Style**
  * Improved comments to clarify default memory allocation behavior.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -457,49 +457,6 @@ endif()
 set(CMAKE_REQUIRED_FLAGS)
 
 #
-# Check for aligned memory allocation support: POSIX
-#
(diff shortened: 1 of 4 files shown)


mtl1979:
Lower project coverage here is mainly due to not testing with invalid parameters, which might warrant separate PR to address as the affected lines are outside the changes in this PR. Otherwise the PR looks pretty clean.

PR #1930, opened 2025-07-08
Add .gitignore to allow run tests with zlib-ng/corpora and local dataset from working copy

phprus:
Add `.gitignore` to allow run tests with zlib-ng/corpora and local dataset from working copy

Without `.gitignore`, the local test dataset will make it difficult working with git.

## Summary by CodeRabbit

* **Chores**
  * Updated the ignore list to exclude additional test data directories from version control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -103,3 +103,7 @@ gzread.c
 MinSizeRel
 RelWithDebInfo
 /_deps/googletest*
+
+# Test datasets
+test/data/corpora/
+test/data/local/


nmoinvaz (approved):

Dead2:
I am thinking perhaps it would be better to have this in the .gitignore file in the root folder. Keeping gitignores in a single location makes it easier to maintain and inspect.

Also, @coderabbitai had a good suggestion that the lines should end with a slash, so that they specifically match a folder and not files. This is not likely to actually cause any problems, but would just make it more readable.

phprus:
@Dead2

Your suggestions have been applied.

Dead2 (approved):
LGTM

PR #1937, opened 2025-07-12
MSVC: Enable CFG

matbech:
Enable Control Flow Guard (CFG) for MSVC builds.
Reference: https://learn.microsoft.com/en-us/cpp/build/reference/guard-enable-control-flow-guard?view=msvc-170
There is no performance impact as indirect calls are infrequent.

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration to enhance security on Windows platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -648,6 +648,7 @@ if(MSVC)
     set(CMAKE_DEBUG_POSTFIX "d")
     add_definitions(-D_CRT_SECURE_NO_DEPRECATE)
     add_definitions(-D_CRT_NONSTDC_NO_DEPRECATE)
+    add_compile_options(/guard:cf)
 endif()
 
 #


phprus:
> There is no performance impact as indirect calls are infrequent.

zlib-ng uses indirection calls (function pointers) to do optimized functions.
I think performance testing is needed.

KungFuJesus:
So I _think_ the way this works, at least from what I recall in clang, is this is analyzed at compile time.  Something gets decorated in the binary for what indirect calls are allowed and from where.  It's supposed to be relatively cheap.  
https://clang.llvm.org/docs/ControlFlowIntegrityDesign.html

The Microsoft implementation sounds like it embeds something in the read only sections of the binary for this.  The clang version kind of sounds like it puts a bunch of code to check against known valid locations at the call site.

Dead2:
I am not sure about this one..

Lets consider Linux distros for a second. Most distros prefer some protection settings or other, and therefore we let them specify what they want. That way we don't have to consider tradeoffs of performance vs prevention of misdeeds or keeping on top of what protections are best, and available for each cpu architecture, compiler version, etc. Also some systems are embedded or otherwise closed off from the internet or other sources of intrusion, and thus might prefer not to sacrifice performance on this.

For windows, this is a little different of course, but I still hesitate to start maintaining the protection switches.

mtl1979:
At the very least, this should be an optional feature that is off by default... Most of these "protection" switches should eventually be handled by CMake's toolchain file for MSVC instead of individual project files as there is very little benefit if only one "subproject" has the protection enabled.

Dead2:
Closing as there has been no progress and there does not seem like there will be any.

PR #1942, opened 2025-07-22
Disable NEON workaround on Clang 20 and above, and enable it for non-mobile platforms

Un1q32:
This seems like it reduces performance, (if not, maybe it should be always applied?), and it isn't needed after Clang 20 so disable it if Clang is new enough.

Also enable the workaround when building with older Clang versions targeting non-mobile platforms like normal Linux.  The original issue had discussion of Alpine Linux breaking and testing with qemu-user emulation confirms that builds with Clang 19 are broken without the workaround on Alpine.

## Summary by CodeRabbit

* **Refactor**
  * Updated internal logic to improve compatibility of NEON intrinsics with older Clang versions on 32-bit ARM devices. No changes to user-facing features or functionality.

### DIFF
--- a/arch/arm/neon_intrins.h
+++ b/arch/arm/neon_intrins.h
@@ -27,8 +27,8 @@
 
 #  if defined(__arm__) && defined(__clang__) && \
     (!defined(__clang_major__) || __clang_major__ < 20)
-/* Clang versions before 20 have too strict of an alignment
- * requirement (:256) for x4 NEON intrinsics */
+/* Clang versions before 20 have too strict of an
+ * alignment requirement (:256) for x4 NEON intrinsics */
 #    undef ARM_NEON_HASLD4
 #    undef vld1q_u16_x4
 #    undef vld1q_u8_x4
--- a/arch/arm/neon_intrins.h
+++ b/arch/arm/neon_intrins.h
@@ -26,9 +26,9 @@
 } while (0)
 
 #  if defined(__arm__) && defined(__clang__) && \
-    (!defined(__clang_major__) || __clang_major__ < 20) && \
-    (defined(__ANDROID__) || defined(__APPLE__))
-/* Android & iOS have too strict alignment requirement (:256) for x4 NEON intrinsics */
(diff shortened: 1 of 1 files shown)


Un1q32:
220-224: Consider narrowing the Apple platform detection.

    The current condition uses __APPLE__ which covers all Apple platforms (macOS, iOS, tvOS, watchOS). If the performance issue is specific to iOS as mentioned in the PR objectives, consider using more specific macros like TARGET_OS_IOS (requires #include <TargetConditionals.h>) to avoid unnecessarily restricting macOS builds.

    However, if testing shows the issue affects all Apple ARM platforms, the current broad condition is appropriate.


The only platform that can define both `__APPLE__` and `__arm__` besides iOS is watchOS, which is based on iOS so almost certainly also affected.

mtl1979:
If Clang 19 breaks a workaround that works for older versions, we should just disable the workaround for Clang 19 and later instead of doing the complete opposite. In current state, it seems this PR introduces a lot of dead code.

Un1q32:
> If Clang 19 breaks a workaround that works for older versions, we should just disable the workaround for Clang 19 and later instead of doing the complete opposite. In current state, it seems this PR introduces a lot of dead code.

I've only tested with Clang 19 and 20, the workaround was supposed to fix an issue affecting all clang versions, but Clang 20 fixed it.  Clang 18, 17, etc all probably also break, I'll test rq to make sure tho.

Un1q32:
Okay maybe I've gone insane because suddenly Clang 19 is working again despite nothing changing from yesterday, I'll just rework the patch to disable the workaround on Clang 20 so it can have better performance I guess.

Un1q32:
This is what I originally imagined this PR as anyway, but then I got issues when building with Clang 19 with the workaround enabled, so I changed it thinking the code had re-broken at some point and was better off disabled.

Dead2:
Looks good to me, but I am no ARM expert.

@ccawley2011 Any comments?

ccawley2011 (approved):

Dead2 (approved):
LGTM

PR #1944, opened 2025-07-29
X86_AVX512VNNI: test _mm256_dpbusd_epi32 too

haampie:
On RHEL9 the GCC is new enough to support AVX512-VNNI, but its assembler
(binutils) is not and errors with

```
Error: unsupported instruction vpdpbusd
```

This was already addressed earlier in #1562 to some extent, except that
a check for `_mm256_dpbusd_epi32` was not added, which is what the
assembler errors over. So, check the 256 bit version as well.

After some more digging, seems that the difference is that when just
`-mavx512vl -mavx512vnni` flags are given, the following assembly is
generated, which is not problematic for the assembler:

```assembly
      vpdpbusd        %ymm0, %ymm0, %ymm0
```

whereas with `-march=sapphirerapids` the following is, which is apparently
problematic for the assembler:

```assembly
      {vex} vpdpbusd  %ymm0, %ymm0, %ymm0
```

## Summary by CodeRabbit

* **Chores**
  * Updated the system check for AVX512-VNNI support to test both 512-bit and 256-bit variants during configuration. This enhances compatibility detection on different hardware platforms. No changes to user-facing features or public interfaces.

### DIFF
--- a/cmake/detect-intrinsics.cmake
+++ b/cmake/detect-intrinsics.cmake
@@ -164,18 +164,17 @@ macro(check_avx512vnni_intrinsics)
     endif()
     # Check whether compiler supports AVX512vnni intrinsics
     set(CMAKE_REQUIRED_FLAGS "${AVX512VNNIFLAG} ${NATIVEFLAG} ${ZNOLTOFLAG}")
-    check_c_source_compiles([=[
-#include <immintrin.h>
-int main(void) {
-    const __m512i z512 = _mm512_setzero_si512();
-    const __m256i z256 = _mm256_setzero_si256();
-    volatile __m512i r512 = _mm512_dpbusd_epi32(z512, z512, z512);
-    volatile __m256i r256 = _mm256_dpbusd_epi32(z256, z256, z256);
-    (void)r512;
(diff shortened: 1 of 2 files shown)


nmoinvaz (approved):

cmake/detect-intrinsics.cmake
```diff
@@ -164,13 +164,18 @@ macro(check_avx512vnni_intrinsics)
...
+int main(void) {
+    const __m512i z512 = _mm512_setzero_si512();
+    const __m256i z256 = _mm256_setzero_si256();
+    volatile __m512i r512 = _mm512_dpbusd_epi32(z512, z512, z512);
+    volatile __m256i r256 = _mm256_dpbusd_epi32(z256, z256, z256);
+    (void)r512;
+    (void)r256;
+    return 0;
+}
+]=]
```
mtl1979: Quoting style doesn't match rest of the file.

haampie: Fixed

Dead2 (approved):
LGTM

PR #1947, opened 2025-08-16
Check the proper bit for BMI2

KungFuJesus:
We were actually checking for BMI1 support here. This is unlikely to have caused any issues because to date there have not been any x86 CPUs with AVX2 support but no BMI2 support.

## Summary by CodeRabbit

- Bug Fixes
  - Corrected CPU capability detection on x86 systems to accurately recognize BMI2 and AVX-512 support. This ensures features are enabled only when supported, reducing potential illegal-instruction errors and improving stability.
  - Users on affected CPUs may see more reliable startup, fewer compatibility issues, and performance improvements where advanced instructions are now correctly utilized.


Dead2 (approved):
LGTM

PR #1948, opened 2025-08-16
Retemplatize the elided copy with the adler checksum

KungFuJesus:
At some point, GCC became too clever and stopped compiling this compile time constant input as a separately inlined function. What this meant was both implementations suffered a branch when compiled, resulting in slower performance for both our internal usage and well as external callers who simply want to checksum a buffer.

This change effectively reverts that behavior so now the compiler can compile out this branch again, making the loop just a tad bit faster.

## Summary by CodeRabbit

- Performance Improvements
  - Faster Adler-32 checksum throughput on modern x86 CPUs with AVX2/AVX-512, especially for large buffers and combined copy+checksum paths.
  - Reduced latency and more consistent performance across varying input sizes.

- Refactor
  - Consolidated vectorized logic into shared templates, simplifying the codebase while preserving existing behavior and compatibility.

- Stability
  - Streamlined code paths reduce divergence across CPU feature levels, improving reliability without changing public interfaces.

### DIFF
--- a/arch/x86/adler32_avx2.c
+++ b/arch/x86/adler32_avx2.c
@@ -10,136 +10,18 @@
 #ifdef X86_AVX2
 
 #include "zbuild.h"
-#include <immintrin.h>
-#include "adler32_p.h"
-#include "adler32_avx2_p.h"
-#include "x86_intrins.h"
 
-extern uint32_t adler32_fold_copy_sse42(uint32_t adler, uint8_t *dst, const uint8_t *src, size_t len);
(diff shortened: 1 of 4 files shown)


KungFuJesus:
Hmm, I may close this PR, this apparently doesn't seem to be the case anymore with GCC 15.  Weird....

KungFuJesus:
Yeah, even on the machine I saw it on I can't reproduce it since updating.  Perhaps this was a one off weird gcc regression or just a weird combination of compilation flags that manage to make this happen.  Closing this for now.

PR #1956, opened 2025-08-22
Add benchmark for insert_string.

nmoinvaz:
## Summary by CodeRabbit

- Tests
  - Added comprehensive performance benchmarks for insertion routines, covering both standard and quick paths across multiple input sizes.
  - Improved determinism and timing control for consistent, reproducible measurements.
  - Integrated new benchmarks into the existing benchmark suite/target.

- Chores
  - Updated benchmark build configuration to include the new tests.
  - No changes to public interfaces or user-facing behavior.

### DIFF
--- a/test/benchmarks/benchmark_insert_string.cc
+++ b/test/benchmarks/benchmark_insert_string.cc
@@ -1,0 +1,164 @@
+/* benchmark_insert_string.cc -- benchmark insert_string variants
+ * Copyright (C) 2025 Nathan Moinvaziri
+ * For conditions of distribution and use, see copyright notice in zlib.h
+ */
+
+#include <limits.h>
+#include <cstring>
+
+#include <benchmark/benchmark.h>
+
+extern "C" {
+#  include "zbuild.h"
+#  include "zutil_p.h"
+#  include "deflate.h"
+#  include "arch_functions.h"
+#  include "../test_cpu_features.h"
+}
+
+#define MAX_WSIZE 32768
+#define TEST_WINDOW_SIZE (MAX_WSIZE * 2)
+
+// Base class with common setup/teardown for both insert_string benchmarks
+class insert_string_base: public benchmark::Fixture {
+protected:
+    deflate_state *s;
+
+public:
+    void SetUp(const ::benchmark::State& state) {
+        s = (deflate_state*)zng_alloc(sizeof(deflate_state));
+        memset(s, 0, sizeof(deflate_state));
+
+        // Set up window parameters
+        s->w_size = MAX_WSIZE;
+        s->w_bits = 15;
+        s->w_mask = MAX_WSIZE - 1;
+        s->window_size = TEST_WINDOW_SIZE;
+
+        // Allocate window
+        s->window = (uint8_t*)zng_alloc(TEST_WINDOW_SIZE);
+
+        // Allocate hash tables
+        s->head = (Pos*)zng_alloc(HASH_SIZE * sizeof(Pos));
+        s->prev = (Pos*)zng_alloc(MAX_WSIZE * sizeof(Pos));
+
+        // Initialize hash tables
+        memset(s->head, 0, HASH_SIZE * sizeof(Pos));
+        memset(s->prev, 0, MAX_WSIZE * sizeof(Pos));
+
+        // Initialize rolling hash state for rolling variant
+        s->ins_h = 0;
+
+        // Fill window with deterministic data patterns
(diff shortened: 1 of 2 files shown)


test/benchmarks/benchmark_insert_string.cc
```diff
@@ -0,0 +1,179 @@
...
+        Bench(state, fptr); \
+    } \
+    BENCHMARK_REGISTER_F(insert_string_bench, name) \
+        ->Args({100, 1})        /* Insert 1 position (minimal case) */ \
+        ->Args({100, 16})       /* Insert 16 consecutive positions */ \
+        ->Args({100, 64})       /* Insert 64 consecutive positions */ \
+        ->Args({100, 256})      /* Insert 256 consecutive positions */ \
+        ->Args({100, 4096})     /* Insert 4KB consecutive positions */ \
+        ->Args({100, 16384})    /* Insert 16KB consecutive positions (half window) */ \
+        ->Args({100, 32765})    /* Insert nearly full window */ \
```
Dead2: **note:** I actually don't think it is ever used for such big inserts. I stuck a printf in there, and it printed mostly 3 and 7, some 8 and sprinkled with anything from 4 to ~50-70 or something like that.

Probably data dependent, but might be worth looking into to make sure the benchmark doesn't tell us an optimization is awesome for sizes that never happen. I too assumed fill_window would do these massive inserts.

nmoinvaz: Yeah I had some difficulty determining what sizes to put here and assumed that it must be doing inserts for a full window at some point

nmoinvaz: I have updated the values using distribution of `count` from `silesia.tar` and `silesia-small.tar`:
https://gist.github.com/nmoinvaz/f6de3ca6301a43203247f4e3078cea24

Majority of calls happen with `count = 3`.

KungFuJesus:
Ahh, great idea. These improvements are a bit harder to measure even in the compression benchmark.

test/benchmarks/benchmark_insert_string.cc
Dead2: That is great 👍 

I do wonder whether it would be worth tracing where those short and long ones come from and possibly split it into two functions. One optimized for <10 and one for >14 or whatever values make sense.

Were there none below 3 btw? If that is a rule, then we could perhaps add a comment and assert to that effect and allow future optimizations to assume as much.

nmoinvaz: Yeah there was nothing below 3, at least for deflate medium.

test/benchmarks/benchmark_insert_string.cc
```diff
@@ -0,0 +1,164 @@
...
+        ->Args({100, 32})       /* Transition point */ \
+        ->Args({100, 127})      /* Large cluster around powers of 2 */ \
+        ->Args({100, 255})      /* Near maximum observed values */ \
+        ->Unit(benchmark::kMicrosecond);
+
+// Benchmark the standard integer hash variant
+BENCHMARK_INSERT_STRING(integer_hash, ::insert_string, 1);
+
+// Benchmark the rolling hash variant
+BENCHMARK_INSERT_STRING(rolling_hash, ::insert_string_roll, 1);
```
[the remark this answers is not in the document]
Dead2: This could actually be useful to add as a common define for all benchmarks at some point, would make it more obvious what that parameter means while reading the benchmarks.

Dead2:
It'd be nice if you applied these changes:
```patch
         ->Args({100, 255})      /* Near maximum observed values */ \
-        ->Unit(benchmark::kMicrosecond);
+        ->Unit(benchmark::kNanosecond);

         ->Args({16000, 100})    /* 100 insertions at mid window (different hash distribution) */ \
-        ->Unit(benchmark::kMicrosecond);
+        ->Unit(benchmark::kNanosecond);

```

Since each iteration takes around 300-600ns, this seems more appropriate :)

test/benchmarks/benchmark_insert_string.cc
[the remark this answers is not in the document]
nmoinvaz: Yeah it should probably be a different PR though I think.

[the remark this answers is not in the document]
Dead2: I agree.

nmoinvaz:
Done and rebased.

Dead2 (approved):
LGTM

PR #1957, opened 2025-08-25
Upgrade to GCC 14 for macOS UBSAN ci.

nmoinvaz:
Newer versions of macOS SDK throw _bounds.h error with GCC 13.4 when installed from homebrew.
```sh
/opt/homebrew/Cellar/gcc@13/13.4.0/lib/gcc/13/gcc/aarch64-apple-darwin24/13/include-fixed/_stdio.h:78:10: fatal error: _bounds.h: No such file or directory
   78 | #include <_bounds.h>
```

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -622,11 +622,12 @@ jobs:
 
           - name: macOS GCC UBSAN (ARM64)
             os: macos-latest
-            compiler: gcc-13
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Looks like this issue will be automatically fixed by GitHub today:
https://github.com/Homebrew/homebrew-core/issues/228131#issuecomment-3199052086

PR #1958, opened 2025-08-29
Allow C17 for newer CMake versions

alxvth:
The [C17](https://en.cppreference.com/w/c/17.html) standard is available since [CMake 3.21](https://cmake.org/cmake/help/v3.29/prop_tgt/C_STANDARD.html). Since it pretty much just contains a couple of defect reports, it's probably save to allow it here.
At a quick glance I couldn't find a specific reason why C17 should not be allowed, but maybe there is something?

I'm not familiar with the code base, so I'm not sure if it would be OK to also allow [C23](https://en.cppreference.com/w/c/23.html), as that version actually removes some features.

## Summary by CodeRabbit

* **New Features**
  * Optional support for building with C17 and C23 when using a modern CMake (users can opt in during configuration).

* **Chores**
  * Configuration validation still restricts C standard choices to supported values and yields clearer errors for unsupported selections.
  * Continuous integration expanded to test C17 and C23 across multiple platforms and toolchains.


CMakeLists.txt
```diff
@@ -21,6 +21,9 @@ if(NOT DEFINED CMAKE_C_EXTENSIONS)
     set(CMAKE_C_EXTENSIONS OFF)       # Boolean specifying whether compiler specific extensions are requested
 endif()
 set(VALID_C_STANDARDS "99" "11")
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.21)
```
phprus: `VERSION_GREATER_EQUAL` - cmake >= 3.7, but zlib-ng require cmake 3.5.1+

I would suggest removing the C standard version check.

alxvth: Ah yes, it should be `VERSION_GREATER 3.20` to work with 3.5.1 as well.

alxvth: I couldn't find anything on why the version check is there but assumed that it's there for some reason, so I didn't want to just take it out myself.

mtl1979:
I don't see any reason why we shouldn't allow compiling with more recent C standard as we have already decided that we support at least 3 most recent compiler major versions. Removing features isn't an issue if there is a replacement that works at least as good as the currently used function or feature.

In most cases only extra addition we should require is adding the new C standard in new CI workflow, so we immediately know if there is breaking change.

alxvth:
I added some CI workflows for C17 and C23, please adjust them as necessary.

- C23 is quite well supported [by GCC and to a good extend by Clang](https://en.cppreference.com/w/c/compiler_support/23.html).
- GCC actually defaults to C17 starting from [GCC 8](https://gcc.gnu.org/gcc-8/changes.html) and to C23 starting from [GCC 15](https://gcc.gnu.org/gcc-15/changes.html).
- MSVC [does not even know about C23](https://learn.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=msvc-170#c-standards-support-1), so I didn't add a workflow for that.

Edit: I just noticed that `clang-15`, which is used in the CI, does not know C23 yet (only C2y), see [this example on godbolt](https://godbolt.org/z/jTT4Evj3r).

mtl1979:
Latest MSVC version is too old to have C23 support, but there is chattering about new MSVC release near November 2025... We should see added features when first public betas are available... What comes to the added CI workflows, we can comment on them after they have been approved and finished running. Basically we need to look for any failed tests, be it a busy or idle loop, crash, checksum mismatch, or data corruption.

Dead2:
@alxvth Could you please rebase this PR? Bunch of CI fixes have gone in :)

alxvth:
Rebased, the branch is up-to-date again

Dead2 (approved):
LGTM

alxvth:
Thanks!

PR #1959, opened 2025-09-01
Allow zlib-ng to be built as part of parent cmake project without declaring project()

mattiasljungstrom:
Hi, first of all thanks for a great project. 

This pull request fixes two issues I had when using zlib-ng in my own project. 

First of all I have a cmake setup where all libraries are compiled directly into the main project. Because the zlib-ng project is declaring project() it creates conflicts with existing language setting. Furthermore, changes to the cmake build settings will affect the whole project. The solution is the check if project() has been declared already, and only do changes to build setting if the project is built as stand-alone.

Secondly, I regenerate my cmake often, and the setup of zlib-ng includes many compiler tests which all together are kind of slow, and several that will always be false on MSVC. (These are features that only exist on linux platforms.). I've added `if (NOT MSVC)` around these.

If you prefer this done differently please let me know.

Thanks!

## Summary by CodeRabbit

- New Features
  - Improved compatibility when used as a subproject; behaves correctly whether built standalone or within a larger build.
- Bug Fixes
  - Prevents unintended overrides of parent project settings (e.g., build type and LTO).
  - Skips non-applicable detection checks on MSVC, improving reliability for Windows builds.
- Refactor
  - Streamlined and guarded platform/compiler checks to avoid side effects in mixed environments.
- Chores
  - Hardened global build configuration to respect host project contexts.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,7 +34,13 @@ string(REGEX REPLACE ".*#define[ \t]+ZLIBNG_VERSION[ \t]+\"([-0-9A-Za-z.]+)\".*"
 message(STATUS "ZLIB_HEADER_VERSION: ${ZLIB_HEADER_VERSION}")
 message(STATUS "ZLIBNG_HEADER_VERSION: ${ZLIBNG_HEADER_VERSION}")
 
-project(zlib VERSION ${ZLIB_HEADER_VERSION} LANGUAGES C)
+if (NOT DEFINED PROJECT_NAME)
+    # Only declare zlib project if we're not already part of another cmake project
+    project(zlib VERSION ${ZLIB_HEADER_VERSION} LANGUAGES C)
+    set(ZLIB_PROJECT_DECLARED TRUE)
+else()
+    set(ZLIB_PROJECT_DECLARED FALSE)
+endif()
 
 include(CheckTypeSize)
 include(CheckSymbolExists)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -353,12 +362,15 @@ if(NOT WITH_RUNTIME_CPU_DETECTION)
...
-    endforeach()
+if(ZLIB_PROJECT_DECLARED)
+    # Only change cmake build instructions when building as standalone project
+    if(NOT WITH_NATIVE_INSTRUCTIONS)
+        set(CMAKE_INTERPROCEDURAL_OPTIMIZATION OFF)
+        foreach(_cfg_name IN LISTS CMAKE_CONFIGURATION_TYPES)
+            string(TOUPPER "${_cfg_name}" _cfg_name_uc)
+            set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_${_cfg_name_uc} OFF)
+        endforeach()
+    endif()
```
phprus, line 373: This change is incorrect.
LTO must be disabled for runtime function dispatching to work properly.

mattiasljungstrom: Is this GNU/Unix specific? On MSVC "WITH_NATIVE_INSTRUCTIONS" is always disabled because it's not supported, leading to LTO being disabled.

phprus: This is LTO specific.
If `WITH_NATIVE_INSTRUCTIONS` is disabled, LTO must also be disabled.

MSVC support `WITH_NATIVE_INSTRUCTIONS`:

https://github.com/zlib-ng/zlib-ng/blob/83a84faf7a67cb9926947d6b618f141c8020e8e3/.github/workflows/cmake.yml#L516

but only for AVX and newer instuction sets.

mattiasljungstrom: Ok, I'm curious about this, is there more documentations somewhere why LTO breaks WITH_NATIVE_INSTRUCTIONS disabled? What are the exact errors or faults happening? I've used the modified code with libpng without issues.

My problem is that I need LTO for the rest of my project, so this makes zlib-ng not usable for me. I guess I have to fall back to old zlib if there are subtle errors due to this setup.

phprus: zlib-ng with cmake `project()` allows to disable LTO only for zlib-ng.
LTO can mix code from implementations compiled for different CPU instruction sets.

If your project, like zlib-ng, compiles different files with different CPU instruction sets, you cannot use LTO for that project.

CMake `project()` isolates such settings in a (sub)projects. So I don't see any reason why adding dependencies would require removing `project()` from them.

Andarwinux: > Ok, I'm curious about this, is there more documentations somewhere why LTO breaks WITH_NATIVE_INSTRUCTIONS disabled? What are the exact errors or faults happening? I've used the modified code with libpng without issues.
> 
> My problem is that I need LTO for the rest of my project, so this makes zlib-ng not usable for me. I guess I have to fall back to old zlib if there are subtle errors due to this setup.

This is a gcc problem. clang's ThinLTO and MSVC LTO works fine, and you can force LTO with -DFNO_LTO_AVAILABLE=OFF. We haven't seen any runtime dispatch problems when building zlib-ng this way.

PR #1976, opened 2025-10-02
Use 'block-list' and 'allow-list' terms

jeffhandley:
This was found when running a scanner for unfavorable terms. I kept "gray-listed" in place though since gray areas are OK.

## Summary by CodeRabbit

* Refactor
  * Replaced non-inclusive terminology with inclusive alternatives across detection logic; behavior and outputs unchanged.
* Documentation
  * Updated explanatory text and examples to use the new terminology; no user-facing changes.
* Style
  * Minor wording and readability tweaks in comments and descriptions, preserving identical functionality.

### DIFF
--- a/doc/txtvsbin.txt
+++ b/doc/txtvsbin.txt
@@ -38,15 +38,15 @@ The Algorithm
 
 The algorithm works by dividing the set of bytecodes [0..255] into three
 categories:
-- The white list of textual bytecodes:
+- The allow list of textual bytecodes:
   9 (TAB), 10 (LF), 13 (CR), 32 (SPACE) to 255.
 - The gray list of tolerated bytecodes:
   7 (BEL), 8 (BS), 11 (VT), 12 (FF), 26 (SUB), 27 (ESC).
-- The black list of undesired, non-textual bytecodes:
+- The block list of undesired, non-textual bytecodes:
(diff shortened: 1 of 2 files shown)


nmoinvaz:
I remember an old zlib issue about this. We should see what Mark did.

jeffhandley:
@nmoinvaz I found the corresponding change in zlib. Here's the commit: https://github.com/madler/zlib/commit/8678871f18f4dd51101a9db1e37791f975969079#diff-98dbe62a944e219f1245cf05c7eca4fec56ffbcc992dd3083817f4a720312da2

That led me to find the txtvsbin.txt file having the terms too, and those were updated in zlib so I updated them here too. "gray-list" was left in place there too.

Here's where Mark concluded to make the change: https://github.com/madler/zlib/pull/511#issuecomment-656764961

mtl1979:
I think it's good that we get rid of the wording that do not fit in today's world. As long as the wording doesn't appear in a quote of what someone has said, we don't need to care if Adler already did same change. 

In some parts we should follow what Adler has changed to avoid deviating too much and making importing new patches too much hassle, but there is already other minor differences that require manual resolving of merge conflicts when doing three-way merge.

Dead2 (approved):

PR #1977, opened 2025-10-04
Add Adler32 AVXVNNI support for Intel's bad decisions

KungFuJesus:
Intel, hopping on the ML all the things band wagon, decided that VNNI was just _too_ good to not include on their lesser client CPUs without AVX512. Given how many cycles it saves for the 32 bit widening operation and how close we already were to something that supports this sans AVX512, we may as well include support for it. On a 32k buffer it's about 3 times as fast in my benchmarking on a Raptor Lake on a performance core.

This may need further testing on MSVC, they have no definitive "arch" flag for it as far as I can tell.


KungFuJesus:
Errr, so I thought.  Rebenchmarking that does not currently seem the case.  I think there may need to be more unrolling.

KungFuJesus:
lol, this is somehow faster on the e cores (than the avx2 version) but slower on the p cores.  I might have to make sure there aren't frequency scaling shenanigans afoot.

cmake/detect-intrinsics.cmake
```diff
@@ -139,6 +139,45 @@ macro(check_avx512_intrinsics)
...
+            set(AVX2VNNIFLAG "-mavxvnni")
+            if(NOT MSVC)
+                check_c_compiler_flag("-mtune=raptorlake" HAVE_RAPTOR_LAKE)
+                if(HAVE_RAPTOR_LAKE)
+                    set(AVX2VNNIFLAG "${AVX2VNNIFLAG} -mtune=raptorlake")
+                else()
+                    set(AVX2VNNIFLAG "${AVX2VNNIFLAG} -mtune=skylake")
+                endif()
+                unset(HAVE_RAPTOR_LAKE)
+            endif()
```
mtl1979, line 211: This might cause the failure on package check as configure doesn't seem to add the tune flag and thus reordering doesn't happen.

mtl1979:
MSVC doesn't have full AVX-VNNI support as far as I know... The internal discussion suggest that `/arch:avx2` should be used, as when using `/arch:avx512` the code generation phase is allowed to use AVX-512 instructions that cause crash when not supported by the processor.

MSVC 2022 17.7 is the first release that mentions even partial support for AVX-VNNI, but there is open bug reports for bad code generation.

See: 
1. https://developercommunity.visualstudio.com/t/MSVC-emits-AVX512-code-for-AVX-VNNI-inst/10804143
2. https://devblogs.microsoft.com/cppblog/msvc-backend-updates-since-visual-studio-2022-version-17-3/

cmake/detect-intrinsics.cmake
```diff
@@ -139,6 +139,45 @@ macro(check_avx512_intrinsics)
...
+            const __m256i z256 = _mm256_setzero_si256();
+            volatile __m256i r256 = _mm256_dpbusd_epi32(z256, z256, z256);
+            (void)r256;
+            return 0;
+        }"
+        HAVE_AVX2VNNI_INTRIN
+    )
+    set(CMAKE_REQUIRED_FLAGS)
+endmacro()
+
```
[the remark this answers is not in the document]
mtl1979: Using latest Visual Studio (version 18.0.0 build 11018.127), I get:

`[CMake]  * AVX2VNNI_ADLER32, Support AVX2VNNI adler32, using "/arch:AVX2"`

It also defines `_mm256_dpbusd_epi32` in `zmmintrin.h`.

KungFuJesus:
Ok, finally had some time to address the performance of this thing.  Basically, we need to partially unroll to a factor of about 4x to get the ILP needed for this to be worthwhile.  There's one remaining undesired stack spill at the tail end of the loop that maybe able to be eliminated but so far my attempts to fold in the remaining running elements to save a register iteration to iteration have resulted in the very data dependency bubble we're trying to eliminate.  On my Raptor Lake CPU, it appears we are faster now on both E-cores and P-cores.  Strangely, in https://github.com/zlib-ng/zlib-ng/issues/1985 it was made apparent that these shared mnemonics for the dot products are separately encoded instructions from AVX512 and are incompatible.  Not that this matters, we'd want the wider 512 bit sums for those architectures, anyhow.

# P-Core
```
mythtv@mythtvbox ~/zlib-ng/build_vnni/test/benchmarks $ taskset -c 0 ./benchmark_zlib --benchmark_filter=adler32/avx2
2025-11-08T15:35:46-05:00
Running ./benchmark_zlib
Run on (16 X 4700 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 1280 KiB (x8)
  L3 Unified 20480 KiB (x1)
Load Average: 0.20, 0.06, 0.02
***WARNING*** ASLR is enabled, the results may have unreproducible noise in them.
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
adler32/avx2/1                 2.78 ns         2.78 ns    251501316
adler32/avx2/8                 4.52 ns         4.51 ns    155044309
adler32/avx2/12                5.62 ns         5.61 ns    124453832
adler32/avx2/16                7.25 ns         7.25 ns     96608713
adler32/avx2/32                6.65 ns         6.65 ns    105237708
adler32/avx2/64                7.19 ns         7.19 ns     97325268
adler32/avx2/512               11.7 ns         11.7 ns     59732088
adler32/avx2/4096              59.9 ns         59.9 ns     11711113
adler32/avx2/32768              461 ns          461 ns      1519799
adler32/avx2/262144            3798 ns         3796 ns       184432
adler32/avx2/4194304          78205 ns        78154 ns         8944
adler32/avx2vnni/1             4.49 ns         4.49 ns    155900335
adler32/avx2vnni/8             4.61 ns         4.60 ns    152026278
adler32/avx2vnni/12            5.68 ns         5.55 ns    126309981
adler32/avx2vnni/16            6.92 ns         6.92 ns    101163301
adler32/avx2vnni/32            7.17 ns         7.16 ns     97583298
adler32/avx2vnni/64            8.25 ns         8.24 ns     84856183
adler32/avx2vnni/512           12.5 ns         12.5 ns     55805105
adler32/avx2vnni/4096          52.3 ns         52.3 ns     13350614
adler32/avx2vnni/32768          401 ns          401 ns      1745918
adler32/avx2vnni/262144        3243 ns         3241 ns       215891
adler32/avx2vnni/4194304      69188 ns        69139 ns        10119
```

# E-Core
```
mythtv@mythtvbox ~/zlib-ng/build_vnni/test/benchmarks $ taskset -c 15 ./benchmark_zlib --benchmark_filter=adler32/avx2
2025-11-08T15:36:58-05:00
Running ./benchmark_zlib
Run on (16 X 4700 MHz CPU s)
CPU Caches:
  L1 Data 48 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 1280 KiB (x8)
  L3 Unified 20480 KiB (x1)
Load Average: 0.18, 0.10, 0.03
***WARNING*** ASLR is enabled, the results may have unreproducible noise in them.
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
adler32/avx2/1                 4.33 ns         4.33 ns    161635417
adler32/avx2/8                 7.30 ns         7.30 ns     95914909
adler32/avx2/12                8.89 ns         8.88 ns     78820703
adler32/avx2/16                12.7 ns         12.7 ns     55399760
adler32/avx2/32                11.8 ns         11.8 ns     59220806
adler32/avx2/64                14.8 ns         14.8 ns     47278212
adler32/avx2/512               36.5 ns         36.5 ns     19167458
adler32/avx2/4096               231 ns          231 ns      3037114
adler32/avx2/32768             1836 ns         1835 ns       381324
adler32/avx2/262144           14610 ns        14597 ns        47948
adler32/avx2/4194304         236654 ns       236398 ns         2961
adler32/avx2vnni/1             7.17 ns         7.16 ns     97727312
adler32/avx2vnni/8             7.22 ns         7.22 ns     96950818
adler32/avx2vnni/12            9.10 ns         9.09 ns     76941583
adler32/avx2vnni/16            12.4 ns         12.4 ns     56514161
adler32/avx2vnni/32            12.8 ns         12.8 ns     54603622
adler32/avx2vnni/64            14.3 ns         14.3 ns     48954844
adler32/avx2vnni/512           30.7 ns         30.7 ns     22797723
adler32/avx2vnni/4096           189 ns          189 ns      3707031
adler32/avx2vnni/32768         1513 ns         1512 ns       462995
adler32/avx2vnni/262144       12123 ns        12112 ns        57797
adler32/avx2vnni/4194304     194864 ns       194650 ns         3596
```

KungFuJesus:
Ugh, can't get better throughput on the copy function, though.  For that one, it's faster on E core but slower on P.  Using the same exact loop as the non-copy variant I think results in poor ILP since port utilization is not the problem when there are moves involved.

cmake/detect-intrinsics.cmake
```diff
@@ -139,6 +139,45 @@ macro(check_avx512_intrinsics)
...
+                if(HAVE_RAPTOR_LAKE)
+                    set(AVX2VNNIFLAG "${AVX2VNNIFLAG} -mtune=raptorlake")
+                else()
+                    set(AVX2VNNIFLAG "${AVX2VNNIFLAG} -mtune=skylake")
+                endif()
+                unset(HAVE_RAPTOR_LAKE)
+            endif()
+        elseif(MSVC)
+            set(AVX2VNNIFLAG "/arch:AVX512")
+        endif()
```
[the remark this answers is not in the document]
KungFuJesus: Does MSVC in fact compile in support for AVX-VNNI when building with /arch:AVX2?  Because I need the special 256 bit VNNI instructions that were added on their non-avx512 capable CPUs.

arch/x86/x86_functions.h
```diff
@@ -198,6 +203,13 @@ uint32_t crc32_vpclmulqdq(uint32_t crc32, const uint8_t *buf, size_t len);
 #      undef native_adler32_fold_copy
 #      define native_adler32_fold_copy adler32_fold_copy_avx512_vnni
 #    endif
+// X86 - AVX2 (VNNI)
+#    if defined(X86_AVX2VNNI) && defined(__AVXVNNI__)
+#      undef native_adler32
+#      define native_adler32 adler32_avx2_vnni
+#      undef native_adler32_fold_copy
+#      define native_adler32_fold_copy adler32_fold_copy_avx2
+#    endif
```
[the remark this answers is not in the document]
KungFuJesus: It's not nested, it's just further down.  It is a later feature.  I suppose I can move it up, though.

KungFuJesus:
> MSVC doesn't have full AVX-VNNI support as far as I know... The internal discussion suggest that `/arch:avx2` should be used, as when using `/arch:avx512` the code generation phase is allowed to use AVX-512 instructions that cause crash when not supported by the processor.
> 
> MSVC 2022 17.7 is the first release that mentions even partial support for AVX-VNNI, but there is open bug reports for bad code generation.
> 
> See:
> 
> 1. https://developercommunity.visualstudio.com/t/MSVC-emits-AVX512-code-for-AVX-VNNI-inst/10804143
> 2. https://devblogs.microsoft.com/cppblog/msvc-backend-updates-since-visual-studio-2022-version-17-3/

Hmm, maybe we exclude this from MSVC for now, then?

mtl1979:
> Hmm, maybe we exclude this from MSVC for now, then?

At least until I/we know what version number(s) fix the code generation bug, it might be the best option... Visual Studio 18 (2026) is still in early development (currently Insider preview), so a lot can change in short time.

Dead2:
@KungFuJesus coderabbitai has some suggested changes that you need to consider.
You also need to consider adapting to the changes made by #2048

CMakeLists.txt
```diff
@@ -1714,6 +1727,10 @@ elseif(BASEARCH_LOONGARCH_FOUND)
     add_feature_info(WITH_LSX WITH_LSX "Build with vectorized LSX on LoongArch")
     add_feature_info(WITH_LASX WITH_LASX "Build with vectorized LASX on LoongArch")
 elseif(BASEARCH_X86_FOUND)
+    add_feature_info(WITH_AVX2 WITH_AVX2 "Build with AVX2")
+    add_feature_info(WITH_AVX2VNNI WITH_AVX2VNNI "Build with AVX2VNNI")
```
nmoinvaz, line 1589: I kind of think you should have a `WITH_VNNI` so your current `WITH_AVX2VNNI` will be the equivalent of `WITH_VNNI AND WITH_AX2`, same for AVX512. It would work similar to how VPCLMULQDQ will work in a future PR of mine, it also has a AVX2 and AVX512 variants.

nmoinvaz:
```cmake
cmake_dependent_option(WITH_VNNI "Build with VNNI extensions" ON "WITH_AVX2 OR WITH_AVX512" OFF)
```

nmoinvaz:
@KungFuJesus do you have any intention to revive this?

CMakeLists.txt
KungFuJesus: The problem is that they are two entirely different opcodes, I think. They are incompatible despite sharing the same mnemonic.

KungFuJesus:
> @KungFuJesus do you have any intention to revive this?

Yes, it's just been back burnered due to lack of free time. As is it's currently faster for the pure adler checksum but slower for the copying variant, unless on an e core, and then it's marginally faster to use the copying variant (which is why it's included but not compiled here). Ideally I'd like it to be a win for both variants on both e cores and p cores. Something about the code gcc and clang produce is fighting with the microarchitecture's pipeline. I could rebase this and include just this version for the pure checksum for now if we'd like to get it in soon.

arch/x86/x86_functions.h
```diff
@@ -82,6 +82,11 @@
 uint32_t adler32_copy_avx512_vnni(uint32_t adler, uint8_t *dst, const uint8_t *src, size_t len);
 #endif
 
+#ifdef X86_AVX2VNNI
+uint32_t adler32_avx2_vnni(uint32_t adler, const uint8_t *buf, size_t len);
+//uint32_t adler32_copy_avx2_vnni(uint32_t adler, uint8_t *dst, const uint8_t *src, size_t len);
```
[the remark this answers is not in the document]
KungFuJesus: This is intentional...it's there should someone want to revive this variant if they have an idea for how to make it faster.  It's uncompiled code that we are leaving dark.

configure
mtl1979, line 2404: Here all new architecture flags for x86 should be logged...

mtl1979, line 2552: This block of lines should be updated when support for new x86 intrinsics is added.

mtl1979 (changes requested):
Some modified variables are not propagated correctly from `configure` to `arch/x86/Makefile`.

arch/x86/x86_functions.h
```diff
@@ -168,6 +173,13 @@ uint32_t crc32_copy_vpclmulqdq_avx512(uint32_t crc, uint8_t *dst, const uint8_t
 #    undef native_slide_hash
 #    define native_slide_hash slide_hash_avx2
 #  endif
+// X86 - AVX2 (VNNI)
+#    if defined(X86_AVX2VNNI) && defined(__AVXVNNI__)
+#      undef native_adler32
+#      define native_adler32 adler32_avx2_vnni
+#      undef native_adler32_fold_copy
+#      define native_adler32_fold_copy adler32_fold_copy_avx2
+#    endif
```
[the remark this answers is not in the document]
mtl1979: This is intentional as AVX2VNNI support in MSVC is still work-in-progress, so it should not be enabled.

configure
KungFuJesus: Ahh thank you, that's the glue I was missing. It's been a bit since mucking with compilation flags.

mtl1979: I'm not big fan of CMake, so I use `configure` most of the time... Because these lines or anything close to them were not edited, I had to expand the range of lines GitHub website shows before I could point out what is missing...

configure
```diff
@@ -2347,6 +2397,10 @@ echo ssse3flag = $ssse3flag >> configure.log
 echo sse42flag = $sse42flag >> configure.log
 echo pclmulflag = $pclmulflag >> configure.log
 echo vpclmulflag = $vpclmulflag >> configure.log
+echo avx2flag = $avx2flag >> configure
+echo avx512flag = $avx512flag >> configure
+echo avx512vnniflag = $avx512vnniflag >> configure
+echo avx2vnniflag = $avx2vnniflag >> configure
```
mtl1979: Should have `configure.log`, not `configure`...

mtl1979 (changes requested):
Wrong filename...

cmake/detect-intrinsics.cmake
KungFuJesus: The failure was the reason you speculated earlier - missed that regex in the config script.

arch/x86/adler32_avx2_vnni.c
```diff
@@ -0,0 +1,253 @@
...
+    if (len) {
+        goto rempeel;
+    }
+
+    return adler;
+}
+
+/* There's a data dependency bubble to pop here to maximize throughput but I haven't been able to quite factor
+ * out exactly what gets us there. I'm leaving this here as this is the closest I managed to get. For performance
+ * cores on Raptor Lake, the avx2 copy function is faster. For like for like E-cores, this function is faster. There's
```
mtl1979: "For like for like" seems weird choice of words...

cmake/detect-intrinsics.cmake
```diff
@@ -178,6 +178,43 @@ macro(check_avx512_intrinsics)
...
 endmacro()
 
+macro(check_avx2vnni_intrinsics)
+    if(NOT NATIVEFLAG)
+        if(CMAKE_C_COMPILER_ID MATCHES "Intel")
+            if(CMAKE_HOST_UNIX OR APPLE OR CMAKE_C_COMPILER_ID MATCHES "IntelLLVM")
+                set(AVX2VNNIFLAG "-mavxvnni -mavx2 -mbmi2")
+            else()
+                # It would _seem_ this is the only option for this intrinsic?
+                set(AVX2VNNIFLAG "/arch:AVX512")
```
mtl1979: This is just plain wrong and result in crash... Maybe remove until the support is actually finished.

CMakeLists.txt
mtl1979: Maybe this is something to consider for another PR after this one is merged. Being consistent should not be judged based on future PR that isn't reviewed or merged yet.

cmake/detect-intrinsics.cmake
mtl1979: It took me a while to compare the compiler flags... Then more time to find out why the compiler flags were not propagated to `arch/x86/Makefile`...

cmake/detect-intrinsics.cmake
```diff
@@ -178,6 +178,43 @@ macro(check_avx512_intrinsics)
...
+                if(HAVE_RAPTOR_LAKE)
+                    set(AVX2VNNIFLAG "${AVX2VNNIFLAG} -mtune=raptorlake")
+                endif()
+                unset(HAVE_RAPTOR_LAKE)
+            endif()
+        elseif(MSVC)
+            set(AVX2VNNIFLAG "/arch:AVX512")
+        endif()
+    endif()
+    # Check whether compiler supports AVX512vnni intrinsics
```
mtl1979: Should say "supports AVX2vnni intrinsics".

mtl1979 (changes requested):
Copy pasta

arch/x86/adler32_avx2_vnni.c
KungFuJesus: I thought it was a common English idiom.  I can change it, though.

arch/x86/x86_features.h
```diff
@@ -14,6 +14,7 @@ struct x86_cpu_features {
     int has_avx512vl;
     int has_avx512_common; // Enabled when AVX512(F,DQ,BW,VL) are all enabled.
     int has_avx512vnni;
+    int has_avxvnni;
```
nmoinvaz: This should be `has_avx2vnni` and located directly below `has_avx2` imo, that would match `has_avx512vnni`.

CMakeLists.txt
```diff
@@ -183,6 +183,7 @@ elseif(BASEARCH_X86_FOUND)
     cmake_dependent_option(WITH_AVX512 "Build with AVX512" ON "WITH_AVX2" OFF)
     cmake_dependent_option(WITH_AVX512VNNI "Build with AVX512 VNNI extensions" ON "WITH_AVX512" OFF)
     cmake_dependent_option(WITH_VPCLMULQDQ "Build with VPCLMULQDQ" ON "WITH_PCLMULQDQ;WITH_AVX2" OFF)
+    cmake_dependent_option(WITH_AVX2VNNI "Build with AVX VNNI extensions" ON "WITH_AVX2" OFF)
```
nmoinvaz: Move below `WITH_AVX2` option?

cmake/detect-intrinsics.cmake
```diff
@@ -190,6 +190,44 @@ macro(check_avx512_intrinsics)
...
 endmacro()
 
+macro(check_avx2vnni_intrinsics)
+    if(NOT NATIVEFLAG)
+        if(CMAKE_C_COMPILER_ID MATCHES "Intel")
+            if(CMAKE_HOST_UNIX OR APPLE OR CMAKE_C_COMPILER_ID MATCHES "IntelLLVM")
+                set(AVX2VNNIFLAG "-mavx2 -mavxvnni -mbmi -mbmi2")
+            #else()
+                # It would _seem_ this is the only option for this intrinsic?
+                #set(AVX2VNNIFLAG "/arch:AVX512")
```
mtl1979: This code is likely wrong, we should just add comment that MSVC doesn't support AVX2VNNI correctly yet, and leave the flag empty. Maybe add link to the bug report.

cmake/detect-intrinsics.cmake
```diff
@@ -190,6 +190,44 @@ macro(check_avx512_intrinsics)
...
+            if(NOT MSVC)
+                check_c_compiler_flag("-mtune=raptorlake" HAVE_RAPTOR_LAKE)
+                if(HAVE_RAPTOR_LAKE)
+                    set(AVX2VNNIFLAG "${AVX2VNNIFLAG} -mtune=raptorlake")
+                endif()
+                unset(HAVE_RAPTOR_LAKE)
+            endif()
+        # When AVX2VNNI support is fully baked into MSVC we should uncomment this
+        #elseif(MSVC)
+        #    set(AVX2VNNIFLAG "/arch:AVX512")
```
mtl1979: Leave the flag empty until proper support can be verified. Comment is OK...

mtl1979 (commented):
Needs just final polishing of commented out blocks for incomplete support in MSVC.

cmake/detect-intrinsics.cmake
KungFuJesus: We do leave it empty, that's commented out.  I'll add links to the bug report, though.

mtl1979: I meant literal empty string... Trying to enable AVX512 is definite crash.

nmoinvaz (approved):

nmoinvaz:
Looked at whether any intrinsics could be combined — not really: the `k >= 128` loop already uses all 16 ymm regs with no spills (clang-22 asm), and the critical path is the four `vpdpbusd` chains at ~5c latency, which the 4-accumulator split is there to hide. That register cap is why the AVX-512 path (32 regs) is faster.

One free tweak: `vs3` funnels four adds into one accumulator (a 3-long dep chain). Splitting it in two costs nothing — same instruction count, no new spills:

```diff
@@ -49,6 +49,7 @@
         __m256i vs3 = zero;
+        __m256i vs3b = zero;
@@ -80,11 +80,11 @@
             vs3 = _mm256_add_epi32(vs3, vs1_0a);
-            vs3 = _mm256_add_epi32(vs3, vs1_0b);
+            vs3b = _mm256_add_epi32(vs3b, vs1_0b);
             vs2a = _mm256_dpbusd_epi32(vs2a, vbuf0, dot2v);
             vs2b = _mm256_dpbusd_epi32(vs2b, vbuf1, dot2v_0);
             vs3 = _mm256_add_epi32(vs3, vs1a);
-            vs3 = _mm256_add_epi32(vs3, vs1b);
+            vs3b = _mm256_add_epi32(vs3b, vs1b);
@@ -100,6 +100,7 @@
+        vs3 = _mm256_add_epi32(vs3, vs3b);
         vs3 = _mm256_slli_epi32(vs3, 6);
```

Numerically identical (int32 adds before the `<<6`). Likely neutral on Raptor Cove, maybe a small win on Gracemont. I'm on arm64 so I only verified the asm, not benchmarked it — worth an A/B if you're curious, otherwise ignore.

KungFuJesus:
> Looked at whether any intrinsics could be combined — not really: the `k >= 128` loop already uses all 16 ymm regs with no spills (clang-22 asm), and the critical path is the four `vpdpbusd` chains at ~5c latency, which the 4-accumulator split is there to hide. That register cap is why the AVX-512 path (32 regs) is faster.
> 
> One free tweak: `vs3` funnels four adds into one accumulator (a 3-long dep chain). Splitting it in two costs nothing — same instruction count, no new spills:
> 
> ```diff
> @@ -49,6 +49,7 @@
>          __m256i vs3 = zero;
> +        __m256i vs3b = zero;
> @@ -80,11 +80,11 @@
>              vs3 = _mm256_add_epi32(vs3, vs1_0a);
> -            vs3 = _mm256_add_epi32(vs3, vs1_0b);
> +            vs3b = _mm256_add_epi32(vs3b, vs1_0b);
>              vs2a = _mm256_dpbusd_epi32(vs2a, vbuf0, dot2v);
>              vs2b = _mm256_dpbusd_epi32(vs2b, vbuf1, dot2v_0);
>              vs3 = _mm256_add_epi32(vs3, vs1a);
> -            vs3 = _mm256_add_epi32(vs3, vs1b);
> +            vs3b = _mm256_add_epi32(vs3b, vs1b);
> @@ -100,6 +100,7 @@
> +        vs3 = _mm256_add_epi32(vs3, vs3b);
>          vs3 = _mm256_slli_epi32(vs3, 6);
> ```
> 
> Numerically identical (int32 adds before the `<<6`). Likely neutral on Raptor Cove, maybe a small win on Gracemont. I'm on arm64 so I only verified the asm, not benchmarked it — worth an A/B if you're curious, otherwise ignore.

Yeah, that did something worthwhile.  Not discernibly better on Raptor Cove but ~3% gain on Gracemont.

nmoinvaz (approved):

PR #1978, opened 2025-10-06
Tests: Fix cast and truncation warnings.

mtl1979:
* `compressed_sizes` was unsigned long, but it was assigned `size_t` which is larger on MSVC
* `l0` and `l1` are both `uint16_t` but `rand()` returns `int` which is 32-bit type

## Summary by CodeRabbit

* **Tests**
  * Made random data initialization explicit to avoid implicit casts in a benchmark, improving consistency.
  * Switched compressed size tracking to a larger integer type in another benchmark to prevent overflow and improve accuracy.
  * Enhances reliability and consistency of benchmark results across platforms.
  * No changes to runtime features; this affects only internal testing and does not impact end-user workflows.


mtl1979:
@coderabbitai generate docstrings

Dead2 (approved):

PR #1980, opened 2025-10-09
Fix type mismatch on platforms where int32_t and uint32_t use long instead of int

mtl1979:
Introduce two new private types, `z_int32_t`and `z_uint32_t` in `zbuild.h` to be used in C source files during compiling zlib-ng when the API differs between zlib-compat and zlib-ng, and toolchain use `long` instead of `int` for `int32_t`; and `unsigned long` instead of `unsigned int` for `uint32_t`.

* Based on PR #1934

### DIFF
--- a/compress.c
+++ b/compress.c
@@ -28,8 +28,8 @@
    memory, Z_BUF_ERROR if there was not enough room in the output buffer,
    Z_STREAM_ERROR if the level parameter is invalid.
 */
-int Z_EXPORT PREFIX(compress2)(unsigned char *dest, z_uintmax_t *destLen, const unsigned char *source,
-                        z_uintmax_t sourceLen, int level) {
+z_int32_t Z_EXPORT PREFIX(compress2)(unsigned char *dest, z_uintmax_t *destLen, const unsigned char *source,
+                        z_uintmax_t sourceLen, z_int32_t level) {
     PREFIX3(stream) stream;
     int err;
     const unsigned int max = (unsigned int)-1;
@@ -70,7 +70,7 @@ int Z_EXPORT PREFIX(compress2)(unsigned char *dest, z_uintmax_t *destLen, const 
 
 /* ===========================================================================
  */
-int Z_EXPORT PREFIX(compress)(unsigned char *dest, z_uintmax_t *destLen, const unsigned char *source, z_uintmax_t sourceLen) {
+z_int32_t Z_EXPORT PREFIX(compress)(unsigned char *dest, z_uintmax_t *destLen, const unsigned char *source, z_uintmax_t sourceLen) {
     return PREFIX(compress2)(dest, destLen, source, sourceLen, Z_DEFAULT_COMPRESSION);
 }
 
--- a/zbuild.h
+++ b/zbuild.h
@@ -108,6 +108,17 @@
 #  define z_uintmax_t size_t
 #endif
 
+/* In zlib-compat headers some function return values and parameter types use int or unsigned, but zlib-ng headers use
+   int32_t and uint32_t, which will cause type mismatch when compiling zlib-ng if int32_t is long and uint32_t is
+   unsigned long */
+#if defined(ZLIB_COMPAT)
+#  define z_int32_t int
+#  define z_uint32_t unsigned int
+#else
+#  define z_int32_t int32_t
+#  define z_uint32_t uint32_t
+#endif
+
 /* Minimum of a and b. */
 #define MIN(a, b) ((a) > (b) ? (b) : (a))
 /* Maximum of a and b. */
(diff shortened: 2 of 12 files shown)


nmoinvaz:
In PR https://github.com/zlib-ng/zlib-ng/pull/1934, I noticed `zlib-ng.h.in` was changed. But in this PR, that file is not changed. Are those changes still needed or does this PR superceed #1934?

mtl1979:
> In PR #1934, I noticed `zlib-ng.h.in` was changed. But in this PR, that file is not changed. Are those changes still needed or does this PR superceed #1934?

This PR pretty much supercedes #1934 as changing existing function signatures in `zlib-ng.h.in` is a bad idea as it breaks the externally visible API.

zbuild.h
```diff
@@ -108,6 +108,15 @@
 #  define z_uintmax_t size_t
 #endif
 
+/* In zlib-compat some functions and types use int or unsigned, but zlib-ng use int32_t, which might be long, and uint32_t, which might be unsigned long on some platforms */
```
Dead2: **suggestion:** This could be understood to mean (u)int32_t can be of variable size, while the reverse it true.
Both int and long might be 16/32/64 bits depending on the platform, while (u)int32_t is always 32 bits and ensures predictable sizes.
Maybe rewrite it a little to clarify?
Perhaps something like this? `In zlib-compat many functions and types use int or long or their unsigned equivalents, but zlib-ng instead prefers int32_t and uint32_t to avoid sizes differing between platforms.`

mtl1979: Not really... This block isn't about zlib-compat using `long`, instead it is about zlib-ng (without zlib-compat) using `long` when it should use `int` as underlying type of `int32_t`. As said earlier, using `long` anywhere is relic from past when `int` was only 16 bits, but zlib-ng dropped support for those architectures quite a long time ago.

When `int` and `long` are both 32 bits, it makes no difference to most compilers which one is used in C (not C++), unless using strict mode, as the encoded function signature is same, only underscore is prepended to function name.

Dead2: I understand the technical side of it. I know the stdint types usually map to int/long, although I think it is more relevant to focus on the fact that we want to use a specific length, irrelevant of whether that is mapped to int or long by the compiler.

So what I am saying here is that the comment as-is is can easily be misunderstood. I can in no way interpret that comment as what you just stated that zlib-ng native is using long instead of int when it specifies int32_t, and I also don't see how that is important information for the code block.

If you want to clarify it differently from my suggestion that is fine, but it needs to be improved upon, or removed if you disagree that it could be misunderstood. For example, the AI reviewers would take this comment as a golden fact and it could potentially muddy their future reviews.

mtl1979: I'm all for improving the comment as long as it doesn't change the meaning of the comment and cause the comment to no longer match the intent of the code block below it... Having no comments will pretty much result in AI bots flagging the PR as incomplete as seen on other recent PRs.

What comes to AI not understanding the comment or code block under it, the whole edit to this file was written 75% by AI in Visual Studio 18, except I added the unsigned version to it. Usually I dislike using AI, but when testing new features in a compiler, it doesn't hurt if it thinks the same way as I do.

mtl1979:
Rewrote the comment in `zbuild.h`... The comment is now a lot longer, but hopefully less ambiguous in meaning...

compress.c
```diff
@@ -28,8 +28,8 @@
    memory, Z_BUF_ERROR if there was not enough room in the output buffer,
    Z_STREAM_ERROR if the level parameter is invalid.
 */
-int Z_EXPORT PREFIX(compress2)(unsigned char *dest, z_uintmax_t *destLen, const unsigned char *source,
-                        z_uintmax_t sourceLen, int level) {
+z_int32_t Z_EXPORT PREFIX(compress2)(unsigned char *dest, z_uintmax_t *destLen, const unsigned char *source,
+                        z_uintmax_t sourceLen, z_int32_t level) {
     PREFIX3(stream) stream;
```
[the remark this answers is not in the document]
mtl1979: `zlib.h` is only used when `ZLIB_COMPAT` is defined, so using old integer types is necessary for backwards compatibility. `zlib.h` is publicly visible header and doesn't include `zbuild.h` which is build-time header.

Dead2:
@ccawley2011 Could you verify whether this works for you?

mtl1979:
I'm gonna try building gcc-newlib to test if there is some code in tests or benchmarks that need adjusting... I'm pretty confident that those are the only places it can fail... Building full gcc takes few hours on even decently modern PC.

mtl1979:
Had to disable GoogleTest and benchmarks as those depend on features not available on newlib... Some tests not disabled also fail due to "unimplemented" features...

Dead2 (approved):
LGTM

PR #1988, opened 2025-10-23
Rename cmake target files to avoid overwrite of PACKAGE_VERSION

bebuch:
Originally there where these 5 CMake config files installed:

```
ZLIB-release.cmake
ZLIB-debug.cmake
ZLIB.cmake
zlib-config-version.cmake
zlib-config.cmake
```

The file `zlib-config-version.cmake` is executed by CMake in a private scope, so the changes it make to variables like `PACKAGE_VERSION` don't leak into the scope where `find_package` is called.

The problem was, that the CMake generated `ZLIB.cmake` includes all files that match the pattern `ZLIB-*.cmake`. That's fine on case sensitive filesystems, but on Windows this triggers the execution of `zlib-config-version.cmake`.

So any call of `find_package(ZLIB)` overwrites `PACKAGE_VERSION`. In case of LLVM I get lld and clang in version 1.3.1 instead of 24.1.4. The worst thing is, that LLVM compiles fine. Only after using libclang with Qt6 I got the very strange error, that clang was found in version 1.3.1.

However, renaming the target files with the common `-targets` postfix solves this problem.

```
ZLIB-targets-release.cmake
ZLIB-targets-debug.cmake
ZLIB-targets.cmake
```

I think that is completly backwards compatible.


mtl1979:
We are officially supporting building zlib-ng for Windows and thus also support case-insensitive file systems... This seems quite clean workaround to me, even though on updates, it does require manually deleting the old target files to work as intended.

CMakeLists.txt
```diff
@@ -1516,6 +1516,7 @@ endif()
 if(NOT SKIP_INSTALL_FILES AND NOT SKIP_INSTALL_ALL)
     install(FILES ${ZLIB_PC} DESTINATION "${PKGCONFIG_INSTALL_DIR}")
     install(EXPORT ${EXPORT_NAME}
+        FILE "${EXPORT_NAME}-targets.cmake"
```
mtl1979: Change directly in `EXPORT` parameter...

```
    install(EXPORT ${EXPORT_NAME}-targets
```

mtl1979 (changes requested):
According to CMake documentation default of `FILE` parameter depends on `EXPORT` parameter of `install()` command. Documentation suggests customizing `EXPORT` parameter directly.

bebuch:
Ready for next review! :-)

Sorry for noise, I missunderstood your suggestion in the first shot and didn't test offline carfully, Now it should be fine ;-)

mtl1979 (approved):
LGTM.

Dead2 (approved):

Issue #1985, opened 2025-10-14
NVC compiler support


PERFACCT-JS:
Dear *zlib-ng* community,

I'm building *zlib-ng* version *2.0.6* with the [nvc 23.9-0 compiler](https://docs.nvidia.com/hpc-sdk/archive/23.9/index.html) on a 64-bit Linux system and a *Meteor Lake* CPU. This CPU supports the `avx_vnni` instruction set. The build is performed as part of creating a Docker image. Containers based on that image run fine on the system on which the image was created. However, when I pull the image on another system with an older CPU of *Cascade Lake-X* architecture (this CPU supports `avx512_vnni` but not `avx_vnni`) and try to run a container there, it fails with `Illegal instruction (core dumped)`.

The backtrace looks like this:
```
(gdb) bt
#0  adler32_avx2 () at /.../zlib-ng/arch/x86/adler32_avx.c:78
#1  0x00007ffff1cf9636 in adler32_stub () at /.../zlib-ng/functable.c:238
#2  0x00007ffff1cfe4b3 in inflate () at /.../zlib-ng/inflate.c:1054

(gdb) display/i $pc
=> 0x7ffff1d06926 <adler32_avx2()+742>:	{vex} vpdpwssd %ymm2,%ymm7,%ymm5
```

As can be seen from the debugger, the application crashes because it tries to execute instruction `vpdpwssd` from the `avx_vnni` instruction set on a system without support for it.

An excerpt from running `objdump -d -S .../arch/x86/adler32_avx.c.o` looks as follows:
```
 2cc:	0f 1f 40 00          	nopl   0x0(%rax)
           __m256i vbuf = _mm256_loadu_si256((__m256i*)buf);
 2d0:	c5 fe 6f 36          	vmovdqu (%rsi),%ymm6
           buf += 32;
 2d4:	48 83 c6 20          	add    $0x20,%rsi
 2d8:	c4 e2 4d 04 f9       	vpmaddubsw %ymm1,%ymm6,%ymm7
 2dd:	c4 e2 4d 04 f3       	vpmaddubsw %ymm3,%ymm6,%ymm6
 2e2:	c5 55 f2 c0          	vpslld %xmm0,%ymm5,%ymm8
           vs1 = _mm256_add_epi32(vsum1, vs1);
 2e6:	c4 e2 45 52 ea       	{vex} vpdpwssd %ymm2,%ymm7,%ymm5
[...]

PR #1993, opened 2025-11-02
Initial support for nVidia toolchain

mtl1979:
* Supports native and non-native builds for x86_64 using CMake

Based on comments in #1985 ...


Dead2 (approved):
LGTM

PR #1999, opened 2025-11-11
several RISC-V hwcap fixes

Icenowy:
This PR drops the code that checks uname before accessing AT_HWCAP (because this auxval always exist in RISC-V Linux), and drops the bogus HWCAP bit for Zbc (because it's never part of upstream Linux UAPI).

cc @ziyao233

### DIFF
--- a/arch/riscv/riscv_features.c
+++ b/arch/riscv/riscv_features.c
@@ -13,38 +13,6 @@
 #define ISA_V_HWCAP (1 << ('v' - 'a'))
 #define ISA_ZBC_HWCAP (1 << 29)
 
-int Z_INTERNAL is_kernel_version_greater_or_equal_to_6_5(void) {
-    struct utsname buffer;
-    if (uname(&buffer) == -1) {
-        // uname failed
-        return 0;
-    }
-
-    int major, minor;
-    if (sscanf(buffer.release, "%d.%d", &major, &minor) != 2) {
-        // Something bad with uname()
-        return 0;
-    }
-
-    if (major > 6 || (major == 6 && minor >= 5))
-        return 1;
-    return 0;
-}
-
-void Z_INTERNAL riscv_check_features_compile_time(struct riscv_cpu_features *features) {
-#if defined(__riscv_v) && defined(__linux__)
-    features->has_rvv = 1;
-#else
-    features->has_rvv = 0;
-#endif
-
-#if defined(__riscv_zbc) && defined(__linux__)
-    features->has_zbc = 1;
-#else
-    features->has_zbc = 0;
-#endif
-}
-
 void Z_INTERNAL riscv_check_features_runtime(struct riscv_cpu_features *features) {
 #if defined(__linux__) && defined(HAVE_SYS_AUXV_H)
     unsigned long hw_cap = getauxval(AT_HWCAP);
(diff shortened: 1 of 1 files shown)


nmoinvaz:
@copilot what was the original reason for the inclusion of this code?

mtl1979:
Calling something downstream "bogus" is not going to help getting this PR approved... Only way this PR is going to get approved is to combine it with adding support for hwprobe.

Dead2:
> Only way this PR is going to get approved is to combine it with adding support for hwprobe.

That would have to be after the first 2.3 stable release.
As I said earlier, I want a PR that only fixes the critical problems now before 2.3, proper hwprobe implementation will have to come later.

This PR disables ZBC completely, and while that is certainly not ideal, I am fine with that for now if that fixes a real problem.

ziyao233:
I've discussed the change privately earlier with Icenowy, and would prefer this PR over #1996 since it keeps the detection of RVV version (the asm volatile block) and thus won't break problematic downstream kernels that set HWCAP_ISA_V even with RVV v0.7.1.

We could drop the extra detection later when hwprobe support is introduced: these problematic kernels don't have hwprobe support backported, so hwprobe will fail and we could simply fallback to GC. But serving as a quick fix, keeping most of the logic and only fixing the real problem sounds better to me.

arch/riscv/riscv_features.c
```diff
@@ -11,38 +11,13 @@
...
-        // Something bad with uname()
-        return 0;
-    }
-
-    if (major > 6 || (major == 6 && minor >= 5))
-        return 1;
-    return 0;
-}
 
 void Z_INTERNAL riscv_check_features_compile_time(struct riscv_cpu_features *features) {
```
ziyao233: Should we remove `riscv_check_features_compile_time()` entirely? The only reference to the symbol is in riscv_check_features_runtime(), and it's removed in the PR.

ziyao233 (approved):

mtl1979:
> This PR disables ZBC completely, and while that is certainly not ideal, I am fine with that for now if that fixes a real problem.

Disabling ZBC detection would discourage downstream projects from switching to zlib-ng until we implement hwprobe, which is more than just not ideal. It makes the project look unprofessional. This PR can be split in two by removing any controversial changes and delaying adding support for hwprobe to later zlib-ng release.

We can list in 2.3.0 release notes that minimum supported kernel version for optimised variants is 6.5 as that is already established as safe requirement.

arch/riscv/riscv_features.c
mtl1979: Unnecessary functions that are no longer needed and likely are not needed in future should obviously be removed.

mtl1979:
> I've discussed the change privately earlier with Icenowy, and would prefer this PR over #1996 since it keeps the detection of RVV version (the asm volatile block) and thus won't break problematic downstream kernels that set HWCAP_ISA_V even with RVV v0.7.1.
> 
> We could drop the extra detection later when hwprobe support is introduced: these problematic kernels don't have hwprobe support backported, so hwprobe will fail and we could simply fallback to GC. But serving as a quick fix, keeping most of the logic and only fixing the real problem sounds better to me.

Even after hwprobe syscall support is added in zlib-ng, there will be "short" gap before we change the minimum kernel version requirement... It might be possible when 2.4.0 is eventually released, but not before... This gives downstream projects enough time to switch to kernel that implements hwprobe syscall.

Icenowy:
I am okay to strip the Zbc commit out, although technically this is incorrect and may lead to problems in the future (but at least not now).

Icenowy:
Dropped the Zbc patch and fixed the problem that riscv_check_features_compile_time() is forgot to be removed .

Dead2 (approved):
LGTM

PR #2003, opened 2025-11-11
Reorganize Chorba activation and refactor Chorba dispatch

Dead2:
First commit:
- Disabling Chorba using CMake parameter `WITH_CRC32_CHORBA=OFF` will now only disable the crc32_chorba C fallback.

- SSE2, SSE41 and pclmul variants will still be able to use their own Chorba-algorithm based code, but their fallback to
the generic crc32_chorba C code in SSE2 and SSE41 will be disabled, reducing their performance on really big
input buffers. But this allows the SSE2 and SSE41 functions to still be used in the first place, and they provide a huge win over crc32_braid for buffers over 72 bytes.

- Remove the crc32_c function (and its file crc32_c.c), instead use the normal functable routing to select
between crc32_braid and crc32_chorba.

Second commit:
- Unify crc32_chorba, chorba_sse2 and chorba_sse41 dispatch functions.
- Fixed alignment diff calculation in crc32_chorba.
- Fixed length check to happen early, avoiding extra branches for too short lengths,
this also allows removing one function call to crc32_braid_internal to handle those.
Gbench shows ~0.15-0.25ns saved per call for lengths shorter than CHORBA_SMALL_THRESHOLD.
- Avoid calculating aligned len if buffer is already aligned

PR #2004, opened 2025-11-12
Reorganize Chorba activation

Dead2:
Split out separately from #2003. Easier to review, and with disabled Chorba-sse variants on old MSVC versions.

* WITHOUT_CHORBA will now only disable the crc32_chorba C fallback.

* SSE2, SSE41 and pclmul variants will still be able to use their Chorba-algorithm based code, but their fallback to the generic crc32_chorba C code in SSE2 and SSE41 will be disabled, reducing their performance on really big input buffers (not used during deflate/inflate, only when calling crc32 directly).

* Remove the crc32_c function (and its file crc32_c.c), instead use the normal functable routing to select between crc32_braid and crc32_chorba.

* Disable sse2 and sse4.1 variants of Chorba-crc32 on MSVC older than 2022 due to code generation bug in 2019 causing segfaults.

* Compile either crc32_chorba_small_nondestructive or crc32_chorba_small_nondestructive_32bit, not both. Shaves ~1.1KB off the library size.

* Don't compile crc32_chorba_32768_nondestructive on 32bit arch.


Dead2:
Develop x86-64:
 172705    1336       8  174049   2a7e1 libz-ng.so.2

PR x86-64:
 171593    1336       8  172937   2a389 libz-ng.so.2

Develop rpi5:
 119167    1560       8  120735   1d79f libz-ng.so.2

PR rpi5:
 118303    1560       8  119871   1d43f libz-ng.so.2
 
 Default compile (with chorba enabled) results in library size reductions:
x86_64: 1112 bytes smaller
aarch64: 864 bytes smaller

Performance for default builds is unchanged, as expected since this does not change anything for those builds.
Performance for Chorba-disabled builds are faster on x86(-64) systems that don't have pclmul instructions but have SSE2 or SSE4.1, since those are no longer disabled along with chorba_c. (See benchmarks for the "no-chorba first commit" here: [https://github.com/zlib-ng/zlib-ng/pull/2003](https://github.com/zlib-ng/zlib-ng/pull/2003#issuecomment-3518886860))
Also [v]plcmul fold copy functions are improved, not the plain crc32 ones, used during inflate.

test/benchmarks/benchmark_crc32.cc
```diff
@@ -56,27 +56,21 @@ class crc32: public benchmark::Fixture {
...
-#       endif
-#   endif
+BENCHMARK_CRC32(chorba_c, crc32_chorba, 1);
+#endif
+#if defined(X86_SSE2) && !defined(WITHOUT_CHORBA_SSE)
+BENCHMARK_CRC32(chorba_sse2, crc32_chorba_sse2, test_cpu_features.x86.has_sse2);
+#endif
+#if defined(X86_SSE41) && !defined(WITHOUT_CHORBA_SSE)
+BENCHMARK_CRC32(chorba_sse41, crc32_chorba_sse41, test_cpu_features.x86.has_sse41);
 #endif
```
mtl1979: Nitpick: Outer test should be `!defined(WITHOUT_CHORBA_SSE)`, so the test isn't unnecessarily duplicated.

test/test_crc32.cc
```diff
@@ -317,10 +314,10 @@ TEST_CRC32(pclmulqdq, crc32_pclmulqdq, test_cpu_features.x86.has_pclmulqdq)
...
 TEST_CRC32(vpclmulqdq, crc32_vpclmulqdq, (test_cpu_features.x86.has_pclmulqdq && test_cpu_features.x86.has_avx512_common && test_cpu_features.x86.has_vpclmulqdq))
 #endif
-#if !defined(WITHOUT_CHORBA) && defined(X86_SSE2) && !defined(NO_CHORBA_SSE)
+#if defined(X86_SSE2) && !defined(WITHOUT_CHORBA_SSE)
 TEST_CRC32(chorba_sse2, crc32_chorba_sse2, test_cpu_features.x86.has_sse2)
 #endif
-#if !defined(WITHOUT_CHORBA) && defined(X86_SSE41) && !defined(NO_CHORBA_SSE)
+#if defined(X86_SSE41) && !defined(WITHOUT_CHORBA_SSE)
 TEST_CRC32(chorba_sse41, crc32_chorba_sse41, test_cpu_features.x86.has_sse41)
 #endif
```
mtl1979: Nitpick: Outer test should be `!defined(WITHOUT_CHORBA_SSE)`, so the test isn't unnecessarily duplicated.

mtl1979 (changes requested):
Just some nitpick comments about logically backwards preprocessor tests.

test/benchmarks/benchmark_crc32.cc
Dead2: I did have it the other way initially, but I really didn't like the readability of it. Having it like this makes it easier to quickly scan down the lines and find the correct test since the names are all aligned the same. Figured the duplicated test doesn't really hurt a lot, don't you agree?

mtl1979: For some people, especially "autistic" people, indentation actually helps with reading the code, even more when dependent things are grouped together. Same goes for adding horizontal lines or extra line feeds when large span of text ends.

Dead2: Well, if you really don't like it, I can certainly put it back. Personally I just think it is cleaner and easier to review.
I am autistic myself, as you probably know.

mtl1979: So am I... Autistic people typically don't like big changes, but are good to spot small errors because of strong focus on logical thinking. If something does not look right, they usually know what and why.

As said before, we should also keep the code and comments short and clear enough for AI bots too.

PR #2005, opened 2025-11-13
Add tests for crc32_fold_copy functions

Dead2:
What I wanted was to add benchmarks for crc32_fold_copy, what I ended up implementing was tests.

I reused the tests from the crc32 tests, to make it easier to add new ones.

This passes testing on x86_64 fine.
Added tests for ARMv8 and Loongarch as well, even though those actually just do the crc followed by a memcpy instead of doing the copy during the actual checksumming.

For some reason the [v]pclmul claim 16 bytes is minimum, but in my tests they work with no minimum as well. Not sure whether I was just lucky or that 16byte minimum is not actually a minimum. Perhaps it does read beyond the end of the buffer but as long as that is allowed, it is safe? IDK, so I kept it at 16 for now.

### DIFF
--- a/test/crc32_test_strings_p.h
+++ b/test/crc32_test_strings_p.h
@@ -1,0 +1,183 @@
+#ifndef CRC32_TEST_STRINGS_P_H
+#define CRC32_TEST_STRINGS_P_H
+
+ALIGNED_(16) static uint8_t crc32_fullwin_buf[32768];
+
+static uint8_t* crc32_setup_buf() {
+    for (int i = 0; i < 32768; ++i) {
+        uint8_t ic = (uint8_t)(i % 256);
+        crc32_fullwin_buf[i] = ic;
+    }
+
+    return crc32_fullwin_buf;
+}
+
+static uint8_t *buf32k = crc32_setup_buf();
+
+typedef struct {
+    unsigned long crc;
+    const uint8_t *buf;
+    size_t len;
+    unsigned long expect;
+} crc32_test;
+
+/* (init-)crc, buf, len, expect(-crc) */
+static const crc32_test crc32_tests[] = {
+  {0x0, (const uint8_t *)0x0, 0, 0x0},
+  {0xffffffff, (const uint8_t *)0x0, 0, 0x0},
+  {0x0, (const uint8_t *)0x0, 255, 0x0}, /*  BZ 174799.  */
+  {0x0, (const uint8_t *)0x0, 256, 0x0},
+  {0x0, (const uint8_t *)0x0, 257, 0x0},
+  {0x0, (const uint8_t *)0x0, 32767, 0x0},
+  {0x0, (const uint8_t *)0x0, 32768, 0x0},
+  {0x0, (const uint8_t *)0x0, 32769, 0x0},
+  {0x0, (const uint8_t *)"", 0, 0x0},
+  {0xffffffff, (const uint8_t *)"", 0, 0xffffffff},
+  {0x0, (const uint8_t *)"abacus", 6, 0xc3d7115b},
+  {0x0, (const uint8_t *)"backlog", 7, 0x269205},
+  {0x0, (const uint8_t *)"campfire", 8, 0x22a515f8},
+  {0x0, (const uint8_t *)"delta", 5, 0x9643fed9},
+  {0x0, (const uint8_t *)"executable", 10, 0xd68eda01},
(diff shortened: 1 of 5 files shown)


test/crc32_test_strings_p.h
```diff
@@ -0,0 +1,178 @@
+ALIGNED_(16) static uint8_t crc32_fullwin_buf[32768];
+
+static uint8_t* crc32_setup_buf() {
+    for (int i = 0; i < 32768; ++i) {
+        unsigned char ic = (unsigned char)(i % 256);
```
mtl1979: Should use uint8_t here too.

test/test_crc32_fold_copy.cc
```diff
@@ -0,0 +1,82 @@
...
+    TEST_CRC32_FOLD(armv8, 0, 0, crc32_fold_reset_c, crc32_fold_copy_armv8, crc32_fold_final_c, test_cpu_features.arm.has_crc32)
+#  endif
+#  ifdef X86_PCLMULQDQ_CRC
+    // Is 16 bytes len the minimum for pclmul functions?
+    TEST_CRC32_FOLD(pclmulqdq, 16, 1, crc32_fold_pclmulqdq_reset, crc32_fold_pclmulqdq_copy, crc32_fold_pclmulqdq_final, test_cpu_features.x86.has_pclmulqdq)
+#  endif
+#  ifdef X86_VPCLMULQDQ_CRC
+    TEST_CRC32_FOLD(vpclmulqdq, 16, 1, crc32_fold_pclmulqdq_reset, crc32_fold_vpclmulqdq_copy, crc32_fold_pclmulqdq_final, (test_cpu_features.x86.has_pclmulqdq && test_cpu_features.x86.has_avx512_common && test_cpu_features.x86.has_vpclmulqdq))
+#  endif
+#  if defined(LOONGARCH_CRC)
```
mtl1979: Should use ifdef here too for consistency.

test/test_crc32_fold_copy.cc
```diff
@@ -0,0 +1,82 @@
...
+#  include "crc32_test_strings_p.h"
+}
+
+class crc32_fc_variant : public ::testing::TestWithParam<crc32_test> {
+public:
+    /* Ensure that crc32 fold copy functions returns the correct crc and copies the data */
+    static void crc32_fold_test(size_t minlen, int onlyzero, crc32_fold_reset_func fold_reset, crc32_fold_copy_func fold_copy, crc32_fold_final_func fold_final, crc32_test params) {
+        ALIGNED_(16) crc32_fold crc_st;
+        uint32_t crc;
+        uint8_t dstbuf[32768];
```
mtl1979: This large allocation on stack can cause issues if not the first variable that is allocated in current scope.

mtl1979 (changes requested):

test/test_crc32_fold_copy.cc
Dead2: Fixed.

test/crc32_test_strings_p.h
Dead2: I only moved this code, but fixed now.

test/test_crc32_fold_copy.cc
```diff
@@ -0,0 +1,83 @@
...
+            fold_copy(&crc_st, dstbuf, params.buf, params.len);
+
+            crc = fold_final(&crc_st);
+
+            EXPECT_EQ(crc, params.expect);
+            EXPECT_EQ(0, memcmp(params.buf, dstbuf, params.len));
+        }
+
+        free(dstbuf);
+    }
```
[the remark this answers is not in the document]
Dead2: Added

test/crc32_test_strings_p.h
```diff
@@ -0,0 +1,183 @@
...
+}
+
+static uint8_t *buf32k = crc32_setup_buf();
+
+typedef struct {
+    unsigned long crc;
+    const uint8_t *buf;
+    size_t len;
+    unsigned long expect;
+} crc32_test;
```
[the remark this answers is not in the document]
mtl1979: Changing type of `crc` will break the alignment of subsequent struct fields on 64-bit targets, `expect` should be moved up to fill the gap as the next struct element is a pointer type.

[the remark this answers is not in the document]
Dead2: I am not changing the data structure for the crc32 test strings in this PR, that is just not a part of this PR. Those were only moved from a test to a common include file to facilitate re-use by other tests.

PR #2006, opened 2025-11-13
Rewrite chorba dispatch functions

Dead2:
- Unify crc32_chorba, chorba_sse2 and chorba_sse41 dispatch functions.
- Fix alignment calculation in crc32_chorba.
- Fix length check to happen early, avoiding several extra branches for too short lengths, this also allows removing one function call to crc32_braid_internal to handle fallthrough. Gbench shows ~0.15-0.25ns saved per call for lengths shorter than CHORBA_SMALL_THRESHOLD.
- Avoid calculating aligned len if buffer is already aligned


nmoinvaz:
LGTM

PR #2009, opened 2025-11-16
Disable benchmark for slide_hash_c with Visual C++ too.

mtl1979:
If `WITH_ALL_FALLBACKS` is unset, we need to disable benchmark for slide_hash_c when using Visual C++ to target AMD64.


Dead2 (approved):
LGTM

PR #2017, opened 2025-11-18
Change OS_CODE definition for Windows platform

rzikm:
[gzip RFC](https://datatracker.ietf.org/doc/html/rfc1952#section-2.2) lists the previous value (10) as TOPS-20, this change replaces it with value value 11 (NTFS filesystem) which is more suited to represent Windows OS.

### DIFF
--- a/zutil.h
+++ b/zutil.h
@@ -112,7 +112,7 @@ extern z_const char * const PREFIX(z_errmsg)[10]; /* indexed by 2-zlib_error */
 #endif
 
 #if defined(_WIN32) && !defined(__CYGWIN__)
-#  define OS_CODE  10
+#  define OS_CODE  11
 #endif
 
 #ifdef __APPLE__


Dead2:
Code 11 is already used by system 370 above.
Also, we follow zlib, not gzip, so I guess that is where the discrepancy comes from.
If you want this to change, I think you'll probably have to convince zlib to change its codes first.

rzikm:
Right, but when zlib-ng is used to produce zgipped content, it ends up using that value, e.g. for empty content I get

    1F 8B 08 00 00 00 00 00 00 0A 00 00 00 FF FF 03 00 00 00 00 00 00 00 00 00

so I am trying to understand if it's the implementation or RFC that is wrong.

Dead2:
@rzikm Unfortunately we cannot answer that question, as that depends on the decisions done by the developers of zlib many, many years ago.
The right place to ask that question would be over at https://github.com/madler/zlib

rzikm:
Ok, I will follow up with upstream repo. Thanks.

PR #2018, opened 2025-11-20
Use elf_aux_info() on FreeBSD and OpenBSD ARM / AArch64

brad0:
Use elf_aux_info() as the prefered API for modern FreeBSD and OpenBSD
ARM and AArch64. This adds 32-bit ARM support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -802,6 +802,54 @@ if(WITH_OPTIM)
                 endif()
             endif()
         endif()
+        if(${CMAKE_SYSTEM_NAME} STREQUAL "FreeBSD" OR
+           ${CMAKE_SYSTEM_NAME} STREQUAL "OpenBSD")
+            if("${ARCH}" MATCHES "aarch64")
+                check_c_source_compiles(
+                    "#include <sys/auxv.h>
+                    int main() {
+                        unsigned long hwcap;
+                        elf_aux_info(AT_HWCAP, &hwcap, sizeof(hwcap));
+                        return (hwcap & HWCAP_CRC32);
+                    }"
+                    ARM_AUXV_HAS_CRC32
+                )
+                if(ARM_AUXV_HAS_CRC32)
+                    add_definitions(-DARM_AUXV_HAS_CRC32)
+                else()
+                    message(STATUS "HWCAP_CRC32 not present in sys/auxv.h; cannot detect support at runtime.")
+                endif()
+            else()
+                check_c_source_compiles(
+                    "#include <sys/auxv.h>
+                    int main() {
+                        unsigned long hwcap2;
+                        elf_aux_info(AT_HWCAP2, &hwcap2, sizeof(hwcap2));
+                        return (hwcap2 & HWCAP2_CRC32);
+                    }"
+                    ARM_AUXV_HAS_CRC32
+                )
+                if(ARM_AUXV_HAS_CRC32)
+                    add_definitions(-DARM_AUXV_HAS_CRC32)
+                else()
+                    message(STATUS "HWCAP2_CRC32 not present in sys/auxv.h; cannot detect support at runtime.")
+                endif()
+                check_c_source_compiles(
+                    "#include <sys/auxv.h>
+                    int main() {
+                      unsigned long hwcap;
+                      elf_aux_info(AT_HWCAP, &hwcap, sizeof(hwcap));
+                      return (hwcap & HWCAP_NEON);
+                    }"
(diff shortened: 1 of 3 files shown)


arch/arm/arm_features.c
```diff
@@ -26,12 +26,24 @@
 #endif
 
 static int arm_has_crc32(void) {
-#if defined(__linux__) && defined(ARM_AUXV_HAS_CRC32)
+#if defined(ARM_AUXV_HAS_CRC32)
+#if defined(__FreeBSD__) || defined(__OpenBSD__)
```
mtl1979: Bad indentation

arch/arm/arm_features.c
```diff
@@ -62,12 +74,18 @@ static int arm_has_crc32(void) {
 /* AArch64 has neon. */
 #if !defined(__aarch64__) && !defined(_M_ARM64) && !defined(_M_ARM64EC)
 static inline int arm_has_neon(void) {
-#if defined(__linux__) && defined(ARM_AUXV_HAS_NEON)
+#if defined(ARM_AUXV_HAS_NEON)
+#if defined(__FreeBSD__) || defined(__OpenBSD__)
```
mtl1979: Bad indentation

CMakeLists.txt
```diff
@@ -738,70 +738,87 @@ if(WITH_OPTIM)
...
+                    #endif
+                }"
+                ARM_AUXV_HAS_CRC32
+            )
+            if(ARM_AUXV_HAS_CRC32)
+                add_definitions(-DARM_AUXV_HAS_CRC32)
             else()
                 check_c_source_compiles(
                     "#include <sys/auxv.h>
+                    #include <asm/hwcap.h>
```
mtl1979: This will break on systems where `asm/hwcap.h` is not required and does not exist.

configure
```diff
@@ -568,14 +568,6 @@ case "$uname" in
...
 esac
 
-# Simplify some later conditionals
-case "$uname" in
-Linux* | linux*)
-  LINUX=1 ;;
-*)
-  LINUX=0 ;;
-esac
-
```
mtl1979, line 578: Removing this block has nothing to do with FreeBSD/OpenBSD support.

configure
```diff
@@ -1970,72 +1962,87 @@ EOF
                 echo "Checking for arm_acle.h... No." | tee -a configure.log
             fi
 
-
-            if test $LINUX -eq 1; then
```
mtl1979, line 1974: Removing this test will introduce a lot of unnecessary white space changes with very little common code between Linux and FreeBSD/OpenBSD... It would be cleaner for reviewing purposes to separate them as own blocks.

CMakeLists.txt
```diff
@@ -738,70 +738,87 @@ if(WITH_OPTIM)
...
+                        elf_aux_info(AT_HWCAP2, &hwcap2, sizeof(hwcap2));
+                        return (hwcap2 & HWCAP2_CRC32);
+                    #else
+                        return (getauxval(AT_HWCAP2) & HWCAP2_CRC32);
+                    #endif
+                }"
+                ARM_AUXV_HAS_CRC32
+            )
+            if(ARM_AUXV_HAS_CRC32)
+                add_definitions(-DARM_AUXV_HAS_CRC32)
```
mtl1979: A lot of unnecessary white space changes... Makes reviewing the code really hard.

CMakeLists.txt
```diff
@@ -738,70 +738,87 @@ if(WITH_OPTIM)
     add_definitions(-DWITH_OPTIM)
     if(BASEARCH_ARM_FOUND)
         add_definitions(-DARM_FEATURES)
-        if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
```
mtl1979, line 741: Just add FreeBSD and OpenBSD here instead of removing whole test and introducing a lot of white space changes.

mtl1979 (changes requested):
A lot of unnecessary white space changes...

brad0:
@coderabbitai full review

mtl1979:
Looking a lot better now, but I don't have AArch64 version of FreeBSD and qemu 10 can't run AArch64 binaries on AMD64...

I'm not sure if we should have single variable in `configure` for the code blocks that are currently common with FreeBSD and OpenBSD... If ever we want to add support for NetBSD and it requires identical tests, then it might be cleaner to use something like `BSD` for all BSD-like systems using ELF binaries and having `elf_aux_info()`.

Dead2 (approved):
LGTM

PR #2026, opened 2025-11-29
Rename adler32_fold_copy to adler32_copy

nmoinvaz:
There are no folding techniques in adler32 implementations. It is simply hashing while copying.


arch/x86/adler32_avx2.c
```diff
@@ -170,11 +170,11 @@ static inline uint32_t adler32_fold_copy_impl(uint32_t adler, uint8_t *dst, cons
...
 
 Z_INTERNAL uint32_t adler32_avx2(uint32_t adler, const uint8_t *src, size_t len) {
-    return adler32_fold_copy_impl(adler, NULL, src, len, 0);
+    return adler32_copy_impl(adler, NULL, src, len, 0);
 }
 
-Z_INTERNAL uint32_t adler32_fold_copy_avx2(uint32_t adler, uint8_t *dst, const uint8_t *src, size_t len) {
-    return adler32_fold_copy_impl(adler, dst, src, len, 1);
+Z_INTERNAL uint32_t adler32_copy_avx2(uint32_t adler, uint8_t *dst, const uint8_t *src, size_t len) {
+    return adler32_copy_impl(adler, dst, src, len, dst != NULL);
```
Dead2: **question:** Wouldn't this added check also remove the ability for the compiler to optimize away the copy checks when inlining the function? I haven't benchmarked it, but I expect it would be slower.

Dead2: (same for all the other places doing the same of course)

nmoinvaz: It never optimizes it away in this case because it was always 1 before.

Dead2: I mean, when it is always 1; it doesn't need to add branching to check whether it is 1, it can just assume it is 1. Now it would need to add the branches.

phprus: I think there should be an `Assert()` here, since this functions is not called with `dst == NULL`.

nmoinvaz: I have removed all the `dst != NULL` I added and the extra one I found. I don't think we should add an `Assert()` I feel it is excessive because then we have to add one for `src != NULL`, etc and all other conditions. Since it is not called with `dst == NULL` it shouldn't matter.

Dead2 (approved):
LGTM

PR #2036, opened 2025-12-03
Wrap support flag for benchmark and test macros.

nmoinvaz:


Dead2 (approved):
Ah, so THAT is what was happening to my testing previously. It makes sense now 👍
LGTM

PR #2038, opened 2025-12-04
configure: Check cc in addition to gcc for GCC compatible compiler

brad0:
If the first try checking for a GCC compatible compiler did not
find anything with gcc(1) then fallback to cc(1).

FreeBSD / OpenBSD no longer use GCC as the system compiler, so gcc(1)
no longer exists. Running ./configure on its own will not find what
it thinks should be GCC compatible compiler and will build a shared
library without -fPIC and fail, but Clang does exist as cc(1).

Fixes building on FreeBSD.

### DIFF
--- a/configure
+++ b/configure
@@ -283,6 +283,14 @@ case $($cc -v 2>&1) in
   *clang*) gcc=1 ;;
 esac
 
+if test "$gcc" -eq 0; then
+  cc=${CC-${CROSS_PREFIX}cc}
+  case $($cc -v 2>&1) in
+    *gcc*) gcc=1 ;;
+    *clang*) gcc=1 ;;
+  esac
+fi
+
 if test $build32 -eq 1; then
   CFLAGS="${CFLAGS} -m32"
   SFLAGS="${SFLAGS} -m32"


Dead2:
While I see the benefit of auto-detecting cc as well, I do not see any reasoning for why you also change the default.
Doing that will cause a potentially big change in behavior for everyone using configure, not only *bsd. If we were to do that, it'd probably have to wait for another major release, so probably a year+.
I suggest using cc as the fall-back instead, that way it only affects those that don't currently work anyway.

brad0:
> While I see the benefit of auto-detecting cc as well, I do not see any reasoning for why you also change the default.

I mean it ultimately is two separate issues.

> Doing that will cause a potentially big change in behavior for everyone using configure, not only *bsd. If we were to do that, it'd probably have to wait for another major release, so probably a year+. I suggest using cc as the fall-back instead, that way it only affects those that don't currently work anyway.

Ok, I'll take that part out.

configure
```diff
@@ -283,6 +283,14 @@ case $($cc -v 2>&1) in
   *clang*) gcc=1 ;;
 esac
 
+if test "$gcc" -eq 0; then
+  cc=${CC-${CROSS_PREFIX}cc}
+  case $($cc -v 2>&1) in
```
mtl1979, line 288: Not all LLVM-based compilers support `-v`, so this gives false negatives.

mtl1979 (commented):
Outdated way of testing compiler brand... This will cause errors later when "-fno-lto" test is skipped, even though it's not supported, and stays enabled.

configure
brad0: Well it's testing for Clang, not LLVM-based compilers. What exception to the overwhelming majority is this?

mtl1979: Considering a lot of operating systems are dropping or have already dropped support for gcc, this means more and more users are going to use custom LLVM-based compilers until the patches are upstreamed.

The PR title is ambiguous even though the description does explain more about restricted nature of the scope of the PR.

My point was/is that fixing this properly needs more than just checking output of `$CC -v`. Command-line option `-v` commonly means `verbose`, which alone does not make sense and it's pure luck that it has worked correctly so far. For this and the test above to work as expected, we would need to use `$CC --version` instead.

We also need to set default value for `noltoflag` based on value of variable `gcc`, so later than it is now.

When I tested this PR in Ubuntu 24.04.1 without gcc and clang, output of `$CC -v` was warning `No files to process`. Obviously compiling and running tests still worked after removing test for variable `gcc`.

configure
```diff
@@ -283,6 +283,14 @@ case $($cc -v 2>&1) in
...
+    *gcc*) gcc=1 ;;
+    *clang*) gcc=1 ;;
+  esac
```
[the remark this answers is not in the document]
mtl1979: Rechecking compiler filename is pointless as it is unlikely to match anything other than "cc" or gives false positive due to installation path possibly containing "gcc" or "clang" if installed under user's home directory and user name contains such text. We need to make sure no error message is printed on STDERR as that would mean the command can't be found.

configure
```diff
@@ -283,6 +283,14 @@ case $($cc -v 2>&1) in
   *clang*) gcc=1 ;;
 esac
 
+if test "$gcc" -eq 0; then
+  cc=${CC-${CROSS_PREFIX}cc}
```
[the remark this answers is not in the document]
mtl1979: If removing CC here and second test fails, the value of "cc" should be reverted to whatever first test used as the command might still be valid but didn't match any listed patterns, for example when using "nvc" which is partially compatible with gcc but can only target current processor due to missing xgetbv intrinsic.

We essentially need outer test if `CC` is non-empty and in that case neither "${CROSS_PREFIX}gcc" or "${CROSS_PREFIX}cc" is tried.

configure
Dead2: **question:** @brad0 Would this work for you, changing `-v` to `--version`?

brad0: > **question:** @brad0 Would this work for you, changing `-v` to `--version`?

Looking at what the script is scraping for from the output it looks like it would be fine.

```
arm64$ clang --version
OpenBSD clang version 19.1.7
Target: aarch64-unknown-openbsd7.8
Thread model: posix
InstalledDir: /usr/bin

arm64$ clang -v
OpenBSD clang version 19.1.7
Target: aarch64-unknown-openbsd7.8
Thread model: posix
InstalledDir: /usr/bin

arm64$ egcc --version
egcc (GCC) 8.4.0
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

arm64$ egcc -v
Using built-in specs.
COLLECT_GCC=egcc
COLLECT_LTO_WRAPPER=/usr/local/libexec/gcc/aarch64-unknown-openbsd/8.4.0/lto-wrapper
Target: aarch64-unknown-openbsd
Configured with: /usr/obj/ports/gcc-8.4.0/gcc-8.4.0/configure --build=aarch64-unknown-openbsd --with-as=/usr/local/bin/gas --verbose --program-transform-name='s,^,e,' --disable-nls --with-system-zlib --disable-libmudflap --disable-libgomp --disable-libssp --disable-tls --with-gnu-ld --with-gnu-as --enable-threads=posix --enable-wchar_t --with-gmp=/usr/local --enable-languages=c,c++,fortran,objc --disable-libstdcxx-pch --enable-default-pie --without-isl --enable-default-ssp --enable-cpp --prefix=/usr/local --sysconfdir=/etc --mandir=/usr/local/man --infodir=/usr/local/info --localstatedir=/var --disable-silent-rules --disable-gtk-doc
Thread model: posix
gcc version 8.4.0 (GCC)
```

If such a change is made the string be scraped for should be changed to an uppercase GCC.

mtl1979: Changing it to uppercase GCC would mean that it would also detect compilers where the name of the compiler front-end doesn't contain `gcc` - it would capture the `GCC` inside parentheses -, although in case of `egcc` the name does contain `gcc` as a substring. 

As far as I know, use of `egcc` and `gcc` is discouraged, and basically unsupported, on architectures where clang is the default (base system) compiler, and can cause non-recoverable errors.

brad0: > Changing it to uppercase GCC would mean that it would also detect compilers where the name of the compiler front-end doesn't contain `gcc` - it would capture the `GCC` inside parentheses -, although in case of `egcc` the name does contain `gcc` as a substring.

Yup, that is the point if switching from -v to --version since the output is different between the two.

mtl1979:
We could simplify the logic by detecting presence of the actual compiler file using `which` command,  that will output empty string in standard output if the command doesn't exist... That way we don't rely on first attempt to return GCC compatible compiler. Then we also only need to parse version string once. This allows us to separate detecting new compilers to another PR. As mentioned earlier, there are already compilers that the current version parsing doesn't detect. If we don't find any compiler, we can also bail out when cross prefix is specified.

Dead2:
Superseded by #2140, reopen if things are still not working right.

PR #2041, opened 2025-12-06
Micro-optimization for in pointer calculation for inflate_fast REFILL

nmoinvaz:
trifectatechfoundation/zlib-rs#320


Dead2 (approved):
LGTM

PR #2042, opened 2025-12-06
Use __builtin_bitreverse16 in bi_reverse if available.

nmoinvaz:


fallback_builtins.h
```diff
@@ -44,4 +44,16 @@ static __forceinline int __builtin_ctzll(unsigned long long value) {
 #endif // Microsoft AMD64/IA64/x86/ARM/ARM64 test
 #endif // _MSC_VER & !clang
 
+#ifndef HAVE_BUILTIN_BITREVERSE16
+/* Bit reversal for 8-bit values using multiplication method */
+#define bitrev8(value) \
+    (uint8_t)((((uint8_t)(value) * 0x80200802ULL) & 0x0884422110ULL) * 0x0101010101ULL >> 32)
+
+/* General purpose bit reversal for 16-bit values */
+Z_FORCEINLINE static uint32_t __builtin_bitreverse16(uint16_t value) {
```
phprus: Replace `uint32_t` with `uint16_t`.

mtl1979:
On architectures that actually have instruction for bit reverse, the builtin function might make things a lot faster depending on how many cycles the instruction takes, but at least on x86/AMD64, where there is no hardware support, this intrinsic depends on support in compiler's optimizer. At least LLVM can optimize bit reverse even on platforms where there is no hardware instruction.

KungFuJesus:
> On architectures that actually have instruction for bit reverse, the builtin function might make things a lot faster depending on how many cycles the instruction takes, but at least on x86/AMD64, where there is no hardware support, this intrinsic depends on support in compiler's optimizer. At least LLVM can optimize bit reverse even on platforms where there is no hardware instruction.

It can be done cheaply with pext and and a multiplication.

nmoinvaz:
@KungFuJesus that sounds like an interesting idea.

fallback_builtins.h
```diff
@@ -44,4 +44,16 @@ static __forceinline int __builtin_ctzll(unsigned long long value) {
...
+/* Bit reversal for 8-bit values using multiplication method */
+#define bitrev8(b) \
+    (uint8_t)((((uint8_t)(b) * 0x80200802ULL) & 0x0884422110ULL) * 0x0101010101ULL >> 32)
+
+/* General purpose bit reversal for 16-bit values */
+Z_FORCEINLINE static uint32_t __builtin_bitreverse16(uint16_t value) {
+    return (bitrev8(code >> 8) | (uint16_t)bitrev8(code) << 8);
+}
+#define HAVE_BUILTIN_BITREVERSE16
+#endif
```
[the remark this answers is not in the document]
nmoinvaz: Fixed.

deflate_p.h
```diff
@@ -127,13 +127,11 @@ Z_FORCEINLINE static void flush_pending_inline(PREFIX3(stream) *strm) {
...
-static inline uint16_t bi_reverse(unsigned code, int len) {
+Z_FORCEINLINE static uint16_t bi_reverse(unsigned code, int len) {
     /* code: the value to invert */
     /* len: its bit length */
     Assert(len >= 1 && len <= 15, "code length must be 1-15");
-#define bitrev8(b) \
-    (uint8_t)((((uint8_t)(b) * 0x80200802ULL) & 0x0884422110ULL) * 0x0101010101ULL >> 32)
-    return (bitrev8(code >> 8) | (uint16_t)bitrev8(code) << 8) >> (16 - len);
+    return __builtin_bitreverse16((uint16_t)code) >> (16 - len);
 }
```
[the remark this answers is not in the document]
nmoinvaz: Fixed.

fallback_builtins.h
nmoinvaz: Fixed.

Dead2:
MSVC failures:
`inftrees.c(234): warning C4242: 'function': conversion from 'unsigned int' to 'uint16_t', possible loss of data`

mtl1979:
> It can be done cheaply with pext and and a multiplication.

A lot of times when I comment on a PR, it's all about open invite to discuss how to do things better than each compiler's own optimizer does it... That's one of the main reasons we use intrinsics instead of hand-writing C code that average Joe can understand without inline comments. 

In compiler explorer looking at just how many lines of machine code gcc trunk and clang trunk generate, it looks like clang can generate 64-bit wide bit reverse for x86_64 target in one less "step" and first step is a lot shorter. For shorter bit widths clang uses more steps but still less instructions. I didn't try to benchmark the different implementations yet as I'm currently using Windows.

mtl1979:
> MSVC failures: `inftrees.c(234): warning C4242: 'function': conversion from 'unsigned int' to 'uint16_t', possible loss of data`

AI did catch that already, but as it doesn't have large enough context window, it thought the missing explicit truncating cast is not treated as an error...

inftrees.c
```diff
@@ -227,15 +230,8 @@ int Z_INTERNAL zng_inflate_table(codetype type, uint16_t *lens, unsigned codes,
...
-        while (huff & incr)
-            incr >>= 1;
-        if (incr != 0) {
-            huff &= incr - 1;
-            huff += incr;
-        } else {
-            huff = 0;
-        }
+        rhuff += (0x8000u >> (len - 1));
+        huff = __builtin_bitreverse16(rhuff);
```
mtl1979, line 234: Missing explicit truncating cast.

mtl1979 (changes requested):
We require zero warnings for MSVC, thus missing cast breaks build.

Andarwinux:
@mtl1979 clang can use GFNI to directly lower `__builtin_bitreverse16`

https://clang.godbolt.org/z/qzcGfEdfh

mtl1979:
> @mtl1979 clang can use GFNI to directly lower __builtin_bitreverse16
>
> https://clang.godbolt.org/z/qzcGfEdfh

Looks like `-march=znver4` is minimum requirement for that optimization.

Andarwinux:
> > @mtl1979 clang can use GFNI to directly lower __builtin_bitreverse16
> > https://clang.godbolt.org/z/qzcGfEdfh
> 
> Looks like `-march=znver4` is minimum requirement for that optimization.

For Intel it's a bit earlier, it's been supported since icelake.

mtl1979:
> For Intel it's a bit earlier, it's been supported since icelake.

At least `-march=icelake-client` did work for me when I tested...

KungFuJesus:
Yeah the client CPUs didn't get gfni. Unfortunately none of the x86 versions lump that in so the people that benefit the most from it are those who are building with march=native.
 The pext trick that the compiler is maybe employing required bmi2, though I think zen3 implemented pdep and pext poorly, so that too may be guarded behind specific optimization targets.

nmoinvaz:
> MSVC failures: `inftrees.c(234): warning C4242: 'function': conversion from 'unsigned int' to 'uint16_t', possible loss of data`

Fixed.

Andarwinux:
> Yeah the client CPUs didn't get gfni.

What? icelake-client did get GFNI.
https://clang.godbolt.org/z/KjPK8W8W5

> Unfortunately none of the x86 versions lump that in so the people that benefit the most from it are those who are building with march=native.

I'm not sure if it's feasible, but if zlib-ng could introduce a concept like [highway](https://github.com/google/highway?tab=readme-ov-file#targets)'s “target” instead of single features, it could benefit more users.

KungFuJesus:
Ahh, right you are.  I might be confusing this because I recall at the time they were introduced they were given to Xeon SP CPUs well in advance of the consumer chips.  I guess by the time Icelake rolled around the extensions weren't so different between client and server.  

The readme markdown doesn't give a very extension description for how the "targets" are implemented but it doesn't sound too unlike how are existing function table dispatches now (looks like there's a compile time static target you can set or setup dynamic dispatch).  Whether or not that maybe uses function multiversioning from GCC, I'm not sure (though the target nomenclature surely suggests that).  I'm not sure how portable that GCC feature is or how good it might be at collapsing and inlining them when they're nested.  For many targets, supporting the instructions tends to not be a narrow enough vector to assume to use a given function (as with the current pmull+eor3 PR for CRC32).

phprus:
`loongarch64` supports `BITREV` instruction. After merging this PR, I can add usage of `BITREV` instruction.

Dead2:
I am not able to reliably benchmark this it seems.

But how about making bi_reverse take uint16_t code directly? After all it already converts code down to uint16_t, and it returns uint16_t.

In `trees.c` we are upconverting from uint16_t to uint32_t for that function call, only for the function to convert it back down to uint16_t.
`arch/s390/dfltcc_deflate.c` would need to have a cast added.
No other users.

nmoinvaz:
@Dead2 done, I also found another instance where we can use `uint16_t`.

Dead2 (approved):
LGTM

PR #2043, opened 2025-12-06
Fix for potentially uninitialized local variable ft used.

phprus:
The `FUNCTABLE_VERIFY_ASSIGN` macro reads values from `ft` and checks that they are not NULL, but the `ft` variable is not initialized.


nmoinvaz (approved):

mtl1979 (approved):
LGTM

Dead2 (approved):

PR #2044, opened 2025-12-07
Optimize code length counting in inflate_table

nmoinvaz:
I forget where I saw this originally (although it was slighly different), maybe here or a derivative:

Similar to https://github.com/dougallj/zlib-dougallj/commit/f23fa25aa168ef782bab5e7cd6f9df50d7bb5eb2. 

AI said SIMD was slower for ~286 symbols, so instead we just unroll the loop. It said you would need at least 1000 symbols to make it advantageous.

### DIFF
--- a/inftrees.c
+++ b/inftrees.c
@@ -24,22 +24,35 @@ const char PREFIX(inflate_copyright)[] = " inflate 1.3.1 Copyright 1995-2024 Mar
 /* Count number of codes for each code length. */
 static inline void count_lengths(uint16_t *lens, int codes, uint16_t *count) {
     int sym;
-    static const ALIGNED_(32) uint8_t one[32] = {
-        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    static const ALIGNED_(16) uint8_t one[256] = {
         1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0,
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1
     };
 
 #if defined(__ARM_NEON) || defined(__ARM_NEON__)
     uint8x16_t s1 = vdupq_n_u8(0);
     uint8x16_t s2 = vdupq_n_u8(0);
 
-    const uint8_t *p = &one[16];
     if (codes & 1) {
-        s1 = vld1q_u8(&p[-lens[0]]);
+        s1 = vld1q_u8(&one[16 * lens[0]]);
     }
     for (sym = codes & 1; sym < codes; sym += 2) {
-      s1 = vaddq_u8(s1, vld1q_u8(&p[-lens[sym]]));
-      s2 = vaddq_u8(s2, vld1q_u8(&p[-lens[sym+1]]));
+      s1 = vaddq_u8(s1, vld1q_u8(&one[16 * lens[sym]]));
+      s2 = vaddq_u8(s2, vld1q_u8(&one[16 * lens[sym+1]]));
     }
 
     vst1q_u16(&count[0], vaddl_u8(vget_low_u8(s1), vget_low_u8(s2)));
@@ -49,13 +62,12 @@ static inline void count_lengths(uint16_t *lens, int codes, uint16_t *count) {
     __m128i s1 = _mm_setzero_si128();
     __m128i s2 = _mm_setzero_si128();
 
-    const uint8_t *p = (uint8_t*)&one[16];
     if (codes & 1) {
-        s1 = _mm_loadu_si128((const __m128i*)&p[-lens[0]]);
+        s1 = _mm_load_si128((const __m128i*)&one[16 * lens[0]]);
     }
     for (sym = codes & 1; sym < codes; sym += 2) {
-        s1 = _mm_add_epi8(s1, _mm_loadu_si128((const __m128i*)&p[-lens[sym]]));   // vaddq_u8
-        s2 = _mm_add_epi8(s2, _mm_loadu_si128((const __m128i*)&p[-lens[sym+1]]));
+        s1 = _mm_add_epi8(s1, _mm_load_si128((const __m128i*)&one[16 * lens[sym]]));  // vaddq_u8
+        s2 = _mm_add_epi8(s2, _mm_load_si128((const __m128i*)&one[16 * lens[sym+1]]));
     }
 
 #  if defined(__AVX2__)
--- a/inftrees.c
+++ b/inftrees.c
@@ -7,6 +7,12 @@
 #include "zutil.h"
 #include "inftrees.h"
 
+#if defined(__SSE2__)
+#  include "arch/x86/x86_intrins.h"
+#elif defined(__ARM_NEON) || defined(__ARM_NEON__)
+#  include "arch/arm/neon_intrins.h"
+#endif
+
 const char PREFIX(inflate_copyright)[] = " inflate 1.3.1 Copyright 1995-2024 Mark Adler ";
 /*
   If you use the zlib library in a product, an acknowledgment is welcome
@@ -14,6 +20,72 @@ const char PREFIX(inflate_copyright)[] = " inflate 1.3.1 Copyright 1995-2024 Mar
   include such an acknowledgment, I would appreciate that you keep this
   copyright string in the executable of your product.
  */
+
+/* Count number of codes for each code length. */
+static inline void count_lengths(uint16_t *lens, int codes, uint16_t *count) {
+    int sym;
+    static const ALIGNED_(32) uint8_t one[32] = {
+        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    };
+
+#if defined(__ARM_NEON) || defined(__ARM_NEON__)
+    uint8x16_t s1 = vdupq_n_u8(0);
+    uint8x16_t s2 = vdupq_n_u8(0);
+
+    const uint8_t *p = &one[16];
+    if (codes & 1) {
+        s1 = vld1q_u8(&p[-lens[0]]);
+    }
+    for (sym = codes & 1; sym < codes; sym += 2) {
+      s1 = vaddq_u8(s1, vld1q_u8(&p[-lens[sym]]));
+      s2 = vaddq_u8(s2, vld1q_u8(&p[-lens[sym+1]]));
+    }
+
+    vst1q_u16(&count[0], vaddl_u8(vget_low_u8(s1), vget_low_u8(s2)));
+    vst1q_u16(&count[8], vaddl_u8(vget_high_u8(s1), vget_high_u8(s2)));
+
+#elif defined(__SSE2__)
+    __m128i s1 = _mm_setzero_si128();
+    __m128i s2 = _mm_setzero_si128();
+
+    const uint8_t *p = (uint8_t*)&one[16];
+    if (codes & 1) {
+        s1 = _mm_loadu_si128((const __m128i*)&p[-lens[0]]);
+    }
+    for (sym = codes & 1; sym < codes; sym += 2) {
+        s1 = _mm_add_epi8(s1, _mm_loadu_si128((const __m128i*)&p[-lens[sym]]));   // vaddq_u8
+        s2 = _mm_add_epi8(s2, _mm_loadu_si128((const __m128i*)&p[-lens[sym+1]]));
+    }
+
+#  if defined(__AVX2__)
+    __m256i w1 = _mm256_cvtepu8_epi16(s1);
+    __m256i w2 = _mm256_cvtepu8_epi16(s2);
+    __m256i sum = _mm256_add_epi16(w1, w2);
+
+    _mm256_storeu_si256((__m256i*)&count[0], sum);
+#  else
+    __m128i zero = _mm_setzero_si128();
+
+    __m128i s1_lo = _mm_unpacklo_epi8(s1, zero);
+    __m128i s2_lo = _mm_unpacklo_epi8(s2, zero);
+    __m128i sum_lo = _mm_add_epi16(s1_lo, s2_lo);
+    _mm_storeu_si128((__m128i*)&count[0], sum_lo);
+
+    __m128i s1_hi = _mm_unpackhi_epi8(s1, zero);
+    __m128i s2_hi = _mm_unpackhi_epi8(s2, zero);
+    __m128i sum_hi = _mm_add_epi16(s1_hi, s2_hi);
+    _mm_storeu_si128((__m128i*)&count[8], sum_hi);
+#  endif
+#else
+    int len;
+    for (len = 0; len <= MAX_BITS; len++)
+        count[len] = 0;
+    for (sym = 0; sym < codes; sym++)
+        count[lens[sym]]++;
+    Z_UNUSED(one);
+#endif
+}
 
 /*
    Build a set of tables to decode the provided canonical Huffman code.
@@ -28,7 +100,7 @@ const char PREFIX(inflate_copyright)[] = " inflate 1.3.1 Copyright 1995-2024 Mar
    longest code or if it is less than the shortest code.
  */
 int Z_INTERNAL zng_inflate_table(codetype type, uint16_t *lens, unsigned codes,
-                                code * *table, unsigned *bits, uint16_t *work) {
+                                 code * *table, unsigned *bits, uint16_t *work) {
     unsigned len;               /* a code's length in bits */
     unsigned sym;               /* index of code symbols */
     unsigned min, max;          /* minimum and maximum code lengths */
@@ -47,8 +119,8 @@ int Z_INTERNAL zng_inflate_table(codetype type, uint16_t *lens, unsigned codes,
     const uint16_t *base;       /* base value table to use */
     const uint16_t *extra;      /* extra bits table to use */
     unsigned match;             /* use base and extra for symbol >= match */
-    uint16_t count[MAX_BITS+1];  /* number of codes of each length */
-    uint16_t offs[MAX_BITS+1];   /* offsets in table for each length */
+    uint16_t count[MAX_BITS+1]; /* number of codes of each length */
+    uint16_t offs[MAX_BITS+1];  /* offsets in table for each length */
     static const uint16_t lbase[31] = { /* Length codes 257..285 base */
         3, 4, 5, 6, 7, 8, 9, 10, 11, 13, 15, 17, 19, 23, 27, 31,
         35, 43, 51, 59, 67, 83, 99, 115, 131, 163, 195, 227, 258, 0, 0};
@@ -96,10 +168,7 @@ int Z_INTERNAL zng_inflate_table(codetype type, uint16_t *lens, unsigned codes,
      */
 
     /* accumulate lengths for codes (assumes lens[] all in 0..MAXBITS) */
-    for (len = 0; len <= MAX_BITS; len++)
-        count[len] = 0;
-    for (sym = 0; sym < codes; sym++)
-        count[lens[sym]]++;
+    count_lengths(lens, codes, count);
 
     /* bound code lengths, force root to be within code lengths */
     root = *bits;
--- a/arch/x86/x86_intrins.h
+++ b/arch/x86/x86_intrins.h
@@ -1,6 +1,10 @@
 #ifndef X86_INTRINS_H
 #define X86_INTRINS_H
 
+#ifdef __SSE2__
+#include <emmintrin.h>
+#endif
+
 /* Unfortunately GCC didn't support these things until version 10.
  * Similarly, AppleClang didn't support them in Xcode 9.2 but did in 9.3.
  */


inftrees.c
```diff
@@ -98,7 +98,17 @@ int Z_INTERNAL zng_inflate_table(codetype type, uint16_t *lens, unsigned codes,
     /* accumulate lengths for codes (assumes lens[] all in 0..MAXBITS) */
     for (len = 0; len <= MAX_BITS; len++)
         count[len] = 0;
-    for (sym = 0; sym < codes; sym++)
+
+    /* Process symbols in parallel batches of 4 */
+    sym = 0;
+    for (; sym + 3 < codes; sym += 4) {
```
Dead2: **thought:** Are the compilers smart enough to make this a compare against a once-calculated `codes - 3` so it avoids the addition in `sym + 3` every iteration?

nmoinvaz: If you want to inspect the assembly:
https://godbolt.org/z/qTnMeG5q3
But I think the answer is yes.

dougallj:
I'd probably avoid manually unrolling here unless you can show a speedup in benchmarks.

The problem my patch was trying to address is a read-after-write hazard, when a `lens[sym]` value is the same as on previous iterations, which it frequently is. This can cause a variety of problems, from expensive pipeline flushes due to misprediction, to ~5 cycle latency per iteration on Apple M1 when working as expected (as opposed to 2 cycle latency for the SIMD, with two parallel accumulators giving 1c throughput, for a ~5x speedup).  Store to load forwarding is an area that has improved significantly in newer M4/M5 CPUs, so it mightn't be as bad as I think, but this kind of histogramming code tends to bring out the problems and trip up the predictors. I can imagine unrolling helping with this under a somewhat bizarre set of circumstances, but unrolling usually solves instruction-fetch and loop-overhead limits, whereas this loop is limited by memory disambiguation and store-to-load forwarding latency.

dougallj:
(FWIW, I think [Claude's](https://claude.ai/share/39790b2c-5f24-4377-94d0-da081908fc15) and [ChatGPT's](https://chatgpt.com/share/6936fa0d-b710-8004-a23e-37068ddb7564) answers are closer to my understanding than a ~1000 symbol estimate. I'm not sure how that compares to what you're using or what you were asking. Obviously a bad SIMD implementation could cause a slow down, but I've only seen thresholds of that magnitude when flushing the pipeline (the thing I wrote the SIMD code to avoid), or maybe on ancient AVX-512, where using it caused the whole system to slow down for a bit.)

nmoinvaz:
>  I'm not sure how that compares to what you're using or what you were asking. 

I think my AI was cooked. Today, I pushed back on its estimate and it said the only way to know would be to benchmark it, so it essentially gave up. I think it was hung up on increase memory bandwidth and assumed it translated to a slower implementation.

nmoinvaz:
I think if I were to implement SIMD, I would need a `functable` function.., right?

KungFuJesus:
> I think if I were to implement SIMD, I would need a `functable` function.., right?

Not exactly - I think asimd is a required feature for aarch64 support so having an ifdef there would _probably_ work and compile compatible code in all places it's used.  Having a function table entry for inflate_table probably wouldn't be the worst thing in the world but you can get away without it.

dougallj:
Yeah, I don't think it matters enough to justify dynamic dispatch, you can just use the base SIMD ISAs where available  – it doesn't meaningfully benefit from any fancy instructions.

I put together a quick (not tested or benchmarked) sketch of x86 support for my version: https://godbolt.org/z/fojxrEo4T (the AVX2 tail just generates smaller/prettier code, you definitely couldn't measure that in a benchmark)

(Unfortunately ARM64 gcc (not pictured) makes a mess of the code generation, but it doesn't stop the underlying optimisation from working, so I guess we don't need to worry that much.)

KungFuJesus:
It requires SSSE3 but it's slightly better throughput to do pmaddusw with a vector of packed (16 bit) ones. You can do two of those in 5 cycles or so. https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#techs=SSE_ALL,AVX_ALL&ig_expand=1857,4200,4236,4233&text=Maddu

KungFuJesus:
Err an adjacent sum doesn't exactly work for this without shifting bytes and recombining, which may be just as much as a sign extending 8 bit to 16 bit move. Also given that it's the tail handling at the very end I suppose it doesn't matter too much, anyhow.

nmoinvaz:
> Having a function table entry for inflate_table probably wouldn't be the worst thing in the world but you can get away without it.

It would increase the code size * each supported arch. If I don't do functable, then it would really only benefit native architecture compilation.

nmoinvaz:
Here is Quick Benchmark: https://quick-bench.com/q/DW80QBZoRy1UBwUEH-JMTeqYDGA
Shows loop unrolling performs poorly at least on x86. 
AVX2 version 1.5 times faster than scalar version.

dougallj:
Yeah, the choice of random code lengths avoids some of the worst of the problem too – if I change `lens[i] = rand() % (MAX_BITS + 1)` to `lens[i] = 7` I see a 2.7x speedup on quick-bench. I would guess real world files land somewhere between the two.

Conversely, just for your amusement, if you deliberately construct a sequence to avoid all read-after-write hazards, i.e. `lens[i] = i % MAX_BITS` then the scalar code runs about the same speed as the SSE2/AVX2 code. I don't think this is remotely plausible in a real world Huffman tree though.

dougallj:
> If I don't do functable, then it would really only benefit native architecture compilation.

All x86-64 CPUs must have SSE2 and all Armv8-A CPUs must have NEON, so those two don't need `-march=native` compilation to benefit, and that's probably at least 95% of users? But yeah, some other, older architectures might miss out.

KungFuJesus:
Right that's what I was getting at - the compile time check against aarch64 will cover basically any 64 bit arm compilation without issue and if you stick to only SSE2 or so support for the x86 variant, every 64 bit CPU should compile it just fine and be runtime compatible.  The small sliver missing out here might be 32 bit consumers not compiling with SSE support (that is going to be a very very narrow margin).  

For an altivec variant of this it'd probably be better suited to be a permutation vector as arbitrary unaligned loads don't really become a thing until power 7ish and I don't think they become cheap for a little while after.

KungFuJesus:
It's probably technically cacheline pollution but I think the best implementation of that histogram function for all potatoes ends up being aligned loads for all 16 possible permutations (ones becomes 8 times larger).  Unaligned loads for pre-nehalem take a beating (Core 2 era) and for altivec CPUs we'd need to synthesize an unaligned load with vec_vsl/vsr and vec_perm, anyhow.  

Unfortunately all of the cheap ways to push a 1 into the correct bin sans an unaligned load end up requiring compile time immediates for shifts, which would end up inducing a huge inflation of code size and a lot of branching.  It's probably not the worst thing in the world, though, as 256 bytes still fits into like 4 cachelines for a lot of architectures.

nmoinvaz:
I have implemented the vectorized versions of code length counting.

nmoinvaz:
Rebased.

KungFuJesus:
Is there an upper bound on codes? @dougallj's implementation does straight 8 bit additions until the tail handling. This is better if possible.

nmoinvaz:
AFAIK 286. I couldn’t get the godbolt code working it kept crashing so…

KungFuJesus:
If 286 is the upper bound and you split the sums over 2 registers then that should work with straight 8 bit additions

nmoinvaz:
Ok, I threw it back in, lets see what happens, maybe it was something else.

KungFuJesus:
You have to widen s1 and s2 outside of the loop to 16 bits and sum those together (you can overflow 8 bits on any given bin if unlucky). Spreading the sum over two makes it impossible to overflow for each intermediate sum and then you only need to widen at the very end.

Edit: err whoops you are, sorry it's just lower than I expected.

dougallj:
I started looking at the static analysis failure and got a bit puzzled.

It looks like I wrote this code expecting to have to handle a length of 16. Which can't happen in Huffman codes (obviously, since I had to extend the output array by one element, which the scalar version doesn't write to). I guess I really didn't want to compute the unused count[0] sum or something?

It doesn't make a huge difference, but FWIW I'd change my godbolt code like this:

```diff
 void count_lengths(const uint8_t *lens, int codes, uint16_t *count) {
     int sym;
     static const __attribute__((aligned(32))) uint8_t one[32] = {
-        0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1,
         0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
     };
 
 #if defined(__ARM_NEON) || defined(__ARM_NEON__)
     uint8x16_t s1 = vdupq_n_u8(0);
     uint8x16_t s2 = vdupq_n_u8(0);
 
     const uint8_t *p = &one[16];
     if (codes & 1) {
         s1 = vld1q_u8(&p[-lens[0]]);
     }
     for (sym = codes & 1; sym < codes; sym += 2) {
       s1 = vaddq_u8(s1, vld1q_u8(&p[-lens[sym]]));
       s2 = vaddq_u8(s2, vld1q_u8(&p[-lens[sym+1]]));
     }
 
-    count[0] = 0;
-    vst1q_u16(&count[1], vaddl_u8(vget_low_u8(s1), vget_low_u8(s2)));
-    vst1q_u16(&count[9], vaddl_u8(vget_high_u8(s1), vget_high_u8(s2)));
+    vst1q_u16(&count[0], vaddl_u8(vget_low_u8(s1), vget_low_u8(s2)));
+    vst1q_u16(&count[8], vaddl_u8(vget_high_u8(s1), vget_high_u8(s2)));
 
 #elif defined(__SSE2__)
     __m128i s1 = _mm_setzero_si128();
     __m128i s2 = _mm_setzero_si128();
 
     const uint8_t *p = (uint8_t*)&one[16];
     if (codes & 1) {
         s1 = _mm_loadu_si128((const __m128i*)&p[-lens[0]]);
     }
     for (sym = codes & 1; sym < codes; sym += 2) {
         s1 = _mm_add_epi8(s1, _mm_loadu_si128((const __m128i*)&p[-lens[sym]]));   // vaddq_u8
         s2 = _mm_add_epi8(s2, _mm_loadu_si128((const __m128i*)&p[-lens[sym+1]]));
     }
 
-    count[0] = 0;
 
 #if defined(__AVX2__)
     __m256i w1 = _mm256_cvtepu8_epi16(s1);
     __m256i w2 = _mm256_cvtepu8_epi16(s2);
     __m256i sum = _mm256_add_epi16(w1, w2);
 
-    _mm256_storeu_si256((__m256i*)&count[1], sum);
+    _mm256_storeu_si256((__m256i*)&count[0], sum);
 #else
     __m128i zero = _mm_setzero_si128();
 
     __m128i s1_lo = _mm_unpacklo_epi8(s1, zero);
     __m128i s2_lo = _mm_unpacklo_epi8(s2, zero);
     __m128i sum_lo = _mm_add_epi16(s1_lo, s2_lo);
-    _mm_storeu_si128((__m128i*)&count[1], sum_lo);
+    _mm_storeu_si128((__m128i*)&count[0], sum_lo);
 
     __m128i s1_hi = _mm_unpackhi_epi8(s1, zero);
     __m128i s2_hi = _mm_unpackhi_epi8(s2, zero);
     __m128i sum_hi = _mm_add_epi16(s1_hi, s2_hi);
-    _mm_storeu_si128((__m128i*)&count[9], sum_hi);
+    _mm_storeu_si128((__m128i*)&count[8], sum_hi);
 #endif
 #else
     for (len = 0; len <= 15; len++)
         count[len] = 0;
     for (sym = 0; sym < codes; sym++)
         count[lens[sym]]++;
 #endif
 }
```

Then you can revert the array size too:
```diff
-    uint16_t count[MAX_BITS+2]; /* number of codes of each length */
+    uint16_t count[MAX_BITS+1]; /* number of codes of each length */
```

Not that it matters, but that should also make the unused value in `count[0]` be consistent with the scalar version.

(I'm guessing the static analysis noticed that the scalar version didn't initialise the whole array, in which case this should fix that too? Not sure what ifdefs it's using.)

phprus:
Are there any plans to use `functable`?

KungFuJesus:
> Are there any plans to use `functable`?

If this isn't being inlined anywhere I don't there's much harm in doing so (other than the baggage of setting it up in the functable, anyhow). Code gets a little bit larger but if the function call was going to happen anyway, it's probably net neutral.

If we want this to be still be fast for some potatoes, it's probably better to set up 15 or 16 configurations of ones and do purely aligned loads.

nmoinvaz:
@dougallj that seems to have done the trick! Thanks.

KungFuJesus:
[0001-Force-purely-aligned-loads.patch](https://github.com/user-attachments/files/24092553/0001-Force-purely-aligned-loads.patch)
@nmoinvaz, this is what I was referring to.

dougallj:
Yeah, amusingly it makes the aarch64 code very pretty (since aarch64 has a *16 addressing mode, but x86 doesn't).

I don't have the numbers, but somehow four cache-lines just feels like too much to me for such a cold loop. I think the hit on 13+ year old Intel CPUs is likely to be pretty unnoticeable (the loads do take 2 cycles, but the loop just doesn't run for very long). But then, it probably is fine either way.

I'm fond of my own approach, but if we're concerned, maybe Apple's technique (scalar accesses, but split across two tables) is a better, more portable option?

```c
unsigned short cnt[2][16];

__builtin_memset(cnt, 0, sizeof(cnt));
for (sym = 0; sym + 2 <= codes; sym += 2)
{
  cnt[0][lens[sym    ]]++;
  cnt[1][lens[sym + 1]]++;
}
// AVX usage in the kernel is problematic. Thus, use 16 byte vectors.
((packed_ushort8*)count)[0] = ((packed_ushort8*)cnt[0])[0] + ((packed_ushort8*)cnt[1])[0];
((packed_ushort8*)count)[1] = ((packed_ushort8*)cnt[0])[1] + ((packed_ushort8*)cnt[1])[1];
if (sym < codes) count[lens[sym]]++;
```

KungFuJesus:
It's kind of silly to me that there's no compact representation to shift this 1 value around from a base register.  Everything to shift a value arbitrarily within the 128 bit register requires compile time immediates.  Personally I'd like to keep things fast for the potatoes (I was using one when I started using this project).  This also applies equally to a G5 where this is far worse, as you have to do vec_lvsl to do the nearest aligned load and let it produce for you a permutation vector to swizzle subsequent values in.  If that's not bad enough, that pattern is really only meant for a single alignment offset, where you can recycle a permutation vector and leverage the previous half of the load at the top of a loop.  Of course, this is a machine from 2005, but I do think one of the cool things about this project is that we are still useful on hardware like this (the vectorized Chorba CRC sets us apart for pre clmul capable CPUs).

If we were to dispatch to the function table of course, I'd conditionally define this and use the more compact one for the newer CPU variants.

mtl1979:
For a long time I did use zlib-ng on a processor model released in 2002... It was pretty much stretching the limits of vector optimizing as just having recent enough operating system that supported new enough cmake was getting harder and harder...

KungFuJesus:
I suspect the cortex a50 something portion of the odroid n2 might benefit from this and that's not super old.

I'll write a microbenchmark and see if I can confirm the gains and more importantly verify no harm to newer CPUs.

dougallj:
Yeah, my only concern is cache misses, which'd hurt more on in-order, cache-starved cores like the Cortex-A53. We don't want to have to wait for four sequential loads from DRAM to run our little ~300 cycle loop. The other risk is this might not show up in microbenchmarks, since the table will be reused with no intervening work. I'd recommend benchmarking inflate instead – it might only be 0.5% or something, but if we can't measure it there, then we're probably wasting our time. (The relative importance of this code will increase as we land other optimisations to the hot loop, but not by a huge amount.)

dougallj:
(Oh, and if you do apply the aligned-access diff, I'd drop `ALIGNED_(32)` to `ALIGNED_(16)`, just so it makes sense. 32 was to avoid crossing a cache line boundary or page boundary with an unaligned access, which can be various degrees of awful on different microarchitectures.)

KungFuJesus:
I don't suspect our trivial inflate benchmark we already have will exercise this much, but the png decoding benchmark with some test pngs I use will (to a small extent, anyway). I'll probably get around to it either tonight or tomorrow.

Err actually, it looks like uncompress_bench has a representative amount of samples in inflate_table so I probably can use that.

KungFuJesus:
Ok, so I'm going to try this over a couple of machines today but the first one I've tried is the m2 on the blizzard cores.  There does seem to be an improvement with the alignment, albeit barely outside of the standard deviation in the benchmark:

# Unaligned
```
adam@m2-asahi:~/zlib-ng/build/test/benchmarks$ taskset -c 0 setarch `uname -m` -R ./benchmark_zlib --benchmark_filter=uncompress_bench/1048576 --benchmark_repetitions=5
2025-12-12T10:06:10-05:00
Running ./benchmark_zlib
Run on (8 X 2424 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB (x8)
  L1 Instruction 128 KiB (x8)
  L2 Unified 4096 KiB (x2)
Load Average: 0.19, 0.09, 0.02
-------------------------------------------------------------------------------------------
Benchmark                                                 Time             CPU   Iterations
-------------------------------------------------------------------------------------------
uncompress_bench/uncompress_bench/1048576            143206 ns       142843 ns         4897
uncompress_bench/uncompress_bench/1048576            143141 ns       142824 ns         4897
uncompress_bench/uncompress_bench/1048576            143213 ns       142950 ns         4897
uncompress_bench/uncompress_bench/1048576            142879 ns       142633 ns         4897
uncompress_bench/uncompress_bench/1048576            142971 ns       142732 ns         4897
uncompress_bench/uncompress_bench/1048576_mean       143082 ns       142796 ns            5
uncompress_bench/uncompress_bench/1048576_median     143141 ns       142824 ns            5
uncompress_bench/uncompress_bench/1048576_stddev        150 ns          120 ns            5
uncompress_bench/uncompress_bench/1048576_cv           0.10 %          0.08 %             5
```

# Aligned
```
adam@m2-asahi:~/zlib-ng/build/test/benchmarks$ taskset -c 0 setarch `uname -m` -R ./benchmark_zlib --benchmark_filter=uncompress_bench/1048576 --benchmark_repetitions=5
2025-12-12T10:04:19-05:00
Running ./benchmark_zlib
Run on (8 X 2424 MHz CPU s)
CPU Caches:
  L1 Data 64 KiB (x8)
  L1 Instruction 128 KiB (x8)
  L2 Unified 4096 KiB (x2)
Load Average: 0.18, 0.07, 0.02
-------------------------------------------------------------------------------------------
Benchmark                                                 Time             CPU   Iterations
-------------------------------------------------------------------------------------------
uncompress_bench/uncompress_bench/1048576            143166 ns       142819 ns         4901
uncompress_bench/uncompress_bench/1048576            143111 ns       142800 ns         4901
uncompress_bench/uncompress_bench/1048576            143027 ns       142771 ns         4901
uncompress_bench/uncompress_bench/1048576            142872 ns       142628 ns         4901
uncompress_bench/uncompress_bench/1048576            142914 ns       142641 ns         4901
uncompress_bench/uncompress_bench/1048576_mean       143018 ns       142732 ns            5
uncompress_bench/uncompress_bench/1048576_median     143027 ns       142771 ns            5
uncompress_bench/uncompress_bench/1048576_stddev        125 ns         90.7 ns            5
uncompress_bench/uncompress_bench/1048576_cv           0.09 %          0.06 %             5
```

That's with 5 trials, with google benchmark doing its google benchmark thing where it samples for each trial until variance is baked to some reasonable level (or at least that's my vague understanding of what happens with that iterations count).

Now, I'd expect Apple has the best behavior for unaligned codes for their "little", I'll try odroid next and then I'll boot up the Q9650 where I'd expect some of the worst results to see the damage.  I guess the important bit is that the purely aligned version is not hurting this benchmark for the shiny relatively new M2.

We lose the fancy 16 modifier for the vectorized load probably but I suspect we can take advantage of the 0 row never being used to shrink this a little.  Though, maybe that one isn't worth it.

KungFuJesus:
So the odroid n2 had some pretty severe swings in variance.  Disabling cpu scaling helped that a bit, but it's still higher.  I upped the benchmark repetitions to 20:

# Unaligned
```
[adam@odroid benchmarks]$ taskset -c 0 setarch `uname -m` -R ./benchmark_zlib --benchmark_filter=uncompress_bench/1048576 --benchmark_repetitions=20
2025-12-12T10:50:53-05:00
Running ./benchmark_zlib
Run on (6 X 1992 MHz CPU s)
Load Average: 0.36, 0.39, 0.45
-------------------------------------------------------------------------------------------
Benchmark                                                 Time             CPU   Iterations
-------------------------------------------------------------------------------------------
uncompress_bench/uncompress_bench/1048576           1115318 ns      1105856 ns          629
uncompress_bench/uncompress_bench/1048576           1109291 ns      1102497 ns          629
uncompress_bench/uncompress_bench/1048576           1104348 ns      1100438 ns          629
uncompress_bench/uncompress_bench/1048576           1107061 ns      1103048 ns          629
uncompress_bench/uncompress_bench/1048576           1105774 ns      1101875 ns          629
uncompress_bench/uncompress_bench/1048576           1108342 ns      1104365 ns          629
uncompress_bench/uncompress_bench/1048576           1107863 ns      1103802 ns          629
uncompress_bench/uncompress_bench/1048576           1105783 ns      1101790 ns          629
uncompress_bench/uncompress_bench/1048576           1107142 ns      1102894 ns          629
uncompress_bench/uncompress_bench/1048576           1106016 ns      1102008 ns          629
uncompress_bench/uncompress_bench/1048576           1106218 ns      1102180 ns          629
uncompress_bench/uncompress_bench/1048576           1106738 ns      1102577 ns          629
uncompress_bench/uncompress_bench/1048576           1104629 ns      1100624 ns          629
uncompress_bench/uncompress_bench/1048576           1106538 ns      1102448 ns          629
uncompress_bench/uncompress_bench/1048576           1106960 ns      1102722 ns          629
uncompress_bench/uncompress_bench/1048576           1106753 ns      1102500 ns          629
uncompress_bench/uncompress_bench/1048576           1105843 ns      1101667 ns          629
uncompress_bench/uncompress_bench/1048576           1107197 ns      1102983 ns          629
uncompress_bench/uncompress_bench/1048576           1106033 ns      1102092 ns          629
uncompress_bench/uncompress_bench/1048576           1106752 ns      1102573 ns          629
uncompress_bench/uncompress_bench/1048576_mean      1107030 ns      1102547 ns           20
uncompress_bench/uncompress_bench/1048576_median    1106745 ns      1102498 ns           20
uncompress_bench/uncompress_bench/1048576_stddev       2259 ns         1192 ns           20
uncompress_bench/uncompress_bench/1048576_cv           0.20 %          0.11 %            20
```

# Aligned
```
[adam@odroid benchmarks]$ taskset -c 0 setarch `uname -m` -R ./benchmark_zlib --benchmark_filter=uncompress_bench/1048576 --benchmark_repetitions=20
2025-12-12T10:39:51-05:00
Running ./benchmark_zlib
Run on (6 X 1992 MHz CPU s)
Load Average: 0.29, 0.82, 0.55
-------------------------------------------------------------------------------------------
Benchmark                                                 Time             CPU   Iterations
-------------------------------------------------------------------------------------------
uncompress_bench/uncompress_bench/1048576           1115306 ns      1105323 ns          630
uncompress_bench/uncompress_bench/1048576           1108632 ns      1102498 ns          630
uncompress_bench/uncompress_bench/1048576           1109593 ns      1103459 ns          630
uncompress_bench/uncompress_bench/1048576           1106959 ns      1101086 ns          630
uncompress_bench/uncompress_bench/1048576           1104837 ns      1100278 ns          630
uncompress_bench/uncompress_bench/1048576           1103126 ns      1098758 ns          630
uncompress_bench/uncompress_bench/1048576           1107648 ns      1103405 ns          630
uncompress_bench/uncompress_bench/1048576           1105302 ns      1100854 ns          630
uncompress_bench/uncompress_bench/1048576           1102912 ns      1098546 ns          630
uncompress_bench/uncompress_bench/1048576           1103877 ns      1099726 ns          630
uncompress_bench/uncompress_bench/1048576           1102345 ns      1098018 ns          630
uncompress_bench/uncompress_bench/1048576           1101503 ns      1097221 ns          630
uncompress_bench/uncompress_bench/1048576           1105720 ns      1101392 ns          630
uncompress_bench/uncompress_bench/1048576           1107849 ns      1103631 ns          630
uncompress_bench/uncompress_bench/1048576           1105522 ns      1101202 ns          630
uncompress_bench/uncompress_bench/1048576           1102369 ns      1098148 ns          630
uncompress_bench/uncompress_bench/1048576           1102166 ns      1098005 ns          630
uncompress_bench/uncompress_bench/1048576           1102727 ns      1098304 ns          630
uncompress_bench/uncompress_bench/1048576           1104727 ns      1100453 ns          630
uncompress_bench/uncompress_bench/1048576           1105644 ns      1101482 ns          630
uncompress_bench/uncompress_bench/1048576_mean      1105438 ns      1100589 ns           20
uncompress_bench/uncompress_bench/1048576_median    1105069 ns      1100654 ns           20
uncompress_bench/uncompress_bench/1048576_stddev       3289 ns         2269 ns           20
uncompress_bench/uncompress_bench/1048576_cv           0.30 %          0.21 %            20
```

The little cores on this guy also follow the convention that the m2 does (0-1), opposite of what raptor lake orders its efficiency vs performance cores.  The variance is all over the place with this thing, though.  I don't have active cooling for it but it also looks as though Majaro ARM (sorry, forgot I wiped ALARM on this) configuration for this device puts it at the n2 rather than n2+ clocks, so I don't _think_ it's thermal throttling.  But it does seem that aligned wins for this thing, too.

KungFuJesus:
Sorry, went on a side quest.  Updated the whole distro, booted a new kernel, and then ethernet was broken.  Assumed it was a goofy dtb/firmware thing so I managed to SPI flash the bootrom and boot generic manjaro arm with older kernel.  Switched to "testing" and now no ethernet again.  Putting that down and start on the Q9650 shortly.

KungFuJesus:
Sorry, almost have a situation where I can test this.  As it turns out, kernel 6.14 introduced something that hangs on bootup (it wasn't just my kernel compilation config, evidently) with Core 2's and/or Q45 chipsets.  I'm booting a live media with a slightly older Ubuntu version to hopefully get me up there.  Otherwise, I do have a working Gentoo installation on an SSD I could also boot up but I'd need to defang some stuff before putting it on the network.

KungFuJesus:
Ok, Q9650.

# Unaligned
```
mythtv@oldmythbox ~/zlib-ng/build_alignment/test/benchmarks $ taskset -c 0 setarch `uname -m` -R ./benchmark_zlib --benchmark_filter=uncompress_bench/1048576 --benchmark_repetitions=5
2025-12-12T15:53:36-05:00
Running ./benchmark_zlib
Run on (4 X 3003 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x4)
  L1 Instruction 32 KiB (x4)
  L2 Unified 6144 KiB (x2)
Load Average: 1.47, 0.79, 0.30
-------------------------------------------------------------------------------------------
Benchmark                                                 Time             CPU   Iterations
-------------------------------------------------------------------------------------------
uncompress_bench/uncompress_bench/1048576            462184 ns       462179 ns         1514
uncompress_bench/uncompress_bench/1048576            462013 ns       462013 ns         1514
uncompress_bench/uncompress_bench/1048576            461982 ns       461975 ns         1514
uncompress_bench/uncompress_bench/1048576            462328 ns       462322 ns         1514
uncompress_bench/uncompress_bench/1048576            462317 ns       462310 ns         1514
uncompress_bench/uncompress_bench/1048576_mean       462165 ns       462160 ns            5
uncompress_bench/uncompress_bench/1048576_median     462184 ns       462179 ns            5
uncompress_bench/uncompress_bench/1048576_stddev        163 ns          162 ns            5
uncompress_bench/uncompress_bench/1048576_cv           0.04 %          0.04 %             5
```

# Aligned
```
mythtv@oldmythbox ~/zlib-ng/build_alignment/test/benchmarks $ taskset -c 0 setarch `uname -m` -R ./benchmark_zlib --benchmark_filter=uncompress_bench/1048576 --benchmark_repetitions=5
2025-12-12T15:54:43-05:00
Running ./benchmark_zlib
Run on (4 X 3003 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x4)
  L1 Instruction 32 KiB (x4)
  L2 Unified 6144 KiB (x2)
Load Average: 1.11, 0.78, 0.33
-------------------------------------------------------------------------------------------
Benchmark                                                 Time             CPU   Iterations
-------------------------------------------------------------------------------------------
uncompress_bench/uncompress_bench/1048576            462061 ns       462051 ns         1514
uncompress_bench/uncompress_bench/1048576            462139 ns       462135 ns         1514
uncompress_bench/uncompress_bench/1048576            461966 ns       461966 ns         1514
uncompress_bench/uncompress_bench/1048576            462137 ns       462133 ns         1514
uncompress_bench/uncompress_bench/1048576            462200 ns       462196 ns         1514
uncompress_bench/uncompress_bench/1048576_mean       462101 ns       462096 ns            5
uncompress_bench/uncompress_bench/1048576_median     462137 ns       462133 ns            5
uncompress_bench/uncompress_bench/1048576_stddev       89.9 ns         89.4 ns            5
uncompress_bench/uncompress_bench/1048576_cv           0.02 %          0.02 %             5
```

KungFuJesus:
To perhaps no surprise, here are those spicy unaligned moves.
unaligned_movs

With alignment:
aligned_2

inftrees.c
```diff
@@ -15,6 +21,86 @@ const char PREFIX(inflate_copyright)[] = " inflate 1.3.1 Copyright 1995-2024 Mar
...
+    }
+
+    vst1q_u16(&count[0], vaddl_u8(vget_low_u8(s1), vget_low_u8(s2)));
+    vst1q_u16(&count[8], vaddl_u8(vget_high_u8(s1), vget_high_u8(s2)));
+
+#elif defined(__SSE2__)
+    __m128i s1 = _mm_setzero_si128();
+    __m128i s2 = _mm_setzero_si128();
+
+    const uint8_t *p = (uint8_t*)one;
```
mtl1979: This looks wrong...  Shouldn't need cast on the right side as the types are same.

mtl1979 (changes requested):
Casting away const-ness and then storing back to const variable is dangerous.

inftrees.c
```diff
@@ -15,6 +21,86 @@ const char PREFIX(inflate_copyright)[] = " inflate 1.3.1 Copyright 1995-2024 Mar
...
+    if (codes & 1) {
+        s1 = _mm_load_si128((const __m128i*)&p[16 * lens[0]]);
+    }
+    for (sym = codes & 1; sym < codes; sym += 2) {
+        s1 = _mm_add_epi8(s1, _mm_load_si128((const __m128i*)&p[16 * lens[sym] - 1]);  // vaddq_u8
```
mtl1979: Missing `)`?

dougallj:
> Casting away const-ness and then storing back to const variable is dangerous.

How so?

KungFuJesus:
We aren't storing to what we are casting const. But either way we're probably upsetting some linter, regardless.

We can probably remove the p assignment entirely and just use one directly, I was just trying to reuse the shorthand but now that we aren't computing an offset into it, it hardly matters.

inftrees.c
```diff
@@ -15,6 +21,86 @@ const char PREFIX(inflate_copyright)[] = " inflate 1.3.1 Copyright 1995-2024 Mar
...
+        s2 = _mm_add_epi8(s2, _mm_load_si128((const __m128i*)&p[16 * lens[sym+1] - 1]));
```
dougallj: Hmm, I think the two `- 1`s go out of bounds and should be deleted?

KungFuJesus: Ah whoops, sorry that was a remnant when I was (wrongly) trying to disregard the first bin.  The first bin does occur.  Here's a revised patch.
[0001-Force-purely-aligned-loads.patch](https://github.com/user-attachments/files/24147280/0001-Force-purely-aligned-loads.patch)

mtl1979:
> > Casting away const-ness and then storing back to const variable is dangerous.
> 
> 
> 
> How so?

1. The whole point of keeping variables const is that optimizer knows memory is never modified, even outside the current function

2. Some compilers ignore casts with no effect, others issue a warning that can be treated as an error. It interferes with static checking.

inftrees.c
nmoinvaz: Patch applied.

Dead2:
### Develop
```
 155156    1352       8  156516   26364 libz-ng.so.2

------------------------------------------------------------------------------------
Benchmark                                          Time             CPU   Iterations
------------------------------------------------------------------------------------
inflate/inflate_nocrc/1                         13.7 ns         13.7 ns    204194534
inflate/inflate_nocrc/64                         111 ns          111 ns     25132817
inflate/inflate_nocrc/1024                       261 ns          260 ns     10569587
inflate/inflate_nocrc/16384                     3732 ns         3727 ns       733649
inflate/inflate_nocrc/131072                   14401 ns        14381 ns       195287
inflate/inflate_nocrc/1048576                  96520 ns        96343 ns        28995

inflate/inflate_nocrc/1                         13.7 ns         13.7 ns    204164512
inflate/inflate_nocrc/64                         111 ns          111 ns     25236921
inflate/inflate_nocrc/1024                       256 ns          255 ns     10652691
inflate/inflate_nocrc/16384                     3725 ns         3720 ns       727272
inflate/inflate_nocrc/131072                   14539 ns        14519 ns       195265
inflate/inflate_nocrc/1048576                  96710 ns        96537 ns        29012
```

### PR
```
 155604    1352       8  156964   26524 libz-ng.so.2

inflate/inflate_nocrc/1             13.7 ns         13.7 ns    204206895
inflate/inflate_nocrc/64             111 ns          111 ns     25343109
inflate/inflate_nocrc/1024           269 ns          269 ns      9991159
inflate/inflate_nocrc/16384         3589 ns         3584 ns       790033
inflate/inflate_nocrc/131072       14302 ns        14283 ns       198693
inflate/inflate_nocrc/1048576      95759 ns        95586 ns        29240

inflate/inflate_nocrc/1             13.7 ns         13.7 ns    204260586
inflate/inflate_nocrc/64             111 ns          111 ns     25239428
inflate/inflate_nocrc/1024           269 ns          268 ns     10075528
inflate/inflate_nocrc/16384         3602 ns         3598 ns       783068
inflate/inflate_nocrc/131072       14260 ns        14240 ns       197131
inflate/inflate_nocrc/1048576      95948 ns        95775 ns        29255
```

Bigger code.
Tests show 1k is slightly slower, but 16k, 128k and 1M are faster.
The 1k is well within variance. The rest are also close, but I'd say this shows the PR is faster.

Dead2 (approved):

PR #2037, opened 2025-12-03
Change code table access from pointer to value in inflate_fast

nmoinvaz:
Benchmark data can be found here under 1000 patch.
https://github.com/zlib-ng/zlib-ng/issues/1998#issuecomment-3598125242
Further benchmark info can be found here:
https://github.com/zlib-ng/zlib-ng/issues/1998#issuecomment-3599244280

@dougallj please review.

PR #2047, opened 2025-12-08
Reorder code struct fields for better access patterns

nmoinvaz:
Place bits field before op field in code struct to optimize memory access. The bits field is accessed first in the hot path, so placing it at offset 0 may improve code generation on some architectures.

Here are my benchmarks from @dougallj original patch set:
https://gist.github.com/nmoinvaz/e393ce50c72cbbe3eb4dc33b46eae02b

Since I didn't write this patch, I asked AI to summarize the potential benefits.

image


Dead2:
Not seeing any reliably measurable difference on 8700BE or RPI5 with either gcc or clang.

Dead2 (approved):

dougallj:
I guess to clarify a bit, this is important if (and probably only if) the `code` is loaded as a single register (as we're attempting to ensure in #2037).

The purpose is to remove a single "shift-right-by-8" instruction on all major architectures (at least 32/64-bit x86, arm and risc-v). These architectures all have shift instructions that read the shift distance from the least-significant bits, and ignore the data in the higher bits for free. You can see this here: https://godbolt.org/z/x8oeKKxbE

The shift that we are removing is on the critical path from decoding one code to the next, so it should save a cycle of latency, although #2037 and possibly other changes are needed for this to pay off. There is also no significant penalty for accessing the `op` field in its new location – it never needs to be shifted into the low bits in the hot loop.

Since it sounds like this is performance-neutral in the absence of #2037, landing this first might improve/fix the benchmark results for that PR.

PR #2053, opened 2025-12-21
Improve Loongarch64 toolchain and switch to new intrinsics

phprus:
1) Toolchain cmake script allow set compiler name suffix (version suffix).
2) GCC16 and LLVM22 introduce new LASX and LSX conversion intrinsics. Switching to using these functions and simplifying assembler blocks for older compilers.

GCC commit: https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=c2013267642fea4a6e89b826940c8aa80a76089d
LLVM commit: https://github.com/llvm/llvm-project/commit/9be30e50c2bf878bd15ac8ed1270f1714c32b30f


mtl1979:
When CI runners start to support gcc-16 and clang-22, we should add CI workflows to test that everything is (still) working... Cleaning up existing code is always a good thing as long as it doesn't break supporting older compilers, which we currently are not testing on CI, but after this PR, we could theoretically, just not sure what is the oldest usable compiler version for either gcc or clang.

phprus:
GCC-16 and Clang-22 will be released in 2026.
We will be able to add CI for GCC-16 ( https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test ) in Ubuntu 26.04.

CI currently uses GCC-14.
GCC-14 is the minimum version with full support LSX/LASX intrinsics.

mtl1979:
> CI currently uses GCC-14. GCC-14 is the minimum version with full support LSX/LASX intrinsics.

Only reason for testing with GCC-13 (or older) would be checking that LSX/LASX optimized code paths get disabled correctly. Personally I think testing single architecture with at most 3 different compiler versions is enough... As a reviewer I consider this PR as first part of multi-part patch set, with adding support for newer compilers as subsequent patches in same series, thus adding CI workflow for gcc-15 is not required in this PR itself if we add support for gcc-16 when it becomes available.

phprus:
> Only reason for testing with GCC-13 (or older) would be checking that LSX/LASX optimized code paths get disabled correctly.

GCC-13 is the first mainline version with support LoongArch64.
I don't think it's necessary to test gcc versions earlier than GCC-14.
If such tests are needed in the future, they can easily be added after this PR.

Yes, this is planned to be the first patch in a series.
I'm currently exploring code optimization methods, including using GCC-16 (master).
Unfortunately, GCC-16 (master) is only in the Ubuntu 26.04 repositories, but it's not released yet, so I can't easily add it to Github Actions.

Dead2 (approved):
LGTM

phprus:
Rebased

Issue #1926, opened 2025-06-23
Warning "bitbuffer may be used uninitialized" in crc32_chorba_118960_nondestructive function (GCC 14.2.0)


phprus:
GCC 14.2.0 produces warnings in the ``crc32_chorba_118960_nondestructive`` function:

```
[ 26%] Building C object CMakeFiles/zlib.dir/arch/generic/crc32_chorba_c.c.o
/home/runner/work/zlib-ng/zlib-ng/arch/generic/crc32_chorba_c.c: In function ‘crc32_chorba_118960_nondestructive’:
/home/runner/work/zlib-ng/zlib-ng/arch/generic/crc32_chorba_c.c:419:66: warning: ‘bitbuffer’ may be used uninitialized [-Wmaybe-uninitialized]
  419 |         in1 = inputqwords[i / sizeof(uint64_t)] ^ bitbufferqwords[(i / sizeof(uint64_t)) % bitbuffersizeqwords];
      |                                                   ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/zlib-ng/zlib-ng/arch/generic/crc32_chorba_c.c:34:27: note: ‘bitbuffer’ declared here
   34 |     ALIGNED_(16) z_word_t bitbuffer[bitbuffersizezwords];
      |                           ^~~~~~~~~
/home/runner/work/zlib-ng/zlib-ng/arch/generic/crc32_chorba_c.c:420:70: warning: ‘bitbuffer’ may be used uninitialized [-Wmaybe-uninitialized]
  420 |         in2 = inputqwords[i / sizeof(uint64_t) + 1] ^ bitbufferqwords[(i / sizeof(uint64_t) + 1) % bitbuffersizeqwords];
      |                                                       ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/runner/work/zlib-ng/zlib-ng/arch/generic/crc32_chorba_c.c:34:27: note: ‘bitbuffer’ declared here
   34 |     ALIGNED_(16) z_word_t bitbuffer[bitbuffersizezwords];
      |                           ^~~~~~~~~
[...]

PR #2055, opened 2025-12-23
Suppress warning "bitbuffer may be used uninitialized"

phprus:
Workaround for issue #1926


mtl1979:
We should only suppress the warning after it's confirmed to be false positive... With uninitialised data, simple assert is enough to confirm uninitialised data while running test suite.

PR #1659, opened 2024-02-03
Disable dynamic function dispatching for native or arch-specific builds

phprus:
Another implementation of PR #1645.

Main differences:
- Separate CPU-specific function prototypes and CPU-arch features checker.
- Reducing dependencies between header files.

PR #2057, opened 2025-12-25
LoongArch64 and e2k has 8-byte general-purpose registers.

phprus:
LoongArch64 and e2k has 8-byte general-purpose registers.

### DIFF
--- a/crc32_braid_p.h
+++ b/crc32_braid_p.h
@@ -9,7 +9,8 @@
 /* Define BRAID_W and the associated z_word_t type. If BRAID_W is not defined, then a braided
    calculation is not used, and the associated tables and code are not compiled.
  */
-#if defined(__x86_64__) || defined(_M_AMD64) || defined(__aarch64__) || defined(_M_ARM64) || defined(__powerpc64__)
+#if defined(__x86_64__) || defined(_M_AMD64) || defined(__aarch64__) || defined(_M_ARM64) || defined(__powerpc64__) \
+    || (defined(__loongarch__) && __loongarch_grlen == 64) || defined(__e2k__)
 #  define BRAID_W 8
     typedef uint64_t z_word_t;
 #else


mtl1979:
We don't even have basic support for Elbrus (Эльбрус) processors, so this would be essentially initial support for yet another architecture. Always with all new architectures, it will be very hard to review and verify that any additions are correct.

phprus:
@mtl1979

We do not have any declared support for E2K architecture. But after merged PR #1659 we actually have it.

This is not a significant change for E2K, as E2K builds reuse x86 optimizations in native build mode (without dynamic dispatch).
So I added a fix for both 64-bit platforms (LA64 and E2K) in one PR.

mtl1979 (approved):
LGTM

mtl1979:
> This is not a significant change for E2K, as E2K builds reuse x86 optimizations in native build mode (without dynamic dispatch). So I added a fix for both 64-bit platforms (LA64 and E2K) in one PR.

I know Elbrus 2000 can run basic x86 code with binary translation, but I was referring to when running code compiled for Elbrus architecture itself (type of VLIW, very long instruction word) using LCC as it could possibly allow some own optimisations later on...

phprus:
> but I was referring to when running code compiled for Elbrus architecture itself (type of VLIW, very long instruction word) using LCC as it could possibly allow some own optimisations later on...

LCC compiler can use x86 intrinsics to create native code for e2k. This allows reuse of some x86-optimized code.
We can add the required settings to CMake for native building on e2k, but without an open emulator we won't be able to add this build to CI.

mtl1979:
> LCC compiler can use x86 intrinsics to create native code for e2k. This allows reuse of some x86-optimized code. We can add the required settings to CMake for native building on e2k, but without an open emulator we won't be able to add this build to CI.

We don't have most tier 3 architectures or operating systems on CI even though some could be added using third-party CI providers, so not being able to add them to GitHub Actions isn't a show stopper.

* Tier 1 is pretty much every operating system or architecture GitHub Actions supports without qemu. (Windows 10/11 using MSVC, Linux, MacOS; x86, AMD64 and AArch64)
* Tier 2 is what we can test on GitHub actions using qemu or other emulators. (including PowerPC (G3 or G4), Power9, Power10, SPARC, MIPS; MinGW, Cygwin, MSYS2)
* Tier 3 is everything that crash qemu or qemu doesn't have support for. (including FreeBSD, OpenBSD, NetBSD; all unlisted operating systems and achitectures)

Dead2 (approved):
LGTM

PR #2056, opened 2025-12-24
Small optimization in 256 bit wide chunkset

KungFuJesus:
It turns out Intel only parses the bottom 4 bits of the shuffle vector. This makes it already a sufficient permutation vector and saves us a small bit of latency.

PR #2058, opened 2025-12-27
Remove redundant instructions in 256 bit wide chunkset on LoongArch64

phprus:
PR #2056 for LoongArch64.

My implementation of `lasx_shuffle_b` only used 4 bits of the shuffle vector.
And I didn't notice that the code being removed in this PR doesn't do anything.


Dead2 (approved):
LGTM

PR #2065, opened 2025-12-29
Improves compile-time and Windows checks for arm

Andarwinux:
1. Do compile-time checks first, not last

2. Add SHA3 runtime check support for Windows
ref: https://github.com/mingw-w64/mingw-w64/blob/8d02d610f707b5f6af74653c6ebb0cdfa4df9212/mingw-w64-tools/widl/include/winnt.h#L1120

### DIFF
--- a/arch/arm/arm_features.c
+++ b/arch/arm/arm_features.c
@@ -23,10 +23,18 @@
 #  include <sys/sysctl.h>
 #elif defined(_WIN32)
 #  include <windows.h>
+#  ifndef PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE
+#    define PF_ARM_V8_CRYPTO_INSTRUCTIONS_AVAILABLE 30
+#  endif
+#  ifndef PF_ARM_SHA3_INSTRUCTIONS_AVAILABLE
+#    define PF_ARM_SHA3_INSTRUCTIONS_AVAILABLE 64
+#  endif
 #endif
 
 static int arm_has_crc32(void) {
-#if defined(ARM_AUXV_HAS_CRC32)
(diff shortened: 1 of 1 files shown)


nmoinvaz (approved):
Looks good, thank you.

mtl1979:
Primary purpose of doing run-time checks first is to avoid crashes when the library is run on older processor than the one used for compiling. Even with 64-bit ARM processors there is differences in support for certain features.

nmoinvaz:
I believe this is for native instructions support.

phprus:
> I believe this is for native instructions support.

Native build does not use any runtime functions:

https://github.com/zlib-ng/zlib-ng/blob/e2ecd2665b8ddf5eadae6925c0f0c2f2da86fed3/arch/arm/arm_functions.h#L70-L76

I think there shouldn't be any compile-time checking in these functions.

Dead2:
Do we know what compilers do set these defines, and what compilers do not? We have generally preferred to do this kind of detection in CMake where it can better report back what is/is not detected.

mtl1979:
> I think there shouldn't be any compile-time checking in these functions.

We allow "forcing" certain features on if the compiler doesn't support checking the feature at run-time... But this logic is always off by default as normal user of zlib-ng would never read the documentation to see that feature checks are short-circuited by default and wonder why all applications using the library crash. Run-time feature detection is skipped only for native builds as it allows better optimisations including less indirect jumps, reduces size of the library, and it's already assumed the library is only used on same exact processor.

Andarwinux:
The reality typically lies somewhere in between: runtime checks below the compile-time baseline are unnecessary, but those above the baseline still hold value.

If users choose to build zlib-ng with some flags, it means they are fully aware of the consequences of using those flags.

PR #2072, opened 2026-01-04
Add ALIGN_DIFF to perform alignment needed to next boundary

nmoinvaz:
This simplies some of the align calculations and checks.. I've also use constants in some of the crc32 code. It just makes the code easier to read, doesn't rely on a typedef, and there are so many other areas of code that use constants in these situations of `len` checks it makes more sense.

### DIFF
--- a/arch/arm/crc32_armv8.c
+++ b/arch/arm/crc32_armv8.c
@@ -10,12 +10,7 @@
 #include "crc32.h"
 
 Z_INTERNAL Z_TARGET_CRC uint32_t crc32_armv8(uint32_t crc, const uint8_t *buf, size_t len) {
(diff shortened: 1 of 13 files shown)


mtl1979:
ARM version dumps core, can't see obvious reason, but the MIN() usage seems suspicious. Using len as misalignment might cause unaligned access.

Dead2:
### Develop
```
   text	   data	    bss	    dec	    hex	filename
 151836	   1328	      8	 153172	  25654	libz-ng.so.2
```

### PR
```
   text	   data	    bss	    dec	    hex	filename
 151772	   1328	      8	 153108	  25614	libz-ng.so.2
```

Dead2 (approved):
LGTM

PR #2073, opened 2026-01-05
[CI] Add workflow with no AVX512VNNI

mtl1979:
* This adds coverage with optimizations that have versions for both AVX512 and AVX512VNNI


nmoinvaz (approved):

mtl1979:
Currently there is no option in GitHub Actions to select minimum sub-architecture for the runner's processor, so coverage shows 0% for both AVX512 and AVX512VNNI specific files when the runner hardware doesn't support them... However there is demand and feature request to add such feature.

See also:

1. https://github.com/orgs/community/discussions/183550

PR #2076, opened 2026-01-06
Explicitly define the __SSE__ and __SSE2__ macros, since starting with MSVS 2012 the default instruction set is SSE2

phprus:
Workaround for MSVC. By default MSVC does not define the `__SSE*__` macros.
We explicitly define the `__SSE__` and `__SSE2__` macros, since starting with MSVS 2012 the default instruction set is SSE2.

MSVS 2015: https://learn.microsoft.com/en-us/cpp/build/reference/arch-x86?view=msvc-140
MSVS 2013: https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2013/7t5yh4fd(v=vs.120)
MSVS 2012: https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2012/7t5yh4fd(v=vs.110)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,6 +240,16 @@ elseif(MSVC)
             set(NEONFLAG "/arch:VFPv4")
         endif()
     endif()
+
+    # Workaround for MSVC. By default MSVC does not define the __SSE*__ macros.
+    # We explicitly define the __SSE__ and __SSE2__ macros, since starting with MSVS 2012
+    # the default instruction set is SSE2.
+    if(BASEARCH_X86_FOUND)
+        add_definitions(
+            -D__SSE__=1
+            -D__SSE2__=1
+        )
+    endif()
 elseif(CMAKE_C_COMPILER_ID MATCHES "GNU" OR CMAKE_C_COMPILER_ID MATCHES "Clang" OR CMAKE_C_COMPILER_ID MATCHES "NVHPC")
     # Enable warnings in GCC and Clang
     set(WARNFLAGS -Wall)
@@ -361,8 +371,6 @@ if(NOT WITH_RUNTIME_CPU_DETECTION)
         set(CMAKE_REQUIRED_FLAGS)
         if(MSVC_IS_ENABLED_AVX)
             add_definitions(
-                -D__SSE__=1
-                -D__SSE2__=1
                 -D__SSE3__=1
                 -D__SSSE3__=1
                 -D__SSE4_1__=1


inftrees.c
```diff
@@ -8,7 +8,7 @@
 #include "inftrees.h"
 #include "fallback_builtins.h"
 
-#if defined(__SSE2__)
+#if defined(__SSE2__) || defined(_M_X64) || (defined(_M_IX86_FP) && _M_IX86_FP >= 2)
```
mtl1979: This breaks if SSE2 is disabled by build configuration on 32-bit Windows, should use `X86_SSE2` instead.

inftrees.c
```diff
@@ -59,7 +59,7 @@ static inline void count_lengths(uint16_t *lens, int codes, uint16_t *count) {
     vst1q_u16(&count[0], vaddl_u8(vget_low_u8(s1), vget_low_u8(s2)));
     vst1q_u16(&count[8], vaddl_u8(vget_high_u8(s1), vget_high_u8(s2)));
 
-#elif defined(__SSE2__)
+#elif defined(__SSE2__) || defined(_M_X64) || (defined(_M_IX86_FP) && _M_IX86_FP >= 2)
```
mtl1979: This breaks if SSE2 is disabled by build configuration on 32-bit Windows, should use `X86_SSE2` instead.

mtl1979 (changes requested):

inftrees.c
phprus: I don't understand why?
On x86-64, SSE2 cannot be disabled (condition: `defined(_M_X64)`).
On x86 (32-bit) predefined macro `_M_IX86_FP` is defined with value `2` only if the `/arch:SSE2`, `/arch:AVX*` compiler option was set.

A 32-bit Windows build with SSE2 disabled by build configuration does not use the `/arch:SSE2` compiler option, and the `_M_IX86_FP` macro value cannot be great or equal to `2`.

The `X86_SSE2` and `ARM_NEON` macros are only needed if we want to explicitly disable vectorized code. Fix it in commit "Allow explicitly disable vectorized code by build options".

mtl1979: We don't use `/arch:SSE2` outside `arch/x86` so autovectorization doesn't try to vectorize code outside the "safe" blocks. This still allows using intrinsics in code paths where vector optimizations are permitted.

phprus: You're absolutely right! I misremembered the CMakeLists.txt logic.

In this case, using the `X86_SSE2` and `ARM_NEON` macros would only be useful for disabling manual vectorization in `inftrees.c`. I added this in commit https://github.com/zlib-ng/zlib-ng/pull/2076/commits/11c57c5c30c0f1bdbb98375c26b451c024adfe20

inftrees.c
```diff
@@ -8,9 +8,9 @@
 #include "inftrees.h"
 #include "fallback_builtins.h"
 
-#if defined(__SSE2__)
+#if defined(X86_SSE2) && (defined(__SSE2__) || defined(_M_X64) || (defined(_M_IX86_FP) && _M_IX86_FP >= 2))
```
mtl1979: We only need `defined(_M_IX86)` for the last check as `X86_SSE2` already implies that SSE2 is supported.

inftrees.c
```diff
@@ -59,7 +59,7 @@ static inline void count_lengths(uint16_t *lens, int codes, uint16_t *count) {
     vst1q_u16(&count[0], vaddl_u8(vget_low_u8(s1), vget_low_u8(s2)));
     vst1q_u16(&count[8], vaddl_u8(vget_high_u8(s1), vget_high_u8(s2)));
 
-#elif defined(__SSE2__)
+#elif defined(X86_SSE2) && (defined(__SSE2__) || defined(_M_X64) || (defined(_M_IX86_FP) && _M_IX86_FP >= 2))
```
mtl1979: We only need defined(_M_IX86) for the last check as X86_SSE2 already implies that SSE2 is supported.

mtl1979 (changes requested):
Check `defined(_M_IX86_FP) && _M_IX86_FP >= 2)` will fail on non-native builds. Native builds are unsupported for MSVC, so most builds for 32-bit Windows will fail.

inftrees.c
phprus: The `X86_SSE2` macro indicates that the compiler/build supports SSE2.
Checking `X86_SSE2` and `defined(_M_IX86)` is incorrect, as such a build will not work on processors without SSE2 (if there are any left and zlib-ng currently runs on them).


```c++
defined(X86_SSE2) // Check support sse2 by compiler/build
&&                // AND
(                 // Compiler specific conditions
    defined(__SSE2__) // SSE2 is enabled for this file (gcc-like predefined macros)
    ||                // OR
    defined(_M_X64)   // MSVC x86_64 build (on x86_64 SSE2 always present)
    ||                // OR
    (defined(_M_IX86_FP) && _M_IX86_FP >= 2) // MSVC 32-bit build. Is SSE2 is enabled for this file then `_M_IX86_FP` has value `2` or greater. If not enabled - less `2`.
)
```

If `X86_SSE2` is not defined, SSE2 code will not compile.
If the compiler does not declare SSE2 support for this file, SSE2 code will not compile.

What combinations of conditions have I missed?

phprus:
> Check `defined(_M_IX86_FP) && _M_IX86_FP >= 2)` will fail on non-native builds. Native builds are unsupported for MSVC, so most builds for 32-bit Windows will fail.

What fail are you talking about?
32-bit Windows compilation without errors: https://github.com/zlib-ng/zlib-ng/actions/runs/20749244086/job/59574046268?pr=2076

inftrees.c
phprus: Or do we want this SSE2 code to be enabled always in 32-bit windows?

mtl1979: As I already said, intrinsic support in MSVC doesn't require "/arch:sse2", so compiler doesn't declare support either... It's true that those builds don't run on processors without SSE2, but that's intentional. For intrinsics, specifying "/arch" is only required for VNNI as those instructions can use either AVX2 or AVX512 encodings, but none of the current MSVC versions support the AVX2 encodings and this file doesn't use those instructions.

phprus: Should the SSE2 code in `inftrees.c` always compile when using MSVC?

nmoinvaz:
@phprus have you seen this in CMake?
https://github.com/zlib-ng/zlib-ng/blob/6bb27e3726010be652d33497995abdb915460db5/CMakeLists.txt#L362

phprus:
> @phprus have you seen this in CMake?
> 
> https://github.com/zlib-ng/zlib-ng/blob/6bb27e3726010be652d33497995abdb915460db5/CMakeLists.txt#L362

This cmake `if` and `add_definitions` for Native build only.

This PR enable intrinsic-optimized code for MSVC x86_64 and for MSVC x86 (if SSE2 is explicitly enabled by compiler options).

inftrees.c
mtl1979: `CMakeLists.txt` already defines `__SSE2__` for MSVC when it's needed by some parts of the build script, so we can assume that SSE2 code should always be compiled in when SSE2 is supported.

mtl1979:
> @phprus have you seen this in CMake?
> 
> https://github.com/zlib-ng/zlib-ng/blob/6bb27e3726010be652d33497995abdb915460db5/CMakeLists.txt#L362

IMHO, this is something that should be moved to `zbuild.h` as it's not limited to run-time feature detection.

mtl1979:
> > Check `defined(_M_IX86_FP) && _M_IX86_FP >= 2)` will fail on non-native builds. Native builds are unsupported for MSVC, so most builds for 32-bit Windows will fail.
> 
> What fail are you talking about? 32-bit Windows compilation without errors: https://github.com/zlib-ng/zlib-ng/actions/runs/20749244086/job/59574046268?pr=2076

Preprocessor test always returning false doesn't always equal to compilation failing with error.

inftrees.c
phprus: https://github.com/zlib-ng/zlib-ng/blob/6bb27e3726010be652d33497995abdb915460db5/CMakeLists.txt#L347-L375

CMakeLists.txt defines `__SSE2__` for MSVC only to support native builds when using `/arch:AVX` and higher. Otherwise, there is no such definition and it will be incorrect.

If we add the `__SSE2__` macro definition for MSVC in any case, zlib-ng will not work on 32-bit processors without SSE2.

phprus:
> IMHO, this is something that should be moved to `zbuild.h` as it's not limited to run-time feature detection.

This is workaround to MSVC native build with `/arch:AVX` and higher.
Since this is a workaround for MSVC native builds, macros must be defined for all source files. Moving the definition to zbuild.h may break the MSVC native build if this file isn't included first.

phprus:
> > > Check `defined(_M_IX86_FP) && _M_IX86_FP >= 2)` will fail on non-native builds. Native builds are unsupported for MSVC, so most builds for 32-bit Windows will fail.
> > 
> > 
> > What fail are you talking about? 32-bit Windows compilation without errors: https://github.com/zlib-ng/zlib-ng/actions/runs/20749244086/job/59574046268?pr=2076
> 
> Preprocessor test always returning false doesn't always equal to compilation failing with error.

MSVC allow overriding `/arch`: https://godbolt.org/z/j39K5j8h5 The last value will be used.

For a 32-bit build, we can set:
`-DCMAKE_C_FLAGS="/arch:SSE2"`
In this case, the condition will return true and the SSE2 code will be compiled. And dynamic dispatch will work fine.

Test for this case: https://github.com/phprus/zlib-ng/actions/runs/20818675008/job/59800949459

inftrees.c
mtl1979: The logic in `CMakeLists.txt` is incomplete anyways and needs to be fixed... As I already asserted, for maintainability concern, it would be easier to move it to `zbuild.h` where it's easier to be found. This way we can "fix" it easier to work also on processors that only support SSE or SSE2 and not AVX. Repeating same fix for 32-bit Windows builds clutters the files anyways...

mtl1979:
> MSVC allow overriding `/arch`: https://godbolt.org/z/j39K5j8h5 The last value will be used.
> 
> For a 32-bit build, we can set: `-DCMAKE_C_FLAGS="/arch:SSE2"` In this case, the condition will return true and the SSE2 code will be compiled. And dynamic dispatch will work fine.
> 
> Test for this case: https://github.com/phprus/zlib-ng/actions/runs/20818675008/job/59800949459

Recent MSVC versions (at least 2022 and early builds of 2026) have bug with `/arch` that makes it enable instructions that should not be enabled... This is one of the reasons why we avoid using it outside `arch/x86`. However for 32-bit builds it might be the only working solution.

MSVC 2019 is last to officially support 32-bit builds and it's already past end of support. I've used clang-cl for building 32-bit binaries on 2022 and 2026, but it's also possible to just install older compiler binary from 2015, 2017.9 or 2019 toolchain.

inftrees.c
phprus: Reimplemented.
I added explicitly define the `__SSE__` and `__SSE2__` macros for MSVC, since starting with MSVS 2015 the default instruction set is SSE2.

The minimum supported MSVC version for zlib-ng is 2022 ( https://github.com/zlib-ng/zlib-ng/wiki ).

mtl1979: The minimum supported MSVC version just means the oldest version that can enable most optimizations, except obviously AVX2VNNI, which is recent addition... We can still use older versions but some optimizations are not available due to compiler bugs that are likely not fixed ever. We actually test all the way back to MSVC 2015 toolchain (v140).

CMakeLists.txt
mtl1979, line 225: Should bump the requirement to:

1900 = Visual Studio 14.0/2015

mtl1979 (commented):
If we continue supporting VS2013, we should still conditionally enable SSE and SSE2 optimizations.

CMakeLists.txt
```diff
@@ -240,6 +240,16 @@ elseif(MSVC)
             set(NEONFLAG "/arch:VFPv4")
         endif()
     endif()
+
+    # Workaround for MSVC. By default MSVC does not define the __SSE*__ macros.
+    # We explicitly define the __SSE__ and __SSE2__ macros, since starting with MSVS 2015
+    # the default instruction set is SSE2.
+    if(BASEARCH_X86_FOUND)
```
nmoinvaz, line 247: Should this be `&& ARCH_64BIT`?

mtl1979: No... Needs definitely to work on 32-bit Windows.

phprus:
> If we continue supporting VS2013, we should still conditionally enable SSE and SSE2 optimizations.

On Visual Studio 2012, 2013 the default instruction on x86 platforms is SSE2:
https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2013/7t5yh4fd(v=vs.120)
https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2012/7t5yh4fd(v=vs.110)

CMakeLists.txt
phprus: This is not necessary, since MSVS 2012,2013 also defaults to SSE2.

mtl1979 (approved):
LGTM.

CMakeLists.txt
phprus: MSVS 2012+ uses the SSE2 instruction set by default on the x86.

I don't know if it makes sense to support `/arch:IA32` and `/arch:SSE`.
CI doesn't implement such builds.
The release process also doesn't support them.

Dead2 (approved):
LGTM

Dead2:
@phprus Would you please rebase this?

phprus:
@Dead2 
Rebased

PR #2077, opened 2026-01-07
Add adler32_copy unit test and fix inconsistent short-len handling

nmoinvaz:


Dead2:
@nmoinvaz I think you either found a bug or made one :raised_eyebrow:

test/hash_test_strings_p.h
```diff
@@ -0,0 +1,385 @@
...
+static uint8_t* hash_setup_buf() {
+    for (int i = 0; i < 615336; ++i) {
+        uint8_t ic = (uint8_t)(i % 256);
+        hash_fullwin_buf[i] = ic;
+    }
+
+    return hash_fullwin_buf;
+}
+
+static uint8_t *buf32k = hash_setup_buf();
```
Dead2: Since this buffer is no longer 32k, perhaps change the name to bigtest or something.

test/hash_test_strings_p.h
```diff
@@ -0,0 +1,385 @@
...
+    'y','K','B','A','5','2','7','b','y','R','K','q','A','u','3','J'};
+
+typedef struct {
+    uint32_t crc;
+    uint32_t adler;
+    const uint8_t *buf;
+    uint32_t len;
+    uint32_t expect_crc;
+    uint32_t expect_adler;
+} hash_test;
```
Dead2, line 202: Nitpick: I would have preferred the order be `len, buf, initial_crc, expect_crc, initial_adler, expect_adler`. The checksums are less interesting when scanning the list, length and the buffer content is a bit more interesting. I'd also prefer the `crc` and `adler` names be prefixed with `initial_` to avoid confusion with the `expect_` names when reading the tests (if you only see the input is named `crc`, do you not then think that is the expected crc, until you notice that the other one is named expect_crc?)

I won't refuse over these comments though, I know reordering fields is kind of a pain.

nmoinvaz: I can reorder it.

nmoinvaz: Done. Although I put adler before crc for alphabetical.

test/hash_test_strings_p.h
nmoinvaz: Done.

nmoinvaz:
> @nmoinvaz I think you either found a bug or made one 🤨

Looks like we had some inconsistent handling of src/buf = NULL or len == 0 across the implementations.

nmoinvaz:
Now generating code coverage reports failing. Maybe codecov is having a problem or we reached a limit..

phprus:
New gcovr version 8.5.

Version 8.4 works without errors.

In version 8.5:

```
gcovr.exceptions.GcovrMergeAssertionError: /home/runner/work/zlib-ng/zlib-ng/chunkset_tpl.h:27 Got function <unknown function> on multiple lines: 27, 54.
    You can run gcovr with --merge-mode-functions=MERGE_MODE.
    The available values for MERGE_MODE are described in the documentation.
```



Full log ( https://github.com/zlib-ng/zlib-ng/actions/runs/20835121384/job/59858889742?pr=2077#step:20:1766 ):

```
(ERROR) - MainThread - Traceback (most recent call last):
  File "/home/runner/work/zlib-ng/zlib-ng/venv/lib/python3.12/site-packages/gcovr/formats/gcov/workers.py", line 96, in worker
    work(*args, **kwargs)
  File "/home/runner/work/zlib-ng/zlib-ng/venv/lib/python3.12/site-packages/gcovr/formats/gcov/read.py", line 531, in process_datafile
    done = run_gcov_and_process_files(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/zlib-ng/zlib-ng/venv/lib/python3.12/site-packages/gcovr/formats/gcov/read.py", line 946, in run_gcov_and_process_files
    process_gcov_text_data(
  File "/home/runner/work/zlib-ng/zlib-ng/venv/lib/python3.12/site-packages/gcovr/formats/gcov/read.py", line 326, in process_gcov_text_data
    covdata.insert_file_coverage(filecov, merge_mode)
  File "/home/runner/work/zlib-ng/zlib-ng/venv/lib/python3.12/site-packages/gcovr/data_model/container.py", line 210, in insert_file_coverage
    self.data[key].merge(filecov, options)
  File "/home/runner/work/zlib-ng/zlib-ng/venv/lib/python3.12/site-packages/gcovr/data_model/coverage.py", line 2353, in merge
    self.__functions.merge(other.__functions, options)
  File "/home/runner/work/zlib-ng/zlib-ng/venv/lib/python3.12/site-packages/gcovr/data_model/coverage_dict.py", line 57, in merge
    self[key].merge(item, options)
  File "/home/runner/work/zlib-ng/zlib-ng/venv/lib/python3.12/site-packages/gcovr/data_model/coverage.py", line 2079, in merge
    self.raise_merge_error(
  File "/home/runner/work/zlib-ng/zlib-ng/venv/lib/python3.12/site-packages/gcovr/data_model/coverage.py", line 148, in raise_merge_error
    raise GcovrMergeAssertionError(
gcovr.exceptions.GcovrMergeAssertionError: /home/runner/work/zlib-ng/zlib-ng/chunkset_tpl.h:27 Got function <unknown function> on multiple lines: 27, 54.
    You can run gcovr with --merge-mode-functions=MERGE_MODE.
    The available values for MERGE_MODE are described in the documentation.
GCOV data file of merge source is:
   {('CMakeFiles/zlib-ng.dir/arch/x86/chunkset_avx512.c.gcda', '/home/runner/work/zlib-ng/zlib-ng/chunkset_tpl.h##d8f3ab1325cc743e6646b742327f5733.gcov')}
and of merge target is:
   {('CMakeFiles/zlib-ng.dir/arch/x86/chunkset_avx2.c.gcda', '/home/runner/work/zlib-ng/zlib-ng/chunkset_tpl.h##d8f3ab1325cc743e6646b742327f5733.gcov')}

```

nmoinvaz:
Rebased.

test/hash_test_strings_p.h
```diff
@@ -0,0 +1,384 @@
+#ifndef HASH_TEST_STRINGS_P_H
+#define HASH_TEST_STRINGS_P_H
+
+ALIGNED_(16) static uint8_t hash_fullwin_buf[615336];
+
+static uint8_t* hash_setup_buf() {
+    for (int i = 0; i < 615336; ++i) {
```
Dead2: I do agree with the coderabbitai nitpick that moving the define above these, and using the define here instead of the hardcoded value would be better.

Dead2 (approved):
Other than the one possible improvement, this looks good.

test/hash_test_strings_p.h
nmoinvaz: Done.

nmoinvaz:
Fixed casting warning on Windows after that last change.

arch/loongarch/adler32_lasx.c
```diff
@@ -35,13 +35,18 @@ extern uint32_t adler32_copy_lsx(uint32_t adler, uint8_t *dst, const uint8_t *sr
...
     adler1 = (adler >> 16) & 0xffff;
     adler0 = adler & 0xffff;
 
+    /* in case user likes doing a byte at a time, keep it fast */
+    if (UNLIKELY(len == 1))
+        return adler32_copy_len_1(adler0, dst, src, adler1, COPY);
+
+    /* initial Adler-32 value (deferred check for len == 1 speed) */
+    if (UNLIKELY(src == NULL))
+        return 1L;
```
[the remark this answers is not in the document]
Dead2: @nmoinvaz It looks like these two checks do need to switch places, don't they?

Dead2:
Seems this fails CI UBSAN. (Also see my previous comment)
```
/home/runner/work/zlib-ng/zlib-ng/arch/x86/adler32_avx512.c:35:57: runtime error: shift exponent 64 is too large for 64-bit type 'long unsigned int'
    #0 0x558a982e7b45 in adler32_copy_impl /home/runner/work/zlib-ng/zlib-ng/arch/x86/adler32_avx512.c:35
    #1 0x558a982e7b45 in adler32_copy_avx512 /home/runner/work/zlib-ng/zlib-ng/arch/x86/adler32_avx512.c:109
    #2 0x558a9841643d in adler32_copy_variant::adler32_copy_test(unsigned int (*)(unsigned int, unsigned char*, unsigned char const*, unsigned long), hash_test) /home/runner/work/zlib-ng/zlib-ng/test/test_adler32_copy.cc:28
    #3 0x558a983fc8df in adler32_copy_variant_avx512_Test::TestBody() /home/runner/work/zlib-ng/zlib-ng/test/test_adler32_copy.cc:74
```

phprus:
@nmoinvaz 

Missed check:
```
if (UNLIKELY(len == 0)) return adler;
```

`len == 0` causes an UB in the `storemask` calculation.

Dead2:
Does the last commit actually belong in this PR though? It seems like it was supposed to go into #2075.
Edit: Well, not really a good fit there either. :man_shrugging:

nmoinvaz:
The last commit fixes a bug that cause copy tests to fail. 

It is only exposed after I merged all test strings which expose inconsistent handling of some of these implementations at lower lens.

nmoinvaz:
image

arch/x86/adler32_avx512.c
```diff
@@ -16,23 +16,26 @@
...
 rem_peel:
     if (len < 64) {
         /* This handles the remaining copies, just call normal adler checksum after this */
         if (COPY) {
-            __mmask64 storemask = (0xFFFFFFFFFFFFFFFFUL >> (64 - len));
-            __m512i copy_vec = _mm512_maskz_loadu_epi8(storemask, src);
-            _mm512_mask_storeu_epi8(dst, storemask, copy_vec);
+            return adler32_copy_avx2(adler, dst, src, len);
+        } else {
+            return adler32_avx2(adler, src, len);
```
phprus: > There is no guarantee that size of dst is greater than size of src by enough bytes to make the final copy or a multiple of 64 bytes.

Why?
`adler32_copy` has only one parameter `len` for `src` and `dst` buffers.

The masked implementation performed the copy in a single read/write.
`_mm512_maskz_loadu_epi8` / `_mm512_mask_storeu_epi8` reads and writes only the masked bytes. There will be no buffer overflow.

UB only if `len==0`. In this case, the 64-bit int will be shifted by 64.

I guess that solution with `_mm512_maskz_loadu_epi8` / `_mm512_mask_storeu_epi8` may be faster than `adler32_copy_avx2` for short strings even with check for `len == 0`.

test/hash_test_strings_p.h
```diff
@@ -0,0 +1,384 @@
...
+  {0, (const uint8_t *)0x0, 0xffffffff, 0x1, 0xffffffff, 0x0},
+  {255, (const uint8_t *)0x0, 0x0, 0x1, 0x0, 0x0}, /*  BZ 174799.  */
+  {256, (const uint8_t *)0x0, 0x0, 0x1, 0x0, 0x0},
+  {257, (const uint8_t *)0x0, 0x0, 0x1, 0x0, 0x0},
+  {32767, (const uint8_t *)0x0, 0x0, 0x1, 0x0, 0x0},
+  {32768, (const uint8_t *)0x0, 0x0, 0x1, 0x0, 0x0},
+  {32769, (const uint8_t *)0x0, 0x0, 0x1, 0x0, 0x0},
+  {0, (const uint8_t *)"", 0x1, 0x1, 0x0, 0x0},
+  {1, (const uint8_t *)"", 0x1, 0x10001, 0x0, 0xd202ef8d},
+  {0, (const uint8_t *)"", 0xffffffff, 0x000e000e, 0xffffffff, 0xffffffff},
```
phprus:
```
 len, buf, initial_adler, expect_adler, initial_crc, expect_crc
{  0, 0x0, 0x0          , 0x1         , 0x0        , 0x0        },
{  0, 0x0, 0xffffffff   , 0x1         , 0xffffffff , 0x0        },
{  0, "" , 0x1          , 0x1         , 0x0        , 0x0        },
{  0, "" , 0xffffffff   , 0x000e000e  , 0xffffffff , 0xffffffff },
```

Why if `len == 0` and `initial_adler == 0xffffffff`, the values of `expect_adler` are different?
According to the algorithm ( https://en.wikipedia.org/wiki/Adler-32 ), `expect_adler` should not depend on the `src` if `len == 0`.

And why if `len == 0` and `initial_crc == 0xffffffff`, the values of `expect_crc` are different?

mtl1979: AFAIK empty string should have `len` of 1 to cause different checksum, as it is essentially single `NUL` byte.

phprus: > AFAIK empty string should have len of 1 to cause different checksum, as it is essentially single NUL byte.

Then the following two tests contradict each other:

```
  {0, (const uint8_t *)"", 0x1, 0x1, 0x0, 0x0},
  {1, (const uint8_t *)"", 0x1, 0x10001, 0x0, 0xd202ef8d},
```

If the first string is implicitly interpreted as a string of length 1, then the hash values ​​must be the same.

phprus: The Adler-32 algorithm on Wikipedia differs from the implementation at https://github.com/madler/zlib/blob/develop/adler32.c

There are no errors in `hash_tests`. The values are the same as `madler/zlib`.

mtl1979: The algorithm in Wikipedia assumes Adler-32 has initial value of 1, which means that it's same as our implementation on first round, but doesn't account chaining calls. I think the confusion raised if modulo should be applied even if `len` is 0 or not. When chaining calls the initial Adler-32 checksum in real life is never 0xFFFFFFFF as the modulo is applied before returning, so it essentially tests if we can handle invalid checksums and resume.

nmoinvaz: > Why if len == 0 and initial_adler == 0xffffffff, the values of expect_adler are different?

The tests are not entirely equal. One has `buf == NULL` and the other `buf == ""`, so the `initial_adler` is irrelevant and the case is probably just copied and merged from `crc32_test_strings_p.h`.

> According to the algorithm ( [en.wikipedia.org/wiki/Adler-32](https://en.wikipedia.org/wiki/Adler-32) ), expect_adler should not depend on the src if len == 0.

That function doesn't have any parameter checking, but in our adler32 functions we do have it: 

```c
    /* initial Adler-32 value (deferred check for len == 1 speed) */
    if (UNLIKELY(src == NULL))
        return 1L;
```

I don't mind removing this check, however the behavior would differ from `adler_z` in `madler/zlib`. None of the `crc32` hash functions check if `src == NULL`. @dead2 thoughts?

> Then the following two tests contradict each other:
>  {0, (const uint8_t *)"", 0x1, 0x1, 0x0, 0x0},
>  {1, (const uint8_t *)"", 0x1, 0x10001, 0x0, 0xd202ef8d},
> If the first string is implicitly interpreted as a string of length 1, then the hash values ​​must be the same.

One has a `len = 0`, and the other `len = 1`, the empty string is irrelevant.

phprus: https://godbolt.org/z/Mcaf9qW6f
The difference is in the handling of zero length.

phprus: @nmoinvaz 
The implementations in zlib-ng and zlib are the same.

arch/x86/adler32_avx512.c
nmoinvaz: > The masked implementation performed the copy in a single read/write.
_mm512_maskz_loadu_epi8 / _mm512_mask_storeu_epi8 reads and writes only the masked bytes. There will be no buffer overflow.

I see. Originally I thought about `if (COPY && len)`..

TBH, I would prefer just to do `adler32_copy_len_64` as it is more straightfoward and similar to other implementations. 

There are probably more overhead for calling a separate function `adler32_avx2/adler32_copy_avx2` on at least < 16, 32?. And in some cases it appears to be calling another function ssse3/sse42, so several jumps... I would think it would only be worth the setup costs if len > 64 for most vector implementations.

test/hash_test_strings_p.h
nmoinvaz: > The implementations in zlib-ng and zlib are the same.

We can consider just moving the check to:
```c
#ifdef ZLIB_COMPAT
unsigned long Z_EXPORT PREFIX(adler32)(unsigned long adler, const unsigned char *buf, unsigned int len) {
    if (UNLIKELY(buf == NULL))
        return 1L;
    return (unsigned long)FUNCTABLE_CALL(adler32)((uint32_t)adler, buf, len);
}
#else
uint32_t Z_EXPORT PREFIX(adler32)(uint32_t adler, const unsigned char *buf, uint32_t len) {
    return FUNCTABLE_CALL(adler32)(adler, buf, len);
}
#endif
```

arch/x86/adler32_avx512.c
nmoinvaz: We should add this mask logic into `adler32_copy_len_64`.

nmoinvaz: Anyways, I decided to do `if COPY && len)` for now.

arch/x86/adler32_avx512.c
```diff
@@ -16,17 +16,22 @@
...
     adler1 = (adler >> 16) & 0xffff;
     adler0 = adler & 0xffff;
 
+    /* in case user likes doing a byte at a time, keep it fast */
+    if (UNLIKELY(len == 1))
+        return adler32_copy_len_1(adler0, dst, src, adler1, COPY);
+
+    /* initial Adler-32 value (deferred check for len == 1 speed) */
+    if (UNLIKELY(src == NULL))
+        return 1L;
```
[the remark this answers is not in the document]
nmoinvaz: This is not for this PR.

arch/loongarch/adler32_lasx.c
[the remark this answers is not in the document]
nmoinvaz: No the code comments suggest otherwise. AFAIk these are original to Madler/zlib just taken from other arch impl.

test/test_crc32_copy.cc
```diff
@@ -9,34 +9,30 @@ extern "C" {
...
             GTEST_SKIP();
         }
 
-        crc = copyfunc(params.crc, dstbuf, params.buf, params.len);
+        uint32_t crc = copyfunc(params.initial_crc, dstbuf, params.buf, params.len);
 
-        EXPECT_EQ(crc, params.expect);
+        EXPECT_EQ(crc, params.expect_crc);
         EXPECT_EQ(0, memcmp(params.buf, dstbuf, params.len));
     }
```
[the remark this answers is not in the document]
nmoinvaz: Not relevant to this PR

arch/loongarch/adler32_lasx.c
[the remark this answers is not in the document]
Dead2: Wouldn't that cause undefined behavior in adler32_copy_len_1 though?
`uint8_t c = *buf;` when buf is NULL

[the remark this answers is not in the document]
nmoinvaz: Correct. It is a pre-existing issue across all adler32 implementations where I copied it from and even from madler/zlib. It is intentional to keep `len = 1` fast.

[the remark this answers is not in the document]
nmoinvaz: I have another comment in this PR here https://github.com/zlib-ng/zlib-ng/pull/2077#discussion_r2679865460. I suggest moving this check into the user API. Possibly can do it in another PR?

test/hash_test_strings_p.h
mtl1979: If we move checks to the wrapper functions, those are only used by external code, not zllb-ng itself, as zlib-ng will not use the wrapper functions due to extra jump on each call.

nmoinvaz: That is correct. It would be the best of both worlds, we don't do the check internally (same as crc32), but the user API still gets that check.

nmoinvaz: You might be asking, how crc32 tests pass if those implementations don't check for NULL.. well the crc32 gtests specifically bypass that scenario.

mtl1979: I've noticed we don't test for a lot of error conditions as coverage for those lines is almost non-existent...

Dead2: I agree, it would be better to check for NULL at the exported function. If we pass a nullpointer internally, that'd have to be a bug, and we'd rather find and fix that rather than keep checking for this in each function.
We should have asserts there though, that helps prevent new bugs and also helps static analyzers understand.

mtl1979: When I comment, it's not always because I'm against something or think something is wrong... Sometimes it's just to clarify I understood the change correctly... Asserts work in debug builds only and we rarely use debug builds. Some compilers allow adding attributes to function parameters to signify that parameter shouldn't be a NULL pointer.

Dead2: > Some compilers allow adding attributes to function parameters to signify that parameter shouldn't be a NULL pointer.

That is something I have been wanting to add to zlib-ng for its external api. It does make zlib-ng.h a bit more messy though. But it would let applications' compilers and static analyzers recognize that our functions don't want NULL. We'd still have to check for NULL though.

mtl1979: > That is something I have been wanting to add to zlib-ng for its external api. It does make zlib-ng.h a bit more messy though. But it would let applications' compilers and static analyzers recognize that our functions don't want NULL. We'd still have to check for NULL though.

Among the compilers we support, Clang and MSVC are the ones that definitely fully support function parameter attributes. Pointer and array checking is part of C11 standard, but there is also compiler specific attributes. I haven't checked the status of function parameter attributes in recent versions of GCC, but I can assume that there is partial support at least.

phprus: GCC supports the `nonnull` attribute: https://gcc.gnu.org/onlinedocs/gcc-8.5.0/gcc/Common-Function-Attributes.html

nmoinvaz: > 
> We can consider just moving the check to:
> 
> ```c
> #ifdef ZLIB_COMPAT
> unsigned long Z_EXPORT PREFIX(adler32)(unsigned long adler, const unsigned char *buf, unsigned int len) {
>     if (UNLIKELY(buf == NULL))
>         return 1L;
>     return (unsigned long)FUNCTABLE_CALL(adler32)((uint32_t)adler, buf, len);
> }
> #endif
> ```

Surprisingly crc32 user-API had the same checks in it..

nmoinvaz: I like the NONNULL idea, it is good improvement idea for the future.

test/hash_test_strings_p.h
```diff
@@ -0,0 +1,378 @@
...
+    const uint8_t *buf;
+    uint32_t initial_adler;
+    uint32_t expect_adler;
+    uint32_t initial_crc;
+    uint32_t expect_crc;
+} hash_test;
+
+static const hash_test hash_tests[] = {
+  {0, (const uint8_t *)0x0, 0x0, 0x0, 0x0, 0x0},
+  {0, (const uint8_t *)0x0, 0xffffffff, 0x000e000e, 0xffffffff, 0xffffffff},
```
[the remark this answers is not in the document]
nmoinvaz: These are two different things.

test/hash_test_strings_p.h
mtl1979: In the past, issue with GCC was that even though it recognized the function parameter attributes, it didn't enforce them. But we can test what compiler versions both support and enforce them, and then conditionally use them.

mtl1979:
Seems like PowerPC64 is failing CRC32 test with...

```
[ RUN      ] crc32/crc32_variant.power8/1
/home/runner/work/zlib-ng/zlib-ng/test/test_crc32.cc:22: Failure
Expected equality of these values:
  crc
    Which is: 0
  param.expect_crc
    Which is: 4294967295

[  FAILED  ] crc32/crc32_variant.power8/1, where GetParam() = 32-byte object <00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00 FF-FF FF-FF 00-0E 00-0E FF-FF FF-FF FF-FF FF-FF> (3 ms)
```

test/hash_test_strings_p.h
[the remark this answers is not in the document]
mtl1979: With NULL buffer, crc32_power8() returns 0, not 0xffffffff. This is inconsistent with other versions that don't have NULL check and only check for `len == 0`.

mtl1979:
Now that NULL buffer and empty buffer are handled the same, we shouldn't try passing NULL buffer in the tests, as the result isn't what the specification says, instead NULL buffer is treated the same as empty buffer. This can cause confusion and the deviating result isn't documented anywhere... Anyone who compares the old CRC32 test matrix and the new combined Adler32/CRC32 test matrix will notice the discrepancy.

nmoinvaz:
Ok, I have removed them. Who knew merging these test strings would be such a big ordeal.. /facepalm

crc32.c
```diff
@@ -19,24 +19,28 @@ const uint32_t * Z_EXPORT PREFIX(get_crc_table)(void) {
...
+    if (buf == NULL)
+        return CRC32_INITIAL_VALUE;
     return (unsigned long)PREFIX(crc32_z)((uint32_t)crc, buf, len);
 }
 #else
 uint32_t Z_EXPORT PREFIX(crc32)(uint32_t crc, const unsigned char *buf, uint32_t len) {
+    if (buf == NULL)
+        return CRC32_INITIAL_VALUE;
     return PREFIX(crc32_z)(crc, buf, len);
 }
```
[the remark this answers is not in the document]
nmoinvaz: Bro you are getting super annoying.

test/test_crc32_copy.cc
```diff
@@ -9,34 +9,26 @@ extern "C" {
...
+        ASSERT_LE(params.len, HASH_TEST_MAX_LENGTH);
 
-        crc = copyfunc(params.crc, dstbuf, params.buf, params.len);
+        uint32_t crc = copyfunc(params.initial_crc, dstbuf, params.buf, params.len);
 
-        EXPECT_EQ(crc, params.expect);
+        EXPECT_EQ(crc, params.expect_crc);
         EXPECT_EQ(0, memcmp(params.buf, dstbuf, params.len));
     }
 };
```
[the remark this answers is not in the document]
nmoinvaz: It is no longer possible for the test strings.

mtl1979:
> Ok, I have removed them. Who knew merging these test strings would be such a big ordeal.. /facepalm

Sometimes those facepalm edits reveal subtle inconsistencies that might cause bugs when two implementations handle edge cases differently.

nmoinvaz:
Yay!

PR #2079, opened 2026-01-07
Add ARCH defines for better code clarity

nmoinvaz:


fallback_builtins.h
```diff
@@ -2,7 +2,7 @@
 #define FALLBACK_BUILTINS_H
 
 #if defined(_MSC_VER) && !defined(__clang__)
-#if defined(_M_IX86) || defined(_M_AMD64) || defined(_M_IA64) ||  defined(_M_ARM) || defined(_M_ARM64) || defined(_M_ARM64EC)
+#if defined(ARCH_X86) || defined(ARCH_IA64) || defined(ARCH_ARM)
```
Dead2: Maybe you could fix some indentation here at the same time? This whole file is a mess..

zbuild.h
```diff
@@ -287,27 +288,26 @@
...
 #if !defined(OPTIMAL_CMP)
-#  if defined(__x86_64__) || defined(_M_X64) || defined(__amd64__) || defined(_M_AMD64)
-#    define OPTIMAL_CMP 64
-#  elif defined(__i386__) || defined(__i486__) || defined(__i586__) || \
-        defined(__i686__) || defined(_X86_) || defined(_M_IX86)
-#    define OPTIMAL_CMP 32
-#  elif defined(__aarch64__) || defined(_M_ARM64) || defined(_M_ARM64EC)
-#    if defined(__ARM_FEATURE_UNALIGNED) || defined(_WIN32)
-#      define OPTIMAL_CMP 64
+#  ifdef ARCH_64BIT
```
Dead2, line 291: This is so much more readable now. 👍

Dead2 (commented):
Some really good changes here I think, some places become a lot more readable.
But I also think this is a benefit for those platforms we don't actively work on and test, with this we don't need to add every possible arch to some of these checks, improving both portability and maintainability.

fallback_builtins.h
nmoinvaz: Done.

Dead2 (approved):
LGTM

zbuild.h
pps83: even though it's might be more readable, it makes lib less portable now. This makes the code unnecessarily dependent on these redundant build flags to be set by the build system, while all of that could be handled in the code directly (compiler knows for a fact if it compiles for 32 or 64 bit, and for what platform, except if code is being compiled for some obscure system/compiler with unknown flags, only in this case testing withing buildsystem makes sense). 
in short, it would be better to have all that in some header that's used during compilation only just like many libs do it.

arch/arm/acle_intrins.h
```diff
@@ -9,7 +9,7 @@
 #endif
 
 #ifdef ARM_CRC32
-#if defined(__aarch64__)
+#if defined(ARCH_ARM) && defined(ARCH_64BIT)
```
pps83, line 12: imo it might have been more clear to have `ARCH_ARM64` (`ARCH_ARM64` could also mean that `ARCH_ARM` is also defined or not). That is 32bit-only arm test could be `defined ARCH_ARM` or `defined ARCH_ARM && !defined ARCH_ARM64`. Same for `ARM_X86` (and it's 64bit version).

zbuild.h
nmoinvaz: I thought about that when I did this but this seemed easier. It think it only sets 1/2 extra defines, right? And either way people need to go through CMake or configure.

We could make detect-arch.c logic a shared header or something.

pps83: Yes, something like `zlib-arch.h` (or `zlibng-arch.h`).
I simply use amalgamated [zlib.cpp](https://github.com/pps83/zlib-ng/blob/master/zlib.cpp) where all that logic to flip required flags is present. It broke with this update (I'll have to add these new defines). This allows me to add this file to any project without any need to have anything defined/configured (similar to [stb libs](https://github.com/nothings/stb))

pps83: currently, this is what cmake build sets when I try to build. It's kind of messy (note redundant defines like `__SSE2__` and `X86_SSE2`). This is what I get for `zlib-ng-static` for x64 release:
```
%(PreprocessorDefinitions)
WIN32
_WINDOWS
NDEBUG
ARCH_X86
ARCH_64BIT
ZLIB_COMPAT
WITH_GZFILEOP
__SSE__=1
__SSE2__=1
NO_FSEEKO
HAVE_CPUID_MS
HAVE_BUILTIN_ASSUME_ALIGNED
_CRT_SECURE_NO_DEPRECATE
_CRT_NONSTDC_NO_DEPRECATE
WITH_OPTIM
X86_FEATURES
X86_HAVE_XSAVE_INTRIN
X86_SSE2
X86_SSSE3
X86_SSE41
X86_SSE42
X86_PCLMULQDQ_CRC
X86_AVX2
X86_AVX512
X86_AVX512VNNI
X86_VPCLMULQDQ_CRC
_MBCS
CMAKE_INTDIR="RelWithDebInfo"
```
most of the defines shouldn't be passed. Ideally, it should be empty list for default build. All those like `X86_PCLMULQDQ_CRC` should be implied by runtime CPU detection, which, in tern, should be implied if omitted.

pps83: ms compiler basically enables compilation of any avx-512 even if you compile for pentinum-1. With gcc/clang if you compile for cpu detection you can use pragmas in code to compile for any arch even if default passed flags are not adequate to build code for recent arch.
See `ZLIB_TARGET_REGION("bmi2,avx2,pclmul,sse4.1")` in my zlib.cpp

nmoinvaz: Those X86 flags are feature flags too. So the build system determine if the compiler has it and if the user has enabled it in the build system.

pps83: all the feature flags are kind of irrelevant when building for runtime cpu detection, right? Basically when I build release x64 for runtime cpu detection **all** these feature flags (related to arch options) have to be enabled.

nmoinvaz: Not if you don't need certain implementations and want to reduce code size. For example Windows has 5 or so different implementations for adler32.

PR #2082, opened 2026-01-09
Reset compiler flags after completing intrinsic tests

Dead2:
Add missing reset of compiler flags after completing a test, avoids the next test inheriting the previous flags.


nmoinvaz (approved):

mtl1979 (approved):
LGTM

PR #2090, opened 2026-01-10
CI: Add benchmarks to codecov ignore

Dead2:
We already avoid collecting coverage when running benchmarks because the benchmarks do not perform most error checking, thus even though they might code increase coverage, they won't detect most bugs unless it actually crashes the whole benchmark.

Now, lets also remove benchmark code from being counted in the zlib-ng code coverage percentage.

### DIFF
--- a/.codecov.yaml
+++ b/.codecov.yaml
@@ -108,11 +108,6 @@ component_management:
         - test/**
         - "!test/minigzip.c"
         - "!test/minideflate.c"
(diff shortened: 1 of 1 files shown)


Dead2:
Not sure how to completely remove the benchmarks from coverage data..
They are still being reported as "no data" https://app.codecov.io/gh/zlib-ng/zlib-ng/pull/2090/indirect-changes

PR #2093, opened 2026-01-10
Simplify the gzread.c name mangling workaround

Dead2:
Simplify the gzread.c name mangling workaround by splitting out just the workaround into a separate file.

This enables:
- We can now browse gzread.c with code highlighting without adding `*.c.in` as a special case in editors
- Codecov can record coverage data for gzread.c

Likely other tools like static analyzers will also have an easier time understanding this file now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,7 +79,7 @@ a.out
 Makefile.tmp
 /makecrct
 /maketrees
-gzread.c
+gzread_mangle.h
 
 /Debug
 /example.dir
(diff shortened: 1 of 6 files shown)


nmoinvaz (approved):
Very clever.

phprus:
https://github.com/zlib-ng/zlib-ng/blob/298f65952faf958b4664bd5be8d816e9599fb7cb/Makefile.in#L238-L248

In `Makefile.in`, the path to `gzread.c` must be specified with `$(SRCDIR)/`.

PR #2094, opened 2026-01-10
Refine codecov config

Dead2:
Based on what we have seen in CI codecov reports since the recent changes, it seems the informational setting is not effective, perhaps it would need to be set for each component? Documentation does not indicate so though.

This PR attempts to refine the coverage-related CI status messages, hopefully they will not fail as often and rather become relevant information when they do fail.

- Don't count tests and tools towards overall project coverage.
- Set coverage targets to 80%.
- Loosen project coverage reduction threshold to 10% to avoid failing coverage tests when CI happens to run on hosts that do not support AVX-512.

### DIFF
--- a/.codecov.yaml
+++ b/.codecov.yaml
@@ -12,14 +12,22 @@ coverage:
   status:
     project:
       default:
-        threshold: 0.07
-        informational: true
+        target: 80%      # Target 80% project coverage
+        threshold: 10    # Allow pass even when project coverage drops by 10% (because of AVX512)
+        paths:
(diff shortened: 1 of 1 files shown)


nmoinvaz (approved):

Dead2:
@coderabbitai The `codecov/project/deflate` (and others) still fail coverage tests because it decreased by 0.06%, even though the coverage is still 84.97%.
What settings need to be set to actually allow codecov "components" to drop a little in coverage without throwing a CI error?
@copilot You might also answer this.

PR #2096, opened 2026-01-11
CI: Misc CI coverage improvements and CI cleanups

Dead2:
- Improve detection of compiler code coverage
  - Enable -fcondition-coverage if it is supported
- Enable codecov for more CI jobs
  - Disable codecov if -O1 or higher is requested, as coverage generation sets -O0
  - Disable codecov if tests are not run
  - Add comments for jobs without codecov, makes it easier to see what is missing and why
- Merge several jobs that do not need to be run separately
  - Remove some jobs that end up being redundant, saving some CI time
- Expand testing of configure parameters
- Use clang-20 for most builds
  - Let one job use clang-15 and one clang-19
- Add compiler defines information output to CMake
- Fix .codecov.yaml parameter placement warnings

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -381,8 +381,8 @@ jobs:
 
           - name: ${{ github.repository == 'zlib-ng/zlib-ng' && 'EL10' || 'Ubuntu' }} Clang S390X DFLTCC ${{ (github.repository == 'zlib-ng/zlib-ng' && 'MSAN') || 'Compat' }}
             os: ${{ github.repository == 'zlib-ng/zlib-ng' && 'z15' || 'ubuntu-latest' }}
-            compiler: ${{ github.repository == 'zlib-ng/zlib-ng' && 'clang' || 'clang-15' }}
-            cxx-compiler: ${{ github.repository == 'zlib-ng/zlib-ng' && 'clang++' || 'clang++-15' }}
+            compiler: clang
+            cxx-compiler: clang++
             cmake-args: >-
               ${{ github.repository == 'zlib-ng/zlib-ng' && '-GNinja -DWITH_SANITIZER=Memory' || '-DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-s390x.cmake -DZLIB_COMPAT=ON' }}
               -DWITH_DFLTCC_DEFLATE=ON -DWITH_DFLTCC_INFLATE=ON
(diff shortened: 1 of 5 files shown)


.github/workflows/cmake.yml
```diff
@@ -433,177 +425,184 @@ jobs:
...
+          - name: Ubuntu Clang-20 Undefined Symbols
             os: ubuntu-latest
-            compiler: clang-15
-            cxx-compiler: clang++-15
+            compiler: clang-20
+            cxx-compiler: clang++-20
             cmake-args: -DCMAKE_SHARED_LINKER_FLAGS="-fuse-ld=lld -Wl,--no-undefined-version" -DZLIBNG_ENABLE_TESTS=OFF
             build-shared: ON
-            packages: clang-15 llvm-15 lld
+            packages: clang-20 llvm-20 lld
```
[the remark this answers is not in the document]
Dead2: The "ON" parameter here is passed to CMake.

CMakeLists.txt
```diff
@@ -1677,6 +1677,16 @@ if(BUILD_TESTING)
...
+#============================================================================
+# Config summary
+#============================================================================
+message("")
+message(STATUS "The following defines have been set (partial list):\n")
+get_property(propvalraw DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR} PROPERTY COMPILE_DEFINITIONS)
+list(SORT propvalraw)
+string(REPLACE ";" " " propval "${propvalraw}")
+message("CURRENT_SOURCE_DIR: ${propval}\n")
+
```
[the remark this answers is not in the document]
Dead2: The list contains the defines that are added for the current source directory, thus combined with the header printed before it, the naming makes sense.

mtl1979:
This is bad... Reveals internal issues in LLVM... Several CI runs fail. Some possible infinite loops that need to be verified.

Dead2:
> This is bad... Reveals internal issues in LLVM... Several CI runs fail. Some possible infinite loops that need to be verified.

Yeah, #2101 and #2097 fix two of the three problems.
Have not decided what to do about the MSAN problem yet, probably try a newer llvm version.

At least now coverage works with clang, but it only works for the default version of clang, not if you install a specific version, CMake test does not succeed. That does work for me locally in an ubuntu 24.04 docker however, so not sure what is actually going on there.

mtl1979:
> > This is bad... Reveals internal issues in LLVM... Several CI runs fail. Some possible infinite loops that need to be verified.
> 
> Yeah, #2101 and #2097 fix two of the three problems. Have not decided what to do about the MSAN problem yet, probably try a newer llvm version.
> 

Need to hold this PR until all the issues are fixed, then rebase... A lot easier than breaking this PR to part that doesn't cause failure(s) and part that depends on other PRs...

#2097 is almost ready for merging... I did already remove that one "internal" entry in the map file, but if there is something that still needs to be done and doesn't break the logic, then just comment on that PR and I will see what can be done...

cmake/detect-coverage.cmake
```diff
@@ -1,29 +1,42 @@
...
-    set(CMAKE_REQUIRED_LINK_OPTIONS -coverage)
-    check_c_compiler_flag(-coverage HAVE_COVERAGE)
+    # Check for --coverage -fcondition-coverage flag support
+    set(CMAKE_REQUIRED_LINK_OPTIONS -coverage -fcondition-coverage)
+    check_c_compiler_flag("--coverage -fcondition-coverage -Wno-coverage-too-many-conditions" HAVE_CONDITION_COVERAGE)
     set(CMAKE_REQUIRED_LINK_OPTIONS)
+    if(HAVE_CONDITION_COVERAGE)
+        add_link_options(-coverage -fcondition-coverage)
+        add_compile_options(--coverage -fcondition-coverage -Wno-coverage-too-many-conditions)
+        message(STATUS "Code coverage enabled using: --coverage -fcondition-coverage")
```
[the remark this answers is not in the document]
Dead2: That is why we test whether the compiler supports the flag, then fall back to something less ideal.

CMakeLists.txt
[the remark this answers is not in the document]
Dead2: I already told you this does make sense, because it comes under the heading "The following defines have been set", and those are the defines that were set for the source dir specifically.

.github/workflows/cmake.yml
```diff
@@ -433,177 +425,184 @@ jobs:
...
             os: ubuntu-latest
-            compiler: clang-18
-            cxx-compiler: clang++-18
-            cmake-args: -DCMAKE_C_STANDARD=23
-            packages: clang-18 llvm-18 llvm-18-tools
+            compiler: clang-20
+            cxx-compiler: clang++-20
+            cmake-args: -DCMAKE_LINKER_TYPE=LLD -DCMAKE_C_STANDARD=23
+            packages: clang-20 lld
+            # codecov disabled for clang-20, errors
```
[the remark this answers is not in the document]
Dead2: It has been working fine thus far, so I don't think that will be necessary.

Dead2:
Refactored and reordered most of this to result in smaller individual commits, split a part out to #2101, and rebased.

.github/workflows/configure.yml
mtl1979: Fails here, needs to be absolute path.

mtl1979 (changes requested):
Relative build path is not allowed for `Upload build errors`

.github/workflows/configure.yml
Dead2: What do you suggest to fix it?

mtl1979: My initial thought was to use `**/configure.log`... It's very unlikely that the pattern would match anything unnecessary...

mtl1979: `Error: Invalid pattern '../build/configure.log'. Relative pathing '.' and '..' is not allowed.`

Dead2: fixed, I think

Dead2:
Rebased and passing all CI checks

.github/workflows/cmake.yml
```diff
@@ -217,19 +213,12 @@ jobs:
...
-            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-aarch64.cmake -DWITH_ARMV8=OFF -DWITH_SANITIZER=Undefined
-            packages: qemu-user gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libc-dev-arm64-cross
-            gcov-exec: aarch64-linux-gnu-gcov
-            codecov: ubuntu_gcc_aarch64_no_armv8
-
-          - name: Ubuntu GCC AARCH64 No NEON UBSAN
+          - name: Ubuntu GCC AARCH64 No Neon No ARMv8 UBSAN
             os: ubuntu-22.04
-            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-aarch64.cmake -DWITH_NEON=OFF -DWITH_SANITIZER=Undefined
+            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-aarch64.cmake -DWITH_NEON=OFF -DWITH_ARMV8=OFF -DWITH_SANITIZER=Undefined
```
mtl1979: This is basically same as setting just `-DWITH_OPTIM=OFF` as AArch64 is same as ARMv8.1A. We don't support anything beyond ARMv8.1A yet, even though work has been done to add partial support for v8.4A and v9.0A.

.github/workflows/configure.yml
```diff
@@ -77,17 +75,10 @@ jobs:
...
             compiler: aarch64-linux-gnu-gcc
-            configure-args: --warn --without-armv8
-            chost: aarch64-linux-gnu
-            packages: qemu-user gcc-aarch64-linux-gnu libc-dev-arm64-cross
-
-          - name: Ubuntu GCC AARCH64 No NEON
-            os: ubuntu-latest
-            compiler: aarch64-linux-gnu-gcc
-            configure-args: --warn --without-neon
+            configure-args: --warn --without-neon --without-armv8
```
mtl1979: Same as `--without-optimizations` ... There is nothing else you can disable for optimizations.

.github/workflows/configure.yml
```diff
@@ -128,6 +119,13 @@ jobs:
             chost: powerpc-linux-gnu
             packages: qemu-user gcc-powerpc-linux-gnu libc-dev-powerpc-cross
 
+          - name: Ubuntu GCC PPC No Power9
+            os: ubuntu-latest
+            compiler: powerpc-linux-gnu-gcc
+            configure-args: --warn --without-power9
+            chost: powerpc-linux-gnu
+            packages: qemu-user gcc-powerpc-linux-gnu libc-dev-powerpc-cross
+
```
mtl1979: This makes no sense as PPC doesn't even support Power9 (or Power7 or Power8 or Power10)...

For the sake of testing impossible combinations, we could disable both Power8 and Power9 and that would reduce the configure time to minimum.

Linux for 64-bit Power processors used to have 32-bit userland and the same gcc build is still used for 32-bit PowerPC.

mtl1979 (changes requested):
Needs some cleaning up...

.github/workflows/cmake.yml
```diff
@@ -59,8 +59,8 @@ jobs:
             build-src-dir: ../zlib-ng
             # Build with readonly project directory require CMake 3.19+
             readonly-project-dir: true
-            codecov: ubuntu_gcc_osb
             cflags: -O3
+            # Skipping codecov to allow -O3
```
nmoinvaz, line 63: Good idea

.github/workflows/cmake.yml
nmoinvaz: I had the same thoughts. The reason why we had `-D WITH_NEON=OFF` only was to insolate only certain implementations for code coverage. So previously it would cover `slide_hash_armv6.c`, `crc32_armv8.c`, and `crc32_armv8_pmull_eor3.c`, but now it will only cover  `slide_hash_armv6.c`. Maybe this is intentional, it is hard to tell.

cmake/detect-coverage.cmake
```diff
@@ -1,29 +1,42 @@
 # detect-coverage.cmake -- Detect supported compiler coverage flags
 # Licensed under the Zlib license, see LICENSE.md for details
 
+# Attempt to enable gcov-style code coverage
 macro(add_code_coverage)
-    # Check for -coverage flag support for Clang/GCC
-    set(CMAKE_REQUIRED_LINK_OPTIONS -coverage)
-    check_c_compiler_flag(-coverage HAVE_COVERAGE)
+    # Check for --coverage -fcondition-coverage flag support
+    set(CMAKE_REQUIRED_LINK_OPTIONS -coverage -fcondition-coverage)
```
nmoinvaz: Why not check for `--coverage` first and then check for `--coverage` with `--fcondition-coverage`? So this detection would go under the `HAVE_COVERAGE` branch.

.github/workflows/configure.yml
nmoinvaz: See my previous comment, there is still `--without-armv6` so it will only test `slide_hash_armv6.c`. I originally thought it was the same also. I'm not sure what benefit `--widthout-armv8` really has, other than forcing crc32 to use `generic` implementations. But the generic crc32 implementation is tested already by `-DWITH_OPTIM=OFF`.

.github/workflows/cmake.yml
mtl1979: AArch64 dropped support for ARMv6 intrinsics, so this doesn't need to worry about them.

.github/workflows/configure.yml
mtl1979: No ARMv6 intrinsics support on AArch64.

.github/workflows/cmake.yml
Dead2: Removed, using WITH_OPTIM=OFF to test both cases since they are not overlapping.

.github/workflows/configure.yml
Dead2: Removed, using --without-optimization to test both cases since they are not overlapping.

.github/workflows/configure.yml
Dead2: Removed

cmake/detect-coverage.cmake
Dead2: Changed

mtl1979 (approved):
LGTM

PR #2071, opened 2026-01-01
Add elf_aux_info() support on FreeBSD/OpenBSD for PMULL and EOR3

brad0:
Use elf_aux_info() to detect PMULL and EOR3 on FreeBSD and OpenBSD
aarch64.

PR #2099, opened 2026-01-13
Cleanup arm feature checks and add PMULL&EOR3 for missing OSes

nmoinvaz:
Superseeds #2071.

This clean up prevents a lot of `return 0` that ended up and messy logic that was necessary in #2071.

### DIFF
--- a/arch/arm/arm_features.c
+++ b/arch/arm/arm_features.c
@@ -26,205 +26,179 @@
 #endif
 
 static int arm_has_crc32(void) {
+    int has_crc32 = 0;
 #if defined(ARM_AUXV_HAS_CRC32)
 #  if defined(__FreeBSD__) || defined(__OpenBSD__)
 #    ifdef HWCAP_CRC32
        unsigned long hwcap = 0;
        elf_aux_info(AT_HWCAP, &hwcap, sizeof(hwcap));
-       return (hwcap & HWCAP_CRC32) != 0 ? 1 : 0;
+       has_crc32 = (hwcap & HWCAP_CRC32) != 0;
 #    else
        unsigned long hwcap2 = 0;
        elf_aux_info(AT_HWCAP2, &hwcap2, sizeof(hwcap2));
-       return (hwcap2 & HWCAP2_CRC32) != 0 ? 1 : 0;
+       has_crc32 = (hwcap2 & HWCAP2_CRC32) != 0;
 #    endif
 #  else
 #    ifdef HWCAP_CRC32
-       return (getauxval(AT_HWCAP) & HWCAP_CRC32) != 0 ? 1 : 0;
+       has_crc32 = (getauxval(AT_HWCAP) & HWCAP_CRC32) != 0;
 #    else
-       return (getauxval(AT_HWCAP2) & HWCAP2_CRC32) != 0 ? 1 : 0;
+       has_crc32 = (getauxval(AT_HWCAP2) & HWCAP2_CRC32) != 0;
 #    endif
 #  endif
 #elif defined(__FreeBSD__) && defined(__aarch64__)
-    return getenv("QEMU_EMULATING") == NULL
+    has_crc32 = getenv("QEMU_EMULATING") == NULL
       && ID_AA64ISAR0_CRC32_VAL(READ_SPECIALREG(id_aa64isar0_el1)) >= ID_AA64ISAR0_CRC32_BASE;
 #elif defined(__OpenBSD__) && defined(__aarch64__)
-    int hascrc32 = 0;
     int isar0_mib[] = { CTL_MACHDEP, CPU_ID_AA64ISAR0 };
     uint64_t isar0 = 0;
     size_t len = sizeof(isar0);
     if (sysctl(isar0_mib, 2, &isar0, &len, NULL, 0) != -1) {
-      if (ID_AA64ISAR0_CRC32(isar0) >= ID_AA64ISAR0_CRC32_BASE)
-          hascrc32 = 1;
+      has_crc32 = ID_AA64ISAR0_CRC32(isar0) >= ID_AA64ISAR0_CRC32_BASE;
     }
-    return hascrc32;
 #elif defined(__APPLE__)
-    int hascrc32;
-    size_t size = sizeof(hascrc32);
-    return sysctlbyname("hw.optional.armv8_crc32", &hascrc32, &size, NULL, 0) == 0
-      && hascrc32 == 1;
+    int has_feat = 0;
+    size_t size = sizeof(has_feat);
+    has_crc32 = sysctlbyname("hw.optional.armv8_crc32", &has_feat, &size, NULL, 0) == 0
+        && has_feat == 1;
 #elif defined(_WIN32)
-    return IsProcessorFeaturePresent(PF_ARM_V8_CRC32_INSTRUCTIONS_AVAILABLE);
+    has_crc32 = IsProcessorFeaturePresent(PF_ARM_V8_CRC32_INSTRUCTIONS_AVAILABLE);
 #elif defined(ARM_NOCHECK_CRC32)
-    return 1;
-#else
-    return 0;
+    has_crc32 = 1;
 #endif
+    return has_crc32;
 }
 
 static int arm_has_pmull(void) {
+    int has_pmull = 0;
 #if defined(__linux__) && defined(HAVE_SYS_AUXV_H)
 #  ifdef HWCAP_PMULL
-    return (getauxval(AT_HWCAP) & HWCAP_PMULL) != 0 ? 1 : 0;
-#  else
+    has_pmull = (getauxval(AT_HWCAP) & HWCAP_PMULL) != 0;
+#  elif defined(HWCAP_AES)
     /* PMULL is part of crypto extension, check for AES as proxy */
-#    ifdef HWCAP_AES
-    return (getauxval(AT_HWCAP) & HWCAP_AES) != 0 ? 1 : 0;
-#    else
-    return 0;
-#    endif
+    has_pmull = (getauxval(AT_HWCAP) & HWCAP_AES) != 0;
 #  endif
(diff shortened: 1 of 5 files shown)


arch/arm/arm_features.c
```diff
@@ -26,209 +26,192 @@
...
+#elif defined(__FreeBSD__) || defined(__OpenBSD__)
+#  ifdef HWCAP_CRC32
+    unsigned long hwcap = 0;
+    elf_aux_info(AT_HWCAP, &hwcap, sizeof(hwcap));
+    has_crc32 = (hwcap & HWCAP_CRC32) != 0;
+#  elif defined(HWCAP2_CRC32)
+    unsigned long hwcap2 = 0;
+    elf_aux_info(AT_HWCAP2, &hwcap2, sizeof(hwcap2));
+    has_crc32 = (hwcap2 & HWCAP2_CRC32) != 0;
+#  endif
```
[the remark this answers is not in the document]
nmoinvaz: I have fixed HAVE_SYS_AUXV_H in arm_features. The power changes should probably wait for another PR.. maybe..

arch/arm/arm_features.c
```diff
@@ -26,209 +26,192 @@
...
         || strncmp(platform, "v7l", 3) == 0
         || strncmp(platform, "v8l", 3) == 0;
 #elif defined(ARM_NOCHECK_SIMD)
-    return 1;
-#else
-    return 0;
+    has_simd = 1;
 #endif
+    return has_simd;
 }
```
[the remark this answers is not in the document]
nmoinvaz: Fixed.

arch/arm/arm_features.c
```diff
@@ -258,23 +241,33 @@ static inline int arm_has_simd(void) {
...
+#  elif defined(HWCAP2_CPUID)
+    has_cpuid = (getauxval(AT_HWCAP2) & HWCAP2_CPUID) != 0;
+#  endif
+#elif defined(__FreeBSD__) || defined(__OpenBSD__)
+    unsigned long hwcap = 0;
+    elf_aux_info(AT_HWCAP, &hwcap, sizeof(hwcap));
+    has_cpuid = (hwcap & HWCAP_CPUID) != 0;
+#endif
+    return has_cpuid;
+}
```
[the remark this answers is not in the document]
nmoinvaz: Fixed.

arch/arm/arm_features.c
```diff
@@ -26,209 +26,193 @@
...
+#if defined(__linux__) && defined(HAVE_SYS_AUXV_H)
+#  ifdef HWCAP_ARM_NEON
+    has_neon = (getauxval(AT_HWCAP) & HWCAP_ARM_NEON) != 0;
+#  elif defined(HWCAP_NEON)
+    has_neon = (getauxval(AT_HWCAP) & HWCAP_NEON) != 0;
 #  endif
+#elif (defined(__FreeBSD__) || defined(__OpenBSD__)) && defined(HAVE_SYS_AUXV_H)
+    unsigned long hwcap = 0;
+    elf_aux_info(AT_HWCAP, &hwcap, sizeof(hwcap));
+    has_neon = (hwcap & HWCAP_NEON) != 0;
```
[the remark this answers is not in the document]
nmoinvaz: Fixed.

arch/arm/arm_features.c
```diff
@@ -26,209 +26,192 @@
...
 #elif defined(__APPLE__)
     /* All Apple Silicon (M1+) has SHA3/EOR3 support */
-#  if defined(__aarch64__) || defined(_M_ARM64)
-    int hassha3;
-    size_t size = sizeof(hassha3);
-    if (sysctlbyname("hw.optional.arm.FEAT_SHA3", &hassha3, &size, NULL, 0) == 0 && hassha3 == 1)
-        return 1;
+    size_t size = sizeof(has_eor3);
+    has_eor3 = sysctlbyname("hw.optional.arm.FEAT_SHA3", &has_eor3, &size, NULL, 0) == 0
+        && has_eor3 == 1;
```
mtl1979: The first test should be isolated, it clobbers the value of `has_eor3`.
The second test seems redundant, it's checking own value...

arch/arm/arm_features.c
```diff
@@ -26,209 +26,192 @@
...
-#  else
-    return 0;
-#  endif
-#elif defined(_WIN32)
-    /* Windows: No direct API for SHA3, return 0 for now */
-    return 0;
+    if (!has_eor3) {
+        size = sizeof(has_eor3);
+        has_eor3 = sysctlbyname("hw.optional.armv8_2_sha3", &has_eor3, &size, NULL, 0) == 0
+            && has_eor3 == 1;
```
mtl1979: Same as above, first test clobbers value of `has_eor3`.

mtl1979 (changes requested):
Variable tests own value...

arch/arm/arm_features.c
nmoinvaz: Why do you need the first value if it fails the test? It seems logically correct.

nmoinvaz: Fixed.

arch/arm/arm_features.c
nmoinvaz: Fixed.

arch/arm/arm_features.c
mtl1979: > Why do you need the first value if it fails the test? It seems logically correct.

The issue is the evaluation order... `== 0` is evaluated after `sysctlbyname` modifies `has_eor3` and returns, so when the first condition is true, the result is overwritten and the second test is always true. Using two separate variables indeed fix the issue.

arch/arm/arm_features.c
```diff
@@ -26,209 +26,199 @@
...
+        && has_feat == 1;
     /* Fallback to legacy name for older macOS versions */
-    size = sizeof(hassha3);
-    return sysctlbyname("hw.optional.armv8_2_sha3", &hassha3, &size, NULL, 0) == 0
-      && hassha3 == 1;
-#  else
-    return 0;
-#  endif
-#elif defined(_WIN32)
-    /* Windows: No direct API for SHA3, return 0 for now */
```
mtl1979, line 171: `PF_ARM_SHA3_INSTRUCTIONS_AVAILABLE`

mtl1979:
```
diff --git a/cmake/detect-intrinsics.cmake b/cmake/detect-intrinsics.cmake
index 510910d2..118bfbc2 100644
--- a/cmake/detect-intrinsics.cmake
+++ b/cmake/detect-intrinsics.cmake
@@ -78,7 +78,11 @@ macro(check_armv8_pmull_eor3_compiler_flag)
         #else
         #  include <arm_neon.h>
         #endif
+        #ifdef _MSC_VER
+        __n128 f(__n64 a, __n64 b) {
+        #else
         poly128_t f(poly64_t a, poly64_t b) {
+        #endif
             return vmull_p64(a, b);
         }
         uint64x2_t g(uint64x2_t a, uint64x2_t b, uint64x2_t c) {
```

mtl1979: Not 100% sure about this one:
```
diff --git a/arch/arm/crc32_armv8_pmull_eor3.c b/arch/arm/crc32_armv8_pmull_eor3.c
index d7a15074..9b83a5d7 100644
--- a/arch/arm/crc32_armv8_pmull_eor3.c
+++ b/arch/arm/crc32_armv8_pmull_eor3.c
@@ -16,9 +16,15 @@

 /* Carryless multiply low 64 bits: a[0] * b[0] */
 static inline uint64x2_t clmul_lo(uint64x2_t a, uint64x2_t b) {
+#ifdef _MSC_VER
+    return vreinterpretq_u64_p128(vmull_p64(
+        vget_low_p64(vreinterpret_p64_u64(a)),
+        vget_low_p64(vreinterpret_p64_u64(b))));
+#else
     return vreinterpretq_u64_p128(vmull_p64(
         vget_lane_p64(vreinterpret_p64_u64(vget_low_u64(a)), 0),
         vget_lane_p64(vreinterpret_p64_u64(vget_low_u64(b)), 0)));
+#endif
 }

 /* Carryless multiply high 64 bits: a[1] * b[1] */
@@ -28,7 +34,11 @@ static inline uint64x2_t clmul_hi(uint64x2_t a, uint64x2_t b) {

 /* Carryless multiply of two 32-bit scalars: a * b (returns 64-bit result in 128-bit vector) */
 static inline uint64x2_t clmul_scalar(uint32_t a, uint32_t b) {
+#ifdef _MSC_VER
+  return vreinterpretq_u64_p128(vmull_p64(vdup_n_p64((poly64_t)a), vdup_n_p64((poly64_t)b)));
+#else
   return vreinterpretq_u64_p128(vmull_p64((poly64_t)a, (poly64_t)b));
+#endif
 }

 /* Compute x^n mod P (CRC-32 polynomial) in log(n) time, where P = 0x104c11db7 */
```

mtl1979 (changes requested):
Windows/Visual C++ does support SHA3/PMULL on AArch64.

arch/arm/arm_features.c
nmoinvaz: Seemed to compile for me locally. I have added it as a separate commit with you as author.

mtl1979: I don't have AArch64 device with Windows, so I couldn't run tests on it, so that's why I said I'm not 100% sure about it being correct (or fastest implementation)... Windows doesn't have `poly128_t` or `uint128_t`, so it uses own types instead of mixing them in same intrinsic. `vdup_n_p64` basically converts between MSVC's own type (`__n64`) and `poly64_t` which is just alias to `unsigned long long`.

nmoinvaz: I use Windows ARM in Parallels Desktop on macOS. It runs on M3, so not a true Windows ARM with Snapdragon CPU.

mtl1979 (approved):

Dead2 (approved):

PR #2101, opened 2026-01-13
Update static analyzer to GCC-14, fix infinite loop warning, add Z_UNREACHABLE

Dead2:
Updating to the newer static analyzer in GCC-14 from GCC-10 revealed a possible infinite loop in gz_fetch(), it doesn't happen, but the compiler/analyzer can't know that.
Ref: https://github.com/zlib-ng/zlib-ng/actions/runs/20957066511/job/60224228645?pr=2096

- Update static analyzer to GCC-14
- Add default case that returns error in gz_fetch()
- According to the comment, gz_fetch() also assumes that state->x.have == 0, so lets add an Assert to that effect.
- Add Z_UNREACHABLE hint and use it for unreachable default cases.

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -5,7 +5,7 @@ concurrency:
   cancel-in-progress: true
 jobs:
   static-analysis:
-    name: GCC
+    name: GCC-14
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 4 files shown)


nmoinvaz (approved):
Looks ok to me.

Issue #2100, opened 2026-01-13
Deflate output not deterministic when the compressor is reused with deflateReset


QrczakMK:
See https://github.com/trifectatechfoundation/zlib-rs/issues/459 for details.

The bug was found in zlib-rs, but zlib-ng has the same bug.

QrczakMK:
Possible fix:

```
--- zlib-ng+/match_tpl.h.bak	2023-03-17 21:27:56.000000000 +0100
+++ zlib-ng+/match_tpl.h	2026-01-13 10:59:04.843864020 +0100
@@ -141,8 +141,10 @@
 
         if (len > best_len) {
             s->match_start = cur_match;
-            /* Do not look for matches beyond the end of the input. */
-            if (len > lookahead)
+            /* Do not look for better matches if the current match reaches
+             * or exceeds the end of the input.
+             */
+            if (len >= lookahead)
                 return lookahead;
             best_len = len;
             if (best_len >= nice_match)
```

QrczakMK:
To reproduce:

```
#include <stddef.h>

#include <cstdlib>
#include <iomanip>
#include <iostream>
#include <limits>
#include <memory>
#include <string>

#include "zlib.h"

using namespace std::literals;

std::string Compress(std::string_view src, z_stream* z_stream_ptr) {
  std::string compressed;
  const size_t buffer_size = 4096;
  const std::unique_ptr<char[]> buffer(new char[buffer_size]);
  z_stream_ptr->next_in =
      const_cast<z_const Bytef*>(reinterpret_cast<const Bytef*>(src.data()));
  for (;;) {
    size_t avail_in = static_cast<size_t>(
        src.data() + src.size() -
        reinterpret_cast<const char*>(z_stream_ptr->next_in));
    int op = Z_FINISH;
    if (avail_in > std::numeric_limits<uInt>::max()) {
      avail_in = size_t{std::numeric_limits<uInt>::max()};
      op = Z_NO_FLUSH;
    }
    z_stream_ptr->avail_in = static_cast<uInt>(avail_in);
    z_stream_ptr->next_out = reinterpret_cast<Bytef*>(buffer.get());
    z_stream_ptr->avail_out = static_cast<uInt>(buffer_size);
    const int zlib_code = deflate(z_stream_ptr, op);
    compressed.append(buffer.get(),
                      static_cast<size_t>(reinterpret_cast<const char*>(
                                              z_stream_ptr->next_out) -
                                          buffer.get()));
[...]

PR #2102, opened 2026-01-13
Make deflate output deterministic if PREFIX3(stream) is reused after deflateReset

phprus:
Issue: #2100

@QrczakMK

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -178,15 +178,16 @@ if(WITH_GTEST)
 
         if(NOT TEST_STOCK_ZLIB)
             list(APPEND TEST_SRCS
-                test_adler32.cc             # adler32_neon(), etc
-                test_adler32_copy.cc        # adler32_copy implementations
-                test_compare256.cc          # compare256_neon(), etc
-                test_compare256_rle.cc      # compare256_rle(), etc
-                test_crc32.cc               # crc32_armv8(), etc
-                test_crc32_copy.cc          # crc32_copy implementations
-                test_inflate_sync.cc        # expects a certain compressed block layout
-                test_main.cc                # cpu_check_features()
-                test_version.cc             # expects a fixed version string
+                test_adler32.cc               # adler32_neon(), etc
+                test_adler32_copy.cc          # adler32_copy implementations
+                test_compare256.cc            # compare256_neon(), etc
+                test_compare256_rle.cc        # compare256_rle(), etc
+                test_crc32.cc                 # crc32_armv8(), etc
+                test_crc32_copy.cc            # crc32_copy implementations
+                test_deflate_deterministic.cc # deterministic output after deflateReset
+                test_inflate_sync.cc          # expects a certain compressed block layout
+                test_main.cc                  # cpu_check_features()
+                test_version.cc               # expects a fixed version string
                 )
         endif()
 
--- a/test/test_deflate_deterministic.cc
+++ b/test/test_deflate_deterministic.cc
@@ -1,0 +1,71 @@
+/* test_deflate_deterministic.cc - Test deterministic output after deflateReset */
+
+#include "zbuild.h"
+#ifdef ZLIB_COMPAT
+#  include "zlib.h"
+#else
+#  include "zlib-ng.h"
+#endif
+
+#include <string>
+#include <gtest/gtest.h>
+
+#include "deflate.h"
+#include "test_shared.h"
+
+
+
+/* Issue: https://github.com/zlib-ng/zlib-ng/issues/2100 */
+
+/* len(data_b) must be greater len(data_a) */
+static const uint8_t data_a[] = {  0 , 'A', 'A', 'A', 'A',  0 , 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A' };
+static const uint8_t data_b[] = { 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B' };
+
+static std::string compress_data(const uint8_t *src, size_t len, PREFIX3(stream)* z_stream) {
+    const uint32_t buffer_size = 1024;
+    uint8_t buffer[buffer_size];
+    int err;
+
+    z_stream->next_in = (z_const uint8_t *)src;
+    z_stream->avail_in = (uint32_t)len;
+    z_stream->next_out = buffer;
+    z_stream->avail_out = buffer_size;
+    err = PREFIX(deflate)(z_stream, Z_FINISH);
+    EXPECT_EQ(err, Z_STREAM_END);
+    return std::string((const char *)buffer, (size_t)(z_stream->next_out - buffer));
+}
+
+TEST(deflate, deterministic) {
+    const int compression_level = 6;
+    const int window_bits = 15;
+    int err;
+
+    PREFIX3(stream) a_stream;
+    memset(&a_stream, 0, sizeof(a_stream));
+
+    /* Compress a with newly created z_stream. */
+    err = PREFIX(deflateInit2)(&a_stream, compression_level, Z_DEFLATED, window_bits, 8, Z_DEFAULT_STRATEGY);
+    EXPECT_EQ(err, Z_OK);
+
+    const std::string a_compressed = compress_data(data_a, sizeof(data_a), &a_stream);
+    err = PREFIX(deflateEnd)(&a_stream);
+    EXPECT_EQ(err, Z_OK);
+
+    /* Compress b with newly created z_stream. */
+    PREFIX3(stream) b_stream;
+    memset(&b_stream, 0, sizeof(b_stream));
+
+    err = PREFIX(deflateInit2)(&b_stream, compression_level, Z_DEFLATED, window_bits, 8, Z_DEFAULT_STRATEGY);
+    EXPECT_EQ(err, Z_OK);
+    const std::string b_compressed = compress_data(data_b, sizeof(data_b), &b_stream);
+
+    /* Reset the stream. */
+    err = PREFIX(deflateReset)(&b_stream);
+    EXPECT_EQ(err, Z_OK);
+
+    const std::string a_compressed2 = compress_data(data_a, sizeof(data_a), &b_stream);
+    err = PREFIX(deflateEnd)(&b_stream);
+    EXPECT_EQ(err, Z_OK);
+
+    EXPECT_EQ(a_compressed, a_compressed2);
+}
--- a/match_tpl.h
+++ b/match_tpl.h
@@ -158,8 +158,10 @@ Z_INTERNAL uint32_t LONGEST_MATCH(deflate_state *const s, uint32_t cur_match) {
             uint32_t match_start = cur_match - match_offset;
             s->match_start = match_start;
 
-            /* Do not look for matches beyond the end of the input. */
-            if (len > lookahead)
+            /* Do not look for better matches if the current match reaches
+             * or exceeds the end of the input.
+             */
+            if (len >= lookahead)
                 return lookahead;
             if (len >= nice_match)
                 return len;


test/test_deflate_reset_deterministic.cc
```diff
@@ -0,0 +1,381 @@
...
+;
+
+
+std::string compress_data(const uint8_t *src, size_t len, PREFIX3(stream)* z_stream) {
+    std::string compressed;
+    const uint32_t buffer_size = 4096;
+    char buffer[buffer_size];
+    int err;
+
+    z_stream->next_in = src;
```
mtl1979: Missing explicit cast to remove constness.

mtl1979 (changes requested):
Can't use implicit cast to remove constness of pointer.

test/test_deflate_reset_deterministic.cc
phprus: Fixed.

Dead2:
The bug fix might be correct, but is this actually the root cause though?
It seems to me like something doesn't get reset correctly, otherwise why would anything be different?

mtl1979:
> The bug fix might be correct, but is this actually the root cause though? It seems to me like something doesn't get reset correctly, otherwise why would anything be different?

I don't think this is the root cause... There is likely off-by-one error elsewhere (somewhere in the compare code) and this PR masks it.

dzetkulic:
> > The bug fix might be correct, but is this actually the root cause though? It seems to me like something doesn't get reset correctly, otherwise why would anything be different?
> 
> I don't think this is the root cause... There is likely off-by-one error elsewhere (somewhere in the compare code) and this PR masks it.

https://github.com/trifectatechfoundation/zlib-rs/issues/459 has some investigation on the root cause.

IIUC 8 byte value is read from the internal buffer: `scan_start = zng_memread_8(scan);` It has the tail of the data from the new deflate stream plus the garbage bytes which are left over from the previous compression session. These garbage bytes should not affect the result of the compression, but they do.

To make this deterministic, we could memset the internal buffer to any pattern (zeros maybe), but that it not a good solution in my opinion for two reasons:

1. It will make reset more expensive
2. It will replace garbage bytes with a different garbage bytes. Instead of trying to find a match with garbage, we would try to make a match for zeros. This makes no sense because those bytes should never matter and we should not match them no matter what they are.

mtl1979:
> IIUC 8 byte value is read from the internal buffer: `scan_start = zng_memread_8(scan);` It has the tail of the data from the new deflate stream plus the garbage bytes which are left over from the previous compression session. These garbage bytes should not affect the result of the compression, but they do.

So to avoid the garbage bytes, during reset `scan` needs to be advanced past the bytes from the old session?

QrczakMK:
The high level logic of the relevant part of `LONGEST_MATCH()` is to iterate through matches in a chain. Whenever the current match is longer than the best match so far, look at it closer, and then update the best match so far. At the end, the best match so far is the answer.

Looking closer involves checking whether the preliminary length exceeds the remaining length. If so, then garbage bytes must have artificially increased the preliminary length, so the actual matching length is only the remaining length. Furthermore, there is no need to iterate further through the chain, because the match is already as long as possible. The function exits early. (This differs from the original zlib, where the length is only trimmed at this stage, without an early exit.)

The problem is that, while comparing whether the current match is longer than the best match so far, the length is still a preliminary length, which may involve comparison of garbage bytes after the data. If a match logically reaches the end of the data, then it is unpredictable whether the preliminary length is equal to the remaining length, or greater.

This causes divergence of the behavior in case the best match so far reached the end of the data, while garbage bytes did not match then, so there was no early exit, and we now have another match which logically reaches the end of the data. Depending on garbage bytes, the preliminary length of the second match can be equal to or greater than the remaining length, and thus the new match can be considered either as good as or better than the best match so far. Only a strictly better match is looked at closer, and ultimately this affects the distance used in the answer.

After the fix, the early exit applies as soon as the match logically reaches the end of the data, not only when the preliminary length exceeds it, i.e. independently from garbage bytes. The first match which reaches the end of the data is already as long as possible, so we exit early.

Dead2:
@QrczakMK That is some real good detective work, I am impressed :100: 
Given your explanation, I do agree that this might very well be the fix we need.

We'll have to do some testing of course, but it makes sense and I agree with that @dzetkulic that resetting those garbage bytes might not be needed and might even help us catch match regressions in the future.

QrczakMK:
As for testing, we use zlib-rs in many places in Google. It has been originally transcribed from zlib-ng and has the same issue. I tested the analogous fix to zlib-rs, and all our tests passed.

Dead2 (approved):
All of my tests pass.
Benchmarking-wise, this has negligible difference. It does look like it is very slightly faster on average (I'd expect so from sometimes not having to look for another match at end of window) over dozens of gbench runs, but benchmarks are too close to really tell for sure.

test/test_deflate_reset_deterministic.cc
```diff
@@ -0,0 +1,381 @@
+/* test_deflate_reset_deterministic.cc - Test deterministic output after deflateReset */
+
+#include "zbuild.h"
+#ifdef ZLIB_COMPAT
+#  include "zlib.h"
+#else
+#  include "zlib-ng.h"
+#endif
+
+#include <stdio.h>
```
Dead2: stdio.h, stdint.h, stdlib.h and string.h are already included from zbuild.h so doesn't need to be included here. string is not though.

QrczakMK:
`compress_data()` is likely more complicated than it deserves.

It evolved from a particular wrapper for streaming compression.  For the curious: https://github.com/google/riegeli/blob/master/riegeli/zlib/zlib_writer.cc

Splitting input and output into buffers does not impact the outcome. For simplicity, it might use some appropriate string-to-string one-shot compressor, as long as the `z_stream` is being reused.

test/test_deflate_reset_deterministic.cc
mtl1979: Often the header names without `.h` are just wrappers around the header with the `.h` extension. This is because C++ headers should essentially contain same functions and types and symbol mangling can be suppressed just for C++ compilers. Including both variants is generally unnecessary. Exception is the namespace `std` which is unnecessary for C, but separates standard functions in own namespace for C++.

mtl1979: `string` is C++ header, so it makes no sense to have it in `zbuild.h` which should only contain C headers.
C++ equivalent of `string.h` is `cstring`.

Dead2: @mtl1979 I didn't mean that we need to add it to `zbuild.h`, I meant the ones that are already in `zbuild.h` should be removed from the test source, as they don't need to be parsed twice.
I don't know whether `string` itself is needed by this test or not.

phprus: Rebased, headers removed.

`string` is used for the output buffer.
Manual memory management is not needed, and the `EXPECT_EQ` macro can be used to compare values. If the values ​​don't match, gtest will print them.

mtl1979: > @mtl1979 I didn't mean that we need to add it to `zbuild.h`, I meant the ones that are already in `zbuild.h` should be removed from the test source, as they don't need to be parsed twice.
> 
> I don't know whether `string` itself is needed by this test or not.

Most of the C++ headers are designed to be included before the equivalent C header as they include the C header after defining some preprocessor defines. For a lot of standard C headers the C++ equivalent name is prefixed with `c` and the extension is omitted.

nmoinvaz:
Can we shorten the filename of the test.. maybe something like dtrmin or whatever?

phprus:
> Can we shorten the filename of the test.. maybe something like dtrmin or whatever?

Maybe `test_deflate_deterministic.cc`?

QrczakMK:
In https://github.com/trifectatechfoundation/zlib-rs/pull/461 we might find simpler test data for triggering this.

folkertdev:
I was able to manually reduce it to

```rust
    const DATA_A: &[u8] = b"\0AAAA\0AAAAAAAA";
    const DATA_B: &[u8] = &[1u8; DATA_A.len() + 1];
```

phprus:
Rebased.
Simpler test data was used.

test/test_deflate_reset_deterministic.cc
```diff
@@ -0,0 +1,81 @@
...
+
+#include "test_shared.h"
+
+#include <gtest/gtest.h>
+
+
+/* Issue: https://github.com/zlib-ng/zlib-ng/issues/2100 */
+
+static const char data_a[] = "\0AAAA\0AAAAAAAA";
+static const char data_b[] = "\1\0AAAA\0AAAAAAAA";
```
QrczakMK: I think it would be cleaner if this consisted of `\1` bytes or `B` bytes, with the same length (longer than `data_a`).

It does not need to have anything in common with `data_a`. It is meant to ensure that `data_a`, written on top of `data_b`, is not followed by `\0` (nor anything else which would interfere with matching).

phprus: Test data:
```c++
/* sizeof(data_b) must be greater sizeof(data_a) */
static const uint8_t data_a[] = {  0 , 'A', 'A', 'A', 'A',  0 , 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A' };
static const uint8_t data_b[] = { '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1' };
```

Error:
```
[ RUN      ] deflate.reset_deterministic
/Users/phprus/Devel/phprus/zlib-ng/test/test_deflate_reset_deterministic.cc:81: Failure
Expected equality of these values:
  a_compressed
    Which is: "x\x9C" "cptttd\0\x11 \x6\0\x14\xE0\x3\r"
  a_compressed2
    Which is: "x\x9C" "cptttd\0\x11 \f\0\x14\xE0\x3\r"

[  FAILED  ] deflate.reset_deterministic (0 ms)
```

After apply fix:
```
[ RUN      ] deflate.reset_deterministic
[       OK ] deflate.reset_deterministic (0 ms)
```

test/test_deflate_reset_deterministic.cc
```diff
@@ -0,0 +1,82 @@
...
+#include "test_shared.h"
+
+#include <gtest/gtest.h>
+
+
+/* Issue: https://github.com/zlib-ng/zlib-ng/issues/2100 */
+
+/* len(data_b) must be greater len(data_a) */
+static const uint8_t data_a[] = {  0 , 'A', 'A', 'A', 'A',  0 , 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A' };
+static const uint8_t data_b[] = { '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1' };
```
QrczakMK: (I meant numeric `1` or `'B'` but this does not really matter now.)

test/test_deflate_deterministic.cc
```diff
@@ -0,0 +1,82 @@
...
+#include <gtest/gtest.h>
+
+
+/* Issue: https://github.com/zlib-ng/zlib-ng/issues/2100 */
+
+/* len(data_b) must be greater len(data_a) */
+static const uint8_t data_a[] = {  0 , 'A', 'A', 'A', 'A',  0 , 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A' };
+static const uint8_t data_b[] = { '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1', '1' };
+
+static std::string compress_data(const uint8_t *src, size_t len, PREFIX3(stream)* z_stream) {
```
QrczakMK, line 24: I am pretty sure this function can be simplified to some one-shot string-to-string compression. Splitting the data into buffers does not influence the behavior, only reusing the `z_stream` is important.

This implementation was derived from a streaming compression. Given that we have about 20 bytes, a 4KB buffer for compressed is somewhat excessive, `avail_in > 0xFFFFFFFF` never applies, and if the output space is sufficient then no loop is needed. I guess you have something handy in other tests.

mtl1979: Any variable smaller than 16 kB on stack doesn't really waste memory unless the total size of all stack variables exceed 16 kB. However it's a lot cleaner if the buffer_size is as small as possible without causing potential overflow. Any dead code should obviously be removed if function can't be called from other translation units.

QrczakMK (approved):

test/CMakeLists.txt
```diff
@@ -184,6 +184,7 @@ if(WITH_GTEST)
                 test_compare256_rle.cc      # compare256_rle(), etc
                 test_crc32.cc               # crc32_armv8(), etc
                 test_crc32_copy.cc          # crc32_copy implementations
+                test_deflate_reset_deterministic.cc # deterministic output after deflateReset
```
nmoinvaz: Please align all #.

nmoinvaz (commented):
> Maybe test_deflate_deterministic.cc?

That would be fine to me. Don't forget to update the filename in the .cc header comment.

phprus:
> That would be fine to me. Don't forget to update the filename in the .cc header comment.

Renamed to `test_deflate_deterministic.cc`.

test/test_deflate_deterministic.cc
```diff
@@ -0,0 +1,72 @@
...
+    uint8_t buffer[buffer_size];
+    int err;
+
+    z_stream->next_in = (z_const uint8_t *)src;
+    z_stream->avail_in = (uint32_t)len;
+    z_stream->next_out = buffer;
+    z_stream->avail_out = buffer_size;
+    err = PREFIX(deflate)(z_stream, Z_FINISH);
+    EXPECT_EQ(err, Z_STREAM_END);
+    return std::string((const char *)buffer, (size_t)(z_stream->next_out - buffer));
```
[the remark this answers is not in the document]
mtl1979: Early return is unnecessary, any corruption or truncation would already make the string comparison fail. This suggestion is pointless.

test/test_deflate_deterministic.cc
```diff
@@ -0,0 +1,72 @@
...
+#  include "zlib-ng.h"
+#endif
+
+#include <string>
+
+#include "deflate.h"
+
+#include "test_shared.h"
+
+#include <gtest/gtest.h>
```
mtl1979: These should be grouped with all <> together and after them all "" ones. Having empty line to separate every single include seems overkill...

phprus: Done.

mtl1979 (approved):
LGTM.

test/test_deflate_deterministic.cc
```diff
@@ -0,0 +1,71 @@
...
+
+
+
+/* Issue: https://github.com/zlib-ng/zlib-ng/issues/2100 */
+
+/* len(data_b) must be greater len(data_a) */
+static const uint8_t data_a[] = {  0 , 'A', 'A', 'A', 'A',  0 , 'A', 'A', 'A', 'A', 'A', 'A', 'A', 'A' };
+static const uint8_t data_b[] = { 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B', 'B' };
+
+static std::string compress_data(const uint8_t *src, size_t len, PREFIX3(stream)* z_stream) {
```
nmoinvaz, line 24: I wonder why you don't use the compress2 function instead of compress_data

phprus: We need to use the `PREFIX3(stream)` structure and call `PREFIX(deflateReset)`.

nmoinvaz: Reuse you mean. Got it. Didn't see stream was passed in as an arg. Thanks! 🙏🏻

nmoinvaz (approved):

folkertdev:
Do you have any indication on when this fix will make it into a release?

Dead2:
@folkertdev I can't promise anything, but I am hoping to backport this and a few other fixes and make a release later this week.

PR #2103, opened 2026-01-14
Add misalignment to benchmark buffers

Dead2:
- Unify benchmark code.
- Add misalignment to buffers (might be changed)

TODO: Further research into expected real-world alignments

---
As can be seen below, currently 16/32/64 byte buffers are all faster than 8 and 12 byte buffers.
While this is accurate, it is only true for 16-byte aligned input.

Before:
```
crc32/pclmulqdq/1                   2.08 ns         2.08 ns   1005886062
crc32/pclmulqdq/8                   13.4 ns         13.4 ns    156492829
crc32/pclmulqdq/12                  19.9 ns         19.9 ns    104951835
crc32/pclmulqdq/16                  6.83 ns         6.83 ns    307242360
crc32/pclmulqdq/32                  7.74 ns         7.74 ns    273097404
crc32/pclmulqdq/64                  9.95 ns         9.95 ns    213542134
crc32/pclmulqdq/512                 31.0 ns         31.0 ns     67606989
crc32/pclmulqdq/4096                 179 ns          179 ns     11667196
crc32/pclmulqdq/32768               1406 ns         1406 ns      1519562
crc32/pclmulqdq/262144             11055 ns        11055 ns       185596
crc32/pclmulqdq/4194304           179210 ns       179209 ns        11699
```

After:
```
crc32/pclmulqdq/1                   2.06 ns         2.06 ns   1019868718
crc32/pclmulqdq/8                   13.4 ns         13.4 ns    156972088
crc32/pclmulqdq/12                  19.9 ns         19.9 ns    105753664
crc32/pclmulqdq/16                  25.0 ns         25.0 ns     83933752
crc32/pclmulqdq/32                  25.3 ns         25.3 ns     82825755
crc32/pclmulqdq/64                  26.2 ns         26.2 ns     80324722
crc32/pclmulqdq/512                 41.2 ns         41.2 ns     51184838
crc32/pclmulqdq/4096                 193 ns          193 ns     11049203
crc32/pclmulqdq/32768               1396 ns         1396 ns      1342525
crc32/pclmulqdq/262144             10872 ns        10872 ns       193004
crc32/pclmulqdq/4194304           175952 ns       175950 ns        11920
```

### DIFF
--- a/test/benchmarks/benchmark_adler32.cc
+++ b/test/benchmarks/benchmark_adler32.cc
@@ -3,65 +3,79 @@
  * For conditions of distribution and use, see copyright notice in zlib.h
  */
 
-#include <stdio.h>
-#include <assert.h>
-
 #include <benchmark/benchmark.h>
 
 extern "C" {
 #  include "zbuild.h"
-#  include "zutil_p.h"
 #  include "arch_functions.h"
 #  include "../test_cpu_features.h"
 }
 
-#define MAX_RANDOM_INTS (1024 * 1024)
-#define MAX_RANDOM_INTS_SIZE (MAX_RANDOM_INTS * sizeof(uint32_t))
+#define BUFSIZE ((4 * 1024 * 1024) + 64)
 
 class adler32: public benchmark::Fixture {
 private:
-    uint32_t *random_ints;
+    uint32_t *testdata;
 
 public:
-    void SetUp(const ::benchmark::State&) {
-        /* Control the alignment so that we have the best case scenario for loads. With
-         * AVX512, unaligned loads can mean we're crossing a cacheline boundary at every load.
-         * And while this is a realistic scenario, it makes it difficult to compare benchmark
-         * to benchmark because one allocation could have been aligned perfectly for the loads
-         * while the subsequent one happened to not be. This is not to be advantageous to AVX512
-         * (indeed, all lesser SIMD implementations benefit from this aligned allocation), but to
-         * control the _consistency_ of the results */
-        random_ints = (uint32_t *)zng_alloc(MAX_RANDOM_INTS_SIZE);
-        assert(random_ints != NULL);
+    void SetUp(::benchmark::State& state) {
+        testdata = (uint32_t *)zng_alloc_aligned(BUFSIZE, 64);
+        if (testdata == NULL) {
+            state.SkipWithError("malloc failed");
+            return;
+        }
 
-        for (int32_t i = 0; i < MAX_RANDOM_INTS; i++) {
-            random_ints[i] = rand();
+        for (uint32_t i = 0; i < BUFSIZE/sizeof(uint32_t); i++) {
+            testdata[i] = rand();
         }
     }
 
-    void Bench(benchmark::State& state, adler32_func adler32) {
+    // Benchmark Adler32, with rolling buffer misalignment for consistent results
+    void Bench(benchmark::State& state, adler32_func adler32, const int DO_ALIGNED) {
+        int misalign = 0;
         uint32_t hash = 0;
 
         for (auto _ : state) {
-            hash = adler32(hash, (const unsigned char *)random_ints, (size_t)state.range(0));
+            hash = adler32(hash, (const unsigned char*)testdata + misalign, (size_t)state.range(0));
+            if (misalign >= 63)
+                misalign = 0;
+            else
+                misalign += (DO_ALIGNED) ? 16 : 1;
         }
 
+        // Prevent the result from being optimized away
         benchmark::DoNotOptimize(hash);
     }
 
     void TearDown(const ::benchmark::State&) {
-        zng_free(random_ints);
+        zng_free_aligned(testdata);
     }
 };
 
-#define BENCHMARK_ADLER32(name, fptr, support_flag) \
+#define BENCHMARK_ADLER32_MISALIGNED(name, hashfunc, support_flag) \
     BENCHMARK_DEFINE_F(adler32, name)(benchmark::State& state) { \
         if (!(support_flag)) { \
             state.SkipWithError("CPU does not support " #name); \
         } \
-        Bench(state, fptr); \
+        Bench(state, hashfunc, 0); \
     } \
-    BENCHMARK_REGISTER_F(adler32, name)->Arg(1)->Arg(8)->Arg(12)->Arg(16)->Arg(32)->Arg(64)->Arg(512)->Arg(4<<10)->Arg(32<<10)->Arg(256<<10)->Arg(4096<<10)
+    BENCHMARK_REGISTER_F(adler32, name)->Arg(1)->Arg(8)->Arg(12)->Arg(16)->Arg(32)->Arg(64)->Arg(512)->Arg(4<<10)->Arg(32<<10)->Arg(256<<10)->Arg(4096<<10);
+
+// Aligned
+#define ALIGNED_NAME(name) name##_aligned
+#define BENCHMARK_ADLER32_ALIGNED(name, hashfunc, support_flag) \
+    BENCHMARK_DEFINE_F(adler32, ALIGNED_NAME(name))(benchmark::State& state) { \
+        if (!(support_flag)) { \
+            state.SkipWithError("CPU does not support " #name); \
+        } \
+        Bench(state, hashfunc, 1); \
+    } \
+    BENCHMARK_REGISTER_F(adler32, ALIGNED_NAME(name))->Arg(8)->Arg(12)->Arg(16)->Arg(32)->Arg(64)->Arg(512)->Arg(4<<10)->Arg(32<<10)->Arg(256<<10)->Arg(4096<<10);
+
+// Queue both misaligned and aligned for each benchmark
+#define BENCHMARK_ADLER32(name, hashfunc, support_flag) \
+    BENCHMARK_ADLER32_MISALIGNED(name, hashfunc, support_flag); \
+    BENCHMARK_ADLER32_ALIGNED(name, hashfunc, support_flag);
 
 BENCHMARK_ADLER32(c, adler32_c, 1);
 
--- a/test/benchmarks/benchmark_adler32_copy.cc
+++ b/test/benchmarks/benchmark_adler32_copy.cc
@@ -3,75 +3,88 @@
  * For conditions of distribution and use, see copyright notice in zlib.h
  */
 
-#include <stdio.h>
-#include <assert.h>
-#include <string.h>
-
 #include <benchmark/benchmark.h>
 
 extern "C" {
 #  include "zbuild.h"
-#  include "zutil_p.h"
 #  include "arch_functions.h"
 #  include "../test_cpu_features.h"
 }
 
-#define MAX_RANDOM_INTS (1024 * 1024)
-#define MAX_RANDOM_INTS_SIZE (MAX_RANDOM_INTS * sizeof(uint32_t))
-
-typedef uint32_t (*adler32_cpy_func)(uint32_t adler, unsigned char *dst, const uint8_t *buf, size_t len);
+// Hash copy functions are used on strm->next_in buffers, we process
+// 512-32k sizes (x2 for initial fill) at a time if enough data is available.
+#define BUFSIZE (65536 + 64)
 
 class adler32_copy: public benchmark::Fixture {
 private:
-    uint32_t *random_ints_src;
-    uint32_t *random_ints_dst;
+    uint32_t *testdata;
+    uint8_t *dstbuf;
 
 public:
-    void SetUp(const ::benchmark::State&) {
-        /* Control the alignment so that we have the best case scenario for loads. With
-         * AVX512, unaligned loads can mean we're crossing a cacheline boundary at every load.
-         * And while this is a realistic scenario, it makes it difficult to compare benchmark
-         * to benchmark because one allocation could have been aligned perfectly for the loads
-         * while the subsequent one happened to not be. This is not to be advantageous to AVX512
-         * (indeed, all lesser SIMD implementations benefit from this aligned allocation), but to
-         * control the _consistency_ of the results */
-        random_ints_src = (uint32_t *)zng_alloc(MAX_RANDOM_INTS_SIZE);
-        random_ints_dst = (uint32_t *)zng_alloc(MAX_RANDOM_INTS_SIZE);
-        assert(random_ints_src != NULL);
-        assert(random_ints_dst != NULL);
+    void SetUp(::benchmark::State& state) {
+        testdata = (uint32_t *)zng_alloc_aligned(BUFSIZE, 64);
+        dstbuf = (uint8_t *)zng_alloc_aligned(BUFSIZE, 64);
+        if (testdata == NULL || dstbuf == NULL) {
+            state.SkipWithError("malloc failed");
+            return;
+        }
 
-        for (int32_t i = 0; i < MAX_RANDOM_INTS; i++) {
-            random_ints_src[i] = rand();
+        for (uint32_t i = 0; i < BUFSIZE/sizeof(uint32_t); i++) {
+            testdata[i] = rand();
         }
     }
 
-    void Bench(benchmark::State& state, adler32_cpy_func adler32_func) {
+    // Benchmark Adler32_copy, with rolling buffer misalignment for consistent results
+    void Bench(benchmark::State& state, adler32_copy_func adler32_copy, const int DO_ALIGNED) {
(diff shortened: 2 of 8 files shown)


mtl1979:
Especially on big-endian platforms (f.ex. 32-bit PowerPC) where unaligned access is not allowed, testing with misalignment makes no sense as the real code is skipped when the length is short and misalignment is almost as big as the iteration length. In most cases the buffer is already guaranteed to be unaligned and length is not multiple of 16 or 64 bytes. This is also why some custom allocators don't work on 32-bit PowerPC as you can't just add sizeof(void*) to allocation length.

Dead2:
> Especially on big-endian platforms (f.ex. 32-bit PowerPC) where unaligned access is not allowed, testing with misalignment makes no sense as the real code is skipped when the length is short and misalignment is almost as big as the iteration length. In most cases the buffer is already guaranteed to be unaligned and length is not multiple of 16 or 64 bytes. This is also why some custom allocators don't work on 32-bit PowerPC as you can't just add sizeof(void*) to allocation length.

I'm not following you. I don't see how it is not useful to benchmark misaligned if real-world access is misaligned..

mtl1979:
> > Especially on big-endian platforms (f.ex. 32-bit PowerPC) where unaligned access is not allowed, testing with misalignment makes no sense as the real code is skipped when the length is short and misalignment is almost as big as the iteration length. In most cases the buffer is already guaranteed to be unaligned and length is not multiple of 16 or 64 bytes. This is also why some custom allocators don't work on 32-bit PowerPC as you can't just add sizeof(void*) to allocation length.
> 
> I'm not following you. I don't see how it is not useful to benchmark misaligned if real-world access is misaligned..

When benchmarking this PR, the adler32_vmx code was 4 times as slow as the scalar version. Before the patch, scalar and VMX versions were almost the same speed. Misalignment of 63 bytes is overkill when maximum misalignment is 15.

Dead2:
> > I'm not following you. I don't see how it is not useful to benchmark misaligned if real-world access is misaligned..
> 
> When benchmarking this PR, the adler32_vmx code was 4 times as slow as the scalar version. Before the patch, scalar and VMX versions were almost the same speed. Misalignment of 63 bytes is overkill when maximum misalignment is 15.

64 bytes misalignment rotation is because that is normally the size of a cacheline, so it goes through all alignment variants in relation to a cacheline. 16-byte alignment is enough for vector instructions, but that would still leave the cacheline-lottery effect from benchmark to benchmark. AFAIK 64 bytes is the biggest we should really care about in this regard. There is of course page-alignment as well, but I think I'll just consider that overkill for now, although it would be interesting to see what effects it gives as it could then also avoid using one extra entry in the TLB-buffer.

But I am not sure how it is not useful to know that adler32_vmx can be slower than the scalar function when the input buffer is unaligned. What benefit can we get from hiding/ignoring that?

mtl1979:
> > > I'm not following you. I don't see how it is not useful to benchmark misaligned if real-world access is misaligned..
> > 
> > 
> > When benchmarking this PR, the adler32_vmx code was 4 times as slow as the scalar version. Before the patch, scalar and VMX versions were almost the same speed. Misalignment of 63 bytes is overkill when maximum misalignment is 15.
> 
> 64 bytes misalignment rotation is because that is normally the size of a cacheline, so it goes through all alignment variants in relation to a cacheline. 16-byte alignment is enough for vector instructions, but that would still leave the cacheline-lottery effect from benchmark to benchmark. AFAIK 64 bytes is the biggest we should really care about in this regard. There is of course page-alignment as well, but I think I'll just consider that overkill for now, although it would be interesting to see what effects it gives as it could then also avoid using one extra entry in the TLB-buffer.
> 
> But I am not sure how it is not useful to know that adler32_vmx can be slower than the scalar function when the input buffer is unaligned. What benefit can we get from hiding/ignoring that?

With adler32_vmx specifically the penalty is quite big until about 32 kB, where the initial penalty of setting up the vectors get compensated. It's basically hitting misprediction in both head and tail handling with minimal iterations in inner vectorized loop. When you cross cachelines on 32-bit PowerPC, it will cause access violation resulting in context switch from userspace to kernel and back, and that is painfully slow on processor running only on 33 MHz clock... This is also why memcpy() and memset() should not be used on early 32-bit PowerPC processors with VMX.

Dead2:
> With adler32_vmx specifically the penalty is quite big until about 32 kB, where the initial penalty of setting up the vectors get compensated. It's basically hitting misprediction in both head and tail handling with minimal iterations in inner vectorized loop. When you cross cachelines on 32-bit PowerPC, it will cause access violation resulting in context switch from userspace to kernel and back, and that is painfully slow on processor running only on 33 MHz clock... This is also why memcpy() and memset() should not be used on early 32-bit PowerPC processors with VMX.

Interesting, that might mean that it should not be used unless lengths are really long, and we are already limited by the 32k window size so we might not actually see a benefit that often.

nmoinvaz:
> Interesting, that might mean that it should not be used unless lengths are really long, and we are already limited by the 32k window size so we might not actually see a benefit that often.

@mtl1979 do you we should add something like what Chorba does where if the len is more than a certain amount, only then it does `_vmx`? Maybe that is something you could investigate?

test/benchmarks/benchmark_adler32.cc
```diff
@@ -32,35 +32,61 @@ class adler32: public benchmark::Fixture {
...
         int misalign = 0;
         uint32_t hash = 0;
 
-        for (auto _ : state) {
-            hash = adler32(hash, (const unsigned char *)testdata + misalign, (size_t)state.range(0));
-            if (misalign >= 63)
-                misalign = 0;
-            else
-                misalign++;
+        if (do_aligned) {
```
nmoinvaz: Nice idea! I suggest for variables where the compiler is expected to optimize out a certain path, we do all caps. `DO_ALIGNED` and `const int`.

test/benchmarks/benchmark_adler32.cc
```diff
@@ -3,66 +3,91 @@
...
+            state.SkipWithError("CPU does not support " #name); \
+        } \
+        Bench(state, adlerfunc, 1); \
+    } \
+    BENCHMARK_REGISTER_F(adler32, ALIGNED_NAME(name))->Arg(8)->Arg(12)->Arg(16)->Arg(32)->Arg(64)->Arg(512)->Arg(4<<10);
+
+// Queue both misaligned and aligned for each benchmark
+#define BENCHMARK_ADLER32(name, copyfunc, support_flag) \
+    BENCHMARK_ADLER32_MISALIGNED(name, copyfunc, support_flag); \
+    BENCHMARK_ADLER32_ALIGNED(name, copyfunc, support_flag);
```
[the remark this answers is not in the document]
nmoinvaz: I suggest just `unaligned` vs `misaligned` like we have used in many places in the code.

nmoinvaz (commented):
Overall I think this PR is nice improvement. It should allow us to get a better picture into performance.

test/benchmarks/benchmark_adler32.cc
Dead2: I thought about that but I was not sure how the gbench templating works, does it actually make two separate functions for this? If so, then yes that'd be a good idea..

test/benchmarks/benchmark_adler32.cc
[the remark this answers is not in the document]
Dead2: Well, I kind of wanted to avoid that since we are intentionally misaligning the buffer for every iteration. Most places we handle unaligned inputs, but often that means it is likely to be aligned just that we need to also handle the cases where it is not.
While here we are actively misaligning by bruteforce for testing, not just passing it a regular malloc() buffer that has no guarantee of being aligned or not.

So that is the reasoning behind the naming, I wanted it to stand out specifically because it is not just benchmarking using a buffer that is not guaranteed to be aligned, it is actively not aligned.

I also considered rotaligned, but that'll just cause confusion I think.

test/benchmarks/benchmark_compare256_rle.cc
```diff
@@ -3,59 +3,64 @@
...
         str2[match_len] = 0;
         for (auto _ : state) {
-            len = compare256_rle((const uint8_t *)str1, (const uint8_t *)str2);
+            len = compare256_rle((const uint8_t *)str1 + misalign, (const uint8_t *)str2 + misalign);
+            if (misalign >= 63)
+                misalign = 0;
+            else
+                misalign++;
         }
         str2[match_len] = 'a';
```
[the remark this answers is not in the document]
Dead2: This is well observed. Will fix.

test/benchmarks/benchmark_adler32.cc
Dead2: Changed.

test/benchmarks/benchmark_adler32_copy.cc
```diff
@@ -3,137 +3,184 @@
...
                         const uint8_t *buf, size_t len) -> uint32_t { \
             memcpy(dst, buf, (size_t)len); \
-            return fptr(init_sum, buf, len); \
-        }); \
+            return hashfunc(init_sum, buf, len); \
+        }, 1); \
     } \
-    BENCHMARK_REGISTER_F(adler32_copy, name)->Range(8192, MAX_RANDOM_INTS_SIZE);
+    BENCHMARK_REGISTER_F(adler32_copy, MEMCPY_ALIGNED_NAME(name))->Arg(32)->Arg(512)->Arg(8<<10)->Arg(32<<10)->Arg(64<<10);
+#endif
```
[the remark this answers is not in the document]
Dead2: It actually does not execute it when it is set to skip it here.

test/benchmarks/benchmark_compare256_rle.cc
```diff
@@ -3,59 +3,67 @@
...
+
+            len = compare256_rle((const uint8_t *)str1 + misalign, (const uint8_t *)str2 + misalign);
+
+            str2[match_len + misalign] = 'a'; // Reset match limit
+
+            if (misalign >= 63)
+                misalign = 0;
+            else
+                misalign++;
         }
```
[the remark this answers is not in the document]
Dead2: The buffer is 256 bytes + 64bytes to allow for misalignment.
The biggest comparison being done is 256bytes (as is the biggest possible for the compare256 function).
`BENCHMARK_REGISTER_F(compare256, name)->Arg(1)->Arg(10)->Arg(40)->Arg(80)->Arg(100)->Arg(175)->Arg(256);`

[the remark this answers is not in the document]
Dead2: Ah, fix incoming.

mtl1979:
> @mtl1979 do you we should add something like what Chorba does where if the len is more than a certain amount, only then it does `_vmx`? Maybe that is something you could investigate?

The original code had only 32 byte (2 vector) inner loop which worked with first generation processors, I tested with 16 byte (1 vector) inner loop too but it was slower than scalar. I haven't yet tested which one of the intrinsics is the bottleneck, but there really isn't many possibilities. Original code did actually mix scalar and VMX to avoid intrinsics that might be slow. There is no issue with VMX in slide_window.

test/benchmarks/benchmark_adler32.cc
```diff
@@ -3,65 +3,89 @@
...
 
-        for (auto _ : state) {
-            hash = adler32(hash, (const unsigned char *)random_ints, (size_t)state.range(0));
+        if (DO_ALIGNED) {
+            for (auto _ : state) {
+                hash = adler32(hash, (const unsigned char*)testdata + misalign, (size_t)state.range(0));
+                if (misalign >= 63)
+                    misalign = 0;
+                else
+                    misalign+=16;
```
nmoinvaz: Spaces between +=

test/benchmarks/benchmark_adler32_copy.cc
```diff
@@ -3,137 +3,184 @@
...
-        random_ints_dst = (uint32_t *)zng_alloc(MAX_RANDOM_INTS_SIZE);
-        assert(random_ints_src != NULL);
-        assert(random_ints_dst != NULL);
-
-        for (int32_t i = 0; i < MAX_RANDOM_INTS; i++) {
-            random_ints_src[i] = rand();
+    void SetUp(::benchmark::State& state) {
+        testdata = (uint32_t *)zng_alloc_aligned(BUFSIZE, 64);
+        dstbuf = (uint8_t *)zng_alloc_aligned(BUFSIZE, 64);
+        if(testdata == NULL || dstbuf == NULL) {
```
nmoinvaz: Space after if

test/benchmarks/benchmark_adler32.cc
```diff
@@ -3,65 +3,89 @@
...
-         * (indeed, all lesser SIMD implementations benefit from this aligned allocation), but to
-         * control the _consistency_ of the results */
-        random_ints = (uint32_t *)zng_alloc(MAX_RANDOM_INTS_SIZE);
-        assert(random_ints != NULL);
-
-        for (int32_t i = 0; i < MAX_RANDOM_INTS; i++) {
-            random_ints[i] = rand();
+    void SetUp(::benchmark::State& state) {
+        testdata = (uint32_t *)zng_alloc_aligned(BUFSIZE, 64);
+        if(testdata == NULL) {
```
nmoinvaz: Space after if

test/benchmarks/benchmark_adler32_copy.cc
```diff
@@ -3,137 +3,184 @@
...
-        for (auto _ : state) {
-            hash = adler32_func(hash, (unsigned char *)random_ints_dst,
-                                (const unsigned char*)random_ints_src, (size_t)state.range(0));
+        if (DO_ALIGNED) {
+            for (auto _ : state) {
+                hash = adler32_copy(hash, dstbuf + misalign, (const unsigned char*)testdata + misalign, (size_t)state.range(0));
+                if (misalign >= 63)
+                    misalign = 0;
+                else
+                    misalign+=16;
```
nmoinvaz: Space around +=

test/benchmarks/benchmark_adler32_copy.cc
```diff
@@ -3,137 +3,184 @@
...
-    void Bench(benchmark::State& state, adler32_cpy_func adler32_func) {
+    // Benchmark Adler32_copy, with rolling buffer misalignment for consistent results
+    void Bench(benchmark::State& state, adler32_copy_func adler32_copy, const int DO_ALIGNED) {
+        int misalign = 0;
         uint32_t hash = 0;
 
-        for (auto _ : state) {
-            hash = adler32_func(hash, (unsigned char *)random_ints_dst,
-                                (const unsigned char*)random_ints_src, (size_t)state.range(0));
+        if (DO_ALIGNED) {
```
nmoinvaz: misaligned += (DO_ALIGNED) ? 16 : 1.

test/benchmarks/benchmark_compare256.cc
```diff
@@ -3,61 +3,70 @@
...
 
 public:
-    void SetUp(const ::benchmark::State&) {
-        str1 = (uint8_t *)zng_alloc(MAX_COMPARE_SIZE);
-        assert(str1 != NULL);
-        memset(str1, 'a', MAX_COMPARE_SIZE);
+    void SetUp(::benchmark::State& state) {
+        str1 = (uint8_t *)malloc(MAX_COMPARE_SIZE);
+        str2 = (uint8_t *)malloc(MAX_COMPARE_SIZE);
+        if(str1 == NULL || str2 == NULL) {
```
nmoinvaz: Space after if

test/benchmarks/benchmark_compare256_rle.cc
```diff
@@ -3,61 +3,70 @@
...
 
 public:
-    void SetUp(const ::benchmark::State&) {
-        str1 = (uint8_t *)zng_alloc(MAX_COMPARE_SIZE);
-        assert(str1 != NULL);
-        memset(str1, 'a', MAX_COMPARE_SIZE);
+    void SetUp(::benchmark::State& state) {
+        str1 = (uint8_t *)malloc(MAX_COMPARE_SIZE);
+        str2 = (uint8_t *)malloc(MAX_COMPARE_SIZE);
+        if(str1 == NULL || str2 == NULL) {
```
nmoinvaz: Space after if

nmoinvaz (commented):
I've haven't mentioned all spaces issue but enough to get the idea. It looks like it needs a bit of formatting still.

test/benchmarks/benchmark_adler32_copy.cc
Dead2: Clever :+1:

Issue #2109, opened 2026-01-17
[Bug] Integer Overflow in `minigzip`


ShangzhiXu:
Hi team,

 Happy new year~

I think I found a small bug in `minigzip`. It occurs in `gz_compress_mmap` at `line 87, minigzip.c`

```cpp
 static int gz_compress_mmap(FILE *in, gzFile out) {
     int len;
     int ifd = fileno(in);
     char *buf;    
     off_t buf_len;  
     struct stat sb;
 
     if (fstat(ifd, &sb) < 0) return Z_ERRNO;
     buf_len = sb.st_size;
    ...
     
    len = PREFIX(gzwrite)(out, buf, (unsigned)buf_len); // crash point
```

Here, `buf_len = sb.st_size` is the file size controlled by user, and it is of singed type `off_t`, but casted to unsigned in the following code, this leads to integer overflow if `sb.st_size` is too large, for example `4294967306` bytes (4GB + 10). 

While this may not lead to a crash, it might causes silent data loss, only the truncated length is written to the compressed output.

### PoC
```cpp
// We first make a large file
z5500277@katana1:.../fuz/zlib-ng $ echo -n "START_MARKER" > poc.img
z5500277@katana1:.../fuz/zlib-ng $ dd if=/dev/urandom of=poc.img bs=1 count=10 seek=4294967296 conv=notrunc 2>/dev/null
z5500277@katana1:.../fuz/zlib-ng $ ls -l poc.img
-rw-r--r--. 1 z5500277 unsw 4294967306 Jan 17 19:45 poc.img
z5500277@katana1:.../fuz/zlib-ng $ ls -lh poc.img
-rw-r--r--. 1 z5500277 unsw 4.1G Jan 17 19:45 poc.img


// then compress and decompress
 
z5500277@katana1:.../fuz/zlib-ng $ ./build-afl/minigzip poc.img
z5500277@katana1:.../fuz/zlib-ng $ ls -lh poc.img.gz
-rw-r--r--. 1 z5500277 unsw 30 Jan 17 19:46 poc.img.gz
z5500277@katana1:.../fuz/zlib-ng $ gzip -d poc.img.gz
z5500277@katana1:.../fuz/zlib-ng $ ls -lh poc.img
-rw-r--r--. 1 z5500277 unsw 10 Jan 17 19:46 poc.img
```

As we can see, after decompression, only the last 10 bytes remain.
[...]

PR #2110, opened 2026-01-17
Fix integer overflow in gz_compress_mmap

phprus:
Fix for issue #2109

Added a 2GB (`INT_MAX`) mmap I/O limit.

### DIFF
--- a/test/fuzz/fuzzer_minigzip.c
+++ b/test/fuzz/fuzzer_minigzip.c
@@ -70,26 +70,27 @@ static void error(const char *msg) {
  * success, Z_ERRNO otherwise.
  */
 static int gz_compress_mmap(FILE *in, gzFile out) {
-    int len;
     int err;
     int ifd = fileno(in);
-    char *buf;      /* mmap'ed buffer for the entire input file */
-    off_t buf_len;  /* length of the input file */
+    void *buf;      /* mmap'ed buffer for the entire input file */
+    size_t buf_len; /* length of the input file */
+    size_t len;
     struct stat sb;
 
     /* Determine the size of the file, needed for mmap: */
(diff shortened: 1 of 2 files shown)


mtl1979:
Like I mentioned in the issue, gzip file format doesn't support files that are 2 GB or larger due to file systems using signed 32-bit integers for file size in `stat` structure on 32-bit operating systems without large file support.

`gzip` silently truncates file size after checksum in the end of the stream to modulo 2^32. Adding support for larger files would require using one of the reserved bits in the gzip header to signify new file format version that increase the size field to 8 bytes. This obviously breaks compatibility with older versions of both `gzip` and `minigzip`.

It's debatable if we should gracefully fail both when using mmap and when not using.

CodeCov failure is due to not being in sync with current development tree, currently 19 commits behind.

nmoinvaz (commented):
I think what we should be using is:
```
// Instead of:
len = PREFIX(gzwrite)(out, (char *)buf, (unsigned)buf_len);

// Use:
len = PREFIX(gzfwrite)((char *)buf, 1, buf_len, out);
```
f prefix doesn't imply `FILE*` rather it was part of a set of changes that utilized `size_t`.

phprus:
@mtl1979

You're right.
But I think work compression and truncated file size in gzip trailer is better than unexpected reading just part of the input file.

Code without MMAP is also susceptible to the gzip format issue.

@nmoinvaz 

I changed the `gz_compress` function so that it now uses the `PREFIX(gzfwrite)`.

nmoinvaz (approved):
LGTM!

mtl1979 (approved):
LGTM

Dead2 (approved):

PR #2111, opened 2026-01-17
Use compile time architecture detection

nmoinvaz:
@pmqs please take a look.


cmake/detect-arch.c
```diff
@@ -1,116 +1,14 @@
...
-    #error archfound wasm32
-
-// return 'unrecognized' if we do not know what architecture this is
-#else
-    #error archfound unrecognized
-#endif
+int main(void) {
+    fprintf(stderr, "archfound %s\n", ARCH_NAME);
+    return 0;
+}
```
phprus, line 14: This method breaks cross-compilation if an emulator is not available.

We can preprocess the file (`-E` for `gcc`, `/EP` for msvc):

```c
#include "../zarch.h"
arch: ARCH_NAME
```

and parse line with `arch: ` tag

https://godbolt.org/z/4b7avcd5q

nmoinvaz: MSVC doesn't output a file.

phprus: > MSVC doesn't output a file.

With the `/P` option, creates a file with the `.i` extension: https://learn.microsoft.com/en-us/cpp/build/reference/p-preprocess-to-a-file?view=msvc-140

nmoinvaz: I commited another method.

zarch.h
```diff
@@ -0,0 +1,188 @@
...
+#    define ARCH_NAME "wasm64"
+#  else
+#    define ARCH_32BIT
+#    define ARCH_NAME "wasm32"
+#  endif
+
+/* Unrecognized architecture */
+#else
+#  define ARCH_NAME "unrecognized"
+#endif
```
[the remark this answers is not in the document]
nmoinvaz: Added.

cmake/detect-arch.c
phprus: Your method will work even without an emulator.
Good idea!

zarch.h
```diff
@@ -0,0 +1,193 @@
...
+#elif defined(__aarch64__) || defined(__arm64__) || defined(_M_ARM64) || defined(_M_ARM64EC)
+#  define ARCH_ARM
+#  define ARCH_64BIT
+#  define ARCH_NAME "aarch64"
+
+/* ARM 32-bit */
+#elif defined(__arm__) || defined(__arm) || defined(_M_ARM) || defined(__TARGET_ARCH_ARM)
+#  define ARCH_ARM
+#  define ARCH_32BIT
+#  if defined(__ARM64_ARCH_8__) || defined(__ARMv8__) || defined(__ARMv8_A__)
```
mtl1979: Missing `__ARM_ARCH_8A__`

mtl1979 (changes requested):
Doesn't correctly detect ARMv8.4A when compiling 32-bit code... This is most recent instruction set my 32-bit ARM GNUEABI compiler supports.

pmqs:
> @pmqs please take a look.

Wow, this is impressive!

There are a LOT of platforms and their variants detected in this change.  How can this change be tested for correctness against _all_ of them? I presume emulation will get most/all of the way there? That implies a workflow and possibly some coverage testing.

I assume the contents of `zarch.h` has come from another project? Does it have any testing that can be borrowed for use here?

nmoinvaz:
The contents of `zarch.h` has come from `detect-arch.c`

zarch.h
nmoinvaz: Good looking out, seems it was missed in the original `detect-arch.c`.

zarch.h
```diff
@@ -0,0 +1,195 @@
...
+#  define ARCH_32BIT
+#  define ARCH_NAME "s370"
+#elif defined(__s390__)
+#  define ARCH_S390
+#  define ARCH_32BIT
+#  define ARCH_NAME "s390"
+#elif defined(__s390x) || defined(__zarch__)
+#  define ARCH_S390
+#  define ARCH_64BIT
+#  define ARCH_NAME "s390x"
```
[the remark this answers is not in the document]
nmoinvaz: Fixed.

mtl1979:
> There are a LOT of platforms and their variants detected in this change. How can this change be tested for correctness against _all_ of them? I presume emulation will get most/all of the way there? That implies a workflow and possibly some coverage testing.

I mostly use cross-gcc to test what preprocessor macros are defined for each supported CPU, then compare that the most specific preprocessor macro is listed. It is possible to make shell script that executes preprocessor for every supported CPU and then verifies that the "default" architecture is never returned.

nmoinvaz:
Rebased.

cmake/detect-arch.c
```diff
@@ -1,116 +1,14 @@
-// archdetect.c -- Detect compiler architecture and raise preprocessor error
-//                 containing a simple arch identifier.
-// Copyright (C) 2019 Hans Kristian Rosbach
-// Licensed under the Zlib license, see LICENSE.md for details
+/* detect-arch.c -- Detect compiler architecture and print to stderr
+ *                  containing a simple arch identifier.
+ * Copyright (C) 2019 Hans Kristian Rosbach
```
mtl1979, line 3: Just a nitpick, but this should be updated...

nmoinvaz: I saw that but wasn't really sure what to do about the date, since I didn't originate the code, and it doesn't really matter to me.

zarch.h
```diff
@@ -0,0 +1,195 @@
+/* zarch.h -- Detect compiler architecture and define ARCH_* macros
+ * Copyright (C) 2019 Hans Kristian Rosbach
```
mtl1979, line 2: Just a nitpick, but this line should be updated.

mtl1979 (commented):
We're living year 2026, so copyright year of 2019 for new or edited file looks weird.

nmoinvaz:
Yeah but I’m not original author so what authority do I have to change that? IMO it should be updated once a year in all the source or the years should be removed from all files.

mtl1979:
> Yeah but I’m not original author so what authority do I have to change that? IMO it should be updated once a year in all the source or the years should be removed from all files.

When there is no substantial edits by other people, then just changing it to `2019-2026` would be the only "legal" alternative. This repository might be named "zlib-ng", but in legal sense, it still belongs to @Dead2. As such, it's not considered a fork, where copyright can "expire" after substantial rewrite or code can be considered abandonware.

nmoinvaz:
We don't update the years when we change the files where Mark Adler has his named in the copyright.

I suggest we just remove the years from all authors who agree and whoever doesn't agree we just always leave the year alone, but it should be a separate PR.

That way we don't always have to ask contributors to update the year.

mtl1979:
> We don't update the years when we change the files where Mark Adler has his named in the copyright.
> 
> I suggest we just remove the years from all authors who agree and whoever doesn't agree we just always leave the year alone, but it should be a separate PR.
> 
> That way we don't always have to ask contributors to update the year.

Mark Adler hasn't made any pull requests in this repository, so that justification makes no sense... In legal sense, copyright expires 15 to 25 years after last edit when the repository is forked, but 70 to 100 years on the original repository depending on legislation.

IMHO new files should always have current year as end of the range, but I have no strong opinion on existing files with minor edits.

phprus (approved):
LGTM

Dead2 (approved):
LGTM

PR #2119, opened 2026-01-20
Add "None" for sanitizer option

pps83:


pps83:
when using cmake gui if you touch `WITH_SANITIZER` option it presents you with list of choices without an option to opt out from using any sanitizer:

image


this PR adds "None" as the first option, changing this list to the following:


image

nmoinvaz (approved):

benoit-pierre:
I think this should use "OFF" instead (a CMake "falsy" value), so the features report shows `WITH_SANITIZER` has disabled by default or when explicitly using `-DWITH_SANITIZER=OFF`:

```diff
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,8 +114,8 @@
 Default is no prefix (empty prefix).")
 
 # Add multi-choice option
-set(WITH_SANITIZER AUTO CACHE STRING "Enable sanitizer support")
-set_property(CACHE WITH_SANITIZER PROPERTY STRINGS "None" "Memory" "Address" "Undefined" "Thread")
+set(WITH_SANITIZER OFF CACHE STRING "Enable sanitizer support")
+set_property(CACHE WITH_SANITIZER PROPERTY STRINGS "OFF" "Memory" "Address" "Undefined" "Thread")
 
 if(BASEARCH_ARM_FOUND)
     cmake_dependent_option(WITH_ARMV6 "Build with ARMv6 SIMD" ON "NOT ARCH MATCHES \"aarch64\"" OFF)
```

pps83:
> I think this should use "OFF" instead (a CMake "falsy" value), so the features report shows `WITH_SANITIZER` has disabled by default or when explicitly using `-DWITH_SANITIZER=OFF`:


I don't know cmake much, but if that's the convention then it should be changed to "OFF". The way it is now, anything will work there afaik, any non-matching string will simply work as OFF. In short, `-DWITH_SANITIZER=OFF` should work now

@coderabbitai what's your take?

pps83:
added pr: https://github.com/zlib-ng/zlib-ng/pull/2141

PR #2122, opened 2026-01-22
Move DoNotOptimize in benchmark loop to prevent it being optimized away

nmoinvaz:
In some cases I've noticed incorrect 0 benchmark results from compiler optimizing away len during hash benchmarks.

It happened while testing changes to compare256 implementations. DoNotOptimize should go inside for loop.

### DIFF
--- a/test/benchmarks/benchmark_adler32.cc
+++ b/test/benchmarks/benchmark_adler32.cc
@@ -41,10 +41,10 @@ public:
                 misalign = 0;
             else
                 misalign += (DO_ALIGNED) ? 16 : 1;
-        }
 
-        // Prevent the result from being optimized away
-        benchmark::DoNotOptimize(hash);
+            // Prevent the result from being optimized away
+            benchmark::DoNotOptimize(hash);
+        }
     }
 
     void TearDown(const ::benchmark::State&) {
--- a/test/benchmarks/benchmark_adler32_copy.cc
+++ b/test/benchmarks/benchmark_adler32_copy.cc
@@ -45,10 +45,10 @@ public:
                 misalign = 0;
             else
                 misalign += (DO_ALIGNED) ? 16 : 1;
-        }
 
-        // Prevent the result from being optimized away
-        benchmark::DoNotOptimize(hash);
+            // Prevent the result from being optimized away
+            benchmark::DoNotOptimize(hash);
+        }
     }
 
     void TearDown(const ::benchmark::State&) {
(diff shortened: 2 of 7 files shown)


mtl1979:
I've seen zero benchmark times too... First I thought it was just because the length was small and branch predictor was lucky, but then it happened also with longer lengths -- on a toaster. Getting ultra fast results with a computer that gets just 50 bogomips is impossible.

phprus:
DoNotOptimize is missing:
https://github.com/zlib-ng/zlib-ng/blob/develop/test/benchmarks/benchmark_inflate.cc
https://github.com/zlib-ng/zlib-ng/blob/develop/test/benchmarks/benchmark_insert_string.cc (insert_string_bench)

Dead2:
I guess inlining (and especially forceinline) is making it easier for the compiler to realize that the function has no side-effects, enabling it to just skip the whole function. That also enables it to optimize out other side-effects and checks (no null check, it might already know the alignment and skips that, it don't need code for handling len=! what it is benchmarking currently, and so on).

I think this is a problem that only affects functions that get included into the benchmark compile unit, like a few of compare256 variants do, and of course any function in LTO builds. I guess that is why I have never seen zero timings.

But this also means that the proposed fix might not be enough to avoid the compiler optimizing specifically for the benchmark.
I don't think any google benchmark using LTO can be trusted, because they don't represent how the code will be optimized in the library itself. External benchmarks using minigzip or accessing the API would of course be valid in LTO builds as well.

We really should somehow forcibly avoid inlining functions that are being benchmarked. Not sure how we could do that though.

nmoinvaz:
@phprus as far as I can tell those other benchmarks don't need DoNotOptimize. It is primarily for these hash functions which have potential to be inlined because they are included directly into the benchmark, they don't get called though functable.

mtl1979:
`DoNotOptimize` is only needed when the return value of the benchmarked function is not used in subsequent code and thus is ignored. Optimizer can't know if any function parameters cause side-effects unless they are passed by reference. 

Some people think side-effects can be deduced from dereferencing pointers, but most compilers by default only check the first "array item" pointed by the pointer, not the subsequent ones as there is no attribute to tell how many "items" the "array" the pointer points to has. 

Some compilers allow adding attribute that tells minimum length of the array the pointer points to, others allow specifying index of the function parameter that tell the array size.

Dead2 (approved):

PR #2123, opened 2026-01-22
Remove unnecessary CHUNK_SIZE define

pps83:


arch/riscv/chunkset_rvv.c
Dead2: You probably forgot to remove this.

pps83: thanks. updated

nmoinvaz (approved):

Dead2 (approved):

PR #2125, opened 2026-01-24
Minor memory layout alignment changes

nmoinvaz:
`matches` and `sym_*` are now in the same cache line.
`strm->avail_in` and `strm->next_out` assignments are ordered sequentially.
`wrap` and `flags` are used often together.
`whave` and `wnext` can now be loaded together on arm in `inflate_fast`.

### DIFF
--- a/deflate.h
+++ b/deflate.h
@@ -249,6 +249,14 @@ struct ALIGNED_(64) internal_state {
     struct tree_desc_s d_desc;               /* desc. for distance tree */
     struct tree_desc_s bl_desc;              /* desc. for bit length tree */
 
+                /* Cacheline 46 - Symbol tallying (accessed with dyn_ltree/dyn_dtree) */
+
+    unsigned int sym_next;        /* running index in symbol buffer */
+    unsigned int sym_end;         /* symbol table full when sym_next reaches this */
+    unsigned int matches;         /* number of string matches in current block */
+    unsigned int opt_len;         /* bit length of current block with optimal trees */
+    unsigned int static_len;      /* bit length of current block with static trees */
+
     uint16_t bl_count[MAX_BITS+1];
     /* number of codes at each bit length for an optimal tree */
 
@@ -281,6 +289,8 @@ struct ALIGNED_(64) internal_state {
      *   - I can't count above 4
      */
 
+                /* Cacheline 92 */
+
 #ifdef LIT_MEM
 #   define LIT_BUFS 5
     uint16_t *d_buf;              /* buffer for distances */
@@ -289,15 +299,6 @@ struct ALIGNED_(64) internal_state {
 #   define LIT_BUFS 4
     unsigned char *sym_buf;       /* buffer for distances and literals/lengths */
 #endif
-
-    unsigned int sym_next;        /* running index in symbol buffer */
-    unsigned int sym_end;         /* symbol table full when sym_next reaches this */
-    unsigned int matches;         /* number of string matches in current block */
-
-                /* Cacheline 92 - Block statistics */
-
-    unsigned int opt_len;         /* bit length of current block with optimal trees */
-    unsigned int static_len;      /* bit length of current block with static trees */
 
     deflate_allocs *alloc_bufs;
 
--- a/deflate.h
+++ b/deflate.h
@@ -135,7 +135,7 @@ typedef struct deflate_allocs_s {
 } deflate_allocs;
 
 struct ALIGNED_(64) internal_state {
-                /* Cacheline 0 */
+                /* Cacheline 0 - Stream and pending buffer management */
     PREFIX3(stream)      *strm;            /* pointer back to this zlib stream */
     unsigned char        *pending_buf;     /* output still pending */
     unsigned char        *pending_out;     /* next pending byte to output to the stream */
@@ -153,7 +153,7 @@ struct ALIGNED_(64) internal_state {
      * This is set to 1 if there is an active block, or 0 if the block was just closed.
      */
 
-                /* Cacheline 1 */
+                /* Cacheline 1 - Window and hash table pointers */
 
     unsigned int  lookahead;    /* number of valid bytes ahead in window */
     unsigned int strstart;      /* start of string to insert */
@@ -198,7 +198,7 @@ struct ALIGNED_(64) internal_state {
     int          match_available;    /* set if previous match exists */
     uint32_t     prev_match;         /* previous match (used by deflate_slow) */
 
-                /* Cacheline 2 */
+                /* Cacheline 2 - Match finding and compression parameters */
 
     unsigned int match_start;        /* start of matching string */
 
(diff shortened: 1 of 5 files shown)


deflate.h
```diff
@@ -295,10 +294,12 @@ struct ALIGNED_(64) internal_state {
 
     unsigned int sym_next;        /* running index in symbol buffer */
     unsigned int sym_end;         /* symbol table full when sym_next reaches this */
-
+    unsigned int matches;         /* number of string matches in current block */
     unsigned int opt_len;         /* bit length of current block with optimal trees */
     unsigned int static_len;      /* bit length of current block with static trees */
 
+    int32_t padding5[4];
+
```
mtl1979: We usually don't add padding to random places... There is no clear indication where cacheline 3 ends or cacheline 4 starts.

deflate.h
```diff
@@ -240,7 +239,7 @@ struct ALIGNED_(64) internal_state {
     int32_t padding1[1];
 
                 /* Cacheline 3 */
-    uint8_t ALIGNED_(16) padding4[68];
+    uint8_t ALIGNED_(16) padding4[52];
```
mtl1979: Is this supposed to align everything to end of cacheline 3 or add implicit padding to end of cacheline 2?

More than 12 bytes of padding is overkill. More than 32 bytes is facepalm and requires 128 byte cachelines, while most systems only have 64 byte cachelines.

mtl1979 (commented):
Math doesn't Math...

deflate.h
nmoinvaz: At this point it is not cacheline 4, it is like cacheline 92.

nmoinvaz: And the padding is there to ensure that the rest of the cachelines are still consistent.

deflate.h
nmoinvaz: I was shifting padding in order to attempt to get everything to line up just nice. But I think it is easier to just remove that original padding.

deflate.h
nmoinvaz: I have removed that extra padding and just added more comments so you can see where the cachelines are.

mtl1979: > At this point it is not cacheline 4, it is like cacheline 92.

In the original code, the large padding filled whole cacheline 3 and start of cacheline 4, so there wasn't easy way to mark the boundary of cacheline, thus not easy way to see where the subsequent cacheline boundaries are. We really need to mark boundaries of cachelines where we need padding, so reviewers don't need to start from very beginning of the struct to verify padding is in sane location.

On most 32-bit processors 12 bytes padding can be needed after single (16 byte aligned) 32-bit integer or pointer if the next struct member requires 16-byte alignment, but on 64-bit processors only maximum of 8 bytes of padding are required as pointers are bigger and it's possible to shuffle around struct members to fill 4 byte gaps.

deflate.h
mtl1979: Recent versions of Visual Studio have a tool to align struct members to cacheline boundaries, but it requires commenting out existing padding... I've used it to verify previous PRs that shuffle struct members for better cache locality.

nmoinvaz: Claude is great at doing this. I have it manually counting. But even that is not enough because the compiler will add padding implicitly so I told Claude to create a test file that uses offsetof to print out the offset of each field to ensure the boundary of each cacheline was where it said it was.

mtl1979: When the compiler itself does the aligning, implicit padding is not an issue as it's basically what the reported padding requirement is. If the layout is beyond fixing, the reported padding is usually larger than sane value, so easy to spot. I always compile for both 32-bit and 64-bit to see architecture specific padding.

For hints to the compiler, it's usually enough to align first struct member of each cacheline to 64 byte boundary.

mtl1979:
This is how I verified:

```
diff --git a/deflate.h b/deflate.h
index d04627f7..74e200ce 100644
--- a/deflate.h
+++ b/deflate.h
@@ -134,6 +134,10 @@ typedef struct deflate_allocs_s {
     Pos             *head;
 } deflate_allocs;

+#ifdef _MSC_VER
+#pragma warning(push)
+#pragma warning(error:4820)
+#endif
 struct ALIGNED_(64) internal_state {
                 /* Cacheline 0 - Stream and pending buffer management */
     PREFIX3(stream)      *strm;            /* pointer back to this zlib stream */
@@ -229,6 +233,9 @@ struct ALIGNED_(64) internal_state {
     unsigned int insert;        /* bytes at end of window left to insert */
     int32_t bi_valid;           /* Number of valid bits in bi_buf.
                                  * All bits above the last valid bit are always zero. */
+#ifdef ARCH_32BIT
+    int32_t padding0;
+#endif
     uint64_t bi_buf;            /* Output buffer.
                                  * Bits are inserted starting at the bottom (least significant bits). */

@@ -243,6 +250,7 @@ struct ALIGNED_(64) internal_state {
     struct ct_data_s dyn_ltree[HEAP_SIZE];   /* literal and length tree */
     struct ct_data_s dyn_dtree[2*D_CODES+1]; /* distance tree */
     struct ct_data_s bl_tree[2*BL_CODES+1];  /* Huffman tree for bit lengths */
+    int32_t padding2;

     struct tree_desc_s l_desc;               /* desc. for literal tree */
     struct tree_desc_s d_desc;               /* desc. for distance tree */
@@ -270,6 +278,7 @@ struct ALIGNED_(64) internal_state {
     uint8_t padding3[3];        /* explicit padding for lit_bufsize alignment */

     unsigned int  lit_bufsize;
+    unsigned int  padding4;
     /* Size of match buffer for literals/lengths.  There are 4 reasons for
      * limiting lit_bufsize to 64K:
      *   - frequencies can be kept in 16 bit counters
@@ -310,11 +319,20 @@ struct ALIGNED_(64) internal_state {
 #ifdef ZLIB_DEBUG
     unsigned long compressed_len; /* total bit length of compressed file mod 2^32 */
     unsigned long bits_sent;      /* bit length of compressed data sent mod 2^32 */
+#else
+    unsigned long padding5[2];
 #endif

     /* Reserved for future use and alignment purposes */
-    int32_t reserved[19];
+#ifdef ARCH_32BIT
+    int32_t reserved[26];
+#else
+    int32_t reserved[24];
+#endif
 };
+#ifdef _MSC_VER
+#pragma warning(pop)
+#endif

 typedef enum {
     need_more,      /* block not completed, need more input or more output */
```

Dead2:
I am not seeing any performance improvement here, it does seem to be very slightly slower both on x86_64 and Aarch64.
Compiled size is also same.

When I do changes to the structs, I always have to benchmark for every change to avoid regressions, it is extremely ticklish.
Btw, there is really no need to move things closer when you can't get it into the same cacheline, since as long as it is in the same TLB, it'll not benefit from it. So moving anything closer to the trees that span lots of cachelines is really not needed, it's more important that those are in the same cacheline as other stuff that get used at the same time. There is also some consideration about what is written vs only read.

nmoinvaz:
Some of the changes are looking at assembly instructions and getting them load in fewer instructions if they are closer to one another in the struct. Other changes are just moving loads to be sequential in memory, you would think that the compiler would do that for you, but it doesn't in every instance.

I don't expect performance improvement to be measurable.

I think I will let you do your struct layout changes.

PR #2127, opened 2026-01-25
Check cpuid availability on FreeBSD/OpenBSD when detecting fast/pmull

nmoinvaz:
`arm_has_cpuid` already has these ifdefs in it.


Dead2 (approved):
LGTM

PR #2129, opened 2026-01-26
Port to Elbrus processors

phprus:
Added support for the e2k architecture to the build system.
Only the cmake is supported, as the native build is used.

MCST LCC compiler allows reuse of x86 optimizations.

Test results with CI enabled: https://github.com/phprus/zlib-ng/actions/runs/21358175395
CI in PR is disabled to avoid overloading the site https://dev.mcst.ru/ (downloading compilers and qemu).


.github/workflows/e2k.yml.disable
```diff
@@ -0,0 +1,115 @@
...
+jobs:
+  cmake:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 80
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: Ubuntu LCC e2kv4
```
nmoinvaz: Can you incorporate these into cmake.yaml for code coverage, instead of these being separate workflow?

mtl1979: As far as I understood, incorporating these to cmake.yml would most likely need using CDN for the toolchain and qemu downloads to avoid overloading the server.

nmoinvaz: It does that now.

nmoinvaz: `if: contains(matrix.name, "e2k")`
can be added to any step.

mtl1979: > It does that now.

```
C:\build\git\zlib-ng>nslookup dev.mcst.ru
Server:  dna.wifi
Address:  fe80::e867:1bff:fe2c:2451

Non-authoritative answer:
Name:    dev.mcst.ru
Address:  185.179.190.4
```

That's clearly single IP address...

nmoinvaz: it may be possible to use `actions/cache` to cache the file.

phprus: @nmoinvaz 
I'm afraid Github Actions might overload the dev.mcst.ru server. I asked if there are plans to use a CDN, but I haven't received an answer yet.

Therefore I moved the e2k build and testing into a separate workflow, which is disabled by default.


I think it's better to leave the workflow disabled for now.
If I get an answer, I'll move the e2k workflow to the cmake.yml in a separate PR.

mtl1979: Would help to know what is the total size of all the files... Then we could figure out if we can host the binaries elsewhere, where we can control the traffic amount better... I don't know what's the current status of CloudFlare in Russia currently, but it's the easiest and cheapest way to avoid overloading servers that are not load-balanced.

phprus: @mtl1979 
It looks like we can use `actions/cache`.
I'll try to make a prototype using `actions/cache`.

> the current status of CloudFlare in Russia currently

:((

> the total size of all the files

~785 Mb.

mtl1979: I don't see anything else obvious in this PR that warrants fixing, but we'll see if using `actions/cache` allows us to run CI successfully. Populating the cache might still cause quite a load on the server, but if everything is wired correctly, it should be one-time inconvenience...

nmoinvaz: If that fails we can always host it from a Dropbox link 😆

mtl1979: I have 20 TB/month bandwidth with 160 GB storage.

phprus:
Ready for review

.github/workflows/cmake.yml
```diff
@@ -797,8 +857,8 @@ jobs:
       run: cmake --build . --verbose -j2 --config ${{ matrix.build-config || 'Release' }}
 
     - name: Run test cases
-      # Don't run tests on Windows ARM
-      if: runner.os != 'Windows' || contains(matrix.name, 'ARM') == false
+      # Don't run tests on Windows ARM and on e2kv6 (qemu: unimplemented operation CLMULL)
+      if: (runner.os != 'Windows' || contains(matrix.name, 'ARM') == false) && !contains(matrix.name, 'e2kv6')
```
mtl1979: Currently CLMULL is disabled only for v5 and older but for me it looks like we should disable it for v6 too on CI until qemu supports it. For x86 we already have build option to disable use of PCLMUL, which is essentially the same.

phprus: v5 and older hardware don't have the CLMULL instruction.
v6 hardware has the CLMULL instruction.

CI is disabled for v6 due to qemu limitations (which will be fixed in the next version).
For this reason, I added a build for v6 but disabled running tests.

At the CMake, the CLMULL instruction is only enabled for v6 and later ( https://github.com/zlib-ng/zlib-ng/pull/2129/changes#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR147-R153 ).

When a next version of qemu is released, the url will need to be changed and this condition will need to be removed.

mtl1979: I know a lot about limitations in qemu emulation... I'm using very old version of qemu myself (version 2.8.0) for similar reason. I was just saying that we could make CLMULL optional, so the tests don't need to be disabled until qemu properly supports CLMULL.

phprus: The next release of qemu with support for the CLMULL instruction should be published soon.
There is currently a private beta version in which all tests pass without errors.

Therefore, I propose using the current CI implementation to simplify future update of qemu in CI.

phprus:
Rebased.

.github/workflows/cmake.yml
mtl1979: It was just an observation, not request to change. I'm still going through all the comments AI bot made before approving the "final" version...

.github/workflows/cmake.yml
```diff
@@ -752,6 +776,42 @@ jobs:
...
+        key: cache-lcc-${{ matrix.toolchain }}
+
+    - name: Install MCST LCC ${{ matrix.toolchain }} (Ubuntu)
+      if: contains(matrix.name, 'e2k') && steps.cache-lcc.outputs.cache-hit != 'true'
+      continue-on-error: false
+      shell: bash
+      run: |
+        wget -q https://dev.mcst.ru/downloads/2025-06-27/cross-sp-${{ matrix.toolchain }}_64.tgz
+        sudo tar -xzf cross-sp-${{ matrix.toolchain }}_64.tgz -C /
+        rm -f cross-sp-${{ matrix.toolchain }}_64.tgz
```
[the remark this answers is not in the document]
mtl1979: We are trying to use only tools that are absolutely necessary so we don't need to install more packages. We don't verify checksums for other downloads.

cmake/detect-arch.cmake
```diff
@@ -143,6 +144,31 @@ elseif("${ARCH}" MATCHES "wasm(32|64)")
     else()
         set(ARCH_BITS 32)
     endif()
+elseif("${ARCH}" MATCHES "e2k")
+    set(BASEARCH "e2k")
+    set(BASEARCH_E2K_FOUND TRUE)
+    set(ARCH_BITS 64)
+
+    # Compile detect-e2k-isa.c and read the architecture version from the binary
+    try_compile(
```
nmoinvaz: I wonder why we can't use the normal `detect-arch.c` here? In `zarch.h` just do:
```c
#define E2K_STRINGIZE(X) E2K_DOSTRINGIZE(X)
#define E2K_DOSTRINGIZE(X) #X
#define ARCH_VERSION E2K_STRINGIZE(__iset__)
```

Then in `detect-arch.c`:
```c
int main(void) {
    fprintf(stderr, "archfound " ARCH_NAME "\n");
#ifdef ARCH_VERSION
    fprintf(stderr, "archversion " ARCH_VERSION "\n");
#endif 
    return 0;
}
```

nmoinvaz: Then in this cmake `elseif("${ARCH}" MATCHES "e2k")` branch just don't run `try_compile` again, instead just parse `archversion` from `COMPILE_RESULT` that already ran.

mtl1979: It would make sense if there were more architectures where the architecture version does matter, but currently there isn't... Now the logic is quite isolated, so there is minimal risk of breaking the logic for other architectures.

phprus: I think it's better to isolate this logic, since only Elbrus requires architecture version detection.
It's worth noting that architecture version is only needed by CMake to configure compiler flags, but is not required in the sources.

I agree with @mtl1979

nmoinvaz: It adds unnecessary source file and duplicates cmake calls that's not needed. Btw, archverison is not the only thing that might need to be done this way in the future. We could easily also print out the ARCH_BITS if we needed to. Also we might want to use archversion for ARM in the future. I don't know why is it that E2K has to diverge to the point of it needing its own detect.c when none of the other architectures do?

phprus: Dynamic function dispatching doesn't make sense for Elbrus, and native build is used.
Therefore, the architecture version must be determined at the CMake stage for the select optimization options.

~~I will make changes to merge `detect-arch.c` and `detect-e2k-isa.c`~~

phprus: It's not possible to merge detect-arch.c and detect-e2k-isa.c.

`detect-arch.c` is not used during cross-compilation:

https://github.com/zlib-ng/zlib-ng/blob/401654554c748ff0e06566aa1e27a4507848d56b/cmake/detect-arch.cmake#L11-L21

We need to get the architecture version from the cross compiler.

nmoinvaz: In that case, does it make sense to just have two different toolchain files, one for each version?

nmoinvaz: For the power toolchains, there are some that have `-cpu power8` and some that have `-cpu power9`, so it seems to some extent we are doing that for other archs.

And on at least one of the toolchain files we call `add_compile_options`.

phprus: I think splitting toolchain files will only make them more difficult to use.

nmoinvaz: I do think more so it should be two separate toolchain files, especially since you are passing in different `QEMU_CPU` variables on each instance. And ideally with two toolchain files you do not need to even have a `detect-e2k-arch.c`. 

But there is something else I discovered, is that this is not necessary:
```cmake
elseif(CMAKE_CROSSCOMPILING)
    set(ARCH ${CMAKE_C_COMPILER_TARGET})
```
It is only done previously because we were using `try_run` instead of `try_compile` (changed in commit cda9b506).

phprus: > I do think more so it should be two separate toolchain files

10 files for all possible combination of architectures and cpus.
The current toolchain file can be used for all configurations due to its parameters.

I simplified the architecture detection code based on your example.

nmoinvaz: This is looking good, thanks!

zarch.h
```diff
@@ -182,6 +182,18 @@
...
+/* Elbrus 2000 (aka e2k) */
+#elif defined(__e2k__)
+#  define ARCH_E2K
+/* e2k reuse x86 optimizations */
+#  define ARCH_X86
+#  define ARCH_64BIT
+#  define ARCH_NAME "e2k"
+
+#  define E2K_STRINGIZE(X) E2K_DOSTRINGIZE(X)
+#  define E2K_DOSTRINGIZE(X) #X
```
mtl1979: If ever these need to be used for other architectures, we basically need to duplicate these... It would be better if these two macros would have generic name, so they could be reused for any necessary architecture and also could be moved outside architecture detection on top of the file.

nmoinvaz: The nice thing about zarch.h currently is it requires no outside includes.

mtl1979: > The nice thing about zarch.h currently is it requires no outside includes.

Yup... This means we don't need to care about other headers using conflicting names for preprocessor macros or defines.

phprus: The macro has been moved to the top of the file and renamed to `ZARCH_STRINGIZE`.

zarch.h
```diff
@@ -182,6 +185,15 @@
...
 #  endif
 
+/* Elbrus 2000 (aka e2k) */
+#elif defined(__e2k__)
+#  define ARCH_E2K
+/* e2k reuse x86 optimizations */
+#  define ARCH_X86
+#  define ARCH_64BIT
+#  define ARCH_NAME "e2k"
+#  define ARCH_VERSION_STR ZARCH_STRINGIZE(__iset__)
```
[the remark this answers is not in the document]
phprus: This is wrong.
Clang defines the `__iset__` macro for Elbrus.

- https://gitflic.ru/project/e2khome/llvm-project-e-2-k/blob?file=clang%2Flib%2FBasic%2FTargets%2FElbrus.cpp&branch=llvmorg-21-e2k#line-num-132
- https://gitflic.ru/project/e2khome/llvm-project-e-2-k/blob?file=clang%2Flib%2FBasic%2FTargets%2FElbrus.cpp&branch=llvmorg-19-e2k#line-num-123
- https://gitflic.ru/project/e2khome/llvm-project-e-2-k/blob?file=clang%2Flib%2FBasic%2FTargets%2FElbrus.cpp&branch=llvmorg-17.0.1-e2k#line-num-123

cmake/toolchain-e2k-lcc.cmake
```diff
@@ -0,0 +1,32 @@
...
+find_program(C_COMPILER_FULL_PATH NAMES lcc PATHS ${MCST_LCC_PREFIX}/bin)
+if(NOT C_COMPILER_FULL_PATH)
+    message(FATAL_ERROR "Cross-compiler for ${CMAKE_C_COMPILER_TARGET} not found")
+endif()
+set(CMAKE_C_COMPILER ${C_COMPILER_FULL_PATH})
+
+find_program(CXX_COMPILER_FULL_PATH NAMES l++ PATHS ${MCST_LCC_PREFIX}/bin)
+if(CXX_COMPILER_FULL_PATH)
+    set(CMAKE_CXX_COMPILER ${CXX_COMPILER_FULL_PATH})
+endif()
```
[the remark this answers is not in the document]
mtl1979: Adding check that the toolchain prefix is set to valid path gives clear indication how to use the toolchain file locally and makes sure any regression or corruption in CI workflow file is caught.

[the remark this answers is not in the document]
phprus: Checks added.

cmake/toolchain-e2k-lcc.cmake
```diff
@@ -0,0 +1,44 @@
...
+set(CMAKE_SYSTEM_VERSION 1)
+
+set(CMAKE_C_COMPILER_TARGET e2k-linux-gnu)
+set(CMAKE_CXX_COMPILER_TARGET e2k-linux-gnu)
+
+set(CMAKE_CROSSCOMPILING TRUE)
+
+# Require explicit toolchain prefix
+if(NOT MCST_LCC_PREFIX)
+    message(FATAL_ERROR "MCST_LCC_PREFIX must be set to the LCC install prefix (e.g., /opt/mcst/lcc-<ver>)")
```
mtl1979: The comma after `e.g.` is unnecessary.

Dead2:
I see no explanation of why native build is forced, we don't do that for any other architecture. At the very least, there should be a comment in CMake detailing why this is done.

And please enable codecov, if possible. If not, add comments to state why it is not (similar to all the other workflow jobs).

phprus:
Elbrus (e2k) is a VLIW architecture where performance is highly dependent on the compiler. The e2k architecture versions are backward compatible: v4, v5 will work on v6 but with less efficiency.

Therefore, it makes no sense to implement dynamic dispatching of functions - this will lead to a decrease in performance, not an increase.
This is why the native build is used.

This is the usual approach to compiling applications for Elbrus.

Dead2:
> Elbrus (e2k) is a VLIW architecture where performance is highly dependent on the compiler. The e2k architecture versions are backward compatible: v4, v5 will work on v6 but with less efficiency.
> 
> Therefore, it makes no sense to implement dynamic dispatching of functions - this will lead to a decrease in performance, not an increase. This is why the native build is used.
> 
> This is the usual approach to compiling applications for Elbrus.

That makes sense then :+1: , please add a comment with this info at CMakelists.txt:141, most people who are going to edit that file do not know Elbrus well enough to know this.

phprus:
> And please enable codecov, if possible. If not, add comments to state why it is not (similar to all the other workflow jobs).

Enabling codecov increases the tests execution time to 40 minutes.

Comment added.

.github/workflows/cmake.yml
```diff
@@ -841,7 +901,7 @@ jobs:
         COVERALLS_REPO_TOKEN: "${{ secrets.COVERALLS_REPO_TOKEN }}"
 
     - name: Test benchmarks (crashtest only, no coverage data collection)
-      if: contains(matrix.cmake-args, '-DWITH_BENCHMARKS=ON')
+      if: contains(matrix.cmake-args, '-DWITH_BENCHMARKS=ON') && !contains(matrix.name, 'e2kv6')
```
Dead2: Why enable benchmarks for e2kv6 when you don't run them? Might as well not enable them in the first place, then there is no need for this override.

.github/workflows/cmake.yml
Dead2: Since tests are slow for emulated architectures, might want to add to this to skip the extra test corpora.

Dead2: Should help with enabling coverage testing being slow..

.github/workflows/cmake.yml
phprus: This is a temporary workaround due to an issue in qemu-e2k. The issue has already been fixed, but the fix has not yet been published.

.github/workflows/cmake.yml
```diff
@@ -665,8 +699,8 @@ jobs:
       uses: actions/checkout@v6
       # Don't test against all corpora with MinGW due to Wine being unable to run parallel jobs
       # without connection timeout. Without parallel jobs test runs using Wine take close to an hour.
-      # Also don't test the extra corpora with Sanitizer builds or RISC-V, due to their slow testing.
-      if: ${{ !contains(matrix.name, 'MinGW') && !contains(matrix.name, 'RISC-V') && !contains(matrix.cmake-args, 'WITH_SANITIZER') }}
+      # Also don't test the extra corpora with Sanitizer builds or RISC-V or e2kv5 with codecov, due to their slow testing.
```
Dead2: You'll need to disable this for all of them, spending 23+ min on a single job is too much, the shipped tests should be good enough.

.github/workflows/cmake.yml
```diff
@@ -655,6 +655,40 @@ jobs:
...
+            ldflags: -static
+
+          - name: Ubuntu LCC e2kv5 (codecov)
+            os: ubuntu-latest
+            cmake-args: -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-e2k-lcc.cmake -DMCST_LCC_PREFIX=/opt/mcst/lcc-1.29.12.e2k-v5.8c2.linux-6.1 -DQEMU_CPU=elbrus-v5
+            toolchain: 1.29.12.e2k-v5.8c2.linux-6.1
+            cflags: -march=elbrus-v5
+            cxxflags: -march=elbrus-v5
+            ldflags: -static
+            codecov: linux_mcst_lcc_e2kv5
```
Dead2: codecov should be the last line of the job, same as all the others. If codecov is not enabled, put a comment there instead with the reason why it is disabled.
But might as well enable codecov for all of these I'd think, and also no need for one with and one without codecov testing the same things.

.github/workflows/cmake.yml
Dead2: I'd rather you disable the building of the benchmarks entirely for that job rather than add another hard-to-maintain special-case here.

Dead2 (commented):
We're trying to keep total runtime of the whole CI down to about 20min, so we can't have single jobs taking that long unfortunately, those jobs easily block other PRs when CI is running on multiple commits at the same time.

.github/workflows/cmake.yml
phprus: Codecov disables compiler optimizations. For e2k VLIW, compiler optimizations are important, so it is important to test code with compiler optimizations enabled.

.github/workflows/cmake.yml
mtl1979: I already suggested disabling PCLMUL/CLMULL optimizations in CI until qemu is updated as that is what causes the issue.

.github/workflows/cmake.yml
```diff
@@ -752,6 +776,42 @@ jobs:
...
+        rm -f cross-sp-${{ matrix.toolchain }}_64.tgz
+
+    # Get qemu-e2k-static from github actions cache or download from dev.mcst.ru and save to cache
+    - name: Cache MCST QEMU (Ubuntu)
+      if: contains(matrix.name, 'e2k')
+      id: cache-qemu-e2k-static
+      uses: actions/cache@v5
+      with:
+        path: /usr/local/bin/qemu-e2k-static
+        key: cache-qemu-e2k-static-20251226
```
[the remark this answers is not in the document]
mtl1979: Snapshot builds don't have usable version number. Whenever a new architecture is added, there will be several snapshots for testing before all the changes are pushed upstream.

Dead2 (approved):
LGTM

PR #2131, opened 2026-01-28
Fix building on FreeBSD/OpenBSD

brad0:
Put the checks in the right order. Newer before older.


nmoinvaz (approved):

nmoinvaz:
Thanks, I had thought the 64-bit stuff was newer.

mtl1979:
> Thanks, I had thought the 64-bit stuff was newer.

One of the reasons we should be careful when reorganising code... Can't know or remember everything and assuming is generally a bad idea...

The bot likes suggesting fixing "code" and white-space issues in same PR, but it's generally frowned upon... The suggestions also contain extra characters that I can't seem to "teach" it to not introduce. Seems like recent regression.

Dead2 (approved):

PR #2078, opened 2026-01-07
Use GCC's may_alias attribute for access to buffers in crc32_chorba

phprus:
Converting a pointer from type `z_word_t*` (on 32-bit systems: `uint32_t*`) to `uint64_t*` breaks strict aliasing rules.

On GCC `-fstrict-aliasing` is `on` by default at `-O2` or above ( https://gcc.gnu.org/onlinedocs/gcc-10.2.0/gcc/Common-Type-Attributes.html ).

Fix #1926 
The compiler warning is no longer present.

Develop: https://github.com/zlib-ng/zlib-ng/actions/runs/20773110891/job/59653058832#step:17:46
PR: https://github.com/phprus/zlib-ng/actions/runs/20795098755/job/59726505687#step:17:46

@mtl1979 @samrussell

PR #2134, opened 2026-01-30
2.3.3 Release

Dead2:
This is a bugfix release, fixing several relatively minor bugs and updating CI workflows.

Changes since 2.3.2:
-----
### Bug fixes
- Make deflate output deterministic if stream is reused after deflateReset #2102 
- minigzip: Fix integer overflow in gz_compress_mmap #2110 
- Use GCC's may_alias attribute for access to buffers in crc32_chorba #2078
- Fix false-positive infinite loop warning detected by GCC-14 static analyzer #2101
- Fix warning for potentially uninitialized local variable ft used. #2043 

### Tests
- Fixed casting warning in benchmark_uncompress on MSVC #2035 

### Buildsystem
- Simplify the gzread.c name mangling workaround #2093
- Fix symbol mangling and prefix gz_error #2097
- CMake
  - Reset compiler flags after completing intrinsic tests #2082 
  - Add "None" for sanitizer option #2119 
- Configure
  - Fix detecting -fno-lto support #2045 

### Cleanup
- Remove leftover memory alloc test in deflateinit #2092

### CI
- Sync Github Actions workflows with develop


mtl1979:
This contains only the first part of the `configure` fix... The remaining parts are still incomplete and unmerged.

Dead2:
> This contains only the first part of the `configure` fix... The remaining parts are still incomplete and unmerged.

That PR contains only one commit. What other PR are you talking about?
To avoid unwanted side-effects, I am trying to be very conservative with pulling these, so I might have skipped some intentionally if I do not feel confident they are safe.

mtl1979:
> > This contains only the first part of the `configure` fix... The remaining parts are still incomplete and unmerged.
> 
> 
> 
> That PR contains only one commit. What other PR are you talking about?
> 
It was split out of a branch that did depend on #2038 because that part didn't overlap and was easy to split as own PR.

Dead2:
> It was split out of a branch that did depend on #2038 because that part didn't overlap and was easy to split as own PR.

Yeah, that one is much more likely to cause a regression, so I'd like to do that for 2.4.0 instead. Develop pretty much has to become 2.4 at this point, with all the refactoring changes.

mtl1979:
> > It was split out of a branch that did depend on #2038 because that part didn't overlap and was easy to split as own PR.
> 
> 
> 
> Yeah, that one is much more likely to cause a regression, so I'd like to do that for 2.4.0 instead. Develop pretty much has to become 2.4 at this point, with all the refactoring changes.

I'm not confident yet that we don't have enough bug fixes that there wouldn't be need for one more point release before switching to work on 2.4.0. I have only one unfinished branch currently but that is more of an improvement than bug fix. Lately I've focused more on going through other people's pull requests and issues.

mtl1979 (approved):
LGTM

PR #2141, opened 2026-02-04
Change "None" to "OFF" for sanitizer option

pps83:
_Originally posted by @benoit-pierre in https://github.com/zlib-ng/zlib-ng/issues/2119#issuecomment-3844711790_

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,7 +110,7 @@ Useful when embedding into a larger library.
 Default is no prefix (empty prefix).")
 
 # Add multi-choice option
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -111,7 +111,7 @@ Default is no prefix (empty prefix).")
 
 # Add multi-choice option
 set(WITH_SANITIZER AUTO CACHE STRING "Enable sanitizer support")
```
benoit-pierre: You also need to change the default:
```suggestion
set(WITH_SANITIZER OFF CACHE STRING "Enable sanitizer support")
```

nmoinvaz (approved):

PR #2143, opened 2026-02-06
Cleanup formatting for crc32_chorba files

nmoinvaz:
I remember when the chorba source files were added. I didn't mention all the formatting issues at the time because I feared the PR might die on the vine which happens because perfect becomes the enemy of good. So now here is the formatting cleanup.


Dead2 (approved):
LGTM

PR #2145, opened 2026-02-09
[zconf] Fix LFS support on Windows

### LINKED ISSUE #2144: WIN32 and _LARGEFILE64_SOURCE is not compiling
If somebody compiles with MSVC and define _LARGEFILE64_SOURCE there is a problem in zconf.h

For Windows, zconf.h define next lines:
```
#if 0    /* was set to #if 0 by configure/cmake/etc */
#  define Z_HAVE_UNISTD_H
#endif

```

And some bellow lines we have these other lines:
```
#if defined(Z_HAVE_UNISTD_H) || defined(_LARGEFILE64_SOURCE)
#  include <unistd.h>         /* for SEEK_*, off_t, and _LFS64_LARGEFILE */
#  ifndef z_off_t
#    define z_off_t off_t
#  endif
#endif

```
As we have defined _LARGEFILE64_SOURCE in windows, compiler tries to use "**unistd.h**" that doesn't exist in windows.

mtl1979:
* Windows doesn't have unistd.h, so z_off_t declaration only can depend on value of Z_HAVE_UNISTD_H.

Fixes #2144

### DIFF
--- a/zconf-ng.h.in
+++ b/zconf-ng.h.in
@@ -132,7 +132,7 @@ typedef PTRDIFF_TYPE ptrdiff_t;
 #  undef _LARGEFILE64_SOURCE
 #endif
 
-#if defined(Z_HAVE_UNISTD_H) || defined(_LARGEFILE64_SOURCE)
+#if defined(Z_HAVE_UNISTD_H)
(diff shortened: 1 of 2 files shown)


Dead2 (approved):
LGTM

PR #2146, opened 2026-02-09
defalte_medium: Make readbility init {0} structs for matches

GermanAizek:
visually, the code is simpler

### DIFF
--- a/deflate_medium.c
+++ b/deflate_medium.c
@@ -163,14 +163,11 @@ static void fizzle_matches(deflate_state *s, struct match *current, struct match
 
 Z_INTERNAL block_state deflate_medium(deflate_state *s, int flush) {
     /* Align the first struct to start on a new cacheline, this allows us to fit both structs in one cacheline */
-    ALIGNED_(16) struct match current_match;
-                 struct match next_match;
+    ALIGNED_(16) struct match current_match = {0};
+                 struct match next_match = {0};
 
     /* For levels below 5, don't check the next position for a better match */
     int early_exit = s->level < 5;
-
-    memset(&current_match, 0, sizeof(struct match));
-    memset(&next_match, 0, sizeof(struct match));
 
     for (;;) {
         uint32_t hash_head = 0;    /* head of the hash chain */


mtl1979:
This doesn't actually work on all GCC versions even though C standard defines zero initializer for structures... CI failures are however due to server crash...

nmoinvaz:
We already use zero initializers in a few parts of code. We should come to a consensus on using them or not.

mtl1979:
> We already use zero initializers in a few parts of code. We should come to a consensus on using them or not.

I'm not against zero initializers for structs or non-struct types, I just memorize all the compiler bugs and deviations from standards. This is exactly why with "some" compilers we don't allow any warnings on CI and treat any warnings as errors.

Dead2:
I am neutral to zero initializers, either way is really fine with me.

mtl1979:
IMHO, safest, but not cleanest way, is to explicitly initialize all members of small structs to `0`, instead of relying on compiler implicitly and silently zeroing all but first element of struct. This means we should only use `memset` for large structs, where explicitly initializing to zero each element would result in very long initializer list.

Dead2:
> IMHO, safest, but not cleanest way, is to explicitly initialize all members of small structs to `0`, instead of relying on compiler implicitly and silently zeroing all but first element of struct. This means we should only use `memset` for large structs, where explicitly initializing to zero each element would result in very long initializer list.

I'll amend my previous statement :slightly_smiling_face:. Zeroing individual members is not something I'd prefer over memset.

mtl1979:
> > IMHO, safest, but not cleanest way, is to explicitly initialize all members of small structs to `0`, instead of relying on compiler implicitly and silently zeroing all but first element of struct. This means we should only use `memset` for large structs, where explicitly initializing to zero each element would result in very long initializer list.
> 
> I'll amend my previous statement 🙂. Zeroing individual members is not something I'd prefer over memset.

Me neither, but it kinda follows our guideline of preferring explicit over implicit... There would be heated arguments over how many struct members would be enough to warrant using `memset` instead of zeroing individual struct elements. Some other projects have also used assignment of static empty structure as those are zeroed by design. This would essentially change the semantics from `memset` to `memcpy`, which is slightly more efficient.

mtl1979:
Should also fix the spelling error in the PR description... The commit message has the words backwards, in English we write from left to right, unlike in for example German, Korean and Arabic that write from right to left.

nmoinvaz (approved):

Dead2 (approved):

PR #2147, opened 2026-02-09
Added NUMA allocation support for native work on server platforms

GermanAizek:
This commits fix ignoring data locality with the distance to the nearest processor cache on multi-CPU platforms. By default select mask on all CPU sockets and local touch allocation memory.

References:
- https://stackoverflow.com/questions/8154162/numa-aware-cache-aligned-memory-allocation
- https://www.kernel.org/doc/html/v5.6/vm/numa.html
- https://hpc-wiki.info/hpc/NUMA
- https://man7.org/linux/man-pages/man3/numa.3.html
- https://en.wikipedia.org/wiki/Non-uniform_memory_access

Raspberry have increase performance with memory, if use fake numa nodes from linux kernel params.

- https://www.phoronix.com/news/ARM64-NUMA-Emulation-RPi5
- https://www.jeffgeerling.com/blog/2024/numa-emulation-speeds-pi-5-and-other-improvements/

Playstation 3 used NUMA but commercial failed.
In the future, due to the increasing complexity of crystal area development, it will be more profitable to create chiplet multicrystalline systems, so this is becoming more and more relevant. Although most NUMA devices are still and server platforms (Intel Xeon, AMD Epyc, Ampere Ultra).

- https://www.cs.york.ac.uk/rts/docs/ESWEEK-2007-tutorials/PS3_cell_tutorial.pdf

My topology:

image


nmoinvaz:
Have you thought about just passing in a custom `zalloc` and `zfree` to `deflateInit/inflateInit`?

mtl1979:
This has same issues as when we tried to support TLS... For storing the allocation size, we could increase the header padding by few bytes and store the "original" size before the "original" pointer. This way it would be compatible with most standard and custom allocators. It doesn't really matter if we allocate extra bytes as in worst case we allocate with 64 byte alignment instead of 16 byte alignment and that gives us maximum of 64 bytes for the allocator header, assuming the non-adjusted allocation start was already on 64-byte boundary.

Dead2:
Considering that zlib-ng is a library and not an application, I think it would be really bad form for us to change the cpu affinity of the whole application (or one of its threads) without regard for what the application expects. And of course without knowing what the application expects, it's hard to know what NUMA node to allocate memory from as well.

Therefore, the best approach here is to have applications provide a custom allocator through the zlib/zlib-ng API (it is meant for cases just like this), where the application can implement NUMA-aware allocations according to its needs and expectations. We already fix misaligned allocations and make our structures and buffers 64-byte aligned internally no matter what allocator is used, so there is really no downside to doing that.

Modern OSes are also good at preferring to alloc ram from the local numa node (if enough is available), and schedulers are pretty good at moving processes/threads to a cpu core that is local to its memory allocations.

---

And since you mentioned Rpi5, this PR would not further affect performance on those because those are fake numa nodes and our allocation is so small that it would not span multiple nodes anyways. The benefit there is primarily because the memory is segmented and it allows better parallelism in memory accesses on a system-level, and we already get that benefit from the kernel enabling it. (If we had a really big buffer, perhaps it would make sense to split it into slices and allocate them on separate fake numa nodes, but I am guessing probably not much since we don't get much memory parallelism without multithreading).

PR #2150, opened 2026-02-12
Update version info 2.3.3

jalegido:
Last version zlib-ng is 2.3.3 but version info has not been updated since 2.3.1.

I would like to nkow if it is possible include zlib-ng version in this view when compiled with **ZLIB_COMPAT=ON**. Maybe **_Product version_** could be changed to "2.3.3.zlib-ng".
With this information, user knows exactly which zlib-ng version has.

imagen


mtl1979:
Last two releases have been hotfix releases to fix critical bugs, so bumping the version numbers everywhere was not highest priority. Next "normal" version will be 2.4.0 when we get all the remaining improvements merged in.

There is plans to convert the "compat" version to small wrapper library that just converts the function parameters and return values while calling the real functions in the non-compat library... When that happens, we can change the "Product name" to "zlib-ng" and "Product version" to match the zlib-ng version it was designed to work with. 

The change to wrapper library will reduce the total required space when both versions are installed on same system and makes compiling changed tree locally faster as the wrapper functions/library doesn't need full recompile unless `zlib-ng.h` changes.

jalegido:
Understood.
Then you can close this PR.

Thank you.

PR #2156, opened 2026-02-14
Fixed unused function warning for arm_has_cpuid

nmoinvaz:


mtl1979 (approved):
LGTM, Windows has own API for detecting CPU features, it doesn't require low-level access.
Checking `ARCH_ARM` on file in `arch/arm` directory is unnecessary until we start supporting universal binaries, and even then it only needs to be done once.

Dead2 (approved):

PR #2159, opened 2026-02-17
cmake: Fix ARCH is empty in detect-arch

sergio-nsk:
The both `CMAKE_C_COMPILER_TARGET` and `CMAKE_SYSTEM_PROCESSOR` are undefined while configuring UWP/WinRT build with Clang:
`-G Ninja -D CMAKE_SYSTEM_NAME=WindowsStore`.
These variables are undefined because `-m` is not set to Clang.

`CMAKE_C_COMPILER_ARCHITECTURE_ID` could be used, but it would cause a more significant change of the cmake script.


mtl1979 (approved):
LGTM. Checking for both unset `ARCH` and empty string makes sense.

Dead2 (approved):
LGTM

PR #2168, opened 2026-02-19
Set version for the develop branch to 2.3.90

phprus:
Suggestion: Use version policy `2.x.90` for the develop branch so that version is higher than the released stable versions.

Currently, this will be version `2.3.90`.
When the new stable version `2.4.0` is released, the develop branch version should be set to `2.4.90`.


nmoinvaz:
@phprus feel free to join the Mattermost chat server mentioned in the GH wiki pages if you wish.

nmoinvaz:
This PR sounds like a good idea.

zutil.c
```diff
@@ -21,7 +21,7 @@ z_const char * const PREFIX(z_errmsg)[10] = {
 };
 
 const char PREFIX3(vstring)[] =
-    " zlib-ng 2.3.1";
+    " zlib-ng 2.3.90";
```
Dead2: I suggest:
`" zlib-ng 2.3.90 snapshot";`

Dead2 (commented):
Excellent, I was planning to do it very similarly, just haven't gotten around to it 👍

mtl1979:
Sounds like a good idea...

zutil.c
phprus: I added `devel` suffix because that is the documented value for the `ZLIBNG_VER_STATUS(H)` macros.

https://github.com/zlib-ng/zlib-ng/blob/08fa4859f38e3411d321e36b425ffc1f21b9d087/zlib-ng.h.in#L57

PR #2173, opened 2026-02-22
[CI] Add workflow to test MSVC 18 2026.

mtl1979:
* Support for Visual Studio 2026 is currently in public beta, so we can add workflow to follow the progress...

Depends on #2172.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -479,6 +479,12 @@ jobs:
             cmake-args: -DCMAKE_TOOLCHAIN_FILE=${EMSDK}/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake -DCMAKE_C_COMPILER_TARGET=wasm32 -DCMAKE_CROSSCOMPILING_EMULATOR=${EMSDK_NODE} -DZLIB_COMPAT=ON
             # codecov disabled for wasm
 
+          - name: Windows MSVC 2026 v145 Win64 C23
+            os: windows-2025-vs2026
+            compiler: cl
+            cmake-args: -G "Visual Studio 18 2026" -A x64 -T v145 -DCMAKE_C_STANDARD=23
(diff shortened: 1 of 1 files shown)


.github/workflows/cmake.yml
```diff
@@ -479,6 +479,12 @@ jobs:
             cmake-args: -DCMAKE_TOOLCHAIN_FILE=${EMSDK}/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake -DCMAKE_C_COMPILER_TARGET=wasm32 -DCMAKE_CROSSCOMPILING_EMULATOR=${EMSDK_NODE} -DZLIB_COMPAT=ON
             # codecov disabled for wasm
 
+          - name: Windows MSVC 2026 v145 Win64 C23
+            os: windows-2025-vs2026
```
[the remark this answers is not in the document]
mtl1979: You are correct, this run is used for monitoring issues that are specific to C23 support in Visual Studio 2026 and thus is expected to fail often. Any issues will be handled in separate PRs, including https://github.com/zlib-ng/zlib-ng/pull/2172. When the runner finishes the beta phase, we can add another workflow without explicit C23 support, using the default settings.

.github/workflows/cmake.yml
```diff
@@ -479,6 +479,12 @@ jobs:
...
+            compiler: cl
+            cmake-args: -G "Visual Studio 18 2026" -DCMAKE_C_STANDARD=23 -A x64 -T v145
```
nmoinvaz: nit: add ` -DCMAKE_C_STANDARD=23` to the end of the line instead of in the middle to match other `cmake-args` for VS.

nmoinvaz (approved):

.github/workflows/cmake.yml
mtl1979: Done.

PR #2174, opened 2026-02-23
Simplify adler32 alignment loops to advance pointers

nmoinvaz:
Replace `done`-`offset` tracking with direct pointer advancement in NEON, VMX, and SSSE3 adler32 implementations. Use `ALIGN_DIFF` consistently across all architectures for the initial alignment step.

### DIFF
--- a/arch/arm/adler32_neon.c
+++ b/arch/arm/adler32_neon.c
@@ -273,8 +273,6 @@ Z_FORCEINLINE static uint32_t adler32_copy_impl(uint32_t adler, uint8_t *dst, co
         return adler32_copy_tail(adler, dst, src, len, sum2, 1, 15, COPY);
 
     uint32_t pair[2];
-    int n = NMAX;
-    unsigned int done = 0;
 
     /* Split Adler-32 into component sums, it can be supplied by
      * the caller sites (e.g. in a PNG file).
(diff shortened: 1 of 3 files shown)


mtl1979:
There are clear logic errors in the new code as pointed out by the bots... This also makes some of the code a lot less readable.

adler32_p.h
```diff
@@ -18,34 +18,40 @@
...
 #define ADLER_DO16(sum1, sum2, buf)    {ADLER_DO8(sum1, sum2, buf, 0); ADLER_DO8(sum1, sum2, buf, 8);}
 
-Z_FORCEINLINE static void adler32_copy_align(uint32_t *Z_RESTRICT adler, uint8_t *dst, const uint8_t *buf, size_t len,
-                                             uint32_t *Z_RESTRICT sum2, const int MAX_LEN, const int COPY) {
-    Z_UNUSED(MAX_LEN);
-    if (len & 1) {
+Z_FORCEINLINE static size_t adler32_copy_align(uint32_t *Z_RESTRICT adler, uint8_t **dst, const uint8_t **buf, size_t *len,
+                                               uint32_t *Z_RESTRICT sum2, const int ALIGN, const int COPY) {
+    size_t align_diff = ALIGN_DIFF(*buf, ALIGN);
+    if (align_diff == 0 || align_diff > *len)
```
mtl1979: `align_diff` should never be more than `*len`, otherwise the short length handling is missing or broken.

arch/power/adler32_vmx.c
```diff
@@ -132,31 +132,23 @@ Z_INTERNAL uint32_t adler32_vmx(uint32_t adler, const uint8_t *buf, size_t len)
...
-        /* Rather than rebasing, we can reduce the max sums for the
-         * first round only */
-        n -= align_len;
-    }
-    for (size_t i = align_len; i < len; i += n) {
-        int remaining = (int)(len-i);
-        n = MIN(remaining, (i == align_len) ? n : NMAX);
-        if (n < 16)
-            break;
+    size_t n = NMAX - adler32_copy_align(&pair[0], NULL, &buf, &len, &pair[1], 16, 0);
```
mtl1979: Compiler can't optimize this code unless

a. We initialize `pair` after calling `adler32_copy_align` and use `adler` and `sum2` directly for `adler32_copy_align`
b. We pass pointer to `pair` instead of pointer to individual array elements to `adler32_copy_align`

Initialization for `pair` is expensive as it will require realigning the stack.

nmoinvaz: LLVM promotes `pair[0]`/`pair[1]` into registers for the scalar alignment loop despite receiving pointers, thanks to `Z_RESTRICT` and inlining.  The generated assembly is identical in the hot path — both versions use the same stack frame, same registers, same instruction count. Moving the initialization after `adler32_copy_align` actually adds instructions because `pair[]` has to be initialized on two separate code paths instead of one.

arch/arm/adler32_neon.c
```diff
@@ -296,41 +294,29 @@ Z_FORCEINLINE static uint32_t adler32_copy_impl(uint32_t adler, uint8_t *dst, co
...
-
-    while (done < len) {
-        int remaining = (int)(len - done);
-        n = MIN(remaining, (done == align_adj) ? n : NMAX);
+    size_t n = NMAX - adler32_copy_align(&pair[0], &dst, &src, &len, &pair[1], 32, COPY);
 
-        if (n < 16)
-            break;
+    while (len >= 16) {
+        n = MIN((int)len, n);
```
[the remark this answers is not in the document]
nmoinvaz: Fixed

arch/power/adler32_vmx.c
```diff
@@ -132,31 +132,23 @@ Z_INTERNAL uint32_t adler32_vmx(uint32_t adler, const uint8_t *buf, size_t len)
...
+
+    while (len >= 16) {
+        n = MIN((int)len, n);
```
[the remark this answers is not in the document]
nmoinvaz: Fixed

adler32_p.h
nmoinvaz: I have moved that back where it was.

nmoinvaz:
@mtl1979 I have changed the align function back to the way it was before, and just worked on cleaning up ptr advancement for this PR.

arch/arm/adler32_neon.c
```diff
@@ -296,41 +294,39 @@ Z_FORCEINLINE static uint32_t adler32_copy_impl(uint32_t adler, uint8_t *dst, co
...
         pair[0] %= BASE;
         pair[1] %= BASE;
 
-        int actual_nsums = (n >> 4) << 4;
-        done += actual_nsums;
+        int k = (n >> 4) << 4;
+        src += k;
+        if (COPY)
+            dst += k;
+        len -= k;
```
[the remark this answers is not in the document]
nmoinvaz: Fixed.

arch/power/adler32_vmx.c
```diff
@@ -132,31 +132,31 @@ Z_INTERNAL uint32_t adler32_vmx(uint32_t adler, const uint8_t *buf, size_t len)
...
+        vmx_accum32(pair, buf, n / 16);
         pair[0] %= BASE;
         pair[1] %= BASE;
 
-        done += (n / 16) * 16;
+        int k = (n / 16) * 16;
+        buf += k;
+        len -= k;
+        n = NMAX;
     }
```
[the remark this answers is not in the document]
nmoinvaz: Fixed the inconsistency.

arch/power/adler32_vmx.c
mtl1979: PowerPC assembly is optimized for pointer offsets which means it needs only one register for `pair` when accessing array elements. I've noticed that for example LoongArch assembly has same optimization. This means both values are guaranteed to always be stored in contiguous memory locations.

Dead2 (approved):
LGTM

Code size -135 bytes on x86-64.
No performance differences as far as I can tell.

PR #2184, opened 2026-03-05
UBSAN: Don't enable unsigned overflow errors by default

Dead2:
`unsigned-integer-overflow` and `unsigned-shift-base` are apparently not actually undefined in C/C++, the behavior is defined therefore don't warn about these.


nmoinvaz (approved):

mtl1979 (approved):
LGTM

PR #2187, opened 2026-03-06
Add NMAX_ALIGNED32 for 32-byte aligned NEON adler32

nmoinvaz:
NMAX (5552) is not a multiple of 32, so the NEON adler32 implementation could lose 32-byte alignment of src after the first SIMD loop iteration. This adds NMAX_ALIGNED32 (NMAX rounded down to 32, yielding 5536) in adler32_p.h and uses it in the NEON adler32 loop. The first iteration's budget is also rounded down with ALIGN_DOWN after subtracting the alignment preamble bytes, ensuring k is always a multiple of 32 and src stays 32-byte aligned throughout.

Summary generated with Claude Code.

### DIFF
--- a/adler32_p.h
+++ b/adler32_p.h
@@ -11,6 +11,8 @@
 #define BASE 65521U     /* largest prime smaller than 65536 */
 #define NMAX 5552
 /* NMAX is the largest n such that 255n(n+1)/2 + (n+1)(BASE-1) <= 2^32-1 */
+#define NMAX_ALIGNED32 (NMAX & ~31)
+/* NMAX rounded down to a multiple of 32 is 5536 */
 
 #define ADLER_DO1(sum1, sum2, buf, i)  {(sum1) += buf[(i)]; (sum2) += (sum1);}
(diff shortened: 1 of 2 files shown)


ccawley2011:
Should commit ced54ac be reverted as part of this PR now that 32-byte alignment can be guaranteed?

nmoinvaz:
I didn't even see that commit got in.

nmoinvaz:
Seems there are no UBSAN warnings since macOS builds are ok.

https://github.com/zlib-ng/zlib-ng/actions/runs/22780091866/job/66083359687

Dead2 (approved):
LGTM

mtl1979:
> Should commit [ced54ac](https://github.com/zlib-ng/zlib-ng/commit/ced54ac89cb79d8df912d741c25ea7bce9061761) be reverted as part of this PR now that 32-byte alignment can be guaranteed?

I'm OK with reverting that commit... It was just temporary fix to see there is no other issues than the one UBSAN noticed... Would not have been needed if this PR was created and merged before the alignment hint one...

It's really bad idea to keep `develop` branch in failing state for longer than few hours.

PR #2188, opened 2026-03-07
CI: Don't use GCC for ASAN/UBSAN on S390x

Dead2:
GCC on S390x hw does not seem to support sanitizers. Whether the cross-compilers do is unknown.
Also added a separate S390x test using GCC on real hardware, but without sanitizers.


Dead2:
Both CI failures are are just because of failing Coveralls uploads

mtl1979:
Pretty much chicken and egg problem with this PR and #2194... Have to either split fixing the duplicate coverage names as own PR or hold back reviewing one of the PRs...

PR #2189, opened 2026-03-07
CI: Reduce time spent compiling for MSAN

Dead2:
Reduce git clone download to only relevant source folders.
Reduces download from 2.3GB to 176MB.
Also reduce compile time by only compiling shared libcxx.

Libcxx compile goes from ~67sec -> ~32sec on S390x.
Time reduced from ~92sec to ~58sec on Github hosts.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -819,18 +819,24 @@ jobs:
     - name: Compile LLVM C++ libraries (MSAN)
       if: contains(matrix.name, 'MSAN')
       run: |
-        git clone --depth=1 https://github.com/llvm/llvm-project --single-branch --branch release/20.x
-        cmake -S llvm-project/runtimes -B llvm-project/build -G Ninja \
+        # Use sparse-checkout to download only the folders we need (176MB instead of 2302MB)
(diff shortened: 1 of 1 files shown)


.github/workflows/cmake.yml
```diff
@@ -819,18 +819,24 @@ jobs:
...
+          -DLIBCXX_ENABLE_STATIC=OFF \
           -DLIBCXX_INCLUDE_BENCHMARKS=OFF \
           -DLLVM_INCLUDE_TESTS=OFF \
           -DLLVM_INCLUDE_DOCS=OFF
-        cmake --build llvm-project/build -j3 -- cxx cxxabi
-        echo "LLVM_BUILD_DIR=`pwd`/llvm-project/build" >> $GITHUB_ENV
+        cmake --build build -j3 -- cxx cxxabi
+        echo "LLVM_BUILD_DIR=`pwd`/build" >> $GITHUB_ENV
       env:
+        CFLAGS: -O2
```
nmoinvaz, line 839: Is -O2 ok for MSAN? I thought it wanted like -g or something. Maybe that was for ASAN.

Dead2: This is libc, not zlib-ng. It was being compiled with -O3.

nmoinvaz (approved):

PR #2190, opened 2026-03-07
Add two groups of missing UBSAN tests

Dead2:
TODO: The fix in inftrees.c did not work

### DIFF
--- a/inftrees.c
+++ b/inftrees.c
@@ -342,7 +342,7 @@ int Z_INTERNAL zng_inflate_table(codetype type, uint16_t *lens, unsigned codes,
         } while (fill != 0);
 
         /* backwards increment the len-bit code huff */
-        rhuff += (0x8000u >> (len - 1));
+        rhuff = (uint16_t)(rhuff + (0x8000u >> (len - 1)));
         huff = zng_bitreverse16(rhuff);
 
         /* go to next symbol, update count, len */
--- a/test/test_deflate_concurrency.cc
+++ b/test/test_deflate_concurrency.cc
@@ -55,8 +55,9 @@ private:
                 continue;
             if (m_state == State::STOPPED)
                 break;
-            for (uint8_t & i: buf)
-                i++;
+            for (size_t i = 0; i < sizeof(buf); i++) {
+                buf[i] = (uint8_t)(buf[i] + 1);
+            }
         }
     }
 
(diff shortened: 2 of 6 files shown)


mtl1979:
> TODO: The fix in inftrees.c did not work

Still makes me wonder if it handles correctly the case when `len` is `0`.

inftrees.c
```diff
@@ -314,7 +314,7 @@ int Z_INTERNAL zng_inflate_table(codetype type, uint16_t *lens, unsigned codes,
         } while (fill != 0);
 
         /* backwards increment the len-bit code huff */
-        rhuff += (0x8000u >> (len - 1));
+        rhuff += (uint16_t)(0x8000u >> (len - 1));
         huff = zng_bitreverse16(rhuff);
```
mtl1979: The original code used `uint32_t` for `rhuff`with `0x80000000u` and `__builtin_bitreverse32()` when using Clang for AArch64/ARM64, which avoided implicit truncation when `len` is `0` and shift amount becomes negative.

cc: @nmoinvaz

Dead2:
Rebased. Awaiting input from @nmoinvaz

nmoinvaz:
It looks like the overflow is expected, we just need to make it explicit so that the warning goes away. Please try this:
```c
/* backwards increment the len-bit code huff */
rhuff = (uint16_t)(rhuff + (0x8000u >> (len - 1)));
huff = zng_bitreverse16(rhuff);
```

nmoinvaz (approved):

test/test_deflate_concurrency.cc
```diff
@@ -55,8 +55,9 @@ class Mutator {
                 continue;
             if (m_state == State::STOPPED)
                 break;
-            for (uint8_t & i: buf)
-                i++;
+            for (size_t i = 0; i < sizeof(buf); i++) {
```
Dead2, line 58: @iii-i  Could you verify whether this change affects the test or not?

mtl1979:
I still think `implicit-conversion` is too harsh and makes the code unnecessarily less readable. Any variable smaller than 32 bits is implicitly promoted when a temporary is needed. This is obvious when simple unsigned post-increment will trip overflow.

nmoinvaz (approved):
LGTM. Thanks for incorporating the fix I mentioned in my previous comment.

nmoinvaz:
> I still think implicit-conversion is too harsh and makes the code unnecessarily less readable. 

I think if sanitizer fails on these errors it is better because otherwise we will just hear about it from compilers that complain and warning of them and they never surface in the CI.

If these few lines cause it to pass it doesn't seem that bad to me.

Dead2:
I agree that it's annoying that some of these fixes were needed, but there were actually fewer fixes needed than I had thought.
We do get reports of implicit conversion warnings from time to time, and this will hopefully let us catch them in the PRs, and reduce the amount of issue reports.

PR #2202, opened 2026-03-09
Fix incorrect sanitizer cflag ordering and S390x fallback compiler

Dead2:
- Fix incorrect order of compiler flags when using sanitizers
- The S390x HW runner has Clang available, but the qemu fallback used in other repos uses a toolchain specifying gcc, therefore make sure we only install and use gcc instead.

@phprus Are you able to test this? I can't test that this fixes things for the out-of-home-repo fallbacks.

Fixes on of two problems in #2201

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -350,14 +350,14 @@ jobs:
             gcov-exec: ${{ github.repository == 'zlib-ng/zlib-ng' && 'llvm-cov gcov' || 's390x-linux-gnu-gcov' }}
             coverage: ${{ github.repository == 'zlib-ng/zlib-ng' && 'el10_clang_s390x_dfltcc_ubsan' || 'ubuntu_gcc_s390x_dfltcc_ubsan' }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/cmake.yml
mtl1979: This is wrong

mtl1979 (changes requested):
Missing packages for gcc and g++ cross compiler.

.github/workflows/cmake.yml
Dead2: fixed

phprus:
Ubuntu GCC S390X DFLTCC Compat - success

Ubuntu GCC S390X DFLTCC ASAN - fail - https://github.com/phprus/zlib-ng/actions/runs/22891876572/job/66416699437#step:20:596

Dead2:
> Ubuntu GCC S390X DFLTCC ASAN - fail - https://github.com/phprus/zlib-ng/actions/runs/22891876572/job/66416699437#step:20:596

Hmm, I have no ideas why that is happening :face_with_diagonal_mouth:

PR #2203, opened 2026-03-10
Replace macros with inline functions in deflate_quick.

nmoinvaz:
On -O2, Clang produces identical asm output, GCC produces 2 fewer instructions.


mtl1979 (approved):
LGTM.

Dead2 (approved):
LGTM

PR #2206, opened 2026-03-10
Clean up dead assignments in insert_match

nmoinvaz:
When 56d3d985 was reverted in b85cfdf9, it restored dead stores to `match.strstart` and `match.match_length` in `insert_match()` that have no effect since `match` is passed by value. The compiler already eliminated them via dead store elimination, verified by comparing assembly output before and after. This removes them from source to keep the code consistent with what the compiler actually emits.

Summary generated with Claude Code.


mtl1979 (approved):
LGTM

Dead2 (approved):
LGTM

PR #2209, opened 2026-03-10
[CI] Fix lint when using workflow_dispatch.

mtl1979:
* PDF files are binary files, so white-space is expected
* Remove trailing empty line from test/abi/ignore
* ignore all files in test/data, including test/data/lcet10.txt as it is expected to contain trailing white-space

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,4 +16,4 @@ jobs:
         BASE_SHA="${{ github.event.pull_request.base.sha }}"
         BASE_SHA="${BASE_SHA:-4b825dc642cb6eb9a060e54bf8d69288fbee4904}"
         git config core.whitespace blank-at-eol
-        git diff --color --check "$BASE_SHA" -- './*' ':!*.patch'
+        git diff --color --check "$BASE_SHA" -- './*' ':!*.patch' ':!*.pdf' ':!/test/data/'
--- a/test/abi/ignore
+++ b/test/abi/ignore
@@ -9,4 +9,3 @@
 # Size varies with version number
 [suppress_variable]
   name_regexp = z(|ng|libng)_(|v)string
-


.github/workflows/lint.yml
```diff
@@ -16,4 +16,4 @@ jobs:
         BASE_SHA="${{ github.event.pull_request.base.sha }}"
         BASE_SHA="${BASE_SHA:-4b825dc642cb6eb9a060e54bf8d69288fbee4904}"
         git config core.whitespace blank-at-eol
-        git diff --color --check "$BASE_SHA" -- './*' ':!*.patch'
+        git diff --color --check "$BASE_SHA" -- './*' ':!*.patch' ':!*.pdf' ':!/test/data/lcet10.txt'
```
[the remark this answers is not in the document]
mtl1979: Project maintainer didn't want the exclusion list to be too complex... This list is minimal to get the CI workflow to pass. Any new files should be investigated case by case and if deemed necessary, the files will be added to exclusion list.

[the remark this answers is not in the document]
mtl1979: When lint workflow fails with more than one file with same extension and the extension is commonly associated with binary files, whole extension should be excluded instead of excluding individual files... For text files, only reported files in test data should be excluded.

The git diff command can be run manually to check if any new files contain white-space that cause error with lint.

Dead2: Might as well change to match anything in `/test/data/`, we can expect those to be data files where the exact contents matter.

mtl1979: I thought about that, but there are PDF files outside `test/data` and ignoring by extension does catch them better. Makes more sense to not have two filters excluding same files.

Dead2: I am not saying you should remove the pdf filter. But you are already adding a file from the data folder, and we'll never care about syntax for any file in that folder, so we might as well exclude the whole folder to save us the need of modifying the filter for possible future data additions.

mtl1979: The likelihood of any new data file containing trailing white-space is a lot smaller than likelihood of new PDF file containing trailing white-space.

This PR doesn't even focus on newly added files, just files that already exist in the repository.

As far as I can remember, this is only the second time we have had need to amend the exclusion list.

Dead2 (approved):
👍 , now it also spends less compute time for those workflow_dispatch runs.

LGTM

PR #2214, opened 2026-03-11
Expand codeql testing

Dead2:
Expand codeql testing to run on multiple platforms and two configs.
Different configs and platforms expose different defines, macros and source files, improving test coverage.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -7,8 +7,8 @@ on:
 
 jobs:
   analyze:
-    name: Analyze
-    runs-on: ubuntu-latest
+    name: CodeQL (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
     permissions:
       actions: read
       contents: read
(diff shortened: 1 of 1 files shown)


mtl1979:
CI fails with bad configuration, possibly server overload or temporary failure with "Ubuntu GCC ARM HF No Neon No ARMv8 ASAN".

.github/workflows/codeql.yml
```diff
@@ -28,13 +32,22 @@ jobs:
...
+        shell: bash
+        run: |
+          cmake -B build-compat -S . -DZLIB_COMPAT=ON -DWITH_NEW_STRATEGIES=OFF -DWITH_OPTIM=OFF -DWITH_REDUCED_MEM=ON
+          cmake --build build-compat -j4
 
       - name: Perform CodeQL Analysis
         uses: github/codeql-action/analyze@v4
         with:
-          category: "/language:${{ matrix.language }}"
+          category: "/oss:${{ matrix.os }}"
```
mtl1979, line 53: CI fails with `/language:cpp` missing...

Dead2: It fails because the new configuration does not match the old configuration present in develop, thus it warns that no results were uploaded as expected.
This new configuration will replace the old configuration and then all PRs will fail in the same way until rebased on top of this change.

mtl1979:
Not sure why Windows on x86-64 isn't also tested as it is supported by CodeQL...

Dead2:
> Not sure why Windows on x86-64 isn't also tested as it is supported by CodeQL...

Feel free to add it in a followup PR. I don't know how this is done on windows.

PR #2215, opened 2026-03-11
Simplify detection and use of getauxval() and elf_aux_info()

ccawley2011:


arch/power/power_features.c
```diff
@@ -4,41 +4,32 @@
...
-#include "power_features.h"
+#include "cpu_features.h"
 
 #ifdef HAVE_SYS_AUXV_H
 #  include <sys/auxv.h>
 #endif
-#ifdef POWER_NEED_AUXVEC_H
+#ifdef HAVE_LINUX_AUXVEC_H
 #  include <linux/auxvec.h>
 #endif
```
mtl1979, line 17: This is semantically different... In standard distributions `sys/auxv.h` should duplicate information from `linux/auxvec.h`. Renaming the preprocessor macro implies it is always necessary to include both if they both exist.

cpu_features.c
```diff
@@ -23,3 +23,19 @@ Z_INTERNAL void cpu_check_features(struct cpu_features *features) {
...
+    return getauxval(type);
+}
+#elif defined(HAVE_ELF_AUX_INFO)
+#include <sys/auxv.h>
+
+Z_INTERNAL unsigned long zng_getauxval(unsigned long type) {
+    unsigned long val;
+    elf_aux_info(type, &val, sizeof(val));
+    return val;
+}
```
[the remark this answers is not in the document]
mtl1979: Implicit cast avoids truncation warning... Converting from unsigned type to signed type is generally safer as it doesn't result in unexpected large value if the original value is negative. However it is still possible to pass large unsigned value explicitly and it gets truncated to 32 bits.

arch/s390/s390_features.c
```diff
@@ -12,7 +12,9 @@
 #endif
 
 void Z_INTERNAL s390_check_features(struct s390_cpu_features *features) {
-    features->has_vx = getauxval(AT_HWCAP) & HWCAP_S390_VXRS;
+#ifdef HAVE_ZNG_GETAUXVAL
+    features->has_vx = zng_getauxval(AT_HWCAP) & HWCAP_S390_VXRS;
+#endif
```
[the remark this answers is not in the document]
ccawley2011: `s390_check_features` is called from `cpu_check_features`, which initialises the structure using memset.

cpu_features.h
```diff
@@ -42,6 +42,11 @@ struct cpu_features {
 
 void cpu_check_features(struct cpu_features *features);
 
+#if defined(HAVE_GETAUXVAL) || defined(HAVE_ELF_AUX_INFO)
+Z_INTERNAL unsigned long zng_getauxval(unsigned long type);
+#define HAVE_ZNG_GETAUXVAL
```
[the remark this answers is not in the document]
mtl1979: It's more likely that operating system has both the header and function, but not the preprocessor macros to pass to the function. This is because some operating systems have separate header that contains all the preprocessor macros for different auxiliary vectors.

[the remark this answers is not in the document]
ccawley2011: If `sys/auxv.h` is not available, the configure and CMake checks for `getauxval` and `elf_aux_info` will fail since those are also in `sys/auxv.h`, so that shouldn't happen.

arch/power/power_features.c
ccawley2011: Would it cause a problem to always include `linux/auxvec.h` when it's available?

mtl1979: It will trip static analysers as the header would be included twice.

configure
```diff
@@ -1559,45 +1569,18 @@ EOF
...
-#endif
-}
-EOF
-        if try $CC -c $CFLAGS -maltivec $test.c; then
-            HAVE_VMX=1
-            echo "Check whether VMX instructions are available ... Yes." | tee -a configure.log
-        else
-            HAVE_VMX=0
-            echo "Check whether VMX instructions are available ... No." | tee -a configure.log
-        fi
```
mtl1979, line 1583: This block tests that current CPU can have VMX... Anything from PPC7400 should be able to utilize VMX, but we support all big-endian 32-bit PowerPC processors. The first test above tests that compiler can disable VSX instructions if available, as is in case of modern GCC and Clang. This is because enabling VSX instructions by default will cause crash on 32-bit PowerPC processors, but not on 64-bit POWER processors. Modern compilers enable VSX instructions by default, instead of enabling them when `-mvsx` is explicitly specified.

There are toolchains/compilers that pass all the previous checks, but fail this one. This is also why earlier `getauxval` checks pass `0` as the auxiliary vector number instead of one of the preprocessor macros for specific auxiliary vector.

cpu_features.c
```diff
@@ -23,3 +23,19 @@ Z_INTERNAL void cpu_check_features(struct cpu_features *features) {
     loongarch_check_features(&features->loongarch);
 #endif
 }
+
+#if defined(__linux__) && defined(HAVE_SYS_AUXV_H)
```
nmoinvaz, line 27: Suggestion for code clarity:
```c
#ifdef HAVE_SYS_AUXV_H
#include <sys/auxv.h>

Z_INTERNAL unsigned long zng_getauxval(unsigned long type) {
#if defined(__linux__)
    return getauxval(type);
#elif defined(__FreeBSD__) || defined(__OpenBSD__)
    unsigned long val = 0;
    elf_aux_info(type, &val, sizeof(val));
    return val;
#endif
}
#endif
```

cpu_features.h
```diff
@@ -42,6 +42,11 @@ struct cpu_features {
 
 void cpu_check_features(struct cpu_features *features);
 
+#if (defined(__linux__) || defined(__FreeBSD__) || defined(__OpenBSD__)) && defined(HAVE_SYS_AUXV_H)
```
nmoinvaz, line 45: I think `(defined(__linux__) || defined(__FreeBSD__) || defined(__OpenBSD__))` is illrevelant here. We can just probably get away with `ifdef HAVE_SYS_AUXV_H`.

nmoinvaz (commented):
There is a lot in this PR, I suggest breaking out the first commit, the "wrapper" one, into a new PR. It is fairly easy to review and merge and it is a good idea.

CMakeLists.txt
```diff
@@ -523,6 +523,20 @@ if(NOT HAVE_STRERROR)
 endif()
 set(CMAKE_REQUIRED_FLAGS)
 
+set(CMAKE_REQUIRED_FLAGS "${ADDITIONAL_CHECK_FLAGS}")
```
nmoinvaz, line 526: This two code blocks should probably be part of the first commit.

mtl1979:
> There is a lot in this PR, I suggest breaking out the first commit, the "wrapper" one, into a new PR. It is fairly easy to review and merge and it is a good idea.

It's true that there is some controversial or complex parts in this PR that might benefit from breaking the PR to two parts for easier reviewing, but that's why I'm the resident bitch that makes sure any controversial parts get cleaned up before merging... 

I might not know everything about recent additions to POWER architecture but I know a lot about older 32-bit processors and different toolchains, especially when it comes to code that interacts with Linux kernel.

PR #2219, opened 2026-03-12
Add copy fallback for Adler32 ARM when building with no-unaligned-access

nmoinvaz:


Dead2 (approved):
LGTM

ccawley2011:
I'm assuming this way is measured to be faster? I would have assumed that the Adler32 function would be OK since it primarily uses NEON loads and stores rather than integer ones.

nmoinvaz:
> I'm assuming this way is measured to be faster? I would have assumed that the Adler32 function would be OK since it primarily uses NEON loads and stores rather than integer ones.

Using `no-unaligned-access` with this PR, on RPi5 it is faster, on M3 it is slower. Although I don't know why anybody would be building with that flag on M3.

Here is my results on RPi5 and M3:
https://gist.github.com/nmoinvaz/d5734003070f7a05183e82ba285b4867

PR #2221, opened 2026-03-12
Split apart long functions in (v)pclmulqdq

nmoinvaz:


arch/x86/crc32_pclmulqdq_tpl.h
```diff
@@ -140,11 +145,85 @@ static inline void fold_16(__m512i *zmm_crc0, __m512i *zmm_crc1, __m512i *zmm_cr
...
+                                 __m128i *xmm_crc1, __m128i *xmm_crc2, __m128i *xmm_crc3, const int COPY) {
+    *len -= 256;
+
+    __m512i zmm_crc0, zmm_crc1, zmm_crc2, zmm_crc3;
+    __m512i zmm_t0, zmm_t1, zmm_t2, zmm_t3;
+    __m512i z_low0, z_high0;
+    const __m512i zmm_fold4 = _mm512_set4_epi32(
+        0x00000001, 0x54442bd4, 0x00000001, 0xc6e41596);
+    const __m512i zmm_fold16 = _mm512_set4_epi32(
+        0x00000001, 0x1542778a, 0x00000001, 0x322d1430);
```
mtl1979: Do not exceed 120 characters per line, no need to split.

nmoinvaz: fixed

Dead2:
I wonder where the compiled size difference comes from..
```
 172444    1328       8  173780   2a6d4 libz-ng.so.2
```
``` 
 173092    1328       8  174428   2a95c libz-ng.so.2
 ```
Performance numbers are inconclusive. Deflatebench tended to be a little faster, while gbench was a little slower.
Both within margins of error though, so I'll just call it a draw.

nmoinvaz:
We don't force inlining on any of the functions in that file so the compiler is free to do whatever.

Dead2 (approved):

PR #2222, opened 2026-03-12
CI: Add guard against doing lots of testing of clearly broken code

Dead2:
Use OSB workflow as an initial test before queueing all the other tests,
this makes sure we don't spend a lot of CI time testing something that won't even build.


mtl1979:
This is a good idea... I had similar idea at works with `Lint` being the first and then followed by others...

nmoinvaz (approved):

PR #2224, opened 2026-03-14
crc32: use may_alias for chorba buffers

cl2t:
This reworks the large Chorba CRC32 path to use explicit may_alias buffer
  views instead of explicit zeroing.

  The goal is to make the overlapping accesses visible to the compiler so it
  does not try to eliminate or reorder writes in a way that can trigger
  uninitialized-value diagnostics, without paying the performance cost of
  memset-based zeroing.

  Validation:
  - cmake --build build -j4
  - ./gtest_zlib --gtest_filter='inflate.adler32:crc32_large_buf.*'

### DIFF
--- a/arch/generic/crc32_chorba_c.c
+++ b/arch/generic/crc32_chorba_c.c
@@ -75,6 +75,10 @@ Z_INTERNAL uint32_t crc32_chorba_118960_nondestructive(uint32_t crc, const uint8
     chorba_word_t next22 = 0;
     crc = 0;
 
+    /* Make the temporary state explicitly zero-initialized so later XOR updates
+     * do not depend on the compiler inferring the staged write coverage. */
+    memset(bitbuffer, 0, bitbuffer_size_bytes);
+
     // do a first pass to zero out bitbuffer
     for (; i < (14848 * sizeof(chorba_word_t)); i += (32 * sizeof(chorba_word_t))) {
         chorba_word_t in1, in2, in3, in4, in5, in6, in7, in8;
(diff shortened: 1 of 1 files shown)


mtl1979:
Explicit zeroing at start makes the Chorba version a lot slower. This was already discussed earlier. Marking the pointers explicitly aliasing each other guarantees that compiler will not try to eliminate or reorder writes so that reads would try to access uninitialized array elements.

cl2t:
> Explicit zeroing at start makes the Chorba version a lot slower. This was already discussed earlier. Marking the pointers explicitly aliasing each other guarantees that compiler will not try to eliminate or reorder writes so that reads would try to access uninitialized array elements.

Thanks, that makes sense. I dropped the explicit zeroing and reworked the large
  Chorba path to use explicit may_alias buffer views instead, so the compiler sees
  the overlapping accesses without paying the memset cost.

  I also re-ran the local validation:
  - cmake --build build -j4
  - ./gtest_zlib --gtest_filter='inflate.adler32:crc32_large_buf.*'

cl2t:
@mtl1979 Updated the PR to follow your suggestion: the explicit zeroing is removed
  and the large Chorba path now uses explicit may_alias buffer views.

mtl1979:
> @mtl1979 Updated the PR to follow your suggestion: the explicit zeroing is removed and the large Chorba path now uses explicit may_alias buffer views.

PR title and description (first comment) still needs to be updated...

cl2t:
@mtl1979 Sure！I've already updated them.

Dead2:
@samrussell This looks good to me, do you have any comments?

samrussell:
no comment on the C language stuff, if it works then it works :)

Dead2 (approved):

PR #2225, opened 2026-03-14
Use zng_memread_* and zng_memwrite_* in more places

ccawley2011:


zmemory.h
```diff
@@ -69,6 +69,39 @@ static inline void zng_memwrite_8(void *ptr, uint64_t val) {
 #endif
 }
 
+static inline void zng_memcpy_2(void *dst, const void *src) {
+#if defined(HAVE_MAY_ALIAS)
+    zng_memwrite_2(ptr, zng_memread_2(src));
+#else
+    memcpy(dst, src, 2);
+#endif
```
mtl1979, line 77: Duplicated preprocessor test, `zng_memwrite_2` already checks for `HAVE_MAY_ALIAS`.

zmemory.h
```diff
@@ -69,6 +69,39 @@ static inline void zng_memwrite_8(void *ptr, uint64_t val) {
...
+}
+
+static inline void zng_memcpy_4(void *dst, const void *src) {
+#if defined(HAVE_MAY_ALIAS)
+    zng_memwrite_4(ptr, zng_memread_4(src));
+#else
+    memcpy(dst, src, 4);
+#endif
```
mtl1979, line 85: Duplicated preprocessor test, `zng_memwrite_4` already checks for `HAVE_MAY_ALIAS`.

zmemory.h
```diff
@@ -69,6 +69,39 @@ static inline void zng_memwrite_8(void *ptr, uint64_t val) {
...
+}
+
+static inline void zng_memcpy_8(void *dst, const void *src) {
+#if defined(HAVE_MAY_ALIAS)
+    zng_memwrite_8(ptr, zng_memread_8(src));
+#else
+    memcpy(dst, src, 8);
+#endif
```
mtl1979, line 93: Duplicated preprocessor test, `zng_memwrite_8` already checks for `HAVE_MAY_ALIAS`.

zmemory.h
```diff
@@ -69,6 +69,39 @@ static inline void zng_memwrite_8(void *ptr, uint64_t val) {
...
+}
+
+static inline void zng_memcpy_16(void *dst, const void *src) {
+#if defined(HAVE_MAY_ALIAS)
+    zng_memwrite_8(ptr + 0, zng_memread_8(src + 0));
+    zng_memwrite_8(ptr + 8, zng_memread_8(src + 8));
```
mtl1979: `+ 0` here can be replaced with just white-space... `ptr` should be `dst` like the bot already said.

mtl1979 (changes requested):
Fails to compile, unnecessary duplicated preprocessor tests.

zmemory.h
ccawley2011: Fixed.

zmemory.h
ccawley2011: It's not strictly identical since when `HAVE_MAY_ALIAS` is undefined, `zng_memwrite_8(ptr, zng_memread_8(src));` expands to two `memcpy` calls instead of just one.

mtl1979: Nope... Single `memcpy` when undefined. Only the 16 byte version is different.

zmemory.h
```diff
@@ -69,6 +69,39 @@ static inline void zng_memwrite_8(void *ptr, uint64_t val) {
...
+}
+
+static inline void zng_memcpy_16(void *dst, const void *src) {
+#if defined(HAVE_MAY_ALIAS)
+    zng_memwrite_8((uint8_t *)dst,     zng_memread_8((uint8_t *)src    ));
+    zng_memwrite_8((uint8_t *)dst + 8, zng_memread_8((uint8_t *)src + 8));
+#else
+    memcpy(dst, src, 16);
+#endif
+}
```
phprus, line 103: The `zng_memcpy_16` function may be slower than `memcpy` because it will use four `mov` instructions instead of two `movdqu`/`movups`:

```
zng_memcpy_16:
        mov     rax, QWORD PTR [rsi]
        mov     QWORD PTR [rdi], rax
        mov     rax, QWORD PTR [rsi+8]
        mov     QWORD PTR [rdi+8], rax
        ret
raw_memcpy_16:
        movdqu  xmm0, XMMWORD PTR [rsi]
        movups  XMMWORD PTR [rdi], xmm0
        ret
```

https://godbolt.org/z/ToaT3fexs (gcc -O3)

mtl1979: Reading or writing 16 bytes in one pass isn't supported without using vector registers, which means it doesn't work with all compilers... I know GCC and Clang can do it without intrinsics, but for example Visual C++ can't.

We have to remember these functions are defined in a header which is not guaranteed to be compiled with vector instructions enabled.

phprus: I didn't understand you.

MSVC doesn't support ``__may_alias__`` and always uses `memcpy` (and two SSE memory ops without any intrinsics).

For gcc/clang, the new `zng_memcpy_16` implementation will use four memory ops instead of two (without any intrinsics).

See: https://godbolt.org/z/rYK6ref19

mtl1979: With some GCC/Clang versions one can use `uint128_t` type to copy 16 bytes in one pass. This supports aliasing.

Because Visual C++ is a C++ compiler, all named variables are allowed to alias by default and temporary variables are not allowed to alias.

Temporary variables in this sense are created by either explicit or implicit cast without assignment.

phprus: gcc automatically uses 16-byte copying when using memcpy.

ccawley2011: If I modify the Compiler Explorer example to use the `restrict` keyword, it generates identical code for both functions. Would that be an acceptable solution?

zmemory.h
ccawley2011: [This Compiler Explorer example](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,selection:(endColumn:45,endLineNumber:31,positionColumn:45,positionLineNumber:31,selectionStartColumn:45,selectionStartLineNumber:31,startColumn:45,startLineNumber:31),source:'%23include+%3Cstring.h%3E%0A%23include+%3Cstdint.h%3E%0A%0A%23if+defined(__GNUC__)+%26%26+(__GNUC__+%3E%3D+4)%0A//%23++define+HAVE_MAY_ALIAS%0A%23endif%0A%0Astatic+inline+uint64_t+zng_memread_8(const+void+*ptr)+%7B%0A%23if+defined(HAVE_MAY_ALIAS)%0A++++typedef+struct+%7B+uint64_t+val%3B+%7D+__attribute__+((__packed__,+__may_alias__))+unaligned_uint64_t%3B%0A++++return+((const+unaligned_uint64_t+*)ptr)-%3Eval%3B%0A%23else%0A++++uint64_t+val%3B%0A++++memcpy(%26val,+ptr,+sizeof(val))%3B%0A++++return+val%3B%0A%23endif%0A%7D%0A%0A%0A%0Astatic+inline+void+zng_memwrite_8(void+*ptr,+uint64_t+val)+%7B%0A%23if+defined(HAVE_MAY_ALIAS)%0A++++typedef+struct+%7B+uint64_t+val%3B+%7D+__attribute__+((__packed__,+__may_alias__))+unaligned_uint64_t%3B%0A++++((unaligned_uint64_t+*)ptr)-%3Eval+%3D+val%3B%0A%23else%0A++++memcpy(ptr,+%26val,+sizeof(val))%3B%0A%23endif%0A%7D%0A%0Avoid+zng_memcpy_8(void+*restrict+dst,+const+void+*restrict+src)+%7B%0A++++zng_memwrite_8(dst,+zng_memread_8(src))%3B%0A%7D%0A%0A%0Avoid+raw_memcpy_8(void+*restrict+dst,+const+void+*restrict+src)+%7B%0A++++memcpy(dst,+src,+8)%3B%0A%7D'),l:'5',n:'0',o:'C+source+%231',t:'0')),k:51.78797339758173,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:carmug1520,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'1',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:___c,libs:!(),options:'-O3',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+ARM+GCC+15.2.0+(unknown-eabi)+(Editor+%231)',t:'0')),k:48.21202660241826,l:'4',m:99.99999999999999,n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) shows that GCC on ARM does generate two separate `memcpy` calls - one for the `zng_memread_8` call and one for the `zng_memwrite_8` call.

zmemory.h
phprus: I think yes.
Using `restrict` keyword (`Z_RESTRICT` macro) is a good solution.

mtl1979: > I think yes.
> 
> Using `restrict` keyword (`Z_RESTRICT` macro) is a good solution.

`restrict` and `__may_alias__` are mutually exclusive. The assembly difference is expected. The crucial part here is that reads and writes can't be interleaved, so reads for `src` must happen before writes to `dst` even if it requires using multiple general-purpose registers or temporary buffer in memory.

phprus: I don't understand why we should replace one 16-byte copy with two 8-byte copies.

`memcpy` implementation - one  16-byte copy on x86_64 (gcc/clang do this optimization themselves).
`__may_alias`__ implementation - two 8-byte copies.

phprus: > `restrict` and `__may_alias__` are mutually exclusive.

`restrict` for `zng_memcpy_*` function arguments: https://godbolt.org/z/oqe78Preh - one 16-byte copy on x86_64

mtl1979: > I don't understand why we should replace one 16-byte copy with two 8-byte copies.
> 
> `memcpy` implementation - one 16-byte copy on x86_64 (gcc/clang do this optimization themselves). `__may_alias`__ implementation - two 8-byte copies.

`zng_memread_16` and `zng_memwrite_16` are still unimplemented as those can't be written without resorting to compiler-specific implementations. The code has to work with and without optimizations.

Little experimenting: https://godbolt.org/z/TrzWfEPPa

phprus: Please see the godbolt link: https://godbolt.org/z/znYcG5Ya6

For 16-byte copying `zng_memread_16` and `zng_memwrite_16` are not needed. This optimization is performed by the compiler.

ccawley2011: > Please see the godbolt links. For 16-byte copying `zng_memread_16` and `zng_memwrite_16` are not needed. This optimization is performed by the compiler.

It's not performed on all architectures - particularly when unaligned memory access is not supported. The Compiler Explorer example in [this comment](https://github.com/zlib-ng/zlib-ng/pull/2225#discussion_r2936860070) demonstrates this on 32-bit ARM.

phprus: My suggestion:

```c
static inline void zng_memcpy_2(void *Z_RESTRICT dst, const void *Z_RESTRICT src) {
#if defined(HAVE_MAY_ALIAS)
    zng_memwrite_2(dst, zng_memread_2(src));
#else
    memcpy(dst, src, 2);
#endif
}

static inline void zng_memcpy_4(void *Z_RESTRICT dst, const void *Z_RESTRICT src) {
#if defined(HAVE_MAY_ALIAS)
    zng_memwrite_4(dst, zng_memread_4(src));
#else
    memcpy(dst, src, 4);
#endif
}

static inline void zng_memcpy_8(void *Z_RESTRICT dst, const void *Z_RESTRICT src) {
#if defined(HAVE_MAY_ALIAS)
    zng_memwrite_8(dst, zng_memread_8(src));
#else
    memcpy(dst, src, 8);
#endif
}

static inline void zng_memcpy_16(void *Z_RESTRICT dst, const void *Z_RESTRICT src) {
#if defined(HAVE_MAY_ALIAS)
    zng_memwrite_8((uint8_t *)dst,     zng_memread_8((uint8_t *)src    ));
    zng_memwrite_8((uint8_t *)dst + 8, zng_memread_8((uint8_t *)src + 8));
#else
    memcpy(dst, src, 16);
#endif
}
```

This will improve code generation on x86_64.

mtl1979: > My suggestion:
> 
> ```c
> static inline void zng_memcpy_2(void *Z_RESTRICT dst, const void *Z_RESTRICT src) {
> #if defined(HAVE_MAY_ALIAS)
>     zng_memwrite_2(dst, zng_memread_2(src));
> #else
>     memcpy(dst, src, 2);
> #endif
> }
> 
> static inline void zng_memcpy_4(void *Z_RESTRICT dst, const void *Z_RESTRICT src) {
> #if defined(HAVE_MAY_ALIAS)
>     zng_memwrite_4(dst, zng_memread_4(src));
> #else
>     memcpy(dst, src, 4);
> #endif
> }
> 
> static inline void zng_memcpy_8(void *Z_RESTRICT dst, const void *Z_RESTRICT src) {
> #if defined(HAVE_MAY_ALIAS)
>     zng_memwrite_8(dst, zng_memread_8(src));
> #else
>     memcpy(dst, src, 8);
> #endif
> }
> 
> static inline void zng_memcpy_16(void *Z_RESTRICT dst, const void *Z_RESTRICT src) {
> #if defined(HAVE_MAY_ALIAS)
>     zng_memwrite_8((uint8_t *)dst,     zng_memread_8((uint8_t *)src    ));
>     zng_memwrite_8((uint8_t *)dst + 8, zng_memread_8((uint8_t *)src + 8));
> #else
>     memcpy(dst, src, 16);
> #endif
> }
> ```
> 
> This will improve code generation on x86_64.

It's still dead code, as `Z_RESTRICT` and `__may_alias__` attribute contradict with each other.

mtl1979: > Please see the godbolt link: https://godbolt.org/z/znYcG5Ya6
> 
> For 16-byte copying `zng_memread_16` and `zng_memwrite_16` are not needed. This optimization is performed by the compiler.

Your argument solely depends on telling the compiler both that the `src` and `dst` pointers can and can't overlap... This eliminates any checking of overlap, thus making the code both potentially invalid and unsafe. Not all bad code results in compiler warning or error, as it would require run-time or JIT static analysis to detect that two attributes are mutually exclusive. It's up to compiler to determine which one of the conflicting attributes it will honour when generating the assembler code.

phprus: > This eliminates any checking of overlap, thus making the code both potentially invalid and unsafe

This is impossible, because `memcpy` require nonoverlapd src and dst memory regions.

phprus: > It's still dead code, as `Z_RESTRICT` and `__may_alias__` attribute contradict with each other.

Using `restrict` for function arguments, gcc and clang generate better code on x86_64. I agree that `restrict` and `__may_alias__` conflict, but `restrict` lets the compiler know that `src` and `dst` don't overlap.

In this case, I don't see the point in replacing `memcpy` with `zng_memcpy_*`.

mtl1979: > > This eliminates any checking of overlap, thus making the code both potentially invalid and unsafe
> 
> This is impossible, because `memcpy` require nonoverlapd src and dst memory regions.

No, it doesn't... `memmove` semantics say that it should behave like intermediate buffer is used when regions overlap, but `memcpy` semantics give no guarantee whatsoever what the implementation does if the size is not power of 2. In most cases it will do either byte-by-byte copy (if unaligned access is not allowed), or load as many bytes in widest usable register and then store that to destination address before continuing with the remaining bytes; In our cases the size is power of 2, so it's safe.

`__may_alias__` in this context means that memory pointed by `src` can change without any apparent reason, so that means all reads must be written to memory as soon as possible and contents should not stay for longer than absolutely required in register(s), just like if `volatile` qualifier is used. Only difference between `__may_alias__` and `volatile` is that `volatile` allows the another variable to be in completely different process.

phprus: > No, it doesn't... 

`memcpy` requires that buffers not overlap.
Since C99, this has been a standard requirement (the arguments are marked `restrict`).
https://en.cppreference.com/w/c/string/byte/memcpy.html

The code previously used `memcpy`.



> `__may_alias__` in this context...

The original PR said that `__may_alias__` was used only for ensures inlining:

https://github.com/zlib-ng/zlib-ng/pull/1548
> Using the may_alias attribute ensures that the code for reading and writing one byte at a time is inlined

mtl1979: > The original PR said that `__may_alias__` was used only for ensures inlining:

PR descriptions are not facts, they are either personal opinions or observations... What compiler documentation says should be considered facts, or the compiler is buggy... Not all compilers follow C standards, Visual Studio is one good example, GCC is another. Most standards are written based on how compilers commonly handle things... For example, support for 128 bit scalar types isn't part of C standard and likely will not be part of any standard before C26 is ratified, but GCC has had it for many many years and a few processors already can do 128-bit arithmetic.

What `memcpy` prototype says doesn't change the fact that `__may_alias__` and `restrict` keywords are still mutually exclusive even though their effect scope is not exactly same. 'restrict' keyword is invalid if parent scope already has two variables that point to same memory region. Semantics of 'restrict' keyword don't require that the two variables point to same exact memory location, just that they point to same allocation.

Dead2:
@ccawley2011 This needs a rebase

zmemory.h
mtl1979: That Compiler Explorer output shows the expected result, when you tell it that source and destination might overlap, it will use temporary "variable" to avoid corrupting source. That's consistent with how `HAVE_MAY_ALIAS` is documented. If you uncomment the define, it will do 8 zero-extended loads as unaligned access is not allowed by the target processor because unaligned loads need targeting at least ARMv6 and no target processor whatsoever was specified.

This just means that the two options here are not equal and that the code is bad by design like I already pointed out earlier.

nmoinvaz:
CodeRabbit indicates you forgot to add Z_RESTRICT.

mtl1979:
> CodeRabbit indicates you forgot to add Z_RESTRICT.

Not adding `Z_RESTRICT` is at the moment correct... The code needs complete rewrite if `Z_RESTRICT` is added. Depending on `__may_alias__` for inlining short copies is just plain facepalm and is compiler-dependent, so has no effect with for example Visual C++.

GitHub has reached limit for inline comments, so we should avoid those until the core issue is resolved, otherwise it will be very hard for people to track what was already said, because GitHub only shows first comments, not most recent ones.

For me, it still looks like there is too much unnecessary preprocessor tests and those can only be eliminated by dropping all checks for `__may_alias__` support, as `restrict` keyword is more widely supported.

PR #2232, opened 2026-03-19
Replace python3 with python in GitHub Actions workflows

nmoinvaz:
Using python command prevents it from using a different python installation that may exist on the machine. setup-python should redirect python command for us.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -908,10 +908,10 @@ jobs:
       if: matrix.coverage
       shell: bash
       run: |
-        python3 -u -m venv ./venv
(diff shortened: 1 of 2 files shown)


mtl1979:
This seems incomplete and fragile...

nmoinvaz:
> This seems incomplete and fragile...

Why?

mtl1979:
> > This seems incomplete and fragile...
> 
> Why?

The workflow file(s) itself don't use `setup-python` action like the bot already mentioned... This means the container build script must always run it...

nmoinvaz:
https://github.com/zlib-ng/zlib-ng/pull/1744

PR #2234, opened 2026-03-22
Use `__attribute__((constructor))` to initialize the functable

phprus:
As i see the ClickHouse developers are concerned that the current functable pointers initialize may not be thread-safe and leads to TSAN errors (atomic read functions are not used). They are adding a patch that uses `__attribute__((constructor))` to initialize functable.

https://github.com/ClickHouse/zlib-ng/pull/18
https://github.com/ClickHouse/zlib-ng/commit/a2fbeffdc30a8b0ce6d54ee31208e2688eac4c9f

This is false positive TSAN error, but I propose adding early functable initialization to zlib-ng if the compiler supports `__attribute__((constructor))`.

### DIFF
--- a/functable.c
+++ b/functable.c
@@ -472,6 +472,12 @@ static int init_functable(void) {
     return Z_OK;
 }
 
+#if !defined(_WIN32) && (defined(__GNUC__) || defined(__clang__))
+static void __attribute__((constructor)) functable_constructor(void) {
+    FUNCTABLE_INIT_ABORT;
+}
+#endif
+
 /* stub functions */
 static int force_init_stub(void) {
     return init_functable();


mtl1979:
We shouldn't really care what third-party repositories to try meddle with our code... A lot of `functable` code has been mutilated since early days trying to fix non-issues, but causing more harm and breaking things like code coverage. Last time I suggested we just revert all the changes and go back to basics, but that idea was scrapped without even discussion.

Whole `functable` was one of the things I designed during the early days to be as safe as possible. The first version used the static initialization using the stub functions to make sure even reads were safe.

There will always be false positives with sanitizers, and that is one of the reasons we disable some sanitizer options but keep the others enabled.

functable.c
```diff
@@ -471,6 +471,14 @@ static int init_functable(void) {
     return Z_OK;
 }
 
+#if !defined(_WIN32) && (defined(__GNUC__) || defined(__clang__))
+
+static void __attribute__((constructor)) functable_constructor(void) {
+    FUNCTABLE_INIT_ABORT;
+}
+
+#endif
```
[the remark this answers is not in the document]
mtl1979: We don't use explicit atomic functions as not all processors we support have atomic instructions. Neither zlib or zlib-ng guarantees to be thread-safe. There is no reason whatsoever why we should implement partial fix for false positive as zlib-ng is used for testing new compiler versions. For all applicable sanitizers we already have CI tests.

In the past we attempted to use TLS to add thread safety but it was dropped as not all compilers support TLS properly.

In zlib-ng, all threads are guaranteed to return same functable entry pointer as kernels use lowest supported instruction set to build the CPU feature structures. Even if two cores or threads attempt to modify same functable entry, it should only result in cache invalidation, not unpredictable garbage.

sfc-gh-bhannel:
> Neither zlib or zlib-ng guarantees to be thread-safe. 

Really? The zlib-ng FAQ, copied from zlib, [say that it is guaranteed to be thread safe](https://github.com/zlib-ng/zlib-ng/blob/develop/FAQ.zlib#L157).

> This is false positive TSAN error

I'm not sure it is, in the strictest sense. It is unlikely that the current implementation causes wrong results on platforms I know of, but I think that it is Undefined Behavior per the [C11 standard](https://port70.net/~nsz/c/c11/n1570.html#5.1.2.4p25). If thread A and thread B start initialization around the same time, then they may both write and read with no synchronization between them.

IMHO, the approach [here](https://github.com/ClickHouse/zlib-ng/pull/18) is the right one - initialize the functable eagerly with `__attribute__((constructor))` (akin to static initialization) so you don't need to worry about atomicity of reads or races during initialization.

Dead2:
> > Neither zlib or zlib-ng guarantees to be thread-safe.
> 
> Really? The zlib-ng FAQ, copied from zlib, [say that it is guaranteed to be thread safe](https://github.com/zlib-ng/zlib-ng/blob/develop/FAQ.zlib#L157).
> 
> > This is false positive TSAN error
> 
> I'm not sure it is, in the strictest sense. It is unlikely that the current implementation causes wrong results on platforms I know of, but I think that it is Undefined Behavior per the [C11 standard](https://port70.net/~nsz/c/c11/n1570.html#5.1.2.4p25). If thread A and thread B start initialization around the same time, then they may both write and read with no synchronization between them.
> 
> IMHO, the approach [here](https://github.com/ClickHouse/zlib-ng/pull/18) is the right one - initialize the functable eagerly with `__attribute__((constructor))` (akin to static initialization) so you don't need to worry about atomicity of reads or races during initialization.

Lost my first, long, edition of this comment to a browser crash, I'll try again, but a bit more briefly this time.

- I think you are probably right, we don't support multiple threads working on a stream, but functable lives outside of streams.
- I had not seen the "new" clickhouse approach, it does look nice, but not very portable. (Does it even compile on msvc?)
- We officially require C11, but we also try to keep portability for other compilers, even OpenWatcom, if possible. They are on their own, we don't actively support or test such setups, but we do recognize that there can be needs outside of the officially supported compilers/platforms, so we try to accommodate them as long as it does not require too invasive changes.

I would not be opposed to defaulting to using constructors for compilers where it is supported, as long as it does not remove the fallback solution. I figure a lot could be done using #ifdef and macros that either point to functions or are empty.

functable.c
```diff
@@ -471,6 +471,14 @@ static int init_functable(void) {
     return Z_OK;
 }
 
+#if !defined(_WIN32) && (defined(__GNUC__) || defined(__clang__))
+
```
nmoinvaz: nitpick: no need for the extra lines around function imo.

phprus: Extra lines removed.

mtl1979:
> > Neither zlib or zlib-ng guarantees to be thread-safe. 
> 
> 
> 
> Really? The zlib-ng FAQ, copied from zlib, [say that it is guaranteed to be thread safe](https://github.com/zlib-ng/zlib-ng/blob/develop/FAQ.zlib#L157).
> 

Please read the whole text before commenting. No C application or library can be thread-safe as there is no locking or other synchronization between threads.

mtl1979:
> - I had not seen the "new" clickhouse approach, it does look nice, but not very portable. (Does it even compile on msvc?)

Visual C++ and Windows have own way of doing it when zlib-ng is built as a DLL or shared library. Using static libraries is quite uncommon on Windows, so not sure how it can/should be done in that scenario.

Generic way is still to use function whose return value is assigned to global variable. This doesn't differ much from the original version where the functable was initialized with pointers to all the stub functions.

Dead2 (approved):
This is simple and harmless, LGTM

PR #2236, opened 2026-03-24
Replace conditional byte swapping with portable host/LE conversion

amaanq:
### Summary

This is just a backport of a565cfa9 from develop to 2.3.x. It fixes issues for me on aarch64be namely.

I've cross-compiled zlib-ng 2.3.3 for aarch64_be, and ran the test suite under qemu, it works great.

There were a couple of conflicts to resolve, namely adapting to use `quick_insert_string` instead of `quick_insert_value`.

### DIFF
--- a/arch/generic/crc32_braid_c.c
+++ b/arch/generic/crc32_braid_c.c
@@ -97,7 +97,7 @@ Z_INTERNAL uint32_t crc32_braid_internal(uint32_t c, const uint8_t *buf, size_t 
 #endif
 #endif
         /* Initialize the CRC for each braid. */
-        crc0 = ZSWAPWORD(c);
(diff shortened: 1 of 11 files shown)


mtl1979:
@Dead2 will pick any fixes when the next stable release is prepared... Knowing the exact issue that this PR resolves would help in deciding if this patch qualifies for 2.3.x branch as we are already planning first 2.4.x release currently.

amaanq:
Awesome, thanks for response!

mtl1979:
> Awesome, thanks for response!

I try to make sure we continue to support big-endian processors and don't make the code unnecessarily slow on them. On my bi-endian setup the code is about 5 times faster as big-endian than little-endian but there has been more patches to improve speed on little-endian architectures.

PR #2245, opened 2026-04-08
Add --benchmark_cooldown flag to mitigate thermal throttling

nmoinvaz:
Adds a `--benchmark_cooldown=<seconds>` flag that inserts a sleep between benchmark families. This helps produce consistent results on systems where sustained workloads cause thermal throttling and CPU frequency scaling.

### DIFF
--- a/test/benchmarks/README.md
+++ b/test/benchmarks/README.md
@@ -5,7 +5,7 @@ These benchmarks are written using [Google Benchmark](https://github.com/google/
 
 To increase the number of times each benchmark iteration is run use:
 
-```
+```sh
 --benchmark_repetitions=20
 ```
 
@@ -13,8 +13,16 @@ To increase the number of times each benchmark iteration is run use:
 
 To filter out which benchmarks are performed use:
 
-```
+```sh
 --benchmark_filter="adler32*"
+```
+
+*Cooldown*
+
+To insert a cooldown sleep between benchmark families to mitigate thermal throttling:
+
+```sh
+--benchmark_cooldown=3
 ```
 
 There are two different benchmarks, micro and macro.
--- a/test/benchmarks/benchmark_main.cc
+++ b/test/benchmarks/benchmark_main.cc
@@ -1,9 +1,19 @@
 /* benchmark_main.cc -- benchmark suite main entry point
- * Copyright (C) 2022 Nathan Moinvaziri
+ * Copyright (C) 2026 Nathan Moinvaziri
  * For conditions of distribution and use, see copyright notice in zlib.h
  */
 
+#include <stdint.h>
 #include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#ifdef _WIN32
+#  define NOMINMAX
+#  include <windows.h>
+#else
+#  include <unistd.h>
+#endif
 
 #include <benchmark/benchmark.h>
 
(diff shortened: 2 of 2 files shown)


test/benchmarks/README.md
```diff
@@ -17,6 +17,14 @@ To filter out which benchmarks are performed use:
 --benchmark_filter="adler32*"
 ```
 
+*Cooldown*
+
+To insert a cooldown sleep between benchmark runs to mitigate thermal throttling:
+
+```
+--benchmark_cooldown=3
+```
```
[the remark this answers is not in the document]
nmoinvaz: fixed

test/benchmarks/benchmark_main.cc
```diff
@@ -18,15 +26,68 @@ extern "C" {
...
 #endif
 
+    for (int i = 1; i < argc; i++) {
+        if (strncmp(argv[i], "--benchmark_cooldown=", 20) == 0) {
+            cooldown_secs = atoi(argv[i] + 20);
+            break;
+        }
+    }
+
     ::benchmark::Initialize(&argc, argv);
```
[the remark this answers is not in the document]
nmoinvaz: It does not warn and does not care.

test/benchmarks/benchmark_main.cc
```diff
@@ -18,15 +26,68 @@ extern "C" {
...
     cpu_check_features(&test_cpu_features);
 #  endif
 #endif
 
+    for (int i = 1; i < argc; i++) {
+        if (strncmp(argv[i], "--benchmark_cooldown=", 20) == 0) {
+            cooldown_secs = atoi(argv[i] + 20);
+            break;
+        }
+    }
```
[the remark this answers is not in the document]
nmoinvaz: Fixed.

test/benchmarks/benchmark_main.cc
```diff
@@ -18,15 +26,68 @@ extern "C" {
...
 #  ifndef DISABLE_RUNTIME_CPU_DETECTION
     cpu_check_features(&test_cpu_features);
 #  endif
 #endif
 
+    for (int i = 1; i < argc; i++) {
+        if (strncmp(argv[i], "--benchmark_cooldown=", 21) == 0) {
+            cooldown_secs = atoi(argv[i] + 21);
+            break;
+        }
```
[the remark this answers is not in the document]
nmoinvaz: I think it is fine for this tool.

arch/s390/self-hosted-builder/actions-runner-rebuild.sh
```diff
@@ -24,6 +24,9 @@ podman container rm gaplib-actions-runner || true
 # Delete old image
 podman image rm localhost/zlib-ng/actions-runner || true
 
+# Prune all unused podman data
+podman system prune -f || true
+
```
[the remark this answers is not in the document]
nmoinvaz: I've fixed this was due to a bad code merge.

test/benchmarks/benchmark_main.cc
```diff
@@ -18,15 +26,69 @@ extern "C" {
...
+
+    void ReportRunsConfig(double min_time, bool has_explicit_iters,
+                          benchmark::IterationCount iters) override {
+        wrapped_->ReportRunsConfig(min_time, has_explicit_iters, iters);
+    }
+
+    void ReportRuns(const std::vector<Run> &report) override {
+        if (!first_report_) {
+#ifdef _WIN32
+            Sleep(cooldown_secs_ * 1000);
```
mtl1979, line 50: Signed/unsigned mismatch, this can cause overflow and/or unexpected results if `cooldown_secs` is not clamped before it is passed to the class.

test/benchmarks/benchmark_main.cc
```diff
@@ -18,15 +26,69 @@ extern "C" {
...
+
 #ifndef BUILD_ALT
 #  ifndef DISABLE_RUNTIME_CPU_DETECTION
     cpu_check_features(&test_cpu_features);
 #  endif
 #endif
 
+    for (int i = 1; i < argc; i++) {
+        if (strncmp(argv[i], "--benchmark_cooldown=", 21) == 0) {
+            cooldown_secs = atoi(argv[i] + 21);
```
mtl1979: No validation whatsoever... If negative or very large number is passed, the results are unexpected, as `Sleep()` on Windows expects unsigned value, not signed value.

mtl1979 (changes requested):
Unsafe command-line parameter handling, signed/unsigned mismatch

test/benchmarks/README.md
```diff
@@ -5,18 +5,34 @@ These benchmarks are written using [Google Benchmark](https://github.com/google/
...
+
+*Cooldown*
+
+To insert a cooldown sleep between benchmark families to mitigate thermal throttling:
+
+```sh
+--benchmark_cooldown=3
+```
```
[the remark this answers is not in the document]
nmoinvaz: Fixed.

test/benchmarks/benchmark_main.cc
nmoinvaz: I have clamped it at `[0,60]` so it doesn't overflow.

nmoinvaz:
Fixed some issues and rebased.

test/benchmarks/benchmark_main.cc
nmoinvaz: The clamp I added should fix this. Is MSVC throwing a compiler warning or no?

test/benchmarks/benchmark_main.cc
```diff
@@ -4,6 +4,14 @@
  */
 
 #include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#ifdef _WIN32
+#  include <windows.h>
```
mtl1979, line 13: Including whole `windows.h` can cause weird compilation failures when subsequent headers use min/max from standard C++ headers... Defining `NOMINMAX` before including the header will solve it.

Usually it's safer to include `windows.h` as late as possible to avoid conflicts with other headers... If other headers need anything from it, usually they define `NOMINMAX`, or for example `WIN32_LEAN_AND_MEAN`, themselves before including it.

nmoinvaz: I'm not using min/max. I am just wondering if this is a real problem in the current code?

mtl1979:
> Fixed some issues and rebased.

Down to 1 error, 16 warnings... Don't really care about warnings that are not treated as errors...

test/benchmarks/benchmark_main.cc
mtl1979: MSVC still fails to compile before this line so no static analysis can be run. I'm not really fan of downgrading Google Benchmarks just to see which version introduced the issues in its headers.

Latest MSVC doesn't like line `static constexpr int64_t TombstoneValue = std::numeric_limits<int64_t>::max();` in `MemoryManager` class inside `benchmark.h`. That line conflicts with defines included through `windows.h` like I mentioned earlier.

test/benchmarks/benchmark_main.cc
mtl1979: > I'm not using min/max. I am just wondering if this is a real problem in the current code?

The issue is that the line 16 includes header that use `max()`... This obviously depends on what Google Benchmark version is found during configuring.. The offending line was added on March 12, 2025, which was just before release of version 1.9.2.

Ref: https://github.com/google/benchmark/commit/5a4c4595480334560fca0a2bdee168371853b52f#diff-cb62aeda23f73502b1ce471b5fd1a17c187f41d0311f6e6dcc3a4d98f3fb79d8R392

nmoinvaz: Fixed

test/benchmarks/benchmark_main.cc
nmoinvaz: Should be fixed now.

mtl1979 (approved):
LGTM, tested with Visual Studio 2026

Dead2 (approved):

PR #2249, opened 2026-04-09
Fix data race on functable reads

sfc-gh-bhannel:
FUNCTABLE_ASSIGN (added in #1609) writes functable members with atomic stores, but callers read them with non-atomic loads. When init_functable() runs concurrently with a reader, this is a data race. Per [C11 5.1.2.4p25](https://port70.net/~nsz/c/c11/n1570.html#5.1.2.4p25), any such data race is undefined behavior. In practice, this causes TSAN violation reports.

Add FUNCTABLE_READ() to perform matching (relaxed) atomic loads, closing the race.

To reproduce (on the parent commit, before the fix):

```
mkdir build && cd build
cmake .. -DWITH_GTEST=ON -DZLIBNG_ENABLE_TESTS=ON \
    -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ \
    -DCMAKE_C_FLAGS="-fsanitize=thread" \
    -DCMAKE_CXX_FLAGS="-fsanitize=thread" \
    -DCMAKE_SHARED_LINKER_FLAGS="-fsanitize=thread" \
    -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=thread"
make -j$(nproc) gtest_zlib
./gtest_zlib --gtest_filter='functable.thread_safety'
```

```
WARNING: ThreadSanitizer: data race
  Atomic write of size 8 by thread T1:
    #0 init_functable functable.c
    #1 adler32_stub functable.c
    #2 zng_adler32
  Previous read of size 8 by main thread:
    #0 zng_crc32
  Location is global 'functable' of size 144
SUMMARY: ThreadSanitizer: data race functable.c in init_functable
```

With this fix, the same test reports zero warnings.

### DIFF
--- a/adler32.c
+++ b/adler32.c
@@ -55,22 +55,22 @@ Z_INTERNAL uint32_t adler32_c(uint32_t adler, const uint8_t *buf, size_t len) {
 
 #ifdef ZLIB_COMPAT
 unsigned long Z_EXPORT PREFIX(adler32_z)(unsigned long adler, const unsigned char *buf, size_t len) {
-    return (unsigned long)functable.adler32((uint32_t)adler, buf, len);
+    return (unsigned long)FUNCTABLE_READ(adler32)((uint32_t)adler, buf, len);
 }
 #else
 uint32_t Z_EXPORT PREFIX(adler32_z)(uint32_t adler, const unsigned char *buf, size_t len) {
-    return functable.adler32(adler, buf, len);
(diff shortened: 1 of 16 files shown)


sfc-gh-bhannel:
Closing in favor of https://github.com/zlib-ng/zlib-ng/pull/2234 - I didn't see that PR before, and it is a better approach.

PR #2254, opened 2026-04-12
Simplify longest_match lookahead handling

nmoinvaz:
Please see the commit messages themselves.

### DIFF
--- a/match_tpl.h
+++ b/match_tpl.h
@@ -53,8 +53,9 @@ Z_INTERNAL uint32_t LONGEST_MATCH(deflate_state *const s, uint32_t cur_match) {
     /* The code is optimized for STD_MAX_MATCH-2 multiple of 16. */
     Assert(STD_MAX_MATCH == 258, "Code too clever");
 
-    scan = window + strstart;
     best_len = s->prev_length ? s->prev_length : STD_MIN_MATCH-1;
+    if (best_len >= lookahead)
+        return lookahead;
 
     /* Calculate read offset which should only extend an extra byte
      * to find the next best match length.
@@ -66,6 +67,7 @@ Z_INTERNAL uint32_t LONGEST_MATCH(deflate_state *const s, uint32_t cur_match) {
             offset -= 4;
     }
 
+    scan = window + strstart;
     scan_start = zng_memread_8(scan);
     scan_end = zng_memread_8(scan+offset);
     mbase_end  = (mbase_start+offset);
--- a/match_tpl.h
+++ b/match_tpl.h
@@ -107,7 +107,7 @@ Z_INTERNAL uint32_t LONGEST_MATCH(deflate_state *const s, uint32_t cur_match) {
         /* Update offset-dependent variables */
         limit = limit_base+match_offset;
         if (cur_match <= limit)
-            goto break_matching;
+            return best_len;
         mbase_start -= match_offset;
         mbase_end -= match_offset;
     }
@@ -190,7 +190,7 @@ Z_INTERNAL uint32_t LONGEST_MATCH(deflate_state *const s, uint32_t cur_match) {
                     if (pos < next_pos) {
                         /* Hash chain is more distant, use it */
                         if (pos <= limit_base + i)
-                            goto break_matching;
+                            return best_len;
                         next_pos = pos;
                         match_offset = i;
                     }
@@ -214,7 +214,7 @@ Z_INTERNAL uint32_t LONGEST_MATCH(deflate_state *const s, uint32_t cur_match) {
                 if (pos < cur_match) {
                     match_offset = len - (STD_MIN_MATCH+1);
                     if (pos <= limit_base + match_offset)
-                        goto break_matching;
+                        return best_len;
                     cur_match = pos;
                 }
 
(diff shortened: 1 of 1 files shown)


match_tpl.h
```diff
@@ -56,8 +56,9 @@ Z_INTERNAL uint32_t LONGEST_MATCH(deflate_state *const s, uint32_t cur_match) {
     /* The code is optimized for STD_MAX_MATCH-2 multiple of 16. */
     Assert(STD_MAX_MATCH == 258, "Code too clever");
 
-    scan = window + strstart;
     best_len = s->prev_length ? s->prev_length : STD_MIN_MATCH-1;
+    if (best_len >= lookahead)
```
nmoinvaz, line 57: This actually fixes a bug with this function. According to the comment for `longest_match`
```c
 OUT assertion: the match length is not greater than s->lookahead
```
Without this if statement, it is not true. `lookahead` can change per call. `s->prev_length` from a previous call may be larger than the `lookahead` for this call.

nmoinvaz:
Rebased

mtl1979 (approved):
LGTM... This improves readability of the code quite a lot...

We can't really ask for more bytes in `lookahead` here or the distance might get too far, so early exit is the best option and eliminates a lot of unnecessary calculation.

Dead2:
### Develop
```
 176980    1336       8  178324   2b894 libz-ng.so.2
```

### PR
```
 177164    1336       8  178508   2b94c libz-ng.so.2
```

Quite a bit bigger, not sure why.
Not seeing any difference in performance or compressed output in my testing.

nmoinvaz:
It does add an additional if statement across all `longest_match` implementations which on x86 there are a few. (64, sse2, avx2, avx512) * 2 (longest_match + longest_match_roll).

Dead2:
> It does add an additional if statement across all `longest_match` implementations which on x86 there are a few. (64, sse2, avx2, avx512) * 2 (longest_match + longest_match_roll).

6 implementations in this case (no C fallback), so that is 30.66 bytes extra per variant. I could understand +16bytes extra, since that is the alignment step for functions. The _roll variants already had the if check, but I suspect it was de-duplicated during linking.
But maybe the new returns require more instances of function teardown code. :man_shrugging:

nmoinvaz:
Dug into this on x86_64 with both toolchains, -O3. (Analysis by Claude.)

| Symbol | GCC 15.2 base→PR (Δ) | Clang 22.1 base→PR (Δ) |
|---|---:|---:|
| `longest_match_sse2` | 1700 → 1761 (**+61**) | 1776 → 1750 (**−26**) |
| `longest_match_roll_sse2` | 2181 → 2039 (**−142**) | 2345 → 2430 (**+85**) |
| `longest_match_avx2` | 1016 → 1058 (**+42**) | 1000 → 1012 (**+12**) |
| `longest_match_roll_avx2` | 1364 → 1446 (**+82**) | 1645 → 1668 (**+23**) |
| `longest_match_avx512` | 840 → 876 (**+36**) | 814 → 821 (**+7**) |
| `longest_match_roll_avx512` | 1318 → 1314 (**−4**) | 1647 → 1634 (**−13**) |
| **all six functions, instructions only** | **+75** | **+88** |
| **all three .o files (code + unwind data)** | **+168** | **+111** |

Real code growth is comparable on both compilers (~75–88 B). The rest of GCC's +168, and the gap up to your reported +184, is empty padding the linker leaves between functions to keep each one 16-byte aligned.

Two things are happening:

**1. Every function pays for the new early return.**

The new `if (best_len >= lookahead) return lookahead;` runs before `offset` is computed, so both compilers split the function setup into two paths — one for `prev_length == 0` (where `best_len` is just 2) and one for the normal case — to avoid having to materialize `best_len` twice. That's where the ~7–61 B per non-roll variant comes from. Clang handles it more compactly than GCC.

**2. `longest_match_roll_sse2` shrinks by 142 B on GCC, and only there.**

For SSE2 specifically, GCC unrolled the inner `match_offset` loop 13 times. Every unrolled copy of `goto break_matching;` got its own little 13-byte exit fragment (`movl $imm, %r12d; subl -0x70(%rbp), %r12d; jmp break_matching`) sitting at the end of the function. Replacing `goto break_matching;` with `return best_len;` deletes those — they all collapse into one shared return path. That's the ~172 B saving.

GCC didn't unroll the AVX2/AVX512 versions (the wider SIMD intrinsics chew through too many registers for the unroller), and Clang didn't unroll any of them, so the saving only materializes in this one spot. That's why GCC's numbers swing wildly (−142 to +82) while Clang's are clustered tightly between −26 and +85.

In short: the +184 B is mostly the cost of the new early-return check plus 16-byte alignment padding. The big per-variant variance is a quirk of GCC's unroller, not something inherent to the change itself.

nmoinvaz:
It said there were a lot of 1 byte REX prefixes in some `longest_match` variants. I had it try at least 5 different variations in the code, but each change only hurt one compiler over the other. If we prefer GCC then we can add `UNLIKELY`, but it hurt clang in some `longest_match` variants.

nmoinvaz:
I ran a bunch more tests across x86/aarch64/clang/gcc with several variations and it seems like trying to optimize for size like this is not worthwhile adventure. You help one to hurt the other.

Dead2:
It is good to have some degree of understanding of why size changes more than expected, so thanks for looking into that.

Dead2 (approved):

PR #2258, opened 2026-04-15
Rename longest_match_slow to longest_match_roll

nmoinvaz:
The "slow" variant of longest_match uses a 3-byte rolling hash to seed its offset-search lookups after a match has been found. Rename the template gate, the functable entry, and all arch-specific instantiations from *_slow to *_roll to reflect what the variant actually uses, so a separate integer-hash offset-search variant can coexist under its own name.

Pure rename, no behavior change.


nmoinvaz:
Rebased

Dead2 (approved):
LGTM

PR #2261, opened 2026-04-16
Move attribute and builtin detection from build system to C preprocessor

nmoinvaz:
Use `__has_attribute` and `__has_builtin` for compile-time detection of `visibility`, `aligned`, and `assume_aligned` support instead of CMake `check_c_source_compiles` and configure compile tests.

This makes detection build-system-independent and consistent with how `Z_FALLTHROUGH` and `Z_TARGET` are already detected.

Visibility uses "hidden" uniformly instead of preferring "internal" when available. Internal is stricter: it lets the compiler skip PIC register loads for intra-module calls, but forbids the function's address from being used outside the module. The optimization only matters on PPC, x86-32, and older RISC, and Clang treats internal as hidden regardless. Hidden is the safer default for functable dispatch and reduces detection to a single Z_HAS_ATTRIBUTE(visibility) check.


Dead2:
Internal also promises the function cannot be escaped (it will return back to caller). And the function cannot be resolved from the application, with hidden it is hidden but can still both conflict with similarly named symbols during static linking, and can be resolved from the application. I seem to remember something about internal also not putting the function symbols in the elf lookup table, so runtime linker then spends less time setting things up while loading the library.

The PIC/TOC optimization is mostly useful for x86-32, PPC and old RISC however.

So I'm not so sure they are as similar as you think.

However, I think we are actually in breach of one of the rules for internal, the functions called via functable need to be hidden and not internal, since calling by pointer is prohibited for internal.

nmoinvaz:
Good points. I may have overstated the similarity. I have updated the commit message and PR description with a better explanation.

Dead2:
But why remove internal for the non-functable functions though? I don't see any upside to doing that other than a minor simplification of the detection logic.

Dead2:
It'd also be a lot easier to review and verify PRs that change detection logic if they only did that and not also make changes to the outcome.

phprus:
> However, I think we are actually in breach of one of the rules for internal, the functions called via functable need to be hidden and not internal, since calling by pointer is prohibited for internal.

Any call by function pointer is prohibited?

The GCC documentation says:

> Internal visibility is like hidden visibility, but with additional processor specific semantics. Unless otherwise specified by the psABI, GCC defines internal visibility to mean that a function is never called from another module. Compare this with hidden functions which, while they cannot be referenced directly by other modules, can be referenced indirectly via function pointers. By indicating that a function cannot be called from outside the module, GCC may for instance omit the load of a PIC register since it is known that the calling function loaded the correct value.

If I understand correctly, only calls from other modules (DSO) are prohibited. And all usage of a functable occurs within a single module (DSO).

Dead2:
> > However, I think we are actually in breach of one of the rules for internal, the functions called via functable need to be hidden and not internal, since calling by pointer is prohibited for internal.
> 
> Any call by function pointer is prohibited?
> 
> The GCC documentation says:
> 
> > Internal visibility is like hidden visibility, but with additional processor specific semantics. Unless otherwise specified by the psABI, GCC defines internal visibility to mean that a function is never called from another module. Compare this with hidden functions which, while they cannot be referenced directly by other modules, can be referenced indirectly via function pointers. By indicating that a function cannot be called from outside the module, GCC may for instance omit the load of a PIC register since it is known that the calling function loaded the correct value.
> 
> If I understand correctly, only calls from other modules (DSO) are prohibited. And all usage of a functable occurs within a single module (DSO).

You are correct, that limitation seems to be unique to BPF and we don't need to cater to that.
I was wondering why nothing was actually broken :slightly_smiling_face: 
(The problem of spending 5-10min researching something while reviewing vs spending hours actually understanding it while I introduced that in the first place).

So we can indeed use internal attribute for functable :+1:.

nmoinvaz:
I had Claude run an in-depth empirical analysis to see if `internal` made any difference. 

Please review the summary.
https://gist.github.com/nmoinvaz/007cd6f481e2b4be96bb0357aaf71ec4#file-zlib-ng-pr2261-visibility-summary-md

The gist includes all supporting files that were part of the test. 

I tested across all GCC supported platforms using Docker. And I had it also investigate the GCC source code and the source code for popular linkers.

Dead2:
You really need to decide to either change detection logic OR change behavior. This mixing of purposes is not beneficial to anyone.

mtl1979:
This is one of the cases where "cleaning up" too much just complicates the code more instead of less... The whole point of doing detection early is to avoid chicken and egg dilemmas, where the "later" detection login depends on something else, which needs to be detected and be available first...

For non-optimized builds, we don't really set minimum compiler version as there is no reason whatsoever to exclude compiler versions that don't even support the target architecture. This means that someone could still try compiling with for example GCC 4.x on ABIs that are incompatible with later GCC versions. Only hard requirement for zlib-ng is that the architecture uses pointers that are at least 32 bits, which eliminates (almost) no architecture currently in use.

nmoinvaz:
* Every additional check adds built-time especially with MSVC where CMake has to generate a solution and project every check.
* My changes have the added benefit that it reduces both build scripts. 
* `defined(__has_builtin)` is super common throughout GitHub code bases. 
* `Z_HAS_ATTRIBUTE` is already defined in our current code base and uses the same methodology `defined(__has_attribute)`
* I've already shown `internal` is identical to `hidden` empirically. I don't care which one we use.

I know there is lots of push back to not change these things.

mtl1979:
> I know there is lots of push back to not change these things.

Every controversial change needs 2 approvals and if Dead2 thinks something needs revising, then there is very small, almost zero, chance that two other long-time contributors take the opposite view. Getting just 1 approval has over 50% chance.

Splitting out the controversial part is one option, but it isn't always the most meaningful solution.

I don't usually care what others think, but when it comes to my opinion, sometimes it's minority opinion.

Dead2:
In general, moving detection out of the buildsystem and into the code itself makes the detection logic opaque, neither we or the users then have any way of seeing, logging or noticing what was detected correctly or incorrectly without actively going in and making temporary code changes to expose the internals of the detection.
This might cut down on bug reports, but that is because nobody notices missing or incorrect detection anymore. I'm not so sure that is a good thing.

PR #2262, opened 2026-04-16
Remove ptrdiff_t configure/cmake detection

nmoinvaz:
## Summary

Remove the ptrdiff_t configure/cmake detection. `ptrdiff_t` is defined in `<stddef.h>` since C89 and is guaranteed on any C99+ compiler, making the build-time check and the fallback typedef in the zconf headers unnecessary.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -636,32 +636,6 @@ if(HAVE_BUILTIN_ASSUME_ALIGNED)
 endif()
 set(CMAKE_REQUIRED_FLAGS)
 
-#
-# check for ptrdiff_t support
-#
-check_c_source_compiles(
-    "#include <stddef.h>
-     int main() {
-         ptrdiff_t *a;
-         (void)a;
-         return 0;
-    }"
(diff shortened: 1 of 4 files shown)


mtl1979:
Only reason we haven't removed the CPUID detection from build system has been that it shows which one is supported by toolchains which we don't support yet, but might support if everything required works... A lot of the recent ones (released in last 2-3 years) are based on LLVM, but older ones aren't.

I have no opinions on removing tests for `ptrdiff_t` in build system, as personally I already use quite modern compilers. Some things are just relics from zlib sources that we haven't removed yet.

nmoinvaz:
Rebased with Claude Code.

nmoinvaz:
@mtl1979 I have removed the CPU ID stuff and just kept ptrdiff_t removal in this PR.

mtl1979 (approved):
LGTM... This is non-issue in compat mode as we haven't supported compiling zlib-ng with "prehistoric" toolchains since removing support for K&R and 16-bit targets.

Issue #2264, opened 2026-04-17
Broken compilation with -DWITH_CRC32_CHORBA=OFF


matbech:
zlib-ng fails to compile with -DWITH_CRC32_CHORBA=OFF
```
functable.c(124,21): error C2065: 'crc32_chorba_sse2': undeclared identifier
functable.c(151,21): error C2065: 'crc32_chorba_sse41': undeclared identifier
```

mtl1979:
We're reworking that part of the code and had to split some changes to multiple commits... There is active discussion on our Mattermost server regarding the compilation failure...

cc: @nmoinvaz

nmoinvaz:
Confirmed fixed by #2260.

PR #2265, opened 2026-04-17
Fix WITH_CRC32_CHORBA=OFF

matbech:
Fixes issue https://github.com/zlib-ng/zlib-ng/issues/2264

### DIFF
--- a/functable.c
+++ b/functable.c
@@ -120,7 +120,7 @@ static int init_functable(void) {
         ft.longest_match_roll = &longest_match_roll_sse2;
         ft.slide_hash = &slide_hash_sse2;
 #  endif
-#  if !defined(WITHOUT_CHORBA_SSE) && !defined(X86_PCLMULQDQ_NATIVE)
+#  if !defined(WITHOUT_CHORBA) && !defined(WITHOUT_CHORBA_SSE) && !defined(X86_PCLMULQDQ_NATIVE)
         ft.crc32 = &crc32_chorba_sse2;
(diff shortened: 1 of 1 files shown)


Dead2:
This was a known regression in #2139 (and then the reworked #2208)
Hopefully will be fixed by part2 in #2260, but it still needs some changes.

nmoinvaz:
This is already fixed by #2260.

nmoinvaz (dismissed):

functable.c
```diff
@@ -147,7 +147,7 @@ static int init_functable(void) {
     if (cf.x86.has_sse41)
 #  endif
     {
-#  ifndef WITHOUT_CHORBA_SSE
+#  if !defined(WITHOUT_CHORBA) && !defined(WITHOUT_CHORBA_SSE)
```
nmoinvaz, line 150: WITHOUT_CHORBA only disables scalar version. I think we are missing a check in the headers.

mtl1979: https://github.com/zlib-ng/zlib-ng/blob/develop/arch/x86/x86_functions.h#L27

https://github.com/zlib-ng/zlib-ng/blob/develop/arch/x86/x86_functions.h#L52

mtl1979:
The real question that we started discussing on the Mattermost server is that do we want a quick fix before #2260 is ready for merging, or should we wait... It's weekend currently and I wasn't around most of the day to check all the relevant lines in the code until less than 1 hour ago...

nmoinvaz:
I think this can be closed

PR #2267, opened 2026-04-21
Fail early in count_lengths if lengths are invalid.

### LINKED ISSUE #2266: Potential memory safety issue — requesting private contact
I have identified a potential memory safety issue in zlib-ng.
Confirmed on version 2.3.90 (d1afe3a6) with AddressSanitizer.

Please contact me privately to share the full technical report and proof of concept.

mtl1979:
Doing the validity check any earlier or later will either introduce more duplicated code or address sanitizer errors further down the path. We bail out early from `zng_inflate_table` if `count_lengths` detects invalid lengths. The caller still needs to detect error condition after `zng_inflate_table` returns, this fix will only eliminate potential data corruption with direct call to `zng_inflate_table` using intentionally invalid parameters.

Fixes #2266.

### DIFF
--- a/inftrees.c
+++ b/inftrees.c
@@ -25,8 +25,10 @@ const char PREFIX(inflate_copyright)[] = " inflate 1.3.1 Copyright 1995-2024 Mar
   copyright string in the executable of your product.
  */
 
-/* Count number of codes for each code length. */
-static inline void count_lengths(uint16_t *lens, int codes, uint16_t *count) {
+#define CHECK_LEN(len) do { if ((len) > MAX_BITS) return -1; } while (0)
+
+/* Count number of codes for each code length. Return -1 on error */
+static inline int count_lengths(uint16_t *lens, int codes, uint16_t *count) {
     /* IBM...made some weird choices for VSX/VMX. Basically vec_ld has an inherent
      * endianness but we don't want to force VSX to be needed */
     static const ALIGNED_(16) uint8_t one[256] = {
@@ -53,13 +55,16 @@ static inline void count_lengths(uint16_t *lens, int codes, uint16_t *count) {
     vector unsigned char s2 = vec_splat_u8(0);
 
     if (codes & 1) {
+        CHECK_LEN(lens[0]);
         s1 = vec_ld(16 * lens[0], one);
         --codes;
         ++lens;
     }
 
     while (codes) {
+        CHECK_LEN(lens[0]);
         s1 = vec_add(s1, vec_ld(16 * lens[0], one));
+        CHECK_LEN(lens[1]);
         s2 = vec_add(s2, vec_ld(16 * lens[1], one));
         codes -= 2;
         lens += 2;
@@ -77,10 +82,13 @@ static inline void count_lengths(uint16_t *lens, int codes, uint16_t *count) {
     uint8x16_t s2 = vdupq_n_u8(0);
 
     if (codes & 1) {
+        CHECK_LEN(lens[0]);
         s1 = vld1q_u8(&one[16 * lens[0]]);
     }
     for (sym = codes & 1; sym < codes; sym += 2) {
+        CHECK_LEN(lens[sym]);
         s1 = vaddq_u8(s1, vld1q_u8(&one[16 * lens[sym]]));
+        CHECK_LEN(lens[sym+1]);
         s2 = vaddq_u8(s2, vld1q_u8(&one[16 * lens[sym+1]]));
     }
 
@@ -93,10 +101,13 @@ static inline void count_lengths(uint16_t *lens, int codes, uint16_t *count) {
     __m128i s2 = _mm_setzero_si128();
 
     if (codes & 1) {
+        CHECK_LEN(lens[0]);
         s1 = _mm_load_si128((const __m128i*)&one[16 * lens[0]]);
     }
     for (sym = codes & 1; sym < codes; sym += 2) {
+        CHECK_LEN(lens[sym]);
         s1 = _mm_add_epi8(s1, _mm_load_si128((const __m128i*)&one[16 * lens[sym]]));  // vaddq_u8
+        CHECK_LEN(lens[sym+1]);
         s2 = _mm_add_epi8(s2, _mm_load_si128((const __m128i*)&one[16 * lens[sym+1]]));
     }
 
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Just put this above the call to `count_lengths`. It is very simple.
```c
      for (unsigned i = 0; i < codes; i++) {
          if (lens[i] > MAX_BITS)                                                                                          
              return -1;                                       
      }
```

mtl1979:
> Just put this above the call to `count_lengths`. It is very simple.
> 
> ```c
> 
>       for (unsigned i = 0; i < codes; i++) {
> 
>           if (lens[i] > MAX_BITS)                                                                                          
> 
>               return -1;                                       
> 
>       }
> 
> ```

That was an option but I didn't want to duplicate the loop and make the code unnecessarily slower.

nmoinvaz:
I feel like it is going to hurt the SIMD performance by putting all those if statements inline there. Perhaps it needs to be benchmarked.

phprus:
Is it possible to construct a input data for the public API such that `zng_inflate_table` will be called with the wrong `lens` parameter?

If this is not possible, then maybe add `Assert` instead of if-condition on the hot path?
Like this:

```c
#ifdef ZLIB_DEBUG
      for (unsigned i = 0; i < codes; i++) {
         Assert(lens[i] <= MAX_BITS, "Invalid lens");
      }
#endfif
```

mtl1979:
> I feel like it is going to hurt the SIMD performance by putting all those if statements inline there. Perhaps it needs to be benchmarked.

It definitely do hurt performance, but it can also reuse some loads and possibly execute at least latter of the compares in parallel with first vector instruction.

We always benchmark things when it involves multiple architectures as not all architectures can interleave same two instructions.

We have to remember that getting invalid lenghts is very unlikely and `zng_inflate_table` gets called multiple times.

mtl1979:
> Is it possible to construct a input data for the public API such that `zng_inflate_table` will be called with the wrong `lens` parameter?
> 

Yes, it is possible... I have a test executable I used to verify my fix. I didn't include it here as most new regression tests should use GoogleTest and it is harder to use without cmake.

KungFuJesus:
Hah, you could use simd to test all entries in the table as a first pass. It's a second scan through the lens array but I imagine doing 4, 8, or 16 at a time is at least better.

KungFuJesus:
> > Just put this above the call to `count_lengths`. It is very simple.
> > ```c
> >       for (unsigned i = 0; i < codes; i++) {
> > 
> >           if (lens[i] > MAX_BITS)                                                                                          
> > 
> >               return -1;                                       
> > 
> >       }
> > ```
> 
> That was an option but I didn't want to duplicate the loop and make the code unnecessarily slower.

I think this might be faster because it will auto vectorize and it's fewer branches that way. You could in fact OR the result of the comparisons to a single variable and check it once. It's a second scan through the array but I still think it might be better than a branch per entry.

mtl1979:
> I think this might be faster because it will auto vectorize and it's fewer branches that way. You could in fact OR the result of the comparisons to a single variable and check it once. It's a second scan through the array but I still think it might be better than a branch per entry.

The number of entries is not power of 2 and small enough that there isn't much benefit from autovectorization. It would require masking and at least AVX512 to efficiently unroll the whole loop and vectorize it.

Tracing the potential code paths, there is only 1/3 chance that there can be invalid data passed to `zng_inflate_table` from public API and so far nobody has submitted proof-of-concept compressed file that actually trigger the out-of-bounds access, so this PR is just for documentation purposes... The issue this PR "fixes" is likely generated by AI bot, so credibility of any claim is minuscule. This was discussed on the Mattermost server and no core contributor could generate or provide such compressed file either.

KungFuJesus:
Yeah but the compiler can generate fairly efficient code to strip mine the remainders. I'd benchmark a loop that prescans with an invalid flag you or together and exit outside the loop if nonzero.

mtl1979:
> Yeah but the compiler can generate fairly efficient code to strip mine the remainders. I'd benchmark a loop that prescans with an invalid flag you or together and exit outside the loop if nonzero.

The biggest issue is that autovectorization is only allowed in native builds in this specific code path, explicit vector instructions are only used if the base architecture always has those instructions, like SSE2 on x86_64 and GCC/Clang.

This would mean using separate loop for autovectorization would require fallback path for non-native builds, which are majority, thus adding unnecessary complexity to the code.

Using same "loop" for validating the entries allows better debugging of the code as it will allow adding breakpoint at the exact point where the invalid value is accessed. This makes possible also to just clear the invalid entry and resume counting the remaining lengths. Not clearing and resuming from next entry will cause same out-of-bound access later when execution returns back from `count_lengths` to `zng_inflate_table`.

KungFuJesus:
Right but even a lowest common denominator compilation with sse should be sufficient for that. And this file already now has functions that assume the code could be compiled with -march=x86-64-v* (I think this is something rocky or alma Linux do now)

mtl1979:
> Right but even a lowest common denominator compilation with sse should be sufficient for that. And this file already now has functions that assume the code could be compiled with -march=x86-64-v* (I think this is something rocky or alma Linux do now)

Assuming there is often 19 entries, ARM could do 1 entry head with 6 iterations of 3 entries. Other architectures could do 3 entry scalar head or tail with remaining 16 entries vectorized as either 8 iterations of 2 entries or 4 iterations of 4 entries depending on register count.

This would be doable currently on GCC or Clang, but not yet on Visual C++, although with some additional preprocessor workarounds it is technically possible.

nmoinvaz:
> The number of entries is not power of 2

Do you mean a multiple of the lane width?

nmoinvaz:
https://gist.github.com/nmoinvaz/47889c1f1efcda7594281744aca957c5
Seems like a lot of code just for internal use function..

mtl1979:
> > The number of entries is not power of 2
> 
> 
> 
> Do you mean a multiple of the lane width?

No... Same as lane count...

KungFuJesus:
> https://gist.github.com/nmoinvaz/47889c1f1efcda7594281744aca957c5 Seems like a lot of code just for internal use function..

Well sure but the autovectorizer should be able to produce that or near that and the code itself remains a simple scalar loop.

mtl1979:
Really dumb idea but you can just load the bytes to uint64_t and drop lowest 4 bits of each length. This requires no vector instructions at all and as such no expensive or slow instructions. This works because MAX_BITS is 15, which is exactly one less than closest power of 2.

nmoinvaz:
> Really dumb idea 

I benchmarked it anyways just for fun:
* check_lens benchmark: https://gist.github.com/nmoinvaz/41f7368d579fee5f33d4c5b21a9e7ff9
* count_lens benchmark: https://gist.github.com/nmoinvaz/1623fc69ca866ea6886c0e8597ed73fa

mtl1979:
> > Really dumb idea 
> 
> 
> 
> I benchmarked it anyways just for fun:
> 

Autovectorizing might optimize it even better if there is 2 uint64_t reads per iteration. However using uint128_t directly might make the mask constant too big for Visual C++.

Dead2:
I think this can be closed, nobody was able to reproduce the "bug" without access to internal functions.

PR #2275, opened 2026-05-04
Minor cleanups in deflate_rle.c and deflate_fast.c

Dead2:
- An incorrect comment in deflate_fast.c, seemingly copied from deflate_slow.c where it does make sense.
- In deflate_rle: No need to check that match_len is less than 258, as compare_256_rle+2 can only be <= 258.


nmoinvaz (approved):
LGTM

PR #2277, opened 2026-05-04
Allow /delta on fork pull requests

nmoinvaz:
## Summary

The `/delta` workflow rejects fork PRs even though the `author_association` gate already restricts triggers to `OWNER`, `MEMBER`, or `COLLABORATOR`. A maintainer typing `/delta` on a fork PR is the same trust model as checking the PR out locally to benchmark it, so the extra fork check is friction without a clear safety benefit.

This drops the rejection and the now-unused base/head repo id parsing. `actions/checkout` resolves the fork's head SHA via `refs/pull/N/head` in the base repo, so no other plumbing is needed.

### DIFF
--- a/.github/workflows/delta.yml
+++ b/.github/workflows/delta.yml
@@ -81,13 +81,6 @@ jobs:
           BASE_SHA=$(echo "$pr_json" | jq -r .base.sha)
           HEAD_BRANCH=$(echo "$pr_json" | jq -r .head.ref)
           HEAD_SHA=$(echo "$pr_json" | jq -r .head.sha)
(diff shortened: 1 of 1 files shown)


nmoinvaz:
It looks like that triggered the workflow from the default branch instead of my PR.

Dead2:
> It looks like that triggered the workflow from the default branch instead of my PR.

To be fixed in another PR or requires additional parameters to run?

nmoinvaz:
No it just needs to be merged in order to work. I don't know a way to test it outside of merging it due to GH limitations.

PR #2281, opened 2026-05-06
[CI] Cache Ubuntu .deb packages to speed up installing dependencies.

mtl1979:
New cache is created for each week of year, so new packages are added to cache weekly. Caching packages also fixes issues if there is temporary interruption of service in the package repository servers.

### DIFF
--- a/.github/actions/restore-apt-cache/action.yml
+++ b/.github/actions/restore-apt-cache/action.yml
@@ -1,0 +1,44 @@
+name: Restore APT cache
+description: Restores cached .deb packages for APT
+inputs:
+  packages:
+    description: Packages to cache
+    required: true
+  cache-key:
+    description: Cache key to use
+    required: true
+outputs:
+  cache-hit:
+    description: A boolean value to indicate an exact match was found for the primary key
+    value: ${{ steps.load-cache.outputs.cache-hit }}
+runs:
+  using: "composite"
+  steps:
+    - name: Get cache key (Ubuntu)
+      if: runner.os == 'Linux'
+      id: get-ubuntu-cache-key
+      shell: bash
+      run: |
+        echo "release=$(sed -n 's/UBUNTU_CODENAME=\(.*\)/\1/p' /etc/os-release)" >> $GITHUB_OUTPUT
+        echo "pkg_sig=$(printf '%s' "${{ inputs.packages }}" | sha256sum | cut -c1-12)" >> $GITHUB_OUTPUT
+        echo "key=$(/bin/date -u "+%G%V")" >> $GITHUB_OUTPUT
+
+    - name: Load package cache (Ubuntu)
+      if: runner.os == 'Linux'
+      id: load-cache
+      uses: actions/cache@v5
+      with:
+        path: ${{ runner.temp }}/apt-archives/*.deb
+        key: cache-apt-${{ inputs.cache-key || 'default' }}-${{ runner.arch }}-${{ steps.get-ubuntu-cache-key.outputs.release }}-${{ steps.get-ubuntu-cache-key.outputs.pkg_sig }}-${{ steps.get-ubuntu-cache-key.outputs.key }}
+        restore-keys: |
+          cache-apt-${{ inputs.cache-key || 'default' }}-${{ runner.arch }}-${{ steps.get-ubuntu-cache-key.outputs.release }}-${{ steps.get-ubuntu-cache-key.outputs.pkg_sig }}
+          cache-apt-${{ inputs.cache-key || 'default' }}-${{ runner.arch }}-${{ steps.get-ubuntu-cache-key.outputs.release }}
+
+    - name: Restore cached packages (Ubuntu)
+      if: runner.os == 'Linux'
+      shell: bash
+      run: |
+        sudo mkdir -p /var/cache/apt/archives ${{ runner.temp }}/apt-archives
+        if compgen -G "${{ runner.temp }}/apt-archives/*.deb" > /dev/null; then
+          sudo mv -f ${{ runner.temp }}/apt-archives/*.deb /var/cache/apt/archives/
+        fi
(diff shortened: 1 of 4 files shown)


nmoinvaz:
Will this break when GH updates `ubuntu-latest` image?

mtl1979:
> Will this break when GH updates `ubuntu-latest` image?

It might... But this is still quite early draft, so it can be detected and cache key can be changed to include Ubuntu version number.

mtl1979:
Example cache key is now `cache-apt-Ubuntu GCC SSE2 UBSAN-noble-202619`... This means fresh cache is created for each LTS version of Ubuntu.

.github/workflows/cmake.yml
```diff
@@ -722,12 +722,35 @@ jobs:
...
+        echo "release=$(sed -n 's/UBUNTU_CODENAME=\(.*\)/\1/p' /etc/os-release)" >> $GITHUB_OUTPUT
+        echo "key=$(/bin/date -u "+%G%V")" >> $GITHUB_OUTPUT
+
+    - name: Load package cache (Ubuntu)
+      if: runner.os == 'Linux' && !contains(matrix.os, 'z15')
+      uses: actions/cache@v5
+      with:
+        path: ${{ runner.temp }}/apt-archives/*.deb
+        key: cache-apt-${{ matrix.name }}-${{ steps.get-ubuntu-cache-key.outputs.release }}-${{ steps.get-ubuntu-cache-key.outputs.key }}
+        restore-keys: cache-apt-${{ matrix.name }}-${{ steps.get-ubuntu-cache-key.outputs.release }}
```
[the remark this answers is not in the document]
mtl1979: We could theoretically group caches by target architecture, but that would require introducing new matrix variable and installing both gcc and clang even if only gcc is needed for the job.

We don't cache empty directory and there should be no other files than deb packages as we only move them to the temporary directory and not the index files that change more often.

mtl1979:
Example cache key is now: `cache-apt-default-ARM64-noble-b638d81430d6-202619` ... This should make grouping similar jobs quite easy, although the key is getting quite long...

mtl1979:
Rebased and added purging older versions of packages and unneeded dependencies.

nmoinvaz:
I ran analysis with Claude:

Apt-install step (the thing the PR actually changes)                                                                                                            
                                                                             
  Comparing develop's Install packages (Ubuntu) step against the PR's full cache pipeline (Get cache key + Load package cache + Install packages + Post Load      
  package cache):                                                                                                                                                 
                                                                                                                                                                  
  - Total apt-step time across 140 Ubuntu jobs: 6,377s → 3,427s (-46.3%, ~49 min saved per workflow).                                                             
  - Mean per job: 45.5s → 24.5s (-46%).                                                                                                                           
  - Biggest wins are jobs where develop's apt-get hit slow mirrors and dragged on for hundreds of seconds (Pigz/Clang No Optim 682→117s, GCC RISC-V 707→19s, GCC  
  MIPS 340→22s, Configure/PPC 420→19s). The cache makes those reliably fast.                                                                                      
  
  Where it loses                                                                                                                                                  
                                                                                       
  - Cache machinery adds ~5–15s overhead on small/fast jobs. About 30 Ubuntu jobs got slower on the apt step in absolute terms, e.g. GCC ASAN 15→30s, Inflate     
  Strict REDUCED_MEM 10→18s, GCC SSE4.2 UBSAN 22→45s. These are all small absolute regressions.
  - One outlier looks like noise, not a cache effect: Pigz / Ubuntu Clang No Threads apt step went 21s→674s in this PR run — a runner/mirror hiccup unrelated to  
  the change.

nmoinvaz:
Claude suggests to gate whether or not you cache based on whether or not `matrix.packages` is set: see proposal in https://gist.github.com/nmoinvaz/978f248ea7d528c954e3d52fb8dc99c0. If you do that I can rerun the analysis.

mtl1979:
> Claude suggests to gate whether or not you cache based on whether or not `matrix.packages` is set: see proposal in https://gist.github.com/nmoinvaz/978f248ea7d528c954e3d52fb8dc99c0. If you do that I can rerun the analysis.

I was considering that for `cmake.yml` as `configure.yml` already does that... Simplifies somewhat the code for cache key generation...

nmoinvaz:
Since the steps are duplicated between cmake and configure workflows, should move the logic to a [composite action](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action) under `.github\actions` that can be reused.

mtl1979:
> Since the steps are duplicated between cmake and configure workflows, should move the logic to a [composite action](https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action) under `.github\actions` that can be reused.

We don't use composite actions for anything else even though things are duplicated in at least two workflow files. However that could be a separate cleanup PR.

nmoinvaz:
If there is equally complex and duplicate code, we should use composite actions where it makes sense. I recommend it as a separate commit. There is no reason to wait for another PR.

mtl1979:
> If there is equally complex and duplicate code, we should use composite actions where it makes sense. I recommend it as a separate commit. There is no reason to wait for another PR.

This is going to get even more complex once I do the final cleanup of the code... Currently there is still some code in the "Install" step that should be moved to own step... I just wanted some benchmarks to see if the PR is worth polishing...

nmoinvaz:
New analysis looks good: https://gist.github.com/nmoinvaz/c3f096b565323db64e197c09fa961018

.github/actions/restore-apt-cache/action.yml
```diff
@@ -0,0 +1,44 @@
+name: 'Restore APT cache'
```
nmoinvaz: nitpick: might want to consider just removing quotes and double quotes from both composite action keys, it is not necessary for yaml, and it is mix-matched.

mtl1979: Removed... Those were mixed even in the example template in GitHub documentation ;)

nmoinvaz:
Looking good.

nmoinvaz (approved):
Cache key name construction looks good, workflow and composite actions are nice and clean.

PR #2282, opened 2026-05-06
[CI] Use older runner for Visual Studio 2022 jobs.

mtl1979:
Latest Windows runner only has Visual Studio 2026 now. 

After we drop support for 32-bit Windows (x86/ARMv7), we can migrate some of the jobs with recent toolchain to Visual Studio 2026, but that's not relevant until support for Visual Studio 2022 is dropped, or too many of the runners are switched to newer Windows and Visual Studio version causing excess delays waiting for available runner.


Dead2 (approved):
LGTM

PR #2285, opened 2026-05-09
Hoist rolling hash state to a local in INSERT_STRING

nmoinvaz:
## Summary

The rolling-hash variant of `INSERT_STRING` (level 9) configured `HASH_CALC_VAR` as `s->ins_h`, so the inner loop wrote through `s->ins_h` on every byte. Indirect stores via `prev[]`/`head[]` (both `Pos*`) prevented the compiler from caching the running hash in a register — it couldn't prove those writes don't alias the integer field, so it kept committing `s->ins_h` to memory each iteration.

This change lifts the running hash to a stack local. The local has no address taken, so the compiler trusts there's no aliasing and keeps it register-resident across the loop. `s->ins_h` is loaded once on entry and stored once on exit.

The non-rolling variant (levels 1–8) already used a local `h`, so it's unaffected; the assembly for `_insert_string` is byte-identical.

## Benchmark

`insert_string_bench/rolling_hash` on Apple M5, 10 reps, geomean **−3.55%**:

| count | Δ time |
| ----- | ------ |
| 100/4   | −1.60% |
| 100/5   | −0.84% |
| 100/7   | −4.38% |
| 100/14  | −4.87% |
| 100/32  | −5.99% |
| 100/127 | −5.96% |
| 100/255 | −6.81% |

### DIFF
--- a/insert_string_p.h
+++ b/insert_string_p.h
@@ -15,6 +15,7 @@
 #define HASH_CALC_MASK       HASH_MASK
 #define HASH_CALC_VAR        h
 #define HASH_CALC_VAR_INIT   uint32_t h
+#define HASH_CALC_VAR_STORE
 #define HASH_CALC_OFFSET     0
 
 #define UPDATE_HASH          update_hash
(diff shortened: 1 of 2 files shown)


nmoinvaz:
Since we no longer use intrinsics for integer hashing and only have rolling hash and integer hash, we should eventually just remove the `_tpl.h` stuff for insert string since it makes it harder to read.

mtl1979 (approved):
LGTM

PR #2288, opened 2026-05-11
Cross-vary inflate and deflate benchmarks by input data type

nmoinvaz:
## Summary

inflate_bench and deflate_bench now register a separate benchmark per (variant, data_type) pair so the data type appears in the display name. inflate yields five benchmarks per size:

```
inflate_bench/nocrc/{text,short_match,random,realistic_rgb,striped_rgb}/SIZE
```

deflate yields five benchmarks per (variant, size, level) — across level/nocrc/filtered/huffman/rle/fixed — for 280 combinations total:

```
deflate_bench/{level,nocrc,filtered,huffman,rle,fixed}/{text,...}/SIZE/LEVEL
```

Both fixtures share a Dispatch() pattern with an empty SetUp; ->Name() controls the "/"-separated display. Inflate sizes its compressed buffer via deflateBound to handle incompressible inputs.

All data generators move to test/test_data_p.h (renamed from compressible_data_p.h) behind a gen_test_data(test_data_type, bufsize) dispatcher. The five data shapes target distinct deflate stream regions:

- **text** — English-like literals + short/medium matches (the original benchmark shape)
- **short_match** — many short back-references with small distances
- **random** — pseudorandom bytes, deflate falls back to stored blocks
- **realistic_rgb** — RGB photo: short matches at dist=3 plus long inter-row matches
- **striped_rgb** — solid R/G/B stripes: long matches at dist=3


test/test_data_p.h
```diff
@@ -0,0 +1,191 @@
...
+
+        for (size_t r = 0; r < reps && i < bufsize; r++) {
+            for (size_t j = 0; j < plen && i < bufsize; j++) {
+                buf[i++] = pattern[j];
+            }
+        }
+    }
+
+    return buf;
+}
```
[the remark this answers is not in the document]
nmoinvaz: fixed

test/test_data_p.h
```diff
@@ -0,0 +1,190 @@
...
+        return NULL;
+    }
+
+    for (size_t pos = 0; pos < bufsize; ) {
+        pos += append_str(buffer+pos, bufsize-pos, teststr1);
+        pos += append_uint8_t(buffer+pos, bufsize-pos, (uint8_t)(rand() & 0xFF));
+        /* Every so often, add a few other little bits to break the pattern */
+        if (loops % 13 == 0) {
+            pos += append_str(buffer+pos, bufsize-pos, teststr3);
+            pos += append_uint8_t(buffer+pos, bufsize-pos, (uint8_t)(rand() & 0xFF));
```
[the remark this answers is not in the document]
nmoinvaz: I don't wish to change the structure of those atm. Besides I already did a useless trivial thing you suggested which I don't agree with so much.

test/benchmarks/benchmark_inflate.cc
```diff
@@ -15,40 +15,41 @@ extern "C" {
...
 #define MAX_SIZE (1024 * 1024)
 #define NUM_TESTS 6
 
 class inflate_bench: public benchmark::Fixture {
 private:
-    uint8_t *inbuff;
-    uint8_t *outbuff;
+    uint8_t *inbuff = nullptr;
+    uint8_t *outbuff = nullptr;
     uint8_t *compressed_buff[NUM_TESTS];
```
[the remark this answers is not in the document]
nmoinvaz: The matching `benchmark_uncompress.cc` fixture (lines 28-29, predating this PR) has the same uninitialized `compressed_buff[NUM_TESTS]` and the same unconditional `free()` in TearDown. The failure path requires `malloc`/`deflateInit2` to fail partway through setup on a benchmark machine — not a realistic scenario. Keeping the convention consistent with the sibling fixture.

Dead2:
Aren't we expanding the number of benchmarks a bit too much now. It's already so many and so time-consuming that I never run the whole suit anymore, I only choose specific benchmarks. I worry it becomes so much that it effectively loses its value.

nmoinvaz:
The problem is that we are continually only optimizing for a single type of data. It does create more benchmarks, but I don't see another solution. I have never run the entire benchmark suite because usually I am only optimizing a particular part. Each of the data that I have included hits inflate/deflate a different way.

Dead2:
> The problem is that we are continually only optimizing for a single type of data. It does create more benchmarks, but I don't see another solution. I have never run the entire benchmark suite because usually I am only optimizing a particular part. Each of the data that I have included hits inflate/deflate a different way.

Sure, but those benchmarks are really geared for measuring something else, since they have up to 16 size-alternatives. We can't have that per type of data and have it be actually used and useful. If we cut it down to something like 2-3 sizes per data type then perhaps it'll turn out to be useful.

I already spend hours benchmarking most PRs that might affect performance, this'll add to that, or actually it probably will reduce that since I am more likely to stop benchmarking using gbench or just use a specific one if it takes too long and has more values than I can bother to keep track of.

I have mostly trusted the silesia tar files to be a fair balance of real-world data. Sure it's not perfect by any measure, but IMHO it's good enough to catch most regressions that are worth caring about.

nmoinvaz:
The Silesia benchmark does not allow you to hone in on certain data types. Everything is lumped together. You have to have a sensitive benchmark rig in order to see any difference. Where when you have benchmarks dedicated for a particular type of structured data it is quite obvious whether or not there is an improvement.

I will just use these benchmarks privately.

Dead2:
If you instead add a new type of deflate/inflate (raw deflate stream for example) benchmarks that only does a single test of for example 1MB buffer for each type of data, then I imagine that'd probably be very nice.

Edit: What I objected to is the multiplication effect on the number of tests, instead of just adding a few.

nmoinvaz:
One problem currently is the deflate*/inflate* benchmarks are misleading without including the type of data that is being used somewhere in the name. It's not really a performance of deflate or inflate overall, rather it is the performance of it under a particular type of manufactured data.

Even if you split it out into what you are talking about you still want inflate and deflate versions of the same. So it only gets us the same place.

There are ways to filter the benchmarks to only run "deflate_bench/text/*" which would include all the tests you normally run.

You can also only run 1MB tests if you wanted. "deflate_bench/*/*/1048576" (or so).

But running only "deflate*" will be too much. So yes this would involve changing how we use it to further narrow down the filter we pass benchmark_zlib.

Dead2:
I get that, and I know the intentions are good. I just fear that it is adding benchmarks that won't actually get used, and then what good do they do?

I am happy to keep adding benchmarks for all permutations if that is what people want, I just know with myself that I will end up never using most of them. Perhaps I'll use some of them once or twice, but it'll never be to catch regressions in new PRs for example.

I do use filters today. For a new contributor it might (currently) be quite a task to run gbench to look for performance changes before they submit a PR.

In short, I am unsure about what direction we should take with regards to gbench. All The Benchmarks, or just a selection of the most important ones? Where to draw the line, or not.

nmoinvaz:
I've completely changed the way that I run benchmarks. I don't run them on the command line by typing the command myself. I simply tell Claude what benchmarks I want to run or what code I want to benchmark. It can select the right benchmarks or write the command line or run the benchmarks for me before I confirm. Claude builds the base and contender benchmarks and runs them and outputs JSON format which I then run against a Python script to compare the output of those two runs. Doing benchmarks the old manual way is simply too laborious. Even if I would use deflatebench, I would run it through Claude. I've found that as long as I give it the data or benchmarks that it needs, it is able to use that empirical data to better inform me. So, will I run all benchmarks manually all the time? Probably not. Even all the ones in deflate? Maybe. But when I'm tracking down an issue I have another lever I can pull on. If a change I make performs well on text data, I can tell Claude, "great now check how it performs on the other data types". If a change does not do well, I can tell Claude, "go analyze the assembly that changed between the two builds and see why that might be". Claude is better able to analyze my benchmark results, it knows when there is thermal bias. It makes sense of these long tables of numbers that I would otherwise have to yo-yo scroll to compare.

This gist of my contains the guidance I give to Claude: https://gist.github.com/nmoinvaz/90f2e4fcfcef9548a6241b1cd1e10336.

nmoinvaz:
Cross-varying by data-type is the best way to get a complete performance picture when tracking down an issue. `silesia` tests are an amalgamation of data types, so why something is slower or faster is not always readily available.

mtl1979:
Just my worthless 5 cents comment, but could this be handled as build-time option... I don't see a good way to handle this as command-line option for the benchmark executable unless we add separate prefix for the data-type dependent benchmarks and exclude the prefix by default. If a commit doesn't change the relevant paths, we would not need to run all the benchmarks and as such CI execute time wouldn't blow up exponentially...

I'm still against justifying larger benchmarks by claiming that running through AI bot is still easy... A lot of people don't have access to larger AI models due to recent US regulations and due to not using them at work on a server located in US... Currently local Claude AI models are limited to 256k output tokens for non-US people and even 1M tokens is barely enough for serious work. I personally have no problems running AI models that are around 16 GB in size, but the front-end limitation is something that I can't bypass...

Dead2:
I agree, a compile-time opt-in for EXTENSIVE_BENCHMARKS or something would be nice

test/test_data_p.h
```diff
@@ -0,0 +1,345 @@
...
+        case TEST_DATA_SHORT_MATCH:    return gen_short_match_data(bufsize);
+        case TEST_DATA_DNA:            return gen_dna_data(bufsize);
+        case TEST_DATA_RANDOM:         return gen_random_data(bufsize);
+        case TEST_DATA_LITERALS:       return gen_literals_data(bufsize);
+        case TEST_DATA_MIXED:          return gen_mixed_data(bufsize);
+        case TEST_DATA_REALISTIC_RGB:  return gen_realistic_rgb_data(bufsize);
+        case TEST_DATA_STRIPED_RGB:    return gen_striped_rgb_data(bufsize);
+    }
+    return NULL;
+}
```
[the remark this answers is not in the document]
nmoinvaz: Done. Applied through the Dispatch/DoSetUp chain as well so the enum type flows end to end.

nmoinvaz:
@Dead2 I have added a new flag for the data types. `--benchmark_data_types`.

test/benchmarks/benchmark_inflate.cc
```diff
@@ -106,34 +108,31 @@ class inflate_bench: public benchmark::Fixture {
         int index = 0;
         while (sizes[index] != (uint32_t)state.range(0)) ++index;
```
mtl1979: This line smells like it should be handled already in `DoSetup()`... Handling it early would also mean that we don't need to allocate multiple buffers, just one for each test run. It would reduce the test setup time significantly as already pointed out by the all-wise AI bot...

mtl1979 (changes requested):

test/benchmarks/benchmark_inflate.cc
nmoinvaz: Outside the scope of the PR though.

mtl1979: Scope is something that a lot of developers use as an excuse when they change something which end up snowballing and affect larger piece of code than originally anticipated. I usually hear it as an excuse from first-time contributors, not from people who have contributed for longer period of time.

nmoinvaz: Maybe if the scope of the PR was small and the scope of the change was small I would consider your request, but that is not the case.

mtl1979: However it is direct result of the change in this PR. It would be a totally different case if it would not be due to a change in this PR.

nmoinvaz: I see. Done. DoSetUp() now compresses only the size being benchmarked into a single buffer.

Dead2 (approved):

PR #2292, opened 2026-05-12
[CI] Cleanup action caches.

mtl1979:
* Clears branch specific GitHub Actions caches automatically when pull request is closed or when triggered manually on a branch where the workflow exists

### DIFF
--- a/.github/workflows/cache.yml
+++ b/.github/workflows/cache.yml
@@ -1,0 +1,35 @@
+# Cleanup caches by a branch, original version from actions/cache documentation
+name: Cache Cleanup
+on:
+  pull_request:
+    types:
+      - closed
+  workflow_dispatch:
+
+jobs:
+  cleanup:
+    name: Cache Cleanup
+    runs-on: ubuntu-latest
+    permissions:
+      # `actions:write` permission is required to delete caches
+      #   See also: https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28#delete-a-github-actions-cache-for-a-repository-using-a-cache-id
+      actions: write
+      contents: read
+    steps:
+      - name: Cleanup
+        run: |
+          echo "Fetching list of cache keys"
+          cacheKeysForPR=$(gh cache list --ref $BRANCH --limit 100 --json id --jq '.[].id')
+
+          ## Setting this to not fail the workflow while deleting cache keys.
+          set +e
+          echo "Deleting caches..."
+          for cacheKey in $cacheKeysForPR
+          do
+              gh cache delete $cacheKey
+          done
+          echo "Done"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          BRANCH: ${{ github.event.ref || format('refs/pull/{0}/merge', github.event.pull_request.number) }}


.github/workflows/cache.yml
```diff
@@ -0,0 +1,33 @@
...
+          echo "Deleting caches..."
+          for cacheKey in $cacheKeysForPR
+          do
+              gh cache delete $cacheKey
+          done
+          echo "Done"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          BRANCH: ${{ github.event.ref || format('refs/pull/{0}/merge', github.event.pull_request.number) }}
```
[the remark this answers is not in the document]
mtl1979: For `workflow_dispatch` trigger, `github.event.ref` is not empty, it's for example `refs/heads/develop`.

.github/workflows/cache.yml
```diff
@@ -0,0 +1,33 @@
+name: Cleanup caches by a branch
+on:
+  pull_request:
+    types:
+      - closed
+  workflow_dispatch:
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
```
nmoinvaz, line 12: I think we need a `name:` field here. Maybe like "Cache Cleanup"? That way it shows up properly under Actions tab with a friendly name on the left hand side.

.github/workflows/cache.yml
```diff
@@ -0,0 +1,33 @@
+name: Cleanup caches by a branch
```
nmoinvaz: I think this is more of a description rather than a name. Maybe better suited for a `#` comment and use the same name you choose for the `jobs` field.

mtl1979: I think it's that way so it stands out from all the other workflows. It could be shorter though...

.github/workflows/cache.yml
mtl1979: Always open for suggestions... None of these were even visible until I merged the commit to `develop` on my own repository... Original code needed some fixes elsewhere too...

nmoinvaz: I just suggest using "Cache Cleanup". I think it will look nice here. Or even "PR Cache Cleanup". Most of our names (Not Copilot) use first letter cap for all the words for the name.

image

.github/workflows/cache.yml
```diff
@@ -0,0 +1,35 @@
...
+    permissions:
+      # `actions:write` permission is required to delete caches
+      #   See also: https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28#delete-a-github-actions-cache-for-a-repository-using-a-cache-id
+      actions: write
+      contents: read
+    steps:
+      - name: Cleanup
+        run: |
+          echo "Fetching list of cache keys"
+          cacheKeysForPR=$(gh cache list --ref $BRANCH --limit 100 --json id --jq '.[].id')
```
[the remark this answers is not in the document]
mtl1979: The list for all branches was 5 full pages on GitHub website when the cache was full... It's unlikely that any branch can accumulate more than 100 cache entries before going over the 10 GB limit.

.github/workflows/cache.yml
mtl1979: Already changed in latest version...

.github/workflows/cache.yml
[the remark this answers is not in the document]
mtl1979: On my personal copy, there was 123 cache entries after I cleaned up both `develop` and `cache-cleanup` branches for testing purposes, total size was about 10.2 GB, so slightly over 10 GB limit. There was 96 cache entries remaining when I got the total usage below 10 GB limit. Currently the total cache usage on my personal repository is 8.03 GB after cleaning caches older than 24 hours. 

In this repository, the strategy is to cleanup caches that are older than 7 days if the total usage is less than 10 GB to avoid evicting caches that are expensive to recreate.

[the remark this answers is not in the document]
mtl1979: When caches from `develop` can be reused, the real world numbers for first day of week can be around 50 for `develop` branch with 15-40 cache entries for PR branches depending on run order until new calendar week starts. As stated earlier, `develop` isn't always the first branch that is pushed to GitHub.

nmoinvaz (approved):

Issue #64, opened 2016-04-21
Stable release


ioquatix:
Hello.

Sorry to be a burden by my request.

I'm simply interested to know if this is a stable project and whether it can be used in production.

I'm not expecting a definitive answer, but perhaps I'm not the only one with this question.

Dead2:
As a shared library replaceing zlib system-wide; no, zlib-ng has a few crash issues with some applications, presumably due to differences in alignment.

Statically linked or LD_PRELOAD or otherwise for a specific application it should be pretty stable, although you of course have to do your own testing before rolling this out into production systems.

Confidence-wise, I would not use zlib-ng for storage of important data at this point. For compression of html/etc to be delivered to a web-browser on-demand should not be a problem, since if you stumble upon a bug then you can just revert to stock zlib or a bugfixed zlib-ng without loosing important data.

We do need testers though, so any experiences both positive and negative would be appreciated. Currently we have no idea what applications crash with zlib-ng, and whether a recompile of that application to use zlib-ng would fix the problem.

Adenilson:
@Dead2 any news about a release?

ioquatix:
I'm not sure if this helps but I can explain our use case.

We are delivering images compressed using libpng. libpng uses libz for compression. We saw a significant improvement in performance using zlib-ng to generate png files. The png files are delivered to web browsers so they are not critical, but latency is very important. We haven't seen any issues or corruption, we compile the zlib-ng source directly into the application and link directly against it to produce a static binary.

Dead2:
@Adenilson I am planning to get back to zlib-ng soon. We had some big deadlines at work before and during the summer, so time has been very short. Hoping to get the remaining issues worked out this fall.

@ioquatix That is great feedback, thank you. Real-world testing beats narrow benchmarks any day 👍

I have been doing some smaller zlib-ng work since my last commit, mostly testing and validation related, such as static analysis, valgrind runs etc. Haven't had time to test it in any applications yet unfortunately.

I am wondering whether we should rename the library before any release is done though, so it can be parallel-installed with stock zlib. Could be made an optional thing, but I haven't looked at how much needs to change to do this properly. Could be done in relation to the zlib-compat compile option perhaps, so by default we use our new name, but compiling with zlib-compat would make it use zlib-naming as well. Thoughts and PRs on this would be much appreciated.
Distros that want to switch to zlib-ng completely could choose to have a zlibng package and a zlibng-compat package, so both libraries are provided by us.

[...]

PR #2294, opened 2026-05-12
deflate / inflate state cleanup

Dead2:
Remove obsolete padding in deflate and inflate state.
Reorder some elements to better pack and improve cache-locality in deflate state.

### DIFF
--- a/deflate.h
+++ b/deflate.h
@@ -238,33 +238,7 @@ struct ALIGNED_(64) internal_state {
     int heap_len;               /* number of elements in the heap */
     int heap_max;               /* element of largest frequency */
 
-    int32_t padding1[1];
-
-                /* Cacheline 3 */
-    uint8_t ALIGNED_(16) padding4[68];
-
                 /* used by trees.c: */
-    /* Didn't use ct_data typedef below to suppress compiler warning */
-    struct ct_data_s dyn_ltree[HEAP_SIZE];   /* literal and length tree */
-    struct ct_data_s dyn_dtree[2*D_CODES+1]; /* distance tree */
-    struct ct_data_s bl_tree[2*BL_CODES+1];  /* Huffman tree for bit lengths */
-
-    struct tree_desc_s l_desc;               /* desc. for literal tree */
-    struct tree_desc_s d_desc;               /* desc. for distance tree */
-    struct tree_desc_s bl_desc;              /* desc. for bit length tree */
-
-    uint16_t bl_count[MAX_BITS+1];
-    /* number of codes at each bit length for an optimal tree */
-
-    int heap[2*L_CODES+1];      /* heap used to build the Huffman trees */
-    /* The sons of heap[n] are heap[2*n] and heap[2*n+1]. heap[0] is not used.
-     * The same heap array is used to build all trees.
-     */
-
-    unsigned char depth[2*L_CODES+1];
-    /* Depth of each subtree used as tie breaker for trees of equal frequency
-     */
-
     unsigned int  lit_bufsize;
     /* Size of match buffer for literals/lengths.  There are 4 reasons for
      * limiting lit_bufsize to 64K:
@@ -285,6 +259,8 @@ struct ALIGNED_(64) internal_state {
      *   - I can't count above 4
      */
 
+                /* Cacheline 3 */
+
 #ifdef LIT_MEM
 #   define LIT_BUFS 5
     uint16_t *d_buf;              /* buffer for distances */
@@ -299,6 +275,27 @@ struct ALIGNED_(64) internal_state {
 
     unsigned int opt_len;         /* bit length of current block with optimal trees */
     unsigned int static_len;      /* bit length of current block with static trees */
+
+    struct tree_desc_s l_desc;               /* desc. for literal tree */
+    struct tree_desc_s d_desc;               /* desc. for distance tree */
+    struct tree_desc_s bl_desc;              /* desc. for bit length tree */
+
+    uint16_t bl_count[MAX_BITS+1];
+    /* number of codes at each bit length for an optimal tree */
+
+    int heap[2*L_CODES+1];      /* heap used to build the Huffman trees */
+    /* The sons of heap[n] are heap[2*n] and heap[2*n+1]. heap[0] is not used.
+     * The same heap array is used to build all trees.
+     */
+
+    unsigned char depth[2*L_CODES+1];
+    /* Depth of each subtree used as tie breaker for trees of equal frequency
+     */
+
+    /* Didn't use ct_data typedef below to suppress compiler warning */
+    struct ct_data_s dyn_ltree[HEAP_SIZE];   /* literal and length tree */
+    struct ct_data_s dyn_dtree[2*D_CODES+1]; /* distance tree */
+    struct ct_data_s bl_tree[2*BL_CODES+1];  /* Huffman tree for bit lengths */
 
     deflate_allocs *alloc_bufs;
 
--- a/inflate.h
+++ b/inflate.h
@@ -137,11 +137,6 @@ struct ALIGNED_(64) inflate_state {
     uint32_t have;              /* number of code lengths in lens[] */
     code *next;                 /* next available space in codes[] */
 
-#ifdef ARCH_32BIT
-    uint32_t padding[1];
-#endif
-    uint8_t ALIGNED_(16) padding4[68];
-
     uint16_t lens[320];         /* temporary storage for code lengths */
     uint16_t work[288];         /* work area for code table building */
     code codes[ENOUGH];         /* space for code tables */


Dead2:
### Develop
```
       |-          Compress                           -|   |-        Decompress                             -|
 Level  Comp %     Out size  Comptime min/avg/max/stddev    Comp %      In size  Decomptime min/avg/max/stddev
     1 54.185%      8526745  0.0532/0.0541/0.0546/0.0005   44.409%     94127497    0.2083/0.2093/0.2105/0.0006
     2 43.871%      6903702  0.0763/0.0765/0.0767/0.0001   35.519%     75286322    0.2070/0.2076/0.2085/0.0003
     3 42.388%      6670239  0.0910/0.0918/0.0921/0.0002   33.882%     71815206    0.1953/0.1965/0.1977/0.0005
     4 41.647%      6553746  0.1093/0.1102/0.1105/0.0003   33.175%     70317229    0.1912/0.1922/0.1927/0.0003
     5 41.216%      6485936  0.1224/0.1233/0.1235/0.0002   32.661%     69228146    0.1896/0.1904/0.1912/0.0004
     6 41.038%      6457827  0.1581/0.1592/0.1596/0.0003   32.507%     68902143    0.1880/0.1890/0.1902/0.0005
     7 40.778%      6416941  0.2291/0.2299/0.2310/0.0003   32.255%     68366759    0.1901/0.1911/0.1919/0.0004
     8 40.704%      6405249  0.3029/0.3040/0.3052/0.0005   32.167%     68180762    0.1899/0.1906/0.1914/0.0005
     9 40.409%      6358951  0.3391/0.3401/0.3418/0.0006   31.887%     67586442    0.1809/0.1819/0.1828/0.0005

 avg1  42.915%                                    0.1655   34.274%                                      0.1943

 177164    1336       8  178508   2b94c libz-ng.so.2
```

### PR
```
       |-          Compress                           -|   |-        Decompress                             -|
 Level  Comp %     Out size  Comptime min/avg/max/stddev    Comp %      In size  Decomptime min/avg/max/stddev
     1 54.185%      8526745  0.0531/0.0534/0.0537/0.0001   44.409%     94127497    0.2079/0.2088/0.2102/0.0006
     2 43.871%      6903702  0.0755/0.0765/0.0768/0.0003   35.519%     75286322    0.2065/0.2074/0.2085/0.0005
     3 42.388%      6670239  0.0906/0.0918/0.0924/0.0004   33.882%     71815206    0.1955/0.1963/0.1975/0.0005
     4 41.647%      6553746  0.1093/0.1101/0.1106/0.0004   33.175%     70317229    0.1909/0.1919/0.1933/0.0005
     5 41.216%      6485936  0.1224/0.1232/0.1236/0.0003   32.661%     69228146    0.1893/0.1902/0.1912/0.0005
     6 41.038%      6457827  0.1583/0.1592/0.1595/0.0003   32.507%     68902143    0.1878/0.1890/0.1901/0.0006
     7 40.778%      6416941  0.2289/0.2301/0.2313/0.0005   32.255%     68366759    0.1898/0.1909/0.1916/0.0005
     8 40.704%      6405249  0.3026/0.3041/0.3054/0.0007   32.167%     68180762    0.1898/0.1907/0.1919/0.0006
     9 40.409%      6358951  0.3386/0.3400/0.3410/0.0005   31.887%     67586442    0.1807/0.1817/0.1824/0.0005

 avg1  42.915%                                    0.1654   34.274%                                      0.1941

 177100    1336       8  178444   2b90c libz-ng.so.2
```

deflate.h
```diff
@@ -296,10 +272,30 @@ struct ALIGNED_(64) internal_state {
 
     unsigned int sym_next;        /* running index in symbol buffer */
     unsigned int sym_end;         /* symbol table full when sym_next reaches this */
-
```
mtl1979, line 299: Nitpick: We don't mix layout/code changes and unconnected white-space changes.

inflate.h
```diff
@@ -137,11 +137,6 @@ struct ALIGNED_(64) inflate_state {
     uint32_t have;              /* number of code lengths in lens[] */
     code *next;                 /* next available space in codes[] */
 
-#ifdef ARCH_32BIT
-    uint32_t padding[1];
-#endif
-    uint8_t ALIGNED_(16) padding4[68];
```
mtl1979, line 143: This would not have side-effects if the padding was exactly size of a cacheline. There is no other changes to "compensate" or explain the difference (4 bytes more than size of a cacheline on 64-bit targets), unless the padding was incorrectly sized in the old version.

mtl1979 (commented):
Hard to verify cacheline boundaries when they are not marked in neither old or new code. Often updating the cacheline boundary comments is delayed so long that people don't remember the original PR that caused the discrepancy.

nmoinvaz:
Verified by dumping `offsetof(deflate_state, …)` for every field on 64-bit (LP64) and comparing against `develop`.

### Cacheline annotations land on 64-byte boundaries

| Marker | Field | Offset |
|---|---|---|
| `Cacheline 0` | `strm` | 0 |
| `Cacheline 1` | `lookahead` | 64 |
| `Cacheline 2` | `match_start` | 128 |
| `Cacheline 3` | `sym_buf` (or `d_buf`) | 192 |

`Cacheline 3` lines up cleanly because `lit_bufsize` was promoted to slot 60 of cacheline 2 (offset 188–192), so the symbol-buffer pointer falls on the 192-byte boundary without any explicit padding.

### Sizes

| | develop | PR | Δ |
|---|---|---|---|
| `sizeof(deflate_state)` | 6040 | 5968 | −72 B |
| `sizeof(inflate_state)` | 9152 | 9080 | −72 B |

### Layout reorder

The new ordering clusters by usage rather than declaration history:

1. Hot symbol-buffer writes (`sym_buf`/`sym_next`/`sym_end`/`opt_len`/`static_len`) in cacheline 3, adjacent to the match metadata in cacheline 2.
2. Tree descriptors and `bl_count` in cachelines 3–4.
3. `heap` + `depth` (used together in `pqdownheap`/`build_tree`).
4. Bulky `dyn_ltree`/`dyn_dtree`/`bl_tree`.

On `develop`, `sym_buf` and the tree arrays sat ~5.7 KB apart with `lit_bufsize` over 5.7 KB away from the rest of the symbol-buffer state — those are the touch sets that now share cachelines.

### Caveats

The `Cacheline N` markers assume LP64. On ILP32 the 4-byte-narrower pointers shift everything earlier (cacheline 0's group ends at offset 48, not 64). That mismatch was already present on `develop` and is not introduced here.

### Tests

Full `ctest` on `build-default`: 742/742 passing.

nmoinvaz:
While `inflate.h` is open, two small reorders from #2125 might be worth folding in. Both follow the same pattern — moving a cold field out from between two hot ones.

**Swap `havedict` and `flags`** so `wrap` and `flags` sit adjacent.

```
inflate.c gzip-header path repeatedly tests:  state->flags & X  &&  state->wrap & Y

  develop                              after swap
  ────────────────────                 ────────────────────
  off 16  │ wrap     │ ◄─┐             off 16  │ wrap     │ ◄┐
  off 20  │ havedict │   │ (cold)      off 20  │ flags    │ ◄┴── ldr [#16]  (8B pair)
  off 24  │ flags    │ ◄─┘             off 24  │ havedict │   (cold, DICTID only)
```

The `(state->flags & 0x0…) && (state->wrap & 4)` pattern shows up ~10 times in the gzip-header parser in `inflate.c`. After the swap a single 8-byte load covers both. `havedict` moves to offset 24, still in cacheline 0, only touched on the cold DICTID path.

**Move `wbufsize` past `wnext`** so the order becomes `wsize, whave, wnext, wbufsize, window`.

```
inffast_tpl.h:125–128 reads:  wsize → whave → wnext → window

  develop                              after move
  ────────────────────                 ────────────────────
  off 64  │ wsize    │ ◄── ldr         off 64  │ wsize    │ ◄┐
  off 68  │ wbufsize │   (cold)        off 68  │ whave    │ ◄┴── ldp [#64]
  off 72  │ whave    │ ◄┐              off 72  │ wnext    │ ◄┐
  off 76  │ wnext    │ ◄┴── ldp [#72]  off 76  │ wbufsize │ ◄┴── ldp [#72] *
  off 80  │ window   │ ◄── ldr         off 80  │ window   │ ◄── ldr

  * wbufsize loaded into a paired register, immediately discarded
```

Instruction count is the same (3 loads either way), but all three hot 32-bit fields land in one contiguous 16-byte block. `wbufsize` (cold, alloc-time only) doubles as the alignment filler that keeps `window` 8-byte aligned at offset 80.

Both reorders are mechanical — no padding, no field renames.

inflate.h
Dead2: The padding was placed there to replace the old vector crc temporary storage that was being removed, to avoid the struct changes affecting benchmark numbers from the crc changes.

deflate.h
Dead2: Yeah, missed that change when I reverted some other changes that didn't pan out.

Dead2:
The cacheline boundary comments are not really important, they are mostly useful for me manually counting out the cacheline, so I don't have to count all the way from the start of the struct while reviewing changes.
Also they are only accurate for 64-bit platforms (Actually the M2 cpu apparently has 128-byte cachelines, but that is only better for us, not a complication)

When I need to verify they are correct, I use pahole to do so.

Dead2:
> While `inflate.h` is open, two small reorders from #2125 might be worth folding in. Both follow the same pattern — moving a cold field out from between two hot ones.
> 
> **Swap `havedict` and `flags`** so `wrap` and `flags` sit adjacent.
> 
> ```
> inflate.c gzip-header path repeatedly tests:  state->flags & X  &&  state->wrap & Y
> 
>   develop                              after swap
>   ────────────────────                 ────────────────────
>   off 16  │ wrap     │ ◄─┐             off 16  │ wrap     │ ◄┐
>   off 20  │ havedict │   │ (cold)      off 20  │ flags    │ ◄┴── ldr [#16]  (8B pair)
>   off 24  │ flags    │ ◄─┘             off 24  │ havedict │   (cold, DICTID only)
> ```
> 
> The `(state->flags & 0x0…) && (state->wrap & 4)` pattern shows up ~10 times in the gzip-header parser in `inflate.c`. After the swap a single 8-byte load covers both. `havedict` moves to offset 24, still in cacheline 0, only touched on the cold DICTID path.
> 
> **Move `wbufsize` past `wnext`** so the order becomes `wsize, whave, wnext, wbufsize, window`.
> 
> ```
> inffast_tpl.h:125–128 reads:  wsize → whave → wnext → window
> 
>   develop                              after move
>   ────────────────────                 ────────────────────
>   off 64  │ wsize    │ ◄── ldr         off 64  │ wsize    │ ◄┐
>   off 68  │ wbufsize │   (cold)        off 68  │ whave    │ ◄┴── ldp [#64]
>   off 72  │ whave    │ ◄┐              off 72  │ wnext    │ ◄┐
>   off 76  │ wnext    │ ◄┴── ldp [#72]  off 76  │ wbufsize │ ◄┴── ldp [#72] *
>   off 80  │ window   │ ◄── ldr         off 80  │ window   │ ◄── ldr
> 
>   * wbufsize loaded into a paired register, immediately discarded
> ```
> 
> Instruction count is the same (3 loads either way), but all three hot 32-bit fields land in one contiguous 16-byte block. `wbufsize` (cold, alloc-time only) doubles as the alignment filler that keeps `window` 8-byte aligned at offset 80.
> 
> Both reorders are mechanical — no padding, no field renames.

Do you have any benchmarks to back that up?

And how is a 8-byte contiguous load beneficial? If the variables end up in separate registers, they cannot share a single 8-byte load from cache. And all loads from memory are cacheline-sized, so there is no benefit there either unless they are otherwise in different cachelines. Alignment does matter, but they are already aligned. At least that goes for x86 architectures, don't know about other architectures, but I'd imagine they operate similarly.

None of my systems show any benefit from this, more often a slight detriment actually, probably because the code size goes up again.

For example on the 8700GE:
```
 Level  Comp %      In size  Decomptime min/avg/max/stddev
     1 44.409%     94127497    0.2085/0.2094/0.2105/0.0004
     2 35.519%     75286322    0.2068/0.2074/0.2079/0.0003
     3 33.882%     71815206    0.1959/0.1965/0.1976/0.0004
     4 33.175%     70317229    0.1915/0.1922/0.1933/0.0004
     5 32.661%     69228146    0.1898/0.1904/0.1909/0.0003
     6 32.507%     68902143    0.1883/0.1893/0.1901/0.0004
     7 32.255%     68366759    0.1904/0.1910/0.1918/0.0003
     8 32.167%     68180762    0.1904/0.1910/0.1919/0.0004
     9 31.887%     67586442    0.1816/0.1821/0.1828/0.0003

 avg1  34.274%                                      0.1944

 177164    1336       8  178508   2b94c libz-ng.so.2
```

nmoinvaz:
> Do you have any benchmarks to back that up?

No, my benchmarking rig is not sensitive enough to be able to detect changes like that.

> And how is a 8-byte contiguous load beneficial? If the variables end up in separate registers, they cannot share a single 8-byte load from cache. And all loads from memory are cacheline-sized, so there is no benefit there either unless they are otherwise in different cachelines. Alignment does matter, but they are already aligned. At least that goes for x86 architectures, don't know about other architectures, but I'd imagine they operate similarly.

I think you are right about it being cacheline-sized anyways. Aarch64 does have [ldp](https://developer.arm.com/documentation/ddi0596/2020-12/Base-Instructions/LDP--Load-Pair-of-Registers-) instruction that can load 2 registers at a time. So the only win would be assembly generation.

> None of my systems show any benefit from this, more often a slight detriment actually, probably because the code size goes up again.

I double checked the size of the library and it did go up with those suggestions from my original PR by a few bytes on Aarch64. So it looks like it doesn't generate that `ldp` instruction like I originally thought it did or in one instance it might have generated slightly differently.

nmoinvaz (approved):

Dead2:
> I think you are right about it being cacheline-sized anyways. Aarch64 does have [ldp](https://developer.arm.com/documentation/ddi0596/2020-12/Base-Instructions/LDP--Load-Pair-of-Registers-) instruction that can load 2 registers at a time. So the only win would be assembly generation.

Right, and 32-bit arm has ldm. However you only really benefit if performance is limited by load capacity, and I don't really think we're close to that.

PR #2296, opened 2026-05-12
[CI] Extend caching apt packages to pigz workflow.

mtl1979:
* This cache should be able to bootstrap from existing `clang15` cache in `cmake.yml` as there is overlap in installed packages
* Because the primary keys are different between `cmake.yml` and `pigz.yml`, using same `cache-key` doesn't result in fighting over which workflow file is run first


mtl1979:
Looking at the logs, 3 packages loaded from restore-key cache, 4 packages downloaded...

nmoinvaz (approved):

PR #2298, opened 2026-05-13
Move LFS64 detection to preprocessor

### LINKED ISSUE #2002: Building for iOS simulator (x86_64) fails on check_type_size
Hello! I'm afraid I'm back with another zlib-ng for iOS build issue. 

Last time you helped me see that we had to build zlib-ng for each architecture separately, whereas other projects can take advantage of XCode's ability to build multiple architectures simultaneously (and then combine them into a fat binary). That's still my goal, but I'm having trouble building for x86_64. I think last time I mentioned that I might be able to drop x86_64, but apparently not.

The problem I'm having is that when its generating the build files it fails on `check_type_size(off64_t OFF64_T)` (root CMakeLists.txt, line 426 or so). I've found that if I just remove that section entirely (OFF64_T) it generates just fine and builds without issue. If I leave it in, I get an error like this:

```
CMake Error at /usr/local/share/cmake-3.25/Modules/CheckTypeSize.cmake:148 (try_compile):
  Cannot copy output executable

    ''

  to destination specified by COPY_FILE:

    '/Users/Andrew/Documents/Andrew/iOS/ThirdParty/cmake/zlib-ng/CMakeFiles/CheckTypeSize/OFF64_T.bin'

  Recorded try_compile output location doesn't exist:

    /Users/Andrew/Documents/Andrew/iOS/ThirdParty/cmake/zlib-ng/CMakeFiles/CMakeScratch/TryCompile-InE4W4/Debug/cmTC_c86fe.app/cmTC_c86fe

  /usr/local/share/cmake-3.25/Modules/CheckTypeSize.cmake:278 (__check_type_size_impl)
  CMakeLists.txt:426 (check_type_size)
```

I think I mentioned this error last time but we moved past it quickly to the multi-architecture problem. When I Google this, I find a bunch of hits, but they're all related to opencv and Kitware & opencv seem to point the finger at each other. It sounds like there was a bug in CMake 3.25.0 and they fixed, but also that opencv was doing something hinky with their test. Given how straight forward your call is, I'm pretty confident you're not doing anything like opencv seems to have done.

Any suggestions on what I might try? The error doesn't give me much in the way of ideas to explore.

Any help appreciated! Thanks!

nmoinvaz:
Replace the CMake `check_type_size(off64_t OFF64_T)` probe (plus `_off64_t` / `__off64_t` follow-ups) with an unconditional `#define _LARGEFILE64_SOURCE 1` in `zbuild.h`. The real gating already lives in `zconf.h.in` via `_LFS64_LARGEFILE`, which glibc/musl set in `<unistd.h>` when LFS64 is supported, and the `HAVE_OFF64_T*` results were never consumed.

Avoiding `try_compile` here sidesteps a class of CMake bundle-generator bugs that misreport the test binary path on iOS, Android NDK, and other multi-arch SDK builds. Fixes #2002.

Bundled dead-code removals exposed by the move: `NO_FSEEKO` detection (zlib-ng calls `lseek*` directly), `gzguts.h` feature-macro juggling (only meaningful when `_LARGEFILE64_SOURCE` was conditional), and a redundant `unlink` forward declaration in the minigzip tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -491,23 +491,6 @@ if(HAVE_ASM_HWPROBE_H)
 endif()
 
 #
-# Check to see if we have large file support
-#
-set(CMAKE_REQUIRED_DEFINITIONS -D_LARGEFILE64_SOURCE=1 -D__USE_LARGEFILE64)
-check_type_size(off64_t OFF64_T)
-if(HAVE_OFF64_T)
-    add_definitions(-D_LARGEFILE64_SOURCE=1 -D__USE_LARGEFILE64)
-else()
-    check_type_size(_off64_t _OFF64_T)
-    if(HAVE__OFF64_T)
-        add_definitions(-D_LARGEFILE64_SOURCE=1 -D__USE_LARGEFILE64)
-    else()
-        check_type_size(__off64_t __OFF64_T)
-    endif()
-endif()
-set(CMAKE_REQUIRED_DEFINITIONS) # clear variable
-
-#
 # Check for fseeko and other optional functions
 #
 set(CMAKE_REQUIRED_FLAGS "${ADDITIONAL_CHECK_FLAGS}")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -491,15 +491,8 @@ if(HAVE_ASM_HWPROBE_H)
 endif()
 
 #
-# Check for fseeko and other optional functions
+# Check for strerror
 #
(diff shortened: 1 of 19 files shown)


mtl1979:
Shuffling all the headers so that `zbuild.h` is first again could be own cleanup PR... We don't have FreeBSD and OpenBSD in CI yet, so I'm not really 100% sure that there is any regressions until I boot up to at least FreeBSD and test the changes...

Dead2 (approved):

PR #2299, opened 2026-05-14
Make CRC32 checking of headers use crc32_small directly

Dead2:
Make CRC32 checking of headers use crc32_small directly, instead of taking a detour through functable to spin up a vector optimized function when the header is too small to be handled by it.

Also removes UPDATE macros that apparently are no longer used.

### DIFF
--- a/crc32.c
+++ b/crc32.c
@@ -9,12 +9,21 @@
 
 #include "zbuild.h"
 #include "functable.h"
-#include "crc32_braid_tbl.h"
+#include "crc32_p.h"
 
 /* ========================================================================= */
 
 const uint32_t * Z_EXPORT PREFIX(get_crc_table)(void) {
     return (const uint32_t *)crc_table;
+}
+
+/* crc32 function meant for short buffers like gzip headers */
+Z_INTERNAL uint32_t crc32_small(uint32_t crc, const uint8_t *buf, size_t len) {
+    if (UNLIKELY(len >= 32)){
+        return FUNCTABLE_CALL(crc32)(crc, buf, len);
+    }
+
+    return ~crc32_copy_small(~crc, NULL, buf, len, 32, 0);
 }
 
 #ifdef ZLIB_COMPAT
--- a/crc32_p.h
+++ b/crc32_p.h
@@ -8,6 +8,8 @@
 #ifndef CRC32_P_H
 #define CRC32_P_H
 
+#include "crc32_braid_tbl.h"
+
 #define CRC_DO1(c, buf, i) c = crc_table[(c ^ buf[i]) & 0xff] ^ (c >> 8)
 #define CRC_DO2(c, buf, i) {CRC_DO1(c, buf, i); CRC_DO1(c, buf, i+1);}
 #define CRC_DO4(c, buf, i) {CRC_DO2(c, buf, i); CRC_DO2(c, buf, i+2);}
--- a/deflate.c
+++ b/deflate.c
@@ -52,6 +52,7 @@
 #include "deflate.h"
 #include "deflate_p.h"
 #include "insert_string_p.h"
+#include "arch_functions.h"
 
 /* Avoid conflicts with zlib.h macros */
 #ifdef ZLIB_COMPAT
@@ -754,7 +755,7 @@ Z_INTERNAL void PREFIX(flush_pending)(PREFIX3(stream) *strm) {
 #define HCRC_UPDATE(beg) \
     do { \
         if (s->gzhead->hcrc && s->pending > (beg)) \
-            strm->adler = PREFIX(crc32)(strm->adler, s->pending_buf + (beg), s->pending - (beg)); \
+            strm->adler = crc32_small((uint32_t)strm->adler, s->pending_buf + (beg), s->pending - (beg)); \
     } while (0)
 
 /* ========================================================================= */
@@ -874,7 +875,7 @@ int32_t Z_EXPORT PREFIX(deflate)(PREFIX3(stream) *strm, int32_t flush) {
             if (s->gzhead->extra != NULL)
                 put_short(s, (uint16_t)s->gzhead->extra_len);
             if (s->gzhead->hcrc)
-                strm->adler = PREFIX(crc32)(strm->adler, s->pending_buf, s->pending);
+                strm->adler = crc32_small((uint32_t)strm->adler, s->pending_buf, s->pending);
             s->gzindex = 0;
             s->status = EXTRA_STATE;
         }
--- a/inflate.c
+++ b/inflate.c
@@ -11,6 +11,7 @@
 #include "inflate_p.h"
 #include "inffixed_tbl.h"
 #include "functable.h"
+#include "arch_functions.h"
 
 /* Avoid conflicts with zlib.h macros */
 #ifdef ZLIB_COMPAT
@@ -624,7 +625,7 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
                         }
                     }
                     if ((state->flags & 0x0200) && (state->wrap & 4)) {
-                        state->check = PREFIX(crc32)(state->check, next, copy);
+                        state->check = crc32_small((uint32_t)state->check, next, copy);
                     }
                     have -= copy;
                     next += copy;
@@ -647,7 +648,7 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
                         state->head->name[state->length++] = (unsigned char)len;
                 } while (len && copy < have);
                 if ((state->flags & 0x0200) && (state->wrap & 4))
-                    state->check = PREFIX(crc32)(state->check, next, copy);
+                    state->check = crc32_small((uint32_t)state->check, next, copy);
                 have -= copy;
                 next += copy;
                 if (len)
@@ -670,7 +671,7 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
                         state->head->comment[state->length++] = (unsigned char)len;
                 } while (len && copy < have);
                 if ((state->flags & 0x0200) && (state->wrap & 4))
-                    state->check = PREFIX(crc32)(state->check, next, copy);
+                    state->check = crc32_small((uint32_t)state->check, next, copy);
                 have -= copy;
                 next += copy;
                 if (len)
--- a/inflate_p.h
+++ b/inflate_p.h
@@ -9,6 +9,7 @@
 
 #include "zendian.h"
 #include "zmemory.h"
+#include "crc32_braid_tbl.h"
 
 /* Architecture-specific hooks. */
 #ifdef S390_DFLTCC_INFLATE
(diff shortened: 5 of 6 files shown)


crc32.c
```diff
@@ -9,14 +9,18 @@
...
 /* ========================================================================= */
 
 const uint32_t * Z_EXPORT PREFIX(get_crc_table)(void) {
     return (const uint32_t *)crc_table;
 }
 
+Z_INTERNAL uint32_t crc32_small(uint32_t crc, const uint8_t *buf, size_t len) {
+    return ~crc32_copy_small(~crc, NULL, buf, len, 128, 0);
+}
+
```
mtl1979, line 28: This should be declared in a header, so it can be inlined like the function it calls... I would move it to `crc32_p.h` and include that when needed.

mtl1979 (changes requested):
Making small wrapper function always inlined would let compiler optimize it better.

crc32.c
Dead2: In that case I'd just have used crc32_copy_small() directly, unfortunately the compiler inlines a lot of code because we don't know ahead of time how big the header data is, and it ends up with both bigger code and slower than this implementation.

mtl1979: There are lines where the `len` is a small constant that benefit from inlining... The real issue that needs to be resolved is mixing function calls where the `len` is constant with calls where it isn't a constant.

Because using `crc32_copy_small` needs bit negating both input and output, it is less readable than both the original code and using this function.

crc32_p.h
```diff
@@ -27,23 +29,23 @@ Z_FORCEINLINE static uint32_t crc32_copy_small(uint32_t crc, uint8_t *dst, const
...
+        if (DO_COPY) {
             memcpy(dst, buf, 2);
             dst += 2;
         }
         CRC_DO2(crc, buf, 0);
         buf += 2;
     }
     if (len & 1) {
-        if (COPY)
+        if (DO_COPY)
```
mtl1979: Renaming a function parameter should be separate PR or dropped completely... It bloats the diff unnecessarily.

Dead2: The name COPY is already defined in inflate

crc32.c
Dead2: > There are lines where the `len` is a small constant that benefit from inlining...

Could you point them out? I didn't see any.

Dead2:
That UBSAN error does not make sense to me.. ideas?

inflate_p.h
```diff
@@ -47,26 +48,18 @@
...
-#  define UPDATE(check, buf, len) FUNCTABLE_CALL(adler32)(check, buf, len)
-#endif
-
 /* check macros for header crc */
 #ifdef GUNZIP
 #  define CRC2(check, word) \
     do { \
         uint16_t tmp = Z_U16_TO_LE((uint16_t)(word)); \
-        check = PREFIX(crc32)(check, (const unsigned char *)&tmp, 2); \
+        check = ~crc32_copy_small(~check, NULL, (const uint8_t *)&tmp, 2, 2, 0); \
```
mtl1979: This is becoming a lot of dead code and barely readable... Bit negating both input and output should be handled in whatever function we decide to use, just like in the original code, not duplicated here and few lines below again...

Maybe once in prehistoric times this was simple enough to be handled in preprocessor but now it's complex enough that it would be cleaner as inlined function. 

Even using `CRC_DO2` directly would make this more readable.

inflate_p.h
```diff
@@ -47,26 +48,18 @@
...
     } while (0)
 
 #  define CRC4(check, word) \
     do { \
         uint32_t tmp = Z_U32_TO_LE((uint32_t)(word)); \
-        check = PREFIX(crc32)(check, (const unsigned char *)&tmp, 4); \
+        check = ~crc32_copy_small(~check, NULL, (const uint8_t *)&tmp, 4, 4, 0); \
```
mtl1979: As above, this is a lot of dead code now... Instead of `CRC_DO2` above, this would require using `CRC_DO4`.

inflate_p.h
nmoinvaz: Yeah very interesting it is named `CRC2` and does work of `CRC_DO2`, same with the one below. Crazy we've only noticed this now..

nmoinvaz: Actually, since `max_len/len` in call to `crc32_copy_small` is 2 it should only inline `CRC_DO2`. Would have to look at asm generation probably.

nmoinvaz: I tested it locally, the way it is, and Clang did very good on aarch64, exactly like what we would expect.

deflate.c
```diff
@@ -754,7 +755,7 @@ Z_INTERNAL void PREFIX(flush_pending)(PREFIX3(stream) *strm) {
 #define HCRC_UPDATE(beg) \
     do { \
         if (s->gzhead->hcrc && s->pending > (beg)) \
-            strm->adler = PREFIX(crc32)(strm->adler, s->pending_buf + (beg), s->pending - (beg)); \
+            strm->adler = crc32_small(strm->adler, s->pending_buf + (beg), s->pending - (beg)); \
```
nmoinvaz: `HCRC_UPDATE` can be called from within `EXTRA_STATE` which can be up-to 64KB. I've never heard of that field until now, and who knows if anybody uses it, but I thought I'd mention it. I think it is okay to leave it as is.

inflate_p.h
mtl1979: > Yeah very interesting it is named `CRC2` and does work of `CRC_DO2`, same with the one below. Crazy we've only noticed this now..

Pretty much only difference is that bit negating on input and output that bots call "conditioning"... Whole purpose of that is to mitigate the limitation that checksum of any sequence of zero bytes is same, but not any other byte value.

mtl1979: > Actually, since `max_len/len` in call to `crc32_copy_small` is 2 it should only inline `CRC_DO2`. Would have to look at asm generation probably.

"It should" sounds a lot like excuse... The byte swap on line above is only required because we need to extract lowest 16 (or 32 bits below) of the input and that can't be done effectively on `uint8_t` array when using just preprocessor... However it can be done if it is transformed into inline function.

Currently the code first narrows the input value to 2 bytes and then right after as aliased pointer to 1 byte, which is kind of redundant.

mtl1979: `inflate_state.check` is `unsigned long` or `uint64_t` but `crc32_copy_small` requires `uint32_t`, which is same as `unsigned int`. Thus UBSAN will complain about truncation.

nmoinvaz: It is hard to understand what you mean. But I put your comment through Claude and you make a good point. `CRC_DO2` prevents having to do `Z_U32_TO_LE` in code. @Dead2 this is some options from Claude: https://gist.github.com/nmoinvaz/8e994315219c81c750213e7eb52ec7cb if you are interested.

nmoinvaz: However both of ways (`crc32_copy_small` and `CRC_DOX`) produce the same assembly, at least in Clang.

mtl1979: > It is hard to understand what you mean. But I put your comment through Claude and you make a good point. 

Yeah... I think like an AI bot when I'm tired... It's almost midnight here and I've been working on several PRs with almost no sleep for days... Plus the current version trips UBSAN, so anything else is better...

nmoinvaz: I just chalk it up to me not knowing a particular naunce that you are zeroed in on.

mtl1979: Being almost blind means I focus on different things than people who have normal vision... And I have photographic memory, so when I zero on to something, I can compare patterns instead of individual characters or lines.

crc32_p.h
mtl1979: There is no conflict if it isn't defined by preprocessor.

Dead2: The compiler errors out when I don't rename, that's why I did the rename in the first case.

mtl1979: For me it looks like the two headers are included in "wrong" order... Compiler "thinks" any header that even indirectly includes this header must become before `inflate.h`... This is mostly issue with capitalized function parameter names and inlined functions when compiling C code.

That doesn't change my view that this change should be own PR... Nothing stops from merging that PR before this PR, thus the end result is the same, but doesn't need to wait for this PR to be finished.

Less invasive method is to use `#define COPY DO_COPY` before including this header and then right after use `#undef COPY`. This allows inserting rationale in inline comment before that code block.

Dead2: It would have been cleaner as `CRC_COPY`, but I am abandoning it.

inflate_p.h
```diff
@@ -47,26 +48,26 @@
...
-
 /* check macros for header crc */
+#define CRC_DO1_B(c, b)    c = crc_table[(c ^ (b)) & 0xff] ^ (c >> 8)
+
 #ifdef GUNZIP
 #  define CRC2(check, word) \
     do { \
-        uint16_t tmp = Z_U16_TO_LE((uint16_t)(word)); \
-        check = PREFIX(crc32)(check, (const unsigned char *)&tmp, 2); \
+        uint32_t crc = ~(check); \
```
mtl1979: Missing cast to `uint32_t`.

inflate_p.h
```diff
@@ -47,26 +48,26 @@
...
+        CRC_DO1_B(crc, (word)     ); \
+        CRC_DO1_B(crc, (word) >> 8); \
+        (check) = ~crc; \
     } while (0)
 
 #  define CRC4(check, word) \
     do { \
-        uint32_t tmp = Z_U32_TO_LE((uint32_t)(word)); \
-        check = PREFIX(crc32)(check, (const unsigned char *)&tmp, 4); \
+        uint32_t crc = ~(check); \
```
mtl1979: Missing cast to `uint32_t`.

inflate_p.h
Dead2: I was wondering about that, but went with the AI code to see.

mtl1979: I just noticed that the CI builds still failed... The line UBSAN mentions contain the macro name, but line is outside this patch, so the cause is not obvious although I did mention the reason earlier.

deflate.c
mtl1979: There are quite a lot of gzip header fields that `minigzip` doesn't use but normal `gzip` does support... If any field can be larger than 128 bytes or whatever we set as threshold for small CRC, then it shouldn't use `crc32_small` at all...

We could have a test that fill all the existing gzip header fields to maximum length and test that gunzip can parse the header without throwing an error.

Dead2: Changed crc32_small to forward to functable for longer lengths.

crc32.c
mtl1979: You already changed the 2 and 4 byte checksums to not use `crc32_small` or `crc32_copy_small`...

crc32_p.h
mtl1979: Now the patch is almost as clean as it can become...

inflate_p.h
```diff
@@ -47,26 +48,26 @@
...
     do { \
-        uint32_t tmp = Z_U32_TO_LE((uint32_t)(word)); \
-        check = PREFIX(crc32)(check, (const unsigned char *)&tmp, 4); \
+        uint32_t crc = ~(uint32_t)(check); \
+        CRC_DO1_B(crc, (word)      ); \
+        CRC_DO1_B(crc, (word) >>  8); \
+        CRC_DO1_B(crc, (word) >> 16); \
+        CRC_DO1_B(crc, (word) >> 24); \
+        (check) = ~crc; \
     } while (0)
```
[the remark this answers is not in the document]
mtl1979: The helper function makes the code less readable... Partial duplicating of code in only two places is borderline case. Whole purpose of the change was to make the code more readable and rely less on compiler optimizing the code.

inflate.c
```diff
@@ -624,7 +625,7 @@ int32_t Z_EXPORT PREFIX(inflate)(PREFIX3(stream) *strm, int32_t flush) {
                         }
                     }
                     if ((state->flags & 0x0200) && (state->wrap & 4)) {
-                        state->check = PREFIX(crc32)(state->check, next, copy);
+                        state->check = crc32_small(state->check, next, copy);
```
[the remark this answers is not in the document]
mtl1979: As long as both types are unsigned and there is no extra parentheses, explicit widening is optional between 32-bit and 64-bit types, but not from smaller types. With temporary variables caused by parentheses compiler is free to cast first to `signed int` and then widen to 64 bits.

nmoinvaz (approved):

mtl1979 (approved):
LGTM. 

A lot more readable code with less dead code.

PR #2303, opened 2026-05-16
build: add missing versioned symbols

gordonmessmer:
Symbols added to the map in this commit were present in 1.1.4 and are not versioned in the current release.

### DIFF
--- a/zlib.map.in
+++ b/zlib.map.in
@@ -1,3 +1,48 @@
+ZLIB_1.1.4 {
+  global:
+    @ZLIB_SYMBOL_PREFIX@adler32;
+    @ZLIB_SYMBOL_PREFIX@compress2;
+    @ZLIB_SYMBOL_PREFIX@compress;
+    @ZLIB_SYMBOL_PREFIX@crc32;
+    @ZLIB_SYMBOL_PREFIX@deflate;
+    @ZLIB_SYMBOL_PREFIX@deflateCopy;
+    @ZLIB_SYMBOL_PREFIX@deflateEnd;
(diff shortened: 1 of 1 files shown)


gordonmessmer:
In order to improve reliable dependency resolution, I would like to add versioned symbols to the libraries that do not include them today.

This change is intended to support Fedora's recommendation that shared libraries provide versioned symbols: https://docs.fedoraproject.org/en-US/packaging-guidelines/C_and_C++/#_versioned_symbols

zlib uses versioned symbols for symbols introduced in 1.2.0 and later, but not for earlier symbols, and zlib-ng does the same. As a result, programs that link to either library and use only the older API do not use any versioned symbols and provide dependency generators with limited information.

I've asked zlib to version the older symbols, and I believe they will merge that change: https://github.com/madler/zlib/pull/1237

Once they do, zlib-ng should also version those symbols for continued ABI compatibility.

mtl1979:
Once the symbol file is updated on stock zlib, we can consider updating it too... Generally we sync with stock zlib shortly before minor releases when we are sure the new changes don't introduce any regressions or serious bugs. Recent versions of stock zlib have introduced some bad code/changes so the PR to sync hasn't been merged yet.

nmoinvaz (approved):

nmoinvaz:
@Dead2 why not merge this?

Dead2:
It needs to be merged upstream first, so we can stay compatible.

nmoinvaz:
But this provides even more compatibility that zlib, so I guess that's not compatible?

PR #2304, opened 2026-05-17
Show --benchmark_cooldown in benchmark_zlib --help output

nmoinvaz:
## Summary
- Use the custom help printer callback in `benchmark::Initialize` to append the `--benchmark_cooldown` flag to the standard help text.


mtl1979 (approved):
LGTM... The purpose of third parameter is quite self-explanatory even though some common people might not understand syntax of lambda functions. Indent seems to be consistent with existing help text.

PR #2307, opened 2026-05-19
Always check that distance is too far back in inflateBack.

### LINKED ISSUE #2244: OOB read in `inflateBack`
With some very specific input, the `inflateBack` function will read out of bounds. `AddressSanitizer` is able to detect this:

``` 
=================================================================
==163008==ERROR: AddressSanitizer: SEGV on unknown address 0x514ffffffe81 (pc 0x5d288b3a2519 bp 0x7fff7978a470 sp 0x7fff7978a220 T0)
==163008==The signal is caused by a READ memory access.
    #0 0x5d288b3a2519 in inflateBack /tmp/zlib-ng/infback.c:484:30
    #1 0x5d288b39ba08 in main /tmp/infback_oob.c:84:11
    #2 0x764ffa42a1c9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #3 0x764ffa42a28a in __libc_start_main csu/../csu/libc-start.c:360:3
    #4 0x5d288b2c5a74 in _start (/tmp/infback_oob_asan+0xa9a74)
```

The input program and commands used to compile are at https://gist.github.com/folkertdev/c31ef8a3c0388ad734274cf9269edb80. With that input `state->wsize == 512` while `state->offset == 513`, leading to an unsigned subtraction with overflow on line 472:

https://github.com/zlib-ng/zlib-ng/blob/c2712b8a345191f6ed79558c089777df94590087/infback.c#L472-L485

That means the `else` branch is taken, constructing an invalid pointer that is eventually dereferenced.

---

Right above this piece of logic is 

https://github.com/zlib-ng/zlib-ng/blob/c2712b8a345191f6ed79558c089777df94590087/infback.c#L461-L466

with `INFLATE_STRICT` enabled, the UB is never reached. So, it seems to me like this check should just always run, and error on invalid input. To my knowledge `inflateBack` is not really used in the wild, so having it be correct seems more important than the (likely negligible) performance hit of the additional check.

---

This issue was found by @inahga as part of a security review of zlib-rs.

mtl1979:
* Making the distance check conditional did trap AddressSanitizer with 512 byte (9 bits) window. Although 32 kilobyte window is the default, `inflateBackInit_` allows setting smaller window size. With larger window sizes, it is not possible to generate large enough distance code, even manually, to trigger the OOB read.

Fixes #2244

### DIFF
--- a/infback.c
+++ b/infback.c
@@ -460,12 +460,10 @@ int32_t Z_EXPORT PREFIX(inflateBack)(PREFIX3(stream) *strm, in_func in, void *in
                 state->offset += BITS(state->extra);
                 DROPBITS(state->extra);
             }
-#ifdef INFLATE_STRICT
             if (state->offset > state->wsize - (state->whave < state->wsize ? left : 0)) {
                 SET_BAD("invalid distance too far back");
                 break;
             }
-#endif
             TRACE_DISTANCE(state->offset);
 
             /* copy match from window to output */


mtl1979:
@coderabbit review

nmoinvaz:
We need to convert the test case in the GH issue to a gtest unit test.

mtl1979:
> We need to convert the test case in the GH issue to a gtest unit test.

I was waiting for you to do it ;)

Dead2 (approved):

PR #2309, opened 2026-05-20
Reject asymmetric C vs C++ machine flags

nmoinvaz:
Different machine flags in C vs C++ flags produce undefined-reference link failures.

### DIFF
--- a/cmake/check-compiler-flags.cmake
+++ b/cmake/check-compiler-flags.cmake
@@ -1,0 +1,28 @@
+# check-compiler-flags.cmake -- Check/validate compiler flags
+# Licensed under the Zlib license, see LICENSE.md for details
+
+# Extract machine flags from multi-flag command line argument.
+function(extract_machine_flags out_var flags_str)
+    string(REGEX MATCHALL
+        "-(m(arch|cpu|tune|fpu|float-abi|fp16-format|abi)|q(arch|tune)|tp|target)=[^ \t]+|/arch:[^ \t]+"
+        matches "${flags_str}")
+    string(REPLACE "\"" "" matches "${matches}")
+    string(REPLACE "'"  "" matches "${matches}")
+
+    list(SORT matches)
+    set(${out_var} "${matches}" PARENT_SCOPE)
+endfunction()
+
+# Error on asymmetric C vs C++ machine flags.
+function(check_machine_flag_symmetry)
+    extract_machine_flags(m_flag_c   "${CMAKE_C_FLAGS}")
+    extract_machine_flags(m_flag_cxx "${CMAKE_CXX_FLAGS}")
+
+    if(NOT "${m_flag_c}" STREQUAL "${m_flag_cxx}")
+        message(FATAL_ERROR
+            "C and C++ machine flags differ:\n"
+            "    CMAKE_C_FLAGS:   ${m_flag_c}\n"
+            "    CMAKE_CXX_FLAGS: ${m_flag_cxx}\n"
+            "Pass the same machine flags to both, or to neither.")
+    endif()
+endfunction()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,7 +59,9 @@ cmake_dependent_option(WITH_BENCHMARKS "Build test/benchmarks" OFF "BUILD_TESTIN
 cmake_dependent_option(WITH_BENCHMARK_APPS "Build application benchmarks" OFF "BUILD_TESTING" OFF)
 
 if(WITH_GTEST OR WITH_FUZZERS OR WITH_BENCHMARKS)
-  enable_language(CXX)
+    enable_language(CXX)
+    include(cmake/check-compiler-flags.cmake)
+    check_machine_flag_symmetry()
 endif()
 
 include(cmake/detect-arch.cmake)


cmake/check-compiler-flags.cmake
```diff
@@ -0,0 +1,26 @@
...
+# Licensed under the Zlib license, see LICENSE.md for details
+
+# Extract machine flags from multi-flag command line argument.
+function(extract_machine_flags out_var flags_str)
+    string(REGEX MATCHALL "-(march|mcpu|mtune|mfpu)=[^ \t]+" matches "${flags_str}")
+    string(REPLACE "\"" "" matches "${matches}")
+    string(REPLACE "'"  "" matches "${matches}")
+
+    list(SORT matches)
+    set(${out_var} "${matches}" PARENT_SCOPE)
```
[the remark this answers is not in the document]
nmoinvaz: Unrealistic scenario

mtl1979:
Checks like this feels like going to extreme measures to detect user errors... We should implement these only if there is a known case where this kind of mismatch has caused a build failure. Too many checks can result in bad interactions between individual checks and edge cases where one check thinks there is a mismatch, but it has no real-life effect in build outcome.

cmake/check-compiler-flags.cmake
```diff
@@ -0,0 +1,26 @@
+# check-compiler-flags.cmake -- Check/validate compiler flags
+# Licensed under the Zlib license, see LICENSE.md for details
+
+# Extract machine flags from multi-flag command line argument.
+function(extract_machine_flags out_var flags_str)
+    string(REGEX MATCHALL "-(march|mcpu|mtune|mfpu)=[^ \t]+" matches "${flags_str}")
```
mtl1979: This pattern will work with GCC/Clang but likely will not be enough for other toolchains like NVHPC.

nmoinvaz: I can further expand the regex, but there will always be a small gap. At least if it covers 90% of cases I'm happy with that. Too many times I've seen people make this honest mistake.

mtl1979: I only know about the compilers I have recently used, but NVHPC (nvc/nvcc/nvc++) will understand both "-tp px" and "-tp=px", so at least that isn't currently supported here...

I guess we should at least try to support all that CMake/configure scripts currently support in our repository...

nmoinvaz: I've updated the regex.

mtl1979: I think it's better, but trying to handle "-tp" with both space and equal sign isn't going to be easy... Especially if space is used to delimit or separate the individual options. I initially thought that the "space or equal sign" needs to be normalized to just equal sign before trying to compare the C and C++ flags.

configure
```diff
@@ -256,6 +256,35 @@ case "$1" in
...
+
+# Error on asymmetric C vs C++ machine flags.
+check_machine_flag_symmetry()
+{
+    if [ -z "${CFLAGS-}" ] && [ -z "${CXXFLAGS-}" ]; then
+        return
+    fi
+    m_flag_c=$(extract_machine_flags "${CFLAGS-}")
+    m_flag_cxx=$(extract_machine_flags "${CXXFLAGS-}")
+    if [ "${m_flag_c}" != "${m_flag_cxx}" ]; then
```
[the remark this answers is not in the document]
mtl1979: `configure` doesn't propagate `CXXFLAGS` at all currently, so even trying to use C++ compiler in itself is an error... Whole check should be removed from `configure` and only keep for CMake builds.

configure
```diff
@@ -256,6 +256,35 @@ case "$1" in
...
+            echo "C and C++ machine flags differ:"
+            echo "    CFLAGS:   ${m_flag_c}"
+            echo "    CXXFLAGS: ${m_flag_cxx}"
+            echo "Pass the same machine flags to both, or to neither."
+        } | tee -a configure.log >&2
+        leave 1
+    fi
+}
+check_machine_flag_symmetry
+
```
mtl1979: Ineffective code as `configure` only use C compiler and doesn't try to build any C++ code.

mtl1979 (changes requested):

configure
nmoinvaz: Great point.

cmake/check-compiler-flags.cmake
```diff
@@ -0,0 +1,28 @@
...
+    list(SORT matches)
+    set(${out_var} "${matches}" PARENT_SCOPE)
+endfunction()
+
+# Error on asymmetric C vs C++ machine flags.
+function(check_machine_flag_symmetry)
+    extract_machine_flags(m_flag_c   "${CMAKE_C_FLAGS}")
+    extract_machine_flags(m_flag_cxx "${CMAKE_CXX_FLAGS}")
+
+    if(NOT "${m_flag_c}" STREQUAL "${m_flag_cxx}")
```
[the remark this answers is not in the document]
nmoinvaz: That is possible, but it is not really the scenario we encounter users doing a lot.

cmake/check-compiler-flags.cmake
```diff
@@ -0,0 +1,28 @@
+# check-compiler-flags.cmake -- Check/validate compiler flags
+# Licensed under the Zlib license, see LICENSE.md for details
+
+# Extract machine flags from multi-flag command line argument.
+function(extract_machine_flags out_var flags_str)
+    string(REGEX MATCHALL
+        "-(m(arch|cpu|tune|fpu|float-abi|fp16-format|abi)|q(arch|tune)|tp|target)=[^ \t]+|/arch:[^ \t]+"
```
mtl1979, line 7: This (still) doesn't match "-tp px" style... When the separator isn't "=", we need to check that the next character after space isn't minus sign and if if isn't, then grab everything until next space including a-z, 0-9 and minus sign. This doesn't unnecessarily match if some other compiler understands "-tp" in different way or it is followed by something that looks like filename with extension.

Currently these are valid values for "-tp":

```
    px              Generic Linux/Windows x86_64 Processor.
    bulldozer       AMD Bulldozer processor
    piledriver      AMD Piledriver processor
    bdver3          AMD Family 15h core
    bdver4          AMD Family 15h core
    znver1          AMD Zen architecture (Epyc, Ryzen)
    znver2          AMD Zen 2 architecture (Ryzen 2)
    znver3          AMD Zen 3 architecture (Ryzen 3)
    znver4          AMD Zen 4 architecture (Ryzen 4)
    sandybridge     Intel SandyBridge processor
    haswell         Intel Haswell processor
    broadwell       Intel Broadwell processor
    skylake         Intel Skylake Xeon processor
    skylakeclient   Intel Skylake Core processor
    icelake         Intel Ice Lake Xeon processor
    icelake-client  Intel Ice Lake client processor
    cannonlake      Intel Cannon Lake Xeon processor
    cascadelake     Intel Cascade Lake Xeon processor
    cooperlake      Intel Cooper Lake Xeon processor
    tigerlake       Intel Tiger Lake Xeon processor
    alderlake       Intel Alder Lake Xeon processor
    rocketlake      Intel Rocket Lake Xeon processor
    sapphirerapids  Intel Sapphire Rapids Xeon processor
    graniterapids   Intel Granite Rapids Xeon processor
    x86-64-v2       x86-64 microarchitectural level including SSE
    x86-64-v3       x86-64 microarchitectural level including AVX2
    x86-64-v4       x86-64 microarchitectural level including some AVX512 variants
    host            Link native version of HPC SDK cpu math library
    native          Alias for -tp host
```

nmoinvaz: It will should match the -tp= style though. I said from the beginning it wouldn't cover every case from every compiler. Just maybe 90%. I don't plan to make this any more complex than it already is.

mtl1979: > It will should match the -tp= style though. I said from the beginning it wouldn't cover every case from every compiler. Just maybe 90%. I don't plan to make this any more complex than it already is.

Just saying because even the help text and support forum suggest the version with space instead of equal sign is preferred style. In other apps, using space is more common with short option names. GCC/Clang is the oddball here as it uses long names with only single minus sign, when long names should use two and short names one.

Dead2 (approved):
LGTM

PR #2310, opened 2026-05-20
[CI] Add NVHPC to CMake workflow.

mtl1979:
* Fix error with invalid inline assembly
* Don't cache NVHPC package as it is over 4 GB in size

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -500,6 +500,14 @@ jobs:
             msan-options: use_sigaltstack=0
             # Coverage disabled for clang-20, errors
 
+          - name: Ubuntu NVC Native
+            os: ubuntu-latest
+            compiler: nvc
+            cxx-compiler: nvc++
+            cmake-args: -G Ninja -DWITH_NATIVE_INSTRUCTIONS=ON
+            packages: ninja-build nvhpc-26-3
+            # Coverage disabled for NVHPC
+
           - name: Ubuntu Emscripten WASM32
             os: ubuntu-latest
             chost: wasm32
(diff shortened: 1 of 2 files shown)


.github/workflows/cmake.yml
```diff
@@ -500,6 +500,15 @@ jobs:
             msan-options: use_sigaltstack=0
             # Coverage disabled for clang-20, errors
 
+          - name: Ubuntu NVC Native
+            os: ubuntu-latest
+            cache-key: nvc
+            compiler: /opt/nvidia/hpc_sdk/Linux_x86_64/2026/compilers/bin/nvc
```
nmoinvaz: You can set the `PATH` variable in the download step. 
```yaml
echo "/opt/nvidia/hpc_sdk/Linux_x86_64/2026/compilers/bin/" >> $GITHUB_PATH`
```

.github/workflows/cmake.yml
```diff
@@ -500,6 +500,15 @@ jobs:
...
+            cxx-compiler: /opt/nvidia/hpc_sdk/Linux_x86_64/2026/compilers/bin/nvc++
```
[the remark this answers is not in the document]
mtl1979: This is intentional, so minor version updates don't break the compiler path... `/opt/nvidia/hpc_sdk/Linux_x86_64/2026` is symlink to the current version. NVHPC allows installing multiple major versions of the toolchain side-by-side.

.github/workflows/cmake.yml
mtl1979: I know... I just didn't see it necessary as multiple versions of NVHPC can be present in build system.

nmoinvaz: I mean multiple versions of many apps can be installed on the same machine. At least it makes it a bit cleaner in the matrix.

mtl1979: Already changed it ;)

nmoinvaz (approved):

PR #2311, opened 2026-05-22
Fix warnings triggered by NVHPC/nvc.

mtl1979:
* Fix jump with `goto` past variable declaration
* Fix variable initialized with itself / use before initialization
* Mark unreachable code with `Z_UNREACHABLE()` inside macro definition

Depends on: #2306, #2310


Dead2 (approved):
LGTM

PR #2317, opened 2026-06-02
Fix inflateBack() out-of-bounds read past the caller-supplied window (#2316)

### LINKED ISSUE #2316: [Fuzzing] Out-of-bounds read: inflateBack() reads past caller-supplied window buffer
## Summary

`zng_inflateBack()` reads past the end of the caller-supplied window buffer, an out-of-bounds **read** on a contract-correct allocation. Discovered via fuzzing (libFuzzer + ASan/UBSan) of the native `zng_` API (`ZLIB_COMPAT=OFF`) on `develop` @ `88d89eb`.

- **Severity:** High
- **Type:** Memory safety — heap/window out-of-bounds READ
- **Affected function:** `zng_inflateBack()`
- **Affected files:** `infback.c:57-58` (window sizing), `infback.c:371` (call into `inflate_fast`); fault surfaces in `arch/x86/chunkset_avx512.c` via `inffast_tpl.h` / `chunkset_tpl.h`

## Root cause

`inflateBack()` decodes back-references with the same chunked SIMD `inflate_fast` loop used by `inflate()`. That loop reads the back-reference *source* out of the sliding window a full SIMD chunk at a time (up to 64 B on AVX-512) via `loadchunk` / `CHUNKCOPY`, even when only a few bytes near the window's end are logically needed.

- Regular `inflate()` **over-allocates** its window precisely for this: `inflate.c:158` allocates `(1 << MAX_WBITS) + 64` ("64B padding for chunksize"), documented as existing "in order to safely read the last < chunksize bytes of the window."
- `inflateBack()` does **not** get that padding. `infback.c:57-58` sets `state->wsize = state->wbufsize = 1U << windowBits` over the **caller-supplied, exact-size** buffer, then `infback.c:371` runs `inflate_fast` on it.
- The existing safe-copy commits (`c79998f` / `887dce6`) and `CHUNKCOPY_SAFE`'s `safe` bound protect only the **write/output** side of the shared allocation. The window-**source** read (`from = window + (wsize - op)`) is **not** bounded, so a back-reference whose source lands in the last `< chunksize` bytes of the window over-reads past the buffer end.

This is a genuine library bug, not harness misuse: the public contract documents the `inflateBack` window as **exactly** `2**windowBits` bytes (`zlib-ng.h.in:1066-1070`, `infback.c:29-30`, identical to upstream zlib), and the fuzz harness honors it with `window = malloc(1 << windowBits)`, as any real caller (e.g. zlib's own gzip example) would.

This is **new** — the read-side over-read is not covered by the prior write-side safe-copy fixes, and the recent inflateBack commit `da22434` (distance-too-far) is unrelated.

## Reproduced stack trace

```
AddressSanitizer: heap-buffer-overflow READ of size 1
  #0 GET_HALFCHUNK_MAG      arch/x86/chunkset_avx512.c:150
  #1 chunkmemset_avx512     chunkset_tpl.h:161
  #2 inflate_fast_avx512    inffast_tpl.h:259
  #3 zng_inflateBack        infback.c:371
  #4 LLVMFuzzerTestOneInput fuzz/fuzzer_inflate_back.c:139
0x... is located 0 bytes after 32768-byte region   <- exact 2**15 window over-read
```

(ASan labels the fault SEGV / heap-buffer-overflow / heap-use-after-free interchangeably depending only on what occupies the heap immediately after the exact-size window — all one defect.)

## Reproduction

The harness allocates the `inflateBack` window as `malloc(1 << windowBits)` (exact size, per the documented contract), then drives `inflateBackInit` / `inflateBack` / `inflateBackEnd`. Input protocol: `byte[0]` = windowBits (8..15), `byte[1]` = `in()` chunk size, `byte[2]` = force `out()` abort, remaining bytes = raw-deflate stream.

```bash
./fuzz/fuzzer_inflate_back crash-11c5a5c0e954a42fb0ae84771200877129eec77b
```

The crash reproduces on a contract-correct, exact-size `2**windowBits` window. 40 distinct crash inputs were collected (16 + 24 across two fuzzing rounds), all collapsing to this single root cause. Smallest reproducers are 83–94 bytes.

## Exploitability

Attacker-controlled raw-deflate stream → out-of-bounds **read** on a contract-correct buffer → DoS (SEGV) and potential disclosure of adjacent heap contents into the decompressed output. It is a read, not a write, so **not directly RCE**.

**Caveat:** reproduced via the AVX-512 chunkset selected by runtime dispatch on the test host. The defect is in the inflateBack/chunked-window contract, not AVX-512 specifically — other SIMD chunksets read the window the same way and are expected to over-read too, but this was not directly reproduced (no CPU-feature override was exercised).

## Suggested fix

Bound the window-**source** read in `inflateBack`, by either:

1. Allocating/requiring chunk padding for the `inflateBack` window so `wbufsize` reflects the padded size (mirroring `inflate.c:158`), or
2. Forcing `inflateBack` to use the fully `*_safe` copy variants that bound the window-source read as well as the output write.

---

*Found via fuzzing with libFuzzer + AddressSanitizer + UndefinedBehaviorSanitizer on `develop` @ `88d89eb`, native `zng_` API (`ZLIB_COMPAT=OFF`), clang 23.0.0.*

parasol-aser:
## Summary

Fixes #2316 — an out-of-bounds **read** past the end of the caller-supplied `inflateBack()` window.

`inflateBack()` decodes back-references with the same chunked SIMD `inflate_fast` loop as `inflate()`. That loop reads the back-reference **source** out of the sliding window a full SIMD chunk at a time (up to 32 B on AVX-512), even when only a few logical bytes near the window's end are needed. `inflate()` over-allocates its window with 64 B of chunk padding to absorb this (`inflate.c`: `(1 << MAX_WBITS) + 64`), but `inflateBack()` uses the **exact-size** `2**windowBits`, caller-supplied window with **no** padding (`infback.c`). A back-reference whose source lands in the last `< chunksize` bytes therefore over-reads past the buffer end:

```
AddressSanitizer: heap-buffer-overflow READ of size 1
  #0 GET_HALFCHUNK_MAG   arch/x86/chunkset_avx512.c
  #1 chunkmemset_avx512  chunkset_tpl.h
  #2 inflate_fast_avx512 inffast_tpl.h   (CHUNKCOPY_SAFE from window source)
  #3 zng_inflateBack     infback.c
0x... is located 0 bytes after 32768-byte region   ← exact 2**15 window over-read
```

This is a contract-correct allocation: the public API documents the `inflateBack` window as **exactly** `2**windowBits` bytes (identical to upstream zlib, and used that way by zlib's own gzip example), so this is a genuine library defect rather than caller misuse.

### Why only AVX-512 faulted

This directly answers the maintainer question on the issue ("not sure how this wasn't detected by fuzzers when using AVX2"):

- **Non-masked builds (e.g. AVX2):** `CHUNKCOPY_SAFE` has a `#ifndef HAVE_MASKED_READWRITE` byte-wise fallback that triggers when the source distance is `< sizeof(chunk_t)`. For `inflateBack`, output *is* the window, so that guard coincidentally also bounds the **source** read. These paths did not over-read.
- **Masked-readwrite builds (AVX-512):** the masked arms unconditionally call the SIMD `CHUNKCOPY_SAFE` → `CHUNKMEMSET`, which builds the repeat "magazine" with over-wide/unmasked loads from `from` and **skips** the byte-wise source guard (it assumes the buffer is padded). Hence only AVX-512 over-read the exact-size window.

## The fix

This takes the issue's **option 2** (force source-bounded copies), which also addresses the maintainer's point that the real requirement is to *"detect if the read spans past end of allocated buffer"* in a way that does not depend on AVX-512 masked reads:

- **`inffast_tpl.h`** — derive `extra_safe = safe_mode && (wbufsize < wsize + CHUNKSIZE())`, i.e. the window has no chunk padding. This is true **only** for `inflateBack` (`wbufsize == wsize`) and never for regular `inflate` (whose window is padded by ≥ 64 B). When set, route all three `safe_mode` window/output back-reference copy sites through the scalar, source-bounded `chunkcopy_safe` — on **every** build, masked or not. The hot `inflate()` fast path (including its AVX-512 safe-mode tail) is byte-for-byte unchanged.
- **`infback.c`** — harden the slow-path `MATCH` bailout copy: replace the SIMD `chunkmemset_safe` (which over-reads the source on masked builds) with the source-bounded `chunkcopy_safe` (bounded by `put + left == window + wsize`), and reject a too-far-back distance on the MATCH bailout path before the copy (the bailout stores the distance before it is validated).
- **`test/`** — add `test_inflate_back_window_oob.cc` (wired into `test/CMakeLists.txt`) as a permanent regression guard.

### Why not option 1 (pad the window)

The window is caller-supplied and the public contract mandates exactly `2**windowBits` bytes. Requiring callers to over-allocate would break the ABI, zlib-compat, and the documented gzip example. Allocating an internal padded window and redirecting `out()` to it is a larger redesign that wastes memory and diverges from upstream. Rejected as primary; kept only as a considered alternative.

### Performance

`extra_safe` is true only for `inflateBack` and never for `inflate`, so the performance-critical `inflate()` path is untouched. `inflateBack` trades SIMD window copies for byte-wise copies — correctness over speed is the right call for an exact caller buffer, and `inflateBack` is far less perf-critical than `inflate`.

## Test plan

New `test/test_inflate_back_window_oob.cc` decodes valid raw-deflate streams through `inflateBack()` over an **exact-size** `malloc(1 << windowBits)` window — so any over-read lands in an ASan red zone — covering:

- Happy-path boundary round-trips across the full `windowBits` 8..15 range (back-reference sources swept across the window boundary so the window portion `op` of a match takes every small value `1..CHUNKSIZE-1`), including tiny-input-chunk re-entry.
- Targeted match shapes: `dist == 1` RLE, small power-of-two distances 2/4/8/16 (magazine builders), and near-max distance right after the window fills.
- The non-fast `MATCH` bailout path (long-match data) and the `out()`-abort path.
- Direct replay of committed crash reproducers from #2316.

Verified locally (host is AVX-512, the affected dispatch):

```
cmake build-asan   # Debug, WITH_SANITIZER=Address, WITH_GTEST=ON, ZLIB_COMPAT=OFF
cmake --build build-asan --target gtest_zlib -j

gtest_zlib --gtest_filter='*inflate_back_oob*'   → 26 tests, all PASS (no ASan report)
gtest_zlib --gtest_filter='*inflate*:*deflate*'  → 207 tests, all PASS
```

Before the fix the new tests (and the original fuzz reproducers) trip an ASan `heap-buffer-overflow READ` on the AVX-512 path; after the fix they return cleanly and the decoded output matches byte-for-byte. The existing `inflateBack` round-trip/`infcover` tests continue to pass, confirming no regression.

### DIFF
--- a/infback.c
+++ b/infback.c
@@ -487,12 +487,9 @@ int32_t Z_EXPORT PREFIX(inflateBack)(PREFIX3(stream) *strm, in_func in, void *in
             break;
 
         case MATCH:
-            /* Copy back-reference that inflate_fast() could not complete due to
-               insufficient output space. state->length and state->offset were set
-               by the safe_mode MATCH bailout in inflate_fast(), which stores the
-               distance before it is validated. Reject a distance that is too far
-               back here (mirrors the LEN path check above and inflate.c's MATCH
-               state) so the copy below cannot read before the start of the window. */
+            /* Reject too-far-back distances stored by inflate_fast()'s MATCH
+               bailout before validation, so the copy can't read before the
+               window (mirrors the LEN path and inflate.c's MATCH state). */
             if (state->offset > state->wsize - (state->whave < state->wsize ? left : 0)) {
                 SET_BAD("invalid distance too far back");
                 break;
@@ -506,12 +503,8 @@ int32_t Z_EXPORT PREFIX(inflateBack)(PREFIX3(stream) *strm, in_func in, void *in
                     copy = MIN(copy, state->length);
                     put = chunkcopy_safe(put, from, copy, put + left);
                 } else {
-                    /* The window is the caller-supplied, exact-size buffer with no
-                       chunk padding, so use the scalar, source-bounded copy rather
-                       than the SIMD chunkmemset_safe (which over-reads the source on
-                       masked-readwrite builds). put + left == window + wsize bounds it. */
                     copy = MIN(state->length, left);
-                    put = chunkcopy_safe(put, put - state->offset, copy, put + left);
+                    put = FUNCTABLE_CALL(chunkmemset_safe)(put, put - state->offset, copy, left);
                 }
                 state->length -= copy;
                 left -= copy;
--- a/infback.c
+++ b/infback.c
@@ -489,7 +489,14 @@ int32_t Z_EXPORT PREFIX(inflateBack)(PREFIX3(stream) *strm, in_func in, void *in
         case MATCH:
             /* Copy back-reference that inflate_fast() could not complete due to
                insufficient output space. state->length and state->offset were set
-               by the safe_mode MATCH bailout in inflate_fast(). */
+               by the safe_mode MATCH bailout in inflate_fast(), which stores the
+               distance before it is validated. Reject a distance that is too far
+               back here (mirrors the LEN path check above and inflate.c's MATCH
+               state) so the copy below cannot read before the start of the window. */
+            if (state->offset > state->wsize - (state->whave < state->wsize ? left : 0)) {
+                SET_BAD("invalid distance too far back");
+                break;
+            }
             do {
                 ROOM();
                 copy = state->wsize - state->offset;
@@ -499,8 +506,12 @@ int32_t Z_EXPORT PREFIX(inflateBack)(PREFIX3(stream) *strm, in_func in, void *in
                     copy = MIN(copy, state->length);
                     put = chunkcopy_safe(put, from, copy, put + left);
                 } else {
+                    /* The window is the caller-supplied, exact-size buffer with no
+                       chunk padding, so use the scalar, source-bounded copy rather
+                       than the SIMD chunkmemset_safe (which over-reads the source on
+                       masked-readwrite builds). put + left == window + wsize bounds it. */
                     copy = MIN(state->length, left);
-                    put = FUNCTABLE_CALL(chunkmemset_safe)(put, put - state->offset, copy, left);
+                    put = chunkcopy_safe(put, put - state->offset, copy, put + left);
                 }
                 state->length -= copy;
                 left -= copy;
--- a/test/test_inflate_back_window_oob.cc
+++ b/test/test_inflate_back_window_oob.cc
@@ -1,0 +1,450 @@
+/* test_inflate_back_window_oob.cc - Regression tests for issue #2316:
+ *   inflateBack() out-of-bounds READ past the end of the caller-supplied,
+ *   exact-size (2**windowBits, no padding) window.
+ *
+ * These tests are written purely against the behavioural contract:
+ *
+ *   - inflateBackInit() takes a window of EXACTLY (1 << windowBits) bytes and
+ *     uses it verbatim (no internal over-allocation / padding).  This is the
+ *     condition under which the SIMD chunked window-source copy could over-read
+ *     past the window's end.
+ *   - inflateBack() must decode valid raw-deflate streams correctly, and must
+ *     never read past window+ (1<<windowBits), regardless of how back-references
+ *     straddle the window boundary, what the match distance is (including 1 and
+ *     small powers of two), or how the output/MATCH bailout path is exercised.
+ *
+ * The oracle for the "no over-read" property is ASan/UBSan: the window is sized
+ * EXACTLY to the contract so any over-read lands in a red zone.  The oracle for
+ * "still correct" is byte-for-byte comparison against the original input.
+ *
+ * NOTE: the window is deliberately malloc()'d at exactly (1 << windowBits).  Do
+ * not "round up" or share a max-size buffer -- that exact sizing is the whole
+ * point; a padded/oversized window would hide the very bug under test.
+ */
+
+#include "zbuild.h"
+#ifdef ZLIB_COMPAT
+#  include "zlib.h"
+#else
+#  include "zlib-ng.h"
+#endif
+
+#include <stdio.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <vector>
+
+#include <gtest/gtest.h>
+
+/* ------------------------------------------------------------------ */
+/* inflateBack() in()/out() callback plumbing                          */
+/* ------------------------------------------------------------------ */
+
+struct in_ctx {
+    const uint8_t *data;
+    size_t len;
+    size_t pos;
+    size_t chunk;       /* 0 => hand everything over in one shot */
+};
+
+/* Mirrors the fuzz harness's in(): a 0-length return at exhaustion is the
+ * normal end-of-input / truncation signal. */
+static z_uint32_t in_cb(void *desc, z_const unsigned char **buf) {
+    in_ctx *st = (in_ctx *)desc;
+    if (st->pos >= st->len) {
+        *buf = (z_const unsigned char *)(st->data + st->pos);
+        return 0;
+    }
+    size_t remaining = st->len - st->pos;
+    size_t n = st->chunk ? (remaining < st->chunk ? remaining : st->chunk)
+                         : remaining;
+    *buf = (z_const unsigned char *)(st->data + st->pos);
+    st->pos += n;
+    return (z_uint32_t)n;
+}
+
+struct out_ctx {
+    uint8_t *buf;       /* capture target, or nullptr to discard */
+    size_t cap;
+    size_t produced;
+    int abort_after;    /* if >0, return non-zero once produced >= abort_after */
+    int overflowed;     /* set if produced would exceed cap */
+};
+
+static z_int32_t out_cb(void *desc, unsigned char *buf, z_uint32_t len) {
+    out_ctx *st = (out_ctx *)desc;
+    if (st->buf) {
+        if (st->produced + len > st->cap) {
+            st->overflowed = 1;
+            return 1;
+        }
+        memcpy(st->buf + st->produced, buf, len);
+    }
+    st->produced += len;
+    if (st->abort_after && st->produced >= (size_t)st->abort_after)
+        return 1;       /* force the out() abort / MATCH-bailout path */
+    return 0;
+}
+
+/* Decode `comp` with inflateBack() over an EXACT-size window.  Captures output
+ * into `out` (may be nullptr) and returns the inflateBack() status. */
+static int run_inflate_back(const uint8_t *comp, size_t comp_len, int wbits,
+                            size_t in_chunk, int abort_after,
+                            uint8_t *out, size_t out_cap, size_t *produced) {
+    PREFIX3(stream) strm;
+    memset(&strm, 0, sizeof(strm));
+
+    /* EXACT-size, caller-supplied, unpadded window -- the bug's oracle. */
+    size_t wsize = (size_t)1 << wbits;
+    uint8_t *window = (uint8_t *)malloc(wsize);
+    if (window == nullptr)
+        return Z_MEM_ERROR;
+
+    int err = PREFIX(inflateBackInit)(&strm, wbits, window);
+    if (err != Z_OK) {
+        free(window);
+        return err;
+    }
+
+    in_ctx ist = { comp, comp_len, 0, in_chunk };
+    out_ctx ost = { out, out_cap, 0, abort_after, 0 };
+
+    strm.next_in = nullptr;
+    strm.avail_in = 0;
+
+    err = PREFIX(inflateBack)(&strm, in_cb, &ist, out_cb, &ost);
+
+    PREFIX(inflateBackEnd)(&strm);
+    free(window);
+
+    if (produced)
+        *produced = ost.produced;
+    return err;
+}
+
+/* Raw-deflate compress `src` (no zlib/gzip wrapper) at `wbits`.
+ *
+ * deflate() does not support a raw 8-bit window (it requires >= 9), so the
+ * compressor window is clamped up to 9.  Callers that inflateBack at
+ * windowBits==8 must therefore use data whose back-references are all at
+ * distance < 256 (e.g. constant RLE runs, which only ever match at distance 1)
+ * so the produced stream is still decodable in a 256-byte window.  For
+ * windowBits >= 9 the compressor and inflate windows match, so deflate can
+ * never emit a distance exceeding the inflateBack window. */
+static std::vector<uint8_t> compress_raw(const uint8_t *src, size_t n, int wbits,
+                                         int level = Z_BEST_COMPRESSION) {
+    int cwbits = wbits < 9 ? 9 : wbits;
+    PREFIX3(stream) c;
+    memset(&c, 0, sizeof(c));
+    int err = PREFIX(deflateInit2)(&c, level, Z_DEFLATED, -cwbits,
+                                   MAX_MEM_LEVEL, Z_DEFAULT_STRATEGY);
+    EXPECT_EQ(err, Z_OK);
+
+    z_uintmax_t bound = PREFIX(deflateBound)(&c, (z_uintmax_t)n);
+    std::vector<uint8_t> out(bound ? bound : 1);
+
+    c.next_in = (z_const unsigned char *)src;
+    c.avail_in = (uint32_t)n;
+    c.next_out = out.data();
+    c.avail_out = (uint32_t)out.size();
+
+    err = PREFIX(deflate)(&c, Z_FINISH);
+    EXPECT_EQ(err, Z_STREAM_END);
+    out.resize(c.total_out);
+    PREFIX(deflateEnd)(&c);
+    return out;
+}
+
+/* ================================================================== */
+/* A. Happy-path round trips across the full windowBits range          */
+/* ================================================================== */
+
+/* Build data that forces many back-references whose source straddles the
+ * inflate window boundary.  The single-byte "rotation" emitted each cycle sweeps
+ * the alignment of the window-sized copy relative to inflateBack's periodic
+ * window flush, so over a long stream the window portion `op` of a match takes
+ * every small value (1..CHUNKSIZE-1) -- exactly the over-read trigger.  Small
+ * fixed-distance runs additionally exercise the dist==1 / 2 / 4 / 8 / 16
+ * chunkmemset "magazine" copy paths. */
+static std::vector<uint8_t> gen_boundary_data(int wbits, int cycles) {
+    uint32_t wsize = 1u << wbits;
+    std::vector<uint8_t> b;
+    b.reserve((size_t)wsize * (cycles + 2));
+
+    uint32_t lcg = 0x9e3779b9u ^ (uint32_t)wbits;
+    auto rnd = [&]() -> uint8_t {
+        lcg = lcg * 1103515245u + 12345u;
+        return (uint8_t)(lcg >> 16);
+    };
+
+    /* Seed block: a full window of varied bytes (mostly literals). */
+    for (uint32_t i = 0; i < wsize; i++)
+        b.push_back(rnd());
+
+    static const uint32_t small_d[5] = { 1, 2, 4, 8, 16 };
+
+    for (int c = 0; c < cycles; c++) {
+        /* Rotate alignment by a few fresh literals each cycle. */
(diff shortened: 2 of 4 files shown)


test/test_inflate_back_window_oob.cc
```diff
@@ -0,0 +1,450 @@
...
+static void replay_fuzz_input(const uint8_t *data, size_t size) {
+    if (size < 4)
+        return;
+    int wbits = 8 + (data[0] % 8);
+    size_t chunk = (size_t)data[1] + 1;
+    int abort_out = data[2] & 1;
+    /* Just decode -- the assertion is "no ASan report / clean return". */
+    int err = run_inflate_back(data + 3, size - 3, wbits, chunk,
+                               abort_out ? 1 : 0, nullptr, 0, nullptr);
+    (void)err;  /* any status is acceptable; not crashing is the point. */
```
mtl1979: For consistency, should use `Z_UNUSED(err);`.

test/test_inflate_back_window_oob.cc
```diff
@@ -0,0 +1,450 @@
...
+ *   exact-size (2**windowBits, no padding) window.
+ *
+ * These tests are written purely against the behavioural contract:
+ *
+ *   - inflateBackInit() takes a window of EXACTLY (1 << windowBits) bytes and
+ *     uses it verbatim (no internal over-allocation / padding).  This is the
+ *     condition under which the SIMD chunked window-source copy could over-read
+ *     past the window's end.
+ *   - inflateBack() must decode valid raw-deflate streams correctly, and must
+ *     never read past window+ (1<<windowBits), regardless of how back-references
```
mtl1979: Inconsistent white-space... Either use `window + (1<<windowBits)` or `window+(1<<windowBits)`.

test/test_inflate_back_window_oob.cc
```diff
@@ -0,0 +1,450 @@
...
+    std::vector<uint8_t> got(1 << wbits);
+    size_t produced = 0;
+    int err = run_inflate_back(junk, sizeof(junk), wbits, 0, 0,
+                               got.data(), got.size(), &produced);
+    EXPECT_NE(err, Z_STREAM_END);
+}
+
+/* ================================================================== */
+/* D. Direct replay of committed crash reproducers (#2316 corpus)      */
+/* ================================================================== */
```
mtl1979, line 389: NITPICK: Inconsistent comment length

test/test_inflate_back_window_oob.cc
```diff
@@ -0,0 +1,450 @@
...
+        std::vector<uint8_t> orig(n);
+        for (size_t i = 0; i < n; i++)
+            orig[i] = (uint8_t)(i % 5);          /* len=258 dist=5 tokens */
+        expect_roundtrip(orig, wbits, /*in_chunk*/ 7);
+    }
+}
+
+/* ================================================================== */
+/* C. Error / abort paths must stay clean (no crash, sane status)      */
+/* ================================================================== */
```
mtl1979, line 339: NITPICK: Inconsistent comment length.

test/test_inflate_back_window_oob.cc
```diff
@@ -0,0 +1,450 @@
...
+/* Cover the full documented windowBits domain 8..15, with the two boundary
+ * values from the plan (8 -> 256B window, 15 -> 32KiB).  For windowBits 8 the
+ * compressor clamps to a 9-bit window whose MAX_DIST (512-262=250) still fits a
+ * 256-byte inflateBack window, so the stream remains decodable there. */
+INSTANTIATE_TEST_SUITE_P(all_window_bits, inflate_back_oob,
+                         ::testing::Values(8, 9, 10, 11, 12, 13, 14, 15));
+
+/* ================================================================== */
+/* B. Targeted distance / match-shape edge cases                       */
+/* ================================================================== */
```
mtl1979, line 257: NITPICK: Inconsistent comment length.

test/test_inflate_back_window_oob.cc
```diff
@@ -0,0 +1,450 @@
...
+    err = PREFIX(deflate)(&c, Z_FINISH);
+    EXPECT_EQ(err, Z_STREAM_END);
+    out.resize(c.total_out);
+    PREFIX(deflateEnd)(&c);
+    return out;
+}
+
+/* ================================================================== */
+/* A. Happy-path round trips across the full windowBits range          */
+/* ================================================================== */
```
mtl1979, line 162: NITPICK: Inconsistent comment length.

test/test_inflate_back_window_oob.cc
```diff
@@ -0,0 +1,450 @@
...
+#include <stdlib.h>
+#include <string.h>
+
+#include <vector>
+
+#include <gtest/gtest.h>
+
+/* ------------------------------------------------------------------ */
+/* inflateBack() in()/out() callback plumbing                          */
+/* ------------------------------------------------------------------ */
```
mtl1979, line 43: NITPICK: Inconsistent comment length.

mtl1979 (changes requested):
Just some style cleanup needed... Someone with AVX512-capable processor needs to verify the code does what it says...

parasol-aser:
Thanks for the review — pushed 1c35fd6 addressing the feedback:

**@coderabbitai (wrap-around tail copy):** Good catch. The earlier `CHUNKCOPY_SAFE(out, from, op, safe)` at the "some from end of window" segment was the one remaining window-source copy not gated on `extra_safe`; when the wrapped tail is < CHUNKSIZE() it could still over-read past `window + wsize` on masked builds. It now routes through the source-bounded scalar `chunkcopy_safe` under `extra_safe`, matching the other three sites.

**@mtl1979 (style):** Aligned the banner comment closing `*/` markers (they were one column off), switched `(void)err;` to `Z_UNUSED(err);`, and normalized the `window + (1<<windowBits)` spacing.

All 27 regression tests still pass under ASan/UBSan.

KungFuJesus:
So the AVX512 path never assumed padding, it assumed it was able to accommodate a safe read distance by constructing a mask from the offset of the read that never extended beyond the bounds of a safe read.  It'd be preferable if the solution involved fixing it that way rather than calling the scalar chunkcopy_safe.

mtl1979:
> So the AVX512 path never assumed padding, it assumed it was able to accommodate a safe read distance by constructing a mask from the offset of the read that never extended beyond the bounds of a safe read. It'd be preferable if the solution involved fixing it that way rather than calling the scalar chunkcopy_safe.

If you have code suggestion, just post a better patch... As I said, I don't have AVX512-capable processor, so I can't test any changes or comment on if some change is better than other, but others who have "better" processor can...

nmoinvaz:
@parasol-aser what AI bot are you?

nmoinvaz (changes requested):
The code comments need to be more terse.

Why wouldn't you just increase the size for the window allocation in inflateBack like you mentioned in your problem report?

Why do you need extra_safe stuff back? We already have a safe_mode variable that is always 1 when called from inflateBack.

mtl1979:
> @parasol-aser what AI bot are you?

Claude, just like you ;)

KungFuJesus:
> > So the AVX512 path never assumed padding, it assumed it was able to accommodate a safe read distance by constructing a mask from the offset of the read that never extended beyond the bounds of a safe read. It'd be preferable if the solution involved fixing it that way rather than calling the scalar chunkcopy_safe.
> 
> If you have code suggestion, just post a better patch... As I said, I don't have AVX512-capable processor, so I can't test any changes or comment on if some change is better than other, but others who have "better" processor can...

Fair enough, I'll tackle it this weekend. I have a rough idea of what's going on, this is likely able to be handled with a simple carefully placed mask read.

mtl1979:
> > > So the AVX512 path never assumed padding, it assumed it was able to accommodate a safe read distance by constructing a mask from the offset of the read that never extended beyond the bounds of a safe read. It'd be preferable if the solution involved fixing it that way rather than calling the scalar chunkcopy_safe.
> > 
> > 
> > If you have code suggestion, just post a better patch... As I said, I don't have AVX512-capable processor, so I can't test any changes or comment on if some change is better than other, but others who have "better" processor can...
> 
> Fair enough, I'll tackle it this weekend. I have a rough idea of what's going on, this is likely able to be handled with a simple carefully placed mask read.

No rush at all... Faster code is always better when it can still fix the underlying issue. AI bots often try suggesting the easy "safe" solution, which isn't always the best.

test/test_inflate_back_window_oob.cc
```diff
@@ -0,0 +1,450 @@
...
+static std::vector<uint8_t> compress_raw(const uint8_t *src, size_t n, int wbits,
+                                         int level = Z_BEST_COMPRESSION) {
+    int cwbits = wbits < 9 ? 9 : wbits;
+    PREFIX3(stream) c;
+    memset(&c, 0, sizeof(c));
+    int err = PREFIX(deflateInit2)(&c, level, Z_DEFLATED, -cwbits,
+                                   MAX_MEM_LEVEL, Z_DEFAULT_STRATEGY);
+    EXPECT_EQ(err, Z_OK);
+
+    z_uintmax_t bound = PREFIX(deflateBound)(&c, (z_uintmax_t)n);
```
mtl1979, line 145: Second parameter should be `unsigned long`.

test/test_inflate_back_window_oob.cc
```diff
@@ -0,0 +1,450 @@
...
+ * shouldn't.  A distance-too-far block in particular must be caught before any
+ * window-source read. */
+TEST(inflate_back_oob_errors, invalid_stream_rejected) {
+    int wbits = 10;
+    /* A stored-block header claiming a length but no payload, plus noise. */
+    const uint8_t junk[] = {
+        0x01, 0xff, 0xff, 0x00, 0x00,   /* final stored block, len=0xffff */
+        0xde, 0xad, 0xbe, 0xef, 0x55, 0xaa, 0x12, 0x34
+    };
+    std::vector<uint8_t> got(1 << wbits);
```
mtl1979, line 380: Implicit widening in constructor from 32-bit to 64-bit. This will fail in CI. Shift is 32-bit operation by default.

mtl1979 (changes requested):
Fails CI tests...

test/test_inflate_back_window_oob.cc
```diff
@@ -0,0 +1,450 @@
...
+    ASSERT_EQ(produced, orig.size());
+    EXPECT_EQ(memcmp(got.data(), orig.data(), orig.size()), 0);
+}
+
+/* Cover the full documented windowBits domain 8..15, with the two boundary
+ * values from the plan (8 -> 256B window, 15 -> 32KiB).  For windowBits 8 the
+ * compressor clamps to a 9-bit window whose MAX_DIST (512-262=250) still fits a
+ * 256-byte inflateBack window, so the stream remains decodable there. */
+INSTANTIATE_TEST_SUITE_P(all_window_bits, inflate_back_oob,
+                         ::testing::Values(8, 9, 10, 11, 12, 13, 14, 15));
```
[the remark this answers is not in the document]
mtl1979: This is non-sense, fails to detect it is a template.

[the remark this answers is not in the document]
mtl1979: This is non-sense, fails to detect it is a template.

test/test_inflate_back_window_oob.cc
```diff
@@ -0,0 +1,450 @@
...
+
+/* ================================================================== */
+/* B. Targeted distance / match-shape edge cases                      */
+/* ================================================================== */
+
+/* Decode `orig` via raw deflate + inflateBack at `wbits`, asserting an exact
+ * round trip.  Shared by the small focused cases below. */
+static void expect_roundtrip(const std::vector<uint8_t> &orig, int wbits,
```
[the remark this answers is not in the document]
mtl1979: This is non-sense, can't detect if function is inlined in current translation unit.

KungFuJesus:
@parasol-aser can you provide a fuzzed input here with a unit test that necessitates all the changes you made in inffast?  As far as I can tell, I only hit your asan fault from the unit test in the current head of develop _if_ the bounds check in "MATCH" is removed for "too far back".  

It'd help to have a case where swapping that chunkmemset_safe for chunkcopy_safe was actually necessary.  Then we can go about figuring out those boundaries.  Basically, provided your margin for safety covers both read and write lengths, in the AVX512 variants it should be perfectly safe to call these things.  We want safe _and_ optimal behavior, if possible.  The entire premise of the masked read and writes is we let the masked reads and writes handle sub vector lengths.

parasol-aser:
Pushed d9a8c2f addressing the feedback by **shrinking the fix to its root cause**:

**@KungFuJesus** — you were right. I built at develop HEAD on an AVX512 host (avx512f/bw/dq/vl, `compare256.avx512` active) under ASan/UBSan and confirmed all 27 regression tests (including the 3 fuzz-crash corpus replays) pass with **only** the MATCH `too-far-back` distance guard in `infback.c` — no `inffast_tpl.h` changes and no `chunkmemset_safe` swap. The ASan fault is reachable solely via the unvalidated distance reading *before* the window start; the masked AVX512 reads/writes already keep sub-vector copies in bounds. So:

- **`inffast_tpl.h`: fully reverted to base.** No `extra_safe`/`wbufsize`; the AVX512 fast path stays optimal.
- **`infback.c`: MATCH copy restored to `chunkmemset_safe`**; kept only the distance check.

**@nmoinvaz** — dropped `extra_safe` (you're right, `safe_mode` already covers inflateBack), and made the remaining comment terse. On the window allocation: it's the caller-supplied exact-size buffer (`inflateBackInit(strm, windowBits, window)`), so it can't be over-allocated without an API change — validating the distance is the in-contract fix.

Net diff is now `infback.c` (+6/-2) plus the regression test. ASan/UBSan green: 27/27 OOB tests, 0 failures across the full gtest suite.

nmoinvaz:
CI failures:
```
 D:\a\zlib-ng\zlib-ng\test\test_inflate_back_window_oob.cc(145): warning C4267: 'argument': conversion from 'size_t' to 'unsigned long', possible loss of data [D:\a\zlib-ng\zlib-ng\test\gtest_zlib.vcxproj]
         D:\a\zlib-ng\zlib-ng\test\test_inflate_back_window_oob.cc(380): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?) [D:\a\zlib-ng\zlib-ng\test\gtest_zlib.vcxproj]

```

test/test_inflate_back_window_oob.cc
```diff
@@ -0,0 +1,450 @@
...
+
+    PREFIX(inflateBackEnd)(&strm);
+    free(window);
+
+    if (produced)
+        *produced = ost.produced;
+    return err;
+}
+
+/* Raw-deflate compress `src` (no zlib/gzip wrapper) at `wbits`.
```
nmoinvaz, line 127: Seems like a bit too much info here. Only need one or two lines. Anything code specific can be a comment above the specific line in code.

infback.c
```diff
@@ -487,9 +487,13 @@ int32_t Z_EXPORT PREFIX(inflateBack)(PREFIX3(stream) *strm, in_func in, void *in
             break;
 
         case MATCH:
-            /* Copy back-reference that inflate_fast() could not complete due to
-               insufficient output space. state->length and state->offset were set
-               by the safe_mode MATCH bailout in inflate_fast(). */
+            /* Reject too-far-back distances stored by inflate_fast()'s MATCH
+               bailout before validation, so the copy can't read before the
+               window (mirrors the LEN path and inflate.c's MATCH state). */
```
nmoinvaz, line 492: > (mirrors the LEN path and inflate.c's MATCH state).

Where does it do this in inflate.c? Are you referring to `INFLATE_ALLOW_INVALID_DISTANCE_TOOFAR_ARRR`? Maybe you are looking at an old commit?

mtl1979:
This needs either a rebase to see what still is needed after merging #2370 or to be closed as abandoned...

nmoinvaz:
The core issue in this PR has been fixed in https://github.com/zlib-ng/zlib-ng/pull/2370. I wonder if the additional tests in this PR might be overkill.

Dead2:
Fixed by #2370

PR #2319, opened 2026-06-05
fix integer overflow in gz_buffer_alloc buffer sizing

dxbjavid:
Noticed gz_buffer_alloc reads state->want into an int and then computes want*2, so a large gzbuffer request overflows the int before it ever reaches zng_alloc_aligned. The total allocation is really 3*want for both read and write, but gzbuffer only guards that 2*want fits in an unsigned, so the size can wrap to something far smaller than state->size and the later gz_write/gz_read copies overrun it. I hit it with gzbuffer(file, 0x55555558) under asan. Compute the sizes as size_t and bail out when the total would not fit the unsigned the allocator takes.

### DIFF
--- a/gzlib.c
+++ b/gzlib.c
@@ -69,8 +69,8 @@ static void gz_reset(gz_state *state) {
 
 /* Allocate in/out buffers for gzFile */
 int Z_INTERNAL gz_buffer_alloc(gz_state *state) {
-    size_t want = state->want;
-    size_t in_size = want, out_size = want;
+    unsigned want = state->want;
+    unsigned in_size = want, out_size = want;
 
     if (state->mode == GZ_WRITE) {
         in_size = want * 2; // double input buffer for compression (ref: gzprintf)
@@ -80,9 +80,7 @@ int Z_INTERNAL gz_buffer_alloc(gz_state *state) {
         out_size = want * 2;  // double output buffer for decompression
     }
 
-    /* gzbuffer caps want at GZBUFSIZE_MAX, so the total here stays inside the
-       unsigned that zng_alloc_aligned takes */
-    state->buffers = (unsigned char *)zng_alloc_aligned((unsigned)(in_size + out_size), 64);
+    state->buffers = (unsigned char *)zng_alloc_aligned((in_size + out_size), 64);
     state->in = state->buffers;
     if (out_size) {
         state->out = state->buffers + (in_size); // Outbuffer goes after inbuffer
@@ -95,7 +93,7 @@ int Z_INTERNAL gz_buffer_alloc(gz_state *state) {
         return -1;
     }
 
-    state->size = (unsigned)want; // mark state as initialized
+    state->size = want; // mark state as initialized
     return 0;
 }
 
@@ -338,6 +336,7 @@ z_int32_t Z_EXPORT PREFIX(gzbuffer)(gzFile file, z_uint32_t size) {
 #ifdef ZLIB_COMPAT
         size = GZBUFSIZE_MAX;   /* silently clamp to stay compatible with zlib */
 #else
+        PREFIX(gz_error)(state, Z_MEM_ERROR, "out of memory");
         return -1;              /* too large */
 #endif
     }
--- a/gzlib.c
+++ b/gzlib.c
@@ -106,7 +106,7 @@ int Z_INTERNAL gz_buffer_alloc(gz_state *state) {
         return -1;
     }
 
-    state->size = want; // mark state as initialized
+    state->size = (unsigned)want; // mark state as initialized
     return 0;
 }
 
--- a/gzlib.c
+++ b/gzlib.c
@@ -80,15 +80,8 @@ int Z_INTERNAL gz_buffer_alloc(gz_state *state) {
         out_size = want * 2;  // double output buffer for decompression
     }
 
-    /* the total reaches 3 * want but gzbuffer only guards 2 * want; cap against
-       INT_MAX (minus zng_alloc_aligned's overhead) and reject rather than wrap */
-    const size_t aligned_overhead = sizeof(void *) + 64;
-    const size_t max_size = (size_t)INT_MAX - aligned_overhead;
-    if (out_size > max_size || in_size > max_size - out_size) {
-        PREFIX(gz_error)(state, Z_MEM_ERROR, "out of memory");
-        return -1;
-    }
-
+    /* gzbuffer caps want at GZBUFSIZE_MAX, so the total here stays inside the
+       unsigned that zng_alloc_aligned takes */
     state->buffers = (unsigned char *)zng_alloc_aligned((unsigned)(in_size + out_size), 64);
     state->in = state->buffers;
     if (out_size) {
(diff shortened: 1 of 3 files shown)


gzlib.c
```diff
@@ -80,7 +80,15 @@ int Z_INTERNAL gz_buffer_alloc(gz_state *state) {
...
-    state->buffers = (unsigned char *)zng_alloc_aligned((in_size + out_size), 64);
+    /* the total here reaches 3 * want, but gzbuffer only guarantees 2 * want
+       fits in an unsigned, and zng_alloc_aligned takes an unsigned size; bail
+       out rather than truncate the request into an undersized allocation */
+    if (in_size > UINT_MAX - out_size) {
+        PREFIX(gz_error)(state, Z_MEM_ERROR, "out of memory");
+        return -1;
+    }
+
+    state->buffers = (unsigned char *)zng_alloc_aligned((unsigned)(in_size + out_size), 64);
```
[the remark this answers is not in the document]
mtl1979: On some platforms, including for example FreeBSD, the maximum allocation size is actually `INT_MAX`. Allocators can internally use signed type, which causes wraparound earlier than unsigned type.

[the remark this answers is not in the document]
dxbjavid: Good point, INT_MAX is the right cap then. The allocator also adds sizeof(void*) + align on top before mallocing, so I capped the total at INT_MAX minus that overhead rather than UINT_MAX. Pushed that. Rejecting the request cleanly now (Z_MEM_ERROR) instead of letting it wrap, and a normal gzbuffer roundtrip still checks out under asan.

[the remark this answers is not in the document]
mtl1979: Even on 64-bit architectures the heap size per application is limited, so using `INT_MAX` as hard limit is safer... Most operating systems like Linux and Windows have own API for larger allocations (1 GB and above), but it's kinda out of scope for this change.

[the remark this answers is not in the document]
dxbjavid: Agreed, INT_MAX as the hard cap is the safe choice here and the current guard rejects the wrapping request cleanly. The bigger-than-INT_MAX allocation path would mean pulling in the platform-specific large-alloc APIs, which is well out of scope for this fix.

gzlib.c
mtl1979: Changing type of `want` means explicit cast is required here.

dxbjavid: Good catch, state->size is unsigned so assigning the size_t want there narrows. Added an explicit (unsigned) cast, pushed. want is always within the INT_MAX cap by then so the value is safe, it just makes the conversion explicit.

gzlib.c
```diff
@@ -80,7 +80,20 @@ int Z_INTERNAL gz_buffer_alloc(gz_state *state) {
         out_size = want * 2;  // double output buffer for decompression
     }
 
-    state->buffers = (unsigned char *)zng_alloc_aligned((in_size + out_size), 64);
+    /* the total here reaches 3 * want, but gzbuffer only guarantees 2 * want
```
nmoinvaz: nit: this is a long comment.

dxbjavid: trimmed it down to two lines, just keeps the bit about the 3*want total and the INT_MAX cap.

nmoinvaz (approved):

Dead2:
This looks correct, and is one way of doing it, but;

I am considering maybe we should be doing an alternate approach instead, actually limit the requested buffer to 1GB for example. That is way way way more than should be beneficial (for reference, 8KB is default in zlib).

It'd need to be added to the zlib-ng api documentation as well.

I figure the compat api could silently clamp to 1GB), while the zlib-ng api returns memory error for too big sizes.
If application requests more memory than can actually be alloced, that of course also returns memory error in any case.

dxbjavid:
Yeah, that makes sense. A hard 1GB cap is well above anything useful in practice, since the decompression speedup flattens out long before that, so clamping there rather than only catching the wrap reads cleaner.

I'd put it in gzbuffer itself: the compat gzbuffer silently clamps the request down to 1GB while zng_gzbuffer returns Z_MEM_ERROR for anything larger, and note the limit in the zlib-ng.h docs next to the existing buffer text. The current gz_buffer_alloc guard can stay as a backstop for the internal 3*want total.

Happy to fold that into this PR, or keep this one as the minimal overflow fix and do the clamp separately, whichever you'd prefer.

Dead2:
If we replace the check in gzbuffer, then we don't need to have another check in gz_buffer_alloc.
I think that will be both a bit cleaner, keeps the checks close to the source (API) like the rest of zlib-ng does, and it keeps gz_buffer_alloc simple.

dxbjavid:
Done that way. Moved the limit into gzbuffer: zng_gzbuffer now returns -1 for anything over 1 GiB and the compat gzbuffer silently clamps to 1 GiB, so the checks sit at the api like you suggested and gz_buffer_alloc is back to being simple (dropped the backstop). Documented the cap in the zlib-ng.h gzbuffer text too.

Checked both paths under asan: native rejects the oversized request and a 1 GiB request still allocates, compat clamps the old 0x55555558 case and the write no longer overruns. Normal 128K roundtrips are unaffected.

gzlib.c
```diff
@@ -332,8 +334,13 @@ z_int32_t Z_EXPORT PREFIX(gzbuffer)(gzFile file, z_uint32_t size) {
         return -1;
 
     /* check and set requested size */
-    if ((size << 1) < size)
-        return -1;              /* need to be able to double it */
+    if (size > GZBUFSIZE_MAX) {
+#ifdef ZLIB_COMPAT
+        size = GZBUFSIZE_MAX;   /* silently clamp to stay compatible with zlib */
+#else
+        return -1;              /* too large */
```
Dead2, line 340: Needs
`PREFIX(gz_error)(state, Z_MEM_ERROR, "out of memory");` before return.

gzlib.c
```diff
@@ -69,8 +69,8 @@ static void gz_reset(gz_state *state) {
 
 /* Allocate in/out buffers for gzFile */
 int Z_INTERNAL gz_buffer_alloc(gz_state *state) {
-    int want = state->want;
-    int in_size = want, out_size = want;
+    size_t want = state->want;
```
Dead2: data type changes in gz_buffer_alloc are no longer needed. It should go back to being unchanged in this PR.

gzlib.c
dxbjavid: Added the gz_error call before the return there, so the strict path reports the memory error now instead of just returning -1.

gzlib.c
dxbjavid: Makes sense, the size_t and the casts weren't really needed. I kept the locals as unsigned rather than going all the way back to int though: with the cap at 1 GiB a write buffer's in_size + out_size still reaches 3*want, so 3 GiB, which is past INT_MAX and would wrap in plain int. unsigned matches state->want and the size zng_alloc_aligned takes, so it stays simple and the casts are gone. Happy to drop the cap lower instead if you'd rather keep it int.

Dead2 (approved):
LGTM

dxbjavid:
Thank you for the review.

dxbjavid:
TY!

PR #2320, opened 2026-06-08
Fix undefined warnings to allow -Wundef to be used

Dead2:
This allows me to enable `-Wundef` during reviews and testing. This is for sure a candidate for -Werror during CI to catch hard-to-spot mistakes early.

The changes to `zendian.h` defines has no effect, they already default to 0 when missing.
The changed cygwin define should have no effect, but I don't know for sure whether it should actually be checking the value or whether presence check is enough.

But the changes to BRAID_W / BRAID_N means these are now actually taken into effect instead of always using the fallback values when choosing what crc tables to use in `crc32_braid_tbl.h`.
I am a bit surprised this hasn't bitten anyone.

The offenders that included `crc32_braid_tbl.h` without also including `crc32_braid_p.h` were:
- crc32_p.h
- inflate_p.h
- arch/power/crc32_power8.c

The two includefiles caused warnings to crop up when compiling all kinds of .c files.

### DIFF
--- a/zendian.h
+++ b/zendian.h
@@ -57,6 +57,28 @@
 #  include <endian.h>
 #endif
 
+#if !defined(BYTE_ORDER) || (!defined(LITTLE_ENDIAN) && !defined(BIG_ENDIAN))
+#  error "Unknown byte order"
+#endif
+
+#if !defined(LITTLE_ENDIAN)
+#  if BIG_ENDIAN == 1234
+#    define LITTLE_ENDIAN 4321
+#  elif BIG_ENDIAN == 4321
+#    define LITTLE_ENDIAN 1234
+#  else
+#    define LITTLE_ENDIAN -1
+#  endif
+#elif !defined(BIG_ENDIAN)
+#  if LITTLE_ENDIAN == 1234
+#    define BIG_ENDIAN 4321
+#  elif LITTLE_ENDIAN == 4321
+#    define BIG_ENDIAN 1234
+#  else
+#    define BIG_ENDIAN -1
+#  endif
+#endif
+
 #if BYTE_ORDER == LITTLE_ENDIAN
 #  define Z_U16_TO_LE(x)    (x)
 #  define Z_U32_TO_LE(x)    (x)
@@ -83,8 +105,6 @@
 #  define Z_U16_FROM_BE(x)  (x)
 #  define Z_U32_FROM_BE(x)  (x)
 #  define Z_U64_FROM_BE(x)  (x)
-#else
-#  error "Unknown byte order"
 #endif
 
 #endif
--- a/arch/generic/crc32_braid_c.c
+++ b/arch/generic/crc32_braid_c.c
@@ -13,7 +13,6 @@
 /* Used by chorba fallback and by arch-specific implementations (s390 vx, risc-v zbc). */
 #ifdef CRC32_BRAID_FALLBACK
 
-#include "crc32_braid_p.h"
 #include "crc32_braid_tbl.h"
 #include "crc32_p.h"
 
--- a/arch/generic/crc32_chorba_c.c
+++ b/arch/generic/crc32_chorba_c.c
@@ -8,7 +8,6 @@
 #endif
 #include "zmemory.h"
 #include "crc32_chorba_p.h"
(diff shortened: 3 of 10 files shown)


zendian.h
```diff
@@ -57,6 +57,13 @@
 #  include <endian.h>
 #endif
 
+#if !defined(LITTLE_ENDIAN)
+#  define LITTLE_ENDIAN 0
+#endif
+#if !defined(BIG_ENDIAN)
+#  define BIG_ENDIAN 0
+#endif
```
mtl1979: This we don't want. It makes testing new architectures and toolchains impossible. Not having these defined means we are missing an essential header.

mtl1979 (changes requested):

zendian.h
Dead2: We already error out if BYTE_ORDER is not set, shouldn't that be enough?
This change does not change current behavior.

The problem this solves is that we for example check `BYTE_ORDER == BIG_ENDIAN`, but on LITTLE_ENDIAN it is not guaranteed to be defined and thus currently silently defaults to 0.
And as can be seen above, we also can't default to 1234 or 4321 since it differs depending on OS.

mtl1979: The worst case is when neither is defined, which this change doesn't handle correctly. I thought about possibility to set these to unique random values, but not sure how much better it would be.

This changes the diagnostic message when `BYTE_ORDER`, `LITTLE_ENDIAN` and `BIG_ENDIAN` are all undefined. This can happen as for example on Linux `endian.h` conditionally define the macros without leading underscores.

zendian.h
```diff
@@ -57,6 +57,28 @@
...
+#  endif
+#elif !defined(BIG_ENDIAN)
+#  if LITTLE_ENDIAN == 1234
+#    define BIG_ENDIAN 4321
+#  elif LITTLE_ENDIAN == 4321
+#    define BIG_ENDIAN 1234
+#  else
+#    define BIG_ENDIAN -1
+#  endif
+#endif
```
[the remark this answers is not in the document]
Dead2: I believe this is excessive checking, the purpose of this PR is to avoid one of the endian defines being undefined, not to handle compilers/environments that do not define BYTE_ORDER correctly.

If the endian conversion macros do stay undefined, it will result in compile errors and thus be detected.

zendian.h
```diff
@@ -57,6 +57,28 @@
 #  include <endian.h>
 #endif
 
+#if !defined(BYTE_ORDER) || (!defined(LITTLE_ENDIAN) && !defined(BIG_ENDIAN))
+#  error "Unknown byte order"
+#endif
```
mtl1979, line 62: This looks much more better... We needed early check that `BYTE_ORDER` is defined, otherwise any other checks just fail when using `-Wundef`. Checking that either `LITTLE_ENDIAN` or `BIG_ENDIAN` is defined fits the existing check in the old version.

zendian.h
mtl1979, line 80: I think setting the missing one of the defines to `-1` is enough of a random value when the existing one doesn't fit the two common possibilities. The important part is that `BIG_ENDIAN` can't be same value as `LITTLE_ENDIAN`, because `LITTLE_ENDIAN` is tested first. This is all preprocessor code so speed penalty is not a deciding factor.

mtl1979 (approved):
LGTM

zendian.h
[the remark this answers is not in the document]
mtl1979: If we ever want to support `PDP_ENDIAN` or other less common endianness, then it should be handled as own PR, but it is very unlikely as the byte order conversion macros would become quite complex. Most of the - especially modern - architectures are `LITTLE_ENDIAN` with remaining architectures we already support being `BIG_ENDIAN`. We don't support bi-endian architectures that have runtime option to change endianness, so these are treated by either `LITTLE_ENDIAN` or `BIG_ENDIAN` paths depending on the default endianness decided by operating system.

In this case we want the conversion macros to be unset for any unsupported target, just like we error out on any target where `BYTE_ORDER` macro is not set at all, so we can handle it cleanly in future PRs without reverting changes made in this PR.

PR #2324, opened 2026-06-10
Clean up insert_string code

Dead2:
- Remove templating, it is obsolete after removal of arch-specific functions.
- Remove unused functions.
- Simplify and clean up the code a bit.
- Readability is significantly improved.


insert_string_p.h
```diff
@@ -1,44 +1,164 @@
-#ifndef INSERT_STRING_P_H_
-#define INSERT_STRING_P_H_
-
-/* insert_string_p.h -- insert_string function generator
+/* insert_string_p.h -- static insert_string and insert_string_roll functions
  *
  * Copyright (C) 1995-2024 Jean-loup Gailly and Mark Adler
+ *
+ * Copyright (C) 2013 Intel Corporation. All rights reserved.
```
nmoinvaz: These authors were only for the crc32 intrinsics hashing right, we don't use that anymore?

insert_string_p.h
```diff
@@ -1,44 +1,164 @@
...
+    for (uint32_t idx = str; strstart < strend; idx++, strstart++) {
+        uint32_t val, h, head;
 
-#define UPDATE_HASH          update_hash_roll
-#define INSERT_STRING        insert_string_roll_static
-#define QUICK_INSERT_STRING  quick_insert_string_roll
-#define QUICK_INSERT_VALUE   quick_insert_value_roll
+        val = strstart[0];
+        h = ((s->ins_h << 5) ^ ((uint8_t)val)) & (32768u - 1u);
+        s->ins_h = h;
```
nmoinvaz: hoist `s->ins_h` to local var.

nmoinvaz (commented):
suggest new naming...
```
update_hash -> update_hash_int
update_hash_roll -> update_hash_roll
quick_insert_value -> insert_value_int
quick_insert_value_roll -> insert_value_roll
quick_insert_string -> insert_string_one_int
quick_insert_string_roll -> insert_string_one_roll
insert_string -> insert_string_many_int
insert_string_roll -> insert_string_many_roll
```

insert_string_p.h
Dead2: I was wondering the same actually.

insert_string_p.h
Dead2: I don't think that is useful now, it's only 1 read and 1 write now after the cleanup commit. It was 4 accesses before the cleanup.

Dead2:
> suggest new naming...
> 
> ```
> update_hash -> update_hash_int
> update_hash_roll -> update_hash_roll
> quick_insert_value -> insert_value_int
> quick_insert_value_roll -> insert_value_roll
> quick_insert_string -> insert_string_one_int
> quick_insert_string_roll -> insert_string_one_roll
> insert_string -> insert_string_many_int
> insert_string_roll -> insert_string_many_roll
> ```

I don't know, not really feeling for those names.
In any case I think that should be a separate PR if needed, it'll touch a lot of places.

insert_string_p.h
Dead2: Sorry, on phone this looked like it was referencing a different function, without the for loop. Will change and test.

insert_string_p.h
```diff
@@ -1,44 +1,152 @@
...
+ *    previous key instead of complete recalculation each time.
+ */
+Z_FORCEINLINE static uint32_t update_hash(uint32_t h, uint32_t val) {
+    h = ((val * 2654435761U) >> 16);
+    return h & HASH_MASK;
+}
+
+Z_FORCEINLINE static uint32_t update_hash_roll(uint32_t h, uint32_t val) {
+    h = ((h << 5) ^ ((uint8_t)val));
+    return h & (32768u - 1u);
```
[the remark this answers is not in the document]
Dead2: I have not verified whether this is right, but the fix is definitely not correct/incomplete.
Level 9 compression goes from 43% to only 46% with this change.

insert_string_p.h
```diff
@@ -1,44 +1,150 @@
...
+    }
+    return head;
+}
+
+Z_FORCEINLINE static uint32_t quick_insert_value_roll(deflate_state *const s, uint32_t str, uint32_t val) {
+    uint32_t h, head;
+
+    h = ((s->ins_h << 5) ^ ((uint8_t)val));
+    h &= (32768u - 1u);
+    s->ins_h = h;
```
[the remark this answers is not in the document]
Dead2: The functions need to have identical ABI, so this won't work.

nmoinvaz:
> I don't know, not really feeling for those names.

The whole `quick_*` is confusing because of `deflate_quick` but they are not related. For example, `s->prev` is not needed for `deflate_quick`.

Dead2:
> > I don't know, not really feeling for those names.

Not saying I'm against renaming, just need to agree on some new names that make sense and are not too cumbersome.

> The whole `quick_*` is confusing because of `deflate_quick` but they are not related. For example, `s->prev` is not needed for `deflate_quick`.

I never really associated those two, but sure, that might be confusing to newcomers.
I always associated the `quick_*` with not having a loop and just doing a single quick insert.

nmoinvaz (approved):

PR #2325, opened 2026-06-11
Add -mbmi to AVX2 and AVX512 compile flags

nmoinvaz:
The AVX2 and AVX512 compile flags enable BMI2 but not BMI1, and TZCNT is a BMI1 instruction. GCC sidesteps the gap by emitting the `rep bsf` encoding, which executes as TZCNT on BMI hardware, but clang gates strictly on the feature bit and emits plain BSF, which is slower on AMD.

With `-mbmi` added, clang emits TZCNT at all three bit-scan sites in `longest_match_avx2` (verified by disassembly; two of them are in compare256's match length calculation). Every CPU with AVX2 also has BMI1, so the flags only affect code that is already behind AVX2 runtime detection, and both build systems probe the flags before use.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Dead2 (approved):
LGTM

PR #2327, opened 2026-06-18
Skip redundant hash maintenance for deflate_quick and huffman/rle

nmoinvaz:
`deflate_quick` only inspects the head of each hash chain — it never walks the `prev` links — and `Z_HUFFMAN_ONLY` / `Z_RLE` read neither `head` nor `prev`. This branch removes hash maintenance those strategies never use.

The main lever is the per-window-slide work. A new `slide_hash_chain` functable entry slides a single table, so `deflate_quick` slides only `head` (skipping the ~64 KB `prev` slide) while huffman/rle skip the slide entirely. Chain-walking strategies keep calling the existing `slide_hash` (both tables, inlined) and are unchanged. `slide_hash_chain` is implemented across all SIMD backends.

`deflate_quick`'s per-position inserts also drop the now-unused `prev` write.

Output is byte-identical at every level and strategy. On an Apple M5 the level-1 deflate path is about 8% faster on the no-CRC benchmark from the slide change, plus roughly 2% from the insert change.

The first commit is a no-op rename that distinguishes the single- and multi-insert helpers and frees the `quick` name for these variants.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -135,6 +135,15 @@ static const config configuration_table[10] = {
  * meaning.
  */
 
+/* Level 1 uses deflate_quick, which only reads the hash chain head, so it can
+ * skip prev maintenance. When quick is compiled out level 1 falls back to
+ * deflate_fast, which walks the chains like every other level. */
+#ifdef NO_QUICK_STRATEGY
+#  define HAVE_QUICK_STRATEGY 0
+#else
+#  define HAVE_QUICK_STRATEGY 1
+#endif
+
 /* rank Z_BLOCK between Z_NO_FLUSH and Z_PARTIAL_FLUSH */
 #define RANK(f) (((f) * 2) - ((f) > 4 ? 9 : 0))
 
@@ -623,9 +632,10 @@ int32_t Z_EXPORT PREFIX(deflateParams)(PREFIX3(stream) *strm, int32_t level, int
     int hashless = level == 0 || strategy == Z_HUFFMAN_ONLY || strategy == Z_RLE;
     int was_hashless = s->level == 0 || s->strategy == Z_HUFFMAN_ONLY || s->strategy == Z_RLE;
 
-    /* Stale if the hash usage flipped (to/from huffman/rle/stored) or the hash
-     * function changed at level 9. */
-    int stale_chain = (hashless != was_hashless) || (level >= 9) != (s->level >= 9);
+    /* Stale if the hash usage flipped (to/from huffman/rle/stored), the hash
+     * function changed at level 9, or quick at level 1 left prev unmaintained. */
+    int stale_chain = (hashless != was_hashless) || (level >= 9) != (s->level >= 9) ||
+                      (HAVE_QUICK_STRATEGY && s->level == 1 && level != 1);
 
     /* Rebuild the hash chains when fill_window is called. */
     if (stale_chain && !hashless) {
@@ -1200,6 +1210,8 @@ void Z_INTERNAL PREFIX(fill_window)(deflate_state *s) {
 
     if (level >= 9)
         insert_batch = insert_roll_batch;
+    else if (HAVE_QUICK_STRATEGY && level == 1)
+        insert_batch = insert_knuth_batch_head;
     else
         insert_batch = insert_knuth_batch;
 
@@ -1249,7 +1261,10 @@ void Z_INTERNAL PREFIX(fill_window)(deflate_state *s) {
             if (UNLIKELY(level >= 9)) {
                 s->ins_h = update_hash_roll(window[str], window[str+1]);
             } else if (str >= 1) {
-                insert_knuth(s, window, str + 2 - STD_MIN_MATCH);
+                if (HAVE_QUICK_STRATEGY && level == 1)
+                    insert_knuth_head(s, window, str + 2 - STD_MIN_MATCH);
+                else
+                    insert_knuth(s, window, str + 2 - STD_MIN_MATCH);
             }
             unsigned int count = s->insert;
             if (UNLIKELY(s->lookahead == 1)) {
--- a/deflate.c
+++ b/deflate.c
@@ -1233,7 +1233,14 @@ void Z_INTERNAL PREFIX(fill_window)(deflate_state *s) {
             s->block_start -= (int)wsize;
             if (s->insert > s->strstart)
                 s->insert = s->strstart;
-            FUNCTABLE_CALL(slide_hash)(s);
+            if (s->strategy != Z_HUFFMAN_ONLY && s->strategy != Z_RLE) {
+                /* Z_HUFFMAN_ONLY and Z_RLE never read the hash chain. deflate_quick
+                 * reads the chain head but never walks prev, so it slides head only. */
+                if (HAVE_QUICK_STRATEGY && level == 1)
+                    FUNCTABLE_CALL(slide_hash_head)(s);
+                else
+                    FUNCTABLE_CALL(slide_hash)(s);
+            }
             more += wsize;
         }
         if (strm->avail_in == 0)
--- a/functable.h
+++ b/functable.h
@@ -34,6 +34,7 @@ struct functable_s {
     uint32_t (* longest_match)      (deflate_state *const s, uint32_t cur_match);
     uint32_t (* longest_match_roll) (deflate_state *const s, uint32_t cur_match);
     void     (* slide_hash)         (deflate_state *s);
+    void     (* slide_hash_head)    (deflate_state *s);
 };
 
 Z_INTERNAL extern struct functable_s functable;
--- a/insert_string_p.h
+++ b/insert_string_p.h
@@ -47,6 +47,21 @@ Z_FORCEINLINE static uint32_t insert_knuth_val(deflate_state *const s, uint32_t 
 }
 
 /* ===========================================================================
+ * Insert string str using a pre-read value, returning the previous head of the
+ * hash chain. The prev link is left untouched since deflate_quick only inspects
+ * the chain head and never walks the chain.
+ */
+Z_FORCEINLINE static uint32_t insert_knuth_val_head(deflate_state *const s, uint32_t str, uint32_t val) {
+    uint32_t h, head;
+
+    UPDATE_HASH_KNUTH(h, val);
+
+    head = s->head[h];
+    s->head[h] = (Pos)str;
+    return head;
+}
+
+/* ===========================================================================
  * Insert string str in the dictionary and set match_head to the previous head
  * of the hash chain (the most recent string with same hash key). Return
  * the previous length of the hash chain.
@@ -63,6 +78,23 @@ Z_FORCEINLINE static uint32_t insert_knuth(deflate_state *const s, unsigned char
         s->prev[str & W_MASK(s)] = (Pos)head;
         s->head[h] = (Pos)str;
     }
+    return head;
+}
+
+/* ===========================================================================
+ * Insert string str read from the window, returning the previous head of the
+ * hash chain. Like insert_knuth but leaves the prev link untouched for
+ * deflate_quick, which only inspects the chain head.
+ */
+Z_FORCEINLINE static uint32_t insert_knuth_head(deflate_state *const s, unsigned char *window, uint32_t str) {
+    uint8_t *strstart = window + str;
+    uint32_t val, h, head;
+
+    val = Z_U32_FROM_LE(zng_memread_4(strstart));
+    UPDATE_HASH_KNUTH(h, val);
+
+    head = s->head[h];
+    s->head[h] = (Pos)str;
     return head;
 }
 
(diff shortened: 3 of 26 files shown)


insert_string_p.h
```diff
@@ -39,11 +39,26 @@ Z_FORCEINLINE static uint32_t quick_insert_value(deflate_state *const s, uint32_
 }
 
 /* ===========================================================================
- * Quick insert string str in the dictionary and set match_head to the previous head
+ * Insert string str using a pre-read value, returning the previous head of the
+ * hash chain. The prev link is left untouched since deflate_quick only inspects
+ * the chain head and never walks the chain.
+ */
+Z_FORCEINLINE static uint32_t insert_string_byval_quick(deflate_state *const s, uint32_t str, uint32_t val) {
```
mtl1979: Using `byval` instead of `byvalue` only saves two characters... As function names should be self-explaining, I see no reason to abbreviate the function name.

insert_string_p.h
```diff
@@ -21,11 +21,11 @@ Z_FORCEINLINE static uint32_t update_hash_roll(uint32_t h, uint32_t val) {
 }
 
 /* ===========================================================================
- * Quick insert string str in the dictionary using a pre-read value and set match_head
+ * Insert string str in the dictionary using a pre-read value and set match_head
  * to the previous head of the hash chain (the most recent string with same hash key).
  * Return the previous length of the hash chain.
  */
-Z_FORCEINLINE static uint32_t quick_insert_value(deflate_state *const s, uint32_t str, uint32_t val) {
+Z_FORCEINLINE static uint32_t insert_string_byval(deflate_state *const s, uint32_t str, uint32_t val) {
```
mtl1979: Using `byval` instead of `byvalue` only saves two characters... As function names should be self-explaining, I see no reason to abbreviate the function name.

test/benchmarks/benchmark_insert_string.cc
```diff
@@ -89,19 +89,19 @@ class insert_string_bench: public insert_string_base {
...
 
-#define BENCHMARK_INSERT_STRING(name, fptr, support_flag) \
-    BENCHMARK_DEFINE_F(insert_string_bench, name)(benchmark::State& state) { \
+#define BENCHMARK_INSERT_STRING_COUNT(name, fptr, support_flag) \
+    BENCHMARK_DEFINE_F(insert_string_count_bench, name)(benchmark::State& state) { \
         if (!(support_flag)) { \
             state.SkipWithError("Function " #name " not supported"); \
         } \
         Bench(state, fptr); \
     } \
```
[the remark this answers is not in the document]
mtl1979: None of the benchmarks seem to do this yet, so it might warrant separate PR for consistency, we only use explicit `return;` when using Google Test, not Google Benchmark, as there the code requires C++23, which means older compilers can't tell if the function returns or not.

Here `SkipWithError` does always return, and no compile-time or run-time error is thrown.

Dead2:
What happens when you change from level 1 to level 6 mid-stream?
Doesn't this mean we need to do some extra work during that change to ensure the window is as it should be?

mtl1979:
Seems to already fail tests when `WITH_NEW_STRATEGIES` is set to `OFF`.

deflate.c
```diff
@@ -1194,9 +1194,11 @@ void Z_INTERNAL PREFIX(fill_window)(deflate_state *s) {
     Assert(s->lookahead < MIN_LOOKAHEAD, "already enough lookahead");
 
     if (level >= 9)
-        insert_string_func = insert_string_roll;
+        insert_string_count_func = insert_string_count_roll;
+    else if (level == 1)
+        insert_string_count_func = insert_string_count_quick;
     else
```
mtl1979: `#ifndef NO_QUICK_STRATEGY`

deflate.c
```diff
@@ -1216,7 +1218,14 @@ void Z_INTERNAL PREFIX(fill_window)(deflate_state *s) {
             s->block_start -= (int)wsize;
             if (s->insert > s->strstart)
                 s->insert = s->strstart;
-            FUNCTABLE_CALL(slide_hash)(s);
+            if (s->strategy != Z_HUFFMAN_ONLY && s->strategy != Z_RLE) {
+                /* Z_HUFFMAN_ONLY and Z_RLE never read the hash chain. deflate_quick
+                 * reads the chain head but never walks prev, so it slides head only. */
+                if (s->level == 1)
+                    FUNCTABLE_CALL(slide_hash_chain)(s->head, HASH_SIZE, (uint16_t)wsize);
+                else
```
mtl1979: `#ifndef NO_QUICK_STRATEGY`

mtl1979 (changes requested):
`deflate_quick` is only used when `NO_QUICK_STRATEGY` is unset.

insert_string_p.h
nmoinvaz: I could also just do `insert_string_value`...

functable.h
```diff
@@ -34,6 +34,7 @@ struct functable_s {
     uint32_t (* longest_match)      (deflate_state *const s, uint32_t cur_match);
     uint32_t (* longest_match_roll) (deflate_state *const s, uint32_t cur_match);
     void     (* slide_hash)         (deflate_state *s);
+    void     (* slide_hash_chain)   (Pos *table, uint32_t entries, uint16_t wsize);
```
nmoinvaz: Another name I could use is `slide_hash_head` or `slide_hash_noprev`, and the function signature would just be `(delfate_state *s)`.

insert_string_p.h
mtl1979: You could... Just opening the discussion what is the best naming style...

functable.h
nmoinvaz: Similarily instead of `insert_string_quick`, I could do `insert_string_noprev`.

mtl1979:
> What happens when you change from level 1 to level 6 mid-stream? Doesn't this mean we need to do some extra work during that change to ensure the window is as it should be?

This is what for we should have tests... If things work correctly only the latest match is propagated through the level change unless we reinsert and hash all the bytes from current window up to current match. If things go wrong, it reads random garbage like the level 1 does now.

`slide_hash` has the main responsibility to make sure the `prev` chain stays valid. I would be really careful with modifying the logic there especially if there can be multiple level changes, both from lower to higher and higher to lower level, in single window.

nmoinvaz:
> What happens when you change from level 1 to level 6 mid-stream? Doesn't this mean we need to do some extra work during that change to ensure the window is as it should be?

What would happen is that the `s->prev` chain would be empty until level 6 started to insert strings. So you'd get a slightly worse compression rate in that rare instance until a full window size is processed.

functable.h
mtl1979: I would avoid names like `noprev`... For someone who isn't really familiar with the internals of `insert_string` yet, it isn't obvious what it does...

Dead2:
> > What happens when you change from level 1 to level 6 mid-stream? Doesn't this mean we need to do some extra work during that change to ensure the window is as it should be?
> 
> What would happen is that the `s->prev` chain would be empty until level 6 started to insert strings. So you'd get a slightly worse compression rate in that rare instance until a full window size is processed.

That would be a regression though. Zlib ensures you can change compression levels without losing history.
In fact this is already a problem we have on changes to/from level 9, so I guess a separate PR to fix this problem would be ok.

I wonder whether we should call deflateSetDictionary (or make a similar function) on level change, just to backfill history when it is needed. I have heard of applications that change the compression level depending on the type of data they are processing (spend less time trying to compress the incompressible image embedded in the otherwise compressible stream for example), no idea whether that is common or not but it is an intelligent way to compress .pdf files for examnple.

nmoinvaz:
Will split the first commit into its own PR.

deflate.c
```diff
@@ -1221,7 +1233,14 @@ void Z_INTERNAL PREFIX(fill_window)(deflate_state *s) {
...
                 s->insert = s->strstart;
-            FUNCTABLE_CALL(slide_hash)(s);
+            if (s->strategy != Z_HUFFMAN_ONLY && s->strategy != Z_RLE) {
+                /* Z_HUFFMAN_ONLY and Z_RLE never read the hash chain. deflate_quick
+                 * reads the chain head but never walks prev, so it slides head only. */
+                if (HAVE_QUICK_STRATEGY && level == 1)
+                    FUNCTABLE_CALL(slide_hash_head)(s);
+                else
+                    FUNCTABLE_CALL(slide_hash)(s);
+            }
```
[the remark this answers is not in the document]
nmoinvaz: Declining this one — `slide_hash` is already unreachable at `level == 0`.

`deflate()` dispatches `level == 0` to `deflate_stored` before any strategy check (deflate.c:1015), and `deflate_stored` never calls `fill_window` (it manages the window itself; the old `s->matches` slide bookkeeping was dropped in b0e408e7). So `fill_window`, and the `slide_hash` inside it, never run at level 0 for any strategy — the proposed `level != 0` guard would always be true here, making it a no-op that just adds a comparison to the hot path.

nmoinvaz:
## Benchmark results (base → head)

Measured PR base (`05b270cc`, upstream/develop) against head (`7e5cd3ca`) with the Google Benchmark suite. Head is faster on every affected path with output byte-identical (all compression ratios unchanged), and the untouched control benchmark stays flat, confirming the deltas aren't thermal drift.

```
benchmark                              base(ns)    head(ns)   delta
level-1 deflate_quick (head-only inserts + head-only slide):
  nocrc/text/16384/1   (no slide)        42316       41355    -2.27%
  level/text/131072/1                   353096      336565    -4.68%
  nocrc/text/131072/1                   351805      338112    -3.89%
  level/text/1048576/1                 3775720     3679567    -2.55%
  nocrc/text/1048576/1                 3738541     3580852    -4.22%
huffman / rle (slide skipped entirely):
  huffman/text/1048576/6               3108823     2960492    -4.77%
  rle/text/1048576/6                   3439316     3273811    -4.81%
  huffman/dna/1048576/6                3351841     3185223    -4.97%
  rle/dna/1048576/6                    5844878     5646872    -3.39%
control (deflate_medium, untouched by this PR):
  level/text/1048576/6                17907921    17870144    -0.21%
```

Geomean over the affected benchmarks: **-3.96%**. The control's -0.21% is the noise floor; every affected path clears it, and related benchmarks agree (level 1 clusters at 2.5-4.7%, huffman/rle at 3.4-5%).

Note this is below the "~8% on level-1 no-CRC" from the commit message, which was measured on different hardware. Level 1 is only exercised for the `text` data type since the non-text ladders start at level 3.

**Method:** interleaved base/head per benchmark (two adjacent passes each, pooled median of 30 reps) so the two binaries share the same thermal state, with the control benchmark as a drift check. Apple M5 throttles under sustained load, so a plain back-to-back run biased whichever binary ran second; interleaving plus median removes that.

**Machine:** Apple M5 (4P+6E, 32 GB), macOS 26.5.2 arm64, Apple clang 21.0.0. Release, static libs, `WITH_BENCHMARKS=ON`.

nmoinvaz:
## Level-1 benchmarks across non-text data types

Following up on the earlier results: the suite only registers level 1 for the `text` data type (non-text ladders start at level 3), so those numbers covered deflate_quick under text only. I added level 1 to the non-text ladder in a throwaway build and measured base (`05b270cc`) vs head (`7e5cd3ca`) across every data type at 1 MB. The level-1 speedup holds everywhere, output stays byte-identical, and the untouched control benchmark is flat.

| Data type (level 1, 1 MB) | Base time | `level` Δ | `nocrc` Δ | Output |
|---|---:|---:|---:|:---:|
| striped_rgb | 0.28 ms | -21.36% | -19.20% | identical |
| mixed | 3.82 ms | -7.75% | -7.35% | identical |
| random | 9.73 ms | -4.77% | -4.44% | identical |
| short_match | 4.26 ms | -4.59% | -3.74% | identical |
| realistic_rgb | 8.92 ms | -1.93% | -4.45% | identical |
| literals | 9.35 ms | -2.20% | -3.30% | identical |
| dna | 4.55 ms | -2.44% | -1.56% | identical |
| text *(cross-check)* | 3.78 ms | — | -4.52% | identical |
| **control** — text, level 6 | 17.77 ms | **-0.07%** | — | identical |

Geomean over the affected level-1 benchmarks is **-6.43%** (-4.1% excluding the striped_rgb outlier). The control's -0.07% is the noise floor.

Base time and win size are inversely correlated: the faster the data compresses at level 1, the larger the fraction of time spent on hash maintenance, so dropping the `prev` writes helps more. striped_rgb compresses in ~0.28 ms and gains ~20%; dna takes ~4.5 ms and gains ~2%. The `level` and `nocrc` columns track each other, as expected since the change is in the shared deflate_quick hash path, not the checksum.

**Method:** interleaved base/head per benchmark (two adjacent passes each, pooled median of 30 reps) so both binaries share the same thermal state, with the control as a drift check. Apple M5 (4P+6E, 32 GB), macOS 26.5.2 arm64, Apple clang 21.0.0, Release + static libs.

Dead2 (approved):
LGTM

PR #2330, opened 2026-06-18
Add initial support for musttail attribute.

mtl1979:
* Add initial support for experimental `musttail` attribute for GCC, Clang and Visual C++
* Visual C++ only allows `musttail` on last line of the function, GCC and Clang allow on any line

### DIFF
--- a/adler32.c
+++ b/adler32.c
@@ -17,7 +17,7 @@ unsigned long Z_EXPORT PREFIX(adler32_z)(unsigned long adler, const unsigned cha
 uint32_t Z_EXPORT PREFIX(adler32_z)(uint32_t adler, const unsigned char *buf, size_t len) {
     if (buf == NULL)
         return ADLER32_INITIAL_VALUE;
-    return FUNCTABLE_CALL(adler32)(adler, buf, len);
+    Z_MUSTTAIL_FT return FUNCTABLE_CALL(adler32)(adler, buf, len);
 }
 #endif
 
--- a/zbuild.h
+++ b/zbuild.h
@@ -72,6 +72,34 @@
 #  endif
 #endif
 
+#ifndef Z_MUSTTAIL
+#  if defined(__clang__) && __clang_major__ >= 17 && !defined(ARCH_WASM)
+#    define Z_MUSTTAIL [[clang::musttail]]
+#    define Z_MUSTTAIL2 [[clang::musttail]]
+#    if defined(ARCH_POWER) && !defined(DISABLE_RUNTIME_CPU_DETECTION)
+#      define Z_MUSTTAIL_FT
+#    else
+#      define Z_MUSTTAIL_FT [[clang::musttail]]
+#    endif
+#  elif defined(__GNUC__) && __GNUC__ >= 15
+#    define Z_MUSTTAIL [[gnu::musttail]]
+#    define Z_MUSTTAIL2 [[gnu::musttail]]
+#    if defined(ARCH_POWER) && !defined(DISABLE_RUNTIME_CPU_DETECTION)
+#      define Z_MUSTTAIL_FT
+#    else
+#      define Z_MUSTTAIL_FT [[gnu::musttail]]
+#    endif
+#  elif defined(_MSC_VER) && _MSC_VER >= 1950 && defined(NDEBUG)
+#    define Z_MUSTTAIL [[msvc::musttail]]
+#    define Z_MUSTTAIL2
+#    define Z_MUSTTAIL_FT [[msvc::musttail]]
+#  else
+#    define Z_MUSTTAIL
+#    define Z_MUSTTAIL2
+#    define Z_MUSTTAIL_FT
+#  endif
+#endif
+
 #ifndef Z_TARGET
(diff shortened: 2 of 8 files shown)


zbuild.h
```diff
@@ -72,6 +72,18 @@
 #  endif
 #endif
 
+#ifndef Z_MUSTTAIL
+#  if defined(__clang__) && __clang_major__ >= 17
+#    define Z_MUSTTAIL [[clang::musttail]]
+#  elif defined(__GNUC__) && __GNUC__ >= 15
+#    define Z_MUSTTAIL [[gnu::musttail]]
+#  elif defined(_MSC_VER) && _MSC_VER >= 1950
+#    define Z_MUSTTAIL [[msvc::musttail]]
```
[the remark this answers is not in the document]
mtl1979: This is still a draft, so we will detect failures after we add the attribute to `return` statements that could benefit from using it... We will eventually disable the attribute for target architectures where it is not supported.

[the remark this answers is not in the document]
mtl1979: MSVC 2026 will issue warning when tail call can't be generated or there might be a performance issue. As we treat warnings as errors when using MSVC, we will catch them early.

adler32.c
```diff
@@ -17,7 +17,7 @@ unsigned long Z_EXPORT PREFIX(adler32_z)(unsigned long adler, const unsigned cha
 uint32_t Z_EXPORT PREFIX(adler32_z)(uint32_t adler, const unsigned char *buf, size_t len) {
     if (buf == NULL)
         return ADLER32_INITIAL_VALUE;
-    return FUNCTABLE_CALL(adler32)(adler, buf, len);
+    Z_MUSTTAIL return FUNCTABLE_CALL(adler32)(adler, buf, len);
```
nmoinvaz: How is that going to work with `FUNCTABLE` though?

mtl1979: It's just an indirect jump... As long as the parameter types and return type match, it does work.

mtl1979:
Some functions will need reordering to be able to support `Z_MUSTTAIL` as it must be the last line of the function. Others will need link-time optimisation (LTO) enabled as the function calls cross compilation unit boundaries.

We still need to investigate what platforms will throw error when the attribute is not supported, I already tested that MSVC doesn't support the attribute for Debug builds.

Architectures that don't support:

* WebAssembly (emcc define `__clang__`)

adler32.c
nmoinvaz: That sounds like magic.

mtl1979: I don't have all the answers yet, but already I know it can't be used if function returns `void`, rest I'm still learning...

zbuild.h
```diff
@@ -72,6 +72,18 @@
 #  endif
 #endif
 
+#ifndef Z_MUSTTAIL
+#  if defined(__clang__) && __clang_major__ >= 17 && !(defined(ARCH_POWER) && defined(ARCH_64BIT))
+#    define Z_MUSTTAIL [[clang::musttail]]
```
mtl1979: Clang crash in backend when compiling for 64-bit PowerPC, so until I know which version fix the ICE, disable use of the attribute.

zbuild.h
```diff
@@ -72,6 +72,18 @@
...
+#  elif defined(__GNUC__) && __GNUC__ >= 15
+#    define Z_MUSTTAIL [[gnu::musttail]]
+#  elif defined(_MSC_VER) && _MSC_VER >= 1950 && defined(NDEBUG)
+#    define Z_MUSTTAIL [[msvc::musttail]]
```
mtl1979: Currently Visual C++ requires at least `/O2`, so we need to disable use of the attribute for Debug builds.

adler32.c
KungFuJesus: I mean gcc already applies this optimization here but I'm more than happy to ensure it with keywords.

mtl1979: So far I only got MSVC and Clang to throw error when tail call optimization failed... GCC silently did just ignore where it couldn't optimize... For PPC64, Clang did ICE when trying TCO.

Adding the attribute is mainly to catch possible regressions if the function parameters or return type are modified without removing the attribute. According to the GCC documentation, it doesn't always automatically try tail call optimization, but for consistency it's better to tag all the possible `return` statements. `Z_MUSTTAIL2` is for cases where current MSVC doesn't want to use TCO, but GCC and Clang will use. If that ever changes, I can add version check just for that macro.

zbuild.h
mtl1979: PowerPC doesn't support `musttail` with indirect or external functions, so use `Z_MUSTTAIL_FT` for indirect calls instead. For "external" function calls to static functions, just don't use `musttail` attribute at all.

adler32.c
mtl1979: Clang seems pretty good at saying when `musttail` attribute shouldn't be used... It starts to complain that static function requires external call which essentially means it rather inline the function than use tail call...

mtl1979:
Rebased to pick some compile fixes.

mtl1979:
Low coverage in `functable.c` has nothing to do with this PR. I warned about it when the stub functions were reworked.

PR #2334, opened 2026-06-19
Fix type mismatch errors in fuzzers.

mtl1979:
* When `ZLIB_COMPAT` is defined, some function parameters use `unsigned long` instead of `size_t`


Dead2 (approved):
LGTM

PR #2337, opened 2026-06-21
Add missing return after SkipWithError in benchmarks

nmoinvaz:
Several benchmark fixtures call `state.SkipWithError(...)` for unsupported CPU features or missing test setup, but do not `return` afterward.

`SkipWithError` only marks the run as skipped, it does not exit the scope (per its own documentation). The benchmark loop body is skipped once the error is set, so this change is behavior neutral today.

It makes the skip handling consistent with the surrounding cases in these files that already `return`, follows Google Benchmark's documented contract that exiting the scope is the caller's responsibility, and keeps the code correct if a fixture is later refactored to do work before the `for (auto _ : state)` loop.

The CPU-support checks live inside the `BENCHMARK_*` registration macros, so the fix is a one-line addition per macro.

### DIFF
--- a/test/benchmarks/benchmark_adler32.cc
+++ b/test/benchmarks/benchmark_adler32.cc
@@ -56,6 +56,7 @@ public:
     BENCHMARK_DEFINE_F(adler32, name)(benchmark::State& state) { \
         if (!(support_flag)) { \
             state.SkipWithError("CPU does not support " #name); \
+            return; \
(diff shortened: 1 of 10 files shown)


nmoinvaz:
This should stop @coderabbitai from flagging these issues.

mtl1979 (approved):
LGTM

PR #2340, opened 2026-06-23
Use macos-15 with Xcode 16.4 for package check workflow

phprus:
Package Check workflow require Xcode 16.4, but it present only in macos-15.

`macos-latest` is now `macos-26` (https://github.com/actions/runner-images/issues/14167).

Error: https://github.com/phprus/zlib-ng/actions/runs/28023368389/job/82945148748#step:6:1


nmoinvaz (approved):

mtl1979:
We could also use `macos-26` and upgrade to Xcode 26.5... We need to do that eventually anyways...

phprus:
Closed in favor of #2341.

PR #2341, opened 2026-06-23
Fix build with macOS 26.

mtl1979:
* After Clang 17 was released originally, target attribute checking was made stricter in the front-end causing all feature checks to fail
* Update feature checks to explicitly state most inclusive target attribute, avoiding including headers from zlib-ng's build tree
* Update `pkgcheck` to use macOS 26 and Xcode 26.5  

Supercedes #2340.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -729,7 +729,7 @@ jobs:
             coverage: macos_clang_arm64_ubsan
 
           - name: macOS Clang (ARM64) Native Instructions
-            os: macos-latest
+            os: macos-26
             compiler: clang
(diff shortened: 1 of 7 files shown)


mtl1979:
Rebased and updated the fix to include DOTPROD feature...

nmoinvaz:
Do you think configure script needs the same changes?

mtl1979:
> Do you think configure script needs the same changes?

I thought we removed support for native builds in `configure` script. This includes specifying `-march=...` manually in CFLAGS.

nmoinvaz:
I think you may be right.

nmoinvaz (approved):

PR #2346, opened 2026-06-30
Force inline compare256 helpers

nmoinvaz:
The compare256_*_static helpers are textually included into match_tpl.h to specialize each longest_match variant, and compare256_rle_* serve the RLE path the same way.

At -O3, clang declines to inline the larger fully-unrolled bodies (the scalar fallback and NEON), emitting an out-of-line call with register spills around it on the hot path instead of inlining the comparison as the per-variant generation intends. GCC already inlines them.

Marking the helpers Z_FORCEINLINE makes clang inline every variant into its consumer, removing the call and surrounding spills. The x86 SIMD variants (SSE2/AVX2/AVX512) were already inlined, so this is a no-op for them.

### DIFF
--- a/arch/arm/compare256_neon.c
+++ b/arch/arm/compare256_neon.c
@@ -11,7 +11,7 @@
 #if defined(ARM_NEON)
 #include "neon_intrins.h"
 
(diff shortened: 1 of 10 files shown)


mtl1979 (approved):
LGTM... Verified with Clang 23.

Issue #1560, opened 2023-08-20
CMake Updates for Superbuilds & find_package


ax3l:
Hi,

Thank you for zlib-ng!

I was working with @FrancescAlted on C-Blosc2 the other week to modernize our large dependency build chain.

In #1557, we noticed a few CMake updates that we could incorporate to zlib-ng would make superbuilds easier and could simplify usage of zlib-ng as drop-in replacement in existing CMake builds that use zlib.

I have the following updates in mind:
- adding alias CMake targets with a namespace: that way installs and superbuilds behave the same (best practice)
- export the CMake targets to the build tree (for superbuilds of downstream consumers)
- add a control option for position independent code for the static target (e.g., https://github.com/Blosc/c-blosc2/pull/547)
- adding alias CMake targets similar to the ones imported by [FindZLIB.cmake](https://cmake.org/cmake/help/latest/module/FindZLIB.html#imported-targets), just with NG suffix, so usage is near-identical as for existing zlib logic
- export a config package, so users do not need to write `FindZLIBNG.cmake` files anymore (e.g., https://github.com/Blosc/c-blosc2/pull/541) and can preserve the CMake targets and their properties

Would you be interested if I took the time and submit PRs for this? :)

mtl1979:
Mixing static and shared libraries is a bad idea... If parent project is built as shared library, all dependencies should also be built as shared libraries. This is because linkers take the first matching symbol they find and ignore any duplicate symbols found later in the command-line, possibly causing parts of binary to use different versions of same function, breaking any internally used "state" variables.

I'm all for adding and exporting CMake alias targets as it has been discussed before, and I fully agree that it makes using zlib-ng as a dependency a lot easier.

ax3l:
Thanks @mtl1979, sounds good!

Yes, it's up to the user to uses statics and shared as they seem necessary. Generally I agree, either go all shared (especially in package managers) or all static (e.g., python wheels) and hide all symbols for encapsulation. Nonetheless, we can expose position independent code as an option, so people are free to mix as they need it. Note that if a static lib is fully "privately" (CMake PRIVATE) consumed into an intermediate share lib, there is no issue with that unless globals are used (what you mention with "state" variables).

mtl1979:
@ax3l That still shifts responsibility for trapping incompatible build parameters from zlib-ng to any project that links against it, which isn't something we should do.

[...]

PR #2347, opened 2026-07-01
Add WITH_PIC option to build the static library as PIC

nmoinvaz:
Adds a `WITH_PIC` CMake option (default `OFF`) that compiles the static library with position independent code by setting `POSITION_INDEPENDENT_CODE` on the static archive target.

Without it, linking zlib-ng's static library into a shared object requires consumers to set `CMAKE_POSITION_INDEPENDENT_CODE` across their entire build. Owning the setting in the library scopes it to the static archive and carries it through the exported CMake target for downstream consumers.

The option resolves to whichever target is the real static archive across the three build configurations, and is a no-op when only a shared library is produced.

Addresses the position-independent-code request from #1560.


mtl1979:
As mentioned in the original issue, this can backfire to us with certain compilers as there is no guarantee that the symbols stay local to the library consuming the `WITH_PIC` enabled version of zlib-ng. Essentially it will cause link-time conflict when two libraries in same executable use zlib-ng with different compilation options. 

This is the reason some developers disabled zlib support in Qt framework before building qmake and recompiled afterwards adding compilation flags that allowed sharing same build of zlib across all Windows binaries that link against either zlib or both zlib and Qt libraries. Without compiling twice, there would be missing symbols.

PR #2355, opened 2026-07-05
Add ABI files for loongarch64 target and update create-abi-files.sh script

phprus:
- Add ABI files for loongarch64 target
- Update `create-abi-files.sh` script for loongarch64 (gcc-14, ubuntu 24.04)

Test for ABI files:
Commit: https://github.com/phprus/zlib-ng/commit/547e46d1197cfb1a238cd08101101e089b8c031d
CI: https://github.com/phprus/zlib-ng/actions/runs/28719249777/job/85166041524#step:9:291


mtl1979:
The whole purpose for uploading the script to GitHub was that other people can improve it by adding more toolchains. I have no objection as long as official Ubuntu repositories have the required toolchains. Only hard requirement is that all the ABI files are created using same exact Ubuntu version.

mtl1979 (approved):
LGTM

PR #2356, opened 2026-07-05
Draft: Split adler32/crc32 copy functions

Dead2:
Splitting the hash_copy functions into:
- one for inflate to handle unaligned buffers of unknown size
- one for deflate to handl 64-byte aligned buffers of known window-sized size

This is a starting point to explore whether we can optimize the deflate hash function to better handle window hashing.

Edit: This turned out to be a failed experiment based on false assumptions.

### DIFF
--- a/arch/x86/adler32_avx512.c
+++ b/arch/x86/adler32_avx512.c
@@ -91,12 +91,72 @@ rem_peel:
     return adler;
 }
 
+// Deflate copying variant
+Z_FORCEINLINE static uint32_t adler32_copy_def(uint32_t adler, uint8_t *dst, const uint8_t *src, size_t len) {
+    uint32_t adler0, adler1;
+    adler1 = (adler >> 16) & 0xffff;
+    adler0 = adler & 0xffff;
+
+    __m512i vbuf, vs1_0, vs3;
+
+    const __m512i dot2v = _mm512_set_epi8(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19,
+                                          20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 32, 33, 34, 35, 36, 37,
+                                          38, 39, 40, 41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, 52, 53, 54, 55,
+                                          56, 57, 58, 59, 60, 61, 62, 63, 64);
+    const __m512i dot3v = _mm512_set1_epi16(1);
+    const __m512i zero = _mm512_setzero_si512();
+
+    while (len >= 64) {
+        __m512i vs1 = _mm512_zextsi128_si512(_mm_cvtsi32_si128(adler0));
+        __m512i vs2 = _mm512_zextsi128_si512(_mm_cvtsi32_si128(adler1));
+        vs1_0 = vs1;
+        vs3 = _mm512_setzero_si512();
+
+        size_t k = MIN(len, NMAX_ALIGNED64);
+        len -= k;
+
+        while (k >= 64) {
+            /*
+               vs1 = adler + sum(c[i])
+               vs2 = sum2 + 64 vs1 + sum( (64-i+1) c[i] )
+            */
+            vbuf = _mm512_load_si512(src);
+
+            _mm512_store_si512(dst, vbuf);
+            dst += 64;
+
+            src += 64;
+            k -= 64;
+
+            __m512i vs1_sad = _mm512_sad_epu8(vbuf, zero);
+            __m512i v_short_sum2 = _mm512_maddubs_epi16(vbuf, dot2v);
+            vs1 = _mm512_add_epi32(vs1_sad, vs1);
+            vs3 = _mm512_add_epi32(vs3, vs1_0);
+            __m512i vsum2 = _mm512_madd_epi16(v_short_sum2, dot3v);
+            vs2 = _mm512_add_epi32(vsum2, vs2);
+            vs1_0 = vs1;
+        }
+
+        vs3 = _mm512_slli_epi32(vs3, 6);
+        vs2 = _mm512_add_epi32(vs2, vs3);
+
+        adler0 = partial_hsum(vs1) % BASE;
+        adler1 = _mm512_reduce_add_epu32(vs2) % BASE;
+    }
+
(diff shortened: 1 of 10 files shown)


test/test_adler32_copy.cc
```diff
@@ -12,23 +12,42 @@ extern "C" {
...
+    static void TearDownTestSuite() {
+        zng_free_aligned(testdata);
+        zng_free_aligned(dstbuf);
+    }
 
 public:
     /* Ensure that adler32 copy functions returns the correct adler and copies the data */
     void adler32_copy_test(adler32_copy_func copyfunc, hash_test params) {
-        ASSERT_LE(params.len, HASH_TEST_MAX_LENGTH);
+        ASSERT_LE(params.len, BUFSIZE);
```
mtl1979, line 35: This line fails to compile, can't expand macro with different types for parameters...

```
D:\a\zlib-ng\zlib-ng\test\test_adler32_copy.cc(35): note: see reference to function template instantiation 'testing::AssertionResult testing::internal::CmpHelperLE<uint32_t,int>(const char *,const char *,const T1 &,const T2 &)' being compiled
                 with
                 [
                     T1=uint32_t,
                     T2=int
                 ]
```

test/test_adler32_copy.cc
```diff
@@ -12,23 +12,42 @@ extern "C" {
 #  include "hash_test_strings_p.h"
 }
 
+#define BUFSIZE 64 * 1024
```
mtl1979, line 15: This is bad... We need unsigned constant.

mtl1979 (changes requested):
Lots of CI failures, incompatible types, warnings treated as errorc...

deflate_p.h
```diff
@@ -167,10 +167,10 @@ Z_FORCEINLINE static unsigned read_buf(PREFIX3(stream) *strm, unsigned char *buf
         memcpy(buf, strm->next_in, len);
 #ifdef GZIP
     } else if (s->wrap == 2) {
-        strm->adler = FUNCTABLE_CALL(crc32_copy)(strm->adler, buf, strm->next_in, len);
+        strm->adler = FUNCTABLE_CALL(crc32_copy_def)(strm->adler, buf, strm->next_in, len);
```
mtl1979, line 170: Fails to compile...

deflate_p.h
```diff
@@ -167,10 +167,10 @@ Z_FORCEINLINE static unsigned read_buf(PREFIX3(stream) *strm, unsigned char *buf
...
 #endif
     } else if (s->wrap == 1) {
-        strm->adler = FUNCTABLE_CALL(adler32_copy)(strm->adler, buf, strm->next_in, len);
+        strm->adler = FUNCTABLE_CALL(adler32_copy_def)(strm->adler, buf, strm->next_in, len);
```
mtl1979, line 173: Fails to compile...

mtl1979 (changes requested):
```native_*_def``` undeclared, fails to compile.

**Comment by Dead2** [comment]
> Lots of CI failures, incompatible types, warnings treated as errorc...

It is a rough draft for local experimentation..

File: arch/x86/adler32_avx512.c [hunk 5]
```diff
@@ -91,12 +91,72 @@ Z_FORCEINLINE static uint32_t adler32_copy_impl(uint32_t adler, uint8_t *dst, co
...
+        len -= k;
+
+        while (k >= 64) {
+            /*
+               vs1 = adler + sum(c[i])
+               vs2 = sum2 + 64 vs1 + sum( (64-i+1) c[i] )
+            */
+            vbuf = _mm512_load_si512(src);
+
+            _mm512_store_si512(dst, vbuf);
```
mtl1979, line 125: Wide store right after a wide load will be really slow. It's usually better to do two 256-bit loads and stores and then combine the halves for rest of the loop. AVX512 should allow interleaving storing the first half and loading the second half. Combining two smaller registers as one larger register shouldn't take many cycles.

KungFuJesus: Yes this was found to be the case for the avx512 VNNI scenario for sure, that's why the whole checksum is using 256 bit vectors for the copy variant.  You actually might need to do that to get the total throughput.  The unrolling was also pretty necessary to get full throughput.

KungFuJesus:
The remainder peeling being absent is a nice little code reduction but ultimately was dead code when calling the existing functions here so other than a shorter sequence to decode won't net much.  The aligned load and stores being explicit won't buy much on these architectures either, that's supposed to be a transparent gain after Nehalem.  The real opportunity might be that we can use _mm256_stream_si256() here instead of _mm256_store_si256(), which might mean we thrash a lot less in cache. Oh and we can also guarantee that we can run this on the lesser, prenehalem CPUs too.  Explicit unaligned loads for the lesser sums will benefit there out of the gate - and we get to subvert memcpy for those cases, too.

arch/x86/adler32_avx512.c
mtl1979: If I remember correctly, AVX512 loads with no specific element width have latency of 7 cycles minimum, going up to few hundred cycles on cache miss, AVX512 stores have latency of 1-5 cycles... But the real bottle neck is the memory controller that doesn't have wide enough bus to handle 512 bits in parallel, so the microcode will break it down to four 128 bit accesses.

AVX256 loads have almost same minimum latency depending on exact processor, but two loads can be interleaved with other instructions, so throughput is better. AVX256 stores have exactly same latency as AVX512 stores.

I've also noticed that some x86-64 processors fetch memory bits in reverse order (end of cache line first), so breaking loads to smaller pieces allows prefetcher to better recognize the pattern and fetch the data to cache early.

mtl1979:
> The remainder peeling being absent is a nice little code reduction but ultimately was dead code when calling the existing functions here so other than a shorter sequence to decode won't net much. The aligned load and stores being explicit won't buy much on these architectures either, that's supposed to be a transparent gain after Nehalem. The real opportunity might be that we can use ```_mm256_stream_si256()``` here instead of ```_mm256_store_si256()```, which might mean we thrash a lot less in cache. Oh and we can also guarantee that we can run this on the lesser, prenehalem CPUs too. Explicit unaligned loads for the lesser sums will benefit there out of the gate - and we get to subvert memcpy for those cases, too.

However ```_mm256_stream_si256``` has about 5-10 times longer latency on certain processors, so it needs more ILP.

PR #2357, opened 2026-07-05
Fix this compilation issue for non-windows machines

KungFuJesus:
I've been manually fixing this up for a bit. We're on C++ so let's just use STL's functions here, which should compile down to min and max instructions.

### DIFF
--- a/test/benchmarks/benchmark_png_shared.h
+++ b/test/benchmarks/benchmark_png_shared.h
@@ -11,6 +11,9 @@ extern "C" {
 #  include <png.h>
 }
 
+#define MAX(a, b) std::max(a, b)
+#define MIN(a, b) std::min(a, b)
+
 typedef struct _png_dat {
     uint8_t *buf;
     int64_t len;


Dead2 (approved):
LGTM

mtl1979:
Why was this not caught by CI? Do we need a CI run to avoid regression or test that the macros are not redefined when the header is included along with some other header?

nmoinvaz:
I personally don't mind if we use `std::min` and `std::max` everywhere in those benchmarks and then remove the macros.

nmoinvaz (approved):

mtl1979:
> I personally don't mind if we use `std::min` and `std::max` everywhere in those benchmarks and then remove the macros.

I don't mind either... Using template functions with C++ is a lot less error-prone than using preprocessor macros. With templates we usually get better diagnostic messages if there is type mismatches and as bonus we usually also avoid double evaluation of expressions passed as parameters.

PR #2360, opened 2026-07-14
Unify slide_hash_sse2 to follow same style as the other variants

Dead2:
This cleans up the code to be simpler and more readable and also prepares it for #2327.

Performance is well within noise in my tests.

Code size increases slightly by 64bytes
`180104    1328       8  181440   2c4c0 libz-ng.so.2`
to:
`180168    1328       8  181504   2c500 libz-ng.so.2`

### DIFF
--- a/arch/x86/slide_hash_sse2.c
+++ b/arch/x86/slide_hash_sse2.c
@@ -15,43 +15,25 @@
 #include "deflate.h"
 
 #include <immintrin.h>
-#include <assert.h>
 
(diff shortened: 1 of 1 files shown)


nmoinvaz (approved):
Thanks I appreciate it. It will make that other commit/PR easier. The old code is harder to follow, and the gains are really minuscule.

mtl1979 (commented):
Changing from 0-based numbering to 1-based numbering adds noise to the patch, but otherwise looks good...

PR #2361, opened 2026-07-14
Clean up unused slide_hash_c code

Dead2:
Clean up slide_hash_c code, removing the original loop that GCC is still struggling to vectorize at -O2.
There really is no upside to keeping the original code anymore.
Also update the comments to reflect current and future #2327 situation.

### DIFF
--- a/arch/generic/slide_hash_c.c
+++ b/arch/generic/slide_hash_c.c
@@ -12,44 +12,21 @@
 #include "deflate.h"
 
 /* ===========================================================================
- * Slide the hash table when sliding the window down (could be avoided with 32
- * bit values at the expense of memory usage). We slide even when level == 0 to
- * keep the hash table consistent if we switch back to level > 0 later.
+ * Slide the hash table when sliding the window down
  */
-static inline void slide_hash_c_chain(Pos *table, uint32_t entries, uint16_t wsize) {
-#ifdef NOT_TWEAK_COMPILER
-    table += entries;
-    do {
-        unsigned m;
-        m = *--table;
-        *table = (Pos)(m >= wsize ? m-wsize : 0);
-        /* If entries is not on any hash chain, prev[entries] is garbage but
-         * its value will never be used.
-         */
-    } while (--entries);
-#else
-    {
-    /* As of I make this change, gcc (4.8.*) isn't able to vectorize
-     * this hot loop using saturated-subtraction on x86-64 architecture.
-     * To avoid this defect, we can change the loop such that
-     *    o. the pointer advance forward, and
-     *    o. demote the variable 'm' to be local to the loop, and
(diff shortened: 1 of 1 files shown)


arch/generic/slide_hash_c.c
```diff
@@ -12,40 +12,19 @@
...
-            Pos t = (Pos)wsize;
-            *q++ = (Pos)(m >= t ? m-t: 0);
-        }
+    unsigned int i;
+    Pos *q = table;
+    for (i = 0; i < entries; i++) {
+        Pos m = *q;
+        Pos t = (Pos)wsize;
+        *q++ = (Pos)(m >= t ? m-t: 0);
     }
```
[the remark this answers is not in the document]
Dead2: This PR does not change how the code works, it is only about cleaning up the dead code.

arch/generic/slide_hash_c.c
```diff
@@ -12,40 +12,19 @@
...
-        for (i = 0; i < entries; i++) {
-            Pos m = *q;
-            Pos t = (Pos)wsize;
-            *q++ = (Pos)(m >= t ? m-t: 0);
-        }
+    unsigned int i;
+    Pos *q = table;
+    for (i = 0; i < entries; i++) {
+        Pos m = *q;
+        Pos t = (Pos)wsize;
```
mtl1979: We should keep this outside the loop as most sane compilers will try to move it outside anyways as it doesn't change between the loop iterations...

arch/generic/slide_hash_c.c
```diff
@@ -12,40 +12,19 @@
 #include "deflate.h"
 
 /* ===========================================================================
- * Slide the hash table when sliding the window down (could be avoided with 32
- * bit values at the expense of memory usage). We slide even when level == 0 to
- * keep the hash table consistent if we switch back to level > 0 later.
+ * Slide the hash table when sliding the window down
  */
 static inline void slide_hash_c_chain(Pos *table, uint32_t entries, uint16_t wsize) {
```
mtl1979: Last parameter should use `Pos` directly, as it would avoid the explicit cast below. It also enforces the contract that `wsize` should always fit in limits of underlying type of `Pos`.

mtl1979 (changes requested):
Some no-op code...

arch/generic/slide_hash_c.c
Dead2: I don't really want to change the code in this PR, mainly remove the obsolete code and comments.

mtl1979: It's basically dead code, as `Pos` currently is same type as `uint16_t`. Doesn't really change the scope of the PR at all...

nmoinvaz (approved):

PR #2362, opened 2026-07-15
Minor cleanup of zng_emit_dist()

Dead2:
Leftovers after a failed attempt at optimization. Compiles the same.
Might make it a tiny bit easier for someone else to understand and optimize it later.

### DIFF
--- a/trees_emit.h
+++ b/trees_emit.h
@@ -120,34 +120,44 @@ static inline uint32_t zng_emit_dist(deflate_state *s, const ct_data *ltree, con
     uint64_t match_bits;
     uint32_t match_bits_len;
 
-    /* Send the length code, len is the match length - STD_MIN_MATCH */
+    /* 1. Process Length Code */
     code = zng_length_code[lc];
-    c = code+LITERALS+1;
+    c = code + LITERALS + 1;
     Assert(c < L_CODES, "bad l_code");
     send_code_trace(s, c);
 
+    /*    Send length code, len is the match length - STD_MIN_MATCH */
     match_bits = ltree[c].Code;
     match_bits_len = ltree[c].Len;
-    /* Get extra bits count and subtract base length from match length */
+
+    /* 2. Get extra bits count and subtract base length */
     lext = lbase_extra[code];
     extra = lext >> 8;
     lc -= lext & 0xff;
-    match_bits |= ((uint64_t)(lc & ((1U << extra) - 1)) << match_bits_len);
+
+    /*    Send length extra bits */
+    uint32_t l_mask = (1U << extra) - 1;
+    match_bits |= (uint64_t)(lc & l_mask) << match_bits_len;
     match_bits_len += extra;
 
+    /* 3. Process Distance Code */
     dist--; /* dist is now the match distance - 1 */
     code = d_code(dist);
     Assert(code < D_CODES, "bad d_code");
     send_code_trace(s, code);
 
-    /* Send the distance code */
+    /*    Send distance code */
     match_bits |= ((uint64_t)dtree[code].Code << match_bits_len);
     match_bits_len += dtree[code].Len;
-    /* Get extra bits count and subtract base distance */
+
+    /* 4. Get extra bits count and subtract base distance */
     lext = dbase_extra[code];
     extra = lext >> 16;
     dist -= lext & 0xffff;
-    match_bits |= ((uint64_t)(dist & ((1U << extra) - 1)) << match_bits_len);
+
+    /*    Send dist extra bits */
+    uint32_t d_mask = (1U << extra) - 1;
+    match_bits |= ((uint64_t)(dist & d_mask) << match_bits_len);
     match_bits_len += extra;
 
     send_bits(s, match_bits, match_bits_len, *bi_buf, *bi_valid);


trees_emit.h
```diff
@@ -120,21 +120,26 @@ static inline uint32_t zng_emit_dist(deflate_state *s, const ct_data *ltree, con
...
+
     /* Get extra bits count and subtract base length from match length */
     lext = lbase_extra[code];
     extra = lext >> 8;
     lc -= lext & 0xff;
-    match_bits |= ((uint64_t)(lc & ((1U << extra) - 1)) << match_bits_len);
+
+    uint32_t l_mask = (1U << extra) - 1;
+    match_bits |= (uint64_t)(lc & l_mask) << match_bits_len;
     match_bits_len += extra;
```
mtl1979, line 141: Single-use variables should be enclosed in own scope to help compiler limit life-time of the variable and avoid unnecessary stack use.

trees_emit.h
```diff
@@ -143,11 +148,14 @@ static inline uint32_t zng_emit_dist(deflate_state *s, const ct_data *ltree, con
...
+
     /* Get extra bits count and subtract base distance */
     lext = dbase_extra[code];
     extra = lext >> 16;
     dist -= lext & 0xffff;
-    match_bits |= ((uint64_t)(dist & ((1U << extra) - 1)) << match_bits_len);
+
+    uint32_t d_mask = (1U << extra) - 1;
+    match_bits |= ((uint64_t)(dist & d_mask) << match_bits_len);
     match_bits_len += extra;
```
mtl1979, line 161: As above, variable is used in one place only, so it doesn't need stack slot.

trees_emit.h
Dead2: What are you suggesting exactly? Enclosing the two lines in `{  ..  }` ?

nmoinvaz: Compiler is super smart these days. If it can't determine the scope of variable it is doomed.

nmoinvaz (approved):

trees_emit.h
mtl1979: > Compiler is super smart these days. If it can't determine the scope of variable it is doomed.

It's pretty much something that depends on C standard version and if extensions are enabled or not. MSVC is notorious for being dumb.

mtl1979: > What are you suggesting exactly? Enclosing the two lines in `{  ..  }` ?

The two lines is minimum that works as own block scope... What is the cleanest might be extending the scope to also the third line...

nmoinvaz: Had Claude analyze it so I'm not misinformed and it said:
> a C block scope controls name visibility (where the identifier is legal to reference), not the variable's live range in codegen. Both clang and gcc compute liveness from actual def→use chains.

I furthermore had it test a simple example, and then also check our code base between the two versions on Clang and GCC using `-O2` and it compiled to the exact same assembly. MSVC too (https://godbolt.org/z/EMzoe8Gv5).

mtl1979: > Had Claude analyze it so I'm not misinformed and it said:
> 
> > a C block scope controls name visibility (where the identifier is legal to reference), not the variable's live range in codegen. Both clang and gcc compute liveness from actual def→use chains.
> 
> 
> 
> I furthermore had it test a simple example, and then also check our code base between the two versions on Clang and GCC using `-O2` and it compiled to the exact same assembly. MSVC too (https://godbolt.org/z/EMzoe8Gv5).

This isn't about simple examples, this is about how required stack space is calculated before the optimizer eliminates any unneeded variables. Some architectures we still support limit stack variables to only 128 bytes as the relative address is encoded in 8 bits in the instruction itself. The remaining 128 bytes are the function parameters. If there is more than 128 bytes of stack variables, the variables that are needed are shifted in memory, which can cause register exhaustion as already seen in our code. Any unnecessary moving of stack variables is drop in performance.

Dead2:
@nmoinvaz I kind of reverted your previous change in order to also have mask part of the tables.
Not sure whether this is optimal or perhaps I should add another field for padding (to 4x8bit and 4x16bit) to make it easier to load.

The coderabbitai suggestions above look.. confused, at least to my first glance.

I am seeing significant speedups for levels 1 (5.4%) and 2 (1.1%), the rest are within noise, so hard to tell.

trees_emit.h
```diff
@@ -115,43 +115,48 @@ static inline void zng_emit_lit(deflate_state *s, const ct_data *ltree, unsigned
...
-    match_bits_len += extra;
 
-    send_bits(s, match_bits, match_bits_len, *bi_buf, *bi_valid);
+    /* 4. Get extra bits count and subtract base distance */
+    const dbase_extra_s dbex = dbase_extra[code];
+    dist -= dbex.base;
+
+    /*    Send dist extra bits */
+    match_bits |= ((uint64_t)(dist & dbex.mask) << match_bits_len);
+    match_bits_len += dbex.extra;
```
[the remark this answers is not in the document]
Dead2: Hmm, suggestion A does seem to work..
Suggestion B did not however.

Added commit for correctness testing.

mtl1979:
> @nmoinvaz I kind of reverted your previous change in order to also have mask part of the tables. Not sure whether this is optimal or perhaps I should add another field for padding (to 4x8bit and 4x16bit) to make it easier to load.

Compilers on architectures except ARM should be quite intelligent in padding anything that is multiple of 3 to be multiple of 4... With all warnings enabled, some compilers might notify about that implicit padding, but adding explicit `struct` field for padding would avoid that warning. ARM is rare case where loading sizes that are multiples of 3 can be optimised.

PR #2364, opened 2026-07-16
Optimize zng_emit_dist alternative B

Dead2:
Optimize zng_emit_dist by having removing base and instead adding mask to extra tables.

This alternative uses structs instead of shifts, resulting in cleaner code.


mtl1979:
This altenative smells less voodoo than the A version... Compilers might also able to optimize this better.

Dead2:
Closing in favor of #2363

PR #2368, opened 2026-07-18
X86 alignment tweaks

KungFuJesus:
This mitigates some rather nasty x86 performance penalties we've been hitting.  I'm going to run a battery of benchmarks against -march=x86-64-v4 compiled binaries (but otherwise vanilla build flags).

We still are losing a very advantageous code alignment lottery that we were hitting for calgary/geo at one point or another.  Oddly it seems to show up heavily in top down analysis as "tma_bad_speculation".  Somehow we started mispredicting much worse for that dataset than the others.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -434,6 +434,34 @@ else()
     add_compile_options(${WARNFLAGS} ${WARNFLAGS_DISABLE})
 endif()
 
+#
+# Intel JCC Erratum mitigation alignment for x86 architectures
+# see: https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/best-practices/mitigation-strategies-jcc-microcode.html
+#
+if(BASEARCH_X86_FOUND)
+    if(CMAKE_C_COMPILER_ID MATCHES "Clang" OR CMAKE_C_COMPILER_ID MATCHES "IntelLLVM")
+        check_c_compiler_flag(-mbranches-within-32B-boundaries HAVE_BRANCHES_WITHIN32)
+        if(HAVE_BRANCHES_WITHIN32)
+            add_compile_options("-mbranches-within-32B-boundaries")
+        endif()
+    elseif(CMAKE_C_COMPILER_ID MATCHES "GNU")
+        check_c_compiler_flag(-Wa,-mbranches-within-32B-boundaries HAVE_BRANCHES_WITHIN32)
+        if(HAVE_BRANCHES_WITHIN32)
+            add_compile_options("-Wa,-mbranches-within-32B-boundaries")
+        endif()
+    elseif(CMAKE_C_COMPILER_ID MATCHES "NVC" OR CMAKE_C_COMPILER_ID MATCHES "NVHPC")
+        check_c_compiler_flag(-Wa,-mbranches-within-32B-boundaries HAVE_BRANCHES_WITHIN32)
+        if(HAVE_BRANCHES_WITHIN32)
+            add_compile_options("-Wa,-mbranches-within-32B-boundaries")
+        endif()
+    elseif (MSVC)
+        check_c_compiler_flag(/QIntel-jcc-erratum HAVE_BRANCHES_WITHIN32)
+        if(HAVE_BRANCHES_WITHIN32)
+            add_compile_options("/QIntel-jcc-erratum")
+        endif()
+    endif()
+endif()
+
 # Set code coverage compiler flags
 if(WITH_CODE_COVERAGE)
     add_code_coverage()
--- a/configure
+++ b/configure
@@ -158,6 +158,10 @@ OBJC='$(OBJZ)'
 PIC_OBJC='$(PIC_OBJZ)'
 INSTALLTARGETS="install-shared install-static"
 UNINSTALLTARGETS="uninstall-shared uninstall-static"
+branch_within32_clang="-mbranches-within-32B-boundaries"
+branch_within32_gcc="-Wa,-mbranches-within-32B-boundaries"
+# Need this set to something that will initially fail the compiler check test
+branch_within32_flag="UNSUPPORTED"
 
 TEST="teststatic"
 
@@ -1346,6 +1350,31 @@ EOF
     fi
 }
 
+check_branch_within32_compiler_flag() {
+    # Check whether the JCC mitigation workaround is supported
+    if test $clang -eq 1; then
+        branch_within32_flag="$branch_within32_clang"
+    elif test $gcc -eq 1; then
+        branch_within32_flag="$branch_within32_gcc"
+    elif test $nvc -eq 1; then
+        branch_within32_flag="$branch_within32_gcc"
+    else
+        BRANCH_WITHIN32_SUPPORTED=0
+    fi
+
+    cat > $test.c << EOF
+int main(void) { return 0; }
+EOF
+    if try $CC -c $CFLAGS $branch_within32_flag $test.c; then
+        BRANCH_WITHIN32_SUPPORTED=1
+        echo "Check whether $branch_within32_flag is available ... Yes." | tee -a configure.log
+    else
+        BRANCH_WITHIN32_SUPPORTED=0
+        branch_within32_flag=
+        echo "Check whether $branch_within32_flag is available ... No." | tee -a configure.log
+    fi
+}
+
 check_neon_compiler_flag() {
     if test "$ARCH" = "aarch64"; then
         neonflag="-march=armv8-a+simd"
@@ -1966,6 +1995,13 @@ case "${ARCH}" in
 
             if test ${MTUNE_RAPTORLAKE_AVAILABLE} -eq 1; then
                 avx2vnniflag="${avx2vnniflag} -mtune=raptorlake"
+            fi
+
+            check_branch_within32_compiler_flag
+
+            if test ${BRANCH_WITHIN32_SUPPORTED} -eq 1; then
+                CFLAGS="${CFLAGS} ${branch_within32_flag}"
+                SFLAGS="${SFLAGS} ${branch_within32_flag}"
             fi
 
             check_avx512_intrinsics
--- a/zbuild.h
+++ b/zbuild.h
@@ -356,4 +356,15 @@
 #  define OPTIMAL_CMP 16
 #endif
 
+/* Allow the developer to specify a power of two sized alignment for a given
+ * code segment. This mitigates some bad DSB2MITE penalties on several Intel
+ * CPUs */
+#if defined(__GNUC__) || defined(__clang__)
+#   define ALIGN_TO_STR_(x) #x
+#   define ALIGN_TO_STR(x) ALIGN_TO_STR_(x)
+#   define P2ALIGN(power) __asm__ volatile (".p2align " ALIGN_TO_STR(power))
+#else
+#   define P2ALIGN(power)
+#endif
+
 #endif
--- a/arch/x86/chunkset_avx512.c
+++ b/arch/x86/chunkset_avx512.c
@@ -89,6 +89,7 @@ static inline uint8_t* CHUNKCOPY(uint8_t *out, uint8_t const *from, size_t len) 
     size_t rem = len % sizeof(chunk_t);
 
     if (len < sizeof(chunk_t)) {
+        P2ALIGN(5);
         mask_t rem_mask = gen_mask(rem);
         chunk = _mm256_maskz_loadu_epi8(rem_mask, from);
         _mm256_mask_storeu_epi8(out, rem_mask, chunk);


KungFuJesus:
### Benchmark Results: CPU Time Mean Comparison

| Corpus / File Path | CPU Change (%) | Old Mean (CPU) | New Mean (CPU) |
| :--- | :---: | :---: | :---: |
| **artificial/a.txt** | $\color{green}-17.49\%$ | 34 | 28 |
| **artificial/aaa.txt** | $\color{green}-15.04\%$ | 6,322 | 5,371 |
| **artificial/alphabet.txt** | $\color{green}-7.95\%$ | 6,693 | 6,161 |
| **artificial/alphabet.txt.hex** | $\color{green}-0.48\%$ | 26,526 | 26,397 |
| **artificial/random.txt** | $\color{green}-8.91\%$ | 301,806 | 274,929 |
| **calgary/bib** | $\color{green}-13.22\%$ | 148,993 | 129,296 |
| **calgary/book1** | $\color{green}-13.17\%$ | 1,225,987 | 1,064,552 |
| **calgary/book2** | $\color{green}-13.04\%$ | 821,905 | 714,690 |
| **calgary/geo** | $\color{green}-10.14\%$ | 268,464 | 241,255 |
| **calgary/news** | $\color{green}-12.54\%$ | 666,579 | 582,969 |
| **calgary/obj1** | $\color{green}-6.29\%$ | 30,677 | 28,746 |
| **calgary/obj2** | $\color{green}-11.79\%$ | 348,098 | 307,073 |
| **calgary/paper1** | $\color{green}-12.03\%$ | 70,413 | 61,946 |
| **calgary/paper2** | $\color{green}-13.26\%$ | 114,702 | 99,496 |
| **calgary/paper3** | $\color{green}-11.34\%$ | 67,353 | 59,714 |
| **calgary/paper4** | $\color{green}-7.07\%$ | 19,715 | 18,321 |
| **calgary/paper5** | $\color{green}-7.78\%$ | 17,878 | 16,488 |
| **calgary/paper6** | $\color{green}-11.78\%$ | 50,724 | 44,747 |
| **calgary/pic** | $\color{green}-4.32\%$ | 278,610 | 266,578 |
| **calgary/progc** | $\color{green}-11.75\%$ | 50,812 | 44,841 |
| **calgary/progl** | $\color{green}-9.10\%$ | 59,285 | 53,889 |
| **calgary/progp** | $\color{green}-9.67\%$ | 41,719 | 37,683 |
| **calgary/trans** | $\color{green}-14.35\%$ | 82,850 | 70,959 |
| **canterbury/alice29.txt** | $\color{green}-13.23\%$ | 208,107 | 180,582 |
| **canterbury/asyoulik.txt** | $\color{green}-13.30\%$ | 194,840 | 168,931 |
| **canterbury/cp.html** | $\color{green}-5.46\%$ | 28,567 | 27,008 |
| **canterbury/fields.c** | $\color{green}-4.52\%$ | 11,543 | 11,021 |
| **canterbury/grammar.lsp** | $\color{green}-2.72\%$ | 5,958 | 5,795 |
| **canterbury/kennedy.xls** | $\color{green}-5.37\%$ | 637,618 | 603,385 |
| **canterbury/lcet10.txt** | $\color{green}-12.83\%$ | 555,949 | 484,609 |
| **canterbury/plrabn12.txt** | $\color{green}-13.24\%$ | 766,596 | 665,070 |
| **canterbury/ptt5** | $\color{green}-3.93\%$ | 279,190 | 268,220 |
| **canterbury/sum** | $\color{green}-10.26\%$ | 48,674 | 43,678 |
| **canterbury/xargs.1** | $\color{green}-6.45\%$ | 7,574 | 7,086 |
| **large/E.coli** | $\color{green}-13.59\%$ | 5,314,092 | 4,591,657 |
| **large/bible.txt** | $\color{green}-13.71\%$ | 4,376,631 | 3,776,609 |
| **large/world192.txt** | $\color{green}-12.32\%$ | 3,041,272 | 2,666,577 |
| **miscellaneous/pi.txt** | $\color{green}-13.65\%$ | 1,835,616 | 1,584,972 |
| **neuro/asl16.nii** | $\color{green}-8.68\%$ | 29,903,072 | 27,308,592 |
| **neuro/asl32.nii** | $\color{green}-4.32\%$ | 44,855,236 | 42,916,274 |
| **silesia/dickens** | $\color{green}-13.20\%$ | 14,744,224 | 12,798,311 |
| **silesia/mozilla** | $\color{green}-9.82\%$ | 73,184,914 | 65,997,760 |
| **silesia/mr** | $\color{green}-9.36\%$ | 11,856,562 | 10,746,712 |
| **silesia/nci** | $\color{green}-11.75\%$ | 16,211,202 | 14,306,765 |
| **silesia/ooffice** | $\color{green}-11.96\%$ | 13,493,910 | 11,879,424 |
| **silesia/osdb** | $\color{green}-9.71\%$ | 11,924,981 | 10,766,986 |
| **silesia/reymont** | $\color{green}-13.00\%$ | 7,438,085 | 6,471,071 |
| **silesia/samba** | $\color{green}-10.17\%$ | 20,835,153 | 18,715,679 |
| **silesia/sao** | $\color{green}-12.90\%$ | 16,722,405 | 14,565,963 |
| **silesia/webster** | $\color{green}-12.46\%$ | 50,313,884 | 44,046,095 |
| **silesia/x-ray** | $\color{green}-7.32\%$ | 16,468,564 | 15,263,202 |
| **silesia/xml** | $\color{green}-11.51\%$ | 3,105,197 | 2,747,738 |
| **snappy/fireworks.jpeg** | $\color{green}-6.17\%$ | 76,723 | 71,986 |
| **snappy/geo.protodata** | $\color{green}-3.00\%$ | 45,830 | 44,455 |
| **snappy/html** | $\color{green}-6.14\%$ | 56,665 | 53,186 |
| **snappy/html_x_4** | $\color{green}-19.23\%$ | 235,030 | 189,829 |
| **snappy/kppkn.gtb** | $\color{green}-13.25\%$ | 158,552 | 137,537 |
| **snappy/paper-100k.pdf** | $\color{red}+10.11\%$ | 187,072 | 205,988 |
| **snappy/urls.10K** | $\color{green}-11.90\%$ | 1,110,937 | 978,681 |

KungFuJesus:
Sorry, was fighting Gemini to format the raw data out to a legible pretty table in markdown.  Surprisingly painful.  Here's the raw results from compare.py that I ran on my Cascade Lake system. 
[results.txt](https://github.com/user-attachments/files/30154217/results.txt)

@Dead2 You should see a pretty sizeable boost from this on your Rocket Lake system.  I'm curious if AMD sees an improvement given that the justification for this is purely Intel microarchitecture born.

mtl1979:
This fails package check, as `configure` isn't updated in same PR.

KungFuJesus:
> This fails package check, as `configure` isn't updated in same PR.

Ugh, yes.  Hmm, hopefully this isn't too painful.

CMakeLists.txt
```diff
@@ -434,6 +434,18 @@ else()
...
+#
+# Intel JCC Erratum mitigation alignment for x86 architectures
+# see: https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/best-practices/mitigation-strategies-jcc-microcode.html
+#
+if(BASEARCH_X86_FOUND)
+    if(CMAKE_C_COMPILER_ID MATCHES "Clang")
+        add_compile_options("-mbranches-within-32B-boundaries")
+    elseif(CMAKE_C_COMPILER_ID MATCHES "GNU")
+        add_compile_options("-Wa,-mbranches-within-32B-boundaries")
+    endif()
```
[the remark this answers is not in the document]
mtl1979: Adding check for minimum Clang version is a good idea, but we can't check for binutils version, only matching GCC version. With all optimizations disabled, we don't really have minimum supported compiler version, unless the compiler doesn't even support the target architecture, which is unlikely for x86/x86_64/AMD64 base architecture. Some compiler versions might have bugs that affect success of compilation, but those are handled case by case.

mtl1979:
Funnily enough even MSVC does partially support this in Release builds since VC++ 2019 (16.5), using `/QIntel-jcc-erratum`. No easy way to emit assembler directives on 64-bit builds though.

Ref: https://devblogs.microsoft.com/cppblog/jcc-erratum-mitigation-in-msvc/

KungFuJesus:
> Funnily enough even MSVC does partially support this in Release builds since VC++ 2019 (16.5), using `/QIntel-jcc-erratum`. No easy way to emit assembler directives on 64-bit builds though.
> 
> Ref: https://devblogs.microsoft.com/cppblog/jcc-erratum-mitigation-in-msvc/

I'll give that a shot once I get the pure GCC and clang builds passing CI.

mtl1979:
> > Funnily enough even MSVC does partially support this in Release builds since VC++ 2019 (16.5), using `/QIntel-jcc-erratum`. No easy way to emit assembler directives on 64-bit builds though.
> > Ref: https://devblogs.microsoft.com/cppblog/jcc-erratum-mitigation-in-msvc/
> 
> I'll give that a shot once I get the pure GCC and clang builds passing CI.

No rush... Just posting the link for reference, as some features/options in MSVC are only mentioned in the blog and not official documentation.

configure
```diff
@@ -1346,6 +1348,26 @@ EOF
...
+EOF
+    if try $CC -c $CFLAGS $branch_within32_flag $test.c; then
+        BRANCH_WITHIN32_SUPPORTED=0
+        echo "Check whether $branch_within32_flag is available ... Yes." | tee -a configure.log
+    else
+        BRANCH_WITHIN32_SUPPORTED=0
+        echo "Check whether $branch_within32_flag is available ... No." | tee -a configure.log
+    fi
+}
+
```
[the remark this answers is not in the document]
mtl1979: Still missing early exit for compilers that are not detected as compatible with either gcc or clang... We don't have every one of them on CI, but that doesn't mean we don't still support them.

configure
```diff
@@ -158,6 +158,8 @@ OBJC='$(OBJZ)'
 PIC_OBJC='$(PIC_OBJZ)'
 INSTALLTARGETS="install-shared install-static"
 UNINSTALLTARGETS="uninstall-shared uninstall-static"
+branch_within32_clang="-mbranches-within-32B-boundaries"
+branch_within32_gcc="-Wa,-mbranches-within-32B-boundaries"
```
mtl1979, line 162: Missing default for `branch_within32_flag`, as it is used unconditionally on the compile test. This will result in hard failure on unsupported compilers instead of silently appending unnecessary white-space to compiler command-line.

mtl1979 (changes requested):
Use of uninitialized variable on unsupported compilers.

CMakeLists.txt
```diff
@@ -434,6 +434,29 @@ else()
...
+    if(CMAKE_C_COMPILER_ID MATCHES "Clang")
+        check_c_compiler_flag(-mbranches-within-32B-boundaries HAVE_BRANCHES_WITHIN32)
+        if(HAVE_BRANCHES_WITHIN32)
+            add_compile_options("-mbranches-within-32B-boundaries")
+        endif()
+    elseif(CMAKE_C_COMPILER_ID MATCHES "GNU")
+        check_c_compiler_flag(-Wa,-mbranches-within-32B-boundaries HAVE_BRANCHES_WITHIN32)
+        if(HAVE_BRANCHES_WITHIN32)
+            add_compile_options("-Wa,-mbranches-within-32B-boundaries")
+        endif()
```
mtl1979, line 456: This skips NVHPC even though it does support `-Wa` to pass arguments to assembler.

mtl1979 (changes requested):
GCC-compatible compilers don't set `CMAKE_C_COMPILER_ID` to `GNU` anymore. NVHPC for example sets it to `NVHPC`.

CMakeLists.txt
KungFuJesus: Added.  We'll just have to see if it supports it.

CMakeLists.txt
```diff
@@ -434,6 +434,34 @@ else()
...
+    if(CMAKE_C_COMPILER_ID MATCHES "NVC" OR CMAKE_C_COMPILER_ID MATCHES "NVHPC")
```
mtl1979: Should be `elseif`... `NVC` is not listed as valid value.

mtl1979 (changes requested):
Fails CI...

CMakeLists.txt
KungFuJesus: Yep, just tweaked that.  Should hopefully bake out now.

CMakeLists.txt
```diff
@@ -434,6 +434,34 @@ else()
...
+#
+# Intel JCC Erratum mitigation alignment for x86 architectures
+# see: https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/best-practices/mitigation-strategies-jcc-microcode.html
+#
+if(BASEARCH_X86_FOUND)
+    if(CMAKE_C_COMPILER_ID MATCHES "Clang")
+        check_c_compiler_flag(-mbranches-within-32B-boundaries HAVE_BRANCHES_WITHIN32)
+        if(HAVE_BRANCHES_WITHIN32)
+            add_compile_options("-mbranches-within-32B-boundaries")
+        endif()
```
mtl1979: ICX/ICPX likes this compiler flag too... CMake detects it as `IntelLLVM`.

configure
```diff
@@ -158,6 +158,9 @@ OBJC='$(OBJZ)'
...
+branch_within32_flag=
```
mtl1979: Empty default will result in the test always passing on line 1367, this is because value of `BRANCH_WITHIN32_SUPPORTED` is ignored after it is set to `0` at line 1361.

Dead2:
AMD 8700GE

### Develop
```
 Level  Comp %      In size  Decomptime min/avg/max/stddev
     1 56.391%     97151495    0.2173/0.2188/0.2194/0.0004
     2 47.581%     81974191    0.2178/0.2190/0.2196/0.0004
     3 46.097%     79417634    0.2113/0.2122/0.2128/0.0004
     4 45.619%     78594336    0.2077/0.2087/0.2093/0.0004
     5 45.172%     77824033    0.2041/0.2058/0.2063/0.0004
     6 45.100%     77699889    0.2040/0.2050/0.2054/0.0004
     7 44.809%     77197491    0.2014/0.2023/0.2029/0.0004
     8 44.788%     77162125    0.2011/0.2022/0.2028/0.0004
     9 43.427%     74816862    0.2085/0.2096/0.2102/0.0004

 avg1  46.554%                                      0.2093

 189832    1336       8  191176   2eac8 libz-ng.so.2

inflate_bench/nocrc/text/1_median             17.7 ns         17.7 ns           15
inflate_bench/nocrc/text/64_median             647 ns          647 ns           15
inflate_bench/nocrc/text/1024_median          1866 ns         1866 ns           15
inflate_bench/nocrc/text/16384_median        10451 ns        10451 ns           15
inflate_bench/nocrc/text/131072_median       60519 ns        60518 ns           15
inflate_bench/nocrc/text/1048576_median     634075 ns       634071 ns           15
```

### PR
```
 Level  Comp %      In size  Decomptime min/avg/max/stddev
     1 56.391%     97151495    0.2215/0.2225/0.2230/0.0003
     2 47.581%     81974191    0.2236/0.2246/0.2251/0.0004
     3 46.097%     79417634    0.2168/0.2176/0.2180/0.0003
     4 45.619%     78594336    0.2127/0.2139/0.2144/0.0004
     5 45.172%     77824033    0.2097/0.2104/0.2109/0.0003
     6 45.100%     77699889    0.2088/0.2095/0.2099/0.0003
     7 44.809%     77197491    0.2057/0.2064/0.2069/0.0003
     8 44.788%     77162125    0.2056/0.2063/0.2067/0.0003
     9 43.427%     74816862    0.2129/0.2140/0.2145/0.0003

 avg1  46.554%                                      0.2139

 198508    1336       8  199852   30cac libz-ng.so.2

inflate_bench/nocrc/text/1_median             18.1 ns         18.1 ns           15
inflate_bench/nocrc/text/64_median             652 ns          652 ns           15
inflate_bench/nocrc/text/1024_median          1892 ns         1892 ns           15
inflate_bench/nocrc/text/16384_median        10768 ns        10768 ns           15
inflate_bench/nocrc/text/131072_median       65269 ns        65269 ns           15
inflate_bench/nocrc/text/1048576_median     657647 ns       657644 ns           15
```

That is a lot bigger code size, and a 2.2% slower decompression in deflatebench on average.

I am wondering whether it would be worth applying this only to AVX2/AVX512 codepaths, and split those into Intel and AMD. That way we might even get a small performance boost for each by applying mtune=intel/amd. I figure the AMD path could be the fallback, it is closer to VIA/Zhaoxin anyway. I have been thinking about experimenting with that already, and this might well be what makes it worthwhile.

That would still leave all the generic code without the workaround, but especially for inflate, all the hot code is in avx2/avx512 codepaths IIRC.

My 9900K is down for the moment, but I don't doubt there is an improvement there.

configure
KungFuJesus: Alright, forced it to be something that has to fail during the check

KungFuJesus:
You may be right about the 9900K, it appears to be based on Sunny Cove.  For CPUs with microcode in the affected list though it's pretty stark.  And that does appear to be a pretty wide breadth of CPUs (nearly everything Skylake derived).  It's a wide enough population of CPUs that it might be worth doing but if we're worried about code size, bifurcating two x86 avx2/av512 deflate/inflate dispatches _might_ be an equally hard thing to stomach.  

That compile time argument does have some adjustments that could be made as for where and when to apply the padding, maybe there's a middle ground that imposes least harm?

Dead2:
> It's a wide enough population of CPUs that it might be worth doing but if we're worried about code size, bifurcating two x86 avx2/av512 deflate/inflate dispatches _might_ be an equally hard thing to stomach.

While I do want to keep the library size down if we can, having alternate functions (effectively meaning all other than the one you use is dead code) is less of a problem than adding size to the specific codepaths that are used. If you understand what I mean. A not-selected function does not push other code out of instruction cache for example.

> That compile time argument does have some adjustments that could be made as for where and when to apply the padding, maybe there's a middle ground that imposes least harm?

We should at least investigate whether applying it to only AVX code is good enough, that way we don't apply it to the whole codebase.

mtl1979:
All compilers allow limiting the effect of the mitigation to at least translation unit or function level... I contacted Microsoft about feature request that would increase the granularity to block scope level, but I haven't gotten reply yet...

The current configuration system in zlib-ng (both configure and CMake) allow limiting "extra" compiler flags to just files using AVX2 or AVX512 with minimal effort, so excluding files that have highest performance penalty should be quite clean solution.

zbuild.h
```diff
@@ -356,4 +356,15 @@
 #  define OPTIMAL_CMP 16
 #endif
 
+/* Allow the developer to specify a power of two sized alignment for a given
+ * code segment. This mitigates some bad DSB2MITE penalties on several Intel
+ * CPUs */
+#if defined(__GNUC__) || defined(__clang__)
+#   define ALIGN_TO_STR_(x) #x
```
nmoinvaz, line 363: I suggest just to make this in zutil.h, they are general purpose and I can see them being reused in the future.
```c
#define STRINGIFY_(x) #x
#define STRINGIFY(x) STRINGIFY_(x)
```

Then use it here:
```c
#define P2ALIGN(power) __asm__ volatile (".p2align " STRINGIFY(power))
```

configure
```diff
@@ -160,6 +158,10 @@ OBJC='$(OBJZ)'
 PIC_OBJC='$(PIC_OBJZ)'
 INSTALLTARGETS="install-shared install-static"
 UNINSTALLTARGETS="uninstall-shared uninstall-static"
+branch_within32_clang="-mbranches-within-32B-boundaries"
```
nmoinvaz, line 161: Perhaps these should be moved closer to similar flags like below `s390vxflag`.

zbuild.h
mtl1979: We already have stringify-style macro, it's defined in `zarch.h`, which is included from `zbuild.h`.

configure
mtl1979: Grouping similar flags close to each other makes it easier to find them, especially when later adding new flags. Only exception to the guideline is when the "default" value depends on value of another variable, which is not the case here.

zbuild.h
nmoinvaz: Ah, perhaps we should move it and just name it `Z_STRINGIFY` then. Crazy I couldn't find it because I look for it before

mtl1979: I'm glad it's someone else who can't find something... Finding things with grep can be confusing sometimes when there is either too much noise or have to know what to look for when things can be spelled more than one way...

PR #2370, opened 2026-07-20
validate distance in inflateBack MATCH state

dxbjavid:
The safe_mode bailout in inflate_fast() sets state->mode = MATCH and stores the decoded distance before it reaches the dist > op / op > whave checks a few lines below, so the offset arrives at the MATCH state unvalidated. inflate() copes with that because its own MATCH case re-checks the distance against whave, but inflateBack()'s MATCH case copies straight away, and its DIST case is the only place that validation happens. Feeding inflateBack() a raw deflate stream whose match distance exceeds the caller's window then makes wsize - offset wrap and chunkmemset_safe read from before the caller-supplied window buffer, with those bytes going out through the out() callback. I hit it under asan with a 256-byte window and a length-258 match at distance 400, which lands late enough in a window pass that the bailout fires.

Adding the check that inflateBack()'s DIST case already does, so the stream is rejected as invalid rather than copied. The regression test covers the same shape and fails on the unpatched tree.

### DIFF
--- a/infback.c
+++ b/infback.c
@@ -491,7 +491,12 @@ int32_t Z_EXPORT PREFIX(inflateBack)(PREFIX3(stream) *strm, in_func in, void *in
         case MATCH:
             /* Copy back-reference that inflate_fast() could not complete due to
                insufficient output space. state->length and state->offset were set
-               by the safe_mode MATCH bailout in inflate_fast(). */
+               by the safe_mode MATCH bailout in inflate_fast(), which bails out
+               before validating the distance, so check it here. */
+            if (state->offset > state->wsize - (state->whave < state->wsize ? left : 0)) {
+                SET_BAD("invalid distance too far back");
+                break;
+            }
             do {
                 ROOM();
                 copy = state->wsize - state->offset;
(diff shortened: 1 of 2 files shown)


nmoinvaz:
Is this the same as https://github.com/zlib-ng/zlib-ng/pull/2317?

dxbjavid:
Yes, the infback.c hunk overlaps — #2317 has the same MATCH-state distance check, just as one piece of a bigger change for #2316 (the over-read past the end of the exact-size window, plus the inffast_tpl.h extra_safe routing). I only spotted it after opening this, sorry for the noise.

The difference is scope: this one is just the missing check plus a regression test in the existing test_inflate_back.cc. If you'd rather take #2317 whole then this is redundant and I'm happy to close it. If that one ends up needing more discussion because of the inffast changes, the check on its own is independent of the padding issue and could go in separately. Your call, either way.

nmoinvaz (approved):

mtl1979:
PR #2317 is failing CI and hasn't had any updates in quite a while... This one passed the CI tests, so I don't see anything bad in overlap... We usually only avoid having two active PRs changing same parts of the code from two different collaborators unless the PRs can be merged sequentially.

dxbjavid:
Understood, I'll leave this one open then. The check here is self-contained, so if #2317 gets picked back up later the window padding part of it should still apply on top without conflicting with this.

Issue #1997, opened 2025-11-08
RISC-V Zbc extension detection is broken


ziyao233:
In `arch/riscv/crc32_zbc.c`, RISC-V Zbc extension is detected by checking `getauxval(AT_HWCAP)` and seeing whether `HWCAP_ISA_ZBC` (defined as `1 << 29`) is set[1]. However, current Linux kernel doesn't define bit 29 of `AT_HWCAP`[2]. Testing in QEMU with Zbc enabled and Linux 6.17.4 also shows has_zbc isn't correctly set. The detection logic of Zbc seems broken.

As a note, it's unlikely to have a separate bit for Zbc extension in AT_HWCAP[3]. An alternative detection method is through `riscv_hwprobe(2)` syscall[4], which however only provides information of Zbc extension since Linux 6.8[5].

[1]: https://github.com/zlib-ng/zlib-ng/blob/20083c8efc259255c77ed5fd57f8d58110b3b2fb/arch/riscv/riscv_features.c#L14
[2]: https://github.com/torvalds/linux/blob/e811c33b1f137be26a20444b79db8cbc1fca1c89/arch/riscv/include/uapi/asm/hwcap.h#L11-L24
[3]: https://github.com/torvalds/linux/blob/0d7bee10beeb59b1133bf5a4749b17a4ef3bbb01/arch/riscv/kernel/cpufeature.c#L611
[4]: https://docs.kernel.org/arch/riscv/hwprobe.html
[5]: https://github.com/torvalds/linux/commit/be6bef2acb75ca980671de19d406c0310d646d2b

mtl1979:
Initially I thought custom kernel is needed for ZBC until official support is added and standardized. We can add support for hwprobe later, but guarding it based on kernel version might open another Pandora's box.

Icenowy:
The extension Zbc itself uses no features from new kernels, but the hwprobe key is only in 6.8 .

In addition, there should be no need for guarding hwprobe against kernel version -- it will just get ENOSYS on older kernels.

Here's a test program:

```
#include <sys/hwprobe.h>
#include <stdio.h>

int main()
{
	int ret;

	struct riscv_hwprobe probe = {
		.key = RISCV_HWPROBE_KEY_IMA_EXT_0
	};

	ret = __riscv_hwprobe(&probe, 1, 0, NULL, 0);
	if (ret) {
		fprintf(stderr, "Failed to probe the hardware: %d\n", ret);
		return ret;
	}

[...]

PR #2371, opened 2026-07-21
riscv: remove ineffective Zbc HWCAP fallback

deskflag-cfc:
The `AT_HWCAP` fallback path in `riscv_features.c` tests `ISA_ZBC_HWCAP`, defined as `1 << 29`. No such hwcap bit exists: Linux only encodes single-letter ISA extensions in `AT_HWCAP` (the highest defined bit is `COMPAT_HWCAP_ISA_V`, `'V' - 'A'` = 21), and the kernel has ruled out adding hwcap bits for multi-letter extensions — hwprobe is the designated mechanism ([uapi/asm/hwcap.h](https://github.com/torvalds/linux/blob/master/arch/riscv/include/uapi/asm/hwcap.h), [cpufeature.c comment](https://github.com/torvalds/linux/blob/0d7bee10beeb59b1133bf5a4749b17a4ef3bbb01/arch/riscv/kernel/cpufeature.c#L611)).

So this check can never succeed on any mainline kernel, and Zbc goes undetected on the hwcap fallback path either way. Worse, since bit 29 is unassigned, a downstream kernel could give it a different meaning, in which case `has_zbc` would false-positive and crc32 would execute `clmul` on hardware without Zbc (SIGILL).

This removes the define and the assignment, and documents why `has_zbc` intentionally stays 0 on the hwcap path.

Behavior is unchanged on all kernels:
- The `cpu_features` struct is zeroed in `cpu_features_setup()` before `riscv_check_features()` runs, so `has_zbc` simply remains 0 when hwprobe is unavailable — exactly what the bit-29 test already yields today.
- The working detection via hwprobe (`RISCV_HWPROBE_EXT_ZBC`, Linux 6.8+), added in #2130, is untouched.

References #1997, which documented this with the same kernel citations.

**Tested:** SpaceMiT K3 (X100: Zbc + Zvbc silicon), Linux 6.18, native build with `-DWITH_GTEST=ON -DWITH_BENCHMARKS=ON`: full ctest suite passes (71/71); Zbc is still detected via hwprobe and the `crc32/riscv` benchmark variant runs unchanged (~6.5 GB/s at 256 KiB). `grep -rn ISA_ZBC_HWCAP` confirms no other references in the tree.

### DIFF
--- a/arch/riscv/riscv_features.c
+++ b/arch/riscv/riscv_features.c
@@ -18,7 +18,6 @@
 #endif
 
 #define ISA_V_HWCAP (1 << ('v' - 'a'))
-#define ISA_ZBC_HWCAP (1 << 29)
 
 static int riscv_check_features_runtime_hwprobe(struct riscv_cpu_features *features) {
 #if defined(__NR_riscv_hwprobe) && defined(RISCV_HWPROBE_KEY_IMA_EXT_0)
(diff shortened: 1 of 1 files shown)


mtl1979:
This was already discussed and rejected as misinformation. We support any kernel, not just mainline kernels as a lot of development happens before patches are merged upstream.

deskflag-cfc:
Fair enough — closing this

PR #2384, opened 2026-07-26
Use z_const for setting msg to literal strings.

mtl1979:
Update SET_BAD macro to cast error message strings as z_const char * instead of char *. Split out of #2242.

Upstream: https://github.com/madler/zlib/commit/2ba25b2d


nmoinvaz (approved):

Dead2 (approved):
LGTM

PR #2385, opened 2026-07-26
Avoid out-of-bounds pointer arithmetic in inflateCopy().

mtl1979:
Set distcode along with lencode when building code length tables to avoid UBSan complaints about out-of-bounds pointer computation. Split out of #2242.

Upstream: https://github.com/madler/zlib/commit/f7d01aae


nmoinvaz (approved):

Dead2 (approved):
LGTM

PR #2387, opened 2026-07-26
Reject invalid MAX_WBITS values at compile time.

mtl1979:
Deflate algorithm doesn't support window bits more than 15 as it would break `slide_hash` that require 1 extra bit for saturated subtraction to work. We don't currently support unofficial Deflate64/Deflate9 algorithm which however does use 16 window bits, but also wider variables. Split out of #2242.

Upstream: https://github.com/madler/zlib/commit/ef24c4c7


nmoinvaz (approved):

Dead2 (approved):
LGTM

PR #2388, opened 2026-07-26
Avoid a fruitless memory copy in gzread.c.

mtl1979:
Split out of #2242.

Upstream: https://github.com/madler/zlib/commit/aa27ba46

### DIFF
--- a/gzread.c
+++ b/gzread.c
@@ -79,10 +79,12 @@ static int gz_avail(gz_state *state) {
         if (strm->avail_in) {       /* copy what's there to the start */
             unsigned char *p = state->in;
             unsigned const char *q = strm->next_in;
-            unsigned n = strm->avail_in;
-            do {
-                *p++ = *q++;
-            } while (--n);
+            if (q != p) {
+                unsigned n = strm->avail_in;
+                do {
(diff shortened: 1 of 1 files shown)


nmoinvaz (approved):

gzread.c
```diff
@@ -79,10 +79,12 @@ static int gz_avail(gz_state *state) {
...
-            unsigned n = strm->avail_in;
-            do {
-                *p++ = *q++;
-            } while (--n);
+            if (q != p) {
+                unsigned n = strm->avail_in;
+                do {
+                    *p++ = *q++;
+                } while (--n);
+            }
```
Dead2, line 87: Couldn't this copying be handled more efficiently by chunkcopy_safe?
It is not relevant for this PR though.

LGTM

Dead2 (approved):

gzread.c
nmoinvaz: Good point, we would probably need to add the same pointer check to`chunkcopy_safe`.

mtl1979: Not my area of expertise, but someone else can make another PR to improve what's now inside the `if` block...

PR #2391, opened 2026-07-26
Initialize data_type in inflateResetKeep and update inflatePrime docs.

mtl1979:
Assure that data_type is correct after initialization and a Z_NEED_DICT return. Elaborate on the inflatePrime() error return and correct data_type description in inflate() docs. Split out of #2242.

Upstream: https://github.com/madler/zlib/commit/473f7853


nmoinvaz (approved):
I've never understood the need for inflatePrime().. maybe one day..

PR #2393, opened 2026-07-26
Note that gzseek() requests are deferred until the next operation.

mtl1979:
Split out of #2242.

Upstream: https://github.com/madler/zlib/commit/aaf94f38


nmoinvaz (approved):
Seems like an ok comment.

PR #2395, opened 2026-07-26
Allow gzflush() to write empty gzip members.

mtl1979:
Before this, a gzflush() with Z_FINISH and no data to write would do nothing. Now it will write an empty gzip member. Split out of #2242.

Upstream: https://github.com/madler/zlib/commit/a2b61271


nmoinvaz (approved):

nmoinvaz:
I'm really not sure the purpose of this type of thing but I don't see a problem with it.

PR #2399, opened 2026-07-26
Clean up the use of switch cases.

mtl1979:
Use default for exhaustive cases, initialize variables at declaration. Split out of #2242. Removing some dummy assignments broke build with Visual C++, so reverted.

Upstream: https://github.com/madler/zlib/commit/fb8dd62f

### DIFF
--- a/inftrees.c
+++ b/inftrees.c
@@ -143,27 +143,27 @@ static inline void count_lengths(uint16_t *lens, int codes, uint16_t *count) {
  */
 int Z_INTERNAL zng_inflate_table(codetype type, uint16_t *lens, unsigned codes,
                                  code * *table, unsigned *bits, uint16_t *work) {
-    unsigned len;               /* a code's length in bits */
-    unsigned sym;               /* index of code symbols */
-    unsigned min, max;          /* minimum and maximum code lengths */
-    unsigned root;              /* number of index bits for root table */
-    unsigned curr;              /* number of index bits for current table */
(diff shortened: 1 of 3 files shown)


nmoinvaz (approved):

inftrees.c
```diff
@@ -159,9 +159,9 @@ int Z_INTERNAL zng_inflate_table(codetype type, uint16_t *lens, unsigned codes,
     unsigned mask;              /* mask for low root bits */
     code here;                  /* table entry for duplication */
     code *next;                 /* next available space in table */
-    const uint16_t *base;       /* base value table to use */
-    const uint16_t *extra;      /* extra bits table to use */
-    unsigned match;             /* use base and extra for symbol >= match */
+    const uint16_t *base = NULL; /* base value table to use */
```
nmoinvaz: Do we want to re-align all the comments abvoe and below?

mtl1979: We haven't done that for larger blocks before, only if there is just a few lines. Generally mixing white-space and code changes is frowned upon, but if it's adding one of two spaces to align the comments correctly, then it's kinda border case.

nmoinvaz: Fixing it looks good, thanks!

nmoinvaz (approved):

inftrees.c
mtl1979: It takes just a few seconds of my time and the whole point of splitting the huge PR was to make things easier to review, fix and clean up...

PR #2402, opened 2026-07-27
Clarify naming of bits/code trace macros

Dead2:
Clarify naming of bits/code trace macros.
They don't send and trace, they only trace.

Their meaning was easy to confuse when we also use `send_bits()` and `send_code()`.


nmoinvaz (approved):
Good change.

PR #2403, opened 2026-07-29
Bump zlib compatibility to 1.3.2

mtl1979:
* Update copyright years to 2026.
* Update version strings in deflate and inflate copyrights to "1.3.2"
* Update zlib version macros to 1.3.2
* Update ABI check base commit to zlib 1.3.2

Final part split from #2242.

NOTE: All controversial parts from 1.3.2 have been excluded, non-controversial parts should be merged before this PR.


nmoinvaz (approved):

Dead2:
@mtl1979 Perhaps this should up the abi checks to compare against zlib 1.3.2 as well?

mtl1979:
> @mtl1979 Perhaps this should up the abi checks to compare against zlib 1.3.2 as well?

The way ABI check works, updating the script and ABI files must be separate PR as it is the only way to have stable commit hash to compare against. We should update base commit for both compat and non-compat builds.

I prefer all other PRs are merged before updating the base commit for non-compat build as it will avoid any issues when other commits alter the exports.

mtl1979:
Updated ABI check to compare against last zlib commit that actually builds.

mtl1979:
Seems like `-m32` is now broken, so we need to use gcc targeting i686 instead...

`gcc-multilib` conflicts with a lot of toolchain packages, thus it can't be installed and `/usr/include/asm` will not exist while building zlib.

nmoinvaz (approved):

Dead2:
@mtl1979 This needs a rebase

PR #2407, opened 2026-08-06
Raise longest_match search floor for Z_FILTERED in deflate_slow

nmoinvaz:
`Z_FILTERED` discards matches of length 5 and shorter in `deflate_slow`, but only after `longest_match` has already walked the hash chain to find them. Every search started from a floor of 2, accepted length 3-5 candidates it was about to throw away, and the discard reset `prev_length` to 2 so the next search started from the weakest floor again.

This seeds the search floor through `prev_length` around the `longest_match` call. Short candidates now fail the wider prefilter mask instead of reaching `compare256`, and the level 9 roll variant can run its entry offset search immediately. The floor is clamped below `good_match`, otherwise the phantom length triggers the reduced chain search and hurts compression where `good_match` is small, such as level 4 with `NO_MEDIUM_STRATEGY`.

The strategy check is hoisted out of the per-position loop. On AArch64 the floor bump and the discard compile to two `csel` with no in-loop strategy load, and on x86-64 neither clang nor gcc adds loop-carried spills.

Benchmark results on Apple M5, `deflate_bench/filtered` at level 9 (Δ time vs develop):

| Data type     | 131072 | 1048576 |
|---------------|-------:|--------:|
| literals      | -38.7% |  -40.8% |
| realistic_rgb | -39.1% |  -39.4% |
| mixed         | -24.2% |  -23.3% |
| random        | -19.9% |  -24.4% |
| short_match   | -15.4% |  -12.7% |
| text          | -3.8% (1024) | -6.9% |
| dna           |  +0.1% |   -0.8% |
| striped_rgb   |  +0.4% |   +0.5% |

Default strategy is unchanged in output and speed. Compressed output is byte-identical across levels 4-9 in both strategy configurations on the test corpora, apart from the level 9 offset search occasionally choosing an equally valid match.


deflate_slow.c
```diff
@@ -68,11 +73,14 @@ Z_INTERNAL block_state deflate_slow(deflate_state *s, int flush) {
...
+            uint32_t prev_length = s->prev_length;
+            s->prev_length = MAX(prev_length, match_floor);
             match_len = longest_match(s, hash_head);
+            s->prev_length = prev_length;
             /* longest_match() sets match_start */
 
-            if (match_len <= 5 && (s->strategy == Z_FILTERED)) {
-                /* If prev_match is also WANT_MIN_MATCH, match_start is garbage
+            if (match_len <= match_discard) {
+                /* No match, or one the strategy discards. match_start may be garbage
```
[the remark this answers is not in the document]
nmoinvaz: To bypass the bookkeeping would introduce a regression, correct?

nmoinvaz:
Some history on this heuristic, since it is easy to wonder who actually uses `Z_FILTERED`.

The strategy predates zlib's first public release. The ChangeLog entry for zlib 0.5 (April 1995) reads:

> renamed the 'filter' parameter of deflateInit2 as 'strategy'. Added Z_FILTERED and Z_HUFFMAN_ONLY constants.

So `deflateInit2` originally had a parameter literally named `filter`, and the strategy API is a generalization of it. zlib was written in 1995 as the compression layer for PNG, whose core idea is row filters/predictors, and `Z_FILTERED` exists for exactly those filtered scanlines. The v0.71 snapshot (14 April 1995) already contains today's documentation wording and the `match_length <= 5` discard alongside `TOO_FAR`, essentially character for character what `deflate_slow` inherited, so the heuristic this PR speeds up is thirty years old.

Its main consumer today is invisible to a code search. libpng's default IDAT strategy is `#define PNG_Z_DEFAULT_STRATEGY 1` in generated pnglibconf.h, applied in pngwutil.c whenever row filtering is enabled, so the `Z_FILTERED` name never appears at call sites. Verified at runtime with libpng 1.6.57 against zlib-ng, writing 512x512 RGB at level 9 with factory defaults: the output is byte-identical to explicitly requesting `Z_FILTERED` (375635 bytes) and differs from `Z_DEFAULT_STRATEGY` (416661 bytes). Pillow also selects it explicitly for all PNG encoding in ZipEncode.c, and its `save(optimize=True)` path forces level 9, which lands on the `deflate_slow` path this PR improves.

deflate_slow.c
```diff
@@ -60,19 +66,22 @@ Z_INTERNAL block_state deflate_slow(deflate_state *s, int flush) {
...
              */
+            s->prev_length = MAX(prev_length, match_floor);
             match_len = longest_match(s, hash_head);
+            s->prev_length = prev_length;
             /* longest_match() sets match_start */
 
-            if (match_len <= 5 && (s->strategy == Z_FILTERED)) {
-                /* If prev_match is also WANT_MIN_MATCH, match_start is garbage
+            if (match_len <= match_discard) {
+                /* No match, or one the strategy discards. match_start may be garbage
```
mtl1979: `or one that the strategy discards`... `that` is mandatory when connecting noun with a sentence.

nmoinvaz:
Measured this end to end through libpng using `benchmark_zlib_apps` (Apple M5, 1024x1024 realistic RGB, whole PNG encode including filtering):

| `png_encode` realistic | Δ time |
|---|---:|
| level 9 | -47.0% (172.0 ms -> 91.2 ms) |
| level 8 | -8.3% |
| level 7 | -8.3% |
| levels 0-6 | flat (`deflate_medium`/`deflate_fast`, no filtered discard) |

One caveat about reproducing this: it requires row filtering to be enabled via `png_set_filter`. libpng selects its deflate strategy from the filter setting, so with `PNG_FILTER_NONE` it uses `Z_DEFAULT_STRATEGY` and this code path never runs. The existing `png_encode` benchmark hardcodes `PNG_FILTER_NONE`, which is why it shows no change; I measured with a variant that passes `PNG_ALL_FILTERS`, matching libpng's default write path where filtered rows select `Z_FILTERED`. Applications that leave libpng's defaults alone get the filtered path automatically.

As a cross-check, extracting the exact filtered scanline stream libpng feeds `deflate` and compressing it directly shows -48% at level 9, so the PNG residual data is more sensitive to the search floor than raw RGB, not less.

deflate_slow.c
nmoinvaz: Fixed.

Dead2:
The comments added in this PR are hard to read, perhaps they can be reworded better?
A comment about the prev_length juggling around the longest_match call would also be good.

nmoinvaz:
> The comments added in this PR are hard to read, perhaps they can be reworded better?

Done

> A comment about the prev_length juggling around the longest_match call would also be good.

Done

mtl1979:
Needs rebase to pick coveralls fixes.

nmoinvaz:
Rebased as requested.

Dead2:
Needs rebase

nmoinvaz:
Rebased.

nmoinvaz:
Added `encode_compressible_filtered` to `benchmark_zlib_apps` so this is reproducible. It enables `PNG_ALL_FILTERS` on realistic RGB, matching libpng's default write path where filtered rows select `Z_FILTERED`.

Apple M5, 1024x1024 realistic RGB, 3 reps, rebased on current develop:

| `png_encode/encode_compressible_filtered` | base | PR | Δ time |
|---|---:|---:|---:|
| level 7 | 106.5 ms | 84.5 ms | -20.7% |
| level 8 | 106.1 ms | 85.3 ms | -19.6% |
| level 9 | 171.2 ms | 91.8 ms | -46.4% |

Levels 7 and 8 gain more than before because the `longest_match` split on develop changed the baseline.

PR #2408, opened 2026-08-10
Cache pending index in local for put_short/put_uint32/put_uint64

nmoinvaz:
The `put_*` helpers read and write `s->pending` around a store to `s->pending_buf[]`. Since `pending_buf` is an `unsigned char *`, the compiler cannot prove the store does not alias `s->pending`, so every helper reloads the field after the buffer write. Caching `s->pending` in a local removes the reload.

On AArch64 (clang, Release) this removes one load per call in the `put_uint64` flush path used by `compress_block`, and trees.c.o drops from 1858 to 1819 instructions with loads of `s->pending` down from 77 to 40. In `zng_tr_stored_block` the LEN/NLEN pair now shares a single `pending += 4` update.

On x86-64 the memory-destination `addl` splits into `lea` + `mov`, so instruction count rises slightly, but each helper still performs one fewer load and no new stack spills appear.

`put_byte` is unchanged because its post-increment reads `s->pending` before the store, so it never had the reload.

### DIFF
--- a/deflate.h
+++ b/deflate.h
@@ -356,9 +356,10 @@ typedef enum {
  * IN assertion: there is enough room in pending_buf.
  */
 static inline void put_short(deflate_state *s, uint16_t w) {
+    uint32_t pending = s->pending;
     w = Z_U16_TO_LE(w);
-    zng_memwrite_2(&s->pending_buf[s->pending], w);
-    s->pending += 2;
+    zng_memwrite_2(&s->pending_buf[pending], w);
+    s->pending = pending + 2;
 }
 
 /* ===========================================================================
@@ -366,9 +367,10 @@ static inline void put_short(deflate_state *s, uint16_t w) {
  * IN assertion: there is enough room in pending_buf.
  */
 static inline void put_short_msb(deflate_state *s, uint16_t w) {
+    uint32_t pending = s->pending;
     w = Z_U16_TO_BE(w);
-    zng_memwrite_2(&s->pending_buf[s->pending], w);
-    s->pending += 2;
+    zng_memwrite_2(&s->pending_buf[pending], w);
+    s->pending = pending + 2;
 }
 
 /* ===========================================================================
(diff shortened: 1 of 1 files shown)


Dead2:
This costs 1 extra mov instruction with clang, but actually 2 extra mov instructions with gcc. On x86_64 that is.
So it doesn't seem like such a good idea for x86_64 unfortunately, not sure why it makes the optimization worse, but it does.

ref: https://godbolt.org/z/GzG3eKd43

nmoinvaz:
Even though there are more instructions, Claude is saying there is 1 less load which came from `add` instruction:

> `add DWORD PTR [rdi+20], 4` has one memory operand but performs two accesses — it reads [rdi+20], adds 4, and writes the result back. The CPU splits it into load + add + store μops internally."

Dead2:
It is AFAIK impossible to benchmark this, and outside of deflate header + footer creation, it is used very rarely. Making it very hard to argue that this has merit. 

I have tried to make zng_tr_flush_bits() and bi_windup() (the only two users of the put_ functions outside of headers) not use a bunch of ifs, I let Gemini create code that tests correctly and only has a single if and put_* call, and looks a lot better in godbolt.. but I haven't made a PR because I cannot measure a win at all, also I couldn't follow the logic enough that I was entirely certain the effects were logically identical.

nmoinvaz:
`put_uint64` is used in `send_bits` which is used very frequently, the other ones yes they are less used.

nmoinvaz:
I had Claude create a micro-benchmark which was able to detect it https://gist.github.com/nmoinvaz/158c34b3a9da9a55cc4c79f8b0e93e90, but I am unable to pick it up on any of our usual benchmarks. Lmk if you want me to close this or not.

Dead2:
@nmoinvaz I think we'll just drop this one, the risk for another spill on gcc is unwanted and the gain is not going to be noticeable anyway.

PR #2409, opened 2026-08-10
Pack frequency, depth and node into 64-bit heap entries in build_tree

nmoinvaz:
`build_tree`'s heap sift-down loads `tree[n].Freq` and `depth[n]` from two scattered arrays for every node it visits. Neither value changes while a node sits in the heap, so both can be packed above the node index in a single 64-bit heap entry. Each comparison becomes one contiguous load and one compare, and the `depth` array plus the `heap_len` state field disappear.

Comparing entries as `entry >> 10` reproduces the old `smaller()` ordering exactly, frequency first with a `<=` tie break on depth, and freq/depth wrap the same way the old `uint16_t`/`unsigned char` fields did. Compressed output is bit-identical, verified by size and checksum at levels 1/3/6/9 for both bulk and `Z_SYNC_FLUSH` chunked streams.

Deflating silesia.tar on Apple M5 (arm64, clang): -2 to -3% at level 6 bulk, -11 to -12% with a 4 KiB `Z_SYNC_FLUSH` cadence, where tree construction is ~70% of `zng_tr_flush_block` time. Level 1 is unaffected since `deflate_quick` never builds dynamic trees.

The packed queue is a 2.3 KiB stack array in `build_tree`, entries `1..heap_len`, while the sorted node output still goes to `s->heap` for `gen_bitlen`.

### DIFF
--- a/deflate.h
+++ b/deflate.h
@@ -259,7 +259,6 @@ struct ALIGNED_(64) internal_state {
     int32_t bi_valid;           /* Number of valid bits in bi_buf.
                                  * All bits above the last valid bit are always zero. */
 
-    int heap_len;               /* number of elements in the heap */
     int heap_max;               /* element of largest frequency */
 
                 /* used by trees.c: */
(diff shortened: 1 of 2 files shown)


nmoinvaz:
Superseded by #2410, which contains this change as its first commit plus the sorted two-queue construction built on top of it.

PR #2411, opened 2026-08-13
Fix minimum Windows API version for benchmarks under MinGW.

mtl1979:
* Fixes missing `SetProcessInformation`
* Don't redefine `NOMINMAX` under MinGW, it's already defined in `bits/os_defines.h`

### DIFF
--- a/test/benchmarks/benchmark_main.cc
+++ b/test/benchmarks/benchmark_main.cc
@@ -10,7 +10,9 @@
 #include <limits.h>
 
 #ifdef _WIN32
-#  define NOMINMAX
+#  ifndef NOMINMAX
+#    define NOMINMAX
+#  endif
 #  include <windows.h>
 #else
 #  include <unistd.h>
@@ -163,7 +165,7 @@ int main(int argc, char** argv) {
     }
 
     if (no_power_throttling) {
-#  ifdef PROCESS_POWER_THROTTLING_CURRENT_VERSION
+#  if _WIN32_WINNT >= _WIN32_WINNT_WIN10
         /* Clear the execution-speed bit to opt out of EcoQoS so the process runs at full
            clock on performance cores. */
         PROCESS_POWER_THROTTLING_STATE state;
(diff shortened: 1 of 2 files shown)


test/benchmarks/benchmark_main.cc
```diff
@@ -10,7 +10,12 @@
 #include <limits.h>
 
 #ifdef _WIN32
-#  define NOMINMAX
+#  ifdef __MINGW32__
+#    undef _WIN32_WINNT
+#    define _WIN32_WINNT _WIN32_WINNT_WIN10
+#  else
```
nmoinvaz:
```suggestion
#  endif
#  ifndef NOMINMAX
#    define NOMINMAX
#  endif
```

test/benchmarks/benchmark_main.cc
```diff
@@ -10,7 +10,12 @@
 #include <limits.h>
 
 #ifdef _WIN32
-#  define NOMINMAX
+#  ifdef __MINGW32__
+#    undef _WIN32_WINNT
+#    define _WIN32_WINNT _WIN32_WINNT_WIN10
```
nmoinvaz: Don't you think `_WIN32_WINNT` should be set by CMake on the project for MinGW instead of in the code?

nmoinvaz: Actually, I think it is better to gate usage of `SetProcessInformation` by this instead:
```diff
    if (no_power_throttling) {
+ #  if _WIN32_WINNT >= _WIN32_WINNT_WIN10
- #  ifdef PROCESS_POWER_THROTTLING_CURRENT_VERSION
        /* Clear the execution-speed bit to opt out of EcoQoS so the process runs at full
           clock on performance cores. */
```
...
```diff
#  else
-       fprintf(stderr, "warning: --benchmark_no_power_throttling requires a Windows 10 SDK\n");
+       fprintf(stderr, "warning: --benchmark_no_power_throttling requires targeting Windows 10 or later\n");
#  endif
```

mtl1979: > Don't you think `_WIN32_WINNT` should be set by CMake on the project for MinGW instead of in the code?

I thought about that but wanted to keep the patch minimalistic as this is the only file that needs functions not available in Windows 7 and we don't support targeting Windows 7 anymore...

Not forgetting that CMake hasn't supported detecting MinGW for long enough that it would be cleaner alternative.

Dead2 (approved):
LGTM

nmoinvaz (approved):

PR #2417, opened 2026-08-19
Deprecate the zng_adler32() and zng_crc32() functions

Dead2:
Deprecate the zng_adler32() and zng_crc32() functions.
Steer people towards the _z suffuxed functions that support size_t lengths.

### DIFF
--- a/test/example.c
+++ b/test/example.c
@@ -790,7 +790,7 @@ static void test_deflate_prime(unsigned char *compr, size_t comprLen, unsigned c
         CHECK_ERR(err, "deflate");
 
     /* Gzip uncompressed data crc32 */
-    crc = PREFIX(crc32)(0, (const uint8_t *)hello, (uint32_t)len);
+    crc = PREFIX(crc32_z)(0, (const uint8_t *)hello, (size_t)len);
     err = deflate_prime_32(&c_stream, crc);
     CHECK_ERR(err, "deflatePrime");
     /* Gzip uncompressed data length */
--- a/test/fuzz/fuzzer_checksum.c
+++ b/test/fuzz/fuzzer_checksum.c
@@ -7,10 +7,10 @@
 #endif
 
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t dataLen) {
-    uint32_t crc0 = PREFIX(crc32)(0L, NULL, 0);
+    uint32_t crc0 = PREFIX(crc32_z)(0L, NULL, 0);
     uint32_t crc1 = crc0;
     uint32_t crc2 = crc0;
-    uint32_t adler0 = PREFIX(adler32)(0L, NULL, 0);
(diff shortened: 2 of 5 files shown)


mtl1979:
This looks so much cleaner... Still waiting for CI to finish if there is any issues... Maybe in zlib-ng 3.x we can revisit dropping the `_z` prefix for the native API.

phprus:
> Maybe in zlib-ng 3.x we can revisit dropping the `_z` prefix for the native API.

First, convince all users to add the `_z` suffix, and then convince them to remove the `_z` suffix? I don't think that's a very good idea.
If the choice is made in favor of names with the `_z` suffix, I think they should remain in future versions.

mtl1979:
> > Maybe in zlib-ng 3.x we can revisit dropping the `_z` prefix for the native API.
> 
> First, convince all users to add the `_z` suffix, and then convince them to remove the `_z` suffix? I don't think that's a very good idea. If the choice is made in favor of names with the `_z` suffix, I think they should remain in future versions.

We need the `_z` suffix now to be able to have zlib compatibility without adding test in every use of the functions... When we switch major version number to 3, most likely Mark Adler is already dead and buried, so we don't need to care about zlib compatibility anymore... Then we can just use preprocessor macro to ease porting to the 3.x API... The file name would become different even on Windows, so no issue whatsoever for linker compatibility.

Dead2:
It'll never be truly safe to remove the _z suffix and use non-z name unfortunately.
This has no relation to the zlib-compat api btw, it is entirely about keeping compatibility with the zlib-ng ABI.

Application compiling using non-z: deprecated warning, change code.
Application compiling using using z: no change
Application built against zlib-ng 2.3 using non-z: works

With #2295 without SYMVER, an application built against a new version but loading an old version, would get ABI problems as application would use size_t but the function expects uint32_t.

Not worth the trouble it seems.

mtl1979:
> It'll never be truly safe to remove the _z suffix and use non-z name unfortunately.
> This has no relation to the zlib-compat api btw, it is entirely about keeping compatibility with the zlib-ng ABI.

Even with `SYMVER` the compatibility is only guaranteed when the major version is the same, this comes from the dynamic linker and how it does use the symbolic links to resolve latest version number. I'm not sure what year zlib-ng 3.0 will be released but I do assume it will be in quite distant future if it will even happen.

```
mika@DESKTOP-AUKPB81:/mnt/c/build/git/zlib-ng/build-linux$ ldd example
        linux-vdso.so.1 (0x00007ffd323f9000)
        libz-ng.so.2 => /mnt/c/build/git/zlib-ng/build-linux/libz-ng.so.2 (0x00007b3854bab000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007b3854800000)
        /lib64/ld-linux-x86-64.so.2 (0x00007b3854be1000)
```

nmoinvaz (approved):

PR #2419, opened 2026-08-21
[Windows] Pass export definition file to linker

mtl1979:
* Linker will compare list of exported symbols with the contents of the def file and abort if symbols are missing


Dead2 (approved):
LGTM

PR #2414, opened 2026-08-16
arch PIC32 XTENSA(ESP32) AVR(Arduino) for Linux RTOS boards

GermanAizek:
This commit also fix MCU arch not support keyword SHARED with dynamic linking

Example building on ARM Cortex M0 (armv6)

```
-- Using CMake version 3.31.6
-- ZLIB_HEADER_VERSION: 1.3.1
-- ZLIBNG_HEADER_VERSION: 2.3.90
CMake Deprecation Warning at zlib-ng/CMakeLists.txt:54 (message):
  ZLIB_ENABLE_TESTS is deprecated.  Please use BUILD_TESTING instead.


-- Arch detected: 'armv6'
-- Basearch of 'armv6' (32bit) has been detected as: 'arm'
-- Performing Test FNO_LTO_AVAILABLE
-- Performing Test FNO_LTO_AVAILABLE - Success
-- Looking for features.h
-- Looking for features.h - not found
-- ARM floating point arch not auto-detected
-- Looking for arm_acle.h
-- Looking for arm_acle.h - found
-- Looking for sys/auxv.h
-- Looking for sys/auxv.h - not found
-- Looking for sys/sdt.h
-- Looking for sys/sdt.h - not found
-- Looking for unistd.h
-- Looking for unistd.h - found
-- Looking for linux/auxvec.h
-- Looking for linux/auxvec.h - not found
-- Looking for asm/hwprobe.h
-- Looking for asm/hwprobe.h - not found
-- Looking for strerror
-- Looking for strerror - found
-- Looking for __cpuid_count
-- Looking for __cpuid_count - not found
-- Performing Test HAVE_CPUID_MS
-- Performing Test HAVE_CPUID_MS - Failed
-- Checking for sanitizer option
--  sanitizer not enabled
-- Performing Test HAVE_NO_INTERPOSITION
-- Performing Test HAVE_NO_INTERPOSITION - Success
-- Performing Test HAVE_ATTRIBUTE_VISIBILITY_HIDDEN
-- Performing Test HAVE_ATTRIBUTE_VISIBILITY_HIDDEN - Success
-- Performing Test HAVE_ATTRIBUTE_VISIBILITY_INTERNAL
-- Performing Test HAVE_ATTRIBUTE_VISIBILITY_INTERNAL - Success
-- Performing Test HAVE_ATTRIBUTE_ALIGNED
-- Performing Test HAVE_ATTRIBUTE_ALIGNED - Success
[...]

PR #2421, opened 2026-08-28
Don't build shared libraries, if target doesn't support them.

mtl1979:
Not all platforms support shared libraries, so we should disable building shared libraries, if CMake already knows it's not supported.

Separated and simplified from #2414.


Dead2 (approved):
LGTM

PR #2423, opened 2026-08-30
Fix import library name on non-MSVC Windows targets

### LINKED ISSUE #1864: Windows: zlib.pc linker flags do not match library name

mtl1979:
* pkgconfig expects import library to be named libz.dll.a or libz-ng.dll.a

Downstream packages already have renamed the import library.

Fixes #1864.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1367,6 +1367,9 @@ if(WIN32)
     # Shared library
     if(NOT DEFINED BUILD_SHARED_LIBS OR BUILD_SHARED_LIBS)
         set_target_properties(zlib-ng PROPERTIES OUTPUT_NAME zlib${SUFFIX})
+        if(NOT MSVC)
+            set_target_properties(zlib-ng PROPERTIES ARCHIVE_OUTPUT_NAME z${SUFFIX})
+        endif()
     endif()
     # Static library
     if(NOT DEFINED BUILD_SHARED_LIBS)


mtl1979:
Cygwin Win64 CI failure is due to mismatch in versions of `python-pip-wheel` and `python39-pip` packages. This has been reported to Cygwin mailing list.

mtl1979:
Rebased to pick the fix for Cygwin Win64 failure...

mtl1979:
Rebased to pick fix for MacOS Intel CI failure, even though it's irrelevant to this PR, but does cause missing coverage data.

nmoinvaz:
> Downstream packages already have renamed the import library.

I am ok with downstream packages having to update to the more compatible name.

mtl1979:
> > Downstream packages already have renamed the import library.
> 
> 
> 
> I am ok with downstream packages having to update to the more compatible name.

And it makes more sense for configure and CMake builds to use same file name for the import library...

Dead2 (approved):
LGTM

PR #2428, opened 2026-08-31
Exclude CHECK_* and EXPECT_* macros from branch coverage.

mtl1979:
* Google Test macros `CHECK_*` and `EXPECT_*` create branches that are expected to be never reached, so lower branch coverage by about 50%


Dead2 (approved):
LGTM

PR #2372, opened 2026-07-22
Split inflate_fast into safe and non-safe specializations

nmoinvaz:
`inflate_fast` took `safe_mode` as a runtime argument and tested it on the hot length/distance path every iteration. On x86-64 the argument did not even stay in a register: it was spilled and re-tested twice per loop iteration (`cmpl $0x0, -0x4c(%rbp)`).

`inffast_tpl.h` is now included twice per architecture, the second time with `USE_SAFE_MODE` defined, so the output space check and the choice of chunk copy routines fold away at compile time. The function table gains an `inflate_fast_safe` slot so `inflate()` and `inflateBack()` call the specialization they need directly instead of branching inside a shared entry point. `inflateBack()` always wanted the safe variant, and previously paid a frame setup, teardown and tail jump to reach it.

Codegen for the bulk decoder improves on every target:

| target | instructions | spill stores | reloads | out-of-line calls |
|---|---|---|---|---|
| x86 sse2 | 425 → 373 | 20 → 16 | 51 → 39 | 8 → 4 |
| x86 ssse3 | 425 → 373 | 20 → 16 | 51 → 39 | 8 → 4 |
| x86 avx2 | 399 → 358 | 20 → 17 | 49 → 45 | 8 → 5 |
| x86 avx512 | 446 → 392 | 16 → 17 | 54 → 41 | 5 → 4 |
| arm64 neon | 339 → 298 | 25 → 20 | 34 → 26 | 7 → 4 |

The safe variant improves sharply on avx512 (446 → 314 instructions, 37 → 24 conditional branches), which is the only variant defining `HAVE_MASKED_READWRITE` and so can use masked stores throughout. On the other targets it grows, because with the mode known at compile time the compiler inlines copy helpers it previously kept out of line. Spill stores drop there anyway.

The second commit builds on that. After the split, `chunkcopy_safe` is reachable only from the safe variant, so inlining it leaves the bulk decoder byte for byte identical (verified by hashing its disassembly) while shortening the path small-buffer callers take.

That inline is gated to ARM. On x86-64 the expansion is far larger and nearly doubles the safe variant, 507 to 990 instructions on sse2 and 482 to 979 on avx2, which measured as a net regression on a Skylake-derived part. avx512 is unaffected either way, since `HAVE_MASKED_READWRITE` routes its safe path through `CHUNKCOPY_SAFE` and it never calls `chunkcopy_safe`. With the gate in place the x86-64 safe variant is byte-identical to the split-only build.

[...]

PR #2432, opened 2026-09-01
Copy two chunks per iteration in the non-safe inflate fast path

nmoinvaz:
`DOUBLECHUNKCOPY` copies two chunks per iteration with both loads issued ahead of both stores, so the compiler pairs them into wider memory operations (`ldp`/`stp q` on AArch64, paired `movups`/`vmovups` on x86-64), and its two-chunk head finishes copies up to two chunks long without entering the loop. The call site guard tightens from `dist >= CHUNKSIZE()` to `dist >= 2 * CHUNKSIZE()` because the paired loads read one chunk beyond the store frontier; overlapped distances below that route to `CHUNKMEMSET`, whose single-chunk `CHUNKCOPY` fallback keeps them correct.

Decompressing the silesia corpus improves 4.0% geomean on Apple M5 (Apple clang 21, Release), with `xml` at -10% and no file regressing.

The second commit addresses the one composition the wide head can only hurt: blocks made entirely of short matches, where the extra chunk of load and store per match is pure cost (see the `dna` rows in the benchmark comments). Such blocks are detectable from the length table, no code for any symbol above 272 caps matches at 34 bytes, so a flag computed once per dynamic block dispatches them to a narrow-copy instantiation of `inflate_fast` that keeps the original single-chunk copy. The dispatch is a tail call on function entry, keeping both hot loops free of per-match branching, the same structure as the safe/non-safe split proposed in #2372. This recovers the `dna` regression entirely (-3% to -7%, back to develop levels) and improves `E.coli` about -1%, with long-match files unchanged.

`test_inflate_overlap.cc` roundtrips every repetition period around the 16, 32, and 64 byte chunk sizes, a case the suite previously never exercised even though relaxing the new guard by one chunk silently corrupts distances 16-31, plus a short-match stream that stays on the narrow variant.

x86-64 was verified at the codegen level only (SSE2 pairs into 32-byte, AVX2 into 64-byte copies, all inlined); runtime numbers there would be welcome.

### DIFF
--- a/test/test_inflate_overlap.cc
+++ b/test/test_inflate_overlap.cc
@@ -1,0 +1,74 @@
+/* test_inflate_overlap.cc - Test inflate of overlapped matches at distances near the chunk size */
+
+#include "zbuild.h"
+#ifdef ZLIB_COMPAT
+#  include "zlib.h"
+#else
+#  include "zlib-ng.h"
+#endif
+
+#include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <gtest/gtest.h>
+
+#define UNCOMPR_SIZE (64 * 1024)
+
+class inflate_overlap : public ::testing::TestWithParam<unsigned> {};
+
+TEST_P(inflate_overlap, roundtrip) {
+    unsigned period = GetParam();
+    int err;
+
+    uint8_t *uncompr = (uint8_t *)malloc(UNCOMPR_SIZE);
+    ASSERT_NE(uncompr, nullptr);
+    for (unsigned i = 0; i < period; i++)
+        uncompr[i] = (uint8_t)(i * 37 + 11);
+    for (unsigned i = period; i < UNCOMPR_SIZE; i++)
+        uncompr[i] = uncompr[i - period];
+
+    size_t compr_size = UNCOMPR_SIZE + 1024;
+    uint8_t *compr = (uint8_t *)malloc(compr_size);
+    ASSERT_NE(compr, nullptr);
+    uint8_t *decompr = (uint8_t *)malloc(UNCOMPR_SIZE);
+    ASSERT_NE(decompr, nullptr);
+
+    PREFIX3(stream) c_stream;
+    memset(&c_stream, 0, sizeof(c_stream));
+    err = PREFIX(deflateInit2)(&c_stream, Z_BEST_COMPRESSION, Z_DEFLATED, -MAX_WBITS,
+                               MAX_MEM_LEVEL, Z_DEFAULT_STRATEGY);
+    ASSERT_EQ(err, Z_OK);
+    c_stream.next_in = (z_const uint8_t *)uncompr;
+    c_stream.avail_in = UNCOMPR_SIZE;
+    c_stream.next_out = compr;
+    c_stream.avail_out = (uint32_t)compr_size;
+    err = PREFIX(deflate)(&c_stream, Z_FINISH);
+    ASSERT_EQ(err, Z_STREAM_END);
+    compr_size = c_stream.total_out;
+    err = PREFIX(deflateEnd)(&c_stream);
+    ASSERT_EQ(err, Z_OK);
+
+    PREFIX3(stream) d_stream;
+    memset(&d_stream, 0, sizeof(d_stream));
+    err = PREFIX(inflateInit2)(&d_stream, -MAX_WBITS);
+    ASSERT_EQ(err, Z_OK);
+    d_stream.next_in = compr;
+    d_stream.avail_in = (uint32_t)compr_size;
+    d_stream.next_out = decompr;
+    d_stream.avail_out = UNCOMPR_SIZE;
+    err = PREFIX(inflate)(&d_stream, Z_FINISH);
+    ASSERT_EQ(err, Z_STREAM_END);
+    ASSERT_EQ(d_stream.total_out, (z_uintmax_t)UNCOMPR_SIZE);
+    err = PREFIX(inflateEnd)(&d_stream);
+    ASSERT_EQ(err, Z_OK);
+
+    EXPECT_EQ(memcmp(uncompr, decompr, UNCOMPR_SIZE), 0);
+
+    free(uncompr);
+    free(compr);
+    free(decompr);
+}
+
+/* Cover every distance around the 16, 32, and 64 byte chunk sizes and their doubles */
+INSTANTIATE_TEST_SUITE_P(periods, inflate_overlap, ::testing::Range(1u, 131u));
--- a/test/test_inflate_overlap.cc
+++ b/test/test_inflate_overlap.cc
@@ -17,16 +17,9 @@
 
 class inflate_overlap : public ::testing::TestWithParam<unsigned> {};
 
-TEST_P(inflate_overlap, roundtrip) {
-    unsigned period = GetParam();
+/* Compress uncompr at level 9, decompress it, and compare; frees uncompr */
+static void roundtrip_verify(uint8_t *uncompr) {
     int err;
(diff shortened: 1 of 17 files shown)


test/test_inflate_overlap.cc
```diff
@@ -0,0 +1,74 @@
...
+
+    EXPECT_EQ(memcmp(uncompr, decompr, UNCOMPR_SIZE), 0);
+
+    free(uncompr);
+    free(compr);
+    free(decompr);
+}
+
+/* Cover every distance around the 16, 32, and 64 byte chunk sizes and their doubles */
+INSTANTIATE_TEST_SUITE_P(periods, inflate_overlap, ::testing::Range(1u, 131u));
```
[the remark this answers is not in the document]
mtl1979: Yet another facepalm comment by a bot that doesn't recognize third-party templates.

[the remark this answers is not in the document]
mtl1979: Yet another facepalm comment by a bot that doesn't recognize third-party templates.

nmoinvaz:
**Correction:** the first version of this comment accidentally compared two head builds (a stale base build directory had picked up the PR sources), which made everything look flat. Corrected `inflate_bench/nocrc` head vs base below, Apple M5 (Apple clang 21, Release); every row over ±2% was re-measured with isolated runs, alternating binaries.

| data type | 16384 | 131072 | 1048576 |
|---|---|---|---|
| text | +0.4% | +2.5% | -8.0% |
| dna | +3.8% | +10.8% | +2.8% |
| literals | -0.0% | +0.6% | -0.5% |
| mixed | +0.3% | +0.1% | -3.2% |
| random | -0.9% | -0.5% | +0.4% |
| realistic_rgb | +0.9% | -0.8% | -0.6% |
| short_match | +1.1% | +1.9% | -3.2% |
| striped_rgb | +1.1% | +0.9% | +0.6% |

Small `text` sizes: 1 B +1.1%, 64 B +0.2%, 1024 B -0.1%.

The 1 MiB rows show the expected wins, `text` at -8.0%. `dna` is a real regression: its 4-letter alphabet compresses into very short matches, which now issue two chunk loads and stores in the copy head where they previously issued one. A follow-up for that band is being investigated.

test/test_inflate_overlap.cc
```diff
@@ -0,0 +1,74 @@
...
+#define UNCOMPR_SIZE (64 * 1024)
+
+class inflate_overlap : public ::testing::TestWithParam<unsigned> {};
+
+TEST_P(inflate_overlap, roundtrip) {
+    unsigned period = GetParam();
+    int err;
+
+    uint8_t *uncompr = (uint8_t *)malloc(UNCOMPR_SIZE);
+    ASSERT_NE(uncompr, nullptr);
```
mtl1979: `ASSERT_NE` doesn't free `uncompr` causing static analysis to complain about potential memory leak. As such `ASSERT_NE` shouldn't be used for `NULL` checks for dynamically allocated local variables.

This is known issue with Google Test as it will use external function for the comparison result, which confuses static analysis, as it will not know if the macro will fall through when comparison result is `not equal`.

See: https://github.com/google/googletest/issues/3800

mtl1979 (changes requested):
Fails static analysis CI.

nmoinvaz:
`inflate_corpora/silesia` head vs base, same machine, each file measured with an isolated sustained run per binary, alternating binaries, stability controls at both ends. The fixture pre-compresses with level 9 internally, so both sides inflate identical streams.

| File | Δ time | File | Δ time |
|---|---|---|---|
| xml | -9.6% | x-ray | -3.7% |
| webster | -6.9% | mozilla | -3.2% |
| nci | -6.6% | ooffice | -1.7% |
| samba | -6.2% | mr | -1.1% |
| reymont | -5.8% | sao | -0.3% |
| osdb | -3.8% | dickens | +1.0% |

-4.0% geomean. The gains track average match length, `xml` and `nci` carry the longest matches in the corpus.

nmoinvaz:
TLDR: `dna` data type regression in the table above is real with just replacing `CHUNKCOPY` in `inflate_fast` with `DOUBLECHUNKCOPY`...

> Closing the loop on the `dna` regression above: infgen on those streams shows 100% of matches are 16 bytes or shorter at one match per ~8 output bytes, so every match paid the wide head's second chunk of load and store with nothing to gain (the +10.8% peak at 131072 is the working set sitting on the L1D boundary; real genomic data like `E.coli` has the same shape but is L2-resident and measured noise-level).

TLDR: I tried a bunch of shit and nothing worked..

> Per-match mitigations all measured worse than the regression itself on Apple M5: a `len <= 16` early-out branch mispredicts on irregular streams (xml +7.2%), a distance-keyed split mispredicts on mixed-distance streams (short_match +27%), and even a perfectly-predictable per-block flag read in the loop costs exactly what the narrow copy saves.

TLDR: Having a `inflate_fast` variant that does `DOUBLECOPYCHUNK` is the key..

> The second commit instead dispatches short-match blocks (no length code above symbol 272, so no match over 34 bytes) to a separate narrow-copy instantiation of `inflate_fast`, selected by tail call on entry, keeping both hot loops free of per-match dispatch. Measured on Apple M5: `dna` -3% to -7% (fully back to develop levels), `E.coli` about -1%, `short_match` -0.7%, long-match files unchanged within noise.

test/test_inflate_overlap.cc
```diff
@@ -0,0 +1,96 @@
...
+
+class inflate_overlap : public ::testing::TestWithParam<unsigned> {};
+
+/* Compress uncompr at level 9, decompress it, and compare; frees uncompr */
+static void roundtrip_verify(uint8_t *uncompr) {
+    int err;
+
+    size_t compr_size = UNCOMPR_SIZE + 1024;
+    uint8_t *compr = (uint8_t *)malloc(compr_size);
+    ASSERT_NE(compr, nullptr);
```
mtl1979, line 26: Still hits a bug in Google Test.

test/test_inflate_overlap.cc
```diff
@@ -0,0 +1,96 @@
...
+    uint8_t *decompr = (uint8_t *)malloc(UNCOMPR_SIZE);
+    ASSERT_NE(decompr, nullptr);
```
mtl1979, line 28: Same here...

test/test_inflate_overlap.cc
```diff
@@ -0,0 +1,96 @@
...
+    free(uncompr);
+    free(compr);
+    free(decompr);
+}
+
+TEST_P(inflate_overlap, roundtrip) {
+    unsigned period = GetParam();
+
+    uint8_t *uncompr = (uint8_t *)malloc(UNCOMPR_SIZE);
+    ASSERT_NE(uncompr, nullptr);
```
mtl1979, line 70: Same here...

test/test_inflate_overlap.cc
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/* Cover every distance around the 16, 32, and 64 byte chunk sizes and their doubles */
+INSTANTIATE_TEST_SUITE_P(periods, inflate_overlap, ::testing::Range(1u, 131u));
+
+/* Repeating 16-byte runs broken up by unique bytes keep every match under 35
+   bytes, so all blocks decode through the narrow-copy fast loop variant */
+TEST(inflate_overlap_narrow, roundtrip) {
+    uint8_t *uncompr = (uint8_t *)malloc(UNCOMPR_SIZE);
+    ASSERT_NE(uncompr, nullptr);
```
mtl1979, line 86: And here...

nmoinvaz:
I will wait on #2372 before picking this up again.

PR #2434, opened 2026-09-04
Use a branchless byte-granular bit writer in compress_block

nmoinvaz:
`send_bits` flushes when the accumulated bits cross the 64-bit boundary, a data-dependent branch that mispredicts constantly with variable-length codes. The new `send_bits_flush` stores the bit buffer unconditionally once per symbol and consumes only the complete bytes, so the flush needs no branch. This is the byte-granular writer pattern used by libdeflate and by igzip's assembly emitters. The literal batch drops from four codes to three since a merge must fit alongside the flush residue in 64 bits, and `zng_emit_dist` is split into a `zng_assemble_dist` helper so the match bit assembly can be shared by the merge path.

The unconditional store reaches up to 7 bytes beyond the output position, so the symbol buffers move up by `PENDING_BUF_PAD` bytes and the pending allocation grows by the same amount.

Output is bit-identical across levels, strategies and flush patterns. On Apple M5, deflate runs 2-5% faster at levels 3 through 9 (geomean -2.7%), `Z_HUFFMAN_ONLY` 13-25% and `Z_RLE` 13-20%. Level 1 is unaffected since `deflate_quick` does not use `compress_block`. x86-64 has not been measured yet.

Full deflate, `deflate_corpora` benchmarks, Apple M5 (macOS 26.6, Apple clang), 5 repetitions, CPU time medians:

| benchmark | level 3 | level 6 | level 9 |
|---|---|---|---|
| silesia/dickens | -3.1% | -2.9% | -2.2% |
| silesia/osdb | -3.6% | -2.9% | -3.7% |
| silesia/xml | -4.3% | -2.0% | -0.5% |
| tars/silesia.tar | | -2.2% | |
| tars/silesia.tar level 1 (control) | | -0.3% | |

Geomean across the set is -2.7%, compressed size delta is exactly 0 bytes on every row.

Strategies where `compress_block` dominates, one-shot deflate at level 6, interleaved A/B, medians of 15:

| strategy | dickens | osdb | x-ray |
|---|---|---|---|
| `Z_HUFFMAN_ONLY` | -25.1% | -15.7% | -13.2% |
| `Z_RLE` | -20.4% | -13.3% | -14.8% |

Static instruction check of `compress_block`, same compiler both sides (LLVM clang 22.1.5, `-O2`):

| | instructions | cond. branches | loads | stores | stack refs |
|---|---|---|---|---|---|
| arm64 develop | 149 | 10 | 34 | 9 | 6 |
| arm64 PR | 116 (-22%) | 7 | 28 | 8 | 4 |
| x86-64 develop | 184 | 10 | 43 | 8 | 6 |
| x86-64 PR | 155 (-16%) | 7 | 38 | 9 | 7 |

Both architectures lose the same three conditional branches (the `send_bits` boundary branch at each emit site) with no spill growth; stores are per-symbol dynamically but fewer static sites.

### DIFF
--- a/trees_emit.h
+++ b/trees_emit.h
@@ -63,6 +63,27 @@ extern Z_INTERNAL const uint32_t dmask_extra[D_CODES];
     bi_valid = total_bits;\
 } while (0)
 
+/* Merge bits into the bit buffer without flushing. The caller must guarantee room,
+ * i.e. bi_valid + len <= 64. Pair with send_bits_flush, which leaves at most 7 bits. */
+#define send_bits_merge(s, t_val, t_len, bi_buf, bi_valid) do {\
+    Assert((bi_valid) + (uint32_t)(t_len) <= 64, "bit buffer overflow");\
+    trace_bits(s, (uint64_t)(t_val), (uint32_t)(t_len));\
+    sent_bits_add(s, t_len);\
+    bi_buf |= (uint64_t)(t_val) << (bi_valid);\
+    bi_valid += (uint32_t)(t_len);\
+} while (0)
+
+/* Store the bit buffer unconditionally and consume its complete bytes, leaving at most
+ * 7 valid bits. The unconditional 8-byte store reaches up to 7 bytes past the logical
+ * output position, which the pending buffer layout must reserve room for
+ * (see PENDING_BUF_PAD in deflate.c). */
+#define send_bits_flush(s, bi_buf, bi_valid, pending) do {\
+    zng_memwrite_8((s)->pending_buf + (pending), Z_U64_TO_LE(bi_buf));\
+    pending += (bi_valid) >> 3;\
+    bi_buf >>= ((bi_valid) & ~7U);\
+    bi_valid &= 7;\
+} while (0)
+
 /* Send a code of the given tree. c and tree must not have side effects */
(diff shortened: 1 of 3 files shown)


trees_emit.h
```diff
@@ -157,6 +190,19 @@ static inline uint32_t zng_emit_dist(deflate_state *s, const ct_data *ltree, con
     match_bits |= ((uint64_t)(dist & mask_ext) << match_bits_len);
     match_bits_len += extra;
 
+    Z_UNUSED(s);
```
mtl1979, line 193: Seems like `deflate_state *s` isn't needed even in debug build, so it should be removed as function parameter completely... Not even `trace_code` macro need it.

mtl1979 (commented):
Some cleanup needed...

Something seems off in the x86-64 statistics... Might need actual assembler dump to see what is going on...

trees_emit.h
nmoinvaz: Good catch, removed the parameter from `zng_assemble_dist` and from `trace_code` in ee66decb.

mtl1979: I try to check if `Z_UNUSED` is actually needed when it is added... Sometimes it's only needed for `Release` builds as some variables (or function parameters) are only used in `Debug` builds.

PR #2436, opened 2026-09-05
Skip redundant hash clearing and window update for stored streams

nmoinvaz:
Level 0 one-shot compression paid two costs that are pure overhead for stored streams.

Every `deflateReset()` cleared the hash table, a 128KB memset that level 0, `Z_HUFFMAN_ONLY`, and `Z_RLE` never read. `deflateParams()` already clears the table when a stream switches from a hashless to a hash-using configuration, so `lm_init()` now skips the clear for hashless configurations.

`deflate_stored()` also copied the last window-size bytes into the sliding window even when writing the final block. A finished stream can only be reset or ended, so that history is never read. The update is now skipped when the last block has been written, saving a 32KB copy per one-shot compression.

Compressed output is byte-identical. Measured on Apple M5, macOS 26.6, Apple clang 21, release build, raw one-shot `deflate` at level 0, medians of 5 reps:

| Input | Size | develop | PR | Δ time |
|---|---|---|---|---|
| calgary/obj1 | 21KB | 2218ns | 498ns | -77.6% |
| snappy/html | 102KB | 4595ns | 1841ns | -59.9% |
| synthetic text | 128KB | 5294ns | 2511ns | -52.6% |
| large/E.coli | 4.6MB | 119.2µs | 119.5µs | +0.2% |
| silesia.tar | 212MB | 6.76ms | 6.73ms | -0.6% |

At 128KB this closes the level 0 gap to libdeflate (~2664ns for the same input on this machine), which has no per-stream hash or window state to maintain. Larger inputs are memcpy-bound and unaffected.

For cross-implementation context, an isolated run of `silesia.tar` level 0 (7 reps, medians) puts all zlib-family implementations in one noise band: zlib-ng (this PR) 7.52ms, madler zlib 7.21ms, zlib-rs 7.59ms, libdeflate 7.08ms, chromium zlib 7.04ms, with run-to-run ordering shuffling within ±5%.

### DIFF
--- a/deflate.c
+++ b/deflate.c
@@ -456,7 +456,8 @@ int32_t Z_EXPORT PREFIX(deflateSetDictionary)(PREFIX3(stream) *strm, const uint8
     /* if dictionary would fill window, just replace the history */
     if (dictLength >= s->w_size) {
         if (wrap == 0) {            /* already empty otherwise */
-            CLEAR_HASH(s);
+            if (!hashless)
+                CLEAR_HASH(s);
             s->strstart = 0;
             s->block_start = 0;
             s->insert = 0;
@@ -1201,7 +1202,9 @@ static void lm_set_level(deflate_state *s, int level) {
 static void lm_init(deflate_state *s) {
     s->window_size = 2 * s->w_size;
 
-    CLEAR_HASH(s);
+    if (!HASHLESS(s->level, s->strategy)) {
+        CLEAR_HASH(s);
+    }
 
     /* Set the default configuration parameters:
      */
--- a/deflate.c
+++ b/deflate.c
@@ -158,6 +158,12 @@ static const config configuration_table[10] = {
     memset((unsigned char *)s->head, 0, HASH_SIZE * sizeof(*s->head)); \
   } while (0)
 
+/* Configurations that never read the hash table. When deflateParams() switches
+ * to a configuration that does read it, it rebuilds the table from the window.
+ */
+#define HASHLESS(level, strategy) \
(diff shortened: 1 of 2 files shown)


nmoinvaz:
After this change we are a bit faster than `libdeflate` in my silesia benchmarks for level 0.

deflate.c
```diff
@@ -1192,7 +1193,20 @@ static void lm_set_level(deflate_state *s, int level) {
...
+     */
+    if (s->level != 0 && s->strategy != Z_HUFFMAN_ONLY && s->strategy != Z_RLE) {
+        CLEAR_HASH(s);
+    }
+#ifdef Z_MEMORY_SANITIZER
+    else {
+        /* fill_window() still chains inserts through the table, so tell MSan
+         * the garbage reads are intentional.
+         */
+        __msan_unpoison(s->head, HASH_SIZE * sizeof(*s->head));
```
mtl1979: Shouldn't avoiding unnecessary reads be faster even when sanitizers are not used?

mtl1979 (commented):
Seems like something is missing in the logic...

deflate.c
nmoinvaz: Good point let me research.

mtl1979: Seems like `deflate_rle` and `deflate_huff` should use trimmed down version of `fill_window` with no hash table access at all...

nmoinvaz: I have added a check

deflate.c
mtl1979, line 466: No test coverage for large dictionary, so CI can't test if this block is correct.

nmoinvaz: done

Dead2 (approved):
LGTM

Issue #883, opened 2021-03-17
Problems with integration zlib-ng to lzbench


data-man:
Hello!

I'm trying to integrate `zlib-ng` to `lzbench` in my https://github.com/inikep/lzbench/pull/100 PR.
And got this error:
>In file included from _lzbench/compressors.cpp:1526:
././zlib-ng/zlib-ng.h:1076:20: error: conflicting declaration ‘typedef uint32_t (* in_func)(void*, const uint8_t**)’
 1076 | typedef uint32_t (*in_func) (void *, const uint8_t * *);
      |                    ^~~~~~~
In file included from _lzbench/compressors.cpp:1502:
././zlib/zlib.h:1092:20: note: previous declaration as ‘typedef unsigned int (* in_func)(void*, unsigned char**)’
 1092 | typedef unsigned (*in_func) OF((void FAR *,
      |                    ^~~~~~~

Any help is very appreciated!

mtl1979:
You're mixing headers from stock zlib and zlib-ng... It won't work...

Dead2:
The problem is caused by including both zlib.h and zlib-ng.h in the same compilation unit, and both of them typedef the above types.

Maybe we need to prefix the names of `in_func` and `out_func`.
I am uncertain what that would mean for users though, do they use those names?

mtl1979:
Those two are callback prototypes... they are not used directly, but as types for callback functions in function signatures...

data-man:
> Maybe we need to prefix the names of in_func and out_func.

Yes, please!

Dead2:
@mtl1979 So it would not be considered an API change?

mtl1979:
@Dead2 It definitely is API change... Even if we fix these two callbacks, there is still quite a few internal functions that cause conflicts when both zlib and zlib-ng are used in same compilation unit.

Dead2:
Yeah, that is what I thought.
So if we do want to change this, it would have to be for the next stable branch, not the 2.0 one.

We should make a complete list so we can decide whether it is worth changing.

A workaround for this case would be either don't compile zlib and zlib-ng at the same time, or you could try moving the zlib-ng include and wrappers into a separate compile unit (.c file)

mtl1979:
@Dead2 I don't think this is the right place for this kind of discussion, but as I mentioned earlier, we shouldn't use the wrapper macros in externally visible headers. This means only functions can have prefixes, not variable types, including callback prototypes.

[...]

PR #2437, opened 2026-09-06
Prefix in_func and out_func typedefs in the zlib-ng native API

nmoinvaz:
Renames the `inflateBack` callback typedefs in `zlib-ng.h` to `zng_in_func` and `zng_out_func`. They were the last unprefixed type names in the native header, so a translation unit including both `zlib-ng.h` and stock `zlib.h` failed on a conflicting `in_func` declaration (zlib-ng adds `const` to the input pointer). This was previously discussed in #883, where the workaround was moving zlib-ng calls into a separate compile unit.

The old `in_func` and `out_func` names remain as `Z_DEPRECATED` aliases, so existing native-API callers keep compiling with a deprecation warning until the aliases are removed in a later release. The aliases are skipped when stock `zlib.h` was included first.

With the rename, both headers compile together when `zlib.h` is included first, leaving only benign `z_const`, `z_off64_t`, and `Z_NULL` macro-redefinition warnings. Including `zlib-ng.h` first still conflicts on the aliases until they are removed. The remaining conflict is `struct gzFile_s`, which only appears when the consumer defines `WITH_GZFILEOP` and is left alone here since fixing it would cascade into renaming the public `gzFile` typedef.

The compat header keeps the standard unprefixed names, `infback.c` selects between them with `PREFIX()`, and the native name-mangling entries follow the rename. Verified native, compat, and `ZLIB_SYMBOL_PREFIX` builds.


mtl1979:
Making it possible to include both zlib.h and zlib-ng.h in same translation unit is almost lowest in our priority list. However we can introduce breaking API changes in zlib-ng 2.4.0 as it was already decided.

Dead2:
I'm not sure, this would complicate supporting multiple versions of zlib-ng.
Is there any way to keep backwards compatibility to at least one version before we remove it completely?
It'd have to have some kind of warning for using the old naming though.

nmoinvaz:
@Dead2 I've added `Z_DEPRECATED`.

nmoinvaz:
Added `in_func` and `out_func` back as `Z_DEPRECATED` aliases of `zng_in_func` and `zng_out_func`. Existing callers keep compiling against 2.4 with a deprecation warning, and we can drop the aliases in the release after.

Code that needs to build against several zlib-ng releases without the warning can bridge the gap on its side:

```c
#if ZLIBNG_VERNUM < 0x02040000
typedef in_func  zng_in_func;
typedef out_func zng_out_func;
#endif
```

Develop snapshots still report 2.3.90, so this only holds once the version is bumped.

The aliases are skipped when stock `zlib.h` was included first, so that include order still works in one translation unit.

Dead2 (approved):
LGTM

PR #2438, opened 2026-09-07
Add define for minimum rolling hash level

nmoinvaz:
Every place that switches between the rolling three-byte hash and the four-byte Knuth hash now reads `MIN_ROLL_LEVEL`, instead of a scattered level 9 comparison.


mtl1979:
Nobody likes "magic numbers" with no explanation what that number is supposed to mean, so replacing magic numbers with more descriptive preprocessor macro(s) make the code more readable.

Dead2:
I suspect we'd need an assert if this is set to a level that is not mapped to deflate_slow. If I remember correctly, other levels assume knuth.

Dead2:
Actually the assert won't fire on release builds..
So maybe what we really need is a preprocessor check that fails the build, otherwise this could end up biting someone silently.

nmoinvaz:
I'm not sure there is a good way to do it, what did you have in mind? I think Assert is rather clean way of doing it, except it is not in Release which is ok, idk why you'd want it in Release.

mtl1979:
> Actually the assert won't fire on release builds.. So maybe what we really need is a preprocessor check that fails the build, otherwise this could end up biting someone silently.

Preprocessor tests can't access struct members, only way we can test this "misconfiguration" is using second `if` condition, which would only fail at run-time, but would also gracefully revert back to Knuth hash.

Dead2:
A preprocessor check could check that level is within 7-9 though, if someone changes the struct they can change that limiter too.
But I think it can be mitigated with a good warning that it can only be used with deflate_slow, placed in the comment where the setting is defined.
The assert can be removed, it won't catch the cases where users carelessly experiment since they are not likely to compile debug builds anyway.

nmoinvaz:
Removed the assert. The `MIN_ROLL_LEVEL` comment now warns that every level at and above it must map to `deflate_slow`, and a preprocessor check fails the build if it drops below 7.

nmoinvaz:
Made a few more wording changes.

Dead2 (approved):
LGTM

PR #2440, opened 2026-09-14
Mark crc32_copy_chorba and adler32_msa as Z_INTERNAL

nmoinvaz:
Both definitions were missing the `Z_INTERNAL` attribute that every other architecture-specific function carries, so they were compiled with default visibility.

When the static library is linked into a shared object, they show up in its dynamic symbol table next to the public API. Verified on arm64 Linux by linking `libz-ng.a` whole into a shared library. Before the change `crc32_copy_chorba` was exported, after it the only non-API exports are `_init` and `_fini`.

`adler32_msa` is the same omission in the MIPS MSA code and is fixed the same way.


Dead2 (approved):
LGTM

PR #2441, opened 2026-09-14
Add zng_crc32_combine_gen and zng_crc32_combine_op to native def

nmoinvaz:
Both functions are declared in `zlib-ng.h` and listed in `zlib-ng.map.in`, but were never added to `win32/zlib-ng.def.in`, so MinGW builds using the def file did not export them.

The compat def files and both name mangling headers were updated for zlib 1.2.12 in 445284c5, and the native def was missed at the same time.


Dead2 (approved):
LGTM

